PR #42, opened 2013-10-30
Add autotools spew to .gitignore

amluto:
This removes all of the garbage I see in 'git status' output.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,32 @@
 # setup-ekam.sh
 /c++/.ekam
 
+# Autotools build artifacts.
+/c++/**/*.o
+/c++/**/*.lo
+/c++/**/*.la
+/c++/**/.dirstamp
+/c++/**/.deps/
+/c++/.libs/
+/c++/m4/
+/c++/stamp-h1
+/c++/config.h.in
+/c++/config.h
+/c++/config.log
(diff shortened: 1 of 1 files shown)


KubaO:
Is there still something that capnproto needs that autotools supports that cmake doesn't? That would make autotools scripts redundant...

kentonv:
@KubaO The cmake build exists because some contributors whose own projects use cmake wanted it badly enough that they contributed cmake files to Cap'n Proto. It then proved useful as well for supporting MSVC. However, I personally don't know cmake, and every time I look at it I become more convinced that I don't want to know it.

Autotools is ugly but it has long been the standard build system for C/C++ projects across the open source ecosystem. Everyone knows what to do when they see that a project includes a `configure` script. A lot of people even immediately know how to install, customize compiler flags, cross-compile, build a distribution tarball, etc.

If we're going to decide that that familiarity isn't important, then cmake is not the build system I'd choose next. My preferred build system -- the one which I use 99% of the time to build Cap'n Proto -- is [Ekam](https://github.com/capnproto/ekam), though admittedly it's probably still too rough to be a serious option. Otherwise I'd probably look into Bazel.

But I'm not interested in making such a change any time soon. We'll be sticking with autotools for now.

PR #55, opened 2013-12-13
Fixes for getting emscripten to build

jparyani:
The changes to get emscripten working with Cap'n Proto are relatively small. The biggest change had to occur in exception.c++ because emscripten doesn't have __cxa_get_globals

### DIFF
--- a/c++/src/kj/async.c++
+++ b/c++/src/kj/async.c++
@@ -35,7 +35,7 @@
 
 #if !KJ_NO_RTTI
 #include <typeinfo>
-#if __GNUC__
+#if __GNUC__ && !EMSCRIPTEN
 #include <cxxabi.h>
 #include <stdlib.h>
 #endif
@@ -419,7 +419,7 @@ _::PromiseNode* Event::getInnerForTrace() {
 }
 
 #if !KJ_NO_RTTI
-#if __GNUC__
+#if __GNUC__ && !EMSCRIPTEN
 static kj::String demangleTypeName(const char* name) {
   int status;
   char* buf = abi::__cxa_demangle(name, nullptr, nullptr, &status);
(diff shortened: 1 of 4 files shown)


mlogan:
+1

For what's it's worth, I'd like to use capnproto in emscripten.

mlogan:
(Also, I have a rebased version of these changes at https://github.com/artillery/capnproto/tree/emscripten)

jplatte:
I had subscribed to this somewhen, just had a look at it again because I saw the `EMSCRIPTEN` macro being used in one of the old commits. It seems that's in the latest version still, so that should be fixed. `EMSCRIPTEN` shouldn't be used anymore, there is `__EMSCRIPTEN__` that replaces it.

kentonv:
If someone would like to send an updated pull request for emscripten support, I'd be happy to consider it.

I'd be interested to know if Cap'n Proto actually compiles to a reasonable amount of code under emscripten/asm.js. Probably some of the "always inline" attributes in layout.c++ aren't the correct trade-off in an emscripten use case; perhaps you'll want to adjust those.

Note also that there is [capnp-js](https://github.com/capnp-js/serialization) these days.

mlogan:
So I actually went with capnp solely because of the smaller build size increase (this is for a game, so serialization/deserialization overhead is peanuts compared to everything else). Protobufs added about 1.5MB of optimized minified JS - capnp was more like 100-200KB.

I'll open a new pull with @jparyani's changes and mine once I have something working on it.

kentonv:
Good to know the Emscripten compiler is good at removing parts of the library you don't use. I'm guessing it chucked out the dynamic and schema APIs, otherwise 200KB seems way too small. :)

FWIW, you might get similar results with protobufs using lite mode (`option optimize_for = LITE_RUNTIME`).

kentonv:
@unicomp21 Well, this PR is almost 7 years old. Both Cap'n Proto and Emscripten have changed a lot in that time, so probably this needs to be started over from scratch.

PR #91, opened 2014-05-28
Added eventloop integration with CFRunLoop (OSX/iOS)

jjwchoy:
This is mostly copied from node-capnproto and async-unix. This commit enables integration with Cocoa's event loop (CFRunLoop).

Notes:
- I haven't committed any tests yet (however I have tried the AsyncIo::SimpleNetwork  test with it and it works both when using `wait`, and running CFRunLoop normally)
- I haven't modified the makefiles (requires `-framework CoreFoundation` to be added to LDFLAGS for libkj-async)
- I've had to add a temporary `AsyncCocoaIoContext` struct as `AsyncIoContext` currently has a unixEventPort field

Is this something you would merge into kj? It would be useful for anyone that wants to integrate kj/capnproto into an OSX/iOS application, and for an Objective-C wrapper of capnproto.

### DIFF
--- a/c++/src/kj/async-cocoa.c++
+++ b/c++/src/kj/async-cocoa.c++
@@ -1,0 +1,496 @@
+// Copyright (c) 2014, Jason Choy <jjwchoy@gmail.com>
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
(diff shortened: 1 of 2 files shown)


jjwchoy:
Opened a new pull request for this: https://github.com/sandstorm-io/capnproto/pull/188

PR #166, opened 2015-01-05
Prevent accidental bool conversion of AutoCloseFd.

mortehu:
Without this change, the following code compiles:

    kj::AutoCloseFd fd;
    if (!fd) return;

In my opinion, this is a bit dangerous, although I could easily agree that it's too much of a corner case to waste header space on.

### DIFF
--- a/c++/src/kj/io.h
+++ b/c++/src/kj/io.h
@@ -257,6 +257,10 @@ public:
   inline operator int() const { return fd; }
   inline int get() const { return fd; }
 
+  operator bool() const = delete;
+  // Deleting this operator prevents accidental use in boolean contexts, which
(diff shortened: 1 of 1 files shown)


joshuawarner32:
What about implementing an `explicit operator bool` that compares against -1?  That would make the most sense to me.

mortehu:
I agree, but this library uses comparison to nullptr in several places, and bool conversion operators in none, so I thought the desire was to keep it that way.

kentonv:
Oh wow, I actually didn't know you could delete operator bool. That's pretty useful.

And yeah, I prefer explicit comparison to null rather than implicit truth values.

Thanks!

PR #168, opened 2015-01-12
added syntax highlighting for emacs

netguy204:
Created an emacs major mode to support capnproto syntax highlighting.

### DIFF
--- a/highlighting/emacs/capnp-mode.el
+++ b/highlighting/emacs/capnp-mode.el
@@ -1,0 +1,75 @@
+;;; capnp-mode.el --- major mode for editing Capn' Proto Files
+
+;; This is free and unencumbered software released into the public domain.
+
+;; Author: Brian Taylor <el.wubo@gmail.com>
+;; Version: 1.0.0
+
+;;; Commentary:
+
+;; Provides basic syntax highlighting for capnp files.
+;;
+;; To use:
+;;
+;; Add something like this to your .emacs file:
+;;
+;; (add-to-list 'load-path "~/src/capnproto/highlighting/emacs")
+;; (require 'capnp-mode)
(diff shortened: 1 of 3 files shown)


dwrensha:
Nice! I would definitely use this.

I'll nitpick that some of the highlighting is not completely accurate -- e.g. if `interface` is used as a method parameter as it is [here](https://github.com/sandstorm-io/sandstorm/blob/cc598730b2735dab0438dfbf8e9e2c249b966853/src/sandstorm/hack-session.capnp#L118), then it will still be highlighted as a keyword. But that's a minor point, and getting it right would probably take a lot more work.

kentonv:
Thanks, Brian!

FWIW, my own syntax highlighting for qtcreator has the same problem with the contextual keywords. :)

Since I don't know emacs at all I'll go on David's review here. David, let me know when I should merge this.

Also, Brian, feel free to add yourself to `CONTRIBUTORS`.

highlighting/emacs/capnp-mode.el
```diff
@@ -0,0 +1,75 @@
...
+;;;###autoload
+(define-derived-mode capnp-mode prog-mode
+  "capn-mode is a major mode for editing capnp protocol files"
+  :syntax-table capnp--syntax-table
+
+  (setq font-lock-defaults '((capnp--font-lock-keywords)))
+
+
+  (setq mode-name "capnp")
+  (define-key capnp-mode-map [remap comment-dwin] 'capnp-comment-dwim))
```
dwrensha: should this be `[remap comment-dwim]`?

netguy204:
@kentonv Thank you! Added.
@dwrensha Thanks for the catch. Fixed.

dwrensha:
Looks good to me.

kaos:
Very nice! thank you for this, Brian :+1:

PR #170, opened 2015-01-23
Stop trying to access /dev/urandom on windows

### LINKED ISSUE #167: `capnp id` fails on windows
... because /dev/urandom does not exist.

Diggsey:
Fixes #167

### DIFF
--- a/c++/src/capnp/compiler/parser.c++
+++ b/c++/src/capnp/compiler/parser.c++
@@ -27,6 +27,8 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <fcntl.h>
+#include <chrono>
+#include <random>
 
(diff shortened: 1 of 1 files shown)


kentonv:
Hi Diggory,

Thanks for this. I've been trying to find the time to deal with it and failing.

This implementation is probably OK in practice, but I'm feeling paranoid that the system timer has significantly less entropy than /dev/urandom and therefore there's a much higher chance that two people will accidentally get the same type ID. So, I'd rather use `CryptGenRandom()`. I've gone ahead and pushed commit 2f30b47 to implement this.

Sorry this took longer than it should have and thanks for pushing me to get it fixed! I'll try to do a 0.5.1 release with this tonight.

-Kenton

PR #171, opened 2015-01-26
add installation doc from brew for Mac OS X

infynyxx:
Makes life easier for OS X users using Homebrew


kentonv:
Thanks! I decided to expand this to mention the Debian package as well.

PR #172, opened 2015-01-27
fix 'capnp id' version string

dwrensha:
Before this patch, `capnp id` reports its version as "0.2":
```
$ capnp id --version
Cap'n Proto multi-tool 0.2
```


kentonv:
doh

PR #173, opened 2015-01-29
MSVS 2015 CTP5 Win64 warning fixes - explicit casts, remove unused list....

mintyc:
Fixes for Visual studio warnings on x64
- One outstanding warning in capnp\encoding-test.c++ line 1598 - truncated value trying to fit 0xFF into an int8_t - not sure how to appropriately tweak the test parameter to avoid truncation

### DIFF
--- a/c++/src/capnp/CMakeLists.txt
+++ b/c++/src/capnp/CMakeLists.txt
@@ -6,7 +6,6 @@ set(capnp_sources_lite
   blob.c++
   arena.c++
   layout.c++
-  list.c++
   any.c++
   message.c++
   schema.capnp.c++
--- a/c++/src/capnp/any-test.c++
+++ b/c++/src/capnp/any-test.c++
@@ -148,7 +148,7 @@ TEST(Any, AnyStruct) {
     MallocMessageBuilder b2;
     auto root2 = b2.getRoot<test::TestAnyPointer>();
     auto sb = root2.getAnyPointerField().initAsAnyStruct(
-        r.getDataSection().size() / 8, r.getPointerSection().size());
+        static_cast<uint>(r.getDataSection().size() / 8), static_cast<uint>(r.getPointerSection().size()));
 
     EXPECT_EQ(48, sb.getDataSection().size());
     EXPECT_EQ(20, sb.getPointerSection().size());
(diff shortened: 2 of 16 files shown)


kentonv:
Hi Paul,

Thanks for helping clean up our code, but it looks like this is the kind of warning-fixing that I prefer to avoid. I don't feel that this warning has a high enough hit rate to be worth the loss in readability it involves.

Can we instead just disable this warning?

mintyc:
I realise you've had this debate with others and I do understand your
viewpoint, but when working on something like MSVS which doesn't have the
'system header' I find it very useful to reduce these trivial warnings so
that other warnings are visible.

There are also a lot of places where blanket pragmas or warnings are closer
to a show-stopper.

Some of those static_casts are a little surprising e.g. subtract two
pointers and assume it fits in a int8_t
- a blanket pragma warning disable for MSVS to stop it bitching may be more
dangerous than adding the casts when you think you understand the limits
the narrowing conversion is bringing

Ultimately its your code, but I am in the camp that says if it can be
quietened down relatively painlessly we should do so. Even if you defer
such things to contributors and just keep doing the excellent and
interesting bits you are doing. Its why I wanted to give you an auto merge
to latest dev to make your life less hassle on such bits and pieces but
sadly the magic MSVS spells to get latest dev working were beyond me :)

pqu:
On 3 February 2015 at 11:47, mintyc <notifications@github.com> wrote:

> I realise you've had this debate with others and I do understand your
> viewpoint, but when working on something like MSVS which doesn't have the
> 'system header' I find it very useful to reduce these trivial warnings so
> that other warnings are visible.
>

This is a good argument for reducing warnings in the public header files
(of which there are only a few changes that need to be made), but less so
for the implementation files.


> Some of those static_casts are a little surprising e.g. subtract two
> pointers and assume it fits in a int8_t
>

Some of these warnings also look like they could be silenced by correcting
the underlying type, instead of casting over it. For example, would it be
cleaner if inRemaining in "serialize-packed.c++" was just declared a size_t?

The values in an operation with a kj::Quantity (e.g., "x * BYTES") should
also be protected when those types are enabled, and the cast makes the
multiplication with the unit quantity redundant.

(Also, should "list.c++" just be removed altogether?)

Philip

Issue #162, opened 2014-12-16
cmake build produces broken pkgconfig files


amluto:
I just installed from a git clone.  capnp.pc contains:

```
Version: 
```

capnp-rpc.pc contains:

```
Version: 
Libs: -L${libdir} -lcapnp-rpc -lkj-async
Requires: capnp = 
```

The result doesn't work (e.g. sandstorm won't build).  Manually shoving 0.5 into all three missing versions fixes it.

kentonv:
Fixed by #177 (long ago)

PR #177, opened 2015-02-01
Also set the version for the CMake build

a-richardson:
Otherwise the installed pkgconfig files are broken. The release script
has also been changed to update the version for CMakeLists.txt.

Fixes issue #162


kentonv:
Thanks!

PR #178, opened 2015-02-01
Add a cmake package config module

RichardsonAlex:
This makes it a lot easier for CMake based projects to use Cap'n Proto.

Example usage:

```
find_package(CapnProto)
capnp_generate_cpp(FOO_SRCS FOO_HDRS foo.capnp)
add_executable(foo main.cpp ${FOO_SRCS})
target_link_libraries(foo CapnProto::capnp)
```

or if they need capnp-rpc replace 'CapnProto::capnp' with
'CapnProto::capnp-rpc' in the target_link_libraries call.

This is a lot better than the previous variable based solution since
linking to nonexistent targets is an error whereas an empty variable
expansion (e.g. due to typos) will be silently ignored. It also makes
sure that the right compiler flags, include directories, defines and
link libraries are passed to the compiler for that target without
needing any other include_directories(), etc.


kentonv:
Hey @pqu, could you take a look at this?

Thanks!

c++/CMakeLists.txt
```diff
@@ -82,11 +82,40 @@ else()
...
+include(CapnProtoMacros)
 add_subdirectory(src)
 
 # Install ======================================================================
 
+include(CMakePackageConfigHelpers)
+write_basic_package_version_file(
+  "${CMAKE_CURRENT_BINARY_DIR}/CapnProtoConfigVersion.cmake"
+  VERSION ${VERSION}
+  COMPATIBILITY AnyNewerVersion
```
pqu: `AnyNewerVersion` should be `ExactVersion`. AFAIK, there aren't any backwards-compatibility guarantees for the API.

c++/CMakeLists.txt
```diff
@@ -82,11 +82,40 @@ else()
...
+)
+set(ConfigPackageLocation ${LIB_INSTALL_DIR}/cmake/CapnProto)
```
pqu: `CONFIG_PACKAGE_LOCATION` for consistency with other variables?

c++/CMakeLists.txt
```diff
@@ -82,11 +82,40 @@ else()
...
+install(EXPORT CapnProtoTargets
+  FILE CapnProtoTargets.cmake
+  NAMESPACE CapnProto::
+  DESTINATION ${ConfigPackageLocation}
+)
+install(FILES
+  cmake/CapnProtoMacros.cmake
+  ${CMAKE_CURRENT_BINARY_DIR}/CapnProtoConfig.cmake
+  ${CMAKE_CURRENT_BINARY_DIR}/CapnProtoConfigVersion.cmake
+  DESTINATION ${ConfigPackageLocation} COMPONENT Devel
```
pqu: Remove `COMPONENT`? Component-based installation isn't used.

c++/src/kj/CMakeLists.txt
```diff
@@ -57,7 +57,15 @@ set(kj-std_headers
...
 )
 add_library(kj ${kj_sources})
-install(TARGETS kj ARCHIVE DESTINATION "${LIB_INSTALL_DIR}")
+#make sure the lite flag propagates to all users (internal + external) of this library
+target_compile_definitions(kj PUBLIC ${CAPNP_LITE_FLAG})
+#make sure external consumers don't need to manually set the include dirs
+target_include_directories(kj INTERFACE $<INSTALL_INTERFACE:include>)
+if(UNIX)
+  #make sure external consumers use C++11 when linking to any of the libs
+  target_compile_options(kj INTERFACE -std=gnu++0x)
```
pqu: I don't think this is necessary, and may cause problems if clients want to pass a different `-std` option. "kj/common.h" will give an error fairly quickly if it isn't being compiled in at least C++11 mode anyway.

c++/CMakeLists.txt
```diff
@@ -82,11 +82,40 @@ else()
...
+set(ConfigPackageLocation ${LIB_INSTALL_DIR}/cmake/CapnProto)
+
+configure_package_config_file(cmake/CapnProtoConfig.cmake.in
+  ${CMAKE_CURRENT_BINARY_DIR}/CapnProtoConfig.cmake
+  INSTALL_DESTINATION ${ConfigPackageLocation}
+  PATH_VARS INCLUDE_INSTALL_DIR
+)
+export(EXPORT CapnProtoTargets
+  FILE "${CMAKE_CURRENT_BINARY_DIR}/CapnProtoTargets.cmake"
+  NAMESPACE CapnProto::
```
pqu: `capnp::` for consistency with the C++ namespace?

pqu:
This is a good improvement, and is a much tidier way to export/import CMake
projects.

I've made some comments on the commit, and my only other reservation is
about discovering autotools installs from a CMake project (which is
necessary to build/link with a shared library version). I think the
"FindCapnProto.cmake" module should be retained (`CAPNP_GENERATE_CPP` could
still be split into "CapnProtoMacros.cmake") so that it can be used by
projects that need it to discover autotools/pkgconfig installs.

c++/CMakeLists.txt
kentonv: If this is the Cap'n Proto API we're talking about, I think it's safe enough to assume that newer versions will be source-compatible, but not binary-compatible. (It's true that we sometimes make breaking changes to the API but only in obscure parts. Anything widely-used will be gracefully deprecated.)

kentonv:
Thanks for reviewing, @pqu!

c++/CMakeLists.txt
a-richardson: I guess that makes sense, This is inconsistent since I copied thiose lines from the CMake docs (http://www.cmake.org/cmake/help/v3.1/manual/cmake-packages.7.html)

c++/CMakeLists.txt
a-richardson: It doesn't do any harm for a normal install and would allow users to skip installing these files.
It's that way in the docs so I did it as well. But I don't mind removing it if you would prefer it that way.

c++/src/kj/CMakeLists.txt
a-richardson: Makes sense, I'll remove it

c++/CMakeLists.txt
a-richardson: Most other config modules I have come across use upper case case names (e.g. Qt5::Core, GLEW::GLEW KF5::Config, etc). Also the Cmake docs use "Upstream::", so I guess it makes sense to be consistent with other packages.

a-richardson:
I'll update this to install a FindCapnProto.cmake when built with autotools and otherwise install the Config.cmake file.

c++/CMakeLists.txt
pqu: I'm not very familiar with it, but from a quick skim of the documentation, it seems to be tied to CPack? I can't see a method for excluding a component from a "make install" without building a CPack installer.

szenker:
It would be very nice to see this improvement on master branch.

katreniak:
Code in this PR was rebased and merged in #314. This PR can be closed.

PR #179, opened 2015-02-02
For MSVS builds allow install target to also copy compiler tools

mintyc:
Useful for custom install prefixes used for subsequent packaging
e.g.
mkdir install
cmake -G "Visual Studio 14 2015 Win64" -DCAPNP_LITE=1 -DEXTERNAL_CAPNP=1 -DCMAKE_INSTALL_PREFIX=./install ../c++
cmake --build . --config Release --target install

### DIFF
--- a/c++/src/capnp/CMakeLists.txt
+++ b/c++/src/capnp/CMakeLists.txt
@@ -6,7 +6,6 @@ set(capnp_sources_lite
   blob.c++
   arena.c++
   layout.c++
-  list.c++
   any.c++
   message.c++
   schema.capnp.c++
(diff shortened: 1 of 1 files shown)


kentonv:
Thanks for the contribution!

Hmm, I don't know cmake very well so may be misinterpreting, but is the idea here that you're copying the "external" capnp binaries into the install location?

If so, I suspect many people don't actually want that, since the user probably already installed the binaries where they want them before building. Particularly when cross-compiling (a typical use case for EXTERNAL_CAPNP), it would be inappropriate to take the host-compiled binaries and install them into the target binpath.

Could you describe a bit more about your use case that motivates this?

mintyc:
Quite often I want to separate the various components of a project and use
a local install prefix to bring together all the files I need for packaging
ready for deployment with other components. "make install" is a packaging
step rather than a global deployment step.

I don't necessarily want to embed the whole of capnproto inc tests and
source into my other project. I also don't want a system 'global' install
when different builds could conflict in their global requirements.

When not asking to build the tools (or in the case of MSVS you can't), if
you've asked to use the external compiler tools and you have explicitly
asked to build the install target (not built by default) you get the
executables copied. You can version and tarball this and then use this on
another build node or at some later time. Doesn't affect you unless you've
selected the install target with an install prefix.

Typically our automated build system will have a project to package up your
latest release. I want to be able to ship just the include/lib/bin as a
tarball to my other projects that will use it as a tarball dependency.

Paul

kentonv:
Hmm. Sorry that this is inconvenient for your build, but I just don't think copying the external capnp into the install location is the right thing to do. The install target should really only be installing things that were produced by the build. This is especially true when cross-compiling. Consider the case where you're building ARM libs for Android -- the host binaries are x86 and won't run on ARM, so installing them together with the ARM binaries is going to cause problems.

Sorry, but I think you'll need to find some other way to get the binaries into the right place for your build process.

mintyc:
Thanks for giving it consideration. I hadn't thought of the cross-compilation scenario.

PR #186, opened 2015-02-25
Ignore stuff that needs ignoring.

benlaurie:


kentonv:
Historically I've intentionally avoided adding build artifacts to `.gitignore` because I like to use `git status` to see when my source tree is unclean, but a number of people have complained now. I suppose I should just get used to using `git status --ignored` if I want to be OCD.

BTW, did you know that autotools support out-of-tree builds? E.g. from the `c++` directory you could do:

    mkdir build
    cd build
    ../configure
    make -j6 check

This would avoid the need to update `.gitignore` as often.

But `autoreconf` unfortunately still needs to run in-tree.

OK, I guess I'll accept this and see whether it bothers me. :)

Thanks!

PR #187, opened 2015-03-02
zero warnings: use KJ_UNUSED for unused constexpr & local vars

johnkdoe:
one function in arena.c++ and several snprintf_result local vars
in string.c++ get optimized away because they're only used in
KJ_DASSERT() invocations.

tthis patch thus identifies those unused declarations with KJ_UNUSED

### DIFF
--- a/c++/src/kj/arena.c++
+++ b/c++/src/kj/arena.c++
@@ -66,7 +66,7 @@ void Arena::cleanup() {
 
 namespace {
 
-constexpr bool isPowerOfTwo(size_t value) {
+constexpr bool KJ_UNUSED isPowerOfTwo(size_t value) {
   return (value & (value - 1)) == 0;
 }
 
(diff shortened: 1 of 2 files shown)


schwa:
+1

kentonv:
Hmm, any particular reason you chose #ifdefs rather than annotating these declarations with KJ_UNUSED?

johnkdoe:
was unfamiliar with KJ_UNUSED … i will so modify the pull request.

johnkdoe:
updated (with push -f to my forked version of the repo)

johnkdoe:
spelled warnings correctly in the commit msg & pull request

kentonv:
Thanks!

kentonv:
PS somehow I had no idea that GCC and Clang let you put __attribute__ in front of a declaration rather than only after it. That hugely simplifies supporting MSVC, where equivalent attributes are expected to appear before. Thanks for making me aware of this.

PR #188, opened 2015-03-04
Async cocoa 2

jjwchoy:
Submitting for early feedback. Changes since the last PR:

I've updated CocoaEventPort due to the recent EventPort interface changes. There's no support for datagram ports or signals yet, but this should be enough to integrate capnp-rpc calls with the framework's event loop.

I've modified `async-io-test.c++` so that each test is now a template function, and introduced a macro, `TEST_IO` that will run `f<AsyncIoContext>()` and `f<AsyncCocoaIoContext>()` if `__APPLE__` is defined. If you have a better way of doing this let me know.

I haven't implemented the relevant tests from `async-unix-test.c++` yet.


kentonv:
Hi Jason!

Thanks for implementing this.

I think the way we want to handle compatibility glue with other libraries is to put them under `kj/compat` and build them as independent libraries so that users can link them optionally.

Moving your code to `compat` seems easy enough. The test will be trickier. We'll need to have separate tests, but perhaps it makes sense to define a test suite independent of event loop implementation which async-io-test and your test call into. Like `capnp/test-util.{h,c++}`, but this is `kj/async-test-util.{h,c++}`. It's OK to make it a header-only template library if that's easiest.

Don't forget to update `Makefile.am` in addition to the cmake files.

I wonder if it makes sense to modify the configure script to detect whether CoreFoundation is available and working vs. just assuming so on Apple targets. In particular I wonder if people who are, say, using GCC from Homebrew will one day find that the CoreFoundation headers only compile under Clang and therefore they can't build Cap'n Proto even though it's an optional dependency. Perhaps that's a silly thing to worry about?

There were some new methods added to AsyncIoStream and ConnectionReceiver recently: `getsockopt`, `setsockopt`, `getsockname`, `getpeername`. The default implementations throw "unimplemented" exceptions but when wrapping an FD you should be sure to implement them.

What does this implementation do with multiple overlapping read()s or write()s? Currently the main implementation doesn't support this, but I would like to fix that in the future, and all the other implementations will eventually need to be updated too. In the long term, overlapping calls should be completed in the order that they were called. No need to implement this immediately but something to keep in mind...

jjwchoy:
Hi Kenton,

Your suggestions sound good. Regarding CoreFoundation an gcc, perhaps these compat libraries should not be part of the default make target. We could have a `make all` to build everything, or you could build the default set + any extra libs you want by invoking make for each one individually.

Overlapping calls to `onReadable()` or `onWritable` are unsupported. Most of the non-apple-specific code was just copied from async-unix - it would be nice to factor this out at some point!

kentonv:
> perhaps these compat libraries should not be part of the default make target.

I think the usual practice is to have `configure` auto-detect if the library exists and is accepted by the compiler, and then build the glue code if so. `configure` should also accept `--with-foo`/`--without-foo` options to explicitly decide what components to build.

kentonv:
BTW, stop me if we talked about this before... but if `UnixEventPort` were an abstract interface, would you be able to implement CFRunLoop support at that level to avoid re-implementing AsyncIoStream? Note that version 0.5.0 significantly changed this interface.

jjwchoy:
> if `UnixEventPort` were an abstract interface, would you be able to implement CFRunLoop support at that level to avoid re-implementing AsyncIoStream? 

I think the right level of abstraction is the `FdObserver`. All the differing logic is encapsulated within that interface. How about an `AsyncIoStream` template that can be specialized by `FdObserver`?

kentonv:
Perhaps we want a new interface `AsyncFd` which basically has the methods of `FdObserver` plus `int getFd()` (to pass to other system calls).

Then we add this to `LowLevelAsyncIoProvider`:

    Own<AsyncFd> wrapFd(int fd, Own<FdObserver> observer, uint flags = 0);

And then the methods of LLAIP that return streams can be given default implementations based on `wrapFd()` (but can still be overridden if someone wants to specialize).

Maybe after this, AsyncIoContext really doesn't need a UnixEventPort anymore... I'll need to update some Sandstorm code, though. And I guess LLAIP needs to be extended to support signals, which may or may not be hard considering how weird signals are. Hmm.

jjwchoy:
Sounds good to me. Should AsyncFd and AsyncStreamFd live in async-unix?

I haven't even attempted to support signals yet.

kentonv:
No, they shouldn't go in async-unix because other EventPorts would be likely to implement the new interface. `AsyncFd` (the interface) should be in `async-io.h`. `AsyncStreamFd` (the implementation of `AsyncIoStream`, now in terms of `AsyncFd`) should stay where it is in `async-io.c++`, probably at the very top.

Does CFRunLoop support signals? I know Apple tries to avoid them.

PR #190, opened 2015-03-14
Hardcode /bin/cat output option for `capnp compile`.

dwrensha:
In the typical build process for a Rust project that uses Cap'n Proto code generation, we have access to capnpc-rust as a library, but not as an executable. (See [this thread](https://github.com/rust-lang/cargo/issues/872) for some discussion.)  The `compile()` function, defined [here](https://github.com/dwrensha/capnpc-rust/blob/master/src/lib.rs), is designed to handle this case. It spawns a child that runs `capnp compile -o/bin/cat`, and forwards the child's stdout to the usual codegen procedure. Things get more complicated when `/bin/cat` is not present, like on Windows, or even when it is not in its usual location (https://github.com/dwrensha/capnpc-rust/issues/9). Adding an option for `capnp compile` to directly write to stdout should make things simpler.

I had thought that "-o -" would be a nice way to specify this option, but I've found that that would interact weirdly with option parsing. For instance, `capnp compile -o -:/tmp` would give an error `-o: missing argument`.

### DIFF
--- a/c++/src/capnp/compiler/capnp.c++
+++ b/c++/src/capnp/compiler/capnp.c++
@@ -242,10 +242,12 @@ public:
     builder.addOptionWithArg({'o', "output"}, KJ_BIND_METHOD(*this, addOutput), "<lang>[:<dir>]",
                              "Generate source code for language <lang> in directory <dir> "
                              "(default: current directory).  <lang> actually specifies a plugin "
-                             "to use.  If <lang> is a simple word, the compiler for a plugin "
+                             "to use.  If <lang> is a simple word, the compiler searches for a plugin "
                              "called 'capnpc-<lang>' in $PATH.  If <lang> is a file path "
                              "containing slashes, it is interpreted as the exact plugin "
-                             "executable file name, and $PATH is not searched.")
+                             "executable file name, and $PATH is not searched.  The special case of "
+                             "'/bin/cat' is hardcoded to dump the request to standard output, as "
+                             "this is a common operation and /bin/cat is not available on all systems.")
            .addOptionWithArg({"src-prefix"}, KJ_BIND_METHOD(*this, addSourcePrefix), "<prefix>",
                              "If a file specified for compilation starts with <prefix>, remove "
                              "the prefix for the purpose of deciding the names of output files.  "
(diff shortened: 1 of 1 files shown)


kentonv:
Hmm, but presumably one would never specify an output directory with this option anyway, right? So does that really matter? Special-casing /bin/cat feels weird.

On the topic of Rust's build system wanting to do everything as a library, have you considered invoking SchemaParser as a library? (Presumably calling C++ code is not too hard? Though maybe that is pretty hard on Windows...)

dwrensha:
Ok, I've switched it to use '-o -'. I suppose the weirdness I was worried about has in fact been present all along. For instance, you can't include the directory `-some-dir` with `capnp compile -I -some-dir`; you instead have to do `capnp compile -import-path=-some-dir`. Of course, directories starting with a '-' are not common. Appending a directory to a  `-o -` option would be even less common, as it would serve no purpose. So I agree that this isn't a big deal.

dwrensha:
> have you considered invoking SchemaParser as a library?

I've considered it, but it doesn't seem like there would be any major advantages.

kentonv:
FWIW, you don't actually need a space between a short option and its argument, so e.g. `-I-some-dir` would work. But, yeah, filenames with leading `-`s are generally problematic for Unix.

bozaro:
:+1: It would be wonderful to build a minor release with these pull request.

Without it, to use capnp + rust under Ms Windows, I need:
 * use patched version of capnproto (release-0.5.1 + d104b7e + 33af12c)
 * use patched version of https://github.com/dwrensha/capnpc-rust.git

Since under Windows has problems with whitch and cat commands :(

dwrensha:
@bozaro: I think the easiest way to work around this for now would be to use the `capnpc-rust` executable. You can get it by checking out the `capnpc-rust` repo and doing `cargo build --release`. If you install it the same place that you've installed `capnpc-c++` and `capnpc-capnp`, you should be able to do `capnp compile -orust <input>`.

bozaro:
@dwrensha: Thanks for suggestion. I already have a workaround, but it would be much better if it worked out of the box :)

kentonv:
Hi! Can you please test this release candidate: https://capnproto.org/capnproto-c++-win32-0.5.2-rc1.zip

bozaro:
Big thanks :)
I checked 0.5.2-rc1 and it works fine in my case :)

kentonv:
OK, release 0.5.2 is up.

PR #192, opened 2015-03-29
Add equal / operator == method

joshuawarner32:
This is a "simplest thing that could possibly work" implementation.  I'm fully expecting to reorganize the code, change the types around, etc - after hearing feedback.  I'll also clean up the git history at that point, which has rather unhelpful commit messages right now.

### DIFF
--- a/c++/src/capnp/any.c++
+++ b/c++/src/capnp/any.c++
@@ -198,7 +198,8 @@ bool AnyPointer::Reader::operator ==(AnyPointer::Reader right) {
     case StructEqualityResult::NOT_EQUAL:
       return false;
     case StructEqualityResult::UNKNOWN_CONTAINS_CAPS:
-      KJ_FAIL_REQUIRE();
+      KJ_FAIL_REQUIRE(
+        "operator== cannot determine equality of capabilities; use equals() instead if you need to handle this case");
     default:
       KJ_UNREACHABLE;
   }
@@ -211,7 +212,8 @@ bool AnyStruct::Reader::operator ==(AnyStruct::Reader right) {
     case StructEqualityResult::NOT_EQUAL:
       return false;
     case StructEqualityResult::UNKNOWN_CONTAINS_CAPS:
-      KJ_FAIL_REQUIRE();
+      KJ_FAIL_REQUIRE(
+        "operator== cannot determine equality of capabilities; use equals() instead if you need to handle this case");
     default:
       KJ_UNREACHABLE;
   }
@@ -224,7 +226,8 @@ bool AnyList::Reader::operator ==(AnyList::Reader right) {
     case StructEqualityResult::NOT_EQUAL:
       return false;
     case StructEqualityResult::UNKNOWN_CONTAINS_CAPS:
-      KJ_FAIL_REQUIRE();
+      KJ_FAIL_REQUIRE(
+        "operator== cannot determine equality of capabilities; use equals() instead if you need to handle this case");
     default:
       KJ_UNREACHABLE;
   }
--- a/c++/src/capnp/any.c++
+++ b/c++/src/capnp/any.c++
@@ -134,7 +134,7 @@ StructEqualityResult equal(AnyList::Reader left, AnyList::Reader right) {
   case ElementSize::FOUR_BYTES:
   case ElementSize::EIGHT_BYTES:
     if(left.getElementSize() == right.getElementSize()) {
-      if(memcmp(left.getData().begin(), right.getData().begin(), left.getData().size()) == 0) {
+      if(memcmp(left.getRawBytes().begin(), right.getRawBytes().begin(), left.getRawBytes().size()) == 0) {
         return StructEqualityResult::EQUAL;
       } else {
         return StructEqualityResult::NOT_EQUAL;
--- a/c++/src/capnp/any.c++
+++ b/c++/src/capnp/any.c++
@@ -109,12 +109,12 @@ StructEqualityResult AnyStruct::Reader::equals(AnyStruct::Reader right) {
     auto l = ptrsL[i];
     auto r = ptrsR[i];
     switch(l.equals(r)) {
-    case StructEqualityResult::EQUAL:
-      break;
-    case StructEqualityResult::NOT_EQUAL:
-      return StructEqualityResult::NOT_EQUAL;
-    case StructEqualityResult::UNKNOWN_CONTAINS_CAPS:
-      return StructEqualityResult::UNKNOWN_CONTAINS_CAPS;
+      case StructEqualityResult::EQUAL:
+        break;
+      case StructEqualityResult::NOT_EQUAL:
+        return StructEqualityResult::NOT_EQUAL;
+      case StructEqualityResult::UNKNOWN_CONTAINS_CAPS:
+        return StructEqualityResult::UNKNOWN_CONTAINS_CAPS;
     }
   }
 
@@ -124,12 +124,12 @@ StructEqualityResult AnyStruct::Reader::equals(AnyStruct::Reader right) {
 
 kj::StringPtr KJ_STRINGIFY(StructEqualityResult res) {
   switch(res) {
-  case StructEqualityResult::NOT_EQUAL:
-    return "NOT_EQUAL";
-  case StructEqualityResult::EQUAL:
-    return "EQUAL";
-  case StructEqualityResult::UNKNOWN_CONTAINS_CAPS:
-    return "UNKNOWN_CONTAINS_CAPS";
+    case StructEqualityResult::NOT_EQUAL:
+      return "NOT_EQUAL";
+    case StructEqualityResult::EQUAL:
+      return "EQUAL";
+    case StructEqualityResult::UNKNOWN_CONTAINS_CAPS:
+      return "UNKNOWN_CONTAINS_CAPS";
   }
 }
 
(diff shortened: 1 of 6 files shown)


kentonv:
I hate to say this, but changing AnyStruct::getDataSection() to return a uint64_t array rather than a byte array violates C++ aliasing rules and so won't work. The only pointer type that is allowed to alias other data is char (including unsigned char). You cannot actually read or write the contents of a struct's data section as a uint64_t array. For example, say you have:

    struct MyType {
      foo @0 :UInt32;
    }

And then you wrote:

    MyType::Builder b = ...;
    b.setFoo(123);
    AnyStruct::Builder(b).getDataSection()[0] = 456;
    uint32_t foo = b.getFoo();

Will the value of `foo` end up being 123 or 456? The answer is: it could be either! Since the value 456 is being set through a uint64_t pointer, and getFoo() reads from a uint32_t pointer, the compiler is allowed to assume that the two cannot point to the same value, and therefore the compiler may reorder the assignment to occur after `getFoo()` (or before `setFoo()`), and `foo` might therefore end up being 123.

This is why the `capnp::word` type is opaque (rather than being a typedef of uint64_t). We use it to clarify pointer arithmetic and alignment, but we never actually read or write a `capnp::word` value directly; we instead cast to something else, or use `memcpy()` which effectively operates on bytes.

Another problem with representing the data section as a uint64_t array is that it might lead to endianness-dependent application code.

joshuawarner32:
I actually wasn't considering saving people from endianness-dependent code, I totally forgot about aliasing rules.  Both very good points.

I changed that primarily for chopping off zeroed words at the end of structs (for canonicalization).  There's no point in chopping off a byte at a time - it'll have no functional difference, but probably be a little slower.  In my hashing implementation, I have very similar code, except it uses a nasty series of casts to get there.

Note that, aliasing rules aside, what I'm doing with the result should be safe (comparing it with 0) - since that's a completely endianness-independent operation.  Perhaps `capnp::word` can allow for that?

Is there room for some sort of middle-ground here, where we can do certain safe operations on them as complete words, but not break aliasing rules?

joshuawarner32:
I went ahead and reverted back to the original `Data::*` return types for `getDataSection` and cleaned up the git history.  Let me know if you notice anything else.

c++/src/capnp/layout.c++
```diff
@@ -2368,24 +2372,26 @@ MessageSizeCounts PointerReader::targetSize() const {
...
-}
-
-bool PointerReader::isList() const {
-  word* refTarget = nullptr;
-  const WirePointer* ptr = pointer;
-  SegmentReader* sgmt = segment;
-  WireHelpers::followFars(ptr, refTarget, sgmt);
-  return ptr->kind() == WirePointer::Kind::LIST;
+PointerType PointerReader::getPointerType() const {
+  if(pointer->isNull()) {
```
kentonv: The previous code for isNull() also returned true if `pointer` itself was null, but you've removed that check here. Better add it back.

c++/src/capnp/layout.h
```diff
@@ -656,6 +654,8 @@ class ListReader {
   Data::Reader asData();
   // Reinterpret the list as a blob.  Throws an exception if the elements are not byte-sized.
 
+  Data::Reader asDataOfAnySize();
```
kentonv: Hmm, perhaps this should be `kj::ArrayPtr<const byte> asRawBytes()`. That way `Data` specifically corresponds to the `Data` type used in Cap'n Proto schemas. This should also be documented as being valid only for data-only lists.

`getData()` in AnyList should similarly become `getRawBytes()` and return ArrayPtr.

`getDataSection()` in AnyStruct should also return ArrayPtr. (It should be safe to make this change.)

c++/src/capnp/any.h
```diff
@@ -806,6 +806,31 @@ inline Orphan<AnyPointer> Orphan<AnyPointer>::releaseAs() {
   return kj::mv(*this);
 }
 
+enum class StructEqualityResult {
+  NOT_EQUAL,
+  EQUAL,
+  UNKNOWN_CONTAINS_CAPS
+};
+
+inline kj::StringPtr KJ_STRINGIFY(StructEqualityResult res) {
```
kentonv: This function is too long to inline.

c++/src/capnp/any.h
```diff
@@ -806,6 +806,31 @@ inline Orphan<AnyPointer> Orphan<AnyPointer>::releaseAs() {
...
+  case StructEqualityResult::NOT_EQUAL:
+    return "NOT_EQUAL";
+  case StructEqualityResult::EQUAL:
+    return "EQUAL";
+  case StructEqualityResult::UNKNOWN_CONTAINS_CAPS:
+    return "UNKNOWN_CONTAINS_CAPS";
+  }
+}
+
+StructEqualityResult equal(AnyStruct::Reader left, AnyStruct::Reader right);
```
kentonv: I'd prefer to see these defined as methods, both for namespacing reasons and so that people can find them when reading the AnyStruct etc. class definitions.

c++/src/capnp/any.c++
```diff
@@ -77,4 +79,127 @@ kj::Own<ClientHook> AnyPointer::Pipeline::asCap() {
...
+  size_t dataSizeR = dataR.size();
+  while(dataSizeR > 0 && dataR[dataSizeR - 1] == 0) {
+    -- dataSizeR;
+  }
+
+  if(dataSizeL != dataSizeR) {
+    return StructEqualityResult::NOT_EQUAL;
+  }
+
+  if(0 != memcmp(dataL.begin(), dataR.begin(), dataSizeL * sizeof(word))) {
```
kentonv: I think `sizeof(word)` is no longer correct here?

c++/src/capnp/any.c++
```diff
@@ -77,4 +79,127 @@ kj::Own<ClientHook> AnyPointer::Pipeline::asCap() {
...
+  auto ptrsL = left.getPointerSection();
+  auto ptrsR = right.getPointerSection();
+
+  size_t i = 0;
+
+  for(; i < kj::min(ptrsL.size(), ptrsR.size()); i++) {
+    auto l = ptrsL[i];
+    auto r = ptrsR[i];
+    switch(equal(l, r)) {
+    case StructEqualityResult::EQUAL:
```
kentonv: stupid style guide nitpick: Please indent case labels two spaces beyond the switch, and case bodies two spaces beyond the labels.

c++/src/capnp/any.c++
```diff
@@ -77,4 +79,127 @@ kj::Own<ClientHook> AnyPointer::Pipeline::asCap() {
...
+    return StructEqualityResult::EQUAL;
+  }
+  }
+}
+
+StructEqualityResult equal(AnyPointer::Reader left, AnyPointer::Reader right) {
+  if(right.isCapability()) {
+    return StructEqualityResult::UNKNOWN_CONTAINS_CAPS;
+  }
+  if(left.isNull()) {
```
kentonv: switch() here on pointer type rather than if/else?

c++/src/capnp/any.c++
```diff
@@ -77,4 +79,127 @@ kj::Own<ClientHook> AnyPointer::Pipeline::asCap() {
...
+      }
+    }
+    return StructEqualityResult::EQUAL;
+  }
+  }
+}
+
+StructEqualityResult equal(AnyPointer::Reader left, AnyPointer::Reader right) {
+  if(right.isCapability()) {
+    return StructEqualityResult::UNKNOWN_CONTAINS_CAPS;
```
kentonv: I think we should only return `UNKNOWN_CONTAINS_CAPS` if forced to compare two caps. When comparing a cap to null, for example, it seems safe to return `NOT_EQUAL`.

c++/src/capnp/any.c++
```diff
@@ -77,4 +79,127 @@ kj::Own<ClientHook> AnyPointer::Pipeline::asCap() {
...
+    if(right.isList()) {
+      return equal(left.getAs<AnyList>(), right.getAs<AnyList>());
+    } else {
+      return StructEqualityResult::NOT_EQUAL;
+    }
+  } else if(left.isCapability()) {
+    return StructEqualityResult::UNKNOWN_CONTAINS_CAPS;
+  } else {
+    // There aren't currently any other types of pointers
+    KJ_FAIL_REQUIRE();
```
kentonv: Use KJ_UNREACHABLE. (I commonly follow a switch() that handles all possible enums with KJ_UNREACHABLE. If a new enumerant is added, the switch() will produce a compiler warning.)

c++/src/capnp/any.c++
```diff
@@ -77,4 +79,127 @@ kj::Own<ClientHook> AnyPointer::Pipeline::asCap() {
...
+}
+
+bool operator ==(AnyPointer::Reader left, AnyPointer::Reader right) {
+  switch(equal(left, right)) {
+  case StructEqualityResult::EQUAL:
+    return true;
+  case StructEqualityResult::NOT_EQUAL:
+    return false;
+  case StructEqualityResult::UNKNOWN_CONTAINS_CAPS:
+    KJ_FAIL_REQUIRE();
```
kentonv: The exception should have a reasonable error message.

    KJ_FAIL_REQUIRE("operator== cannot determine equality of capabilities; use equals() instead if you need to handle this case");

c++/src/capnp/any.c++
kentonv: In fact, _arguably_ we should only use `UNKNOWN_CONTAINS_CAPS` in the case where everything else is equal except for the caps. Otherwise `NOT_EQUAL` is obviously correct. This way a caller could potentially decide that they only care about comparing the non-cap parts, and so could treat `UNKNOWN_CONTAINS_CAPS` as `EQUAL`.

kentonv:
Thanks for working on this! I've added a bunch of comments. Sorry for the delay -- it's crunch time at Sandstorm.

c++/src/capnp/layout.c++
joshuawarner32: Whoops, not sure how I missed that.

c++/src/capnp/layout.h
joshuawarner32: Fine by me.  There's also getDataSectionAsBlob, which may make sense to rename (I haven't as of yet).

c++/src/capnp/any.h
joshuawarner32: Probably true.  I figured it would be fairly unusual to need this linked in, so putting it in the header gives the compiler a better chance of eliminating it as dead code.  Right now, it's only needed for the tests.  I went ahead and removed it.

c++/src/capnp/any.h
joshuawarner32: That's fine.  I added it to a Any{Struct,List,Pointer}::{Reader,Builder} for symmetry.  That leads to some duplication that I'm not sure how to get rid of (`operator ==`).

c++/src/capnp/any.c++
joshuawarner32: Yup.  Good catch.

c++/src/capnp/any.c++
joshuawarner32: I think I mentioned this before, but how do you feel about having an auto-formatter config file checked in?  I'm personally a fan of clang-format.

c++/src/capnp/any.c++
joshuawarner32: Yep.

c++/src/capnp/any.c++
joshuawarner32: Makes sense.  Done, I think.  I'm not sure the best way to test that logic.

joshuawarner32:
Not a problem.  I'm rooting for Sandstorm.

I think I took care of everything.  I also renamed `StructEqualityResult` to `Equality`, since the long name was getting on my nerves.  I'm not attached to it if you have a preference.

I squashed the history, since it didn't seem particularly important.  I still have the original commits, if you'd prefer (they're actually meaningfully separated this time).

kentonv:
Hi Joshua,

It's pretty difficult for me to iterate on a large code review like this if I can't see the incremental changes. Would you mind un-smashing them? Sorry I didn't see this earlier.

kentonv:
(That is to say, the incremental commits from the last time I reviewed the code -- I don't want to re-read the stuff I've seen already.)

joshuawarner32:
@kentonv no problem.  I'll keep that in mind in the future.  Pushed.

kentonv:
Beautiful, much easier to follow, especially with a commit for every comment like that.

Just the one comment left about placement of KJ_UNREACHABLE.

kentonv:
Thanks!

PR #200, opened 2015-05-03
Canonicalize NaN on the hppa architecture

thomaslee:
Original credit to John David Anlin on [Debian bug 781787](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=781787) who noticed the hppa architecture was affected in the same way as mips wrt NaN quirks. In his own words:

> The hppa and mips architecture have the same nan format (i.e., quiet and signalling nans are
> different from x86, etc).  With the attached change, capnproto builds successfully on hppa.  See:
> http://buildd.debian-ports.org/status/fetch.php?pkg=capnproto&arch=hppa&ver=0.4.1-3&stamp=1428027386

### DIFF
--- a/c++/src/capnp/layout.h
+++ b/c++/src/capnp/layout.h
@@ -39,7 +39,7 @@
 #include "blob.h"
 #include "endian.h"
 
-#if __mips__ && !defined(CAPNP_CANONICALIZE_NAN)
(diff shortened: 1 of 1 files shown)


kentonv:
Thanks!

kentonv:
Do you need me to do a point release with this or can you just add the patch to the deb package?

thomaslee:
No need to cut a release, I can use John's patch downstream next time I have 5 mins to spare. Thanks though!

PR #202, opened 2015-05-09
Fall back to sysconf(_SC_IOV_MAX) if both IOV_MAX and UIO_MAXIOV are undefined

thomaslee:
Attempt to fix a build error seen on GNU/Hurd:

https://buildd.debian.org/status/fetch.php?pkg=capnproto&arch=hurd-i386&ver=0.5.2-1&stamp=1431041329

Based loosely on: https://lists.debian.org/debian-hurd/2012/05/msg00120.html


thomaslee:
Notice we're using IOV_MAX in a number of other places, so consider this a WIP for now.

thomaslee:
Updated to patch up IOV_MAX in a few other places.

c++/src/kj/iovmax.h
```diff
@@ -0,0 +1,64 @@
...
+// Apparently, there is a maximum number of iovecs allowed per call.  I don't understand why.
+// Most platforms define IOV_MAX but Linux defines only UIO_MAXIOV and others, like Hurd,
+// define neither.
+//
+// On platforms where both IOV_MAX and UIO_MAXIOV are undefined, we poke sysconf(_SC_IOV_MAX),
+// try to fall back to the POSIX-mandated minimum of _XOPEN_IOV_MAX if that fails.
+//
+// http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/limits.h.html#tag_13_23_03_01
+inline size_t iovMax(size_t count)
+{
```
kentonv: Please match formatting style of the rest of the package:
* Open brace never on its own line.
* `} else {` on one line.
* Two-space indents.

c++/src/kj/iovmax.h
```diff
@@ -0,0 +1,64 @@
+// Copyright (c) 2013-2014 Sandstorm Development Group, Inc. and contributors
```
kentonv: Perhaps instead of a new file, we could put this in `miniposix.h`, which is already a cross-platform compatibility file?

c++/src/kj/iovmax.h
kentonv: Also doc comments go inside the braces, e.g.:

    void foo() {
      // doc comment for foo()

See: https://github.com/sandstorm-io/capnproto/blob/master/style-guide.md#irrelevant-formatting-rules

thomaslee: D'oh, will do.

c++/src/kj/iovmax.h
thomaslee: Hm does seem like an appropriate place, but notice iovmax.h is only used internally within kj & doesn't form part of any public API. Are you okay with exposing what I'm effectively treating as an internal implementation detail?

In particular, I'm not providing an implementation of iovMax for Windows here because it the IOV_MAX thing doesn't really apply there. Should there be some stub implementation of iovMax for Windows just in case somebody tries to call iovMax on POSIX only to see it fail when they build under say MinGW?

kentonv: Actually, `miniposix.h` isn't intended to be a public API itself, and arguably it should not be installed with the rest of the headers.

No need for a Win32 iovMax() since Win32 doesn't have iov in the first place, so no one would ever call it.

thomaslee:
@kentonv how's this look? Moved `miniposix.h` into noinst_HEADERS too given it's not part of the public API.

kentonv:
Thanks!

PR #207, opened 2015-05-20
Fix -Werror build without KJ_HAS_BACKTRACE

emaste:
This avoids an unused function warning for crashHandler.

### DIFF
--- a/c++/src/kj/exception.c++
+++ b/c++/src/kj/exception.c++
@@ -147,6 +147,7 @@ String stringifyStackTrace(ArrayPtr<void* const> trace) {
 #endif
 }
 
+#if KJ_HAS_BACKTRACE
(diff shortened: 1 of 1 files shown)


kentonv:
Thanks!

FWIW, this would have been caught before the next release as -Werror testing is part of the release process. Be warned that we generally recommend against using non-release versions of Cap'n Proto, although if you're working with Sandstorm.io (which is co-developed with Cap'n Proto) then it may be required.

emaste:
Hi @kentonv.

Just a note, my testing and this report comes out of a desire to make sure Cap'n Proto builds and tests properly on FreeBSD, not from a real use case.

kentonv:
Fair enough! :)

PR #208, opened 2015-05-21
Fix Syntax highlighting for Kate/KDevelop

a-richardson:
Apparently Kate requires the `<itemDatas>` to be inside the `<highlighting>`
tag whereas QtCreator also accepts it a level higher.

Tested using this DTD:
https://github.com/jgm/highlighting-kate/blob/master/xml/language.dtd


kentonv:
Merging this one, thanks!

PR #209, opened 2015-05-21
Improvements to syntax highlighting

a-richardson:
I split this up into lots of small commits to make it easier to review. It also contains the commit from #208 since I don't know how to make pull requests depend on each other (is this even possible?).

The commit `Highlight: Use default colors for keywords and types` conflicts with the comment regarding not using dsKeyword and dsDataType since I don't quite understand the problem described in the comment. Does this mean that some of the hardcoded colours would be the same as the ones for keywords or types? In Kate/KDevelop this is not a problem since the colours in the syntax highlighting definition are only a default an can be overridden per language in the settings dialog:
settings

I can also try and contribute this syntax highlighting file to upstream Kate I you don't mind that.

I tested the highlighting using this file: https://paste.kde.org/ptdamq4xz

### DIFF
--- a/highlighting/qtcreator/capnp.xml
+++ b/highlighting/qtcreator/capnp.xml
@@ -106,25 +106,25 @@
         <RegExpr attribute="Annotation" context="#stay" String="'([^'\\]|\\.)*'"/>
       </context>
     </contexts>
+    <itemDatas>
+      <!-- Since my highlighting approach is different from most programming languages, the contexts
+          don't fit nicely into the standard set, so I have to define colors explicitly.  A couple
+          of these, like "keyword" and "type", could be mapped to dsKeyword and dsDataType, but
+          there's a chance the user has mapped the colors for those things to things that would
+          conflict with the manually-defined colors here, which would probably be even more annoying
+          than having the colors be inconsitent from other languages.  So, I use manual colors for
+          everything, except comments, which I figure are less likely to have this problem. -->
+      <itemData name="Normal"       defStyleNum="dsNormal"/>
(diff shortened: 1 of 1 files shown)


kentonv:
Hmm, the problem is, Qt Creator does not appear to offer me the option to configure colors for the specific token types declared by my Cap'n Proto highlighter definition. Instead, Qt Creator seems to offer one giant, static list of things whose colors I can configure. They've included everything relevant to (Qt-flavored) C++ and (Qt-flavored) JavaScript, but I'm offered no way to configure colors for Cap'n Proto. Hence I had to declare them directly in the XML.

So while your changes seem like the right thing to do for KDevelop and the like, it effectively breaks support for Qt Creator. Maybe we need two files? Any ideas?

kentonv:
Looking at the content of your changes, it looks like you're trying to change the highlighter to use a more conventional token-highlighting approach. I had actually intentionally designed this highlighter to highlight contexts rather than tokens, e.g. having `= <default value>` be entirely blue while `$annotation(...)` is entirely yellow. Personally I find this highlighting style to be more useful to me but I understand if some people find it weird and confusing.

So I think this further argues for forking the file rather than changing it in-place...

a-richardson:
Yes, that sounds like it would make sense to split it. Would it be okay if I contribute your highlighting file with my changes to the upstream Kate repository?

kentonv:
Well, I'm a little sad that the standard highlighting then won't be the style I like, but it's certainly better than not having it, and since I'm not volunteering to try to upstream it myself, I suppose you should go for it. :)

PR #211, opened 2015-05-29
Fixed installation of schema files via CMake.

Fraser999:
If building the lite target only (i.e. having run CMake with `-DCAPNP_LITE=1 -DEXTERNAL_CAPNP=1`) then the c++.capnp and schema.capnp files were missing from the install folder, meaning that for example you couldn't set a namespace.

This changes the schema files to be installed along with their respective headers.


kentonv:
Thanks!

@pqu or @joshuawarner32 can you review?

joshuawarner32:
Pretty simple, looks good to me.  Thanks!

PR #218, opened 2015-06-19
Allow trailing commas in parenthesized and bracketed lists.

zarvox:
Add tests to verify this behavior.

Also, include `schema-parser-test.c++` in the CMake and autotools test suites.  Somehow it got left out.

### DIFF
--- a/c++/Makefile.am
+++ b/c++/Makefile.am
@@ -375,6 +375,7 @@ heavy_tests =                                                  \
   src/capnp/capability-test.c++                                \
   src/capnp/schema-test.c++                                    \
   src/capnp/schema-loader-test.c++                             \
+  src/capnp/schema-parser-test.c++                             \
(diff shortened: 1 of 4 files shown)


kentonv:
I hate to say this, but I think the right place to make this edit is actually in `lexer.c++`, around line 159-ish. That's the same place where we handle differentiating an empty list from a one-item list, which similarly detects has to detect that there are no tokens in the list item.

zarvox:
Ahh, yeah, I wasn't sure if the lexer should strip out the commas as though they were never there, or if it should detect them, and then let the parser remove them instead.  But I'll go ahead and revamp this patchset so it's the lexer stripping things out instead.

zarvox:
Okay, reworked to have the lexer detect an empty final list item.

paulproteus:
I just want to say:

+1
​

PR #222, opened 2015-07-01
Workaround nonexistent EPROTO on OpenBSD

### LINKED ISSUE #221: EPROTO isn't defined on OpenBSD
async-io.c++ has a match for EPROTO (to retry the accept), but EPROTO isn't defined on OpenBSD. I guess a simple `#ifndef __OpenBSD__` at the line in question would fix that.

mrdomino:
Fixes #221.

### DIFF
--- a/c++/src/kj/async-io.c++
+++ b/c++/src/kj/async-io.c++
@@ -806,7 +806,9 @@ public:
 
         case EINTR:
         case ENETDOWN:
+#if !__OpenBSD__
         case EPROTO:
+#endif
         case EHOSTDOWN:
         case EHOSTUNREACH:
         case ENETUNREACH:


kentonv:
Thanks! But would it make sense to write `#ifndef EPROTO` instead, to cover any other platforms that also do not define this error value? Also that way if OpenBSD ever added this error, we'd have it covered.

mrdomino:
Perhaps, but then would you get in trouble if EPROTO was defined in an enum
instead?

I'm fine with either approach, as both will get the code building on my
machine. The safest thing might be an autoconf test, but that might be
overkill here...

kentonv:
> Perhaps, but then would you get in trouble if EPROTO was defined in an enum instead?

I think the standard states they must be macros, and in fact there is a lot of existing code that #ifdefs on these and would break on a platform where they are not macros.

That said, you could add a check like:

    #ifndef ERANGE
    #error "It appears errno codes are not macros on your system. We weren't expecting this. Please file a bug report!"
    #endif

mrdomino:
See #223.

PR #223, opened 2015-07-03
Workaround nonexistent EPROTO on OpenBSD

### LINKED ISSUE #221: EPROTO isn't defined on OpenBSD

mrdomino:
Fixes #221.


kentonv:
Thanks!

PR #226, opened 2015-07-17
Update RPC docs for the change to a single bootstrap interface

geofft:
9a32fa73 changed the way that getting an initial / public capability works, but the online docs were not updated to match.

Probably rpc.md (the language-agnostic docs) could also use a sentence or two talking about how you get an initial capability, how you should authenticate users, etc.


kentonv:
Thanks!

PR #232, opened 2015-08-02
Fix bug dereferencing nullptr when forwarding a large skip

skarab:


dwrensha:
Eek, good catch.

I think the correct fix is to do `inner.skip(to_skip);` but not to add the line `size_t to_skip = bytes - bufferAvailable.size();`, as we've already performed that subtraction earlier in the method.

kentonv:
In fact, this won't dereference any null pointer, because assigning an array to `nullptr` is equivalent to setting it to a zero-size array, so `.size()` then returns zero. So the code will actually subtract zero from `bytes`, which as @dwrensha points out is coincidentally exactly correct. If I'm reading right, then by accident, there is no bug here, just some dead code. :)

@skarab Did you observe an actual bug in practice or did you catch thing just by reading through the source code? (It's still a good catch either way.)

skarab:
I just caught it reading through the code, so there is no practical observation. And upon closer examination, I agree that I was mistaken with the magnitude of the skip in my patch. I'll submit another pull request with the dead code removal.

skarab:
The new patch simply removes dead code.

kentonv:
Thanks!

Nice to have people reading through the code. :)

PR #237, opened 2015-08-26
Fix for `CAPNP_GENERATE_CPP` function output path mismatch

MarPiRK:
Before, when no `CAPNPC_OUTPUT_DIR` was defined, output path was ambiguous (set to `.` for custom command).
`SOURCES` and `HEADERS` were statically set to input file name (plus suffix) without concidering where custom command will place generated files.
Now if no `CAPNPC_OUTPUT_DIR` is defined, it's set by default to `CMAKE_CURRENT_BINARY_DIR` (this is basically `.`) and prevents any ambiguity.

### DIFF
--- a/c++/cmake/FindCapnProto.cmake
+++ b/c++/cmake/FindCapnProto.cmake
@@ -72,12 +72,12 @@ function(CAPNP_GENERATE_CPP SOURCES HEADERS)
     endforeach()
   endif()
 
-  if(DEFINED CAPNPC_OUTPUT_DIR)
-    # Prepend a ':' to get the format for the '-o' flag right
-    set(output_dir ":${CAPNPC_OUTPUT_DIR}")
-  else()
-    set(output_dir ":.")
-  endif()
+  if (NOT DEFINED CAPNPC_OUTPUT_DIR)
+    set (CAPNPC_OUTPUT_DIR "${CMAKE_CURRENT_BINARY_DIR}")
+  endif()  
+  
+  # Prepend a ':' to get the format for the '-o' flag right
+  set(output_dir ":${CAPNPC_OUTPUT_DIR}")
 
   if(NOT DEFINED CAPNPC_SRC_PREFIX)
     set(CAPNPC_SRC_PREFIX "${CMAKE_CURRENT_SOURCE_DIR}")
@@ -91,20 +91,16 @@ function(CAPNP_GENERATE_CPP SOURCES HEADERS)
     get_filename_component(file_dir "${file_path}" PATH)
 
     # Figure out where the output files will go
-    if (NOT DEFINED CAPNPC_OUTPUT_DIR)
-      set(output_base "${file_path}")
-    else()
-      # Output files are placed in CAPNPC_OUTPUT_DIR, at a location as if they were
-      # relative to CAPNPC_SRC_PREFIX.
-      string(LENGTH "${CAPNPC_SRC_PREFIX}" prefix_len)
-      string(SUBSTRING "${file_path}" 0 ${prefix_len} output_prefix)
-      if(NOT "${CAPNPC_SRC_PREFIX}" STREQUAL "${output_prefix}")
-        message(SEND_ERROR "Could not determine output path for '${schema_file}' ('${file_path}') with source prefix '${CAPNPC_SRC_PREFIX}' into '${CAPNPC_OUTPUT_DIR}'.")
-      endif()
-
-      string(SUBSTRING "${file_path}" ${prefix_len} -1 output_path)
-      set(output_base "${CAPNPC_OUTPUT_DIR}${output_path}")
+    # Output files are placed in CAPNPC_OUTPUT_DIR, at a location as if they were
+    # relative to CAPNPC_SRC_PREFIX.
+    string(LENGTH "${CAPNPC_SRC_PREFIX}" prefix_len)
+    string(SUBSTRING "${file_path}" 0 ${prefix_len} output_prefix)
+    if(NOT "${CAPNPC_SRC_PREFIX}" STREQUAL "${output_prefix}")
+      message(SEND_ERROR "Could not determine output path for '${schema_file}' ('${file_path}') with source prefix '${CAPNPC_SRC_PREFIX}' into '${CAPNPC_OUTPUT_DIR}'.")
     endif()
+
+    string(SUBSTRING "${file_path}" ${prefix_len} -1 output_path)
+    set(output_base "${CAPNPC_OUTPUT_DIR}${output_path}")
 
     add_custom_command(
       OUTPUT "${output_base}.c++" "${output_base}.h"


kentonv:
@pqu and/or @joshuawarner32, can you have a look at this? Thanks!

joshuawarner32:
I don't happen to have any projects that pull in capnproto in this fashion - I've previously just used direct cmake add_subdirectories, copy-pasta, and git submodules.  Nasty, I know.

One thing that might make testing this enormously easier is having a simple example capnproto project in the source tree that uses cmake, and pulls in capnproto in the "recommended fashion".  That can serve both as a template for people starting capnproto-based projects, and a test for the build/install infrastructure.  Extra points for using something like Docker to easily & cheaply test that across a wide range of distros.

Thoughts?

pqu:
The intent was for the default (`.`) to be the same directory as the `.capnp` file, but it looks like wasn't what `.` meant – it should've been `CMAKE_CURRENT_LIST_DIR`. It's debatable which should be the default (`CMAKE_CURRENT_LIST_DIR` avoids having to mess with paths, but `CMAKE_CURRENT_BINARY_DIR` doesn't pollute the build tree), but I suppose that's why it's configurable. LGTM.

A sample CMake project isn't much more complex than the example at the top of the package file (except for sorting out the paths...). Perhaps a CMakeLists.txt for the address book/calculator samples?

MarPiRK:
I'll create it tomorrow, or on Monday.
28 sie 2015 05:25 "Philip Quinn" <notifications@github.com> napisał(a):

MarPiRK:
There it is.

MarPiRK:
There's another problem with `FindCapnpProto.cmake`. In my system CMake states it hasn't been found:
```
-- checking for module 'capnp'
--   package 'capnp' not found
-- Found CAPNP: /usr/local/include  
-- Configuring done
-- Generating done
-- Build files have been written to: /tmp/samples/build
```

Everything builds just fine, but I'd like to use `feature_summary` in my project and it throws, that:
```
-- The following REQUIRED packages have not been found:

 * CapnProto
```

MarPiRK:
After last commit it's working.

joshuawarner32:
That's awesome!  Tested & working on OS X 10.10.

kentonv:
I remain cmake-illiterate, so @joshuawarner32 let me know when I should merge this. :)

kentonv:
Hmm a couple issues not directly related to the cmake code:

* I see you added a BSD 3-clause license in the samples/cmake directory. I would somewhat rather not have a subdirectory which is gratuitously licensed under a different (but functionally equivalent) license. Could you instead add your name to `CONTRIBUTORS` at the top level, and let the same MIT license apply? Note that you'd still own copyright on your contributions.
* I'd prefer to leave `test.sh` as-is, as it is important to me to test that people can build Cap'n Proto without using cmake. I think it would be great to add a second `test-cmake.sh` script which I'd be happy to add to my test suite.

joshuawarner32:
I was only planning on testing on my primary dev platform (osx), since I'm not relying on this method of importing yet.

So... :+1:

MarPiRK:
Done

c++/samples/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
+cmake_minimum_required (VERSION 2.8.12)
+
+project ("Cap'n'proto sample" CXX)
+
+# Needed to include generated Cap'n'proto headers
+set (CMAKE_INCLUDE_CURRENT_DIR ON)
+
+# Include CMake scripts
+include (FindPkgConfig)
```
pqu: This isn't necessary/used?

c++/samples/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
...
+find_package (Capnp REQUIRED)
+add_definitions(${CAPNP_DEFINITIONS})
+
+# Compile .capnp files
+capnp_generate_cpp(CAPNPSRC_addressbook CAPNPHDR_addressbook "${PROJECT_SOURCE_DIR}/addressbook.capnp")
+capnp_generate_cpp(CAPNPSRC_calculator CAPNPHDR_calculator "${PROJECT_SOURCE_DIR}/calculator.capnp")
+
+# Includes
+include_directories (
+    "${PROJECT_SOURCE_DIR}"
```
pqu: Redundant with `CMAKE_INCLUDE_CURRENT_DIR` above.

c++/samples/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
...
+add_executable (calculator-server
+    calculator-server.c++
+    "${CAPNPSRC_calculator}"
+)
+target_link_libraries (calculator-server ${CAPNP_LIBRARIES})
+
+# Tests
+enable_testing()
+
+set (ADDRESSBOOK_EXEC "${PROJECT_BINARY_DIR}/addressbook")
```
pqu: Could use a generator here instead: `$<TARGET_FILE:addressbook>` (and similarly for the calculator).

c++/samples/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
...
+set (CMAKE_MODULE_PATH
+    "${PROJECT_SOURCE_DIR}/../cmake"
+    "${PROJECT_SOURCE_DIR}/cmake"
+)
+
+# Enable tests
+include(CTest)
+
+# Check for C++11 support
+include (C++11)
```
pqu: I'm not sure this is useful. The macro doesn't work with MSVC, and support for "-std=c++11" doesn't necessarily guarantee *enough* C++11 support.

The checks at the top of `kj/common.h` do a more thorough job, and will halt the build with an error message fairly quick.

c++/samples/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
...
+CheckCXX11(FATAL_ERROR)
+
+# Add "-Wall" to compiler flags
+set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall")
```
pqu: Add `-std=c++0x` and a guard for MSVC.

c++/samples/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
...
+
+# Find Cap'n'proto
+find_package (Capnp REQUIRED)
```
pqu: I don't understand why the module was renamed.

pqu: I don't understand why the module was renamed?

c++/samples/CMakeLists.txt
MarPiRK: `include (FindPkgConfig)` is indeed unneeded

c++/samples/CMakeLists.txt
MarPiRK: It's some kind of CMake counterintuitivity – `find_package` checks if `Capnp_FOUND` is set, before it checked `CapnProto_FOUND`. In FindCapnp.cmake pkg_config searches for `capnp` and if it's found, it sets `Capnp_FOUND`, thus it's easier to rename just find_package module, rather than all variables and pkg_config file.

c++/samples/CMakeLists.txt
MarPiRK: Not really. If I remove `CMAKE_INCLUDE_CURRENT_DIR` on my system it throws:
```
/home/marek/tmp/capnproto/c++/samples/addressbook.c++:39:31: fatal error: addressbook.capnp.h: Nie ma takiego pliku ani katalogu
```

c++/samples/CMakeLists.txt
MarPiRK: OK then, I thought it'd be cleaner this way, but if you want so no problem.

c++/samples/CMakeLists.txt
pqu: Sorry, I meant that you could remove `"${PROJECT_SOURCE_DIR}"` and retain `CMAKE_INCLUDE_CURRENT_DIR` (or, replace it with `"${PROJECT_BINARY_DIR}"` and remove `CMAKE_INCLUDE_CURRENT_DIR`).

c++/samples/CMakeLists.txt
```diff
@@ -0,0 +1,65 @@
...
+set (CMAKE_INCLUDE_CURRENT_DIR ON)
+
+# Include CMake scripts
+set (CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/../cmake")
+
+# Enable tests
+include(CTest)
+
+# Add "-Wall" to compiler flags
+set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -std=c++0x")
```
pqu: Wrap this in an `if(NOT MSVC)...endif()` block.

MarPiRK:
I've changed `C++11.cmake` script with MSVC in mind

pqu:
MSVC doesn't support (or require) `-std=c++0x` and it's inadvisable to compile using `-Wall` (see the setup in the main `CMakeLists.txt` file).

(My concern that a C++11 check has no real utility remains.)

MarPiRK:
Are there any errors or something if compiler has `-std=c++0x` flag? `-Wall` had been passed before – in `test.sh`. I've just copied what I found there. I see there `-std=c++11` too, so maybe it'll have effect for MSVC?
There is also other possibility: if we're not that much concerned about the widest compatibility possible, we can force using CMake ≥3.1 and use `target_compile_features` – it's very convenient, but believe me or not, some distros have CMake 2.8 in repos, so people using these distros won't have possibility to easily compile this sample.
CMake C++11 check is just cleaner than throwing error during compilation – CMake's task is to check system's configuration **before** compilation, so, in my opinion, it should literally check for everything needed to compile **without any errors**.

pqu:
It will at least issue a warning for the unrecognised flag. Passing `Wall` doesn't cause any errors, but MSVC's `Wall` really means *all* warnings – including informational ones (struct packing, inline attempts, etc.).

I don't think that the build requirements for a sample project should be stricter than the main project itself. The current build is quite relaxed and the requirements are well documented.

While I agree with you that errors during compilation due to missing prerequisites are generally bad and can be confusing, in this case the error messages are very clear and specific. Completely testing all preconditions for a build would mean duplicating (and maintaining) the checks at the top of `kj/common.h`.

Also, if a user at the stage of building the samples, they must have already built the main library without issue.

MarPiRK:
1. Is that really bad thing? It was already in `test.sh`.
2. I understand and agree.
3. It's possible to check if `kj/common.h` compiles with `CHECK_CXX_SOURCE_COMPILES`.
4. I think that it shouldn't be assumed – this person may have downloaded binary, we don't know.

Besides, MSVC has C++11 turned on by default (http://stackoverflow.com/a/20167247), but for other compilers appropriate flag has to be added.

pqu:
It doesn't appear that `test.sh` was ever intended to work with MSVC.

I think `CHECK_CXX_SOURCE_COMPILES` is a much more robust way to go about it, and might be useful in the main library build. I'm not sure it should be recommended that *every* project using capnproto do the same thing.

MarPiRK:
I've done requested changes to C++11 check.

Here:
>  I'm not sure it should be recommended that every project using capnproto do the same thing.
you mean to include `kj/common.h` check in main `FindCapnp.cmake`?

pqu:
That too. I meant that I believe it's overkill for clients to be doing (or including) such a check.

MarPiRK:
Why? Does it harm anyone? In my opinion more checks = more confidence in clean build. It's not Autotools, that it takes ages to complete. It's neat and fast.

MarPiRK:
Do I understand correctly that Lite version doesn't require C++11?

MarPiRK:
I moved all C++11 checks to `FindCapnp.cmake`, so it's transparent for end user. It adds required compiler flags and checks if `kj/common.h` compiles.

MarPiRK:
Any more thoughts about this PR?

kentonv:
(I'm waiting on @pqu to say he's happy.)

pqu:
Sorry about the delay. I think my comments make it clear that I disagree with the excessive compiler checking. In particular, the latest patch now requires all users of the `FindCapnp` module to also satisfy a C++11 and `CHECK_CXX_SOURCE_COMPILES` check—even if neither is relevant to their use of the module (this is a very unusual thing for a CMake module to do). Similarly, I think a sample project should demonstrate the essentials of integrating the module, and limit extraneous material (although I should be clear that this is not a concern about technical correctness).

The patches from commits 213faff and 5b18062 (watch the `.gitignore`) LGTM, but I do not agree with the changes introduced by the others.

MarPiRK:
About `C++11.cmake`: I see your point about external module, but it can be directly integrated into the `FindCapnp.cmake` (I thought it'd be cleaner if it was separate, but in reality, as you said, it's not).  Why you disagree with, as you called it, excessive checks of compiler? If I use CMake module I want it to check all requirements *before* actual compilation process, to avoid misleading compilation errors and I don't want to care about some additional dependencies of that module. If Cap'n Proto requires C++11, then in my opinion `FindCapnp.cmake` should make test for it. 
Why don't check if `SOURCE_COMPILES`? If it adds some extra confidence that everything will build cleanly it's not overkill. I would understand your concern, if it took few minutes to pass, but on my computer it takes less than a second.

About sample project, remember that the old `test.sh` is still in tree - it's the essential approach. But if I'd want to integrate Cap'n Proto in my project, I'd like to see clear way to do it without too much effort. If I'm using CMake, I should have sample CMake project as excessive as possible. The same applies for Autotools project, but I'm not really into it, so I haven't included it in my PR.

pqu:
As far as `C++11.cmake` is concerned, it's not a check for "all requirements", it's a check for the `-std=c++11` flag, which is not sufficient to successfully build (you can pass the check and still fail to build). The `CHECK_CXX_SOURCE_COMPILES` check is more robust (and supersedes `C++11.cmake`?), but it requires compiler flags to be setup before the module is loaded and you could still get into trouble if it passes (e.g., forgetting about "lite" mode).

But even if we accept "extra confidence" as being valuable and worth maintaining, the errors that would be caught by these these checks are not "misleading", and are much more informative than "Your compiler probably doesn't support C++11 enough". The library is fairly diligent in checking its preconditions and reporting them with an `#error` or `#warning`.

I don't know how to resolve "clear way to do it without too much effort" with "as excessive [expressive?] as possible".

MarPiRK:
Why it is not? It's checking for C++11 flag for pretty much every compiler you could imagine. I see in main `c++/CMakeLists.txt` that for MSVC there is lite version enabled (I haven't noticed it earlier, but it's no problem to add `-DCAPNP_LITE` flag in `FindCapnp.cmake` to `CMAKE_CXX_FLAGS` for MSVC) and there is `-std=gnu++11` requirement for other compilers (which **is** checked by C++11 module). I think that compilers which don't support GNU extensions should have set supported C++11 flag (`-std=c++11` or `-std=c++0x`) and `-DCAPNP_LITE`.

Potential error or warning messages from `CHECK_CXX_SOURCE_COMPILES` are put into `build/CMakeFiles/CMakeError.log`. It can be mentioned in configuration-time message, where to look for them, but they shouldn't really appear, if all flags were set appropriately, so I don't insist to include that particular check.

In _as excessive as possible_ sentence, I meant that I'd like to be sure that CMake module will run, pass necessary checks on all supported platforms and add needed flags, without too much effort to implement it in main project's `CMakeLists.txt`. If I include package via `find_package (Capnp)`, I'd like to have all needed checks done, without any kind of extra compatibility work. It's in my point of view one of advantages of CMake – easily addable dependencies. Let's keep it that way for Cap'n proto, too.

joshuawarner32:
@pqu - is it possible to get this merged?  IMHO, this does improve the general state of affairs, and while it's not perfect, we can always come back and _improve it again_ at a later date.  I don't want to see this just bitrot.

pqu:
Unfortunately, the current version of the patch breaks the CMake build when performing a clean install.

MarPiRK:
What exactly do you mean? I've just clean cloned it and everything builds just fine.

pqu:
When I try to run `cmake` on a clean system (i.e., without Cap'n Proto already installed), it fails with the error message "kj/common.h compile check failed! Your compiler probably doesn't support C++11 enough".

MarPiRK:
It seems it's not just problem with this test, but previous test for capnp presence in system, which should throw error message before.

MarekPikula:
I was looking through some old PRs I have open and noticed that this one hasn't been merged yet. Let me know if there is an interest in merging it, as the code base has changed over those 8 (sic!) years since I submitted this PR. I'm not currently involved with capnproto, so I'm not even sure if the proposed change is still applicable. If not, I could close this PR.

MarekPikula:
Sounds good 👍 I actually have no idea what was it about anymore, and I don't use capnproto at the moment. Let's close it.

PR #239, opened 2015-08-29
Add observing for urgent data to UnixEventPort::FdObserver

mologie:
Hello,

this patch adds support for observing availability of urgent/out-of-band data by adding a new promise to FdObserver. The patch itself is simple and without side effects: Unless OBSERVE_URGENT is passed when constructing the FdObserver object, no new code is run. The implementation mirrors what was already there for the POLLIN event.

Right, so... why do we need this? A little backstory: I gave Cap'n Proto a shot as RPC server for my home automation system, which controls various devices through a Raspberry Pi B+'s GPIO ports and a i2c GPIO multiplexer. Cap'n Proto did the job flawlessly! However, in order to receive interrupts from the i2c GPIO multiplexer, I have to test for the POLLPRI flag on the interrupt port's fd. Linux's GPIO abstraction layer sets POLLPRI when an edge change is detected on a GPIO port. Extending KJ's UnixEventPort implementation appeared to be the most simple way to achieve this goal and at the same time integrate with the RPC server's event loop. POLLPRI is also used by TCP and many drivers for specialized hardware (such as Bluetooth receivers.)

I would appreciate if this change could be merged so that I do not have to maintain a separate fork. 

I have included a test which verifies that the new event works correctly with Linux's loopback TCP/IP out-of-band message implementation.

Cheers,
Oliver


kentonv:
Hi! Sorry for the delay; we launched the Sandstorm beta on Monday and I've been fire-fighting since then.

This code looks pretty great. The one question I have is about the test: the comment implies that MSG_OOB is a Linux-only feature, but I'm pretty sure I used it on Mac when I uncovered this bug:

https://blog.sandstorm.io/news/2015-04-08-osx-security-bug.html

I would expect it to be available on Unix in general. The implementation is not totally consistent but for the case of sending a single byte I think it should generally work. Have you found otherwise in practice?

kentonv:
Hmm, I think the test could be simplified to use a socketpair (rather than bind, listen, connect, accept), and could avoid using a thread (since you can safely write a byte in the main thread without blocking). A TCP socketpair should still be able to test OOB.

mologie:
Hey kentonv! I just gave the test a shot under OS X 10.11 preview 7. The interface appears to be supported / nothing fails explicitly, but the test gets stuck for no apparent reason. I will investigate further.

`socketpair()` is only implemented for AF_LOCAL/AF_UNIX on Linux and OS X. Local sockets do not support out-of-band messages and cannot be used for this test.

kentonv:
Ah, for some reason I thought socketpair worked for TCP.

As written the test will get stuck if any of the checks fail after the thread is created, because an exception will be thrown and then the thread's destructor will wait forever for the thread to complete. You can fix this by moving the final one-byte write into a KJ_DEFER statement appearing immediately after the thread is created -- this ensures that the byte is written when the function finishes even on an exception.

mologie:
Thanks for the tip, I will do so when I figure out what is wrong. However, this will not fix all cases where the test could get stuck: A syscall in the server thread may throw an exception (so that the OOB message never arrives), or something else could go wrong so that the main thread is never unblocked. Adding timers may help, but I found that this would make the test too complex for what is actually tested. Other tests appear to suffer from the same issue, so I ignored the possibility of it getting stuck.

Back to the original issue: The thing here is that no exception is thrown at all. The OOB message is simply never received on the main thread on OS X.

mologie:
Hey kentonv, I'm just dropping a note to tell you that I'm still alive and have started working on my project which concerns this issue again.

I will look into the OS X TCP socket test issue when I get the chance to - probably somewhen in February or March. This issue calls for diving into kernel: The behavior I observe starts to smell like an implementation issue of XNU's loopback TCP sockets. If this turns out to be the case, then I'd like to ask for including this feature in 'good faith' and keep the test Linux-only. POLLPRI is a generic interface with very implementation-specific behavior. There is little point in testing anything specific other than confirming that the generic integration with libkj works.

Cheers, Oliver

kentonv:
Cool, sounds good to me! Ping me when this is ready to look at again.

mologie:
All done! The issue was that OS X, for some reason, would never signal POLLPRI when sending an OOB byte unless in-band data was also received on the socket and has not been read yet. I've also reshuffled some code to allow the test to fail gracefully when an exception is thrown anywhere. SO_REUSEADDR is set on the server socket so that running the test multiple times with short delay does not fail. The test now passes on OS X 10.11.4 and Linux 4.2.0.

kentonv:
Hi Oliver,

Thanks! It would be great if you could also do the KJ_DEFER trick I mentioned so that the test doesn't get stuck if the main body throws an exception. That is, do something like:

    Thread thread([&]() {
      ...
    });
    c = 'q';
    KJ_DEFER(KJ_SYSCALL(send(clientFd, &c, 1, 0)));

This way, the 'q' byte will always be written to the pipe regardless of whether the scope exits normally or throws an exception.

mologie:
That I've already taken care of: The clientFd's destructor will be invoked prior to the server thread's destructor, which will close the socket, which will unblock the server by making the read syscalls fail. Assuming that the exception thrown in the server thread is properly handled by the test suite, this should make the test fail gracefully without explicitly requiring the KJ_DEFER trick for cleaning up.

mologie:
While looking over this again, I've noticed that things can still get stuck if `connect` fails after the server thread was started. This could happen if the fd resource limit is hit for whatever reason, or if the Unix gods are angry at us.

To solve this, I could insert a KJ_DEFER which closes serverFd and is cancelled after connect succeeded (or I could just explicitly check the error code), but I'm not sure whether it is legal to close an fd which another thread is currently doing a blocking system call on.

kentonv:
Ah, you're right, I didn't catch that the connect FD is now being closed on the way out.

> I'm not sure whether it is legal to close an fd which another thread is currently doing a blocking system call on.

It's legal, but it won't help you. What actually happens is that when a blocking syscall is entered, the syscall effectively dup()s the file descriptor, and then on the way out it closes its dup. So if you close the file descriptor in another thread, the syscall still has its own internal reference and keeps going. You would additionally need to send a signal to wake up the thread, which is ugly.

I'm not sure there is any reasonable way to make this totally safe, but I'm happy enough with what you have.

Meanwhile, though, this has called my attention back to the fact that we're relying on port 44444 to be open, which is kind of sad.

There's another option: If you bind() to port zero, the kernel will automatically assign an unused port. You can read the port back using getsockname(). Would you mind changing the test to do that?

mologie:
> It's legal, but it won't help you. What actually happens is that when a blocking syscall is entered, the syscall effectively dup()s the file descriptor, and then on the way out it closes its dup. So if you close the file descriptor in another thread, the syscall still has its own internal reference and keeps going. You would additionally need to send a signal to wake up the thread, which is ugly.

Interesting, thank you for the detailed explanation. After a bit of research, the Stackoverflow gurus have revealed that calling `shutdown` on the server socket will correctly cancel the `accept` call under Linux and OS X. I've tested this under OS X 10.11.4 and Linux 4.2 with success. Although one could argue that this behavior may still rely on implementation details of the kernel, it sounds like a good safety layer to apply here without increasing complexity of the test, because there are no side effects if no exception is thrown.

> There's another option: If you bind() to port zero, the kernel will automatically assign an unused port. You can read the port back using getsockname(). Would you mind changing the test to do that?

The static port bugged me too and I was not aware of this option. Sure, I'll change it.

kentonv:
Ah, sure, calling `shutdown()` seems reasonable.

c++/src/kj/async-unix-test.c++
```diff
@@ -467,16 +467,15 @@ TEST(AsyncUnixTest, UrgentObserver) {
...
   kj::AutoCloseFd serverFd(tmpFd);
   sockaddr_in saddr = {0};
   saddr.sin_family = AF_INET;
   saddr.sin_addr.s_addr = htonl(INADDR_LOOPBACK);
-  saddr.sin_port = htons(44444);
-  int optval = 1;
-  KJ_SYSCALL(setsockopt(serverFd, SOL_SOCKET, SO_REUSEADDR, &optval, sizeof(optval)));
   KJ_SYSCALL(bind(serverFd, reinterpret_cast<sockaddr*>(&saddr), sizeof(saddr)));
+  socklen_t saddrLen = sizeof(saddr);
+  KJ_SYSCALL(getsockname(serverFd, reinterpret_cast<sockaddr*>(&saddr), &saddrLen));
```
kentonv: Technically your later call to `connect()` should now use `saddrLen`, in case for some reason it's different than `sizeof(saddr)` (won't ever be in practice, I suppose...).

mologie:
Well, I did mess up [the quote](https://www.youtube.com/watch?v=hou0lU8WMgo), but the rest is, as they say, history.

kentonv:
Oh yeah, and you should add yourself to `/CONTRIBUTORS`. :)

mologie:
Thank you for the offer, done. I've reviewed everything again and cannot find anything else amiss, so that would be it from my side. Let me know if there's anything else that has to be taken care of. It was a pleasure working with you. :)

kentonv:
Thanks!

PR #240, opened 2015-08-30
Added method accept() to TwoPartyServer for servicing arbitrary AsyncIoStreams.

chris-turner137:
This patch makes TwoPartyServer more useful. TwoPartyServer could previously only be used with AsyncIoStreams accepted by a ConnectionReceiver listener. Two-way pipes (socketpairs) for example can't be accepted in this way.

It (appears to) work(s) correctly in my application which is providing one end of a socketpair.


kentonv:
Seems reasonable, thanks! (Sorry for the delay, we launched the Sandstorm beta on Monday and I've been firefighting.)

PR #241, opened 2015-09-01
Increase arenaSpace to fix 32-bit builds.

pqu:
Builds on 32-bit systems (x86 and armv7) fail because `arenaSpace` is 4 bytes too small. 

The size of `arenaSpace` was last increased in 62c0270, but it looks like the change to `BuilderArena::localCapTable` in 5413038 introduced a new vtable pointer.


kentonv:
Thanks!

Issue #220, opened 2015-07-01
IPV6_V6ONLY=0 not supported on OpenBSD


mrdomino:
SocketAddress::bind tries to disable IPV6_V6ONLY to allow a single socket to service both v4 and v6 requests (afaict just for wildcard sockets). This doesn't work on OpenBSD, where all sockets are either v4-only or v6-only.

One observable consequence of this is that the AsyncIo_SimpleNetwork test fails.

kentonv:
Hmm, that's unfortunate.

I think the short-term fix is to default to one or the other (I guess probably ipv4, unfortunately).

Longer-term, we should support binding multiple sockets and accepting from all of them. This is also useful for binding to `localhost`, which typically has both an ipv4 and an ipv6 address (currently, we bind to whichever one comes first in the result from getaddrinfo()).

Any interest in working on this? :)

mrdomino:
I wouldn't mind working on it, but usual caveats about dayjob-limited availability apply.

Would the short-term fix apply universally or just on OpenBSD? It's a trade-off between a regression on Linux and more code in compile-time conditionals I suppose.

kentonv:
I think the short-term fix should be limited to platforms that don't support the flag. Is `IPV6_V6ONLY` defined at all on OpenBSD? If not, #ifdef on that.

mrdomino:
It's defined but the setsockopt call fails with EINVAL.

PR #243, opened 2015-09-06
Use ip4 sockets for wildcard on OpenBSD

mrdomino:
Short-term fix for #220.

This just ifdefs on `__OpenBSD__`. Suggestions welcome for better ways of switching behaviour.

I opted to just stick this in the existing wildcard section rather than rearrange things and use the following af-conditioned block because this seemed a bit clearer.


kentonv:
Thanks!

PR #247, opened 2015-09-28
Add TextCodec to expose an API for the schema's text format.

pqu:
I've been looking at using the schema text format as a convenient way of specifying unit test cases that contain Cap'n Proto data, and as a runtime format for user configuration (already backed by a Cap'n Proto schema). But there's no API for reading these messages at runtime (aside from rolling them into a constant in a schema and using `SchemaParser`).

Reading the mailing lists, I see there's resistance to exposing an API for reading text messages because it has some undesirable effects if it's abused as a long-term storage or interchange format. This is understandable, but I think there are valid use cases for using the text format at runtime, and that the caveats are no worse than if JSON or the `SchemaParser` APIs are used instead. In particular, my use cases are similar to that described in [this thread](https://groups.google.com/d/msg/capnproto/nAYn8g9rn_w/Vi-vgcJ2xokJ):
 * Messages are supplied (and are editable) by users,
 * Schema definitions (and the capnp tool) are not distributed with the built application, and
 * Messages are not expected to be transmitted or be viable between versions.

While the `SchemaParser` supports reading this format, it expects to read a complete *schema*, which introduces a fair bit of setup overhead if only a single *message* needs to be read.

As with any other text format, this format is unable to meet the backwards-compatibility guarantees of the binary format—but there is some scope for customising errors, such as unknown fields, through the `ErrorHandler` (although there needs to be a better way to communicate the type of error from the `ValueTranslator`). But in my case, I prefer to treat these as hard-errors.

Also, keeping this in the `capnpc` library should hopefully deter people from using it as a robust alternative for messages.

### DIFF
--- a/c++/Makefile.am
+++ b/c++/Makefile.am
@@ -172,6 +172,7 @@ includecapnp_HEADERS =                                         \
   src/capnp/serialize.h                                        \
   src/capnp/serialize-async.h                                  \
   src/capnp/serialize-packed.h                                 \
+  src/capnp/serialize-text.h                                   \
   src/capnp/pointer-helpers.h                                  \
   src/capnp/generated-header-support.h                         \
   src/capnp/rpc-prelude.h                                      \
@@ -284,7 +285,8 @@ libcapnpc_la_SOURCES=                                          \
   src/capnp/compiler/node-translator.c++                       \
   src/capnp/compiler/compiler.h                                \
   src/capnp/compiler/compiler.c++                              \
-  src/capnp/schema-parser.c++
+  src/capnp/schema-parser.c++                                  \
+  src/capnp/serialize-text.c++
 
 bin_PROGRAMS = capnp capnpc-capnp capnpc-c++
 
@@ -382,6 +384,7 @@ heavy_tests =                                                  \
   src/capnp/dynamic-test.c++                                   \
   src/capnp/stringify-test.c++                                 \
   src/capnp/serialize-async-test.c++                           \
+  src/capnp/serialize-text-test.c++                            \
   src/capnp/rpc-test.c++                                       \
   src/capnp/rpc-twoparty-test.c++                              \
   src/capnp/ez-rpc-test.c++                                    \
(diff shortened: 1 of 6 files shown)


kentonv:
Hi Philip,

> While the `SchemaParser` supports reading this format, it expects to read a complete schema, which introduces a fair bit of setup overhead if only a single message needs to be read.

Are you sure that's not what you want? Using a proper schema file has a huge advantage over raw "text format": variable substitution. Meanwhile this "fair bit of setup overhead" seems pretty light to me: a file ID and a constant declaration.

I've had success with this approach in Sandstorm: apps define their metadata in the form of a file called `sandstorm-pkgdef.capnp`. For example:

https://github.com/kentonv/etherpad-lite/blob/sandstorm/sandstorm-pkgdef.capnp

This file is not parsed by the usual `capnp` tool, but rather by Sandstorm's `spk` tool, using `SchemaParser`. The ability to use variable substitution has proven surprisingly useful both for keeping the content organized and for cases of repetition.

You mention unit test cases as one use case. These tend to be highly repetitive. Would it not be useful to be able to define a set of variables for sub-objects that can be reused in each test case? (In fact, for a unit test, I would normally recommend compiling the constants into the binary rather than reading them at runtime, because why not?)

pqu:
`SchemaParser` is definitely an option that can be made to work, but it feels like a bit of a sledgehammer. That is, for messages I know will never need the complete features of the schema language (closer to glorified INI files than Sandstorm metadata files), a one-line decode API would help keep a simple task simple.

I'll grant that the requirements for dependent schema sources to be available at runtime and that files be complete schemas aren't onerous, and that everything I want to do can be functionally done by `SchemaParser`. I'm primarily trying to reduce the friction for utterly trivial messages here.

(I do make use of more complex constants at build-time through `capnp eval`, so these are a special case.)

kentonv:
OK, after sitting on it a week I decided you're right.

(I had actually forgotten that SchemaParser requires schema files available at runtime, which I do actually find really annoying. I would like to fix that by allowing it to import already-parsed schema nodes, but that's perhaps a bigger project.)

kentonv:
Thanks for writing this!

PR #252, opened 2015-10-06
Update link for Go library

TheHippo:
The Go library in the documentation is deprecated. The author apparently rewrote it from scratch and recommends to use the new one, which should be linked from your documentation now.


TheHippo:
Okay. Not a rewrite but the old library is still deprecated.

kentonv:
This is actually just a rename. The parenthetical that you removed still applies since it's the same code. I've updated the link but kept the parenthetical.

Thanks for the reminder!

PR #253, opened 2015-10-07
Add QtEventPort

nathanhourt:
Contributed code for an EventPort implementation which allows a KJ
event loop to co-exist with a Qt event loop on the same thread.

### DIFF
--- a/c++/src/kj/compat/QtEventPort.cpp
+++ b/c++/src/kj/compat/QtEventPort.cpp
@@ -1,0 +1,58 @@
+// Copyright (c) 2015 Follow My Vote, Inc.
+// Licensed under the MIT License:
+//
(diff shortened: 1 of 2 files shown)


kentonv:
Thanks!

Things to do before we can merge this:

1. Need to match KJ style. So:
    * File names should probably be `async-qt.{h,c++}`.
    * `namespace kj`
    * Preferably follow KJ formatting rules: https://github.com/sandstorm-io/capnproto/blob/master/style-guide.md#irrelevant-formatting-rules

2. Needs build rules of some sort. Ideally we build an additional library kj-async-qt.so if qt development headers are detected to be present at configure time (with the ability to explicitly enable or disable this library based on flags to `configure`.)

3. Needs a unit test.

4. Add yourself to CONTRIBUTORS. :)

PR #258, opened 2015-11-05
Implement most of JsonCodec::decodeRaw

kamalmarhubi:
This is the first step towards JSON decoding, implementing the basic
functionality of JsonCodec::decodeRaw. The main outstanding issues are:

- it allows trailing commas in arrays and objects
- it is too liberal in number syntax, eg allowing a leading +
- it does rejects non-ASCII characters in \u escapes

Refs https://github.com/sandstorm-io/capnproto/issues/255

### DIFF
--- a/c++/src/capnp/compat/json-test.c++
+++ b/c++/src/capnp/compat/json-test.c++
@@ -355,10 +355,44 @@ KJ_TEST("basic json decoding") {
     MallocMessageBuilder message;
     auto root = message.initRoot<JsonValue>();
 
-    // TODO(soon): this should fail, JSON doesn't allow leading +
-    json.decodeRaw("+123", root);
-    KJ_EXPECT(root.which() == JsonValue::NUMBER);
-    KJ_EXPECT(root.getNumber() == 123);
+    KJ_EXPECT_THROW_MESSAGE("input", json.decodeRaw("z", root));
+  }
+
+  {
+    MallocMessageBuilder message;
+    auto root = message.initRoot<JsonValue>();
+
+    // Leading + not allowed in numbers.
+    KJ_EXPECT_THROW_MESSAGE("Unexpected", json.decodeRaw("+123", root));
+  }
+
+  {
+    MallocMessageBuilder message;
+    auto root = message.initRoot<JsonValue>();
+
+    KJ_EXPECT_THROW_MESSAGE("Overflow", json.decodeRaw("1e1024", root));
+  }
+
+  {
+    MallocMessageBuilder message;
+    auto root = message.initRoot<JsonValue>();
+
+    KJ_EXPECT_THROW_MESSAGE("Underflow", json.decodeRaw("1e-1023", root));
+  }
+
+  {
+    MallocMessageBuilder message;
+    auto root = message.initRoot<JsonValue>();
+
+    KJ_EXPECT_THROW_MESSAGE("Unexpected", json.decodeRaw("[00]", root));
+    KJ_EXPECT_THROW_MESSAGE("Unexpected", json.decodeRaw("[01]", root));
+  }
+
+  {
+    MallocMessageBuilder message;
+    auto root = message.initRoot<JsonValue>();
+
+    KJ_EXPECT_THROW_MESSAGE("Expected number", json.decodeRaw("-", root));
   }
 
   {
--- a/c++/src/capnp/compat/json-test.c++
+++ b/c++/src/capnp/compat/json-test.c++
@@ -382,7 +382,7 @@ KJ_TEST("basic json decoding") {
     MallocMessageBuilder message;
     auto root = message.initRoot<JsonValue>();
 
-    KJ_EXPECT_THROW_MESSAGE("Expected number", json.decodeRaw("-", root));
+    KJ_EXPECT_THROW_MESSAGE("ends prematurely", json.decodeRaw("-", root));
   }
 
   {
@@ -426,8 +426,8 @@ KJ_TEST("basic json decoding") {
     KJ_EXPECT_THROW_MESSAGE("Invalid escape", json.decodeRaw(R"("\z")", root));
     KJ_EXPECT_THROW_MESSAGE("ends prematurely", json.decodeRaw(R"(["\n\", 3])", root));
     KJ_EXPECT_THROW_MESSAGE("Invalid hex", json.decodeRaw(R"("\u12zz")", root));
-    KJ_EXPECT_THROW_MESSAGE("Expected number", json.decodeRaw("-", root));
-    KJ_EXPECT_THROW_MESSAGE("Expected number", json.decodeRaw("--", root));
+    KJ_EXPECT_THROW_MESSAGE("ends prematurely", json.decodeRaw("-", root));
+    KJ_EXPECT_THROW_MESSAGE("Unexpected input", json.decodeRaw("--", root));
   }
 }
 
--- a/c++/src/capnp/compat/json-test.c++
+++ b/c++/src/capnp/compat/json-test.c++
@@ -185,21 +185,11 @@ KJ_TEST("encode union") {
 
 KJ_TEST("basic json decoding") {
   // TODO(cleanup): this test is a mess!
-  // TODO(soon): add expected failing cases.
   JsonCodec json;
   {
     MallocMessageBuilder message;
     auto root = message.initRoot<JsonValue>();
     json.decodeRaw("null", root);
-
-    KJ_EXPECT(root.which() == JsonValue::NULL_);
-    KJ_EXPECT(root.getNull() == VOID);
-  }
-
-  {
-    MallocMessageBuilder message;
-    auto root = message.initRoot<JsonValue>();
-    json.decodeRaw(kj::str("null"), root);
 
     KJ_EXPECT(root.which() == JsonValue::NULL_);
     KJ_EXPECT(root.getNull() == VOID);
@@ -411,6 +401,22 @@ KJ_TEST("basic json decoding") {
     json.decodeRaw("-5.5", root);
     KJ_EXPECT(root.which() == JsonValue::NUMBER);
     KJ_EXPECT(root.getNumber() == -5.5);
+  }
+
+  {
+    MallocMessageBuilder message;
+    auto root = message.initRoot<JsonValue>();
+
+    KJ_EXPECT_THROW_MESSAGE("Unexpected input", json.decodeRaw("a", root));
+    KJ_EXPECT_THROW_MESSAGE("ends prematurely", json.decodeRaw("[", root));
+    KJ_EXPECT_THROW_MESSAGE("ends prematurely", json.decodeRaw("{", root));
+    KJ_EXPECT_THROW_MESSAGE("Unexpected input", json.decodeRaw("[}", root));
+    KJ_EXPECT_THROW_MESSAGE("Unexpected input", json.decodeRaw("{]", root));
+    KJ_EXPECT_THROW_MESSAGE("Unexpected input", json.decodeRaw("[}]", root));
+    KJ_EXPECT_THROW_MESSAGE("Unexpected input", json.decodeRaw("[1, , ]", root));
+    KJ_EXPECT_THROW_MESSAGE("Unexpected input", json.decodeRaw("[,]", root));
+    KJ_EXPECT_THROW_MESSAGE("Unexpected input", json.decodeRaw("[, 1]", root));
+    KJ_EXPECT_THROW_MESSAGE("Input remains", json.decodeRaw("11a", root));
   }
 }
 
--- a/c++/src/capnp/compat/json-test.c++
+++ b/c++/src/capnp/compat/json-test.c++
@@ -199,6 +199,15 @@ KJ_TEST("basic json decoding") {
   {
     MallocMessageBuilder message;
     auto root = message.initRoot<JsonValue>();
+    json.decodeRaw(kj::str("null"), root);
+
+    KJ_EXPECT(root.which() == JsonValue::NULL_);
+    KJ_EXPECT(root.getNull() == VOID);
+  }
+
+  {
+    MallocMessageBuilder message;
+    auto root = message.initRoot<JsonValue>();
 
     json.decodeRaw("false", root);
     KJ_EXPECT(root.which() == JsonValue::BOOLEAN);
--- a/c++/src/capnp/compat/json-test.c++
+++ b/c++/src/capnp/compat/json-test.c++
@@ -380,6 +380,45 @@ KJ_TEST("basic json decoding") {
   }
 }
 
+KJ_TEST("maximum nesting depth") {
+  JsonCodec json;
+  auto input = kj::str(R"({"foo": "a", "bar": ["b", { "baz": [-5.5e11] }, [ [ 1 ], {  "z": 2 }]]})");
+  // `input` has a maximum nesting depth of 4, reached 3 times.
+
+  {
+    MallocMessageBuilder message;
+    auto root = message.initRoot<JsonValue>();
+
+    json.decodeRaw(input, root);
+  }
+
+  {
+    json.setMaxNestingDepth(0);
+    MallocMessageBuilder message;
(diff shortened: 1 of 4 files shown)


kamalmarhubi:
I intend to fix the trailing commas and expand \u escapes to cover the basic multilingual plane. I'm leaning towards being lazy about the number syntax, but can be persuaded. :-)

kamalmarhubi:
Oh and also: should there be a nesting depth limit? This could be a DoS issue...

kamalmarhubi:
Ideally I'd like to support surrogate pairs so that people can escape their emoji. However that can't be done inside of `std`. (Other chars can be covered by using wonderfully named [`c16rtomb`](http://en.cppreference.com/w/cpp/string/multibyte/c16rtomb).)

kamalmarhubi:
Another thing I need to do: handle overflow in parsing numbers.

c++/src/capnp/compat/json.c++
```diff
@@ -212,7 +222,7 @@ kj::String JsonCodec::encodeRaw(JsonValue::Reader value) const {
 }
 
 void JsonCodec::decodeRaw(kj::ArrayPtr<const char> input, JsonValue::Builder output) const {
-  KJ_FAIL_ASSERT("JSON decode not implement yet. :(");
+  _::parseJsonValue(input, output);
```
kentonv: Why the indirection to a separate function? Why not move the whole definition of `JsonCodec::decodeRaw()` down to where `parseJsonValue()` is defined?

c++/src/capnp/compat/json.c++
```diff
@@ -372,6 +382,251 @@ Orphan<DynamicValue> JsonCodec::decode(
...
+  void consume(kj::ArrayPtr<const char> expected) {
+    KJ_REQUIRE(remaining_.size() >= expected.size());
+
+    auto prefix = remaining_.slice(0, expected.size());
+    KJ_REQUIRE(prefix == expected, "Unexpected input in JSON message.");
+
+    advance(expected.size());
+  }
+
+  kj::ArrayPtr<const char> consumeWhile(kj::Function<bool(char)> predicate) {
```
kentonv: kj::Function adds virtual call overhead which could be expensive when calling for every byte. Instead, make this a template:

    template <typename Predicate>
    kj::ArrayPtr<const char> consumeWhile(Predicate&& predicate) {

c++/src/capnp/compat/json.c++
```diff
@@ -372,6 +382,251 @@ Orphan<DynamicValue> JsonCodec::decode(
...
+
+    for (auto i : kj::indices(fields)) {
+      object.adoptWithCaveats(i, kj::mv(fields[i]));
+    }
+
+    consume('}');
+  }
+
+  char nextChar() {
+    return remaining_.front();
```
kentonv: Need to require `remaining_.size() > 0`?

c++/src/capnp/compat/json.c++
```diff
@@ -372,6 +382,251 @@ Orphan<DynamicValue> JsonCodec::decode(
 
 // -----------------------------------------------------------------------------
 
+namespace _ {  // private
+
+class Parser {
+public:
+  Parser(kj::ArrayPtr<const char> input) : input_(input), remaining_(input_) {}
+  void parseValue(JsonValue::Builder &output) {
```
kentonv: Style nitpick: group '&' with the type, not the var name, i.e. `JsonValue::Builder& output`.

c++/src/capnp/compat/json.c++
```diff
@@ -372,6 +382,251 @@ Orphan<DynamicValue> JsonCodec::decode(
...
+      // relying on strtod.
+      default: parseNumber(output); break;
+    }
+  }
+
+  void parseNumber(JsonValue::Builder &output) {
+    // TODO(someday): strtod allows leading +, while JSON grammar does not.
+    // strtod consumes leading whitespace, so we don't have to.
+    char *numEnd;
+    output.setNumber(std::strtod(remaining_.begin(), &numEnd));
```
kentonv: Problems:
- If the first char is not a valid part of a number, then `strtod()` will return zero and set `numEnd` equal to `remaining_.begin()`, i.e. no input will be consumed. I think you'll end up in an infinite loop here. Please add a test.
- If `strtod()` reaches the end of the input, it will keep going past it, since our input buffer is not NUL-terminated.

I think you will need to write a scanner that actually matches the JSON number pattern as one step, then pass it to `strtod()` as a second step.

kentonv: Also I think `strtod()` will accept redundant leading zeros. The JSON spec disallows these presumably due to ambiguity with octal. It's probably best that we also disallow them, rather than possibly misinterpret them.

c++/src/capnp/compat/json.c++
```diff
@@ -372,6 +382,251 @@ Orphan<DynamicValue> JsonCodec::decode(
...
+      consume(':');
+      consumeWhitespace();
+
+      auto valueBuilder = builder.getValue();
+      parseValue(valueBuilder);
+
+      fields.add(kj::mv(orphan));
+
+      if (consumeWhitespace(), nextChar() != '}') {
+        // TODO(soon): This incorrectly allows a trailing comma.
```
kentonv: FWIW, I don't mind allowing a trailing comma, since it's totally unambiguous what it means.

c++/src/capnp/compat/json.c++
```diff
@@ -372,6 +382,251 @@ Orphan<DynamicValue> JsonCodec::decode(
...
+      } else if ('a' <= c && c <= 'f') {
+        codePoint |= c - 'a';
+      } else if ('A' <= c && c <= 'F') {
+        codePoint |= c - 'A';
+      } else {
+        KJ_FAIL_REQUIRE("invalid hex digit in unicode escape", c);
+      }
+    }
+
+    // TODO(soon): Support at least basic multi-lingual plane, ie ignore surrogates.
```
kentonv: Encoding a codepoint to UTF-8 is generally pretty easy (much easier than decoding), so I think we should go ahead and do it here.

Surrogates are annoying but probably not too hard either. FWIW, if unpaired surrogates are seen I think we should go ahead and encode them as separate UTF-8 sequences, that is to say we should actually produce WTF-8. But if they are properly paired we should probably decode them.

c++/src/capnp/compat/json.c++
```diff
@@ -372,6 +382,251 @@ Orphan<DynamicValue> JsonCodec::decode(
...
+};  // class Parser
+
+// Array literal used instead of string literal to avoid null terminator.
+const kj::ArrayPtr<const char> Parser::NULL_ = kj::ArrayPtr<const char>({'n','u','l','l'});
+const kj::ArrayPtr<const char> Parser::FALSE = kj::ArrayPtr<const char>({'f','a','l','s','e'});
+const kj::ArrayPtr<const char> Parser::TRUE = kj::ArrayPtr<const char>({'t','r','u','e'});
+
+
+void parseJsonValue(kj::ArrayPtr<const char> input, JsonValue::Builder output) {
+  // TODO(security): should we check there are no non-whitespace characters left in input?
```
kentonv: Seems like a good idea. (Not sure I'd call it a security issue, though.)

c++/src/capnp/compat/json-test.c++
```diff
@@ -181,6 +183,203 @@ KJ_TEST("encode union") {
   KJ_EXPECT(json.encode(root) == "{\"before\":\"a\",\"middle\":44,\"bar\":321,\"after\":\"c\"}");
 }
 
+KJ_TEST("basic json decoding") {
+  // TODO(cleanup): this test is a mess!
+  // TODO(soon): add expected failing cases.
```
kentonv: Definitely add some failing cases. :)

kentonv:
In general this is pretty great -- the code is clean and easy to read. Thanks!

On a procedural note, when you address the issues above, please make a new commit without squashing -- that way I can review the diffs separately from the code I've already read.

> Oh and also: should there be a nesting depth limit? This could be a DoS issue...

Yeah, that's probably a good idea. 64 has always worked well for me. Of course you can make it configurable via a method on `JsonCodec`.

> Ideally I'd like to support surrogate pairs so that people can escape their emoji. However that can't be done inside of std without using <locale>. (Other chars can be covered by using wonderfully named c16rtomb.)

It seems to me it should only take a few lines of code to handle surrogate pairs correctly without a library.

c++/src/capnp/compat/json.c++
kamalmarhubi: Done.

For some reason I decided the order was important...

c++/src/capnp/compat/json.c++
kamalmarhubi: Done.

c++/src/capnp/compat/json.c++
kamalmarhubi: Ah yes! Done.

c++/src/capnp/compat/json.c++
kamalmarhubi: Done.

c++/src/capnp/compat/json.c++
kamalmarhubi: Changed it to "someday" as a TODO.

kamalmarhubi:
Thanks for the review! I'll fix everything but the unicode escape handling on this PR, preferring to get this merged and push that to a later PR.

c++/src/capnp/compat/json.c++
kamalmarhubi: Will do in a followup PR.

c++/src/capnp/compat/json.c++
kamalmarhubi: Argh these are all good points, should have caught them!

Done. (Using `<regex>` for now; this can be fixed in the future.)

c++/src/capnp/compat/json.c++
kamalmarhubi: Done.

kamalmarhubi:
>> Oh and also: should there be a nesting depth limit? This could be a DoS issue...

> Yeah, that's probably a good idea. 64 has always worked well for me. Of course you can make it configurable via a method on JsonCodec.

Done.

c++/src/capnp/compat/json-test.c++
kamalmarhubi: Added some!

kamalmarhubi:
Please take another look!

As mentioned earlier, I'm pushing unicode stuff to a later PR. I think I've addressed the rest of the comments. Let me know if you spot anything else. Thanks!

c++/src/capnp/compat/json.c++
```diff
@@ -372,6 +379,313 @@ Orphan<DynamicValue> JsonCodec::decode(
...
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wglobal-constructors"
+#endif
+
+// Array literal used instead of string literal to avoid null terminator.
+const kj::ArrayPtr<const char> Parser::NULL_ = kj::ArrayPtr<const char>({'n','u','l','l'});
+const kj::ArrayPtr<const char> Parser::FALSE = kj::ArrayPtr<const char>({'f','a','l','s','e'});
+const kj::ArrayPtr<const char> Parser::TRUE = kj::ArrayPtr<const char>({'t','r','u','e'});
+
```
kamalmarhubi: For reasons, I ended up compiling with clang. It warns about these three lines requiring global constructors (ie, construction at program start). Given the triviality, I disabled the warning, but could do something else if you prefer.

kentonv: Even a trivial global constructor is a problem: it forces the code to be paged in from disk at startup, which can be a pretty significant cost.

What you want here is:

    const char Parser::NULL_[4] = {'n', 'u', 'l', 'l'};

A `T[]` can be implicitly converted to a `kj::ArrayPtr<T>`, so you should be able to pass the constant into `consume()`.

Or, better yet, don't use constants at all -- I don't think you need them here. You can do:

    consume(kj::StringPtr("null"))

That should do the right thing. `kj::StringPtr` implicitly converts to `kj::ArrayPtr<const char>`, and the NUL terminator is not included.

kentonv: (That was a random driveby comment -- unfortunately I don't have time to review all your updates at the moment, but will try to get to it soon.)

kamalmarhubi: Done, thanks for the suggestion.

c++/src/capnp/compat/json.c++
kentonv: Well, if we're ever going to change it I think it should be now, but I don't think we ever need to change it so I'd make it a note rather than a TODO.

c++/src/capnp/compat/json.c++
kentonv: I'd definitely like to avoid the `<regex>`. I'd rather fall back to `kj/parse/char.h` if we're going to use a parser framework, though simple recursive descent shouldn't be too hard here.

(KJ parsers have the advantage of being compiled at compile time, whereas regex will compile the expression at runtime (in fact, every time it parses a number, as the code is currently written).)

kentonv: I think the code could look something like:

    tryConsume('-');
    if (!tryConsume('0')) {
      consume([](char c) {'1' <= c && c <= '9'})`
      consumeWhile([](char c) {'0' <= c && c <= '9'})`
    }

    if (tryConsume('.')) {
      consumeWhile([](char c) {'0' <= c && c <= '9'})`
    }
    if (tryConsume('e') || tryConsume('E')) {
      tryConsume('+') || tryConsume('-');
      consumeWhile([](char c) {'0' <= c && c <= '9'})`
    }

kentonv:
OK I think I've covered everything up to this point!

c++/src/capnp/compat/json.c++
kamalmarhubi: Done as suggested. I'd started doing that but was (perhaps irrationally) unwilling to play the role of a regex compiler...

c++/src/capnp/compat/json.c++
kamalmarhubi: Decided to fix it instead. :-)

kamalmarhubi:
Ok, addressed those last things.

c++/src/capnp/compat/json.c++
```diff
@@ -372,6 +379,320 @@ Orphan<DynamicValue> JsonCodec::decode(
...
+      }
+    }
+
+    // TODO(soon): Support at least basic multi-lingual plane, ie ignore surrogates.
+    KJ_REQUIRE(codePoint < 128, "non-ASCII unicode escapes are not supported (yet!)");
+    target.add(0x7f & static_cast<char>(codePoint));
+  }
+
+private:
+  const size_t maxNestingDepth_;
```
kentonv: Hey, sorry to nitpick, but I just noticed you're doing Google-style trailing underscores on members. We don't use that style in KJ code -- could you remove them?

kamalmarhubi: Done.

kentonv:
OK, everything looks good. Can you add yourself to `CONTRIBUTORS`? :)

kamalmarhubi:
Done!

kentonv:
This change broke the build for us, fixed in b820678. I wonder why this didn't affect you? The enum returned by `.which()` doesn't have a stringifier.

kamalmarhubi:
That is weird. I was using `make -f Makefile.ekam continuous` and either of:

```
$ g++ --version
g++ (Debian 4.9.2-10) 4.9.2
Copyright (C) 2014 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

$ clang++ --version
Debian clang version 3.5.0-10 (tags/RELEASE_350/final) (based on LLVM 3.5.0)
Target: x86_64-pc-linux-gnu
Thread model: posix
```

PR #260, opened 2015-11-07
Fix UnixEventPort value on 32-bit big endian archs

thomaslee:
toRegularSiginfo attempts to convert signalfd_siginfo.ssi_ptr (a 64-bit
integer across all architectures) to a pointer. On 32-bit big endian
architectures, sival_int/sival_ptr will be stored in the high 32-bits
of the ssi_ptr value.

Prior to this change, the value was "lost" as we cast away the high
bits of ssi_ptr in the conversion to a 32-bit pointer.

This fixes AsyncUnixTest::SignalWith{,Pointer}Value on affected archs.

The comment in AsyncUnixTest::SignalWithValue mentions sigqueue() is broken on droid bionic -- not entirely clear to me if this is related & I don't have the hardware to check.

### DIFF
--- a/c++/src/kj/async-unix.c++
+++ b/c++/src/kj/async-unix.c++
@@ -458,7 +458,7 @@ static siginfo_t toRegularSiginfo(const struct signalfd_siginfo& siginfo) {
         // we write the pointer, we'll end up with the right value for the int? Presumably the
         // two fields of signalfd_siginfo are actually extracted from one of these unions
         // originally, so actually contain redundant data? Better write some tests...
-        result.si_ptr = reinterpret_cast<void*>(static_cast<uintptr_t>(siginfo.ssi_ptr));
+        memcpy(&result.si_ptr, &siginfo.ssi_ptr, sizeof(result.si_ptr));
         break;
 
       case SI_TIMER:


c++/configure.ac
```diff
@@ -52,6 +52,19 @@ AC_PROG_CXX
 AC_LANG([C++])
 AX_CXX_COMPILE_STDCXX_11
 
+AC_C_BIGENDIAN([CXXFLAGS="$CXXFLAGS -D_KJ_BIGENDIAN=1"],
+               [CXXFLAGS="$CXXFLAGS -D_KJ_BIGENDIAN=0"],
+               [AC_MSG_WARN([could not determine system endianness: assuming little endian])
+                CXXFLAGS="$CXXFLAGS -D_KJ_BIGENDIAN=0"])
```
thomaslee: The `AC_MSG_WARN` feels like something of a cop out, but I found it hard to justify breaking the build over something that only really seems to affect a small set of features.

thomaslee:
This is a direct copy of code I'm using to address this issue in Debian ports -- expect it may need some additional cleanup before merging.

kentonv:
Thanks Thomas.

If I understand correctly, the problem is that `siginfo.ssi_ptr` is `uint64_t` even on 32-bit systems, and the kernel surprisingly stores the 32-bit pointer in the first 4 bytes of the space used by `ssi_ptr`, rather than casting the pointer to a 64-bit integer and storing that. Essentially, it is incorrect to cast `ssi_ptr` to a pointer value -- you must `memcpy()` it. This behavior is undocumented, and to me very surprising, though I can perhaps imagine a contorted world view in which a developer thought this was the right way to do it. More likely, signalfd was insufficiently tested on 32-bit big-endian architectures.

In any case, I think that the easy fix is to replace the assignment with:

    memcpy(&result.si_ptr, &siginfo.ssi_ptr, sizeof(result.si_ptr));

This solution does not require detecting endianness and I think more clearly shows what is going on.

(BTW, it no longer matters for this change, but Cap'n Proto already has endianness checks based on default define probing, which I prefer over configure-time checking; see capnp/endian.h.)

thomaslee:
> If I understand correctly, the problem is that siginfo.ssi_ptr is uint64_t even on 32-bit systems, and the kernel surprisingly stores the 32-bit pointer in the first 4 bytes of the space used by ssi_ptr, rather than casting the pointer to a 64-bit integer and storing that. 

Pretty much. Looks like it might be happening on the way back out of the kernel: http://lxr.free-electrons.com/source/fs/signalfd.c?v=4.2#L151 (where `__put_user` seems to be a glorified memcpy between kernel & user space)

> In any case, I think that the easy fix is to replace the assignment with: (snip)

Can't think of any counter-examples that wouldn't work with the memcpy. Much cleaner, thanks!

> Cap'n Proto already has endianness checks based on default define probing, which I prefer over configure-time checking; see capnp/endian.h

I noticed those, but are they available to code in libkj? I had assumed not.

kentonv:
> I noticed those, but are they available to code in libkj? I had assumed not.

Hmm, true. Well, no matter now!

thomaslee:
My branch has been updated but give me just a moment to try this fix out on ppc to be sure the issue is fixed. Keep you posted.

thomaslee:
Tests pass on ppc -- thanks again!

kentonv:
Thanks!

paulproteus:
BTW, Tom -- I wonder if you'd be interested in setting up a Travis-CI
builder for capnproto ppc with qemu. It may seem like a completely absurd
thought, but I've done similar things and been surprisingly happy with the
results.

Kenton, I know you've expressed a concern that it's nice if the build
automation is all in one place -- Sandstorm Jenkins vs. Travis-CI. Maybe
initially Tom's thing should be in a separate repo that he controls, but in
the long run it might be nice to merge that.

thomaslee:
@paulproteus huh, had it in my head that qemu was just an x86 emulator, didn't realize it could emulate other archs. I'm curious enough to give it a shot sometime -- especially if breakages on more esoteric archs is interesting to you guys too. :smile:

PR #264, opened 2015-11-18
Minor cleanup

joliss:


kentonv:
Whoa did you actually run my test scripts? I kind of expected no one would ever look at them. :)

(Thanks for the cleanups!)

joliss:
Haha, I *tried* running them, but then it errored out for lack of root rights (and I wasn't feeling brave enough to use sudo :D).

PR #266, opened 2015-11-24
adjust the set of white space characters to the set described by RFC7159

eriksjolund:
Vertical tab and form feed are not specified as white space characters 
in RFC7159 but they are used as such in the parsing code:
c++/src/capnp/compat/json.c++
This patch corrects this.

### DIFF
--- a/c++/src/capnp/compat/json-test.c++
+++ b/c++/src/capnp/compat/json-test.c++
@@ -298,6 +298,15 @@ KJ_TEST("basic json decoding") {
     MallocMessageBuilder message;
     auto root = message.initRoot<JsonValue>();
 
+    json.decodeRaw("\r\n\t {\r\n\t }\r\n\t ", root);
+    KJ_EXPECT(root.which() == JsonValue::OBJECT, (uint)root.which());
+    KJ_EXPECT(root.getObject().size() == 0);
+  }
+
+  {
+    MallocMessageBuilder message;
+    auto root = message.initRoot<JsonValue>();
+
     json.decodeRaw(R"({"some": null})", root);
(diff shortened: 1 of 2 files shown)


kentonv:
Hey Erik,

I'm really curious to know what has motivated you to look for these obscure cases where the parser accepts technically-invalid (but unambiguous) input. Not that I'm against it, just wondering. :)

kentonv:
(And on a more practical note, probably you should extend the unit test to cover this.)

eriksjolund:
I just stumbled upon those things when looking at the code. In general I think it is easier
to fix bugs earlier than later in a development process. Once you start accepting non-standard variants of a file format, you need to be careful when you remove support for such things in the future. Another thing: The more varieties of input some software accepts, the more difficult it will be validate that the code-base is secure. And then you also have the issue that all non-standard extensions to a file format need be documented.

Anyway, I'm still learning how the JSON format works. I'll try to add some tests. It may take a week or two as I'm busy with other things.

kentonv:
OK, that's good enough for me.

I do think we should add a test to json-test.c++. This should be pretty easy, though.

eriksjolund:
I read the whole
https://tools.ietf.org/html/rfc7159
and was surprised to see:
"A JSON parser MAY accept non-JSON forms or extensions"

Anyway, I think it it still better to fail on non-JSON input than to silently accept the input.
I'll need to install Ekam to run the tests. I'll give you a notice when I have it done.

kentonv:
Thanks!

kamalmarhubi:
I was sure I'd got the list of whitespace characters from ECMA 404, but turns out I hadn't and it specifies the same list as the RFC. Thanks!

PR #268, opened 2015-12-05
Drop onlick() handlers in documentation, in favor of simple links.

dwrensha:
This should look and act exactly the same as the current version, except:

  1. The header of the currently-selected section is no longer clickable. (Currently it has an `onclick()` handler that maps it to "#".)
  2. You can now command-click (or middle-click or whatever) the links and have them open in a new tab, without changing the state of the current tab.

### DIFF
--- a/doc/_includes/header.html
+++ b/doc/_includes/header.html
@@ -41,18 +41,18 @@
 
       <section id="menu">
         <ul>
-          <li><a href="{{ site.baseurl }}index.html">Introduction</a></li>
-          <li><a href="{{ site.baseurl }}news/">News</a></li>
-          <li><a href="{{ site.baseurl }}install.html">Installation</a></li>
-          <li><a href="{{ site.baseurl }}language.html">Schema Language</a></li>
-          <li><a href="{{ site.baseurl }}encoding.html">Encoding</a></li>
-          <li><a href="{{ site.baseurl }}rpc.html">RPC Protocol</a></li>
-          <li><a href="{{ site.baseurl }}capnp-tool.html">The <code>capnp</code> Tool</a></li>
-          <li><a href="{{ site.baseurl }}cxx.html">C++ Serialization</a></li>
-          <li><a href="{{ site.baseurl }}cxxrpc.html">C++ RPC</a></li>
-          <li><a href="{{ site.baseurl }}otherlang.html">Other Languages</a></li>
-          <li><a href="{{ site.baseurl }}roadmap.html">Road Map</a></li>
-          <li><a href="{{ site.baseurl }}faq.html">FAQ</a></li>
+          <li><a href="{{ site.baseurl }}index.html"><p>Introduction</p></a></li>
+          <li><a href="{{ site.baseurl }}news/"><p>News</p></a></li>
+          <li><a href="{{ site.baseurl }}install.html"><p>Installation</p></a></li>
+          <li><a href="{{ site.baseurl }}language.html"><p>Schema Language</p></a></li>
+          <li><a href="{{ site.baseurl }}encoding.html"><p>Encoding</p></a></li>
+          <li><a href="{{ site.baseurl }}rpc.html"><p>RPC Protocol</p></a></li>
+          <li><a href="{{ site.baseurl }}capnp-tool.html"><p>The <code>capnp</code> Tool</p></a></li>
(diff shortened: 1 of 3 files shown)


paulproteus:
Thanks. Will read & review (and merge, I presume, since that's what I
presume it being assigned to me means) shortly, eta 1ish days.

doc/javascripts/main.js
```diff
@@ -44,32 +44,17 @@ function initSidebar() {
     var href = link.href;
     if (href.lastIndexOf(filename) >= 0) {
       var parent = link.parentNode;
-      var p = document.createElement("p");
 
       while (link.childNodes.length > 0) {
         var child = link.childNodes[0];
         link.removeChild(child);
-        p.appendChild(child);
+        parent.appendChild(child);
```
paulproteus: I can't say I'm excited about this uncommented JS code that does templating on the sidebar. @dwrensha any way I can convince you to add comments, as part of this pull request?

doc/_includes/header.html
```diff
@@ -41,18 +41,18 @@
...
-          <li><a href="{{ site.baseurl }}language.html">Schema Language</a></li>
-          <li><a href="{{ site.baseurl }}encoding.html">Encoding</a></li>
-          <li><a href="{{ site.baseurl }}rpc.html">RPC Protocol</a></li>
-          <li><a href="{{ site.baseurl }}capnp-tool.html">The <code>capnp</code> Tool</a></li>
-          <li><a href="{{ site.baseurl }}cxx.html">C++ Serialization</a></li>
-          <li><a href="{{ site.baseurl }}cxxrpc.html">C++ RPC</a></li>
-          <li><a href="{{ site.baseurl }}otherlang.html">Other Languages</a></li>
-          <li><a href="{{ site.baseurl }}roadmap.html">Road Map</a></li>
-          <li><a href="{{ site.baseurl }}faq.html">FAQ</a></li>
+          <li><a href="{{ site.baseurl }}index.html"><p>Introduction</p></a></li>
```
paulproteus: The reason to use a `p` tag here seems to be so that the generated DOM content (below in JS) creates the same DOM tree as the hard-coded HTML here.

I can't say I love the `p` tag.

I wonder if you can:

- Avoid it entirely, or

- Use a `span` to indicate that this has no particular meaning as an element

doc/javascripts/main.js
```diff
@@ -117,21 +102,15 @@ function setupSidebar() {
       var item = document.createElement("li");
       var p = document.createElement("p");
```
paulproteus: Can I suggest that rather than creating a `p` here, you append a text node to the inside of the `a` instead?

Then you could remove the `p` in the HTML (see my code review comment above) I believe.

doc/stylesheets/stylesheet.css
```diff
@@ -599,10 +599,15 @@ body.narrow #menu {
...
 #menu li {
   margin: 0;
-  padding: 10px 15px 10px 15px;
   list-style-type: none;
   background-color: #212121;
 }
+#menu>ul>li>a>p, #menu>ul>li.selected {
+  padding: 10px 15px 10px 15px;
+}
+#menu>ul>li>a>p {
```
paulproteus: If you can remove the `p`, I think this'll be noticeably cleaner. See above.

paulproteus:
General feedback:

- I'm +1 on this as-is, since it does what it says it does (I tested it locally thanks to https://github.com/sandstorm-io/capnproto/blob/master/doc/README.md yay!)

- I have some requests which are not show-stoppers.

Given that they're not show-stoppers I'm going to merge this immediately and @dwrensha can follow up on the requests, or not, at his option. 

(Note also that I'll be mostly away-from-keyboard until this coming Monday.)

doc/_includes/header.html
dwrensha: The main point of the `<p>` was to allow the links to be sized. In #269 I instead use `display:block` to achieve the same thing.

kentonv:
Whoa, I only just realized that this is against capnproto docs, not Sandstorm docs. I assigned @paulproteus based on thinking it was sandstorm docs. Sorry if I seemed like I was unfairly shoving work on you!

paulproteus:
No worries. It was kind of fun to see how the capnproto docs are built.​

PR #270, opened 2015-12-13
Check num remaining bytes of escaped json character

eriksjolund:
It seems the JSON parsing is potentially reading past the end of the input buffer.

I'm pretty confident the patch should work but I need to verify that by installing Ekam and run the tests. I'll probably be able to do that before the end of this month.


paulproteus:
This is a great pull request, Erik.

This kind of problem can sometimes be found by using "American Fuzzy Lop"
(AFL), a tool to generate invalid inputs and attempt to discover code paths
that crash. So I just wanted to add a comment saying that if you want to
try using AFL to test capnproto's JSON parsing, then the world would
probably be a better place! You don't have to by any means, but if it
sounds fun, then great.

eriksjolund:
"American Fuzzy Lop" looks cool. Unfortunately, there is a bit lack of time for me right now.

kentonv:
Thanks for the good catch!

kentonv:
@kamalmarhubi FYI

kamalmarhubi:
Ugh! Thanks for this @eriksjolund. This makes me realise that the contract *should* be that the `remaining` array is never touched outside of `advance` and `consume*` methods. I think I'll do a little encapsulation of the input array to make it harder to let issues like this creep in.

PR #271, opened 2015-12-16
json: Hide the input array from parser methods

kamalmarhubi:
This encapsulates the input array, hiding it from the parser's methods.
Access is forced to go through `advance` and `consume*` methods on the
new `Input` object. The aim is to make it harder to have errors like
https://github.com/sandstorm-io/capnproto/pull/270


kentonv:
Sorry for the delay -- holidays!

kamalmarhubi:
Holidays happen :-)

PR #272, opened 2015-12-16
Account for the possibility of adopting an empty struct.

dwrensha:
Fixes a bug reported by David Buckley in this thread: https://groups.google.com/forum/#!topic/capnproto/3MIU8xZBX1Q

Adds a test that fails before the fix and succeeds after the fix.


kentonv:
This change introduced a subtle message corruption bug, because it fails to zero the upper 32 bits of the pointer, which may contain any arbitrary garbage if the pointer is being adopted over some other pointer (the calling code is responsible for zeroing whatever the pointer previously pointed to, but not zeroing the pointer itself).

Fixed in 4d4f831.

Issue #273, opened 2015-12-23
Block syntax for functions taking code like KJ_DEFER / KJ_ON_SCOPE_*


mrdomino:
Using an implementation like that on slide 20 of [the uncaught_exceptions proposal](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4152.pdf) (i.e. an overloaded `operator+` of a functor type), we could write deferred blocks like:

```
KJ_DEFER { code; }
```

instead of the current `KJ_DEFER(code);`. It'd probably be a bit, but not much, more work to also do `KJ_ON_SCOPE_SUCCESS` and `KJ_ON_SCOPE_FAILURE` like this. Is it desirable to have this? If so, I can start on a pull request.

kentonv:
Hmm. Users will have to type a semicolon after the closing brace, that is:

    KJ_DEFER { code; };

This may be unintuitive, since it differs from usual control blocks. Note that `KJ_MAP` works this way, but it also evaluates to a value, which perhaps makes it less surprising.

Note that currently, the following syntax works:

    KJ_DEFER({
      code();
      moreCode();
    });

I guess overall I don't see a huge advantage to avoiding the parentheses here. In retrospect I would perhaps have done it that way had I thought of it originally, but I'm not sure it's worth changing now. I wouldn't want to break existing users, nor do I think it's worth having two different accepted syntaxes (if that's even possible; not sure).

Thoughts?

mrdomino:
Yeah, probably not worth it given that you still need the closing semicolon (I hadn't noticed that). Closing for now.

PR #274, opened 2015-12-26
Support extracting the current exception count on MSVC2015.

zarvox:
The MSVC runtime was heavily refactored, so the internal function we call to
get the current exception count changed for newer MSVC versions (as did the
offset).

I have not actually tested this code, as I don't have a native Windows dev
environment available at the moment. I just saw the `uncaught_exceptions`
proposal in #273 and remembered this comment, and realized I could fix it.

### DIFF
--- a/c++/src/kj/exception.c++
+++ b/c++/src/kj/exception.c++
@@ -551,12 +551,12 @@ uint uncaughtExceptionCount() {
 #elif _MSC_VER
 
 #if _MSC_VER >= 1900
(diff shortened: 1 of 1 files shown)


kentonv:
Hmm, the Boost bug discussion you reference suggests that the solution you copied no longer works, but there is another alternative: https://github.com/boostorg/log/commit/4751a165366dd9d1cf55252c290eaacf39a59552

zarvox:
Mea culpa.  Here's a patch based on that information.

PR #275, opened 2015-12-29
Simplify QueuedClient::call().

dwrensha:
It looks to me like `QueuedClient::call()` can be significantly simplified by using a promise / fulfiller pair, as in this patch. Am I missing something here?


kentonv:
The problem is that if `completionPromise` is dropped, the whole call will be canceled, even if the pipeline is still held, because in a promise/fulfiller pair the promise does not own the task that eventually invokes the fulfiller. The fork-based construction allows both the promise and pipeline to jointly own the task, ensuring that it is only canceled if both objects are dropped.

It's fairly common, I think, to drop the promise but keep the pipeline. Seems like we're missing a test here?

PR #277, opened 2016-01-07
Make Arena movable

mortehu:
If a `kj::Arena` is used to hold data needed for the fulfillment of a promise, it may be desirable to attach the arena to the promise using `Promise<T>::attach(...)`.  For this, a move constructor is required.

### DIFF
--- a/c++/src/kj/arena-test.c++
+++ b/c++/src/kj/arena-test.c++
@@ -286,6 +286,20 @@ TEST(Arena, Constructor) {
   EXPECT_EQ("foo", arena.allocate<StringPtr>("foo", 3));
 }
 
+TEST(Arena, MoveConstructor) {
+  TestObject::count = 0;
+
+  Arena arena1;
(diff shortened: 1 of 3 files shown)


kentonv:
Hmm. In general I'm wary of making things movable because it means than no future implementation of this class is allowed to store pointers to itself. So, for instance, if we wanted to add a pointer to `ChunkHeader` pointing back to the `Arena` for some reason, we couldn't, because the `Arena` might move and we wouldn't want to have to update all the `ChunkHeader`s on move. So what we'd probably end up doing then is allocating another intermediate object which all the `ChunkHeader`s can share. But as an alternative, anyone who needs a movable `Arena` could always allocate it on the heap and use `Own<Arena>` instead, which is then totally movable.

Generally my rule of thumb is that things which are fundamentally value types should be movable while resources types should not, and I'd call `Arena` a resource type.

But I'm open to other opinions. Thoughts?

mortehu:
I agree with your assessment.

PR #291, opened 2016-03-14
doc/cxxrpc: use getParams() in server

katreniak:
Server::Context does not have method getRequest


kentonv:
Doh. Thanks!

kentonv:
Site updated.

PR #293, opened 2016-03-17
Add Canonicalization

maurer:
The user facing API is in MessageReader and MessageBuilder

{MessageBuilder,MessageReader}::isCanonical verifies the canonicity of a
message. This is both useful for debugging and for knowing if a received
message can be used for hashes, bytewise equality, etc.

MessageBuilder::canonicalRoot(Reader) can be used to write a canonical
message on a best effort basis, and checks itself using isCanonical.
It should succeed as long as the MessageBuilder in question:
* Has a first segment which is long enough to contain the message
* Has not been used before

Tests have been added in canonicalize-test.c++ which verify that for
crafted examples of canonicalization errors, isCanonical will reject,
and for a canonicalized version of the standard test message, it will
accept.


c++/src/capnp/any.h
```diff
@@ -221,9 +221,14 @@ struct AnyPointer {
     inline void setAs(std::initializer_list<ReaderFor<ListElementType<T>>> list);
     // Valid for T = List<?>.
 
+    template <typename T>
+    inline void setAsCanonical(ReaderFor<T> value);
```
kentonv: Should be `setCanonicalAs`, since the `As` is meant to refer to the template parameter type.

c++/src/capnp/layout.c++
```diff
@@ -1545,23 +1552,43 @@ struct WireHelpers {
...
-    WordCount totalSize = dataSize + value.pointerCount * WORDS_PER_POINTER;
+    WirePointerCount ptrCount = value.pointerCount;
 
-    word* ptr = allocate(ref, segment, capTable, totalSize, WirePointer::STRUCT, orphanArena);
-    ref->structRef.set(dataSize, value.pointerCount);
+
+    if (canonical) {
+      // Truncate the data section
+      while ((dataSize != 0) &&
+             (value.getDataField<uint64_t>(dataSize - 1) == 0)) {
```
kentonv: This is UB unfortunately: only {,signed,unsigned} char is allowed to alias memory which may be accessed as some other type. You'll need to use memcmp.

kentonv: Also: The input struct's data section may be less than a whole word, because we allow List(Primitive) to be upgraded to List(Struct) where Struct's first field is of type Primitive (except for Bool because it's too hard). Therefore this read could be unaligned and/or out-of-bounds.

c++/src/capnp/layout.c++
```diff
@@ -1545,23 +1552,43 @@ struct WireHelpers {
...
+        *reinterpret_cast<char*>(ptr) = value.getDataField<bool>(0 * ELEMENTS);
+      }
     } else {
-      memcpy(ptr, value.data, value.dataSize / BITS_PER_BYTE / BYTES);
+      memcpy(ptr, value.data, dataSize * BYTES_PER_WORD);
     }
 
     WirePointer* pointerSection = reinterpret_cast<WirePointer*>(ptr + dataSize);
-    for (uint i = 0; i < value.pointerCount / POINTERS; i++) {
+    for (uint i = 0; i < ptrCount; i++) {
```
kentonv: Please compile with `-DCAPNP_DEBUG_TYPES` in order to catch unit errors. The `/ POINTERS` that you deleted was part of the unit analysis.

c++/src/capnp/layout.c++
```diff
@@ -1545,23 +1552,43 @@ struct WireHelpers {
...
 
     if (value.dataSize == 1 * BITS) {
-      *reinterpret_cast<char*>(ptr) = value.getDataField<bool>(0 * ELEMENTS);
+      // Data size could be made 0 by truncation
+      if (dataSize != 0) {
+        *reinterpret_cast<char*>(ptr) = value.getDataField<bool>(0 * ELEMENTS);
+      }
     } else {
-      memcpy(ptr, value.data, value.dataSize / BITS_PER_BYTE / BYTES);
+      memcpy(ptr, value.data, dataSize * BYTES_PER_WORD);
```
kentonv: This memcpy is now possibly a buffer overrun; see above.

c++/src/capnp/layout.c++
```diff
@@ -344,7 +345,8 @@ struct WireHelpers {
     //   segment belonging to the arena.  `ref` will be initialized as a non-far pointer, but its
     //   target offset will be set to zero.
 
-    if (orphanArena == nullptr) {
+    // Canonical messages can't be crafted from orphans, don't try
+    if ((orphanArena == nullptr) || canonical) {
```
kentonv: If `orphanArena` is non-null then taking this branch will crash (because `segment` is null).

It doesn't look to me like adding `canonical` as a parameter of this method is accomplishing anything. You throw an exception on one branch, but calling `isCanonical()` later would have detected the problem anyway.

I think it would be best to revert the changes to this method.

c++/src/capnp/layout.c++
```diff
@@ -1609,31 +1636,61 @@ struct WireHelpers {
       return { segment, ptr };
     } else {
       // List of structs.
+      WordCount declDataSize = roundBitsUpToWords(value.structDataSize);
```
kentonv: Before this line can you add:

    KJ_DASSERT(value.structDataSize % BITS_PER_WORD == 0);

? I think it holds here (because `elementSize` wouldn't be INLINE_COMPOSITE for a list of sub-word structs) but I'm not 100% sure. If it doesn't hold there will be a unit test failure. (Make sure you're compiling in debug mode, i.e. without `-DNDEBUG`.)

Assuming it does hold then you can replace the `roundBitsUpToWords` with a simple division.

If it doesn't hold then you will have problems below, similar to the issues in `setStructPointer()`.

c++/src/capnp/layout.c++
```diff
@@ -1609,31 +1636,61 @@ struct WireHelpers {
...
+
+      WordCount dataSize = 0 * WORDS;
+      WirePointerCount ptrCount = 0 * POINTERS;
+
+      if (canonical) {
+        for (auto ec = ElementCount(0); ec < value.elementCount; ec++) {
+          auto se = value.getStructElement(ec);
+          WordCount localDataSize = declDataSize;
+          while ((localDataSize != 0 * WORDS) &&
+                 (se.getDataField<uint64_t>(localDataSize - 1) == 0)) {
```
kentonv: UB again -- need memcmp().

c++/src/capnp/layout.c++
```diff
@@ -1777,6 +1836,11 @@ struct WireHelpers {
         }
 
 #if !CAPNP_LITE
+        if (canonical) {
+          KJ_FAIL_REQUIRE("Cannot create a canonical message with a capability") {
+            goto useDefault;
```
kentonv: I think this should probably be `break;` -- when exceptions are disabled we go on executing the code and simply produce a non-canonical message.

c++/src/capnp/layout.c++
```diff
@@ -2258,7 +2322,7 @@ Text::Builder PointerBuilder::initBlob<Text>(ByteCount size) {
   return WireHelpers::initTextPointer(pointer, segment, capTable, size).value;
 }
 template <>
-void PointerBuilder::setBlob<Text>(Text::Reader value) {
+void PointerBuilder::setBlob<Text>(Text::Reader value, bool canonical) {
```
kentonv: Why pass `canonical` to `setBlob()` at all if it isn't relevant?

c++/src/capnp/layout.c++
```diff
@@ -2468,6 +2534,36 @@ PointerReader PointerReader::imbue(CapTableReader* capTable) const {
...
+    return false;
+  }
+
+  switch (this->getPointerType()) {
+    case PointerType::NULL_:
+      // The pointer is null, we are canonical and do not read
+      return true;
+    case PointerType::STRUCT:
+      bool dataTrunc, ptrTrunc;
+      return (this->getStruct(nullptr).isCanonical(readHead,
```
kentonv: Why the one-parameter-per-line thing here? Note that KJ style says to go out to 100 chars before wrapping.

c++/src/capnp/message.c++
```diff
@@ -130,6 +159,25 @@ Orphanage MessageBuilder::getOrphanage() {
...
+  }
+
+  if (arena()->tryGetSegment(_::SegmentId(1))) {
+    // The message has more than one segment
+    return false;
+  }
+
+  const word* readHead = segment->getStartPtr() + 1;
+  return _::PointerReader::getRoot(segment, nullptr, segment->getStartPtr(),
+                                   INT_MAX)
```
kentonv: Use `kj::maxValue` instead of `INT_MAX`, then remove the `climits` include.

kentonv:
Procedural note: Please commit updates as a separate commit rather than squashing because otherwise I can't see what changed (github code reviews... not so good).

c++/src/capnp/canonicalize-test.c++
```diff
@@ -0,0 +1,147 @@
+// Copyright (c) 2013-2014 Sandstorm Development Group, Inc. and contributors
```
kentonv: 2016 :)

c++/src/capnp/message.h
```diff
@@ -195,6 +199,12 @@ class MessageBuilder {
   void setRoot(Reader&& value);
   // Set the root struct to a deep copy of the given struct.
 
+  template <typename Reader>
+  void canonicalRoot(Reader&& value);
```
kentonv: Perhaps the right interface to canonicalization is more like:

    template <typename Reader>
    kj::Array<word> canonicalize(Reader reader);

This would:

1. Call `reader.totalSize()` to get an upper bound on the canonical size.
2. Allocate a `FlatMessageBuilder`.
3. Call `getRoot<AnyPointer>().setCanonical()` to copy and canonicalize (no need for a new method of `MessageBuilder`).
4. Copy out the allocated part of the first segment and return it.

c++/src/capnp/canonicalize-test.c++
```diff
@@ -0,0 +1,147 @@
...
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00
+  }};
+  kj::ArrayPtr<const word> segments[1] = {
+    kj::arrayPtr(nonTruncatedList.words, 6)
+  };
+  SegmentArrayMessageReader nonTruncated(kj::arrayPtr(segments, 1));
+
+  KJ_ASSERT(!nonTruncated.isCanonical());
+}
+}  // namespace
```
kentonv: Please add a test where a primitive list has been upgraded to a struct list, and then you're trying to canonicalize one struct out of that list -- and therefore it has less-than-word size.

c++/src/capnp/canonicalize-test.c++
```diff
@@ -0,0 +1,147 @@
...
+  }};
+  kj::ArrayPtr<const word> segments[1] = {
+    kj::arrayPtr(nonTruncatedSegment.words, 3)
+  };
+  SegmentArrayMessageReader nonTruncated(kj::arrayPtr(segments, 1));
+
+  KJ_ASSERT(!nonTruncated.isCanonical());
+}
+
+KJ_TEST("isCanonical requires truncation of 0-valued struct fields\
```
kentonv: Please don't escape a newline in a string like this. Adjacent string literals are concatenated by C, so you can have each line be a separate literal. Alternatively, it looks like this whole string should fit on one line (since the limit is 100 char).

kentonv:
OK I think that's all the comments I have for now.

Thanks for working on this!

c++/src/capnp/layout.c++
maurer: While I agree that no message encoded by the library as it stands should generate an inline composite for a non-word-multiple struct, afaict the `structDataSize` just comes off the tag word, and so can be whatever it wants. If someone makes an inline-composite of a byte struct, it might not be spec valid, but I'm not actually sure it would be prevented from making it to this point in the code.

In summary, I'm worried that a crafted message could make this assert trip.

Given this, can I use a `KJ_ASSERT()` instead of a `KJ_DASSERT()`? Otherwise I can just write the other-size tolerant code like I did for `setStructPointer`.

kentonv: In the INLINE_COMPOSITE case, the tag word has the layout of a struct pointer. Within a struct pointer, the data section size is a count of words, so there's no way it would not be evenly divisible by words here, right?

maurer: Err, I realized I was a dummy after I posted that and deleted my comment before I saw your response to try to avoid bothering you. You are correct, and a DASSERT is fine.

maurer:
I don't have the additional test you asked for written yet (I'll get to that soon) but I think I've addressed all the other suggestions.

maurer:
@kentonv I don't have the additional test you asked for written yet (I'll get to that soon) but I think I've addressed all the other suggestions.

c++/src/capnp/message.h
```diff
@@ -414,6 +421,21 @@ class FlatMessageBuilder: public MessageBuilder {
   bool allocated;
 };
 
+template <typename MR>
+kj::Array<word> canonicalize(MR&& reader) {
+  AnyPointer::Reader root = reader.template getRoot<AnyPointer>();
+  WordCount size = root.targetSize().wordCount * WORDS + POINTER_SIZE_IN_WORDS;
+  kj::Array<word> backing = kj::heapArray<word>(size / WORDS);
+  bzero(backing.begin(), backing.asBytes().size());
```
kentonv: bzero() is deprecated by POSIX. Please use memset().

c++/src/capnp/message.h
```diff
@@ -414,6 +421,21 @@ class FlatMessageBuilder: public MessageBuilder {
   bool allocated;
 };
 
+template <typename MR>
```
kentonv: Does this really need to be a template? Couldn't it take `const MessageReader&` as its parameter?

kentonv: I see, my original suggestion was that this function accept *any reader type* as input, not a MessageReader.

kentonv: Yeah, I think we should really make this accept an arbitrary struct reader type, otherwise we might be forcing people to make yet another redundant copy into a MallocMessageBuilder.

I think it's OK to require specifically a struct reader, which means that you can use `.totalSize()` instead of `.targetSize()`. Everything else should work, I think.

maurer:
@kentonv The new test (which can exist now that we're canonicalizing off struct readers) is added, and the user can canonicalize from any `StructReader`, `AnyStruct::Reader`, or capnpc-c++ generated `Foo::Reader` (where Foo is a Struct).

Only big caveat here is that this very slightly bloats generated code by adding an additional passthrough for each `Foo::Reader`.

kentonv:
Yeah, I don't really think we should add a method on every generated class. I was hoping to just change the input of `canonicalize()` to accept any struct reader.

Something like this should work:

    template <typename T>
    kj::Array<word> canonicalize(T&& reader) {
      return PointerHelpers<FromReader<T>>::getInternalReader(reader).canonicalize();
    }

(Sorry for the review latency... swamped as usual. :( )

maurer:
@kentonv Should be good now (removed the stubs from generated code). Thanks for being patient with my C++, I know it's a little out of date stylistically.

kentonv:
Great! Can you please add yourself to `CONTRIBUTORS` in the root dir? :)

(Sorry, I always forget this one until the last moment. :( )

maurer:
@kentonv Added.

kentonv:
Thanks!

PR #294, opened 2016-03-21
capnp/ez-rpc: add method setMainInterface

katreniak:
Add an option to change main interface at run time.

Main motivation for this change is that my server's constructor
depends on event loop being initialized.
So it could not created before EzRpcServer constructor.

### DIFF
--- a/c++/src/capnp/ez-rpc.h
+++ b/c++/src/capnp/ez-rpc.h
@@ -207,6 +207,9 @@ public:
 
   ~EzRpcServer() noexcept(false);
 
+  void setMainInterface(Capability::Client cap);
+  // Set / replace main interface.
+
   void exportCap(kj::StringPtr name, Capability::Client cap);
   // Export a capability publicly under the given name, so that clients can import it.
   //
--- a/c++/src/capnp/ez-rpc-test.c++
+++ b/c++/src/capnp/ez-rpc-test.c++
@@ -44,6 +44,24 @@ TEST(EzRpc, Basic) {
   EXPECT_EQ(1, callCount);
 }
 
+TEST(EzRpc, SetMainInterface) {
+  int callCount = 0;
+  EzRpcServer server(nullptr, "localhost");
+  server.setMainInterface(kj::heap<TestInterfaceImpl>(callCount));
+
+  EzRpcClient client("localhost", server.getPort().wait(server.getWaitScope()));
+
+  auto cap = client.getMain<test::TestInterface>();
+  auto request = cap.fooRequest();
+  request.setI(123);
+  request.setJ(true);
+
+  EXPECT_EQ(0, callCount);
+  auto response = request.send().wait(server.getWaitScope());
+  EXPECT_EQ("foo", response.getX());
+  EXPECT_EQ(1, callCount);
+}
+
 TEST(EzRpc, DeprecatedNames) {
   EzRpcServer server("localhost");
   int callCount = 0;
--- a/c++/src/capnp/ez-rpc.c++
+++ b/c++/src/capnp/ez-rpc.c++
@@ -334,6 +334,10 @@ EzRpcServer::EzRpcServer(int socketFd, uint port, ReaderOptions readerOpts)
 
 EzRpcServer::~EzRpcServer() noexcept(false) {}
 
+void EzRpcServer::setMainInterface(Capability::Client cap) {
+  impl->mainInterface = kj::mv(cap);
+}
+
 void EzRpcServer::exportCap(kj::StringPtr name, Capability::Client cap) {
   Impl::ExportedCap entry(kj::heapString(name), cap);
   impl->exportMap[entry.name] = kj::mv(entry);


kentonv:
This feels a bit ugly to me. It theoretically creates a situation where your server is briefly listening but not yet actually ready to handle messages. In practice it probably isn't an issue if you set the main interface before the event loop has a chance to run at all, but theoretically that's not strictly guaranteed by the interface. E.g. the RPC server could be designed to call accept() and start reading messages if they happen to be available from the kernel already (unlikely, but possible) before the constructor returns.

Here are two other options:

1) Create a promise/fulfiller pair, and pass the promise as the main interface, then fulfill it later with the actual main interface. Like:

    auto paf = kj::newPromiseAndFulfiller<MyMainInterface::Client>();
    EzRpcServer server(kj::mv(paf.promise), ...);
    paf.fulfiller->fulfill(kj::heap<MyMainInterfaceImpl>();

(A `kj::Promise<T::Client>` can implicitly convert to a `T::Client` which queues messages until the promise resolves, then forwards them.)

2) Don't use EzRpcServer. Use `capnp/rpc-twoparty.h` directly. You'll probably find you need to do this eventually anyway. The EzRpc interfaces are intentionally designed to remove features for the sake of simplicity.

Thoughts?

katreniak:
Hi Kenton.

The `Promise<Client>` solves my problem. Thanks a lot!

My thoughts? You ask for it :)

1. I consider it bad design to start serving clients in constructor. It leads to trap that the whole configuration must be passed as arguments to constructor. It is fine to have handy constructor that starts acting, but it should not be the only way to construct the class. The class should either start acting after entering the event loop or after an explicit `start()`.

2. I believe that changing `mainInterface` on the fly is valid use case.

Feel free to close this PR if you don't buy my arguments.

kentonv:
Thanks for the thoughts. But I disagree. :)

1. I believe that classes should be fully-initialized in their constructor. Inside Google I saw a lot of uses of the post-initialization pattern and I also saw people write long rants declaring it to be an anti-pattern. I tend to agree: The spirit of C++ is that the constructor initializes the methods, and once complete all methods are safe to call. The type system is meant to guarantee this -- so that if you have a reference to an EzRpcServer, you know for sure that it is safe to call all its methods. If the class has complex configuration, it is best to either represent that configuration as a separate type which is passed into the constructor, or to create a "factory class" which can be configured and then has a method which actually constructs and returns the final object. I think the specific problem in this case comes from the fact that EzRpcServer is doing too many things at once, and the right solution is to break it down to the underlying interfaces in `rpc-twoparty.h` instead.
2. Allowing people to change `mainInterface` on-the-fly feels very awkward to me, and I am honestly pretty skeptical that it is a good fit for real use cases. However, the underlying interfaces now support something called `BootstrapFactory`, which allows you to specify a factory object to call whenever a client requests the bootstrap interface. `BootstrapFactory` is strictly more powerful than `setMainInterface()`, so if anything I'd rather expose `BootstrapFactory` than provide `setMainInterface()`.

FWIW, I would accept a change that makes `EzRpcServer` support specifying a `BootstrapFactory` instead of a `mainInterface`. I think it would also solve your problem without the `Promise` hack.

katreniak:
Hi Kenton,

While I really like your design decisions, I still have hard time to buy factory over simplicity of post initialization. I have seen too many factories in Java world. But I see now that my patch is not consistent with your vision.

Thanks you for feedback.

katreniak:
auto paf = kj::newPromiseAndFulfiller<MyMainInterface::Client>();
    EzRpcServer server(kj::mv(paf.promise), ...);
    paf.fulfiller->fulfill(kj::heap<MyMainInterfaceImpl>();

This unfortunately does not work for me. It asserts somewhere in EzRpcServer constructor :(
I switched to deprecated `EzRpcServer::exportCap` for time being.

kentonv:
> It asserts somewhere in EzRpcServer constructor :(

That sounds like a bug. What's the stack trace?

katreniak:
Simplest code to reproduce:

    auto rpcInterfacePaf = kj::newPromiseAndFulfiller<Hello::Client>();
    capnp::Capability::Client client(kj::mv(rpcInterfacePaf.promise));

Error message:

    terminate called after throwing an instance of 'kj::ExceptionImpl'
      what():  kj/async.c++:53: failed: expected loop != nullptr; No event loop is running on this thread.
    stack: 0x7f3d5444ff77 0x7f3d5444ffd4 0x7f3d5495ca5e 0x7f3d54953f02 0x405164 0x4049a2 0x7f3d535d0580 0x404a69

Translated stack trace:

    kj::_::Event::Event() + 23 in section .text of /home/brano/dss-sysroot/lib/libkj-async-0.6-dev.so
    kj::_::ForkHubBase::ForkHubBase(kj::Own<kj::_::PromiseNode>&&, kj::_::ExceptionOrValue&) + 52 in section .text of /home/brano/dss-sysroot/lib/libkj-async-0.6-dev.so
    capnp::QueuedClient::QueuedClient(kj::Promise<kj::Own<capnp::ClientHook> >&&) + 94 in section .text of /home/brano/dss-sysroot/lib/libcapnp-rpc-0.6-dev.so
    capnp::newLocalPromiseClient(kj::Promise<kj::Own<capnp::ClientHook> >&&) + 34 in section .text of /home/brano/dss-sysroot/lib/libcapnp-rpc-0.6-dev.so
    capnp::Capability::Client::Client<Hello::Client, void>(kj::Promise<Hello::Client>&&) + 60 in section .text of /home/brano/dss/capnproto-feasibility/src/build/capnp/capnp-hello-server

PS. What is the best way to translate stack addressed to symbols? It would be helpful to have a hint in documentation.

kentonv:
Hmm, dang, I think that error is actually expected. I guess you can't create a PromiseFulfillerPair before the event loop is constructed. So we're back to the original problem.

> What is the best way to translate stack addressed to symbols?

If you compile libkj in debug mode on Linux or OSX it should automatically symbolize the stack trace. On Linux it uses `addr2line`. On OSX it uses `atos`.

katreniak:
> BootstrapFactory is strictly more powerful than setMainInterface(), so if anything I'd rather expose BootstrapFactory than provide setMainInterface().

BootstrapFactory would pollute my demo code with concepts I don't want to show to readers. It is kind of too powerful :) The new `mainInterface` in `EzRpcServer` can be seen as a regression compared to deprecated `exportCap` methods.

I am going to carry this patch on my own. It works, it solves my problem, it is simple and it is non-intrusive. I cannot expect to get all my changes accepted upstream :)

c++/src/capnp/ez-rpc.h
```diff
@@ -207,6 +207,9 @@ class EzRpcServer {
 
   ~EzRpcServer() noexcept(false);
 
+  void setMainInterface(Capability::Client cap);
```
katreniak: This method should be called `setMain` to be consistent with EzRpcClient.

katreniak:
What about adding new EzRpcServer and EzRpcClient constructors that take AsyncIoContext& as their first argument? 

Is it really good idea to wrap the call to `kj::setupAsyncIo()` in these classes? The context reference counting inside EzRpcServer and EzRpcClient is in fact singleton and it allows anybody to get to waitScope by calling `EzRpcClient("").getWaitScope()`.

kentonv:
I suppose passing AsyncIoContext into the constructor would be OK, but I really do think any app that finds EzRpc too limiting really ought to move on to using TwoPartyClient / TwoPartyServer directly.

PR #295, opened 2016-03-21
kj/vector: add methods operator = nullptr and truncate

katreniak:
=nullptr releases memory

https://github.com/sandstorm-io/capnproto/issues/292


kentonv:
Thanks!

PR #296, opened 2016-03-23
json: integrate to build

katreniak:
Json code was sitting outside of build system. This commit integrates json classes into build and tests.
Json code is available in library `capnp-json`, header files are installed to `INSTALL_DIR/capnp/compat/json*`


c++/Makefile.am
```diff
@@ -181,12 +184,14 @@ includecapnp_HEADERS =                                         \
   src/capnp/rpc.capnp.h                                        \
   src/capnp/rpc-twoparty.capnp.h                               \
   src/capnp/persistent.capnp.h                                 \
-  src/capnp/ez-rpc.h
+  src/capnp/ez-rpc.h                                           \
+  src/capnp/compat/json.h                                      \
```
kentonv: Unfortunately automake is terrible and listing the headers here will cause them to be installed in `/usr/include/capnp` (without `compat`). You'll need to define a new variable `includecapnpcompat_HEADERS`, similar to `includekjcompat_HEADERS` above.

c++/regenerate-bootstraps.sh
```diff
@@ -7,4 +7,5 @@ export PATH=$PWD/bin:$PWD:$PATH
 capnp compile -Isrc --no-standard-import --src-prefix=src -oc++:src \
     src/capnp/c++.capnp src/capnp/schema.capnp \
     src/capnp/compiler/lexer.capnp src/capnp/compiler/grammar.capnp \
-    src/capnp/rpc.capnp src/capnp/rpc-twoparty.capnp src/capnp/persistent.capnp
+    src/capnp/rpc.capnp src/capnp/rpc-twoparty.capnp src/capnp/persistent.capnp \
```
kentonv: Hmm, I wonder if json.capnp really needs to be bootstrapped. Since it's in a separate library, it should be possible to set up the build system to run the newly-built capnp tool to generate this within the build, like we do for all the test files. We couldn't do this with most of the other files because they are actually dependencies of the compiler. Then again, arguably rpc.capnp is not a dependency of the compiler, yet we bootstrap it. I can't quite remember why...

If we do need to go with the bootstrap approach, can you please also update:
* `c++/src/capnp/bootstrap-test.ekam-rule`
* `c++/src/capnp/capnpc.ekam-rule`

kentonv:
Thanks for doing this! Just had a couple comments (above).

c++/regenerate-bootstraps.sh
katreniak: Truly speaking I don't understand what I am doing here. I just tried to be consistent and I hoped for review feedback. I will try to address these comments the best I can next week.

c++/Makefile.am
katreniak: I am pretty sure this worked for me. It installed it in `${INSTAL}/include/compat/json.h`. I will check again. But I will rework this as suggested.

c++/regenerate-bootstraps.sh
kentonv: Feel free to stick with this for now and we can file a bug to try to improve it later. Getting it to work as part of the build might turn out surprisingly complicated, at least with automake.

c++/Makefile.am
katreniak: You are right. I was using left over from cmake build. Amended.

kentonv:
Thanks!

For future PRs, please don't squash commits when responding to a review, as it makes it hard for me to see what changed. (I know squashing is considered "good hygiene" but github's code reviews are... limited. We should get Phabricator running on Sandstorm, probably.)

PR #262, opened 2015-11-12
JSON: encode NaNs and infinite floats as `null`

kamalmarhubi:
Such values are not allowed by the JSON spec. We match the behavior of
JSON.stringify in Firefox and Chrome.

fixes https://github.com/sandstorm-io/capnproto/issues/261

Issue #261, opened 2015-11-11
JSON encoding allows inf, -inf, nan which are disallowed by spec


kamalmarhubi:
Compare to

```
» a = [0/0, 1/0, -1/0]
« [NaN, Infinity, -Infinity]
» JSON.stringify(a)
« "[null,null,null]"
```

as seen in console for both Chrome and Firefox.

kamalmarhubi:
The spec also says

> Numeric values that cannot be represented as sequences of digits (such as Infinity and NaN) are not permitted.

kamalmarhubi:
I don't think there's any way to roundtrip those values. I'll send a PR to fix the encoder; unsure what to do in the decoder. Probably replace with nan?

kamalmarhubi:
```
» Number(JSON.parse(JSON.stringify(1/0)))
« 0
```

kamalmarhubi:
Hrm, Python's `json` module allows all them:

```
>>> a = [float('nan'), float('inf'), float('-inf')]
>>> a
[nan, inf, -inf]
>>> json.dumps(a)
'[NaN, Infinity, -Infinity]'
>>> json.loads(json.dumps(a))
[nan, inf, -inf]
```

kamalmarhubi:
Ruby has the same behaviour.

kamalmarhubi:
And Go disallows them. I think it's unsafe the encode these values is the aim is interoperability with JSON-expecting systems. It may be acceptable to decode them, however #258 does not.

kentonv:
How about encoding them as strings? We can make sure our own encoder and parser can round-trip these correctly (albeit treating all NaNs as equivalent, oh well).

Otherwise it seems like there's nothing reasonable we can do except throw an exception, but I think that would probably we worse for most apps.

kamalmarhubi:
The existing encoding was already treating all NaNs as equivalent, so no loss there.

I don't know the main use case for the JSON codec, so it's hard for me to pick between alternatives. How about disallowing them by default, but adding a ready-made type handler for the string encoding?

kentonv:
My worry is that this would be the *only* case where exceptions could be thrown at serialization time. Many users are likely to assume that serialization doesn't throw, so this could end up being a security issue: trick someone into JSON-ifying a message you send them that happens to contain infinity/nan, and now you've caused an exception where perhaps they weren't prepared for it. (Then again, people *should* be prepared for an exception to happen *anywhere*, but...)

[...]

PR #297, opened 2016-03-23
json: implement decode(... input, DynamicStruct::Builder output)

katreniak:
* type and field handlers are not implemented yet.
* top level json object must be struct.

This code covers the basic use case: to convert json string to capnp struct.


c++/src/capnp/compat/json-test.c++
```diff
@@ -183,6 +183,83 @@ KJ_TEST("encode union") {
...
+      kj::heapArray<byte>({7,0,122}).asPtr() == root.getDataList()[0]);
+  CASE(R"({"structList":null})", root.hasStructList() == false);
+  CASE(R"({"structList":[null]})", root.hasStructList() == true);
+  CASE(R"({"structList":[null]})", root.getStructList()[0].getBoolField() == false);
+  CASE(R"({"structList":[{}]})", root.getStructList()[0].getBoolField() == false);
+  CASE(R"({"structList":[{"boolField":false}]})", root.getStructList()[0].getBoolField() == false);
+  CASE(R"({"structList":[{"boolField":true}]})", root.getStructList()[0].getBoolField() == true);
+  CASE(R"({"enumList":["bar"]})", root.getEnumList()[0] == TestEnum::BAR);
+#undef CASE
+}
```
kentonv: Can you add a test that uses initTestMessage() and checkTestMessage() from capnp/test-util.h and does a round-trip to JSON and back?

c++/src/capnp/compat/json.c++
```diff
@@ -379,13 +379,119 @@ void JsonCodec::encodeField(StructSchema::Field field, DynamicValue::Reader inpu
   encode(input, field.getType(), output);
 }
 
+namespace {
+int64_t parseInt64(kj::StringPtr s)
+{
```
kentonv: Please put open braces at end of previous line to match KJ style.

kentonv:
Hooray, thanks for implementing this!

@kamalmarhubi, any comments?

kentonv:
Oh also, please add yourself to CONTRIBUTORS.txt. :)

kamalmarhubi:
> @kamalmarhubi, any comments?

I'm mostly super glad someone did this. I had a branch that did some of it, but it fell by the wayside, and it's been a background sadface guilt item ever since!

I should be able to take a look at the code tomorrow night or Monday.

c++/src/capnp/compat/json-test.c++
```diff
@@ -183,6 +183,83 @@ KJ_TEST("encode union") {
   KJ_EXPECT(json.encode(root) == "{\"before\":\"a\",\"middle\":44,\"bar\":321,\"after\":\"c\"}");
 }
 
+KJ_TEST("decode all types") {
+  JsonCodec json;
+#define CASE(s, f) \
```
kamalmarhubi: nice!

c++/src/capnp/compat/json.c++
```diff
@@ -379,13 +379,119 @@ void JsonCodec::encodeField(StructSchema::Field field, DynamicValue::Reader inpu
...
+  KJ_REQUIRE(endPtr == s.end() && errno != ERANGE, "String is not correct int64 number");
+  return value;
+}
+
+uint64_t parseUInt64(kj::StringPtr s)
+{
+  char *endPtr;
+  errno = 0;
+  uint64_t value = std::strtoull(s.begin(), &endPtr, 10);
+  KJ_ASSERT(endPtr == s.end() && errno != ERANGE, "String is not correct uint64 number");
```
kamalmarhubi: this should be `KJ_REQUIRE` as it's an input problem rather than a bug

c++/src/capnp/compat/json.c++
```diff
@@ -379,13 +379,119 @@ void JsonCodec::encodeField(StructSchema::Field field, DynamicValue::Reader inpu
   encode(input, field.getType(), output);
 }
 
+namespace {
+int64_t parseInt64(kj::StringPtr s)
+{
+  char *endPtr;
+  errno = 0;
+  int64_t value = std::strtoll(s.begin(), &endPtr, 10);
+  KJ_REQUIRE(endPtr == s.end() && errno != ERANGE, "String is not correct int64 number");
```
kamalmarhubi: does silently returning `LLONG_MAX` for overflow make sense, or should this throw in that case? @kentonv what is most KJesque?

c++/src/capnp/compat/json.c++
kamalmarhubi: also ditto re the `ERANGE` handling here

c++/src/capnp/compat/json.c++
```diff
@@ -379,13 +379,119 @@ void JsonCodec::encodeField(StructSchema::Field field, DynamicValue::Reader inpu
...
+uint64_t parseUInt64(kj::StringPtr s)
+{
+  char *endPtr;
+  errno = 0;
+  uint64_t value = std::strtoull(s.begin(), &endPtr, 10);
+  KJ_ASSERT(endPtr == s.end() && errno != ERANGE, "String is not correct uint64 number");
+  return value;
+}
+
+template <typename Set, typename DecodeArray, typename DecodeObject>
```
kamalmarhubi: could you name the template parameters `SetFn`, or comment that they are functions?

c++/src/capnp/compat/json.c++
```diff
@@ -379,13 +379,119 @@ void JsonCodec::encodeField(StructSchema::Field field, DynamicValue::Reader inpu
...
+    case JsonValue::CALL:
+      //TODO(soon)
+      KJ_FAIL_ASSERT("JSON call decode not implemented yet. :(");
+      break;
+  }
+}
+} //namespace
+
+void JsonCodec::decodeArray(List<JsonValue>::Reader input, DynamicList::Builder output) const {
+  KJ_ASSERT(input.size() == output.size(), "Builder must be initialized to input size");
```
kamalmarhubi: message makes it sound like something the user could fix. This is an internal implementation detail, so the message would best be phrased "Builder was not initialized to the correct size".

c++/src/capnp/compat/json.c++
```diff
@@ -379,13 +379,119 @@ void JsonCodec::encodeField(StructSchema::Field field, DynamicValue::Reader inpu
...
+  auto type = output.getSchema().getElementType();
+  for (auto i = 0; i < input.size(); i++) {
+    decodeField(type, input[i],
+        [&](DynamicValue::Reader value) { output.set(i, value); },
```
kamalmarhubi: This capture of `i` can be by value. I don't know about KJ style, but I prefer explicit lambda captures; here that would be `[i, &output]`.

c++/src/capnp/compat/json.c++
```diff
@@ -379,13 +379,119 @@ void JsonCodec::encodeField(StructSchema::Field field, DynamicValue::Reader inpu
...
+  errno = 0;
+  uint64_t value = std::strtoull(s.begin(), &endPtr, 10);
+  KJ_ASSERT(endPtr == s.end() && errno != ERANGE, "String is not correct uint64 number");
+  return value;
+}
+
+template <typename Set, typename DecodeArray, typename DecodeObject>
+void decodeField(Type type, JsonValue::Reader value, Set set, DecodeArray decodeArray,
+    DecodeObject decodeObject) {
+  switch (value.which()) {
```
kamalmarhubi: I think the switch should be on the `type` rather than the `value` discriminant. This makes it easier to ensure that schema type is matching the JSON value type.

c++/src/capnp/compat/json.c++
```diff
@@ -379,13 +379,119 @@ void JsonCodec::encodeField(StructSchema::Field field, DynamicValue::Reader inpu
...
+    case JsonValue::NULL_:
+      break;
+    case JsonValue::BOOLEAN:
+      set(value.getBoolean());
```
kamalmarhubi: Is this relying on `DynamicStruct::Builder::set()` throwing on a type error? I don't know if it does or not but could you add a comment explaining how we know all these conversions are correct?

kamalmarhubi:
I took a pass at review. There are some nits and a couple of places where it wasn't obvious to me things would work because I'm not super familiar with `DynamicStruct`.

c++/src/capnp/compat/json.c++
katreniak: Citation from [KJ style guide](https://github.com/sandstorm-io/capnproto/blob/master/style-guide.md): 

Capture lists may use & ("capture all by reference") but only in cases where it is known that the lambda will not outlive the current stack frame. In fact, they generally should use & in this case, to make clear that there are no lifetime issues to think about.

c++/src/capnp/compat/json.c++
```diff
@@ -379,13 +379,119 @@ void JsonCodec::encodeField(StructSchema::Field field, DynamicValue::Reader inpu
   encode(input, field.getType(), output);
 }
 
+namespace {
+int64_t parseInt64(kj::StringPtr s)
```
katreniak: Kenton, would you accept to merge these parse functions somewhere into `kj`? There is no json dependency in this code. It just  `kj` style wrapper around std function.

c++/src/capnp/compat/json.c++
```diff
@@ -379,13 +379,119 @@ void JsonCodec::encodeField(StructSchema::Field field, DynamicValue::Reader inpu
...
+      set(value.getNumber());
+      break;
+    case JsonValue::STRING:
+      if (type.isInt8() || type.isInt16() || type.isInt32() || type.isInt64()) {
+        set(parseInt64(value.getString()));
+      } else if (type.isUInt8() || type.isUInt16() || type.isUInt32() || type.isUInt64()) {
+        set(parseUInt64(value.getString()));
+      } else {
+        set(value.getString());
+      }
```
katreniak: Kenton, would you accept this string -> number conversion into `DynamicValue::Reader::as<T>`?

c++/src/capnp/compat/json.c++
katreniak: Does it really matter? This parser follows JSON structure and converts it into capnp and it relies a lot on conversions in `DynamicValue::Reader::as<T>`. It feels weir for me to revert the code flow at this place.

c++/src/capnp/compat/json-test.c++
katreniak: That test fails on `inf`, `-inf` and `nan` in float fields. Json encoder stores them as `null`. Can we change encoder to store them as "NaN", "Infinity", ""-Infinity"? This is inline with protobuf and parsing is directly supported in `std::strtod`.

katreniak: NaN can still stay as `null`.

c++/src/capnp/compat/json.c++
katreniak: I refactored the switch as asked. The code is more explicit now. Thanks !

katreniak:
I (hopefully) addressed all review comments. 

Kenton, I changed how NaN, Inf and -Inf is serialized. Do you agree with the change?

kamalmarhubi:
re nan, inf, -inf, I made the change to encode as null in #262. I think a better solution will be to provide type or field handlers that can be used to get this behavior, and keep the base codec very close to the spec.

kamalmarhubi:
I tried cobbling together a GitHub compare URL to see what's new but it didn't work: https://github.com/sandstorm-io/capnproto/compare/9aa52315d17c2a0d926173e871b12787a30f3c0e...419d3f18208d162484b54f7c13d9d293400f444a

@katreniak could you make it that the changes addressing the review are a separate commit? It'll make it easier to see what changed. You could squash prior to merging.

I think a sequence like this should do it:

```
git reset 9aa52315
git cherry-pick 4bb5ec4
git commit -a -m 'Address comments on PR'
```

c++/src/capnp/compat/json.c++
kamalmarhubi: Ah thanks!

c++/src/capnp/compat/json.c++
kamalmarhubi: I wasn't 100% sure it'd be better, but I was going based on the structure I had in my incomplete branch. I figured past me had a reason... glad to know maybe it did :-)

katreniak:
I will force push again tomorrow: reviewed commit + changes.

katreniak:
> nan, inf, -inf, 

Do we really want to lose information in capn -> json -> capnp transformation by default? We don't have problem to represent double as string in json. Is that in line with json spec? Default should be useful for most people.

Why is "Infinity" so big problem? [Protobuf sticks to "Infinity"](https://developers.google.com/protocol-buffers/docs/proto3#json), [Avro sticks to "Infinity"](https://issues.apache.org/jira/browse/AVRO-1290).

kamalmarhubi:
Major browsers encode them as null; stick this in your URL bar: `javascript:alert(JSON.stringify({a: 1/0, b: -1/0, c: 0/0}));`.

In #261 I compared to Python, Ruby, Go and JavaScript. There's nothing super consistent. 

> We don't have problem to represent double as string in json. Is that in line with json spec?

This is a good point. 

> Default should be useful for most people.

Is roundtripping Cap'n Proto to JSON and back the most likely use case? I imagine it's more likely to be used for interop with JSON-consuming systems, in which case the "default" is arbitrary and depends on the implementation details of the other system.

katreniak:
> Is roundtripping Cap'n Proto to JSON and back the most likely use case?

It is our use case :)

Json is great for external API. You cannot push capnp (or protobuf, ...) to 3rd parties. The biggest advantage of json over capnp in 3rd party API is that it does not need schema to decode. We are likely to settle with some binary json encoding at the end.

I really dislike to design for incorrect serialization as the default one.

kamalmarhubi:
> I really dislike to design for incorrect serialization as the default one. 

I guess my big issue is that I don't see any of the other alternatives as being more or less incorrect, just *differently* incorrect! That's why I pushed for closer adherence to the JSON spec.

I think at this point we've got our arguments for our preferences fairly cleanly laid out. We'll need to hear from @kentonv or another contributor / maintainer to see what they think. :-)

katreniak:
Force pushed to restore original commit + commit addressing review comments. I added 3rd commit that changes inf serialization. It can be easily dropped.

c++/src/capnp/compat/json.c++
```diff
@@ -380,72 +380,135 @@ void JsonCodec::encodeField(StructSchema::Field field, DynamicValue::Reader inpu
...
+        setFn(parseInt64(value.getString()));
+      } else {
+        KJ_FAIL_REQUIRE("Expected numeric value");
+      }
+      break;
+    case schema::Type::UINT8:
+    case schema::Type::UINT16:
+    case schema::Type::UINT32:
+    case schema::Type::UINT64:
+      //Rellies on range check in DynamicValue::Reader::as<IntType>
```
kamalmarhubi: typo: Relies (one l)

c++/src/capnp/compat/json-test.c++
```diff
@@ -258,6 +308,38 @@ KJ_TEST("decode all types") {
...
+  initTestMessage(root);
+
+  JsonCodec json;
+  auto encoded = json.encode(root);
+
+  MallocMessageBuilder decodedMessage;
+  auto decodedRoot = decodedMessage.initRoot<TestAllTypes>();
+  json.decode(encoded, decodedRoot);
+
+  //json encode serializes nan, inf and -inf as null.
```
kamalmarhubi: nit: space after `//` in comments (consistent with rest of code)

c++/src/capnp/compat/json.c++
```diff
@@ -380,72 +380,135 @@ void JsonCodec::encodeField(StructSchema::Field field, DynamicValue::Reader inpu
...
+        KJ_FAIL_REQUIRE("Expected numeric value");
+      }
       break;
-    case JsonValue::STRING:
-      if (type.isInt8() || type.isInt16() || type.isInt32() || type.isInt64()) {
-        set(parseInt64(value.getString()));
-      } else if (type.isUInt8() || type.isUInt16() || type.isUInt32() || type.isUInt64()) {
-        set(parseUInt64(value.getString()));
+    case schema::Type::TEXT:
+      if (value.isString()) {
```
kamalmarhubi: you can just `KJ_REQUIRE(value.isString())` here, instead of have the if/else

c++/src/capnp/compat/json.c++
```diff
@@ -380,72 +380,135 @@ void JsonCodec::encodeField(StructSchema::Field field, DynamicValue::Reader inpu
...
+      }
       break;
-    case JsonValue::BOOLEAN:
-      set(value.getBoolean());
+    case schema::Type::INT8:
+    case schema::Type::INT16:
+    case schema::Type::INT32:
+    case schema::Type::INT64:
+      //Rellies on range check in DynamicValue::Reader::as<IntType>
+      if (value.isNumber()) {
```
kamalmarhubi: rather than a sequence of `if / else if / ... / else` here, you could `switch(value.which())`. It *might* produce more performant code, but I think it will certainly read better.

c++/src/capnp/compat/json.c++
```diff
@@ -380,72 +380,135 @@ void JsonCodec::encodeField(StructSchema::Field field, DynamicValue::Reader inpu
...
+        setFn(value.getString());
+      } else {
+        KJ_FAIL_REQUIRE("Expected enum as string value");
+      }
       break;
-    case JsonValue::CALL:
-      //TODO(soon)
-      KJ_FAIL_ASSERT("JSON call decode not implemented yet. :(");
+    case schema::Type::STRUCT: {
+      if (value.isNull()) {
```
kamalmarhubi: empty braces?

c++/src/capnp/compat/json.c++
```diff
@@ -380,72 +380,135 @@ void JsonCodec::encodeField(StructSchema::Field field, DynamicValue::Reader inpu
...
+    case schema::Type::LIST:
+      if (value.isNull()) {
+      } else if (value.isArray()) {
+        decodeArrayFn(value.getArray());
+      } else {
+        KJ_FAIL_REQUIRE("Expected array value");
+      }
+      break;
+    case schema::Type::ENUM:
+      if (value.isString()) {
```
kamalmarhubi: `KJ_REQUIRE(value.isString())`

c++/src/capnp/compat/json.c++
```diff
@@ -380,72 +380,135 @@ void JsonCodec::encodeField(StructSchema::Field field, DynamicValue::Reader inpu
...
-      if (type.isData()) {
-        kj::Vector<byte> data;
-        for (auto arrayObject : value.getArray()) {
+    case schema::Type::DATA:
+      if (value.isArray()) {
+        auto array = value.getArray();
+        kj::Vector<byte> data(array.size());
+        for (auto arrayObject : array) {
           auto x = int(arrayObject.getNumber());
           KJ_REQUIRE(x >= 0 && x <= 255, "Number in array of bytes out of range.");
```
kamalmarhubi: could be `KJ_REQUIRE(byte(x) == x)` I think

kamalmarhubi: That would additionally catch non-integral values. Error message would have to change a tiny bit.

c++/src/capnp/compat/json.c++
```diff
@@ -380,72 +380,135 @@ void JsonCodec::encodeField(StructSchema::Field field, DynamicValue::Reader inpu
...
+        setFn(parseUInt64(value.getString()));
+      } else {
+        KJ_FAIL_REQUIRE("Expected numeric value");
+      }
       break;
-    case JsonValue::NUMBER:
-      set(value.getNumber());
+    case schema::Type::FLOAT32:
+    case schema::Type::FLOAT64:
+      if (value.isNull()) {
```
kamalmarhubi: `switch`

c++/src/capnp/compat/json.c++
```diff
@@ -380,72 +380,135 @@ void JsonCodec::encodeField(StructSchema::Field field, DynamicValue::Reader inpu
...
-        set(value.getString());
+        KJ_FAIL_REQUIRE("Expected string value");
       }
       break;
-    case JsonValue::ARRAY: {
-      if (type.isData()) {
-        kj::Vector<byte> data;
-        for (auto arrayObject : value.getArray()) {
+    case schema::Type::DATA:
+      if (value.isArray()) {
```
kamalmarhubi: `KJ_REQUIRE(value.isArray())`

c++/src/capnp/compat/json.c++
```diff
@@ -380,72 +380,135 @@ void JsonCodec::encodeField(StructSchema::Field field, DynamicValue::Reader inpu
...
       } else {
-        decodeArray(value.getArray());
+        KJ_FAIL_REQUIRE("Expected string value");
       }
       break;
-    }
-    case JsonValue::OBJECT:
-      decodeObject(value.getObject());
+    case schema::Type::LIST:
+      if (value.isNull()) {
```
kamalmarhubi: empty braces?

kamalmarhubi:
Thanks for humouring me with the extra commit. It's looking pretty nice. I made a few mostly stylistic comments, then it's probably time to hand it back over to @kentonv for a final look.

c++/src/capnp/compat/json.c++
katreniak: I converted everything to switch.

katreniak:
Review comments addressed.

c++/src/capnp/compat/json.c++
```diff
@@ -379,13 +382,208 @@ void JsonCodec::encodeField(StructSchema::Field field, DynamicValue::Reader inpu
   encode(input, field.getType(), output);
 }
 
+namespace {
+int64_t parseInt64(kj::StringPtr s) {
+  char *endPtr;
+  errno = 0;
+  int64_t value = std::strtoll(s.begin(), &endPtr, 10);
+  KJ_REQUIRE(endPtr == s.end(), "String does not contain valid number", s);
+  KJ_REQUIRE(errno != ERANGE, "Value out-of-range", s);
```
kamalmarhubi: One thing I'm not sure of: is it best to check for the error return as well as `errno`? In this case, that'd be checking if it was `LLONG_MIN` or `LLONG_MAX`. The general protocol for `errno` is to ignore it unless the return value indicated an error.

c++/src/capnp/compat/json.c++
```diff
@@ -379,13 +382,208 @@ void JsonCodec::encodeField(StructSchema::Field field, DynamicValue::Reader inpu
...
+template <typename SetFn, typename DecodeArrayFn, typename DecodeObjectFn>
+void decodeField(Type type, JsonValue::Reader value, SetFn setFn, DecodeArrayFn decodeArrayFn,
+    DecodeObjectFn decodeObjectFn) {
+  // This code relies on conversions in DynamicValue::Reader::as<T>.
+  switch(type.which()) {
+    case schema::Type::VOID:
+      break;
+    case schema::Type::BOOL:
+      switch (value.which()) {
+        case JsonValue::BOOLEAN:
```
kamalmarhubi: micronit is here:

```
KJ_REQUIRE(value.isBoolean(), "Expected boolean value");
setFn(value.getBoolean();
```

kamalmarhubi: (and for the other single-case switch statements)

c++/src/capnp/compat/json.c++
```diff
@@ -379,13 +382,208 @@ void JsonCodec::encodeField(StructSchema::Field field, DynamicValue::Reader inpu
...
+      }
+      break;
+    case schema::Type::DATA:
+      switch (value.which()) {
+        case JsonValue::ARRAY: {
+          auto array = value.getArray();
+          kj::Vector<byte> data(array.size());
+          for (auto arrayObject : array) {
+            auto x = arrayObject.getNumber();
+            KJ_REQUIRE(byte(x) == x, "Number in array of bytes out of range.");
```
kamalmarhubi: "Number in byte array is not an integer in [0, 255]" or something like that. (integer part is missing here)

c++/src/capnp/compat/json.c++
```diff
@@ -379,13 +382,208 @@ void JsonCodec::encodeField(StructSchema::Field field, DynamicValue::Reader inpu
...
+          break;
+        }
+        default:
+          KJ_FAIL_REQUIRE("Expected data value");
+      }
+      break;
+    case schema::Type::LIST:
+      switch (value.which()) {
+        case JsonValue::NULL_:
+          // nothing to do
```
kamalmarhubi: :+1:

c++/src/capnp/compat/json.c++
```diff
@@ -379,13 +382,208 @@ void JsonCodec::encodeField(StructSchema::Field field, DynamicValue::Reader inpu
...
+          break;
+        case JsonValue::STRING:
+          setFn(parseFloat64(value.getString()));
+          break;
+        default:
+          KJ_FAIL_REQUIRE("Expected float value");
+      }
+      break;
+    case schema::Type::TEXT:
+      switch (value.which()) {
```
kamalmarhubi: single case switch

c++/src/capnp/compat/json.c++
```diff
@@ -379,13 +382,208 @@ void JsonCodec::encodeField(StructSchema::Field field, DynamicValue::Reader inpu
...
+          break;
+        case JsonValue::ARRAY:
+          decodeArrayFn(value.getArray());
+          break;
+        default:
+          KJ_FAIL_REQUIRE("Expected list value");
+      }
+      break;
+    case schema::Type::ENUM:
+      switch (value.which()) {
```
kamalmarhubi: single case switch

c++/src/capnp/compat/json.c++
```diff
@@ -379,13 +382,208 @@ void JsonCodec::encodeField(StructSchema::Field field, DynamicValue::Reader inpu
...
+      decodeField((*fieldSchema).getType(), field.getValue(),
+          [&](DynamicValue::Reader value) { output.set(*fieldSchema, value); },
+          [&](List<JsonValue>::Reader array) {
+            decodeArray(array, output.init(*fieldSchema, array.size()).as<DynamicList>());
+          },
+          [&](List<JsonValue::Field>::Reader object) {
+            decodeObject(object, output.init(*fieldSchema).as<DynamicStruct>());
+          });
+    } else {
+      // Unknown json fields are ignored to allow schema evolution
```
kamalmarhubi: can you just drop the `else` cause then?

kamalmarhubi:
Ok I went over the whole thing one more time, and made a couple of really small nit comments. Otherwise I think this looks pretty good!

c++/src/capnp/compat/json.c++
katreniak: These single case switches make the code slightly more verbose, but they are no performance problem and they are consistent. I would like to keep them as they are. I expect / I want to make it easy to add more conversion. E.g. it makes sense to add conversion from integer to bool (0=false, everything else true) with little conflicts.

c++/src/capnp/compat/json.c++
katreniak: I want to keep the comment to make the intent explicit. I find the comment in empty `else {}` as the most readable way to express the intent.

c++/src/capnp/compat/json.c++
katreniak: I don't see point to complicate the code. Do we really care about strtoll implementation that returns errno==ERANGE  and it does not mean range error? I consider it invalid strtoll implementation. My interpretation is that strltoll is required to not set errno if conversion was successful.

c++/src/capnp/compat/json.c++
katreniak: Thanks. I will push on Monday.

c++/src/capnp/compat/json.c++
kamalmarhubi: :+1:

katreniak:
Improved error message, updated list of contributors. I am moving the change for [inf, nan serialization as string](https://github.com/sandstorm-io/capnproto/pull/297#issuecomment-202842288)  into separate PR #308.

kentonv:
@kamalmarhubi should I interpret your thumbs-up on the last comment as LGTM for the PR?

c++/src/capnp/compat/json.c++
katreniak: DynamicValue::Reader::as<T> throws on integer overflows.

katreniak:
@kentonv, AFAIK I addressed all review comments. Few points worth looking for you:

* Is this `kj`-ish  https://github.com/sandstorm-io/capnproto/pull/297/files#r58242351? But more important question is what should json decode do when it expects `bool` and it finds something else than `json bool`.  Fail? Interpret number 0 as false and other numbers as true? ...

* You asked to add a test that uses `initTestMessage()` and `checkTestMessage()`  on round-trip to JSON and back. This exposed the feature that `nan, inf, -inf` are serialized in json as `null`. I had to hack around this in the test. I proposed to serialize these numbers to "Infinity", "-Infinity", "NaN". Me and @kamalmarhubi [agreed that we disagree](https://github.com/sandstorm-io/capnproto/pull/297#issuecomment-203044775). So I moved the decision into PR #308. Do you have an opinion about this?

* Would you accept to merge these parse functions somewhere into kj? There is no json dependency in this code. It is just kj style wrapper around std function. https://github.com/sandstorm-io/capnproto/pull/297#discussion-diff-57689068

* Would you accept string -> number conversion into DynamicValue::Reader::as<T>? https://github.com/sandstorm-io/capnproto/pull/297#discussion-diff-57689657

kamalmarhubi:
@kentonv yeah, at this point it's LGTM, especially with nan / inf stuff moved to #308. I think @katreniak has summarised it quite well above.

c++/src/capnp/compat/json-test.c++
```diff
@@ -214,7 +214,7 @@ KJ_TEST("decode all types") {
   CASE(R"({"int32Field":"2147483647"})", root.getInt32Field() == 2147483647);
   CASE(R"({"int64Field":-9007199254740992})", root.getInt64Field() == -9007199254740992LL);
   CASE(R"({"int64Field":9007199254740991})", root.getInt64Field() == 9007199254740991LL);
-  CASE(R"({"int64Field":"-9223372036854775808"})", root.getInt64Field() == -9223372036854775808LL);
+  CASE(R"({"int64Field":"-9223372036854775808"})", root.getInt64Field() == -9223372036854775808ULL);
```
katreniak: I have no idea why both gcc and clang warn when this constant is LL. Changing to ULL fixed integer range warning.

katreniak:
I just pushed another commit that fixes compiler warnings in code added in this PR.

kentonv:
@katreniak Answering your questions:

* While I would personally have used `isBool()` rather than a switch, I don't have a strong opinion on this.

* I think we should only accept `true` and `false` for boolean fields. I have never liked implicit conversion of numbers to booleans -- I think it leads to uncaught type errors.

* I actually slightly prefer passing Infinity, -Infinity, and NaN as strings, but this was a point where @kamalmarhubi and I disagreed earlier. Note that `JSON.stringify()` in browsers seems to coerce them to null -- which I find surprising and disappointing, but that's what they do.

* I'd love to have `parseInt` functions in KJ. I am thinking maybe the best place to put these is actually as methods of `kj::StringPtr`. Hmm, maybe instead of `parseInt64()`, `parseUInt64()`, etc., the method could be `parseAs<int64>()`, `parseAs<uint64>()`, etc.? (Of course, they wouldn't actually be defined as a template, but rather as a series of specializations.) Should we also have `tryParse` variants that return Maybe? I think that the integer functions should default to auto-detecting base (pass 0 to `strtol`), but could optionally take a base as a parameter.

* Sorry, I do not like the idea of string -> number conversions in `DynamicValue::Reader::as`. The hack of using strings for int64 is very specific to JSON/Javascript and should not bleed over to other environments IMO.

katreniak:
> I actually slightly prefer passing Infinity, -Infinity, and NaN as strings, but this was a point where @kamalmarhubi and I disagreed earlier.

It is up to you to accept or reject  #308. I don't understand why capnp should align with browser behavior instead of other serialization frameworks. Capnp json encode already diverges from browsers by using string for [u]int64 type. I have hard time to accept the unnecessary loss of information. 

> I'd love to have parseInt functions in KJ. ... `parseAs<int64>(), parseAs<uint64>(), tryParseAs<int64>(), tryParseAs<uint64>()`

I am likely to push some patches over time because the standard C++ api is horrible. Kj-ish api in `kj::StringPtr` will be great. `kj::StringPtr` requires null terminator, but `strtoFoo` methods require it too. 

> I think that the integer functions should default to auto-detecting base

Auto detecting base leads to surprising results when string starts with 0. Most humans interpret such a number with base 10. I fixed few such bugs by explicitly passing base 10. I find it safer to default to base 10.

Anything more to do on this PR to make it mergeable?

kentonv:
@katreniak The PR currently has merge conflicts against master. You'll need to rebase it and fix those.

katreniak:
Conflict in CONTRIBUTORS file. This honor brings complications :)

I merged master branch into my branch to resolve conflict and pushed here. This is rebased branch if you prefer rebase: https://github.com/sandstorm-io/capnproto/compare/master...katreniak:bka-jsonDecodeRebased?expand=1

kentonv:
Thanks!

PR #299, opened 2016-03-26
Fix uninitialized members of ListBuilder

maurer:
Adding a `KJ_DASSERT` in the `setListPointer` logic flagged
non-word-multiple data sections in `INLINE_COMPOSITE` lists, which
should be impossible. This traced back to uninitialized member variables
in `ListBuilder` in the case that it was created from a null pointer.


kentonv:
Thanks!

PR #301, opened 2016-03-30
Fixed access violation in GetFunctorStartAddress()

szenker:
The method to determine, if a pointer to a member function points to a virtual method or not, depends on the ABI. Please see also:
* http://infocenter.arm.com/help/topic/com.arm.doc.ihi0041e/IHI0041E_cppabi.pdf chapter 3.2
* https://blog.mozilla.org/nfroyd/2014/02/20/finding-addresses-of-virtual-functions

To reproduce it, run unit test Async/GetFunctorStartAddress on an ARM hardware. If you don't have one, just compile cap'n'proto with an ARM compiler and run it via QEMU. #

### DIFF
--- a/c++/src/kj/async-inl.h
+++ b/c++/src/kj/async-inl.h
@@ -285,9 +285,14 @@ class PtmfHelper {
 #undef BODY
 
   void* apply(void* obj) {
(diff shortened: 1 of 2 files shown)


kentonv:
Thanks!

PR #302, opened 2016-03-30
Propagate DocComments into binary schema

eqvinox:
Not sure if this is done correctly - I'm not a C++ developer. It does work...

(Tested with pycapnp, accessing `X.schema.node.docComment` and `X.field.proto.docComment`)


paulproteus:
Hi! Thanks for writing this! For what it's worth, it'll take until Friday for @kentonv to have time to thoughtfully review this, but of course I'm not Kenton, so I can't say for sure. I do know from working with Kenton that Fridays are days he typically does a lot of code review.

kentonv:
Thanks!

kentonv:
Unfortunately this is problematic as-is because it adds a lot of bloat to descriptors which often isn't wanted. For example, when encoded descriptors are embedded into generated code, all the comments will be included.

It seems like every code generator that embeds descirptors now has to strip out the docComment fields. We could instead strip them out before passing on to the code generator but of course that would defeat the purpose of having the code generator generate proper doc comments.

kentonv:
I'm sorry, I'm going to need to revert this.

I tired making the C++ code generator strip out the doc comments in the embedded descriptors, but it broke things because it meant that the default values were no longer at the predicted offsets within the descriptor.

I think we will need to consider a different approach in which doc comments are delivered out-of-band, like Protobufs' SourceInfo.

PR #303, opened 2016-03-31
style-guide: fix isConst and UnConst examples

katreniak:
Problems:
* missing ; after struct declarations
* missing static for constexpr bool value
* UnConst_<T> -> UnConst_<const T>
* UnConst = -> UnConst = typename


kentonv:
Indeed, thanks!

PR #305, opened 2016-04-02
headers: check whether CAPNP_HEADER_WARNINGS is defined

### LINKED ISSUE #304: CAPNP_HEADER_WARNINGS and -Wundef
When not installed to the system include directory and `-Wundef` is set for client code, `#if defined(__GNUC__) && !CAPNP_HEADER_WARNINGS`, which is included in a variety of installed headers, triggers compilation warnings.

Would you be interested in a patch that checks that `CAPNP_HEADER_WARNINGS` is defined?

liamstask:
fixes #304


kentonv:
Thanks!

PR #306, opened 2016-04-02
Fix building with Xcode 7.3

mologie:
This pull requests fixes the following compiler error of the master branch with Apple's latest toolchain:

```
src/capnp/compat/json.c++:521:20: error: no member named 'strtod' in namespace 'std'; did you mean simply 'strtod'?
    double value = std::strtod(numberStr.begin(), &endPtr);
                   ^~~~~~~~~~~
                   strtod
/usr/include/stdlib.h:162:9: note: 'strtod' declared here
double   strtod(const char *, char **) __DARWIN_ALIAS(strtod);
         ^
1 error generated.
make: *** [src/capnp/compat/json.lo] Error 1
```

### DIFF
--- a/c++/src/capnp/compat/json.c++
+++ b/c++/src/capnp/compat/json.c++
@@ -20,9 +20,9 @@
 // THE SOFTWARE.
 
 #include "json.h"
-#include <math.h>    // for HUGEVAL to check for overflow in std::strtod
-#include <cstdlib>   // std::strtod
-#include <errno.h>   // for std::strtod errors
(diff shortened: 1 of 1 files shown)


kentonv:
Hmm, I'd actually prefer to remove the `std::` prefix. The `<cwhatever>` headers are widely recognized to be a lost cause, and we already depend on the non-namespaced name at nearly all other call sites.

mologie:
Amen to that. I've updated the PR.

kentonv:
Thanks!

PR #308, opened 2016-04-06
Bka json inf nan as string

katreniak:
Based on #297.

Discussed here: https://github.com/sandstorm-io/capnproto/pull/297#issuecomment-202842288

I split this change into separate PR so that it is easy to discard.


kentonv:
As discussed I think this behavior is preferable, therefore merged. Thanks!

kentonv:
BTW it looks like your email address @digitalstrom.com is not connected to your Github account, therefore your commits show up with unknown author. I recommend adding it here, which will cause Github to retroactively credit your commits: https://github.com/settings/emails

PR #309, opened 2016-04-06
Trivial changes required to support Visual Studio 2015 Update 2

harrishancock:
Here is the first batch of commits ready to merge from my VS2015 branch, rebased onto the current capnproto master. These are primarily trivial changes, as requested: the only workaround included is a small one having to do with MSVC's preprocessor.

There are two bootstrap regeneration commits. I noticed that json.capnp's generated C++ code was out of date on master. This puzzled me, so the first commit is a bootstrap regeneration to get that difference out of the way, so it didn't get folded into the later changes I made.

I decomposed Gordon's work into smaller commits, keeping him as commit author. I dropped some of his changes because they were no longer necessary with VS2015 Update 2 (he compiled against Update 1), some I dropped because I found a better way to do things, and some I held back because they need discussion first.


kentonv:
Awesome work! These all look great. Can you add a commit adding yourself to CONTRIBUTORS? :)

harrishancock:
Done! Thanks for reviewing. Once this is merged, I'll start readying another batch of commits focused on re-enabling lite-mode. I gather from the discussion on node-capnp (https://github.com/kentonv/node-capnp/issues/28#issuecomment-208672744) that that would be of some short-term benefit. I'll be in touch via the capnproto group!

PR #311, opened 2016-04-08
Declare SourceInfo in schema.capnp as a proposal.

kentonv:
@eqvinox This is how I propose we represent doc comments (alternative to #302). We can also use this for things like source code locations (line/column numbers) of declarations.

The `SchemaLoader` and `SchemaParser` classes will probably need to be updated to support loading `SourceInfo`s separately and then looking up `SourceInfo` for a particular node.

Unfortunately I don't have time to work on it right this moment... :/


c++/src/capnp/schema.capnp
```diff
@@ -169,6 +169,36 @@ struct Node {
...
+    # generally used at runtime. Instead of educating everyone on how to remove it when not needed,
+    # we choose to deliver it out-of-band such that use cases that *do* need SourceInfo can request
+    # it explicitly. For example, `CodeGeneratorRequest` delivers a list of `SourceInfo`s in
+    # parallel with the list of `Node`s.
+
+    id @0 :UInt64;
+    # ID of the node we're describing.
+
+    docComment @1 :UInt64;
+    # Doc comemnt attached to the node.
```
zombiezen: nit: comment

c++/src/capnp/schema.capnp
```diff
@@ -169,6 +169,36 @@ struct Node {
...
+    # accidentally included in cases where it is not needed. SourceInfo can be very big and is not
+    # generally used at runtime. Instead of educating everyone on how to remove it when not needed,
+    # we choose to deliver it out-of-band such that use cases that *do* need SourceInfo can request
+    # it explicitly. For example, `CodeGeneratorRequest` delivers a list of `SourceInfo`s in
+    # parallel with the list of `Node`s.
+
+    id @0 :UInt64;
+    # ID of the node we're describing.
+
+    docComment @1 :UInt64;
```
zombiezen: Why isn't this `:Text`?

kentonv: ... apparently because I was typing very quickly and there's no implementation to test against. :P

c++/src/capnp/schema.capnp
kentonv: fixed

c++/src/capnp/schema.capnp
kentonv: fixed

zombiezen:
I like this approach. If these fields got populated, I'd update capnpc-go with this change.

eqvinox:
So... while I can justify sinking some time on the topic, that's of no use as my C++ skills barely covered my original patch :(.  SourceInfo does contain the info I need and I suppose the distinction with ancilliary information makes sense... but I can't provide an implementation.

(Stupidly enough, I can't even locally patch the compiler on my build -- with our python module we only ship the .capnp and it's dynamically processed on the end user's box to generate the docs from setup.py...  I guess I could ship the generated docs though.)

PR #312, opened 2016-04-12
Compile fix for musl libc

xming:
glibc is not the only libc on Linux, this fixes compilation for musl based Linux (like Alpine Linux).

### DIFF
--- a/c++/src/kj/exception.c++
+++ b/c++/src/kj/exception.c++
@@ -19,19 +19,19 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
+#include <stdlib.h>
 #include "exception.h"
 #include "string.h"
 #include "debug.h"
 #include "threadlocal.h"
 #include "miniposix.h"
-#include <stdlib.h>
 #include <exception>
 #include <new>
 #include <signal.h>
 #include <sys/mman.h>
 #include "io.h"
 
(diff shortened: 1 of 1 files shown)


c++/src/kj/exception.c++
```diff
@@ -19,19 +19,19 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
+#include <stdlib.h>
```
kentonv: Is it necessary to move this include up? I prefer to have the corresponding header file be the first include, as this ensures that the header itself includes all of its dependencies.

xming: Yes, musl is very strict about the correct order of includes.

kentonv: If there was an error with the old ordering, that implies that one of the other header files that was previously included before stdlib.h is using declarations from stdlib.h without correctly including it. The correct fix here is to add `#include <stdlib.h>` to the offending header file. Moving the include up inside `exception.c++` is the wrong solution because this is only masking the bug that is in the header file.

Can you please identify which header file is using stdlib declarations without including stdlib.h, and then add the include to that header, rather than moving it here?

justincormack: I don't see any build errors on Alpine 3.4 with the ordering unchanged, so not sure there is an issue.

kentonv:
I think this is superseded by #353, so I'm closing this. Let me know if you disagree.

Issue #235, opened 2015-08-21
Test `Exception/TrimSourceFilename` fails when building from outside source directory.


adrianratnapala:
With autotools, it is possible to build from outside the source directory by running `configure` from your desired build directory.   This works also with capnproto,  except that one of the unit tests fails when you run `make check`.

**To reproduce:**

* Get clean capnproto sources
* Make a build directory beside the sources.  Run `autoreconf` and ` configure` from there.
* Then run `make check`.

For example:

    mkdir test-cpnp
    cd test-cpnp
    git clone https://github.com/sandstorm-io/capnproto.git

    mkdir build
    cd build
    cpDir=../capnproto/c++

    autoreconf -i "$cpDir"
    "$cpDir/configure"
    make -j6 check

**Observation:**

`capnp-test` fails.  If we run it by hand we get

    [ TEST ] kj/exception-test.c++:30: legacy test: Exception/TrimSourceFilename  capnproto/c++/src/kj/exception-test.c++:31: failed: expected (trimSourceFilename("../capnproto/c++/src/kj/exception-test.c++")) == ("kj/exception-test.c++"); trimSourceFilename("../capnproto/c++/src/kj/exception-test.c++") = capnproto/c++/src/kj/exception-test.c++; kj/exception-test.c++
    stack: 0x42accd 0x42a6d5 0x7fe254f7ac6e 0x7fe254d5cd96 0x7fe254f7c241 0x7fe254f7d02d 0x7fe254d69048 0x7fe254d6a9c7 0x7fe254d5cd96 0x7fe254d6ab43 0x7fe254f7983f 0x7fe253f11790 0x41d979
    [ FAIL ] kj/exception-test.c++:30: legacy test: Exception/TrimSourceFilename

**NB:** If I build from inside the source directory, all tests pass.

**Expectation:**

All tests should pass, even when using a separate build directory.

kentonv:
Hmm, I guess we've tested compiling from a `build` subdirectory but not from entirely out-of-tree.

Thanks for the report. FWIW, this only matters for making debug output a bit more readable. I'll try to fix it before the next release (for which there's no particular timeframe currently).

adrianratnapala:
Yes suspected it was not very important.  The "fix" might simply be to weaken the test.  For example, you can make it XFAIL if it detects a "../" in certain place.

frol:
FYI, I'm getting the same fail test using CMake and building in `<project_root>/build` folder (`cmake ../c++`).

[...]

PR #325, opened 2016-05-09
Modify `trimSourceFilename()` to remove all text before a prefix.

pqu:
A quick stab at fixing #235 (and the tests under CMake) by looking for the `PREFIXES` anywhere in the string and simply removing all text up-to-and-including them.

### DIFF
--- a/c++/src/kj/exception.c++
+++ b/c++/src/kj/exception.c++
@@ -227,9 +227,11 @@ kj::StringPtr trimSourceFilename(kj::StringPtr filename) {
   };
 
 retry:
+  const char* prefixLocation = nullptr;
   for (const char* prefix: PREFIXES) {
-    if (filename.startsWith(prefix)) {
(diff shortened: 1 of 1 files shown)


kentonv:
Hmm, I think this isn't quite the right behavior for `../` or the ekam directories. Also it could misfire on directory names that happen to end in `src` or `tmp`.

What does make sense would be looking for `/src/` or `/tmp/` anywhere in the string and, if found, removing it and everything before it.

pqu:
It did seem a bit aggressive. Should `PREFIXES` only contain `/src/` and `/tmp/` then? I don't know what kind of directory structure ekam creates, but I can't think of many cases where the last occurrence of `/src/` wouldn't be a good splitting point.

kentonv:
I was going to type a long explanation of what all the different prefixes mean but then I realized that it would be easier for me to just fix it. :)

I think 22c1b62 should solve all issues. Sorry this took a while.

PR #328, opened 2016-05-18
Implement method kj::StringPtr:parseAs<NUMBER_TYPE>

katreniak:
This  PR moves number parsing code from capnp json decoder to kj::StringPtr as discussed [here](https://github.com/sandstorm-io/capnproto/pull/297#issuecomment-212764806).

Questions:

1. Integer parsing methods use base 10. My problem with auto detected base is that 0 prefixed numbers are correct decimal numbers but they are interpreted as octal. Does anybody use octal numbers? I would be ok to automatically interpret hexadecimal numbers prefixed with "0x" or "0X". 
2. There is no optional base parameter to integer parsing. I am not sure how to write the template magic that accepts optional argument only for integral types. Is this important / necessary?
3. There is one more call to `strtod` in json parser. But it throws when HUGE floating number overflows instead of interpreting the number as INF. Current `parseAs<double>()` code parses it as INF. What is the desired behavior for each of them?

@kentonv @kamalmarhubi

### DIFF
--- a/c++/src/kj/string.c++
+++ b/c++/src/kj/string.c++
@@ -69,7 +69,7 @@ T parseInteger(const StringPtr& s) {
     long long max = static_cast<T>(maxValue);
     return static_cast<T>(parseSigned(s, min, max));
   } else {
-    long long max = static_cast<T>(maxValue);
+    unsigned long long max = static_cast<T>(maxValue);
     return static_cast<T>(parseUnsigned(s, max));
   }
 }
--- a/c++/src/kj/string.c++
+++ b/c++/src/kj/string.c++
@@ -34,12 +34,16 @@ namespace kj {
 #endif
 
 namespace {
+bool isHex(const char *s) {
+  if (*s == '-') s++;
+  return s[0] == '0' && (s[1] == 'x' || s[1] == 'X');
+}
 
 long long parseSigned(const StringPtr& s, long long min, long long max) {
   KJ_REQUIRE(s != nullptr, "String does not contain valid number", s);
   char *endPtr;
   errno = 0;
-  auto value = strtoll(s.begin(), &endPtr, 10);
+  auto value = strtoll(s.begin(), &endPtr, isHex(s.cStr()) ? 16 : 10);
   KJ_REQUIRE(endPtr == s.end(), "String does not contain valid number", s);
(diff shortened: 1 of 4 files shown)


c++/src/kj/string.c++
```diff
@@ -33,6 +34,71 @@ namespace kj {
...
+  KJ_REQUIRE(errno != ERANGE, "Value out-of-range", s);
+  KJ_REQUIRE(value <= max, "Value out-of-range", value, max);
+  KJ_REQUIRE(s[0] != '-', "Value out-of-range", s); //strtoull("-1") does not fail with ERANGE
+  return value;
+}
+
+template <typename T>
+T parseInteger(const StringPtr& s) {
+  if (std::numeric_limits<T>::is_signed) {
+    auto min = (long long) std::numeric_limits<T>::min();
```
kentonv: You could avoid pulling in `std::numeric_limits` here by using `minValue` and `maxValue`.

    if (static_cast<T>(minValue) < 0) {
      long long min = static_cast<T>(minValue);
      long long max = static_cast<T>(maxValue);

katreniak:
* removed #include <limits>
* added support for hex numbers
* added documentation

c++/src/kj/string.c++
```diff
@@ -61,12 +60,12 @@ unsigned long long parseUnsigned(const StringPtr& s, unsigned long long max) {
...
-  if (std::numeric_limits<T>::is_signed) {
-    auto min = (long long) std::numeric_limits<T>::min();
-    auto max = (long long) std::numeric_limits<T>::max();
+  if (static_cast<T>(minValue) < 0) {
+    long long min = static_cast<T>(minValue);
+    long long max = static_cast<T>(maxValue);
     return static_cast<T>(parseSigned(s, min, max));
   } else {
-    auto max = (unsigned long long) std::numeric_limits<T>::max();
+    long long max = static_cast<T>(maxValue);
```
kentonv: This should be `unsigned long long`, right?

katreniak: True. Fixed in 651aaf3b5def9808c3272fb5565122a3921f9aa2

kentonv:
Thanks!

PR #329, opened 2016-05-18
Bka fix union with group in generic struct

katreniak:
Union with group in generic struct leads to compilation error in generated header:

````cpp
In file included from external/capnproto/c++/src/capnp/test_capnp/capnp/test.capnp.c++:4:0:
external/capnproto/c++/src/capnp/test_capnp/capnp/test.capnp.h:9565:10: error: need ‘typename’ before ‘capnproto_test::capnp::test::TestGenerics<Foo, Bar>::Ug::Reader’ because ‘capnproto_test::capnp::test::TestGenerics<Foo, Bar>::Ug’ is a dependent scope
   inline Ug::Reader getUg() const;
````
Relavant parts in header file:

````cpp
template <typename Foo = ::capnp::AnyPointer, typename Bar = ::capnp::AnyPointer>
struct TestGenerics {
  ...
  struct Ug;
};

template <typename Foo, typename Bar>
class TestGenerics<Foo, Bar>::Reader {
  ...
  inline Ug::Reader getUg() const;
};
````

Compiler misses `typename` keyword before Ug::Reader. 

First commit ads test that leads to compilation error. Second commit fixes the problem (for me).

### DIFF
--- a/c++/src/capnp/compiler/capnpc-c++.c++
+++ b/c++/src/capnp/compiler/capnpc-c++.c++
@@ -1121,39 +1121,39 @@ private:
         return FieldText {
             kj::strTree(
                 kj::mv(unionDiscrim.readerIsDecl),
-                "  inline ", titleCase, "::Reader get", titleCase, "() const;\n"
+                "  inline typename ", titleCase, "::Reader get", titleCase, "() const;\n"
                 "\n"),
 
             kj::strTree(
                 kj::mv(unionDiscrim.builderIsDecl),
-                "  inline ", titleCase, "::Builder get", titleCase, "();\n"
-                "  inline ", titleCase, "::Builder init", titleCase, "();\n"
+                "  inline typename ", titleCase, "::Builder get", titleCase, "();\n"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by kentonv** [comment]
Interesting, this adds `typename` even when the context is not generic. I thought that wouldn't work -- I thought the compiler complained when using `typename` outside of a template -- but it looks like C++11 made this allowed as long as the type name is nested (has a `::` in it).

But does this work correctly in MSVC? @harrishancock maybe you could check?

If not, we'll need to add `typename` only when `templateContext.isGeneric()` returns true, I think.

**Comment by harrishancock** [comment]
MSVC compiles this correctly:

```
struct A {
  struct B {
    typename A::B f () { return {}; }
  };
};

int main () {
  auto b = typename A::B().f();
  return 0;
}
```

Testing @katreniak 's branch now.

harrishancock:
@kentonv, MSVC compiles code generated by this branch correctly and all tests pass (save TrimSourceFileName).

kentonv:
Thanks!

PR #331, opened 2016-05-19
Re-enable MSVC lite-mode

harrishancock:
This branch contains the changes required to make lite-mode Cap'n Proto master compile again under MSVC. I tested this branch by using it to do the three-step "external capnp" bootstrapping dance (Linux, then MinGW, then MSVC).

There were a couple tests which needed at least partial disabling because they used either RPC features, or a brand pointer, neither of which are available in lite-mode.

Otherwise, all MSVC-compiled lite-mode tests pass except Exception/TrimSourceFileName: https://gist.github.com/harrishancock/6dcec0534341876a3356d081d841a79c

MSVC choked on the current implementations of both Kind_ and FromAny_. I replaced them with a std::void_t-style implementation (I named the voider alias kj::VoidSfinae). It is more verbose, but MSVC accepts it.

I replaced unistd.h with miniposix.h in all places required by MSVC to compile, including some non-lite-mode areas out of optimism that I'll be able to get the compiler ported next. Linux remains happy with the header changes, but I haven't tested on Mac. I also fiddled with the arguments passed to _pipe() on Windows (including MinGW).

There are three ugly workarounds included in this branch:
- I added a CAPNP_STYLE(T) macro, because style<T>() won't work as a template argument. I hated adding this, but I didn't see a better way to fix it. This is only necessary because of FromAny, by the way. See commit ce30ee3.
- There are a couple circumstances in which MSVC cannot compile the expression `noexcept(instance<T&>().~T())` (used to detect if a type has a noexcept destructor). I worked around it with a big ugly macro and a compiler intrinsic. See commit d147ea9.
- Overloading a constructor on parameters whose types are computed using decltype() doesn't work in MSVC, which affects Quantity::Quantity(decltype(maxValue)). See commit a7ffa38.

Lastly, a few minor header and linker changes were also required.

### DIFF
--- a/c++/src/capnp/encoding-test.c++
+++ b/c++/src/capnp/encoding-test.c++
@@ -1667,6 +1667,8 @@ TEST(Encoding, Embeds) {
     checkTestMessage(reader.getRoot<TestAllTypes>());
   }
 
+#if !CAPNP_LITE
+
   {
     MallocMessageBuilder builder;
     auto root = builder.getRoot<TestAllTypes>();
@@ -1674,6 +1676,8 @@ TEST(Encoding, Embeds) {
     kj::StringPtr text = test::EMBEDDED_TEXT;
     EXPECT_EQ(kj::str(root, '\n').size(), text.size());
   }
+
+#endif // CAPNP_LITE
 
   {
     checkTestMessage(test::EMBEDDED_STRUCT);
--- a/c++/src/capnp/any-test.c++
+++ b/c++/src/capnp/any-test.c++
@@ -247,6 +247,8 @@ TEST(Any, AnyStructListCapInSchema) {
     checkList(anyList.as<List<int>>(), {123, 456, 789});
   }
 
+  #if !CAPNP_LITE
+  // This portion of the test relies on a Client, not present in lite-mode.
   {
     kj::EventLoop loop;
     kj::WaitScope waitScope(loop);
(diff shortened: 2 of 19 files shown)


c++/src/capnp/encoding-test.c++
```diff
@@ -1667,6 +1667,8 @@ TEST(Encoding, Embeds) {
     checkTestMessage(reader.getRoot<TestAllTypes>());
   }
 
+#if !CAPNP_LITE
```
kentonv: A clearer explanation for this commit would be: Stringification is not available in lite mode, so skip test that relies on it.

(Stringification depends on the dynamic API which by definition is not available in lite mode.)

kentonv:
That's all the comments I have. This looks pretty great.

(At first I didn't like the name `VoidSfinae` but when I tried to think of a better name I ended up with the same thing... heh.)

harrishancock:
I know what you mean about VoidSfinae. :) I thought myself in a couple circles before giving up.

I updated the branch:
- commit 1909684 and commit 25ce8b8 now have TODO(msvc)
- the CAPNP_STYLE() macro is replaced by spelling out the style condition in FromAny_'s EnableIf
- the stringification test commit message is worded better

I discovered we actually do need kind<T>() exposed in lite mode for a different reason, so I added one commit at the head of the branch to cover that.

Thanks for the feedback!

kentonv:
Thanks!

PR #333, opened 2016-05-27
Bka cleanups - fix compiler warnings in kj tests

katreniak:


c++/src/kj/test.c++
```diff
@@ -346,6 +346,8 @@ public:
     if (passCount > 0) write(GREEN, kj::str(passCount, " test(s) passed"), "");
     if (failCount > 0) write(RED, kj::str(failCount, " test(s) failed"), "");
     context.exit();
+
+    return true;
```
kentonv: Hmm, which compiler warned about this missing return? `context.exit()` is marked with the `noreturn` attribute, which should prevent such warnings.

Does it work if you put `KJ_UNREACHABLE` here instead of `return true`? I generally prefer that as it is more self-documenting.

katreniak: Warning is generated by gcc 5.3.1. Sorry, I did not really read the code. I will try KJ_UNREACHABLE on Monday.

Clang was fine. 

There is one more buggy warning in capnp where gcc warns that kj::Maybe instance may not be initialized. Do you care about warning free gcc build in master?

kentonv: Yes, I care that the build succeeds with the warning settings fond at the top of `super-test.sh`, namely:

    -Wall -Wextra -Wno-strict-aliasing -Wno-sign-compare -Wno-unused-parameter

However, for GCC (not Clang), we also add `-Wno-maybe-uninitialized` because it seems to report a lot of bogus warnings of this type, especially related to kj::Maybe. The test script uses Valgrind to check for actual uninitialized variable usage, so the GCC warning is unnecessary.

katreniak:
KJ_UNREACHABLE fixed the warning

PR #335, opened 2016-05-31
WIP OpenSSL TLS bindings

kentonv:
My memorial day weekend project... >_>


kentonv:
@benlaurie Would you care to review this, or could you suggest someone who might be able to review this, to verify that I'm using the OpenSSL API correctly and doing all the right things security-wise?

Note that all the interesting code is in the last commit -- 01adf56.

PR #337, opened 2016-06-03
kj/io: Make FdOutputStream::write fallback to OutputStream::write.

tgockel:
Completely irrelevant change, but it bothered me just enough.


kentonv:
Hah, sure, that makes things slightly cleaner. Thanks!

PR #339, opened 2016-06-14
capnpc-c++: fix generated code for union inside template

katreniak:


kentonv:
Thanks!

kentonv:
This change doesn't pass tests:
* The new test throws an exception due to trying to read a union field that isn't active.
* The bootstraps were not regenerated, so the bootstrap test failed.

I've fixed these in fc11baa, but please be sure to run the tests before sending next time. Thanks!

Issue #334, opened 2016-05-30
Generated method asGeneric is broken for Inner template classes.


katreniak:
Capnp declaration:
````cpp
struct TestGenerics(Foo, Bar) {
  struct Inner {
    foo @0 :Foo;
    bar @1 :Bar;
  }
````

Generated header:
````cpp
template <typename Foo, typename Bar>
class TestGenerics<Foo, Bar>::Inner::Builder {
public:
    typename Inner::Builder asGeneric() {
    return typename Inner::Builder(_builder);
  }
````

The generated code should look like this:
````cpp
template <typename Foo, typename Bar>
class TestGenerics<Foo, Bar>::Inner::Builder {
public:
  template <typename Foo2 = ::capnp::AnyPointer, typename Bar2 = ::capnp::AnyPointer>
  typename TestGenerics<Foo2, Bar2>::Builder asGeneric() {
    return typename TestGenerics<Foo2, Bar2>::Inner::Builder(_builder);
  }
````

Call to `node.getParameters()` on `Inner` node in `capnp-c++.c++` returns empty list . I believe that this is incorrect and `NodeTranslator::compileNode` in `node-translator.c++` should add parameters from parent scopes into node parameters.

Any hints?

kentonv:
`Node.parameters` is working as intended -- you need to follow `scopeId` to the parent scope to discover if it has parameters too.

I see what you mean about `asGeneric()`, though. The tricky question is what do you do if both `TestGenerics` and `TestGenerics::Inner` have type parameters? Does `asGeneric()` concatenate them into its own type parameter list? The problem there is that it's intended that you can add new type parameters to the end of the list without breaking source compatibility, but if we're concatenating them then adding parameters to the parent will change the offset of the child parameters, breaking things.

Maybe we need a separate method to re-assign each of the parent scopes.

    innerBuilder.asTestGenerics<MyFoo, MyBar>()

katreniak:
Thank you for great explanation. I believed that I am walking wrong path.

I like your API proposal. I would like to align on more details:

````cpp
struct Outer(T1) {
  struct Middle(T2) {
    struct Inner(T3) {
      t1 @0 :T1;
      t2 @1 :T2;
      t3 @2 :T3;
    }
  }
}
````

Generated code:
````cpp
template <typename T32 = ::capnp::AnyPointer>
typename Inner<T32>::Reader asGeneric() {..}

template <typename T22 = ::capnp::AnyPointer, 
[...]

PR #343, opened 2016-06-20
capnp-c++: implement asGeneric method for outer template contexts

katreniak:
https://github.com/sandstorm-io/capnproto/issues/334

### DIFF
--- a/c++/src/capnp/compiler/capnpc-c++.c++
+++ b/c++/src/capnp/compiler/capnpc-c++.c++
@@ -571,8 +571,8 @@ private:
     TemplateContext(): parent(nullptr) {}
     explicit TemplateContext(schema::Node::Reader node)
         : parent(nullptr), node(node) {}
-    TemplateContext(const TemplateContext& parent, schema::Node::Reader node)
-        : parent(parent), node(node) {}
+    TemplateContext(const TemplateContext& parent, kj::StringPtr name, schema::Node::Reader node)
+        : parent(parent), name(name), node(node) {}
 
     bool hasParams() const {
       return node.getParameters().size() > 0;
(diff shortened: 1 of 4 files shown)


katreniak:
This change solves my problem in #334. Bootstraps regenerated, tests passed locally.

I struggled a bit how to add the functionality to capnp-c++ and to keep the style. Please, let me know if you have better proposal.

c++/src/capnp/compiler/capnpc-c++.c++
```diff
@@ -1719,6 +1724,56 @@ private:
...
+              "    return castAs<", innerType.size() ? "typename " : "", returnType(), ">();\n"
+              "  }\n"
+              "\n");
+          break;
+      }
+    }
+    KJ_IF_MAYBE(p, templateContext.getParent()) {
+      up = makeAsGenericDef(role, *p, p->getName(), kj::strTree(
+          templateContext.hasParams() ? "::template " : "::", type,
+          templateContext.args()).flatten());
```
kentonv: Seems like the `.flatten()` is not needed here?

kentonv: Oh I see, this is the argument to makeAsGenericDef(). Perhaps the `innerType` parameter should be type `StringTree`?

kentonv: Eh, this is an irrelevant nitpick. Never mind.

kentonv:
Looks great! Thanks.

c++/src/capnp/compiler/capnpc-c++.c++
katreniak: The code consumes innerType two times. That did not work with StringTree move only semantics. String and flatten() did the job ...

PR #348, opened 2016-07-11
Add missing include when building on Linux

vlovich:
Discovered trying to build on Android using CMake.  UIO_MAXIOV is only
exported by the linux/uio.h header which was never included.  Not sure
how normal Linux builds work.

There's also a missing reinterpet_cast & linking against -pthread is implicit on Android, so omit the -pthread flag when building against the NDK.

### DIFF
--- a/c++/src/kj/miniposix.h
+++ b/c++/src/kj/miniposix.h
@@ -44,6 +44,10 @@
 #include <sys/types.h>
 #endif
 
+#if !_WIN32
+#include <sys/uio.h>
+#endif
+
 namespace kj {
 namespace miniposix {
 
--- a/c++/CMakeLists.txt
+++ b/c++/CMakeLists.txt
@@ -63,7 +63,11 @@ else()
   #
   # We have to use -std=gnu++0x isntead of -std=c++11 because otherwise we lose
   # GNU extensions that we need.
-  add_compile_options(-std=gnu++0x -Wall -Wextra -Wno-strict-aliasing -Wno-sign-compare -Wno-unused-parameter -pthread)
+  add_compile_options(-std=gnu++0x -Wall -Wextra -Wno-strict-aliasing -Wno-sign-compare -Wno-unused-parameter)
+
+  if (NOT ANDROID)
+    add_compile_options(-pthread)
+  endif()
 endif()
 
 # Source =======================================================================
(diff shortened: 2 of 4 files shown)


kentonv:
Hi Vitali,

Thank for the fixes, though:

* It looks like the correct header is `sys/uio.h`, which really ought to be included on all systems, not just Linux. It looks like this was mistakenly omitted from `miniposix.h` but is included directly from the `.c++` files that use `writev`, which masked the problem.
* The `reinterpret_cast` shouldn't be needed as CMSG_DATA is documented as returning `unsigned char*` which is the same thing as `byte*`. What type is it returning on Andorid? Note that the `reinterpret_cast` may be erroneous on some platforms where CMSG_DATA returns a non-const pointer, so if a cast is needed it may need to be C-style.
* The `-pthread` flag doesn't just link against the pthread library but also instructs GCC to generate thread-safe code. Does anything bad happen if you leave it there on Android?

vlovich:
I'll update the sys/uio.h header.

Regarding the reinterpret_cast, this is the error:

capnproto/c++/src/kj/async-io.c++: In member function 'virtual kj::Promise<void> kj::{anonymous}::DatagramPortImpl::ReceiverImpl::receive()':
capnproto/c++/src/kj/async-io.c++:1232:45: error: invalid conversion from 'void*' to 'const byte* {aka const unsigned char*}' [-fpermissive]

The -pthread flag causes the library to fail to link because there's no libpthread (it's part of Bionic).  I'm just following the advice from SO: http://stackoverflow.com/questions/11048260/does-android-support-thread.  I haven't really done any NDK development before so I just followed the advice & verified that it works & I'm still able to construct & use threads (although whether the code remains thread-safe is unknown, but I'm assuming yes).

kentonv:
For the cast, I guess we'll have to change it to a C-style cast so that varying const-ness doesn't cause a problem.

I guess we'll have to hope that removing the `-pthread` flag doesn't cause GCC to emit thread-unsafe code. It's weird that I haven't seen this issue when compiling for Android in the past (I test every release on Android), but maybe it's because I was using the autotool build rather than the cmake build.

vlovich:
Issues raised have been addressed I believe.

c++/src/kj/miniposix.h
```diff
@@ -44,6 +44,8 @@
 #include <sys/types.h>
 #endif
 
+#include <sys/uio.h>
```
kentonv: Sorry but also this `#include` should not apply on Windows, so you'll need to place it in the "else" block of the `#if _WIN32` above.

(Apologies for the slow turn-around. Swamped as always.)

vlovich: Fixed.  No worries about the slow-turnaround.  Haven't been the quickest on my end either.

kentonv:
Thanks!

PR #353, opened 2016-08-10
Only Glibc on Linux has execinfo.h

justincormack:
Fix as done elsewhere so that it builds correctly on Musl libc
as well as Glibc.

Signed-off-by: Justin Cormack <justin@specialbusservice.com>

### DIFF
--- a/c++/src/kj/exception.c++
+++ b/c++/src/kj/exception.c++
@@ -33,7 +33,7 @@
 #endif
 #include "io.h"
 
-#if (__linux__ && !__ANDROID__) || __APPLE__
+#if (__linux__ && __GLIBC__) || __APPLE__
 #define KJ_HAS_BACKTRACE 1
 #include <execinfo.h>
 #endif


justincormack:
Oh I didn't see https://github.com/sandstorm-io/capnproto/pull/312

I think this is all that is needed, it builds and tests pass fine for me on Alpine 3.4.

paulproteus:
Hi @justincormack !

Thanks for submitting this. I see that in #312 there is an unanswered question about the order of various headers. If you can answer that technical question, which is hopefully somewhat-easy since you have an Alpine Linux build environment handy, it would help us merge #312.

Let me know what you think! And thanks again for submitting this PR and trying to make capnproto more portable. Work like yours makes capnproto available to more people, which makes me smile!

justincormack:
I don't see the issue at all. Also that patch has the now unnecessary `__ANDROID__` left in, which is not required, so I think this patch is the minimal required fix...

paulproteus:
Gotcha! Hooray. I hereby ping @kentonv who will do a final review and/or merge, hopefully within a week or so.

justincormack:
Thanks!

kentonv:
Thanks!

PR #356, opened 2016-08-31
Add accessors for getting DynamicStruct field as another reader or builder

diorcety:
Provide a way to get a struct field as another reader/builder. I'm using it to change the default behavour related to interface. This is more a ask for feature as PR form. Not certain that all is correct, but it works on my cases.

### DIFF
--- a/c++/src/capnp/any.h
+++ b/c++/src/capnp/any.h
@@ -145,7 +145,7 @@ struct AnyPointer {
   public:
     typedef AnyPointer Builds;
 
-    Builder() = delete;
+    Builder() = default;
     inline Builder(decltype(nullptr)) {}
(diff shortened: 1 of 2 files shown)


diorcety:
Hi,
Because get(field).as<T> already interpret/do some work/check on the underlying stuff (cf create of DynamicCapability::Client for interface).

kentonv:
Oh, you're trying to bypass the type checking.

I suggest that instead you cast the DynamicStruct to AnyStruct. AnyStruct gives you direct access to the underlying pointers without enforcing that you're using the right types (it should maybe be called RawStruct...).

I would prefer that DynamicStruct continue to provide strict (runtime) type checking so I'd rather not add these methods. Sorry.

diorcety:
Okay but with anystruct i haven't an anypointer for a given field.

PR #358, opened 2016-09-05
include stdlib.h to fix build on OSX

dwrensha:
Fixes [this problem](https://github.com/sandstorm-io/capnproto/commit/af899ceb5fc1c650090d7ae8ff79e925b1c7e9a9#commitcomment-18895614) reported by @kaos.


dwrensha:
Updated to use angle brackets like `<stdlib.h>`, matching other such includes in this repository.

PR #359, opened 2016-09-05
Fix Mac OS X build failure "undeclared identifier getenv"

mark-grimes:
Currently on Mac OS X there's a build error with

`node-translator.c++:35:10: error: use of undeclared identifier 'getenv'`

This just explicitly adds the `#include <stdlib.h>` which presumably is implicit on linux.

Mac OS X El Capitan
cmake 3.4.2
Apple LLVM version 7.3.0 (clang-703.0.29)


mark-grimes:
Just seen this is a duplicate of #358

PR #362, opened 2016-09-15
Fix segfault race condition when kj::Thread is detached

mark-grimes:
This is to address the point discussed on https://groups.google.com/forum/#!topic/capnproto/Qbn3SEKBqLg

I need to test a bit more (particularly windows) but I thought I'd open early to get comments on the general approach.  This breaks pretty much every kj memory management rule at some point.  In particular, I can see no way of achieving the result without effectively passing an unowned new to the OS thread creation function.  I've tried to mitigate this by having it securely held until confirmation the thread was successfully created, and wrapping back in a smart pointer as soon as the new thread starts.

Also, this now requires two heap allocations in the kj::Thread constructor - one for the pimple idiom members and one for the shared thread data.  This could be cut to one for the shared data but it would mean putting a load of cruft in the header.  Threads aren't often created in kj or capnproto so I'm hoping this isn't too much of an issue.

After writing this I'm starting to wonder if the functor and exception actually need to be protected with a mutex.  The points they're accessed from in each thread are fairly well defined.  It could be quicker to just use std::atomic on the reference counter.  Would that be acceptable use of the std lib in kj?

### DIFF
--- a/c++/src/kj/thread.c++
+++ b/c++/src/kj/thread.c++
@@ -21,6 +21,7 @@
 
 #include "thread.h"
 #include "debug.h"
+#include "mutex.h"
 
 #if _WIN32
 #include <windows.h>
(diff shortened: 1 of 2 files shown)


kentonv:
Hi Mark!

Thanks for working on this. This seems to me, though, to be more code than is really needed here. I wrote up an alternative; please tell me what you think of #364

kentonv:
I went ahead and merged my version in order to include it in the Sandstorm release, but I'm still interested in your feedback!

mark-grimes:
Yes, I possibly over engineered to make myself feel better about the raw `new`, but the extra code doesn't actually changethat problem.  I have a comment but I'll do so on your pull request.

PR #364, opened 2016-09-16
Fix that Thread::detach() prematurely deleted objects the thread might be using.

kentonv:
I haven't actually tested the Win32/MSVC parts but I think capnp still largely doesn't compile on Win32 as-is... so we'll deal with it later I guess.

### DIFF
--- a/c++/src/kj/thread.c++
+++ b/c++/src/kj/thread.c++
@@ -33,8 +33,8 @@ namespace kj {
 
 #if _WIN32
 
-Thread::Thread(Function<void()> func): func(kj::mv(func)) {
-  threadHandle = CreateThread(nullptr, 0, &runThread, this, 0, nullptr);
+Thread::Thread(Function<void()> func): state(new ThreadState { kj::mv(func), nullptr, 2 }) {
+  threadHandle = CreateThread(nullptr, 0, &runThread, state, 0, nullptr);
   KJ_ASSERT(threadHandle != nullptr, "CreateThread failed.");
 }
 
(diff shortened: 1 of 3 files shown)


c++/src/kj/thread.c++
```diff
@@ -33,8 +33,8 @@ namespace kj {
 
 #if _WIN32
 
-Thread::Thread(Function<void()> func): func(kj::mv(func)) {
-  threadHandle = CreateThread(nullptr, 0, &runThread, this, 0, nullptr);
+Thread::Thread(Function<void()> func): state(new ThreadState { kj::mv(func), nullptr, 2 }) {
+  threadHandle = CreateThread(nullptr, 0, &runThread, state, 0, nullptr);
   KJ_ASSERT(threadHandle != nullptr, "CreateThread failed.");
```
mark-grimes: Probably need a

    if( threadHandle==nullptr ) state->unref();

here to do the other thread's cleanup if it doesn't start.  I don't know if CreateThread can throw - if it does this gets more complicated.

c++/src/kj/thread.c++
```diff
@@ -54,37 +54,42 @@ void Thread::detach() {
...
-Thread::Thread(Function<void()> func): func(kj::mv(func)) {
+Thread::Thread(Function<void()> func): state(new ThreadState { kj::mv(func), nullptr, 2 }) {
   static_assert(sizeof(threadId) >= sizeof(pthread_t),
                 "pthread_t is larger than a long long on your platform.  Please port.");
 
   int pthreadResult = pthread_create(reinterpret_cast<pthread_t*>(&threadId),
-                                     nullptr, &runThread, this);
+                                     nullptr, &runThread, state);
   if (pthreadResult != 0) {
     KJ_FAIL_SYSCALL("pthread_create", pthreadResult);
```
mark-grimes: Also think you need

    state->unref();

here or it will leak memory.

c++/src/kj/thread.c++
```diff
@@ -102,18 +107,37 @@ void Thread::detach() {
...
 }
 
 #endif  // _WIN32, else
 
+void Thread::ThreadState::unref() {
+#if _MSC_VER
+  if (_InterlockedDecrement_rel(&refcount)) {
+    _ReadBarrier();
+#else
+  if (__atomic_sub_fetch(&refcount, 1, __ATOMIC_RELEASE) == 0) {
```
mark-grimes: I thought this was gcc specific but I can confirm it works with OSX+clang.  I don't know if the `<stdatomic.h>` versions are an option to widen your supported compilers, but your listed supported compilers are covered.

mark-grimes:
Completely unrelatedly I tried seeing how far you can get compiling with clang on windows, and this indeed breaks the windows build.  The changes are pretty trivial though -

* line 45 change `exception` to `state->exception`
* line 128 clang doesn't like `_InterlockedDecrement_rel` and https://msdn.microsoft.com/hu-hu/f24ya7ct seems to say it's ARM only. Changing to `_InterlockedDecrement` works but also requires changing refcount to long.

I also get bizarre errors with the KJ_LOG on line 136, but I think that's something to do with the mash up of having _MSC_VER defined in a clang build.

I could put in a pull request but I can't currently get the rest of the build to complete, so can't test.

fenollp:
@mark-grimes How about setting up appveyor at the same time?

mark-grimes:
Figured I could just as easily try with MSVC and fix the errors, so I've put in #365 (didn't include my line comments since they're a matter of opinion rather than trivial compile fixes).

@fenollp Do you mean for me personally or for the official builds?  Either way I've not got time to spend on windows, I was just having a play to see if clang would get further than MSVC for non-lite capnp.

PR #365, opened 2016-09-17
Fixes for windows

mark-grimes:
A few trivial fixes to get capnproto mostly compiling on windows.
I'm not a windows expert so I suggest someone who is reviews these changes.  After this change kj and capnproto seem to compile but the full build still fails on something in the test, with

    6>  Compiling Cap'n Proto schema test.capnp
    6>  '""' is not recognized as an internal or external command,
    6>  operable program or batch file.
    6>C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V140\Microsoft.CppCommon.targets(171,5): error MSB6006: "cmd.exe" exited with code 9009.

I suspect it's an issue with cmake not passing on the path of the external capnp executable but I don't know enough about it.

Tested on

* Windows 10 + cmake 3.4.3 + MSVC 2015 (partial build success)
* Ubuntu + gcc 4.8 (`make check` run successfully)

### DIFF
--- a/c++/src/kj/thread.c++
+++ b/c++/src/kj/thread.c++
@@ -24,10 +24,12 @@
 
 #if _WIN32
 #include <windows.h>
+#include "windows-sanity.h"
 #else
 #include <pthread.h>
 #include <signal.h>
 #endif
+
 
 namespace kj {
 
--- a/c++/src/kj/thread.c++
+++ b/c++/src/kj/thread.c++
@@ -30,7 +30,6 @@
 #include <signal.h>
 #endif
 
-
 namespace kj {
 
 #if _WIN32
@@ -128,7 +127,6 @@ void* Thread::runThread(void* ptr) {
 void Thread::ThreadState::unref() {
 #if _MSC_VER
   if (_InterlockedDecrement(&refcount)) {
-    _ReadBarrier();
 #else
   if (__atomic_sub_fetch(&refcount, 1, __ATOMIC_RELEASE) == 0) {
     __atomic_thread_fence(__ATOMIC_ACQUIRE);
--- a/c++/src/kj/thread.c++
+++ b/c++/src/kj/thread.c++
@@ -42,7 +42,7 @@ Thread::~Thread() noexcept(false) {
   if (!detached) {
     KJ_ASSERT(WaitForSingleObject(threadHandle, INFINITE) != WAIT_FAILED);
 
-    KJ_IF_MAYBE(e, exception) {
+    KJ_IF_MAYBE(e, state->exception) {
       kj::throwRecoverableException(kj::mv(*e));
     }
   }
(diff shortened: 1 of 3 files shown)


c++/src/kj/thread.c++
```diff
@@ -125,7 +125,7 @@ void* Thread::runThread(void* ptr) {
 
 void Thread::ThreadState::unref() {
 #if _MSC_VER
-  if (_InterlockedDecrement_rel(&refcount)) {
+  if (_InterlockedDecrement(&refcount)) {
```
kentonv: The compiler will ensure that `refcount` is 32-bit aligned. (In general, the compiler ensures that all variables are aligned to min(variable size, platform bitness).)

I didn't notice that `_rel` is ARM-only -- that's really weird. I would think they'd make the same API available on x86 and have it alias to the full-barrier version. Oh well.

If you have to remove the `_rel` anyway, then I suppose you can remove the `_ReadBarrier()` on the next line, since it's redundant now.

c++/src/kj/thread.h
```diff
@@ -59,7 +59,7 @@ class Thread {
     Function<void()> func;
     kj::Maybe<kj::Exception> exception;
 
-    int refcount;
+    unsigned int refcount;
```
kentonv: Confused here -- the MSDN docs say that `_InterlockedDecrement()` takes a signed value, which is why I made this signed. Were they lying?

BTW KJ defines `uint` as an abbreviation.

c++/src/kj/thread.c++
```diff
@@ -42,7 +44,7 @@ Thread::~Thread() noexcept(false) {
   if (!detached) {
     KJ_ASSERT(WaitForSingleObject(threadHandle, INFINITE) != WAIT_FAILED);
 
-    KJ_IF_MAYBE(e, exception) {
+    KJ_IF_MAYBE(e, state->exception) {
```
kentonv: Hmm, it looks like I totally forgot to make the updates to win32 ~Thread that I did to pthread ~Thread.

c++/src/kj/thread.c++
```diff
@@ -24,11 +24,13 @@
 
 #if _WIN32
 #include <windows.h>
+#include "windows-sanity.h"
 #else
 #include <pthread.h>
 #include <signal.h>
 #endif
 
+
```
kentonv: Extra blank line.

c++/src/kj/thread.h
mark-grimes: It's a signed `long`.  The compiler error I got offered an `unsigned int` overload even though I don't see it listed on the MSDN docs.  I figured the latter was a less intrusive change.

kentonv:
I added 834319e to complete ~Thread() on Windows.

kentonv:
Err, and 562f7ea

mark-grimes:
Oops, I missed that unref.  I also think there should be an extra unref if the thread fails to start but I'll leave that to your judgement.

kentonv:
@mark-grimes Indeed. 8c01fe7

PR #366, opened 2016-09-17
Dockerfile for capnproto

vidbina:
I setup a Dockerfile to allow myself to spawn an environment with all the dependencies available to build, test and run `capnp`. I added some documentation to explain how to use this container.

Upon `docker build .` the current state of the rep is copied into `/tmp/capnproto` and whichever version is there is build and installed. Using `docker run -it $IMAGE -v $PATH_TO_THIS_REP:/tmp/capnproto` one can mount a volume on `/tmp/capnproto` that points to the repository's path on the host such that all changes will be reflected in both environments -- so much easier to develop like this :wink:. 

Shall we setup an official capnproto automated build on Docker hub and Quay?

[](http://microbadger.com/images/vidbina/capnproto "Get your own image badge on microbadger.com")

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,12 @@
+FROM debian:jessie
+RUN \
+  apt-get update && apt-get install -y \
+    autoconf automake libtool g++ git make \
+    && rm -rf /var/lib/apt/lists/*
+COPY . /tmp/capnproto
+RUN cd /tmp/capnproto/c++ && \
+      autoreconf -i && \
+      ./configure && \
+      make -j6 check && \
+      make install
+WORKDIR /tmp/capnproto
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,9 +4,11 @@ RUN \
     autoconf automake libtool g++ git make \
     && rm -rf /var/lib/apt/lists/*
 COPY . /tmp/capnproto
-RUN cd /tmp/capnproto/c++ && \
-      autoreconf -i && \
-      ./configure && \
-      make -j6 check && \
-      make install
+RUN \
+  cd /tmp/capnproto/c++ && \
+    autoreconf -i && \
+    ./configure && \
+    make -j6 check && \
+    make install && \
+    make clean
 WORKDIR /tmp/capnproto
--- a/doc/install.md
+++ b/doc/install.md
@@ -108,7 +108,9 @@ Note: These packages are not maintained by us and are sometimes not up to date w
 If you download directly from Git, you will need to have the GNU autotools --
 [autoconf](http://www.gnu.org/software/autoconf/),
 [automake](http://www.gnu.org/software/automake/), and
-[libtool](http://www.gnu.org/software/libtool/) -- installed.
+[libtool](http://www.gnu.org/software/libtool/) -- installed, unless you choose
+to use build or pull the Docker image which already has the dependencies
+installed.
 
     git clone https://github.com/sandstorm-io/capnproto.git
     cd capnproto/c++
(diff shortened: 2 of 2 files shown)


kentonv:
HI @vidbina,

Thanks for working on this.

Could you describe how you imagine this being used, exactly? Since Cap'n Proto is a combination of a library and command-line tool -- not a microservice -- it's unclear to me how someone would make use of a Docker container that contains only Cap'n Proto. In practice I think people would need to build the libraries for their application's target environment, and would need to integrate the tool into their application's build process. If they are developing inside Docker, they would need to define their own Docker image that includes their other dependencies as well, I would think.

vidbina:
Hi @kentonv, I just use the containers to run the commands so I don't have to worry about installing the tools and all deps on my box. It just simplifies a few things for me.

Generally, I just spin up the container by mounting my work directory `$PWD` into it `docker run -it --rm $IMAGE -v $PWD:/schemas`, cd into my working directory `cd schemas`, and do whatever I need to do there. In fact I have set up a zsh function that just mounts my `$PWD` in `/schemas`, sets `/schemas` as my container's workdir and then performs some `capnp` command in that environment.

```zsh
capnp = function() {
  docker run --rm -w /schemas -v $PWD:/schemas -it vidbina/capnproto:experiment /bin/bash #-c "capnp $*"
}
```
> This helps me jump from computer to computer, because I just need the container and I'm good to go. It's probably a personal misuse of Docker, but I have been packaging a lot of command line tools in containers, since I often find myself having to reinstall certain tools after performing updates.

With the schema `foo.capnp` in my `$PWD` I could basically produce the C++ schema by running `capnp compile -oc++ experiment.capnp`.

It gets tricky when I need plugins for other languages. In that case I basically just take this image and install whatever else needed (e.g.: capnproto2 for golang) to move forward. It's also easier for the team to be able to pull an environment that is guaranteed to work out of the box 🎁 .

kentonv:
Hmm, so you run the `capnp` tool in the container? But if you were generating C++, you would also need to link against the libraries, which requires building them for your target environment, in which case the container doesn't help.

I guess you use this specifically for running the Go code generator? Maybe it makes more sense, then, to submit a Dockerfile to go-capnproto2, that also includes the Go plugin?

Note that if you're installing Cap'n Proto C++ from a release package, the only dependency is a C++ compiler, which most people have. If we did releases more often, then would the Dockerfile no longer be necessary?

vidbina:
My use case may be a rare one considering that my target environment is generally a containerised env as I mostly write services. I only need `capnp` in my dev env and just export the generated code into my projects which I generally outfit with a Dockerfile for runtime, containing just the runtime dependencies.

I didn't just use it for the go code gen, mostly experimented with C++, but then again... the code produced was used in a service that runs inside a container.

I hardly rely on my C++ compiler on my OSX setup because somehow it manages to break with every update of Xcode or macOS. So out of laziness I just packaged this env which is useable wherever I end up having to do my thing.

If there are more releases, automated build could will rebuild the image. Basically running a `docker build .` will always produce an image containing the `HEAD`. You could basically do this with your eyes closed, grab a cup of ☕  and start generating code for schemas after the image is ready to spawn a dev env container.

I will stick to using the container. If you believe the Dockerfile doesn't quite make sense for this project this PR can be closed.

kentonv:
You might find you have another problem with your setup: It's necessary that the version of the `capnp` tool used to generate your code is exactly the same as the version of `libcapnp` than you use at runtime. So if you are generating code using one container that builds `capnp` from HEAD, then you are checking that code into git, and then later on you are building a new container that actually compiles the code against `libcapnp`, it may turn out that a change to Cap'n Proto upstream has broken your project. The breakage would usually be a compile error but it could in some cases be subtle runtime crashes / corruption. We have checks for version mismatches, but they only detect differences between release versions -- if you build from HEAD all bets are off.

That brings up another point, which is that we do not recommend using Cap'n Proto from HEAD. We recommend using a release version. However, this recommendation is currently a bit hypocritical as we use Cap'n Proto from HEAD for Sandstorm and haven't shipped a release in a while. But I hope to do a release in the next few months.

Thanks for trying to improve our tools! Unfortunately due to the issues above I think this Dockerfile would actually create some confusion and might lead people down the wrong path. I think the right way forward starts with us doing an official release. Once the package managers pick up that release, you should be able to do e.g. "brew install capnproto" and get a working capnp tool on your mac, without the need to run it inside a virtual machine.

PR #368, opened 2016-09-26
Add a move constructor to kj::TaskSet

mark-grimes:
TaskSet currently doesn't have a move constructor.  I'm not sure if this is intentional or one has never been required before?  If this is intentional just close the pull request.

I have a particular use case which requires a move constructor, which I can detail if required (basically moving them so that they're deleted on the correct thread).

### DIFF
--- a/c++/src/kj/async.c++
+++ b/c++/src/kj/async.c++
@@ -480,6 +480,9 @@ TaskSet::TaskSet(ErrorHandler& errorHandler)
 
 TaskSet::~TaskSet() noexcept(false) {}
 
+TaskSet::TaskSet(TaskSet&& other)
+    : impl(kj::mv(other.impl)) {}
+
 void TaskSet::add(Promise<void>&& promise) {
   impl->add(kj::mv(promise));
 }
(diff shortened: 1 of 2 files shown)


kentonv:
Hi Mark,

Thanks for checking, but yes, the absence of a move constructor is intentional here.

Adding a move constructor only coincidentally works for TaskSet because its implementation uses the "PIMPL" idiom: its only member is a pointer to a different object that actually implements the logic. If you take a look at TaskSetImpl, you'll find that allowing it to be moved would actually lead to dangling pointers, because each TaskSetImpl::Task contains a reference back to the TaskSetImpl.

The only reason we use PIMPL here is because TaskSetImpl contains an `std::map`, which I don't want to include from a KJ header. Some day it might change to use some other data structure, and then we might want to avoid the extract indirection by merging TaskSetImpl's code back into the TaskSet class.

But if TaskSet has a move constructor, we can never do this.

More generally, the rule I like to go by is that move constructors should only be given to:
- Pure data structures -- that is, anything which could logically have a copy constructor, too.
- Types which are explicitly designed for the purpose of being a handle to a separate resource, such as Own<T> and Promise<T>.

I recommend that you allocate your TaskSet on the heap to get an `Own<TaskSet>`, to which you can then apply move semantics.

PR #370, opened 2016-09-30
Add initial AppVeyor support

harrishancock:
This commit adds an AppVeyor configuration file (appveyor.yml) and a build script (appveyor-build.bat). These files could just as well live in the c++/ directory, although that would add a step in configuring AppVeyor to build the project.

AppVeyor will autoconfigure this project from the configuration stored in appveyor.yml, so setting up an AppVeyor project for Cap'n Proto should be as simple as:
1. Log into ci.appveyor.com. You can use a GitHub account for this.
2. Select "New Project."
3. Select capnproto from the list of GitHub projects.

At this point, you should have a functioning AppVeyor project, and can select "New Build" to build the latest commit on the default branch.

A few things to note:
- To test this commit before it's merged into mainline, you'll need to fork harrishancock/capnproto and change the AppVeyor project's default branch to 'appveyor' in the 'Settings -> General' panel.
- Every new commit to the repository will trigger a new build. This can be configured more granularly in appveyor.yml.
- You can configure AppVeyor to notify you of build statuses under the 'Settings -> Notification' panel, or in appveyor.yml.
- Test failures are reported as build failures. Currently the tests fail on Windows.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,75 @@
+# Cap'n Proto AppVeyor configuration
+#
+# See https://www.appveyor.com/docs/appveyor-yml/ for configuration options.
+#
+# This script configures AppVeyor to:
+#   - Download and unzip MinGW-w64 4.8.5 for x86_64, the minimum gcc version Cap'n Proto advertises
+#     support for.
+#   - Use CMake to ...
+#       build Cap'n Proto with MinGW.
+#       build Cap'n Proto with VS2015 and the MinGW-built capnp tools.
+#       build the Cap'n Proto samples with VS2015 and the MinGW-built capnp tools.
+#   - Archive both Cap'n Proto builds in capnproto-c++-{mingw,vs2015}.zip artifacts.
+
+version: "{build}"
+
+image: Visual Studio 2017
+# AppVeyor build worker image (VM template).
+
+shallow_clone: true
+# Fetch repository as zip archive.
+
+cache:
+  - x86_64-4.8.5-release-win32-seh-rt_v4-rev0.7z
+
+environment:
+  MINGW_DIR: mingw64
+  MINGW_URL: https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win64/Personal%20Builds/mingw-builds/4.8.5/threads-win32/seh/x86_64-4.8.5-release-win32-seh-rt_v4-rev0.7z/download
+  MINGW_ARCHIVE: x86_64-4.8.5-release-win32-seh-rt_v4-rev0.7z
+  BUILD_TYPE: debug
+
+install:
+  - if not exist "%MINGW_ARCHIVE%" appveyor DownloadFile "%MINGW_URL%" -FileName "%MINGW_ARCHIVE%"
+  - 7z x -y "%MINGW_ARCHIVE%" > nul
+  - ps: Get-Command sh.exe -All | Remove-Item
+  # CMake refuses to generate MinGW Makefiles if sh.exe is in the PATH
+
+before_build:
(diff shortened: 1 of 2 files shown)


kentonv:
Thanks, Harris!

I'm hoping to get a chance to set this up and also do some other Windows-related work in the next week or two.

paulproteus:
I want to chime in and say that I'm super excited to see more build automation, especially on Windows where we've been unable to reliably test capnproto ourselves!

harrishancock:
Kenton, awesome -- let me know if you've any questions. :)

And thanks, Asheesh!

harrishancock:
@kentonv I actually need to update this branch to build non-lite mode now, so you might hold off learning you an AppVeyor until I update it. Should be super simple, expect it tomorrow.

kentonv:
OK, cool. Let me know.

harrishancock:
Okay, updated.

To test this before merging into master, you may want to create an AppVeyor project for capnproto first, then fetch this branch into the main repo. As soon as this appveyor branch is uploaded to the main repo, you should see AppVeyor begin building it.

My first iteration of this branch used a separate build script because the Windows capnproto bootstrapping process was rather involved. Since that process has been simplified, I just moved all the scripting into appveyor.yml. At the moment, it is hardcoded to build release configurations, rather than debug.

Do we want to generate releases with AppVeyor? Currently the script generates two artifacts: `capnproto-c++-mingw.zip` and `capnproto-c++-msvc.zip`, which are just archives of each respective install tree. If we wanted to include version numbers in those filenames, we'd either have to hard-code the version numbers in the .yml file (maybe bump them with `sed` in `release.sh`?) or programmatically generate the artifact names, e.g. in a PowerShell `after_build` script.

kentonv:
Hmm I'd like to merge this but apparently there are conflicts.

> At the moment, it is hardcoded to build release configurations, rather than debug.

For CI purposes I think debug is preferred. A lot of checks are skipped in release mode.

> Do we want to generate releases with AppVeyor?

Dunno. Do Windows developers want precompiled libraries? This seems complicated since MSVC has a lot of compiler settings that affect the ABI, so it seems like people need to ensure their libraries are built with the same settings as their app. It seems, then, that people had better build their own libs?

My automated release scripts already know how to build and package the capnp binaries with MInGW, which works on Linux, which I'm happy with. So unless we think that precompiled libs would be valuable, I don't see a strong reason to use AppVeyor for releases.

harrishancock:
Will update.

Cap'n Proto is very easy to build, so I'm happy not worrying about precompiled libs until there's demand.

harrishancock:
Updated!

- Rebased onto master, resolving conflicts (a minor disagreement over whitespace).
- Builds in debug mode, rather than release.
- Removed artifact generation so we don't waste AppVeyor's disk space.
- Broke the long `cmake` invocations which were over 100 characters into multiple lines using YAML block quotes with a newline chomper (`>-`).
- Builds the samples now, too -- @kentonv merged that PR as I was finishing this one up.

I also switched it to use AppVeyor's new Visual Studio 2017 build worker image, but it still won't build yet (blocked by the `NullableValue` ICE I mentioned in #414). VS2015 is available on this build worker image, too, so this is essentially a no-op.

harrishancock:
Blech, I just noticed that commit dee4c91 is braindead now (that was the whitespace conflict, and now I understand why it happened). Will remove in just a second.

harrishancock:
Okay, really done this time!

kentonv:
Woohoo!

harrishancock:
:D Please feel free to add my email address to the project's notifications on AppVeyor.

kentonv:
@harrishancock Done, I think! First build in progress now.

PR #378, opened 2016-10-07
Fix RealmGateway to avoid double-transforms on loopback.

kentonv:
Previously, in the presence of promise capabilities, a save() request could cross over the same gateway twice (or more!). Some gateways (e.g. Sandstorm's) implement one-way transformations, so this could cause the request to fail when it shouldn't.


kentonv:
Pretty confident in this due to the unit tests (yay!) so I'm going to merge for the Sandstorm release, but interested in your take when you get a chance.

PR #384, opened 2016-10-24
KJ Filesystem API

kentonv:
I've long wanted a filesystem API that:
- Is resistant to path injection bugs.
- Uses a representation of paths that is easy to manipulate (flat strings with `/` as separators are actually incredibly tedious!).
- Canonicalizes ".." logically (by removing the previous path component) rather than physically (by climbing the directory tree), especially when dealing with the current working directory.
- Is RAII.
- Does not assume a singleton filesystem (= implements abstract interfaces so that you can provide a different implementation when desired).
- Throws exceptions on error, but lets you opt-in to programatically handling the kinds of errors that one commonly wants to handle (namely, doesn't-exist and already-exists).
- Directly supports the write-to-temp-then-atomically-replace pattern.
- Directly supports the common operations "recursively delete directory" and "automatically create parent directories".
- Directly supports a copy operation, with smart semantics around sparse files.
- Makes mmap easy.
- Automatically opens all file descriptors close-on-exec.
- Utilizes Linux features like renameat2(), FICLONE (copy-on-write files), O_TMPFILE, etc. when available, but falls back to generic POSIX when not.
- Much more...

So this is my attempt.

### DIFF
--- a/c++/src/kj/filesystem-disk-generic-test.c++
+++ b/c++/src/kj/filesystem-disk-generic-test.c++
@@ -1,0 +1,57 @@
+// Copyright (c) 2016 Sandstorm Development Group, Inc. and contributors
+// Licensed under the MIT License:
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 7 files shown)


kentonv:
(I just merged everything except the filesystem API itself. Github seems very confused about this...)

kentonv:
Gonna merge this because I'm using it in a project, although currently only Ekam will build it. I guess we need a Win32 version before we can make it part of the main builds.

PR #386, opened 2016-11-07
Support RPC on Windows (MinGW)

kentonv:
With these changes, all RPC tests work on Windows when compiling with MinGW.

The meat of this change is implementing EventPort and async-io in terms of Winsock and I/O completion ports.

Things still to do:
* Update cmake files.
* Make it compile in MSVC. (May or may not be a lot of work.)
* Support waiting on signalable HANDLEs (a la WaitForMultipleObjects()). This looks like it will require a thread pool since there's no way to wait on IOCP and objects in the same thread.
* The interface change to LowLevelAsyncIoProvider::wrapConnectingSocketFd() should perhaps be used on Unix too, so that they can be consistent. Maybe this method should change to `wrapSocketAndConnect()`.
* debug.c++ has some TODO(now)s.
* It looks like there is one spurious test failure when running on actual Windows rather than Wine. Just noticed this now and haven't had time to investigate.

@harrishancock Are you interested in getting this working in MSVC? If not I'll probably look into it later this month.

### DIFF
--- a/c++/src/kj/exception.c++
+++ b/c++/src/kj/exception.c++
@@ -38,6 +38,13 @@
 #include <execinfo.h>
 #endif
 
+#if _WIN32
+#define WIN32_LEAN_AND_MEAN
+#include <windows.h>
+#include "windows-sanity.h"
+#include <dbghelp.h>
+#endif
+
 #if (__linux__ || __APPLE__) && defined(KJ_DEBUG)
 #include <stdio.h>
 #include <pthread.h>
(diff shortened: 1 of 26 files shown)


harrishancock:
Awesome!! Nice job, Kenton. :)

I'll try making as much of this compile in MSVC as I can tonight, but don't hold the merge up for me.

I suspect it might not be feasible for another month or so, because it is currently impossible to port the dynamic schema API to MSVC. I realize the async I/O implementation ought to be orthogonal to the dynamic API, but last time I tried porting the event loop code over, I ended up having to port the dynamic API first -- I can't remember why, though. :/ Presumably tonight I will either remember, or disprove myself.

By the way, the reason the dynamic API isn't portable right now is this: There is a regression in the latest version of VS2015 (Update 3) in which MSVC ICEs when trying to compile the `anyOfChars` parser combinator. The latest version of VS "15" (Preview 5) fixes this regression, but the latest release candidate of CMake is currently incompatible with VS "15"'s new directory structure. I expect CMake to support it soon, hopefully by the 3.7 release, at which point it should be feasible to complete the port to MSVC, finally.

kentonv:
Ouch. Did VS2015 Update 2 work? Will there be another update to the VS2015 line that fixes the regression? People didn't like it two years ago when capnp required a prelease VS, so it'll be sad if we have to tell them that it once again requires a prerelease.

Maybe we can work around the anyOfChars problem somehow?

Also, I think the parser combinators are only used by the compiler, not by the dynamic API. So maybe we could stick with shipping a prebuild compiler binary and exclude just the parser combinators from the runtime library?

harrishancock:
Update 2 sort of worked, but only with this workaround: https://github.com/harrishancock/capnproto/commit/864c930736b428d3f95890c75b444e582596c41d

(Obviously that macro could be named `anyOfChars`, which would be an improvement in readability, though it could arguably be more of a global namespace pollutant.)

Update 3 ICEs even *with* that workaround. I tried finding a different way for a few hours in September, but gave up. I had already spent a long time trying to find a better way than that `#define` for Update 2. I certainly hope there will be an update to VS2015 that resolves the regression, but lack the required tea leaves to know. :(

I think the only workaround that would work with Update 3 would be excluding the parser combinators from the release, as you suggested. I looked at the code and think you're right: the dynamic API shouldn't depend on the parser combinators. I probably conflated the dynamic API with schema parsing.

By the way, it looks like it is still possible to install Update 2: https://msdn.microsoft.com/en-us/library/mt653628.aspx

I'll try that tonight.

kentonv:
Great, I think it makes sense then to exclude the parser combinators (and hence the compiler and libcapnpc (schema parser library)) in VS2015. That's OK because most apps don't use those at runtime anyway. We can ship a MinGW-built compiler binary for people to use. For people using the confusingly-named VS "15", we can enable that part of the build.

c++/src/kj/exception.c++
```diff
@@ -38,6 +38,13 @@
 #include <execinfo.h>
 #endif
 
+#if _WIN32
+#define WIN32_LEAN_AND_MEAN
+#include <windows.h>
+#include "windows-sanity.h"
```
harrishancock: I recommend:

```
#define WIN32_LEAN_AND_MEAN
#define NOGDI     // Inhibit ERROR define
#define NOMINMAX  // Inhibit min/max defines
#include <windows.h>
```

Then this file will compile in MSVC. The GDI in NOGDI stands for Graphics Device Interface -- presumably cruft we can safely excise in this .c++ file.

harrishancock:
This branch compiles in MSVC under lite mode with the `#define NOMINMAX` suggestion I made in exception.c++. Haven't tried compiling async code yet.

c++/src/kj/exception.c++
```diff
@@ -57,19 +64,142 @@ StringPtr KJ_STRINGIFY(LogSeverity severity) {
...
+      DWORD MachineType,HANDLE hProcess,HANDLE hThread,
+      LPSTACKFRAME64 StackFrame,PVOID ContextRecord,
+      PREAD_PROCESS_MEMORY_ROUTINE64 ReadMemoryRoutine,
+      PFUNCTION_TABLE_ACCESS_ROUTINE64 FunctionTableAccessRoutine,
+      PGET_MODULE_BASE_ROUTINE64 GetModuleBaseRoutine,
+      PTRANSLATE_ADDRESS_ROUTINE64 TranslateAddress);
+  PVOID WINAPI (*symFunctionTableAccess64)(HANDLE hProcess,DWORD64 AddrBase);
+  DWORD64 WINAPI (*symGetModuleBase64)(HANDLE hProcess,DWORD64 qwAddr);
+  WINBOOL WINAPI (*symGetLineFromAddr64)(
+      HANDLE hProcess,DWORD64 qwAddr,PDWORD pdwDisplacement,PIMAGEHLP_LINE64 Line64);
```
harrishancock: Earlier when I said this file compiled in MSVC, I spoke too soon -- I was targeting 32-bit.

To actually get this to compile in MSVC, I had to change WINBOOL to BOOL and change all occurrences of `WINAPI (*foo)` to `(WINAPI *foo)`. Still untested, though.

harrishancock:
@kentonv, I finished a rough draft port of this branch to VS2015U3: https://github.com/harrishancock/capnproto/commits/async-win32-vs2015

There are a handful of commits that I can split off into a useful PR right away. The bulk of it I won't submit until this one (#386) is merged, so I can rebase onto master. There are some other cleanup issues that I need to attend to, as well.

For now, I made it so CMake won't build the compiler, or any parsing code, if EXTERNAL_CAPNP is defined. I think this is now inconsistent with the autotools build script's external-capnp semantics, though. We could add a third variable to control the building of the tools, CAPNP_BUILD_TOOLS or something, or we could consider redefining CAPNP_LITE to mean "build everything but the tools."

I have not yet:
- compiled against VS2015 Update 2, which should get us the compiler tools built, which could potentially get us MSVC-compiled pycapnp
- cleaned up any of the 428 warnings
- debugged any of the 10 failing tests
- ported the pointer-to-member-function helper code

Test results are here: https://gist.github.com/harrishancock/54accdd8beae10b93b5fbd4c652b9e2a

kentonv:
Went ahead and merged this -- I guess there was nothing to wait for.

@harrishancock Please send a pull request for the rest of the MSVC changes whenever you're ready.

harrishancock:
Great, this is exciting! I'll have a PR ready soon.

PR #387, opened 2016-11-15
A grab bag of small changes to support the MSVC port

harrishancock:
I discovered these mostly trivial changes while working on the MSVC port of #386 and factored them out to reduce the footprint of that eventual PR.

Note that this touches code which the build system currently doesn't allow MSVC to compile, but hopefully that will change.


c++/src/capnp/compat/json.h
```diff
@@ -313,6 +313,28 @@ class JsonCodec::Handler<T, Style::POINTER>: private JsonCodec::HandlerBase {
   friend class JsonCodec;
 };
 
+template <>
+class JsonCodec::Handler<DynamicValue>: private JsonCodec::HandlerBase {
```
kentonv: Instead of specializing this, maybe we should actually implement DynamicValue::as<DynamicValue>()? It would just `return *this;` of course.

harrishancock: That does sound better -- I'll give it a shot tonight or tomorrow and update the branch.

harrishancock: Done, see below. Would you like me to remove this commit, and its reversion, from the branch history?

kentonv:
Would you mind rebasing-out the commit/revert? Otherwise looks great to me.

harrishancock:
Done!

kentonv:
Thanks!

PR #396, opened 2016-12-16
Let MSVC build without lite mode

harrishancock:
This branch allows Visual Studio 2015 Update 3 to build everything except parser-dependent code (schema loading, compiler, compiler plugins) in non-lite mode. It is based on the current master branch. On Linux, all tests pass. On Windows, the status is still a bit WIP: 10 of the newly-enabled tests fail, and it builds with many warnings.

@kentonv, this is a rather large PR -- I'm happy to break it up into two or three smaller ones, if you'd like. I won't have time to work on the warnings/test failures and PtmfHelper port until next month, unfortunately, though I'm interested in doing so. I of course wouldn't be offended if you wanted to get some of it done to get a release out the door, too. :)

@nathanhourt, you might be interested in testing this with your codebase, though keep in mind you'll need a MinGW-built capnp.exe.

Notes:
- CMake now only builds parser-dependent code if CAPNP_BUILD_TOOLS is set to ON, which I made the default. We could name this option differently (CAPNP_NO_BUILD_TOOLS?), or we could default it to OFF, or what-have-you. I have no strong preference either way.
- Pointer-to-member-function address extraction is not yet ported, so stack traces will be unhelpful for the time being.
- Atomic operations are a dilemma. For now, I #defined the GNU builtins that RawSchema uses for its initializers to be simple non-atomic loads and stores, per a conversation I had with Kenton a while back. For the atomic sentWake in async-io-win32.{h,c++}, I'd prefer to use std::atomic to decrease the amount of special-casing between MSVC and MinGW.
- One of the overloads of AsyncInputStream::read() was undefined in async-io-win32.c++, which the linker complained about. I copied and pasted the implementation from async-io.c++.
- I had to work around the usual host of compiler bugs. Most of the workarounds turned out okay, but there are a couple (non-const copy capture in lambdas and hard-coding dependency/scope counts) that I wish I had better options for.


nathanhourt:
Need a mingw-built capnp.exe... is that used for anything other than codegen? Could I just do the codegen on another OS and drop the generated files into place manually for Windows builds?

harrishancock:
Yes I believe that would work fine.

kentonv:
Is it redundant to have both `EXTERNAL_CAPNP` and `CAPNP_BUILD_TOOLS`?

If not, it seems to me that you need to add a check that `EXTERNAL_CAPNP` is enabled if `CAPNP_BUILT_TOOLS` is disabled.

harrishancock:
My only reason for making them separate options is that I worried that not having a separate option controlling build tool creation would lead to confusion as to what exactly the semantics of "external capnp" meant in autoconf versus CMake builds. In autoconf configuration, it does not necessarily imply that the capnp tools cannot be built, so I tried to mirror that meaning in CMake configuration.

On the other hand, having a separate option might end up just being line noise to the user -- most will either want `EXTERNAL_CAPNP` on and `CAPNP_BUILD_TOOLS` off, or vice versa. Thoughts?

kentonv:
Ah, right. Another use for EXTERNAL_CAPNP is when cross-compiling. You in theory may actually want to build the capnp executable in this case, but need to use a host-compiled capnp during the build process.

What if instead of introducing a new option we just automatically don't build capnp.exe when using MSVC? EXERNAL_CAPNP will still need to be used to specify what capnp executable to use, but the CAPNP_BUILD_TOOLS option seems to me like forcing the user to make a decision where we already know the right answer...

harrishancock:
@nathanhourt I didn't see your edit. Automating codegen in Docker would probably work fine, too. Building a capnp.exe with MinGW probably wouldn't be any more work than that, though -- but I'd say whatever fits into your existing workflow should be good. :) Btw, you'll need to pass `-DBUILD_TESTING=OFF` to the CMake command line or else it will try to execute capnp.exe.

harrishancock:
> the CAPNP_BUILD_TOOLS option seems to me like forcing the user to make a decision where we already know the right answer

I agree. Hmm. How about we keep the option, mark it as advanced (i.e., hide it in the CMake GUI), default it to OFF if `MSVC` is true, and default it to ON otherwise? Then it would just do the right thing under normal usage, but we can port the parser-dependent code for VS2015U2/VS2017 with the flip of a switch.

kentonv:
Having raised the question I wanted to raise, I'm happy with whatever answer you think is best (including leaving as-is, if you prefer).

harrishancock:
Hehe, thanks for raising it, it made me think a little harder about it. I like the solution I suggested just now best, and will rebase-in a couple adjustments to the relevant commits in a few minutes here.

harrishancock:
Done. Changed commits are 79dc0f1 and 8845da1.

kentonv:
Great! Still need a check that errors out if EXTERNAL_CAPNP is missing in this case, right?

harrishancock:
Oops! You're right. Will update tomorrow morning.

harrishancock:
Updated! Changes are in 9d7680e.

Adding the option made the error message confusing to write, since it is unlikely but conceivable that someone using gcc/clang could set the option off, so CMake needs to error out in that case as well. Nevertheless, I still think having it as an option will be worth it.

Also, I noticed that my changes in this branch broke the Linux CMake build (I only tested the Linux autoconf build while developing). I'll investigate and fix it.

c++/src/kj/async-io-win32.c++
```diff
@@ -1114,6 +1114,16 @@ private:
 Promise<void> AsyncInputStream::read(void* buffer, size_t bytes) {
   return read(buffer, bytes, bytes).then([](size_t) {});
 }
+Promise<size_t> AsyncInputStream::read(void* buffer, size_t minBytes, size_t maxBytes) {
```
kentonv: I'm confused -- how did this manage not to be here before and how did the code compile and run for me without it? All the code around this function was clearly copy/pasted from the corresponding surrounding code in async-io.c++. Did I copy the code and then remove this function definition? So strange.

(And yeah, there's a fair amount of code duplication between async-io.c++ and async-io-win32.c++... I will maybe clean it up eventually.)

c++/src/kj/common.h
```diff
@@ -373,6 +373,23 @@ struct DisallowConstCopy {
   DisallowConstCopy& operator=(DisallowConstCopy&&) = default;
 };
 
+#if _MSC_VER
+
+#define MSVC_CAPTURE(obj) obj=::kj::cp(obj)
```
kentonv: I would call this `KJ_CPCAP()`:
* KJ macros should start with `KJ_` to follow macro namespacing rules.
* There is actually a macro called `KJ_MVCAP()` in the Sandstorm codebase which has a similar purpose but for capture-by-move.

More generally, the need for this macro is unfortunate. I've actually been thinking of giving up on the non-const-copy thing. The purpose was supposed to be to enforce "transitive const" semantics, but in practice it causes quite a lot of problems, including tickling a lot of compiler bugs (there's actually an ABI incompatibility between Clang and GCC triggered by it). Perhaps transitive-const simply is not practical to enforce in C++. Perhaps it's time to do that.

I guess this commit should stay in for now but I'll revert it when I'm ready to roll back transitive-const enforcement.

c++/src/capnp/persistent.capnp.c++
```diff
@@ -69,8 +69,8 @@ static const ::capnp::_::RawSchema* const d_c8cb212fcd9f5691[] = {
 };
 static const uint16_t m_c8cb212fcd9f5691[] = {0};
 const ::capnp::_::RawBrandedSchema::Dependency bd_c8cb212fcd9f5691[] = {
-  { 33554432,  ::capnp::Persistent< ::capnp::AnyPointer,  ::capnp::AnyPointer>::SaveParams::_capnpPrivate::brand },
-  { 50331648,  ::capnp::Persistent< ::capnp::AnyPointer,  ::capnp::AnyPointer>::SaveResults::_capnpPrivate::brand },
+  { 33554432,  ::capnp::Persistent< ::capnp::AnyPointer,  ::capnp::AnyPointer>::SaveParams::_capnpPrivate::brand() },
```
kentonv: So do these initializers run code at startup when compiled with MSVC?

If so that's pretty unfortunate -- dynamic startup-time initializers are something that the Chrome team has in the past worked vigorously to eliminate, to improve startup time. The main problem is not the execution time of the code itself but the fact that the code must be paged in from disk in order to execute.

But maybe MSVC is actually able to optimize these away? Can you check somehow if any dynamic initialization code is being generated?

If dynamic initializers are being generated, maybe we could avoid them by manual inlining? That is, instead of a call to `brand()`, actually refer to `&schema->defaultBrand` directly here?

c++/src/capnp/compiler/capnpc-c++.c++
```diff
@@ -2188,7 +2188,12 @@ private:
   InterfaceText makeInterfaceText(kj::StringPtr scope, kj::StringPtr name, InterfaceSchema schema,
                                   kj::Array<kj::StringTree> nestedTypeDecls,
                                   const TemplateContext& templateContext) {
-    auto fullName = kj::str(scope, name, templateContext.args());
+    auto templateArgs = templateContext.args();
+    auto fullName = kj::str(scope, name, templateArgs);
+    auto fullDisambiguatedName = kj::str(
+        scope.size() ? "typename " : "", scope,
+        scope.size() && templateArgs.size() ? "template " : "", name,
```
kentonv: Hmm, is there a problem here if one of the parent scopes (encoded in `scope`) is a template? It looks like you're only adding `template` to the last component.

Actually, I wonder if this stuff is even necessary. Looking at the code for structs, the `Reads` and `Builds` typedefs there are defined using the unqualified parent type, i.e.:

    struct MyType {
      class Reader {
        typedef MyType Reads;

I'm not sure why the `Calls`, `Reads`, and `Serves` typedefs of interfaces are using fully-qualified names here -- maybe they don't have to?

c++/src/capnp/compiler/capnpc-c++.c++
```diff
@@ -2122,10 +2122,22 @@ private:
...
+    //
+    // The easiest (and only) workaround I found is to use C++14's return type deduction here, thus
+    // the `#if _MSC_VER` hackery in the return type declarations below. We're depending on the fact
+    // that that this function has an inline implementation for the deduction to work.
+
     auto requestMethodImpl = kj::strTree(
         templateContext.allDecls(),
         implicitParamsTemplateDecl,
-        "::capnp::Request<", paramType, ", ", resultType, ">\n",
+        templateContext.isGeneric() ? "#if _MSC_VER\nauto\n#else\n" : "",
```
kentonv: Instead of littering the generated code with these verbose if/elses, what about defining a macro like `CAPNP_AUTO_IF_MSVC(type)` which evaluates to `auto` on MSVC or `type` elsewhere?

c++/src/capnp/raw-schema.h
```diff
@@ -28,6 +28,26 @@
 
 #include "common.h"  // for uint and friends
 
+#if _MSC_VER
+
+// TODO(msvc):  RawSchema and RawBrandedSchema use some atomic operations in their implementations.
+// The only non-deprecated atomic operations that MSVC supports is std::atomic. We could rewrite our
```
harrishancock: I think you mean C11 atomics? And no, they never bothered implementing them.

c++/src/capnp/raw-schema.h
```diff
@@ -28,6 +28,26 @@
...
+// atomic operations to use std::atomic, but this has the awkward consequence of making RawSchema
+// and RawBrandedSchema non-POD, non-trivially-copyable, non-standard-layout, etc. It is likely safe
+// to ignore this consequence -- i.e., declare the atomic variables as Initializer* and cast them to
+// std::atomic<Initializer*> as necessary -- but clearly non-portable beyond MSVC, which defeats the
+// purpose of using std::atomic. Another, perhaps better, option would be to use MSVC's old
+// deprecated operations, which are unlikely to disappear any time soon.
+//
+// Since Cap'n Proto code is required to be single-threaded anyway for the foreseeable future, we
```
harrishancock: Thanks for explaining this, the need is clear to me now.

My hesitance to use the deprecated API is due in large part to lack of familiarity: I never touched an atomic operation until a post-C++11 world, and I don't want to get it wrong. However, given our options, they seem least bad, so I'll give it a shot. :) I'll take a look at `kj::Once` first, though, as it'd be nice to factor the platform-dependent code out somewhere else.

c++/src/kj/async-io-win32.c++
harrishancock: This spooked me, too, because I was also able to get your original async-win32 branch compiling by itself in MSVC, but when I based the work on master, suddenly this cropped up! My best guess is that 559044bf tickled the dead code path elimination routines in the compiler to suddenly reveal the lack of `read`. That's just a guess, though.

c++/src/kj/common.h
harrishancock: Thanks, that's a better name.

I think this is the kind of bug that we could get Microsoft to fix, but it would nevertheless mean waiting a long while before we could remove the macro. I really like the non-const-copy strategy conceptually, so I'd be sad to see it go, but new code is ultimately going to trip over this bug repeatedly.

My biggest concern would be whether or not it affects user code. I.e., how many support requests are going to be generated on the mailing list by confused users trying to port their code to Windows? If that number is low, then I'd say let's not worry about it. If it's high, then sacrificing transitive const safety might be the lesser of two evils.

c++/src/capnp/persistent.capnp.c++
harrishancock: That's a great question. I would expect that the compiler would be smart enough to optimize the startup-time calls away, since they're inline and just return a value, but I will find a way to verify that assumption.

> If dynamic initializers are being generated, maybe we could avoid them by manual inlining? That is, instead of a call to brand(), actually refer to &schema->defaultBrand directly here?

Unfortunately, no. This was the strategy I tried initially, and it works in simple cases, but once brand pointers start depending on other brand pointers through `ChooseBrand` indirection, eventually someone dereferences a null pointer on startup, due to the "static initialization order fiasco."

I wrote a ham-handed explanation of the problem here: https://groups.google.com/d/msg/capnproto/DnlAgoiJGBI/YGVWhfOSBQAJ

Ultimately, hiding the pointers behind a function was the only workable solution I found, but if there's a way to remove the need for `ChooseBrand`, simple direct initialization might be feasible.

harrishancock: Wait, I think I misunderstood your suggestion. You meant have the code generator calculate what the appropriate brand pointer is and generate that directly in the initializer lists in the generated .c++ files? That might be doable. I'll have to think on it!

c++/src/capnp/compiler/capnpc-c++.c++
harrishancock: The lack of more `template` sprinkles might be an oversight: it looks like all the templates this code affects in test.capnp.h are all only nested one-deep.

I'll try using the unqualified parent type name first and remove this stuff. If that doesn't work for some reason, I'll add a test case with multiple nested generics to elicit the problem you described, then fix it.

harrishancock:
Thanks for the copious feedback! I'll prepare some patches addressing everything you raised after Monday.

Would you like updates in the form of new commits or rebased force-pushes? I fear that rebases might obliterate the code comments you wrote, and we can always clean up the history afterward...

harrishancock:
@kentonv, I added commits that address every issue you raised except for the dynamic initializer calls at startup.

The atomic operations implementation is a little messy. I think the next best thing we could do is to refactor atomic ops into a `kj::Atomic` type or `kj::atomicXxx` functions, but I figured I'd run this draft by you first.

I'm holding off rebasing these commits for fear of losing the comment history, but when the time comes, I'll clean it all up.

c++/src/capnp/persistent.capnp.c++
harrishancock: I looked at the disassembly, and dynamic calls *are* generated, unfortunately. Manual inlining will probably be our best option, though I haven't looked into it yet.

c++/src/kj/common.h
kentonv: It's nice to hear that someone understands and likes my unconventional style choices. :)

However, we've had many other problems with the non-const copy approach. The Clang/GCC ABI incompatibility is pretty severe. So it should probably go. C++ just isn't going to be Rust any time soon, I guess.

kentonv:
Sorry for the long wait -- holidays.

Changes look good.

What should we do about the dynamic init? It's important to me that dynamic initializers are not generated on Clang or GCC, but I could accept having them in MSVC temporarily until they fix their bugs, if that's the only option.

harrishancock:
Truthfully, the dynamic init issue slipped my mind over the holidays. Give me the weekend to work on it. If I run into trouble, I'm fine with it merging as-is, but would prefer not to let it slip through the cracks if I can help it.

harrishancock:
@kentonv Thanks for your patience. After some flailing about, I have to admit defeat.

To manually inline the brand pointers, we'd need to reproduce the logic in `ChooseBrand` at this point: https://github.com/harrishancock/capnproto/blob/async-win32-msvc/c%2B%2B/src/capnp/compiler/capnpc-c%2B%2B.c%2B%2B#L811-L813

`ChooseBrand` checks the type of every template parameter of every enclosing scope of the dependency, so we'd need to either reconstruct the template arguments in the dependency's chain of enclosing scopes, or pass the dependency's chain of enclosing scopes through the various `makeBrandDepInitializer` overloads.

The first option (reconstruction) doesn't seem to be feasible, since Node `scopeId`s are allowed to be 0. The second option is probably doable, but looks painful. It would further complicate already complicated code, and I'm not at all confident that I would get it correct for all cases.

Do you see a clean way of implementing the manual inlining? If not, I think I'd prefer to defer the issue until Microsoft has a fix. :/

kentonv:
@harrishancock If you can't find a way I don't expect I'd be able to do any better.

The important thing to me, then, is that we still avoid dynamic initialization on Clang and GCC. Do we need to add some #ifdefs so that they can keep their code the old way?

harrishancock:
I'll examine the gcc and clang assembly output -- I suspect that it already evaluates at compile-time as-is. If not, or simply for peace of mind, maybe we could do something like:

```c++
template <const RawBrandedSchema* ptr>
struct EvalConstexprBrand {
  static constexpr RawBrandedSchema const* value = ptr;
};

#if _MSC_VER
#define CAPNP_EVAL_CONSTEXPR_BRAND(...) __VA_ARGS__
#else
#define CAPNP_EVAL_CONSTEXPR_BRAND(...) EvalConstexprBrand<__VA_ARGS__>::value
#endif
```

Then wrap the worrisome calls in `CAPNP_EVAL_CONSTEXPR_BRAND(SomeType::_capnpPrivate::brand())`.

We could also `#ifdef` the existing raw constexpr pointer code back in, but I think that would have more cognitive overhead for the reader of the code.

kentonv:
I wonder if a more general `KJ_FORCE_CONSTEXPR(x)` which evaluates to `x` but guarantees compile-time evaluation would be useful. I've noticed that Clang has regressed recently in terms of what it evaluates at compile time, sometimes moving what ought to be constexpr evaluations to runtime for no good reason.

Up to you.

c++/src/kj/async-io-win32.c++
kentonv: This change is no longer needed at master -- I moved the default implementations of virtual functions into a common file for Unix and Windows.

harrishancock: Thanks for the heads-up!

harrishancock:
What if we just marked the entire dependency array as `KJ_CONSTEXPR(const)`?

```c++
KJ_CONSTEXPR(const) ::capnp::_::RawBrandedSchema::Dependency bd_c8cb212fcd9f5691[] = {
  { 33554432,  ::capnp::Persistent< ::capnp::AnyPointer,  ::capnp::AnyPointer>::SaveParams::_capnpPrivate::brand() },
  { 50331648,  ::capnp::Persistent< ::capnp::AnyPointer,  ::capnp::AnyPointer>::SaveResults::_capnpPrivate::brand() },
};
```

That ought to provide us the desired guarantee on non-MSVC compilers with almost no code disruption.

Also, FWIW, I have verified that gcc 5.4 behaves correctly: no `.text` section is generated at all using `constexpr` brand functions.

harrishancock:
Cleaned up the history a bit. I still have to fold the `KJ_CONSTEXPR(const)` change in and do some bootstrap regenerations. Will rebase onto master for good measure.

harrishancock:
Done. I added one commit at the end to fix a minor build error in async-io-win32.

@kentonv If the `KJ_CONSTEXPR(const)` dependency array change works for you, then I'd consider this branch good to go. Keep in mind I haven't resolved the failing tests or any of the warnings.

kentonv:
Merged!

PR #399, opened 2016-12-29
Fix conversion build error when building with Clang 4.0

EricWF:
This patch fixes a build error with Clang 4.0. The emitted error is:

> /tmp/capnproto/c++/src/capnp/layout.h:129:65: error: value of type 'bool' is not implicitly convertible to 'capnp::Kind'

However I think this change is a general bug fix, since it appears that the `ElementSizeForType` specialization for `List` was incorrect, and should have always specialized the second parameter as type `Kind` not `bool`.


kentonv:
Thanks!

PR #401, opened 2017-01-04
Add getFd method to FdInputStream and FdOutputStream

zombiezen:
These methods are useful if you're receiving a generic stream in application code and would like to check some non-stream behavior about the file (e.g. stat or isatty).  Of course, downcasting can fail and code should not rely on the file descriptor being available.

(Contribution is copyright Google Inc.)


kentonv:
Seems reasonable.

Issue #413, opened 2017-02-11
sample calculator server compile error with vs2015 update3


nneesshh:
1>..\src\Capnproto\calculator-server.c++(84): error C2558: class 'Calculator::Function::Client': no copy constructor available or copy constructor is declared 'explicit'
1>..\src\Capnproto\calculator-server.c++(125): error C2558: class 'capnp::CallContext<Calculator::Function::Server::CallParams,Calculator::Function::Server::CallResults>': no copy constructor available or copy constructor is declared 'explicit'
1>..\src\Capnproto\calculator-server.c++(173): error C2558: class 'capnp::CallContext<Calculator::Server::EvaluateParams,Calculator::Server::EvaluateResults>': no copy constructor available or copy constructor is declared 'explicit'
========== Build: 0 succeeded, 1 failed, 0 up-to-date, 0 skipped ==========

harrishancock:
The build error here is caused by an MSVC compiler bug: when capturing variables by value in lambda capture lists, MSVC only appears to consider const copy constructors (`Foo::Foo(const Foo&)`) and never mutable copy constructors (`Foo::Foo(Foo&)`). You can see my initial discovery of the problem at f87b306, and some discussion with @kentonv in #396.

There's an easy workaround: wrap variables with mutable copy constructors in `KJ_CPCAP(...)` when capturing them, which forces MSVC to perform a copy manually. I'll have a patch ready soon that does that for this particular instance of the problem.

However, you will probably run into this bug in the future until we can fully resolve the root cause, either by a fix from Microsoft or by removing the use of mutable copy construction in the codebase.

kentonv:
Looks like this was fixed by #414.

PR #414, opened 2017-02-11
Build samples with MSVC

harrishancock:
This PR:

- Installs the external capnp tools if `CAPNP_BUILD_TOOLS=OFF`. This allows MSVC CMake builds to install complete, working distributions of Cap'n Proto.

- Adds a CMakeLists.txt and applies the mutable capture bug workaround for MSVC (`KJ_CPCAP`) to the C++ samples (`addressbook`, `calculator-{client,server}`).

This may resolve the immediate issue in #413, but it doesn't resolve the underlying problem, so I'm not sure if it should be closed yet.

### DIFF
--- a/c++/cmake/CapnProtoConfig.cmake.in
+++ b/c++/cmake/CapnProtoConfig.cmake.in
@@ -12,8 +12,10 @@
 
 set(CapnProto_VERSION @VERSION@)
 
-set(CAPNP_EXECUTABLE $<TARGET_FILE:CapnProto::capnp_tool>)
-set(CAPNPC_CXX_EXECUTABLE $<TARGET_FILE:CapnProto::capnpc_cpp>)
+if(@CAPNP_BUILD_TOOLS@)
+  set(CAPNP_EXECUTABLE $<TARGET_FILE:CapnProto::capnp_tool>)
+  set(CAPNPC_CXX_EXECUTABLE $<TARGET_FILE:CapnProto::capnpc_cpp>)
+endif()
 set(CAPNP_INCLUDE_DIRECTORY "@PACKAGE_CMAKE_INSTALL_FULL_INCLUDEDIR@")
 
 # work around http://public.kitware.com/Bug/view.php?id=15258
--- a/c++/samples/CMakeLists.txt
+++ b/c++/samples/CMakeLists.txt
@@ -1,0 +1,53 @@
+# A Cap'n Proto sample project.
+#
+# To build (non-MSVC):
+#   1. Install Cap'n Proto somewhere ($PREFIX below):
+#
+#      mkdir capnproto/build
+#      cd capnproto/build
+#      cmake ../c++ -DCMAKE_INSTALL_PREFIX=$PREFIX
+#      cmake --build . --target install
+#      export PATH=$PREFIX/bin:$PATH
+#
+#   2. Ensure Cap'n Proto's executables are on the PATH, then build the sample project:
+#
+#      export PATH=$PREFIX/bin:$PATH
+#      mkdir ../build-samples
+#      cd ../build-samples
+#      cmake ../c++/samples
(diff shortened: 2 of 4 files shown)


kentonv:
Whoops, somehow I missed this PR, sorry.

> Installs the external capnp tools if `CAPNP_BUILD_TOOLS=OFF`.

Hmm, you're saying if we are using an external capnp executable, we'll install that executable? What if it's already installed? What if we're using an external capnp because we're cross-compiling, so it's compiled for the host and not the target? This feels weird.

harrishancock:
I agree -- I'm happy to back that commit out, or change it.

The issue I was attempting to address is this: Right now, the CMake install target does not produce a working install tree when building with MSVC. The CMake package configuration file (`c++/cmake/CapnProtoConfig.cmake.in`) requires that `capnp` and `capnpc-c++` be available as the exported targets `capnp_tool` and `capnpc_cpp`, which of course don't exist.

An alternative is to patch the install tree manually to produce a working package: i.e., copy the `bin` directory from a MinGW-built tree to an MSVC-built tree, and copy the relevant exported targets between the `lib/cmake/CapnProtoTargets*.cmake` files.

Another alternative might be to include some other logic in `c++/cmake/CapnProtoConfig.cmake.in`, perhaps searching for the capnp tools based on an environment or CMake variable that the user sets.

harrishancock:
I realize I didn't define what I mean by "working install tree". I mean a working CMake package: one that can be consumed by a CMake-based client project with `find_package(CapnProto)`, assuming the installed `bin` directory is on the PATH.

kentonv:
That does seem like a reasonable argument.

I'm happy to leave it to your judgment, since I don't have a sense for how cmake is used in practice. My intuition derives more from automake but that may be the wrong intuition to apply here.

Or maybe @pqu could provide a third opinion?

harrishancock:
@kentonv another option would be to get the tools building with VS2017, and officially support Cap'n Proto as an installable CMake package only with VS2017. I tried doing that last weekend, but ran into another ICE related to `NullableValue::~NullableValue()`'s `noexcept` specification. I will try again this weekend when I update my AppVeyor PR.

pqu:
My inclination is that if `CAPNP_BUILD_TOOLS` is `OFF`, then `CAPNP_EXECUTABLE`/`CAPNPC_CXX_EXECUTABLE` should simply not be set. This issue effects all platforms (not just MSVC) and the cross-compiling scenario Kenton raised is compelling.

harrishancock:
@pqu thanks for weighing in! I think leaving them unset makes sense. It would break the `capnp_generate_cpp()` function, but client projects could work around this by specifying `CAPNP_EXECUTABLE` etc. on the cmake command line.

To make this easier, perhaps we could use `find_program()` to search for the CLI tools on the PATH if `CAPNP_BUILD_TOOLS` is off, and emit a warning explaining to the user that `capnp_generate_cpp()` will fail, if they're not found? This would avoid forcing the user to pass explicit path(s) on the command line, but might carry a risk of finding the wrong version of the CLI tools. My instinct is that this risk is low, however.

pqu:
I'm a little weary of adding too much logic for what is essentially a workaround (and `capnp_generate_cpp()` will already emit a semi-friendly error). The best solution would be as you mentioned above: enable building the tools with MSVC.

harrishancock:
Thanks for the help! I've followed your advice.

@kentonv I updated this PR:

- Rebased onto master.
- Added one commit to fix a minor build problem (missing `_exit` header in `test-helpers.c++`).
- Changed the `CapnProtoConfig.cmake.in` logic to follow @pqu's suggestion.
- Added more comments to `c++/samples/CMakeLists.txt` to document the workaround needed to find the command line tools when using MSVC.

kentonv:
Thanks!

PR #417, opened 2017-02-18
fix bug in parsing of non-ascii constants

dwrensha:
`spk dev` currently fails on Wekan due to a [non-ascii character](https://github.com/wekan/wekan/blob/81de7bceecf96c31a49baee67abfd6e597652e18/sandstorm-pkgdef.capnp#L166) in its sandstorm-pkgdef.capnp file.

This problem was introduced in https://github.com/sandstorm-io/capnproto/commit/9182e67f146d81cbb6865230172ba6a7bbd81ec6.

This commit fixes the problem.


kentonv:
ARRRRGGGHHHH, why is `char` signed?

PR #420, opened 2017-02-25
Make sure to clear tag word when upgrading a struct list.

### LINKED ISSUE #419: tag not cleared in struct list upgrade
Consider the following schema and program:

```capnp
# upgrade.capnp
@0xca1a6fd5ed3b4bb1;

struct RootStruct {
   field @0 :AnyPointer;
}

struct OldVersion {
   old0 @0 :UInt32;
   old1 @1 :UInt32;
}

struct NewVersion {
   old0 @0 :UInt32;
   old1 @1 :UInt32;
   new2 @2 :UInt64;
}
```

```cpp
// upgrade.c++

#include "upgrade.capnp.h"
#include <capnp/message.h>
#include <capnp/serialize.h>
#include <capnp/serialize.h>

int main() {
  ::capnp::MallocMessageBuilder message;
  auto root = message.initRoot<RootStruct>();
  auto oldList = root.getField().initAs<capnp::List<OldVersion>>(1);
  oldList[0].setOld0(0xab);
  oldList[0].setOld1(0xcd);

  // Reallocate the list with enough space for the new version of the struct.
  auto newList = root.getField().getAs<capnp::List<NewVersion>>();

  capnp::writeMessageToFd(1, message);
}
```

I expect this to write a message that looks like this:

```
0000 0000 0700 0000 // one segment with 7 words
0000 0000 0000 0100 // struct pointer, pointer section has one pointer
0900 0000 1700 0000 // list pointer, offset 2, inline composite, 2 words long
0000 0000 0000 0000 // [zeroed old list tag]
0000 0000 0000 0000 // [zeroed old list body]
0400 0000 0200 0000 // list tag. 1 element. data section has 2 words.
ab00 0000 cd00 0000 // first element, first word
0000 0000 0000 0000 // first element, second word
```

Instead, it writes this message:

```
0000 0000 0700 0000 // one segment with 7 words
0000 0000 0000 0100 // struct pointer, pointer section has one pointer
0900 0000 1700 0000 // list pointer, offset 2, inline composite, 2 words long
0400 0000 0100 0000 // old list tag! (BUG)
0000 0000 0000 0000 // [zeroed old list body]
0400 0000 0200 0000 // list tag. 1 element. data section has 2 words.
ab00 0000 cd00 0000 // first element, first word
0000 0000 0000 0000 // first element, second word
```

dwrensha:
Fixes #419.

### DIFF
--- a/c++/src/capnp/layout.c++
+++ b/c++/src/capnp/layout.c++
@@ -1360,7 +1360,8 @@ struct WireHelpers {
       }
 
       // Zero out old location.  See explanation in getWritableStructPointer().
-      memset(oldPtr, 0, oldStep * elementCount * BYTES_PER_WORD / BYTES);
+      // Make sure to include the tag word.
+      memset(oldPtr - 1, 0, (1 + oldStep * elementCount) * BYTES_PER_WORD / BYTES);
 
       return ListBuilder(origSegment, capTable, newPtr, newStep * BITS_PER_WORD, elementCount,
                          newDataSize * BITS_PER_WORD, newPointerCount, ElementSize::INLINE_COMPOSITE);
--- a/c++/src/capnp/layout.c++
+++ b/c++/src/capnp/layout.c++
@@ -1361,7 +1361,8 @@ struct WireHelpers {
 
       // Zero out old location.  See explanation in getWritableStructPointer().
       // Make sure to include the tag word.
-      memset(oldPtr - 1, 0, (1 + oldStep * elementCount) * BYTES_PER_WORD / BYTES);
+      memset(oldPtr - POINTER_SIZE_IN_WORDS, 0,
+             (POINTER_SIZE_IN_WORDS + oldStep * elementCount) * BYTES_PER_WORD / BYTES);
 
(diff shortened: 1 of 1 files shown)


c++/src/capnp/layout.c++
```diff
@@ -1360,7 +1360,8 @@ struct WireHelpers {
       }
 
       // Zero out old location.  See explanation in getWritableStructPointer().
-      memset(oldPtr, 0, oldStep * elementCount * BYTES_PER_WORD / BYTES);
+      // Make sure to include the tag word.
+      memset(oldPtr - 1, 0, (1 + oldStep * elementCount) * BYTES_PER_WORD / BYTES);
```
kentonv: I think this will fail unit analysis. The `1` needs to be `1 * WORDS`.

dwrensha: How do I run the unit analysis?

kentonv: Build with `-DCAPNP_DEBUG_TYPES`.

kentonv: Ah, yes, that is more correct!

PR #421, opened 2017-02-26
Edit error message for the case where copyPointer() hits a FAR pointer.

dwrensha:
The existing error message suggests this case can only arise if there is a
a bug in the library. However, malformed input can trigger the error too.
In particular, the error gets thrown when a non-double far pointer resolves
to another far pointer.

### DIFF
--- a/c++/src/capnp/layout.c++
+++ b/c++/src/capnp/layout.c++
@@ -738,7 +738,7 @@ struct WireHelpers {
         break;
       }
       case WirePointer::FAR:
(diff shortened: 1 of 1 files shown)


dwrensha:
I've pushed a commit that switches the macro used in this case in `copyPointer()` and `totalSize()` from `KJ_FAIL_ASSERT` to `KJ_FAIL_REQUIRE`, as per [these docs](https://github.com/sandstorm-io/capnproto/blob/01f6d5e4ff05fcd67e968b50120dba3fbbb38afb/c%2B%2B/src/kj/debug.h#L46-L48).

PR #423, opened 2017-03-07
fix various bugs in canonicalization

dwrensha:
These were found with the help of [afl](http://lcamtuf.coredump.cx/afl/).

Both of the added test cases fail when run on the current master branch.

### DIFF
--- a/c++/src/capnp/layout.c++
+++ b/c++/src/capnp/layout.c++
@@ -2965,6 +2965,9 @@ bool ListReader::isCanonical(const word **readHead) {
       }
       auto structSize = (this->structDataSize / BITS_PER_WORD) +
                         (this->structPointerCount * WORDS_PER_POINTER);
+      if (structSize == 0 * WORDS) {
+        return true;
+      }
       auto listEnd = *readHead + (this->elementCount / ELEMENTS * structSize) / WORDS;
       auto pointerHead = listEnd;
       bool listDataTrunc = false;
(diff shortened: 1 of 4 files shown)


dwrensha:
I've pushed a commit that additionally fixes the case where `inlineCompositeWordCount()` is larger than `elementCount * structSize` (but still small enough to pass the bounds check).

dwrensha:
I've pushed a commit that fixes some off-by-one bugs that can cause canonicalization to drop meaningful data from a message.

c++/src/capnp/layout.c++
```diff
@@ -2951,7 +2951,7 @@ ListReader ListReader::imbue(CapTableReader* capTable) const {
   return result;
 }
 
-bool ListReader::isCanonical(const word **readHead) {
+  bool ListReader::isCanonical(const word **readHead, const WirePointer *ref) {
```
kentonv: Did you accidentally add some extra indentation here?

dwrensha: Oops. Fixed.

I think emacs wanted the indentation because of the enclosing `namespace {`.

kentonv:
@dwrensha Is your AFL test setup something that might make sense to commit? I'd like to have AFL in our automated release tests before the next release.

dwrensha:
I've posted the setup [here](https://github.com/dwrensha/fuzz-capnproto). Do with it as you please!

PR #427, opened 2017-03-09
omit needless memset

dwrensha:
`ptr` is freshly allocated, so the memory it points to is already zeroed, so this `memset()` is always a no-op. For reference, it traces back to https://github.com/sandstorm-io/capnproto/commit/bd49b14a657664ddc714f905f2077e6ad99d52e1.


kentonv:
Nice.

PR #436, opened 2017-03-30
doc/cxxrpc: Document lazy execution

ryanofsky:
This change takes a comment about lazy execution from
https://github.com/sandstorm-io/capnproto/issues/263#issuecomment-158564701 and
adds it to the C++ RPC documentation.

Although lazy execution is described in `Promise::then()` method documentation,
it's unfortunately buried in the *Advanced implementation tips* section, and
described as an optimization rather than as a more general feature of the
asynchronous execution model. Since I recently got tripped up by this behavior,
I thought it would be worth trying to document better.


kentonv:
Thanks! I'll be merging and tweaking this as part of the 0.6 release, probably the week after next. (I'm out-of-town next week.)

PR #437, opened 2017-03-31
Actually implement integer overflow guards

kentonv:
Way back in this blog post, I described how I used template metaprogramming to detect integer overflows in Cap'n Proto:

https://capnproto.org/news/2015-03-02-security-advisory-and-integer-overflow-protection.html

Unfortunately, at the time I only got that code up to the point of being able to compile layout.c++ without errors. I didn't ever get it _running_, much less did I merge it. And over the years, it of course got out-of-date.

This PR updates the code to the current master and actually runs, with all tests passing.

I ran all the benchmarks and determined that if there is any performance impact, it is within the noise.

Code size has increased by 5%-ish. Remarkably, though, there is little difference between CAPNP_DEBUG_TYPES vs. not, so this increase seems to be coming from some structural changes rather than from any failure to elide the extra checks.

### DIFF
--- a/c++/src/kj/units.h
+++ b/c++/src/kj/units.h
@@ -376,9 +376,6 @@ private:
   template <typename Number1, typename Number2, typename Unit2>
   friend inline constexpr auto operator*(Number1 a, Quantity<Number2, Unit2> b)
       -> Quantity<decltype(Number1() * Number2()), Unit2>;
-
-  template <typename T>
-  friend inline constexpr T unit();
 };
 
 template <typename T> struct Unit_ {
--- a/c++/src/kj/units.h
+++ b/c++/src/kj/units.h
@@ -182,12 +182,13 @@ private:
   template <typename OtherNumber, typename OtherUnit1, typename OtherUnit2>
   friend class UnitRatio;
 
-  template <typename N1, typename N2, typename U1, typename U2>
+  template <typename N1, typename N2, typename U1, typename U2, typename>
   friend inline constexpr UnitRatio<decltype(N1() * N2()), U1, U2>
       operator*(N1, UnitRatio<N2, U1, U2>);
 };
 
-template <typename N1, typename N2, typename U1, typename U2>
+template <typename N1, typename N2, typename U1, typename U2,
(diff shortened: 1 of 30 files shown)


kentonv:
Note that this was one of the major blockers to doing a release, since I promised to actually merge this for the next release.

kentonv:
@harrishancock I fear that MSVC may hate this, especially when CAPNP_DEBUG_TYPES is on. FWIW we could go back to disabling debug types on MSVC if that helps.

c++/src/kj/units.h
```diff
@@ -347,21 +374,31 @@ class Quantity {
 
   template <typename Number1, typename Number2, typename Unit2>
   friend inline constexpr auto operator*(Number1 a, Quantity<Number2, Unit2> b)
-      -> Quantity<decltype(Number1(1) * Number2(1)), Unit2>;
+      -> Quantity<decltype(Number1() * Number2()), Unit2>;
 
   template <typename T>
   friend inline constexpr T unit();
```
harrishancock: To make MSVC happy, I had to change this friend declaration to match the trailing return type SFINAE style below, i.e.:

```c++
template <typename T>
friend inline constexpr auto unit() -> decltype(Unit_<T>::get());
```

Otherwise MSVC complains about an "ambiguous call to overloaded function", interpreting this friend declaration as a prototype for a function different from the trailing return type overload defined below.

harrishancock:
Other than the issue in units.h that I commented on, MSVC miraculously builds this branch!

kentonv:
@harrishancock Wow! Did you test with CAPNP_DEBUG_TYPES=1, or only without it?

c++/src/kj/units.h
kentonv: Turns out this line is vestigial and can just be deleted. I guess GCC and Clang ignore the friend declaration if it doesn't match any real declaration.

harrishancock:
Oops, only without, it turns out. Actually defining the symbol makes the build explode with errors, you're right.

I imagine MSVC building the type debugging would be a nice-to-have, but not a requirement for the next release?

kentonv:
Yeah, as I said, I don't think it's particularly important to make that work. I view building with debug types as being like running a separate verification tool... not everyone has to do it.

Issue #245, opened 2015-09-15
Protobuf migration tool or technique


ekg:
Has anyone developed a method to port c++ based on protobuf to capnp? It looks like a series of find and replace operations will get me most of the way, but I'm curious about peoples experiences.

paulproteus:
I for one haven't, but I do want to say hi Erik! I'm curious about @kentonv 's perspective on this.

ekg:
Hi Asheesh! :waves: I have a ~20k line codebase that'd I'd like to migrate and test, so any timesaving techniques would be very helpful.

One of the biggest frustrations I've had with protobuf is that messages can't be larger than 64MB. In my case we need to be able to handle messages of up to 1TB or there will be a need to implement hacks on serialization and deserialization. I'm also curious if this kind of use pattern is supported by capnp. Is there a max message size?

ekg:
Whoops that last comment isn't entirely clear. Another way of asking is "how hard is it to implement streaming parsers in capnp?"

kentonv:
Much like with protobufs, there is a default enforced message size (I think of 64MB), but you can increase the default (see `capnp::ReaderOptions`). The default exists in part to prevent resource exhaustion attacks, e.g. where someone sends you a gigabyte message when you were only expecting a dozen bytes.

Protobufs had a theoretical maximum message size of 2^31 bytes. Cap'n Proto's max is 2^64 bytes (though note that a particular List, Text, or Data item cannot exceed 512M elements).

Cap'n Proto will, of course, scale much better to large messages since you can use mmap() and read just what you need rather than parsing upfront.

However, note that Cap'n Proto still isn't great for large _mutable_ data. Cap'n Proto messages use an arena allocator, meaning all new objects are allocated at the end of the message, and space is not reclaimed from "freed" objects. If you are writing your 1TB message all at once, this is fine, but if you intended to represent a 1TB database which you regularly modify, then you'll need to break it up into small pieces that you can modify as units.

To answer the original question: Inside Google, a great deal of effort was spent developing tools to automate a migration from Proto1 to Proto2 even though they had nearly-identical APIs. Moving to Cap'n Proto is a much larger jump. Unfortunately I don't think you'll be able to automate it, though you may be able to automate, say, 75% of it. I'm not aware of any tools written for this purpose currently.

kentonv:
I'm going to close this as I don't think a tool for this is likely to be built. That said, if someone wants to work on it, please do get in touch!

[...]

PR #444, opened 2017-04-12
MallocMessageBuilder: don't allocate segments above the max serializable size

mlin:
MallocMessageBuilder::allocateSegment increases the segment size on each sequential invocation, irrespective of the size actually requested by the caller. After enough invocations, it allocates more than MAX_SEGMENT_WORDS even if the caller didn't request so much. Then BuilderArena rejects the returned segment (verifySegmentSize) leading to breakage.

This would only affect multi-gigabyte messages where we also (foolishly) raise the traversal limit to accommodate. By capping excess segment allocation to MAX_SEGMENT_WORDS, @orodeh and I buy ourselves more headroom before we have to fix our application to not depend on such huge messages. (cc @ekg who had a related question on #245)

Thank you for this awesome tool!

### DIFF
--- a/c++/src/capnp/message.c++
+++ b/c++/src/capnp/message.c++
@@ -242,6 +242,8 @@ MallocMessageBuilder::~MallocMessageBuilder() noexcept(false) {
 }
 
 kj::ArrayPtr<word> MallocMessageBuilder::allocateSegment(uint minimumSize) {
+  KJ_REQUIRE(minimumSize <= MAX_SEGMENT_WORDS, "MallocMessageBuilder asked to allocate segment above maximum serializable size.");
+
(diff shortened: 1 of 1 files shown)


kentonv:
Good catch.

Actually, really huge messages are perfectly fine as long as your application is expecting them. Nothing "foolish" here -- creating huge Cap'n Proto files and mmap()ing them in is very much an intended use case.

It looks like `nextSize` will eventually overflow, if the message grows beyond 4GB. Perhaps we should actually be capping `nextSize` on line 282?

mlin:
Good idea- I reworked it that way!

kentonv:
Thanks!

PR #449, opened 2017-04-19
canonicalization: make sure that pad bits of primitive lists are zeroed

### LINKED ISSUE #429: canonicalize() and isCanonical() don't require primitive list pad bytes to be zero
Consider this message, a list of three single-byte elements:

```
        {0x01, 0x00, 0x00, 0x00, 0x1a, 0x00, 0x00, 0x00,
         0x01, 0x02, 0x03, 0x01, 0x00, 0x00, 0x00, 0x00}, // fourth byte is non-zero!
```

or this message, a list of eleven single-bit elements:

```
        {0x01, 0x00, 0x00, 0x00, 0x59, 0x00, 0x00, 0x00,
         0xee, 0x0f, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00}  // twelfth bit is non-zero!
```

In my opinion, these messages should not be considered canonical, because they have non-zero padding that is inaccessible to the user, and no amount of schema evolution is going to make that non-zero padding data significant. However, the current implementation of capnproto-c++ does in fact consider these messages to be canonical, and `canonicalize()` does not zero out the padding.

dwrensha:
Fixes #429.

I've not yet been able to figure out how to get this to pass the `-DCAPNP_DEBUG_TYPES` analysis. @kentonv, maybe you can help me with that?

### DIFF
--- a/c++/src/capnp/canonicalize-test.c++
+++ b/c++/src/capnp/canonicalize-test.c++
@@ -332,6 +332,61 @@ KJ_TEST("isCanonical requires truncation of 0-valued struct fields in all list m
 
   KJ_ASSERT(!nonTruncated.isCanonical());
 }
+
+KJ_TEST("primitive list with nonzero padding") {
+   AlignedData<3> segment = {{
+     // Struct, one pointer field.
+     0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x01, 0x00,
+
+     // List of three byte-sized elements.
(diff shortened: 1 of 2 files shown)


dwrensha:
OK, with `unbound()` and `assertMax()`, I was able to get the analysis to pass.

kentonv:
Thanks! I tweaked the debug-type-safety a bit in cccc78d.

PR #451, opened 2017-04-21
Add kj-http to cmake build.

kentonv:
@harrishancock does this look right?

(Yay, AppVeyor will test it!)

### DIFF
--- a/c++/src/kj/CMakeLists.txt
+++ b/c++/src/kj/CMakeLists.txt
@@ -126,6 +126,22 @@ if(NOT CAPNP_LITE)
   install(FILES ${kj-async_headers} DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/kj")
 endif()
 
+# kj-http ======================================================================
+
+set(kj-http_sources
+  compat/http.c++
+)
+set(kj-http_headers
+  compat/http.h
+)
+if(NOT CAPNP_LITE)
+  add_library(kj-http ${kj-http_sources})
+  add_library(CapnProto::kj-http ALIAS kj-http)
+  target_link_libraries(kj-http kj-async kj)
+  install(TARGETS kj-http ${INSTALL_TARGETS_DEFAULT_ARGS})
+  install(FILES ${kj-http_headers} DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/kj/compat")
+endif()
+
 # Tests ========================================================================
 
 if(BUILD_TESTING)
@@ -161,8 +177,9 @@ if(BUILD_TESTING)
       one-of-test.c++
       function-test.c++
       threadlocal-pthread-test.c++
+      compat/http-test.c++
     )
-    target_link_libraries(kj-heavy-tests kj-async kj-test kj)
+    target_link_libraries(kj-heavy-tests kj-http kj-async kj-test kj)
     add_dependencies(check kj-heavy-tests)
     add_test(NAME kj-heavy-tests-run COMMAND kj-heavy-tests)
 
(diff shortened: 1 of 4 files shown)


kentonv:
Hmm, it seems that every commit I upload to any branch triggers and appveyor build, *and* any pull request also triggers a build testing a merge with master.

I think I'd prefer to have only PRs trigger builds but I'm not entirely sure how to do that.

kentonv:
Sadness, MSVC ICEs.

This code isn't particularly complicated in terms of C++ features used. @harrishancock any interest in taking a look? (I might also poke at it on Monday but if I can't figure it out quickly I'll probably leave this out of the 0.6 release.)

harrishancock:
> @harrishancock does this look right?

Yes. I'd note that you don't need to explicitly link to `kj` -- `kj-async` publicly links to `kj` already, so `kj-http` will transitively link to it. It's correct either way.

I see that the build failed with:

```
(ClCompile target) -> 
  C:\projects\capnproto\c++\src\kj/parse/char.h(192): fatal error C1001: An internal error has occurred in the compiler. [C:\projects\capnproto\build-vs2015\src\kj\kj-http.vcxproj]
```

If `kj-http` uses the parsing code, then MSVC won't be able to build it for now. :(

harrishancock:
@kentonv This ICE is the `anyOfChars` ICE, which I actually do have a workaround for:

```c++
#define anyOfChars(chars) CharGroup_().orAny(chars)
```

I've been hesitant to submit that, hoping that MSVC would just get its act together, but since VS2017 has a new, and unrelated ICE, it might be worthwhile at least supporting the parsing code with VS2015.

I have to run, but will look at this more closely tonight.

harrishancock:
One last thing: on AppVeyor, under 'Settings -> General', I see there's a "Do not build feature branches with open Pull Requests" checkbox. Perhaps this might partially give you the behavior you want? If so, we can bake it into the `appveyor.yml` with `skip_branch_with_pr: true`.

kentonv:
@harrishancock I went ahead and added your work-around (testing now). If it works then I think it's worthwhile to use it, as it would be wonderful if we don't have to make people jump through as many hoops on MSVC. It is super-ugly, but it's better than not building the file at all...

kentonv:
@harrishancock I've tried enabling `skip_branch_with_pr`, we'll see if it helps. But I think appveyor started building the branch before I even opened the PR last time, so I'm not sure how that's supposed to work... meh, we'll see.

kentonv:
(I enabled `skip_branch_with_pr` after pushing the latest commit so it had already scheduled a branch build, which I now canceled manually, which causes it to show up as an error... oh well.)

kentonv:
Still ICEs. :( @harrishancock, did I implement your work-around correctly?

harrishancock:
Yes, you did. The ICE in appveyor build 16 looks like a new, unrelated ICE. 🙁 There may be other work I had to do to get the parsing code building last year, or it could just be a regression from VS2015U1, which was the only version I ever truly got everything building for. Will look at it tonight. (I know I say that every time...)

harrishancock:
@kentonv, I found two workarounds for this ICE. You can find the first workaround in this branch:

https://github.com/harrishancock/capnproto/commits/cmake-http

The second workaround is simply to omit the trailing return type on the two functions affected (`Sequence_::operator()` and `Sequence_::parseNext()`), relying on C++14's return type deduction. I have no real preference between the two.

The first workaround is similar to cff9ba0, which makes me wonder if it's the same bug, just expressed in a constexpr context.

kentonv:
Wowee, it works!

@harrishancock I'm eager to know if this means we can build the tools and libcapnpc with MSVC. :)

harrishancock:
> I'm eager to know if this means we can build the tools and libcapnpc with MSVC. :)

The answer appears to be yes! However, more workarounds are required -- I'll have a PR up once I've de-uglified them.

PR #461, opened 2017-04-25
Bug fix rollup for 0.6 release

kentonv:


kentonv:
Looks like `KJ_CPCAP` is still needed specifically for `Capability::Client`, which has a non-const copy constructor but doesn't suffer from the ABI compatibility bug since it's a non-default copy constructor.

kentonv:
Note that AppVeyor did actually run and pass on 336f6bd: https://ci.appveyor.com/project/kentonv/capnproto/build/28

But I added a commit after than, then removed it, and AppVeyor apparently started a new build whereas Travis stuck with its previous cached result. Heh.

PR #464, opened 2017-04-28
Have AppVeyor run the tests.

kentonv:


kentonv:
@harrishancock I'm going to try to fix these tomorrow but let me know if you've done any work on this already.

harrishancock:
@kentonv Awesome! I have a fix for the segfault in `kj-heavy-tests.exe` which will be in my tools PR (almost done). I also have an idea for how to implement the `PtmfHelper` class required for `Async/GetFunctorStartAddress`, but it's just an idea last year. I'll propose it once I refresh my memory on it. Beyond that, I haven't investigated any of the other failures yet.

kentonv:
@harrishancock I removed your HeapArrayDisposer change but couldn't reproduce the crash that it was supposed to fix. If you can repro the problem, I'll happily accept a fix PR which moves the test globals to be lazily initialized.

I also removed the ParserRef change and replaced it with one that simply uses `const` on MSVC.

kentonv:
Apparently, appveyor reproduces the kj-heavy-tests crash, even though I don't see it on a local build. Guess I'll iterate slowly.

kentonv:
Apparently all the CI services use GCC 4.8 but the oldest version of GCC I can find to install is 4.9, which fixed the bug in question. Grr.

kentonv:
@harrishancock Maybe we should skip the MinGW build since it means GCC 4.8 bugs (which Travis is already catching) will blocks us from seeing MSVC results, which is a little sad.

Better yet is if we could do the MinGW build in parallel. (And run the tests, too.)

I'm leaving it as-is for now, but just something to think about.

kentonv:
ZOMG IT PASSED

kentonv:
:ship: :shipit: :rocket:

harrishancock:
Awesome!!! Thanks for taking the reins on the tests. :)

The nondeterminism in the `HeapArrayDisposer` and `WSAESHUTDOWN` bugs is weird. FWIW, my home machine reproduces both bugs consistently. It's a ThinkPad Yoga 14 running Windows 10, no VM in between, with VS2017 installed, but compiling with the older VS2015 compiler.

> I also removed the ParserRef change and replaced it with one that simply uses const on MSVC.

Now that I understand that global initializers are scrupulously avoided, this sounds reasonable.

> Better yet is if we could do the MinGW build in parallel. (And run the tests, too.)

Yeah! It looks like AppVeyor's "build matrix" might help us here: https://www.appveyor.com/docs/build-configuration/#build-matrix

harrishancock:
Oh, and to be clear, master branch tests now pass 100% on my home machine.

kentonv:
@harrishancock I used a Windows 10 development VM downloaded from Microsoft, with VS community edition.

I did notice that if I told cmake I wanted VS 2017 rather than 2015, things get ICEy. Which seems weird because I thought 2017 is what fixed a lot of the bugs?

harrishancock:
No, that's a miscommunication on my part. VS2017 indeed has one more ICE left -- I'm puzzling over it right now. VS2015 felt very close to the finish line, so I prioritized work on that.

kentonv:
Hmm, so if people installed VS2015 two years ago, they should be able to compile Cap'n Proto with it? We don't need to tell people VS2017 is required? If so that's pretty great.

harrishancock:
Maybe -- there have been three updates, and we're only building against Update 3. I wouldn't be surprised if Update 1 were able to build it, though. I remember Update 2 had some regressions that made it impossible.

PR #465, opened 2017-04-28
Allow MSVC to build capnp command line tools

harrishancock:
This PR:

- Works around a couple run-of-the-mill MSVC bugs.
- Fixes a static initialization order issue that caused `kj-heavy-tests.exe` to crash (87b7d79).
- Works around the MSVC-doesn't-initialize-constexpr-vtable-pointers bug. For the `ParserRef` workaround, I also had to take some care to avoid the static initialization order fiasco.
- Fixes an error that prevented `capnpc-c++` from being located correctly by CMake.
- Removes the CAPNP_BUILD_TOOLS CMake option.

@kentonv, the static initialization order issues are unfortunate here. :( In general, any use of the `Disposer::instance`-style idiom risks a crash, unless they can be made constexpr. Since all of the variables in question in this PR have vtables, they cannot be made constexpr lest MSVC generate incomplete code. If you worry about dynamic initializers bleeding performance, we could add some `#ifdef` logic to allow constexpr variables to be used by compilers which compile correctly.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,9 +7,8 @@
 #     support for.
 #   - Use CMake to ...
 #       build Cap'n Proto with MinGW.
-#       build Cap'n Proto with VS2015 and the MinGW-built capnp tools.
-#       build the Cap'n Proto samples with VS2015 and the MinGW-built capnp tools.
-#   - Archive both Cap'n Proto builds in capnproto-c++-{mingw,vs2015}.zip artifacts.
+#       build Cap'n Proto with VS2015.
+#       build Cap'n Proto samples with VS2015.
 
 version: "{build}"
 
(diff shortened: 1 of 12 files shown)


harrishancock:
> So the uninitialized vtable problem in constexpr is really "initialized during dynamic init", right? Hence, ordering problems? But once main() starts, all these vtables are initialized?

Not quite -- to clarify: MSVC does not initialize constexpr vtable pointers at all, not even dynamically. Thus calling a constexpr object's virtual function is a guaranteed segfault, no matter what we do. They must either be regular const with external linkage (risking ordering problems), hidden inside a function, or have internal linkage.

I think your proposed solution (initialize test globals on first use post-main()) should indeed work. Then we can use externally-linked `KJ_CONSTEXPR(const)` without fear, I think.

I won't have any more time to work on this today, though. This weekend I will.

kentonv:
I'm going to take a stab at it.

kentonv:
I've merged most of your commits into #464! Closing this in favor of that.

PR #467, opened 2017-04-29
Fix VS2017 ICE and add build matrix to appveyor.yml

### LINKED ISSUE #466: VS2017 ICEs at kj/common.h:824
[Continuing the discussion from #464.]

While the ICE appears to originate from inside `kj::dtor()`, it seems like an exacerbation of whatever bug required the workaround in commit 19096841c85c. Briefly, we have:

```c++
template <class T>
class NullableValue {
public:
  // ...
  inline ~NullableValue() noexcept(__is_nothrow_destructible(T)) { ... }
};
```

If I change the `noexcept` specifier to `noexcept(false)`, or if I omit it entirely, VS2017 compiles everything and all tests pass.

harrishancock:
Fixes #466.

I wanted to have AppVeyor test VS2017 for this PR, too, so I went ahead and added a build matrix. I'm curious to see what this is going to look like...

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,8 +7,8 @@
 #     support for.
 #   - Use CMake to ...
 #       build Cap'n Proto with MinGW.
-#       build Cap'n Proto with VS2015.
-#       build Cap'n Proto samples with VS2015.
+#       build Cap'n Proto with VS2015 and VS2017.
+#       build Cap'n Proto samples with VS2015 and VS2017.
 
 version: "{build}"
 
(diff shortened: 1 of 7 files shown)


harrishancock:
Hmm, I was hoping the GitHub integration would break out the jobs visibly as well. They're nicely visible on the AppVeyor project page, though.

harrishancock:
MinGW samples fail to build because I forgot to add `-gnu++11` in the samples CMakeLists.txt.

Options:

- Add `if(NOT MSVC) set(CMAKE_CXX_FLAGS "-gnu++11") endif()` to the samples CMakeLists.txt.

- Upgrade to CMake 3.1 and use `target_compile_features()` to propagate the C++11 flag requirement transitively to targets that link to `kj`.

Both are equally easy. I'd much prefer the latter, though, as it simplifies capnproto client projects which use CMake.

kentonv:
Beautiful!

PR #473, opened 2017-05-01
kj::Directory::list*() should return sorted lists to avoid non-determinism.

kentonv:


kentonv:
@dwrensha Can you verify that this fixes the test on your machine? It worked fine before on all of mine...

dwrensha:
Yep, after applying this patch the tests pass for me.

PR #476, opened 2017-05-03
Remove references to obsolete isCallersFault

zenhack:
At some point isCallersFault was marked as obsolete, but the description
of abort still referenced it, suggesting particular usage. This patch
removes said references.

Since `git-diff` does *such* a good job with prose: This just removes
the two parentheticals and re-flows the paragraph.

### DIFF
--- a/c++/src/capnp/rpc.capnp
+++ b/c++/src/capnp/rpc.capnp
@@ -233,11 +233,11 @@ struct Message {
 
     abort @1 :Exception;
     # Sent when a connection is being aborted due to an unrecoverable error.  This could be e.g.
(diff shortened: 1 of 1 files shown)


kentonv:
Thanks!

Once upon a time I had this idea that exceptions should specify whether the problem was the thrower's fault or the caller's fault, like HTTP 400 vs. 500 errors, but then I realized that inevitably when it's the caller's fault, the caller probably won't catch the exception, and then someone up-stack will catch it, and by that time it's no longer correct to describe the error as their fault. So now it's just "someone did something wrong" (which is still distinguish from "a connection dropped or requirements changed; please start over" and "resources exhausted; please come back later" and "that function isn't implemented").

PR #481, opened 2017-05-19
Doccomments in schema.capnp

eqvinox:
I forgot about #311 but the result is pretty similar.

**WARNING: I'm a C developer, not C++. This code may set your computer on fire.** It does actually work for me though.

If someone could take this from here I'd seriously appreciate it...


eqvinox:
actually... the field docs don't quite handle descending into unions correctly *sigh*

kentonv:
Hey, just wanted to say that I've seen this and I hope to take a look some time in the next week. Thanks for working on this!

FWIW it looks like the Travis build is failing due to a new unused variable warning.

kentonv:
@eqvinox Sorry for the long delay on this.

Instead of going back-and-forth, I went ahead and tweaked this code myself and merged it. I made some changes to the interface -- see schema.capnp:

https://github.com/capnproto/capnproto/commit/943a8b6885124e6f6b2fed79da41ea0aa84a2c82#diff-6a549b22a680f941abeb69c0b0ea1c6b

In particular I suspect that we may want to put other kinds of info into this structure in the future, such as the locations of tokens in the original source code (useful for IDEs), so I renamed it to SourceInfo.

I also made sure that enum and method docs are picked up, extended SchemaParser to provide access to this stuff, and added some tests of that.

Let me know if you have any issues.

Thanks for working on this! It's nice to finally have doc comments available to code generator plugins!

Issue #479, opened 2017-05-18
Compilation issue on MSVC


hof:
Created a simple capnproto program that uses generic types in the message schema. See https://github.com/hof/capnp_test 

This causes compile errors on MSVC 2015 and 2017, See: https://ci.appveyor.com/project/hof/capnp-test/build/1.0.2/job/qa5cvfk75dxu33t9#L4592

But not on Linux g++ and clang, See https://travis-ci.org/hof/capnp_test/jobs/232762510 

Am I making a mistake here? Suggestions? Thanks

harrishancock:
Thanks for the report! I'll see about reproducing this, but my hunch is it's another compiler bug we'll need to work around.

Couple quick things:

1. Both of your appveyor jobs are using vs2017. You'll need to add `-G "%CMAKE_GENERATOR%"` to your build script's configure step in order to test both.

2. Thank you for hunterizing CapnProto!

Will get back to you once I reproduce.

hof:
Thanks. I updated the appveyor config.

harrishancock:
@hof I have reproduced this and have an idea for a fix. The problem seems to have to do with MSVC's limited constexpr support.

harrishancock:
This bug's back. MSVC 14.13.26128 (from the latest version of VS2017, which AppVeyor now uses) has a regression which causes this exact same build failure, even with our workaround from #489.

PR #489, opened 2017-05-29
Fix MSVC build error when using List<T> with generic T

harrishancock:
Issue #479 turns out to be a run-of-the-mill expression SFINAE / constexpr problem in MSVC. This PR:

- Adds the offending case to `test.capnp`.
- Fixes it by reverting the definition of `CAPNP_KIND(T)` back to the lite mode way of doing it.
- Sneaks in a couple comment improvements in `capnpc++.c++`, unrelated to the problem at hand.

### DIFF
--- a/c++/src/capnp/common.h
+++ b/c++/src/capnp/common.h
@@ -161,15 +161,19 @@ inline constexpr Kind kind() {
   return k;
 }
 
-#if CAPNP_LITE
+#if _MSC_VER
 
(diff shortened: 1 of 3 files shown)


hof:
FWIW, I can confirm that this works for our project that triggered #479. Thanks!

harrishancock:
Thanks for testing, @hof!

kentonv:
Thanks for the fix!

PR #495, opened 2017-06-09
Add versioning to the libraries in the CMake build

Conan-Kudo:
The Autotools and CMake builds are almost the same, except for
the libraries not being generated as versioned libraries in CMake.

Fixing this lets Linux CMake builds be on the same level as Autotools ones.

### DIFF
--- a/c++/src/capnp/CMakeLists.txt
+++ b/c++/src/capnp/CMakeLists.txt
@@ -64,6 +64,8 @@ target_include_directories(capnp INTERFACE
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/..>
   $<INSTALL_INTERFACE:include>
 )
+# Ensure the library has a version set to match autotools build
+set_target_properties(capnp PROPERTIES VERSION ${VERSION})
 install(TARGETS capnp ${INSTALL_TARGETS_DEFAULT_ARGS})
 install(FILES ${capnp_headers} ${capnp_schemas} DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/capnp")
 
@@ -97,6 +99,8 @@ if(NOT CAPNP_LITE)
   add_library(capnp-rpc ${capnp-rpc_sources})
   add_library(CapnProto::capnp-rpc ALIAS capnp-rpc)
   target_link_libraries(capnp-rpc PUBLIC capnp kj-async kj)
+  # Ensure the library has a version set to match autotools build
+  set_target_properties(capnp-rpc PROPERTIES VERSION ${VERSION})
   install(TARGETS capnp-rpc ${INSTALL_TARGETS_DEFAULT_ARGS})
   install(FILES ${capnp-rpc_headers} ${capnp-rpc_schemas} DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/capnp")
 endif()
@@ -118,6 +122,8 @@ if(NOT CAPNP_LITE)
   add_library(capnp-json ${capnp-json_sources})
   add_library(CapnProto::capnp-json ALIAS capnp-json)
   target_link_libraries(capnp-json PUBLIC capnp kj-async kj)
+  # Ensure the library has a version set to match autotools build
+  set_target_properties(capnp-json PROPERTIES VERSION ${VERSION})
   install(TARGETS capnp-json ${INSTALL_TARGETS_DEFAULT_ARGS})
   install(FILES ${capnp-json_headers} ${capnp-json_schemas} DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/capnp/compat")
 endif()
@@ -139,6 +145,8 @@ set(capnpc_sources
 if(NOT CAPNP_LITE)
   add_library(capnpc ${capnpc_sources})
   target_link_libraries(capnpc PUBLIC capnp kj)
+  # Ensure the library has a version set to match autotools build
+  set_target_properties(capnpc PROPERTIES VERSION ${VERSION})
   install(TARGETS capnpc ${INSTALL_TARGETS_DEFAULT_ARGS})
   install(FILES ${capnpc_headers} DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/capnp")
 endif()
--- a/c++/src/kj/CMakeLists.txt
+++ b/c++/src/kj/CMakeLists.txt
@@ -74,6 +74,8 @@ target_include_directories(kj INTERFACE
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/..>
   $<INSTALL_INTERFACE:include>
 )
+# Ensure the library has a version set to match autotools build
+set_target_properties(kj PROPERTIES VERSION ${VERSION})
 install(TARGETS kj ${INSTALL_TARGETS_DEFAULT_ARGS})
 install(FILES ${kj_headers} DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/kj")
 install(FILES ${kj-parse_headers} DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/kj/parse")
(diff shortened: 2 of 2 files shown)


c++/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 project("Cap'n Proto" CXX)
 cmake_minimum_required(VERSION 3.1)
-set(VERSION 0.7-dev)
+set(RELEASE_VERSION 0.7)
```
kentonv: By separating the "0.7" and "-dev" parts here you're actually breaking our release scripts which automatically update this for each release.

It looks like you're doing this so that only "0.7" goes into the SONAME (library version), but that's actually *not* equivalent to what autotools does. The autotools build uses the whole "0.7-dev" as the version string to embed in the SONAME.

Note that, again, our release scripts remove the -dev automatically in the release branch, so only builds from master would ever end up with the -dev tag. (There is also the -rc1, -rc2, etc. tags that designate release candidates, but final releases have no such suffix.)

So can we revert this part of the change?

kentonv:
Aside from my comment this seems reasonable.

Note that generally we recommend using autotools to build libraries for installation. The cmake build is provided mostly to support cmake-based applications who want to bundle Cap'n Proto into their own build (so, not use installed libraries), and for the MSVC build.

But there's no reason not to make the builds match better.

@harrishancock, any opinions?

c++/CMakeLists.txt
Conan-Kudo: The thing I was concerned about is that `0.6` and `0.6.1`. Both of those are supposed to have the same version at the library level.

The other thought I had was using `SOVERSION` and setting up the autotools like major.minor.patch variables and construct this properly, I just wasn't sure how your ABI guarantee worked, so I erred on the side of caution.

This is the variant I'm using in Fedora: http://pkgs.fedoraproject.org/cgit/rpms/capnproto.git/tree/capnproto-0.6.1-CMake-Add-versioning-to-the-libraries.patch

Conan-Kudo:
@kentonv I switched to the CMake build for Fedora downstream because that way I got both the CMake configuration stuff and the pkgconfig files. Some new packages depend on the CMake configuration stuff and Autotools build doesn't provide them.

c++/CMakeLists.txt
kentonv: Actually, we want 0.6.0 and 0.6.1 to be different versions, or at least that's how we've done it historically. Our ABI policy is that ABI compatibility in C++ is basically impossible and we don't try to provide it. There's an exception for security patch releases, where we add a fourth number -- e.g. 0.5.3.1 -- and we are very careful to make sure not to touch the ABI.

That said, in practice micro releases (third number) have been pretty ABI-stable, and adding security releases has some annoying side effects. Maybe we should rethink the policy for 0.7?

kentonv:
@Conan-Kudo Hmm, I'm not sure how comfortable I am with packages being built from the cmake build. I don't actually know cmake myself so it's hard for me to verify that there's nothing missing there. I accepted the files as a contribution but they were never meant to be canonical.

How hard would it be for the automake build to also install cmake configuration files, just like the cmake build installs pkgconfig?

c++/CMakeLists.txt
Conan-Kudo: Sounds like a good idea to rework for 0.7.

Conan-Kudo: But yeah, I can revert this change. Give me a moment and I'll push a replacement patch.

Conan-Kudo:
@kentonv I removed the controversial change.

harrishancock:
> How hard would it be for the automake build to also install cmake configuration files, just like the cmake build installs pkgconfig?

I think it should be pretty straightforward. A quick fix might be to have the automake build install a `FindCapnProto.cmake` (in the appropriate `share` or `lib` subdirectory), which uses `pkg-config` under the hood to set up the appropriate variables and/or `IMPORTED` targets. CMake has a utility which eases the use of `pkg-config` in such a circumstance: [FindPkgConfig](https://cmake.org/cmake/help/v3.0/module/FindPkgConfig.html#module:FindPkgConfig).

This would still have a significant wart: if a system has the autotools version of CapnProto installed, `find_package(CapnProto)` would use the find-module system, while if a system has the CMake version of CapnProto installed, `find_package()` would use the config-file system.

Getting the automake build to install a real `CapnProtoConfig.cmake` and `CapnProtoConfigVersion.cmake` that match the CMake-generated versions would probably be cleaner, but I'm not sure how hard this would be.

@Conan-Kudo could you share which packages depend on the CMake version of Cap'n Proto?

Conan-Kudo:
@harrishancock One notable package is Ubuntu's Mir, which uses capnproto CMake configuration instead of pkgconfig, because it uses CMake itself.

There's a few others that I don't recall off the top of my head, but that was the major one.

harrishancock:
@kentonv I don't have terribly educated opinions on library versioning on Unix systems -- I have typically just fiddled until things worked and haven't thought deeply on ABI compatibility, since my libraries are all just for internal consumption. FWIW, this is what I do in my own projects:

```cmake
project(MyProject VERSION 0.7.0 LANGUAGES CXX)
# By setting the version on the project-line,
# we get PROJECT_VERSION{,_MAJOR,_MINOR,_PATCH,_TWEAK} for free.

set(preRelease "dev")
# I change this to "" for release builds.

set(PROJECT_VERSION_STRING "${PROJECT_VERSION}")
if(preRelease)
  set(PROJECT_VERSION_STRING "${PROJECT_VERSION_STRING}-${preRelease}")
endif()

add_library(foo ...)
set_target_properties(foo
  PROPERTIES
    VERSION ${PROJECT_VERSION}  # or ${PROJECT_VERSION_STRING}, to include pre-release metadata
    SOVERSION ${PROJECT_VERSION_MAJOR}
)
```

I keep the pre-release metadata in a separate variable so that I can use the `project(... VERSION ...)` CMake feature, and so it's easy to reconstruct in a different way for Debian version strings, which have different rules for hyphens. I'm not advocating this pattern for Cap'n Proto, only providing food for thought. :)

@Conan-Kudo:

> One notable package is Ubuntu's Mir, which uses capnproto CMake configuration instead of pkgconfig, because it uses CMake itself.

Thanks, I'm examining it now. From reading their CMakeLists.txt, I think the band-aid I proposed, or something like it, should allow it to build against the autotools version of Cap'n Proto. I can write such a band-aid, but won't have time for a few days.

kentonv:
OK, I'm accepting this change since it's clearly an improvement, but want to reiterate that I would prefer that distro packages use the automake build, as that is the one that goes through rigorous testing and for which I'm able to provide support.

kentonv:
(Thanks for the fix, BTW!)

Conan-Kudo:
@kentonv Thank you for accepting the change. And for what it's worth, I'll be happy to switch back to the autotools build once it exports the same CMake modules for CMake systems to use.

pqu:
Last time I checked this (#162), the libraries CMake produced with version information weren't compatible with the automake libraries due to how they encoded the version information.  Is this not the case anymore, or not an issue?

Conan-Kudo:
@pqu CMake encodes versioning as `foo.so.ver` on *nix and as `foo-VER.dll` on Windows, and as `foo.ver.dylib` on Mac. This is the basic default libtool rules. The Autotools build can be adjusted to follow these rules, too, since it does use libtool.

kentonv:
@Conan-Kudo Cap'n Proto doesn't use the default libtool behavior, though. It uses the `-release` flag (instead of the `-version-info` flag) to specify an arbitrary version string, which turns into a filename like `libcapnp-VERSION.so`, *not* `libcapnp.so.VERSION`.

https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html

https://www.gnu.org/software/libtool/manual/html_node/Release-numbers.html

pqu:
Building with `cmake -DBUILD_SHARED_LIBS=1` now produces libraries named like:

    libcapnp.0.7-dev.dylib

Building with autotools produces libraries named like:

    libcapnp-0.7-dev.dylib

PR #496, opened 2017-06-09
Add cmake_minimum_required to root CMakeLists.txt

graeme-hill:
There was already a minimum version specified in the actual CMakeLists file in the C++ directory, but since the parent file does not have it you get a warning if you build in the idiomatic cmake fashion:

```bash
mkdir build
cd build
cmake .. # <---- warning here because no minimum version specified
```

I used 3.1 as min version because that is the same as specified in child file.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,1 +1,2 @@
+cmake_minimum_required(VERSION 3.1)
 add_subdirectory(c++)


kentonv:
Seems reasonable but I defer to @harrishancock or @pqu on cmake usage.

harrishancock:
Looks correct to me!

kentonv:
Thanks!

PR #501, opened 2017-06-27
encode/decode json text

lcy03406:
I've found a tool for encode/decode json text is useful, and the easist way to make such a tool is based on the existing capnp tool. So I add an option --json to capnp's subcommand encode and decode, expecting/producing text in json format instead of capnp schema language.

### DIFF
--- a/c++/src/capnp/CMakeLists.txt
+++ b/c++/src/capnp/CMakeLists.txt
@@ -156,7 +156,7 @@ if(NOT CAPNP_LITE)
     compiler/module-loader.c++
     compiler/capnp.c++
   )
-  target_link_libraries(capnp_tool capnpc capnp kj)
+  target_link_libraries(capnp_tool capnp-json capnpc capnp kj)
   set_target_properties(capnp_tool PROPERTIES OUTPUT_NAME capnp)
   set_target_properties(capnp_tool PROPERTIES CAPNP_INCLUDE_DIRECTORY
     $<JOIN:$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/..>,$<INSTALL_INTERFACE:${CMAKE_INSTALL_BINDIR}/..>>
--- a/c++/src/capnp/compiler/capnp.c++
+++ b/c++/src/capnp/compiler/capnp.c++
@@ -40,6 +40,7 @@
 #include <capnp/serialize-packed.h>
 #include <errno.h>
 #include <stdlib.h>
+#include <capnp/compat/json.h>
 
 #if _WIN32
 #include <process.h>
@@ -137,6 +138,8 @@ public:
                       "Do not print warning messages about the input being in the wrong format.  "
                       "Use this if you find the warnings are wrong (but also let us know so "
                       "we can improve them).")
+           .addOption({"json"}, KJ_BIND_METHOD(*this, setJson),
+                      "Print in json format.  ")
            .expectArg("<schema-file>", KJ_BIND_METHOD(*this, addSource))
            .expectArg("<type>", KJ_BIND_METHOD(*this, setRootType))
            .callAfterParsing(KJ_BIND_METHOD(*this, decode));
@@ -171,6 +174,8 @@ public:
                       "deflates zero-valued bytes.  (This writes messages using "
                       "capnp::writePackedMessage() from <capnp/serialize-packed.h>.  Without "
                       "this, capnp::writeMessage() from <capnp/serialize.h> is used.)")
+           .addOption({"json"}, KJ_BIND_METHOD(*this, setJson),
+                      "Expect in json format.  ")
            .addOptionWithArg({"segment-size"}, KJ_BIND_METHOD(*this, setSegmentSize), "<n>",
                              "Sets the preferred segment size on the MallocMessageBuilder to <n> "
                              "words and turns off heuristic growth.  This flag is mainly useful "
@@ -613,6 +618,10 @@ public:
     quiet = true;
     return true;
   }
+  kj::MainBuilder::Validity setJson() {
+    json = true;
+    return true;
+  }
   kj::MainBuilder::Validity setSegmentSize(kj::StringPtr size) {
     if (flat) return "cannot be used with --flat";
     char* end;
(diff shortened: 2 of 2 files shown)


c++/src/capnp/compiler/capnp.c++
```diff
@@ -743,7 +752,12 @@ private:
     kj::String text;
     kj::Maybe<kj::Exception> exception;
 
-    {
+    if (json) {
+      auto root = reader.template getRoot<DynamicStruct>(rootType);
+      JsonCodec codec;
+      codec.setPrettyPrint(pretty);
+	  text = codec.encode(root, rootType);
```
katreniak: coding style: spaces instead of tab

c++/src/capnp/CMakeLists.txt
```diff
@@ -156,7 +156,7 @@ if(NOT CAPNP_LITE)
     compiler/module-loader.c++
     compiler/capnp.c++
   )
-  target_link_libraries(capnp_tool capnpc capnp kj)
+  target_link_libraries(capnp_tool capnp-json capnpc capnp kj)
```
katreniak: autoconf build needs to be updated too.

c++/src/capnp/compiler/capnp.c++
```diff
@@ -1234,6 +1248,21 @@ public:
...
+	  codec.decode(allText, builder);
+      kj::FdOutputStream rawOutput(STDOUT_FILENO);
+      kj::BufferedOutputStreamWrapper output(rawOutput);
+      if (packed) {
+        writePackedMessage(output, arena);
+      } else {
+        writeMessage(output, arena);
+      }
+	  return true;
+	}
```
katreniak: coding style: spaces instead of tab

c++/src/capnp/compiler/capnp.c++
```diff
@@ -1234,6 +1248,21 @@ public:
     EncoderErrorReporter errorReporter(*this, allText);
     MallocMessageBuilder arena;
 
+    if (json) {
+      DynamicStruct::Builder builder = arena.initRoot<DynamicStruct>(rootType);
+	  JsonCodec codec;
+	  codec.decode(allText, builder);
```
katreniak: coding style: spaces instead of tab

c++/src/capnp/compiler/capnp.c++
```diff
@@ -171,6 +174,8 @@ public:
                       "deflates zero-valued bytes.  (This writes messages using "
                       "capnp::writePackedMessage() from <capnp/serialize-packed.h>.  Without "
                       "this, capnp::writeMessage() from <capnp/serialize.h> is used.)")
+           .addOption({"json"}, KJ_BIND_METHOD(*this, setJson),
+                      "Expect in json format.  ")
```
katreniak: Extra white space at the end of the string.

c++/src/capnp/compiler/capnp.c++
```diff
@@ -137,6 +138,8 @@ public:
                       "Do not print warning messages about the input being in the wrong format.  "
                       "Use this if you find the warnings are wrong (but also let us know so "
                       "we can improve them).")
+           .addOption({"json"}, KJ_BIND_METHOD(*this, setJson),
+                      "Print in json format.  ")
```
katreniak: Extra white space at the end of the string.

kentonv:
Hi! I'm currently travelling and won't be able to look at this until next week.

kentonv:
This is a good idea but I think it raises a design issue: What if I want to convert between JSON and binary? I'd have to pipe a `decode` to an `encode`, which seems sad.

(Related to this, I notice that you pay attention to the `pretty` setting to decide whether to output JSON with whitespace. However, this setting can only be controlled on the `decode` command, but JSON is only possibly output from the `encode` command.)

I think perhaps we need a new command: `convert` You'd specify two formats out of "text", "binary", "packed", "flat", "json" and it would convert between them.

This command could also be useful for converting between binary formats without a schema (since technically no schema is needed in these cases).

I'll try to work on this soon.

lcy03406:
@kentonv 

My patch IS to convert between JSON and binary.

What I've done is simply a thin wrapper around the JsonCodec's encode and decode functions, which convert things between JSON and binary.

( We don't have the functionality to convert between json text and capnp text. I guess @katreniak 's work about eval command would contribute this. )

The tricky part is, in JsonCodec's point of view, encode() is to convert something(binary) to JSON. But in capnp_tool's eyes, encode() is to convert something to binary and decode() is to convert binary to something(JSON). So I wired up JsonCodec's encode() into the decode command and vise versa.

And so, the 'pretty' setting works just fine. It is setted by the decode command and passed to JsonCodec's encode() function. So that the JsonCodec's encode() fuction outputs a pretty-formated encoded-json-text, which is the pretty-formated text-decoded-from-binary we want.

kentonv:
Oops, sorry, I must have misread the code. I thought this was being set up as an alternative to --flat, --packed, etc., which is what I would intuitively have expected. Still, this raises the opposite question: what if I actually do want to convert between capnp text and JSON? There's also some confusion if we want to extend eval, since in that case --json would be an alternative to --binary/--flat/--packed/etc., whereas for encode/decode it affects the opposite end... hmm.

lcy03406:
Then the all-in-one convert command is the answer.

But I don't think it's a real requirement to convert between capnp text and JSON. We convert data because we want to use them. For a meaningful conversion, the one end is an 'external' format, which is produced or consumed by other system or people, and the other end is the 'internal' format, which is processed and stored in the system. The internal format is usually some binary format, flat or packed. (Maybe someone prefer plain json? But why they use capnproto in the first place?) Only the external format can be some text, if it's written by humen. Honestly, the capnp text format is not popular, and is not likely to be used outside our systems, so it's not the most common 'external' format either.

Still, I see your concern. All other switches control the binary part but the --json controls the text part. It would confuse users. Maybe an iconv-like interface is better, if all formats are created equal.

kentonv:
I've implemented `capnp convert` including JSON support in #527.

It looks like you've added some other features to the JSON parser in this PR, but you haven't mentioned them in the thread. If you still want these merged, can you open a new PR?

lcy03406:
@kentonv , Thank you for the great 'convert' command, and it's awesome to pack/unpack without schema.

The other things I added are infomation about source line and column, and an option to require all field names are known. It's useful for debuging hand written json. In the original version, json codec asserts on invalid json text and does nothing on mismatched schema, both are not very useful, as assettions report the line number of c++ source rather than the line number of input file. So I modified it in #517 and rebased this pr on top of it.

But why "All checks have failed" and how to fix it?

Issue #510, opened 2017-07-06
Failure to compile with uClibc (missing execinfo.h)


sonologic:
uClibc lies about it's identity, and defines __GLIBC__ in features.h. However, uClibc does not implement execinfo.h, and thus compilation of capnproto fails.

I'm creating a pull request to add an explicit check for execinfo.h to configure.ac, and then use config.h in c++/src/kj/exception.c++ to see if we can have backtraces or not instead of checking for (__linux__ && __GLIBC__) || __APPLE__.

kentonv:
Fixed by #511.

aleksander:
For those who faced this bug with capnproto version without #511 merged in: there is another way, uClibc can be built with UCLIBC_HAS_BACKTRACE flag enabled (development/debugging options in uclibc-menuconfig) which enables support for execinfo.h. So there is no need to disable backtrace support for uClibc.

PR #511, opened 2017-07-06
Do not use execinfo.h with uclibc

sonologic:
Fix for #510

### DIFF
--- a/c++/configure.ac
+++ b/c++/configure.ac
@@ -52,8 +52,6 @@ AC_PROG_CXX
 AC_LANG([C++])
 AX_CXX_COMPILE_STDCXX_11
 
-AC_CHECK_HEADERS(execinfo.h)
-
 AS_CASE("${host_os}", *mingw*, [
     # We don't use pthreads on MinGW.
     PTHREAD_CFLAGS="-mthreads"
--- a/c++/configure.ac
+++ b/c++/configure.ac
@@ -52,6 +52,8 @@ AC_PROG_CXX
 AC_LANG([C++])
 AX_CXX_COMPILE_STDCXX_11
 
+AC_CHECK_HEADERS(execinfo.h)
+
 AS_CASE("${host_os}", *mingw*, [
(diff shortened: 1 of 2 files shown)


kentonv:
Hi Koen,

Cap'n Proto has three different build systems -- autotools, cmake, and ekam. If we're going to introduce a config.h, they will all need to support performing these checks, or in the absence of checks we'll need to default to the previous behavior. Currently, your change will disable backtraces for cmake and ekam users, which would be pretty disruptive to those of us using those build systems.

In what context are you finding that execinfo.h is missing? Can we extend the `#if` to check for that context without adding a config-time check?

Or maybe we can use the new `__has_include()` preprocessor feature to ask the compiler if the header exists? This feature is supported by GCC (at least back to 4.9) and Clang (forever), and can itself be checked for using `#ifdef __has_include`.

sonologic:
Hi,

Sorry, I wasn't aware that there were three build systems with capnproto. Obviously my pr is not going to float then.

We can add check whether __UCLIBC__ is not defined, something like this:

#if ((__linux__ && __GLIBC__) || __APPLE__) && not defined(__UCLIBC__)

Which is quite ugly, but works.

If that is acceptable to you though, i'll change my pr.

Could also do it with __has_include, and default to including execinfo.h if __has_include is not defined. Personally, i'd prefer the __UCLIBC__ solution above though.

Let me know.

The context btw is that i'm cross-compiling capnproto with buildroot for use on a target with uclibc.

kentonv:
Checking for `__UCLIBC__` sounds fine to me. Thanks!

kentonv:
Hmm, since uclibc appears to be Linux-specific, could this be simplified to:

    #if (__linux__ && __GLIBC__ && !__UCLIBC__) || __APPLE__

?

sonologic:
I think that is a valid approach, yes :)

kentonv:
Merged. Thanks!

PR #512, opened 2017-07-07
Fix link errors with -fsanitize=vptr

ecatmur:
When building on gcc 5.4.0 with -fsanitize=vptr (enabled by -fsanitize=undefined) linking libcapnp.a fails in layout.c++.o with `undefined reference to `typeinfo for capnp::ClientHook'.

This is because `ClientHook::NULL_CAPABILITY_BRAND` is defined in layout.c++ and fsanitize=vptr makes that definition depend in some way on the vtable. Simplest workaround is to drag the vtable in as well by defining the first virtual function. Fortunately in this case this is `ClientHook::getLocalServer(_::CapabilityServerSetBase&)` which has a trivial implementation and does not drag in any other symbols.

Note https://gcc.gnu.org/bugzilla/show_bug.cgi?id=64670

### DIFF
--- a/c++/src/capnp/capability.c++
+++ b/c++/src/capnp/capability.c++
@@ -61,10 +61,6 @@ ClientHook::ClientHook() {
   setGlobalBrokenCapFactoryForLayoutCpp(brokenCapFactory);
 }
 
-void* ClientHook::getLocalServer(_::CapabilityServerSetBase& capServerSet) {
-  return nullptr;
-}
-
 // =======================================================================================
 
 Capability::Client::Client(decltype(nullptr))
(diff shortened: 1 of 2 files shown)


kentonv:
Hmm, is this a bug in -fsanitize=vptr? I don't understand how the definition of a `uint` could possibly depend on a vtable.

Can you add a comment as the first line of the function body saying: "Defined here rather than capability.c++ because otherwise building with -fsanitize=vptr fails."

ecatmur:
Done, thanks.

My original reasoning was a bit of the mark; it's actually the member function calls e.g. `isNull()` that provoke the vtable dependency. -fsanitize=vptr takes the opportunity of a member function call to verify that the object (e.g. `cap` argument to `setCapabilityPointer`) has dynamic type actually derived from `ClientHook` and not an unrelated polymorphic type.  This catches bugs (or hacks) like:

```
struct A { virtual int f() = 0; };
struct B { virtual int g() = 0; };
struct D : B { int g() override { return 42; } };
int main() { return reinterpret_cast<A*>(new D)->f(); }
```

kentonv:
Thanks! Sorry for the slow response.

Issue #515, opened 2017-07-12
Header miniposix.h is installed by cmake but not installed by autoconf


katreniak:
As I wrote code generator starting by copying `capnpc-c++.c++`, I happened to use `miniposix.h` file. Then I realized that the header file is installed by `cmake` but not by `autoconf`.

What are the long term plans with this file? It was changed to `noinst_HEADER` in https://github.com/capnproto/capnproto/pull/202. Can it be installed in `autoconf`? Can it be non installed in `cmake`?

kentonv:
This header was not intended to be part of the public API, as it's kind of a hack. It looks like it's installed by the cmake build by mistake.

I would suggest copying the parts you need. Copying may sound ugly but these are really trivial definitions -- they are unlikely to change over time in a way that you would want to track. Keeping a private copy ensures that we can make changes to the header (e.g. eliminating it, if we come up with better abstractions) without breaking your code.

FWIW I think that capnpc-c++.c++ only depends on the header for the STDIN_FILENO and STDOUT_FILENO constants and the mkdir() wrapper.

katreniak:
I already did copy the file locally, I will stick to it. Thanks!

 #516 fixes the cmake "mistake"

PR #516, opened 2017-07-13
cmake: do not install header miniposix.h

katreniak:
synchrozines behavior with autoconf build

https://github.com/capnproto/capnproto/issues/515

### DIFF
--- a/c++/src/kj/CMakeLists.txt
+++ b/c++/src/kj/CMakeLists.txt
@@ -41,7 +41,6 @@ set(kj_headers
   exception.h
   debug.h
   arena.h
-  miniposix.h
   io.h
   tuple.h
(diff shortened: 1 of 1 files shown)


kentonv:
Thanks! But I think the right thing to do with non-installed headers is not to list them at all? E.g. capnp/CMakeLists.txt does not list test-util.h at all.

katreniak:
Amended to address review comments.

kentonv:
Thanks!

The reason the file is tracked in automake is because the automake build is responsible for building dist tarballs, and will only include files it knows about in the dist. Non-installed headers thus have to be declared via `noinst` to make sure they are included. (`noinst` actually has a special meaning to automake.)

PR #522, opened 2017-07-20
Clarify RPC spec documentation based on mail thread

zombiezen:
https://groups.google.com/d/topic/capnproto/_1kBtRSC51s/discussion

### DIFF
--- a/c++/src/capnp/rpc.capnp
+++ b/c++/src/capnp/rpc.capnp
@@ -453,15 +453,14 @@ struct Call {
     # - Vat B dutifully forwards the bar() call to Carol.  Let us call this forwarded call bar'().
     #   Notice that bar() and bar'() are travelling in opposite directions on the same network
     #   link.
-    # - The `Call` for bar'() has `sendResultsTo` set to `yourself`, with the value being the
-    #   question ID originally assigned to the bar() call.
+    # - The `Call` for bar'() has `sendResultsTo` set to `yourself`.
     # - Vat A receives bar'() and delivers it to Carol.
-    # - When bar'() returns, Vat A immediately takes the results and returns them from bar().
-    # - Meanwhile, Vat A sends a `Return` for bar'() to Vat B, with `resultsSentElsewhere` set in
-    #   place of results.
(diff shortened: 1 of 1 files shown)


c++/src/capnp/rpc.capnp
```diff
@@ -453,15 +453,14 @@ struct Call {
...
     #   link.
-    # - The `Call` for bar'() has `sendResultsTo` set to `yourself`, with the value being the
-    #   question ID originally assigned to the bar() call.
+    # - The `Call` for bar'() has `sendResultsTo` set to `yourself`.
     # - Vat A receives bar'() and delivers it to Carol.
-    # - When bar'() returns, Vat A immediately takes the results and returns them from bar().
-    # - Meanwhile, Vat A sends a `Return` for bar'() to Vat B, with `resultsSentElsewhere` set in
-    #   place of results.
-    # - Vat A sends a `Finish` for that call to Vat B.
+    # - When bar'() returns, Vat A sends a `Return` for bar'() to Vat B, with `resultsSentElsewhere`
```
kentonv: Vat A sends the `Return` for `bar()` immediately after sending the `Call` for `bar'()`. It doesn't wait for `bar'()` to return first, since that would introduce an unnecessary round-trip of latency.

c++/src/capnp/rpc.capnp
```diff
@@ -951,8 +950,8 @@ struct CapDescriptor {
     # Hopefully this is unusual.
 
     senderHosted @1 :ExportId;
-    # A capability newly exported by the sender.  This is the ID of the new capability in the
-    # sender's export table (receiver's import table).
+    # A capability exported by the sender.  This may or may not be a new ID in the sender's export
```
kentonv: This feels a little unclear to me. Maybe: "This is the ID of the capability in the sender's export table (receiver's import table). It may be a newly-allocated table entry, or an existing entry (incrementing the refcount)."

zombiezen:
PTAL.  I also added a note that `takeFromOtherQuestion` can only be used once.

kentonv:
Thanks!

PR #527, opened 2017-07-28
Add new `capnp convert` command with JSON support.

kentonv:
This supersedes the `capnp encode` and `capnp decode` commands. It didn't make sense to add JSON to those commands since it was unclear if JSON should be thought of as the "encoded" or "decoded" format. `convert` allows mapping anything to anything.

This command is also useful for, say, converting unpacked format to packed format or vice versa, which can now be done without a schema.


lcy03406:
This command is also useful for, say, converting unpacked format to packed format or vice versa, which can now be done without a schema.

Does it mean it can be used as a common tool for zip/unzip all kinds of data if my data has a lot of zeros?

kentonv:
@lcy03406 Sorry I missed your question earlier.

No, the tool as-is won't work for data that isn't a valid capnp message. Even without a schema, it still expects to see a segment table and to be able to traverse the object tree. In theory the packing format could work without this but we'd need to structure the code differently. (Also, the packing format would still require that your input size is a multiple of 8 bytes.)

Issue #523, opened 2017-07-22
autotools build does not install cmake files?


thomaslee:
Just raising this to be sure it's not an oversight:

While packaging up 0.6.1 I noticed that we've now got cmake config files upstream, which is great. I'm a bit confused by the way they're integrated into the build system, though: running a `./configure && make install` on Linux from a source distribution tarball will not lead to the cmake files under `cmake/` being installed. As far as I can tell we don't even try to generate CapnProtoConfigVersion.cmake from CapnProtoConfigVersion.cmake.in. Is this omission intentional?

If I run a cmake build, I can see that these files are correctly generated (and presumably installed, though I didn't try), but the docs on the web site still seem to point me toward the autotools configure script on Unix.

thomaslee:
Perhaps the assumption is folks on \*nix can use `pkg_check_modules` & the pkg-config files instead of "native" cmake config? If so I've probably got to talk through the implications of ripping out the cmake config files on the Debian side -- I don't want to maintain entirely independent native cmake config if everybody on Linux is going to be using the pkg-config files anyways.

harrishancock:
@thomaslee there was some prior discussion of this in #495, but it deserves its own issue -- thanks for raising it.

> As far as I can tell we don't even try to generate CapnProtoConfigVersion.cmake from CapnProtoConfigVersion.cmake.in. Is this omission intentional?

Yes, it is in intentional, though not desirable. Generating a CMake package config is a feature implemented by the CMake build of capnproto, but not (yet) by the autotools build. This has created an awkward situation: the autotools build is the canonical build system of capnproto, but CMake-based client projects have started depending on the CMake package config, which is not installed by the autotools build. Thus CMake-based client projects are locked into the CMake capnproto build.

A workaround might be to package the CMake build for unix OS package managers, which is what the OP of #495 did, I believe. This is potentially dangerous, as @kentonv pointed out in that PR, because the CMake build does not implement the same ABI stability policy (i.e., no ABI stability across versions) as the autotools build. It is possible to emulate this behavior in CMake ([here's how Roger Leigh did it for Xerces-C++](http://svn.apache.org/viewvc/xerces/c/trunk/src/CMakeLists.txt?revision=1800911&view=markup#l1269)), but it's ugly.

[...]

PR #528, opened 2017-07-29
Align CMake-built shared library filenames, sonames, and rpaths with autotools build

harrishancock:
@kentonv This branch improves the CMake build by installing shared library files which match the autotools build in terms of filename, soname, and rpath (as reported by `readelf -d`). Here are a few gists showing the listing of all `.so` files installed by autotools and CMake on my Ubuntu 16.04 virtual machine, as well as their dynamic headers:

1. [autotools-built .so files](https://gist.github.com/harrishancock/b17f148f823406ce994267c902f88bec)
2. [CMake-built .so files (master branch)](https://gist.github.com/harrishancock/5a121fb4094c4ddd52f97630b4a19ca3)
3. [CMake-built .so files (this branch)](https://gist.github.com/harrishancock/e85015377016dd7fa22852dfda5c59a3)

Note that the first and third listings are now nearly identical.

While writing this branch, I realized I have a question. For every library target `foo`, the autotools build installs `libfoo-version.so` and `libfoo.so` with identical contents. I would have expected `libfoo.so` to be a symlink, rather than a copy. CMake can emulate either behavior, but I was wondering if the copy was intentional?

And a few more notes:

- I haven't been able to verify the results on Mac. @pqu would you be able to help with that?

- @Conan-Kudo I'd appreciate your opinion on this branch, too, since it rewrites your work.

- The `kj_add_library()` (`capnp_add_library()`?) function could be factored differently. For instance, it could be called `kj_install_library()` and deal solely will name-version mangling and installing. I threw in the `add_library()` calls because they were so clearly boilerplate, though I'm a bit wary of custom target functions like that. I'm open to opinions.

### DIFF
--- a/c++/src/CMakeLists.txt
+++ b/c++/src/CMakeLists.txt
@@ -43,6 +43,67 @@ if(BUILD_TESTING)
   add_custom_target(check "${CMAKE_CTEST_COMMAND}" -V)
 endif()  # BUILD_TESTING
 
+# CMake support functions ======================================================
+
+function(kj_add_library target)
+  # Wrapper around `add_library()` to implement certain boilerplate practices:
+  #
+  #   1. The project version is appended to the library's output name (if it's a shared library).
+  #   2. The library gets an ALIAS target in the CapnProto:: namespace.
+  #   3. The library is install()ed using the value of INSTALL_TARGETS_DEFAULT_ARGS as arguments.
+  #   4. The installed library is copied to a convenience library without the name mangling from
+  #      (1) (if it's a shared library).
+  #
+  # You should use this function as you would `add_library()`.
+  #
+  # DETAILS
+  #
+  # The Cap'n Proto C++ ABI stability policy assumes no two versions are ABI-compatible. To prevent
+  # accidental dynamic linkage to an ABI-incompatible version of the library, the project version
+  # (-dev metadata and all) is appended to the library target's output name. For example, target
+  # `kj` produces a library whose filename and soname are `libkj-0.7-dev.so`. Additionally, the
+  # autotools build produces a `libkj.so` convenience file (but with the full soname).
+  #
+  # The autotools build implements this name mangling with `libtool -release`. The CMake build can
+  # emulate its behavior by manually setting the OUTPUT_NAME property. Copying the file to a
+  # filename without the version appendage is accomplished with a technique I learned from reading
+  # the Xerces-C++ project's CMakeLists.txt, by Roger Leigh:
+  # http://svn.apache.org/viewvc/xerces/c/trunk/src/CMakeLists.txt?revision=1800911&view=markup#l1269
+
+  add_library(${target} ${ARGN})
+  # Forward all arguments directly to add_library.
+
+  add_library(CapnProto::${target} ALIAS ${target})
+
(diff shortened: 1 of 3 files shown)


Conan-Kudo:
The autotools one should be doing a symlink rather than a full identical copy.

As for CMake, it's actually already emulating the behavior that autotools does (that's why I set `VERSION` instead of `SOVERSION`). When you use `VERSION`, the ELF image and soname are set to the full thing.

That's why with the RPM dependencies, it uses the exact version in the generated dependencies:
```
[neal@linux ~]$ sudo dnf -q repoquery --requires capnproto.x86_64
capnproto-libs(x86-64) = 0.6.1-3.fc26
libc.so.6(GLIBC_2.14)(64bit)
libcapnp.so.0.6.1()(64bit)
libcapnpc.so.0.6.1()(64bit)
libgcc_s.so.1()(64bit)
libgcc_s.so.1(GCC_3.0)(64bit)
libkj.so.0.6.1()(64bit)
libm.so.6()(64bit)
libpthread.so.0()(64bit)
libpthread.so.0(GLIBC_2.2.5)(64bit)
libstdc++.so.6()(64bit)
libstdc++.so.6(CXXABI_1.3)(64bit)
libstdc++.so.6(GLIBCXX_3.4)(64bit)
libstdc++.so.6(GLIBCXX_3.4.18)(64bit)
rtld(GNU_HASH)
[neal@linux ~]$ sudo dnf -q repoquery --provides capnproto-libs.x86_64
capnproto-libs = 0.6.1-3.fc26
capnproto-libs(x86-64) = 0.6.1-3.fc26
libcapnp-json.so.0.6.1()(64bit)
libcapnp-rpc.so.0.6.1()(64bit)
libcapnp.so.0.6.1()(64bit)
libcapnpc.so.0.6.1()(64bit)
libkj-async.so.0.6.1()(64bit)
libkj-http.so.0.6.1()(64bit)
libkj-test.so.0.6.1()(64bit)
libkj.so.0.6.1()(64bit)
```

If I had used `SOVERSION` and set it up like an ABI, the symbol would be something like `libcapnp.so.0()(64bit)`.

So personally, I don't think there's a problem to solve here.

Conan-Kudo:
As for the rpath stuff, I really don't want that behavior. With the autotools build, I have to regenerate the autofoo to make that go away (it's coming from libtool, and it varies by distribution whether rpath info is set).

harrishancock:
@Conan-Kudo Thank you for reviewing. You're right about the versioning already being safe -- I was simply confused after reading too much CMake and libtool documentation. :(

A subtle clarification, though: your code actually is relying on the [SOVERSION](https://cmake.org/cmake/help/latest/prop_tgt/SOVERSION.html) property. From the CMake docs: "If only one of both [VERSION and SOVERSION] is specified the missing is assumed to have the same version number." That doesn't change the fact that there isn't really any problem to be solved, which I agree with.

After thinking about this, this is probably the wrong approach for #523. While this branch "improves" the CMake build outputs by making them match the autotools build, it's not a particularly useful improvement, and it makes the CMake code significantly uglier. I had originally thought it would be the path of least resistance, but modifying the autotools build to install CMake package config files is probably both more elegant and useful. Therefore, I'm going to close this and work on the other approach!

PR #529, opened 2017-07-31
initialize null builder's sizes.

lcy03406:
So they are not random junk numbers and we can safely cast the builder to reader and read the content.

### DIFF
--- a/c++/src/capnp/layout.h
+++ b/c++/src/capnp/layout.h
@@ -498,7 +498,9 @@ private:
 
 class StructBuilder: public kj::DisallowConstCopy {
 public:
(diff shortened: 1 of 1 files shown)


kentonv:
This technically would allow you to convert to a reader and then read the content (getting default values), but it still would not allow you to safely call getter methods directly on an uninitialized builder, since builders assume the invariant that the objects they point at are always big enough for the defined schema.

I don't think you should be relying on uninitialized builders. Treat an uninitialized Builder object like an uninitialized pointer -- the only thing you can safely do is overwrite it.

(Confusingly, uninitialized Readers are safe to use and various applications in fact rely on this... oh well.)

PR #530, opened 2017-08-04
Doc: fix misspelled annotation target

ryanofsky:


kentonv:
Not sure why I never saw this but #1439 ended up applying the same change.

PR #531, opened 2017-08-04
Add WebSocket support to HTTP library

kentonv:


c++/src/kj/compat/http.h
```diff
@@ -563,19 +565,32 @@ kj::Own<HttpClient> newHttpClient(HttpHeaderTable& responseHeaderTable, kj::Asyn
...
-                                kj::Maybe<WebSocket::MaskKeyGenerator&> maskKeyGenerator = nullptr);
+                                kj::Maybe<EntropySource&> maskEntropySource);
 // Create a new WebSocket on top of the given stream. It is assumed that the HTTP -> WebSocket
 // upgrade handshake has already occurred (or is not needed), and messages can immediately be
 // sent and received on the stream. Normally applications would not call this directly.
 //
-// `maskKeyGenerator` is optional, but if omitted, the WebSocket frames will not be masked. Refer
-// to RFC6455 to understand when masking is required.
+// `maskEntropySource` is used to generate cryptographically-random frame masks. If null, outgoing
+// frames will not be masked. Servers are not required to mask their outgoing frames, but clients
```
harrishancock: nit: update comment to "Servers are required NOT to mask their outgoing frames ..."

[The RFC says](https://tools.ietf.org/html/rfc6455#section-5.1):

> A server MUST NOT mask any frames that it sends to the client.

c++/src/kj/compat/http.h
```diff
@@ -518,11 +570,33 @@ kj::Own<HttpClient> newHttpClient(HttpHeaderTable& responseHeaderTable, kj::Asyn
...
 // Adapts an HttpClient to an HttpService and vice versa.
 
+kj::Own<WebSocket> newWebSocket(kj::Own<kj::AsyncIoStream> stream,
+                                kj::Maybe<EntropySource&> maskEntropySource);
+// Create a new WebSocket on top of the given stream. It is assumed that the HTTP -> WebSocket
+// upgrade handshake has already occurred (or is not needed), and messages can immediately be
+// sent and received on the stream. Normally applications would not call this directly.
+//
+// `maskEntropySource` is used to generate cryptographically-random frame masks. If null, outgoing
+// frames will not be masked. Servers are not required to mask their outgoing frames, but clients
```
harrishancock: nit: update comment to "Servers are required NOT to mask their outgoing frames ..."

[The RFC says](https://tools.ietf.org/html/rfc6455#section-5.1):

> A server MUST NOT mask any frames that it sends to the client.

c++/src/kj/compat/http.c++
```diff
@@ -1547,6 +1547,466 @@ private:
...
+    if (currentlySending) {
+      // There is a message-send in progress, so we cannot write to the stream now.
+      //
+      // Note: According to spec, if the server receives a second ping before responding to the
+      //   previous one, it can opt to respond only to the last ping. So we don't have to check if
+      //   queuedPong is already non-null.
+      queuedPong = kj::mv(payload);
+    } else KJ_IF_MAYBE(promise, sendingPong) {
+      // We're still sending a previous pong. Wait for it to finish before sending ours.
+      sendingPong = promise->then(kj::mvCapture(payload, [this](kj::Array<byte> payload) mutable {
```
kentonv: True, though that will change as soon as we switch to C++14 and can change the kj::mvCapture() into a native capture.

I wish there were a way that we could declare all promise continuations mutable, because const is never what you want for them, and it results in confusing error messages. So, often I write `mutable` indiscriminately to avoid confusion later.

harrishancock: Good point! I hadn't considered the future C++14 transition.

PR #533, opened 2017-08-12
RPC: Support abstract unix socket addresses on Linux

ohwgiles:
Since `kj::StringPtr` supports embedded `NULL`s, copy its entire data buffer to the `sockaddr_un` structure instead of using `strcpy`.

On Linux, this allows the use of the abstract socket namespace, where the first character in the path is `\0`

### DIFF
--- a/c++/src/kj/async-io-unix.c++
+++ b/c++/src/kj/async-io-unix.c++
@@ -471,8 +471,11 @@ public:
       KJ_REQUIRE(path.size() < sizeof(addr.unixDomain.sun_path),
                  "Unix domain socket address is too long.", str);
       result.addr.unixDomain.sun_family = AF_UNIX;
-      strcpy(result.addr.unixDomain.sun_path, path.cStr());
-      result.addrlen = offsetof(struct sockaddr_un, sun_path) + path.size() + 1;
+      memcpy(result.addr.unixDomain.sun_path, path.cStr(), path.size() + 1);
+      result.addrlen = offsetof(struct sockaddr_un, sun_path) + path.size();
+      // Linux-specific: abstract namespace for unix sockets
+      // According to unix(7), addrlen of an abstract socket should not include the NULL terminator
+      if (path[0] != '\0') result.addrlen++;
       auto array = kj::heapArrayBuilder<SocketAddress>(1);
       array.add(result);
       return array.finish();
--- a/c++/src/kj/async-io-unix.c++
+++ b/c++/src/kj/async-io-unix.c++
@@ -491,7 +491,9 @@ public:
                  "Unix domain socket address is too long.", str);
       result.addr.unixDomain.sun_family = AF_UNIX;
       result.addr.unixDomain.sun_path[0] = '\0';
-      memcpy(result.addr.unixDomain.sun_path + 1, path.cStr(), path.size());
+      // although not strictly required by Linux, also copy the trailing
+      // NULL terminator so that we can safely read it back in toString
+      memcpy(result.addr.unixDomain.sun_path + 1, path.cStr(), path.size() + 1);
       result.addrlen = offsetof(struct sockaddr_un, sun_path) + path.size() + 1;
       auto array = kj::heapArrayBuilder<SocketAddress>(1);
       array.add(result);
--- a/c++/src/kj/async-io-unix.c++
+++ b/c++/src/kj/async-io-unix.c++
@@ -449,7 +449,11 @@ public:
         return str('[', buffer, "]:", ntohs(addr.inet6.sin6_port));
       }
       case AF_UNIX: {
-        return str("unix:", addr.unixDomain.sun_path);
+        if (addr.unixDomain.sun_path[0] == '\0') {
+          return str("unix-abstract:", addr.unixDomain.sun_path + 1);
+        } else {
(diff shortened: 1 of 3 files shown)


kentonv:
Hmm. While this may allow calling code which is explicitly written for abstract unix sockets, I think that the spirit of this interface is that a program should be able to accept an address as a command-line parameter, or in a config file, etc. Passing a string containing `\0` is going to be problematic all over the place.

How about we define a new prefix, "unix-abstract:", instead of expecting a string that contains `\0`? The parsing code will then take care of the quirks of initializing `sockaddr_un`?

ohwgiles:
I did think about that option. The question then would be what to do when the StringPtr actually does begin with `unix:\0`. Currently, an abstract unix socket is actually created, and (almost) the correct length is passed, but the abstract socket name itself is a string of `NULL`s due to the use of `strcpy`. This actually works in practice, and you can open several abstract sockets as long as they have different lengths. I've even used this in a few projects where capnproto is embedded before I realized this issue exists. If we make a new prefix, this can be "fixed" by replacing the `str.size()` with a call to `strlen()`. This would be a subtle breaking change but it would be consistent with `unix:` **not** supporting the abstract namespace. In this case, I would prefer to add a check if the socket path begins with `\0` and throw an error with an instruction to use the new prefix instead.

Instead of `unix-abstract:`, another option would be to use the `@` convention (like `ss -lx`). `parseAddress` could first recognize `unix:@` and immediately replace the `@` with `\0`, then fall through to the regular unix socket case including the `memcpy` fix above. This would allow both easy use as argv/config file/env var and support any embedded-code expectation that a StringPtr beginning with `\0` should create an abstract socket, without really providing two separate ways to do so.

I slightly lean towards the second option, but it's not a strong preference. If you tell me which you prefer I'm happy to create a new patch.

kentonv:
Ah, I agree that the current behavior is bad. We should perhaps throw an exception if strlen() and str.size() disagree. The exception message can mention the `unix-abstract:` prefix so that people can quickly figure out the right thing to do.

The `@` prefix has a couple (admittedly obscure) problems:

1. Technically `@` is a perfectly valid character to have in a filename, so attempts to reference a socket in the current directory whose name started with `@` would mysteriously fail.
2. A program might chroot or set up a mount namespace to restrict its view of the filesystem, then assume that it's safe to open an attacker-specified unix socket since it would necessarily be under that filesystem, but not realize that a whole non-filesystem-based namespace of sockets exists.

So personally my preference is for `unix-abstract:`, since it makes things nice and explicit.

ohwgiles:
Thanks for the convincing points, nice and explicit it is. For now I didn't enclose it in `#if __linux__`, since it's a bit of code noise for what would probably amount to a different runtime error on non-Linux Unix platforms.

I couldn't reproduce the Travis failure. It passed after what should have been an unrelated commit, so it looks like a spurious failure. Is that a known issue?

kentonv:
Yeah that Travis failure has been popping up a lot lately, but seems to be non-deterministic. :/ I restarted the job.

c++/src/kj/async-io-unix.c++
```diff
@@ -478,6 +485,19 @@ public:
       return array.finish();
     }
 
+    if (str.startsWith("unix-abstract:")) {
+      StringPtr path = str.slice(strlen("unix-abstract:"));
+      KJ_REQUIRE(path.size() + 1 < sizeof(addr.unixDomain.sun_path),
+                 "Unix domain socket address is too long.", str);
+      result.addr.unixDomain.sun_family = AF_UNIX;
+      result.addr.unixDomain.sun_path[0] = '\0';
+      memcpy(result.addr.unixDomain.sun_path + 1, path.cStr(), path.size());
```
kentonv: Since you use memcpy() here rather than strcpy(), the NUL terminator doesn't get copied. I'm not sure if Linux requires the string to be terminated (since it knows the addrlen separately), but it `toString()` you appear to be expecting a NUL-terminated string. Probably to be save we should assume a NUL terminator is required?

It would be nice to have a test that actually binds to or connects to an abstract address, ideally in a way that proves resolution is not filesystem-dependent... could be accomplished with some chdir()s. If you aren't interested in writing this test, though, don't worry about it.

ohwgiles:
OK fair enough, I copied the NULL terminator too. I wanted to stick with memcpy over strcpy since, intentionally or not, abstract unix socket addresses seem to support embedded NULLs (no implied comment on the wisdom of doing so).

I also added a small test, not sure if that's what you had in mind. It seems untidy to leave the test chdir'd, but maybe it's irrelevant?

kentonv:
To handle errors, you should write the `chdir` like:

    KJ_SYSCALL(chdir("/tmp"));

I think it's important to chdir back. So maybe do this:

    int originalDirFd;
    KJ_SYSCALL(originalDirFd = open(".", O_RDONLY | O_DIRECTORY | O_CLOEXEC));
    KJ_DEFER(close(originalDirFd));
    KJ_SYSCALL(chdir("/tmp"));
    KJ_DEFER(KJ_SYSCALL(fchdir(originalDirFd)));

Sorry for all the back-and-forth.

ohwgiles:
Not at all, thanks for the pointers, that's much better.

kentonv:
Thanks!

PR #536, opened 2017-08-14
Add canonical format to capnp convert

zombiezen:


kentonv:
Nice!

You'll need to update the plausibility checks (as Travis caught). Reusing the "flat" checks should be fine; no need to check for canonical ordering.

zombiezen:
Done.

Is it okay that I'm not checking the input being canonical?

kentonv:
Good question. Arguably if someone specifies canonical input instead of flat input, they might want it to be verified. Maybe throw in right after the `SegmentArrayMessageBuilder` declaration:

    if (convertFrom == Format::CANONICAL) {
      KJ_REQUIRE(message.isCanonical());
    }

zombiezen:
Done.

kentonv:
Thanks!

PR #537, opened 2017-08-15
Clarify the RPC spec in some places

talex5:
Based on discussions in the Google group:

https://groups.google.com/forum/#!topic/capnproto/_1kBtRSC51s

### DIFF
--- a/c++/src/capnp/rpc.capnp
+++ b/c++/src/capnp/rpc.capnp
@@ -492,6 +492,9 @@ struct Return {
   # should always set this true.  This defaults true because if level 0 implementations forget to
   # set it they'll never notice (just silently leak caps), but if level >=1 implementations forget
   # to set it to false they'll quickly get errors.
+  #
+  # The receiver should act as if the sender had sent a release message with count=1 for each
+  # CapDescriptor in the original Call message.
 
   union {
     results @2 :Payload;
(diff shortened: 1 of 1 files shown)


c++/src/capnp/rpc.capnp
```diff
@@ -926,6 +933,11 @@ struct Payload {
 
   capTable @1 :List(CapDescriptor);
   # Descriptors corresponding to the cap pointers in `content`.
+  #
+  # Note that it is currently not possible to include a broken capability in a payload.
```
kentonv: Perhaps this comment belongs inside CapDescriptor?

talex5: OK, I've moved it.

c++/src/capnp/rpc.capnp
```diff
@@ -939,6 +946,11 @@ struct CapDescriptor {
   #
   # Keep in mind that `ExportIds` in a `CapDescriptor` are subject to reference counting.  See the
   # description of `ExportId`.
+  #
+  # Note that it is currently not possible to include a broken capability in the CapDescriptor
+  # table.  Instead, create a new export (`senderHosted`) for each broken capability and then
```
kentonv: s/senderHosted/senderPromise/, otherwise you can't resolve it.

talex5: Good point. My implementation exported it as hosted because it wasn't a promise, and had an exception to allow far-refs to resolve to exceptions. But far-refs only need to break when the connection is lost, not through Resolve messages. I've fixed my code and updated the PR.

talex5:
Updated as suggested.

kentonv:
Thanks!

PR #547, opened 2017-09-03
ubsan: cast to unsigned integer before performing left shift

dwrensha:
When I run the test suite with `clang++ -fsanitize=undefined`, I see some errors of the form `runtime error: left shift of negative value -15`. This change makes the errors go away.


kentonv:
Apparently this is undefined to account for machines that don't use two's complement. Sigh.

PR #548, opened 2017-09-03
buffers passed to PackedOutputStream::write() must be word-aligned

dwrensha:
Fixes an error in the test suite that was reported by ubsan, and adds documentation about `PackedOutputStream`'s alignment assumptions.

As far as I can tell, the error cannot occur through normal usage of `serialize_package::write_message()`.


kentonv:
Thanks!

PR #549, opened 2017-09-03
avoid passing invalid reference to Array<T> constructor

dwrensha:
ubsan complains about the invalid reference even if it is unused:

```
../src/kj/parse/../array.h:453:59: runtime error: reference binding to misaligned address 0x7fa57ffffffa for type 'const kj::ArrayDisposer', which requires 8 byte alignment
```

### DIFF
--- a/c++/src/kj/array.h
+++ b/c++/src/kj/array.h
@@ -450,11 +450,17 @@ public:
     // check might catch bugs.  Probably people should use Vector if they want to build arrays
     // without knowing the final size in advance.
     KJ_IREQUIRE(pos == endPtr, "ArrayBuilder::finish() called prematurely.");
-    Array<T> result(reinterpret_cast<T*>(ptr), pos - ptr, *disposer);
+    const ptrdiff_t size = pos - ptr;
+    T* start = reinterpret_cast<T*>(ptr);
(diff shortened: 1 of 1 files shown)


kentonv:
Hmm, maybe `disposer` should be initialized to null in the ArrayBuilder constructor instead?

dwrensha:
ubsan would similarly complain about a null reference:

```
 runtime error: reference binding to null pointer of type 'const kj::ArrayDisposer'
```

It's probably a good idea to initialize `ArrayBuilder.disposer` to null rather than keeping it uninitialized, but we still need to construct the array without forming an invalid (or null) reference.

Another way to avoid the error would be to use a new `Array` construct that takes a pointer to an `ArrayDisposer`, rather than a reference.

kentonv:
Ah, I see. `Array::disposer` is also a pointer but the constructor takes a reference, and it's specifically the reference which is required to refer to a real object. OK, this looks like the right answer, then. Thanks!

kentonv:
Turns out this change introduced a memory leak: Zero-sized arrays are allocated all the time, and are now not being deleted. This caused an Oasis outage last night as the master instance does not seem to gracefully handle running out of memory.

PR #550, opened 2017-09-05
Implement TLS option to verify client certificates.

kentonv:
Use case for me is Cloudflare's "Authenticated Origin Pulls", in which Cloudflare authenticates itself to the origin server using client certificates. Thus the origin server can ensure that it only accepts traffic through Cloudflare, without resorting to IP whitelists.


harrishancock:
Sorry, just saw this! Will review first thing tomorrow.

kentonv:
No rush on this -- it turns out I can't actually make use of it right now anyway.

PR #555, opened 2017-09-12
Extend kj::Network interface for easy SSRF protection

kentonv:
This is an important security feature for many applications which is really difficult to implement on top of the KJ interfaces prior to this change.


harrishancock:
Working on this, should be done in the afternoon.

PR #556, opened 2017-09-17
Make win32Socketpair() more secure

kentonv:
Make win32Socketpair more secure.

In theory if someone guessed the port and connected at just the right time they could break the socket pair. Check peer address to verify this hasn't happened.

(Note that we don't use win32Socketpair() anywhere important today -- mostly just in tests. This is just something that occurred to me white touching related code.)


kentonv:
Ironically the `goto` was especially needed because the existing code (not written by me) was written in a style that used `break`s inside a `for(;;)` as a way of implementing `goto` without actually writing `goto`. IMO these tricks are much worse than just writing `goto`.

PR #557, opened 2017-09-21
Implement HttpClient that automatically manages connections. 

kentonv:
There are actually two new client types:
- One which always connects to a given NetworkAddress, but will automatically manage a pool of reusable connections.
- One which looks up the remote address based on the URL it is given, and manages a pool of connections for each host.

The latter of these two is a "true HTTP client library".

### DIFF
--- a/c++/src/kj/compat/http-test.c++
+++ b/c++/src/kj/compat/http-test.c++
@@ -2355,6 +2355,344 @@ KJ_TEST("newHttpService from HttpClient WebSockets disconnect") {
 
 // -----------------------------------------------------------------------------
 
+// TODO(now): Test NetworkAddressHttpClient:
+// - Serial requests open only one connection.
+// - Parallel requests open multiple connections.
+// - Connections time out.
+
+class CountingIoStream final: public kj::AsyncIoStream {
+  // An AsyncIoStream which waits for a promise to resolve then forwards all calls to the promised
+  // stream.
+  //
+  // TODO(cleanup): Make this more broadly available.
+
+public:
+  CountingIoStream(kj::Own<kj::AsyncIoStream> inner, uint& count)
+      : inner(kj::mv(inner)), count(count) {}
+  ~CountingIoStream() noexcept(false) {
+    --count;
+  }
+
+  kj::Promise<size_t> read(void* buffer, size_t minBytes, size_t maxBytes) override {
+    return inner->read(buffer, minBytes, maxBytes);
+  }
+  kj::Promise<size_t> tryRead(void* buffer, size_t minBytes, size_t maxBytes) override {
+    return inner->tryRead(buffer, minBytes, maxBytes);
+  }
+  kj::Maybe<uint64_t> tryGetLength() override {
+    return inner->tryGetLength();;
+  }
+  kj::Promise<uint64_t> pumpTo(kj::AsyncOutputStream& output, uint64_t amount) override {
+    return inner->pumpTo(output, amount);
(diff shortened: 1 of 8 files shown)


c++/src/kj/compat/http-test.c++
```diff
@@ -2355,6 +2355,344 @@ KJ_TEST("newHttpService from HttpClient WebSockets disconnect") {
...
 // -----------------------------------------------------------------------------
 
+// TODO(now): Test NetworkAddressHttpClient:
+// - Serial requests open only one connection.
+// - Parallel requests open multiple connections.
+// - Connections time out.
+
+class CountingIoStream final: public kj::AsyncIoStream {
+  // An AsyncIoStream which waits for a promise to resolve then forwards all calls to the promised
+  // stream.
```
harrishancock: Inaccurate comment, looks like copy-and-paste error.

c++/src/kj/compat/http-test.c++
```diff
@@ -2355,6 +2355,344 @@ KJ_TEST("newHttpService from HttpClient WebSockets disconnect") {
 
 // -----------------------------------------------------------------------------
 
+// TODO(now): Test NetworkAddressHttpClient:
+// - Serial requests open only one connection.
+// - Parallel requests open multiple connections.
+// - Connections time out.
```
harrishancock: Remove TODO? These look implemented to me.

c++/src/kj/compat/http-test.c++
```diff
@@ -2355,6 +2355,344 @@ KJ_TEST("newHttpService from HttpClient WebSockets disconnect") {
...
+  KJ_EXPECT(tlsAddrCount == 1);
+
+  // Try first host again, no change in connection count.
+  doRequest(false, port1).wait(io.waitScope);
+  KJ_EXPECT(count == 2);
+  KJ_EXPECT(tlsCount == 1);
+  KJ_EXPECT(addrCount == 2);
+  KJ_EXPECT(tlsAddrCount == 1);
+
+  // Multipre requests in parallel forces more connections to that host.
```
harrishancock: /Multipre/Multiple/

c++/src/kj/compat/http-test.c++
```diff
@@ -2355,6 +2355,344 @@ KJ_TEST("newHttpService from HttpClient WebSockets disconnect") {
...
+  promise1.wait(io.waitScope);
+  promise2.wait(io.waitScope);
+  KJ_EXPECT(count == 3);
+  KJ_EXPECT(tlsCount == 1);
+  KJ_EXPECT(addrCount == 2);
+  KJ_EXPECT(tlsAddrCount == 1);
+
+  // Let everything expire.
+  clientTimer.advanceTo(clientTimer.now() + clientSettings.idleTimout * 2);
+  kj::Promise<void>(kj::NEVER_DONE).poll(io.waitScope);
```
harrishancock: `io.waitScope.poll()`? This is the only instance of `Promise::poll()` in these tests, so perhaps this is intentional and I just don't follow some subtle difference in their use.

kentonv: Oops. I implemented Promise::poll() first, then realized that I only ever used it on a NEVER_DONE promise, so added WaitScope::poll() and switched to that, but I guess I forgot to update all users.

Promise::poll() turns out not to be useful in this test, but I know I've written lots of tests in the past that could have used it, so I decided to leave it in for future use.

PR #565, opened 2017-09-24
Add Own<T>::attach() which is much like Promise<T>::attach()

kentonv:
This lets us delete a bunch of code in http.c++.

It's also really useful when you have a class that takes its dependencies by reference (probably to allow you to stack-allocate them), but you need a single Own<T> for it and its dependencies, e.g. to return from some factory function.

### DIFF
--- a/c++/src/kj/memory-test.c++
+++ b/c++/src/kj/memory-test.c++
@@ -65,6 +65,47 @@ TEST(Memory, AssignNested) {
   EXPECT_TRUE(destroyed1 && destroyed2);
 }
 
+struct DestructionOrderRecorder {
+  DestructionOrderRecorder(uint& counter, uint& recordTo)
+      : counter(counter), recordTo(recordTo) {}
+  ~DestructionOrderRecorder() {
+    recordTo = ++counter;
+  }
+
+  uint& counter;
+  uint& recordTo;
+};
+
+TEST(Memory, Attach) {
+  uint counter = 0;
+  uint destroyed1 = 0;
+  uint destroyed2 = 0;
+  uint destroyed3 = 0;
+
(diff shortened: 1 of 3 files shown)


c++/src/kj/memory.h
```diff
@@ -150,6 +150,12 @@ class Own {
     return *this;
   }
 
+  template <typename... Attachments>
+  Own<T> attach(Attachments&&... attachments);
+  // Returns an Own<T> which points to the same object but which also ensures that all values
+  // passed to `attachments` remain alive until after this object is destroyed. Normally
+  // `attachments` are other Own<?>s pointing to objects that this one depends on.
```
harrishancock: Maybe clarify that the order of destruction of `attachments` is the order in which they are passed? The comment in the implementation already makes this clear, but it'd be good to document at the interface definition, I think.

c++/src/kj/memory-test.c++
```diff
@@ -65,6 +65,47 @@ TEST(Memory, AssignNested) {
...
+  KJ_EXPECT(destroyed2 == 0);
+  KJ_EXPECT(destroyed3 == 0);
+
+  combined = nullptr;
+
+  KJ_EXPECT(destroyed1 == 1, destroyed1);
+  KJ_EXPECT(destroyed2 == 2, destroyed2);
+  KJ_EXPECT(destroyed3 == 3, destroyed3);
+}
+
```
harrishancock: I'd recommend adding a `ptr.attach(foo).attach(bar)...` test case as well. Right now this only tests the case when one single `DisposableOwnedBundle` is owned by an `Own<T>` -- it'd be good to verify that it works more deeply nested, as well.

PR #567, opened 2017-10-12
Memcpy null src

ecatmur:
It is invalid to pass null as a pointer argument to memcpy/memcmp/memset, even if the count argument is zero:

> Where an argument declared as size_t n specifies the length of the array for a
function, n can have the value zero on a call to that function. Unless explicitly stated
otherwise in the description of a particular function in this subclause, pointer arguments
on such a call shall still have valid values [...]

Detected by -fsanitize=nonnull-attribute.

Additionally, when constructing StringPtr from String, ensure that content does not bceome null. Test.

### DIFF
--- a/c++/src/capnp/layout.c++
+++ b/c++/src/capnp/layout.c++
@@ -371,15 +371,15 @@ struct WireHelpers {
 #endif
 
   static KJ_ALWAYS_INLINE(void zeroMemory(byte* ptr, ByteCount32 count)) {
-    memset(ptr, 0, unbound(count / BYTES));
+    if (count != 0u) memset(ptr, 0, unbound(count / BYTES));
   }
 
   static KJ_ALWAYS_INLINE(void zeroMemory(word* ptr, WordCountN<29> count)) {
-    memset(ptr, 0, unbound(count * BYTES_PER_WORD / BYTES));
+    if (count != 0u) memset(ptr, 0, unbound(count * BYTES_PER_WORD / BYTES));
   }
 
   static KJ_ALWAYS_INLINE(void zeroMemory(WirePointer* ptr, WirePointerCountN<29> count)) {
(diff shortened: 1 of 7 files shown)


ecatmur:
note: I combined these two commits because they're detected by the same sanitizer, so testing gets confusing otherwise (especially since the `StringPtr(String(nullptr))` issue shows up in the OneOf tests). Happy to separate these commits if you'd prefer.

kentonv:
Does mem*() on any real platform have a problem with the pointer being null when the size is zero?

It seems sad to add all these extra branches, slowing down the code, just to work around a pedantic warning.

ecatmur:
I'm not aware of any production (as opposed to debug) libc that behaves badly when the memcpy etc. symbols are invoked with 0 size and null pointer(s), but it's possible that one exists for a minority platform somewhere.

A more likely issue with production builds would be the optimizer deducing from the call that the pointer arguments are non-null and messing up the calling code.

As far as the branching goes, memcpy is pretty branch-heavy anyway, and there's a win in the size zero case from avoiding the jump and fall through a bunch of internal branches, so it's not totally clear that this would worsen performance, if enough of these size zero calls occur. The branch might also help the optimizer no-op e.g. empty container copy ctors.

If you're unhappy about the impact on source code, I'd be happy to code a patch replacing with <algorithm> equivalents e.g. copy_n<char> is compiled to guarded calls to memcpy (gcc, clang, assuming appropriate use of restrict) or memmove (icc, MSVC).

kentonv:
Ah, good point about the optimizer -- although I suspect such an optimization will never actually be enabled since it would break a lot of existing code, I am sufficiently scared of the possibility after having hit "optimizer changed the meaning of my code" problems in the past. :)

OK, accepting this, thanks!

kentonv:
This change broke the build with CAPNP_DEBUG_TYPES. Fixed in 7c8870e. Guess we should add that to the CI.

PR #568, opened 2017-10-12
Variable-length arrays must have positive bound.

ecatmur:
>  If the size is an expression that is not an integer constant expression [...], each time it is evaluated it shall have a value greater than zero.

Detected by -fsanitize=vla-bound


kentonv:
Thanks!

PR #569, opened 2017-10-12
Don't bind Header reference to a null pointer.

ecatmur:
If recvData is empty, recvData.begin() is a null pointer and binding a Header reference to it is invalid.

Detected by -fsanitize=null.


kentonv:
Thanks!

PR #570, opened 2017-10-12
Don't read past the end of the base64 decode out buffer.

ecatmur:
If we finish decoding in step_a state, there is no current output character, so reading *plainchar will either be an uninitialized read or (if the output buffer is minimally sized) a past-the-end read.

Detected by -fsanitize=address.

### DIFF
--- a/c++/src/kj/encoding-test.c++
+++ b/c++/src/kj/encoding-test.c++
@@ -269,9 +269,21 @@ KJ_TEST("C escape encoding/decoding") {
 
 KJ_TEST("base64 encoding/decoding") {
   {
+    auto encoded = encodeBase64(StringPtr("").asBytes(), false);
+    KJ_EXPECT(encoded == "", encoded, encoded.size());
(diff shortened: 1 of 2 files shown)


kentonv:
Thanks!

Is this bug also present in libb64, from which this code is derived?

kentonv:
Oops, should have waited for the CI. This change sends the MSVC build into an infinite loop. I've reverted it.

ecatmur:
Ah, that's interesting. The test on empty input was resulting in a write to the pointer returned by `operator new(0)`, which is a non-null, distinct, non-writable pointer. MSVC was crashing in the later delete of that pointer. Interesting that asan didn't detect it.

Fixed by conditioning the write on decoder state, similarly to the original patch.

ecatmur:
OK if I submit a new PR? https://github.com/capnproto/capnproto/compare/master...ecatmur:base64-invalid-read

ecatmur:
Yes, libb64 has the same issues. I've emailed the author a patch.

kentonv:
Yep, please submit new PR. Thanks!

PR #571, opened 2017-10-12
Include <wincrypt.h> directly for WIN32_LEAN_AND_MEAN

ecatmur:
Visual Studio 2017 (14.11.25503, 19.11.25508.2) needs `#include <wincrypt.h>` for CryptGenRandom etc.

### DIFF
--- a/c++/src/capnp/compiler/parser.c++
+++ b/c++/src/capnp/compiler/parser.c++
@@ -31,6 +31,7 @@
 #include <fcntl.h>
 
 #if _WIN32
+#define WIN32_LEAN_AND_MEAN
 #include <windows.h>
 #include <wincrypt.h>
 #undef VOID
--- a/c++/src/capnp/compiler/parser.c++
+++ b/c++/src/capnp/compiler/parser.c++
@@ -32,6 +32,7 @@
 
 #if _WIN32
 #include <windows.h>
+#include <wincrypt.h>
 #undef VOID
 #endif
 


ecatmur:
ah, this is because we use WIN32_LEAN_AND_MEAN, which excludes the crypto API from windows.h.

kentonv:
Do you mean that you're compiling this code in some no-standard way with WIN32_LEAN_AND_MEAN always defined? We have tested in MSVC 2017 and never had a problem here.

kentonv:
Seems your latest commit breaks the CI.

ecatmur:
Oops, confused myself there. Unfortunately windows.h (or some indeterminate set of lesser headers; let's not go there) is required; wincrypt.h is not self-contained.

Yes, we use WIN32_LEAN_AND_MEAN to build our own code and prefer to build everything with the same compile flags where possible. WIN32_LEAN_AND_MEAN is non-standard, admitted, but it is a documented way to speed up builds (https://msdn.microsoft.com/en-us/library/windows/desktop/aa383745(v=vs.85).aspx "Faster Builds with Smaller Header Files"). From my point of view I won't be offended in the slightest if you reject this PR as I can keep the patch in our local branch or disable the flag for the capnproto build. Still, it doesn't hurt to have the wincrypt.h include and it could save time for anyone else with the same issue.

kentonv:
We actually explicitly #define WIN32_LEAN_AND_MEAN in most places where we include windows.h, but apparently I forgot in this file. Care to add it? Will help make sure we don't break your build again with later changes.

kentonv:
(Every time I type WIN32_LEAN_AND_MEAN I die a little inside...)

kentonv:
Oh geez, I forget about this, sorry. Looks great.

PR #577, opened 2017-10-19
Base64 decoder: invalid write/read

ecatmur:


kentonv:
Thanks, and sorry for the delay.

PR #581, opened 2017-10-27
Close idle HTTP connections on server drain.

kentonv:
Previously we only closed immediately after a response finished.

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -3295,8 +3295,9 @@ public:
     if (!firstRequest) {
       // For requests after the first, require that the first byte arrive before the pipeline
       // timeout, otherwise treat it like the connection was simply closed.
-      firstByte = firstByte.exclusiveJoin(
-          server.timer.afterDelay(server.settings.pipelineTimeout)
+      firstByte = firstByte
+          .exclusiveJoin(server.timer.afterDelay(server.settings.pipelineTimeout)
+          .exclusiveJoin(server.onDrain.addBranch())
           .then([this]() -> bool {
         timedOut = true;
         return false;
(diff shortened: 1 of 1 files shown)


c++/src/kj/compat/http.c++
```diff
@@ -3295,8 +3295,9 @@ public:
     if (!firstRequest) {
       // For requests after the first, require that the first byte arrive before the pipeline
       // timeout, otherwise treat it like the connection was simply closed.
-      firstByte = firstByte.exclusiveJoin(
-          server.timer.afterDelay(server.settings.pipelineTimeout)
+      firstByte = firstByte
+          .exclusiveJoin(server.timer.afterDelay(server.settings.pipelineTimeout)
```
harrishancock: I would have expected a second parenthesis after ...headerTimeout). (Maybe a nitpick, as it doesn't really matter what the topology of the exclusive-join graph looks like.)

c++/src/kj/compat/http.c++
```diff
@@ -3328,8 +3329,9 @@ public:
 
     if (firstRequest) {
       // On the first request, the header timeout starts ticking immediately upon request opening.
-      receivedHeaders = receivedHeaders.exclusiveJoin(
-          server.timer.afterDelay(server.settings.headerTimeout)
+      receivedHeaders = receivedHeaders
+          .exclusiveJoin(server.timer.afterDelay(server.settings.headerTimeout)
```
harrishancock: Ditto.

c++/src/kj/compat/http.c++
kentonv: Actually it does matter here, because there's also a .then() which specifically applies to the join of the timeout and onDrain, but not firstByte.

kentonv: That said I agree the indentation is confusing and could be improved.

kentonv:
Whoops, accidentally merged this due to branch confusion on my client. Addressed your readability comment in 3174079c.

harrishancock:
Looks good, that makes it a lot clearer!

PR #585, opened 2017-11-14
Extend UnixEventPort with the ability to listen for subprocess exit.

kentonv:
(Later, we should add a nice subprocess API around this, maybe even one that can work on Windows too...)

### DIFF
--- a/c++/src/kj/async-unix.c++
+++ b/c++/src/kj/async-unix.c++
@@ -157,6 +157,9 @@ void UnixEventPort::ChildSet::checkExits() {
 }
 
 Promise<int> UnixEventPort::onChildExit(Maybe<pid_t>& pid) {
+  KJ_REQUIRE(capturedChildExit,
+      "must call UnixEventPort::captureChildExit() to use onChildExit().");
+
   ChildSet* cs;
   KJ_IF_MAYBE(c, childSet) {
     cs = *c;
--- a/c++/src/kj/async-unix.c++
+++ b/c++/src/kj/async-unix.c++
@@ -30,6 +30,8 @@
 #include <limits>
 #include <chrono>
 #include <pthread.h>
+#include <map>
+#include <sys/wait.h>
 
 #if KJ_USE_EPOLL
 #include <unistd.h>
@@ -57,6 +59,8 @@ namespace {
 
 int reservedSignal = SIGUSR1;
 bool tooLateToSetReserved = false;
+bool capturedChildExit = false;
+bool threadClaimedChildExits = false;
 
 struct SignalCapture {
   sigjmp_buf jumpTo;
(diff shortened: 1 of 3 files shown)


c++/src/kj/async-unix.c++
```diff
@@ -109,6 +113,73 @@ pthread_once_t registerReservedSignalOnce = PTHREAD_ONCE_INIT;
...
+            "`pid` must be non-null at the time `onChildExit()` is called")),
+        pidRef(pidRef), fulfiller(fulfiller) {
+    KJ_REQUIRE(childSet.waiters.insert(std::make_pair(pid, this)).second,
+        "already called onChildExit() for this pid");
+  }
+
+  ~ChildExitPromiseAdapter() noexcept(false) {
+    childSet.waiters.erase(pid);
+  }
+
```
harrishancock: Total nit: KJ_DISALLOW_COPY? I know it's just an adapter...

c++/src/kj/async-unix.h
```diff
@@ -99,15 +99,49 @@ class UnixEventPort: public EventPort {
...
+  //   child exit is signaled to the process via SIGCHLD, and Unix does not allow the program to
+  //   control which thread receives the signal. (We may fix this in the future by automatically
+  //   coordinating between threads when multiple threads are expecting child exits.)
+  // WARNING 2: If any UnixEventPort in the process is currently waiting for onChildExit(), then
+  //   *only* that port's thread can safely wait on child processes, even synchronously. This is
+  //   because the thread which used onChildExit() uses wait() to reap children, without specifying
+  //   which child, and therefore it may inadvertently reap children created by other threads.
+
+  static void captureChildExit();
+  // Arranges for child process exit to be captured and handled via UnixEventPort, so than you may
```
harrishancock: /than/that/

c++/src/kj/async-unix.c++
```diff
@@ -109,6 +113,73 @@ pthread_once_t registerReservedSignalOnce = PTHREAD_ONCE_INIT;
...
+Promise<int> UnixEventPort::onChildExit(Maybe<pid_t>& pid) {
+  ChildSet* cs;
+  KJ_IF_MAYBE(c, childSet) {
+    cs = *c;
+  } else {
+    // In theory we should do an atomic compare-and-swap on threadClaimedChildExits, but this is
+    // for debug purposes only so it's not a big deal.
+    KJ_REQUIRE(!threadClaimedChildExits,
+        "only one UnixEvertPort per process may listen for child exits");
```
harrishancock: Should we also `KJ_REQUIRE(capturedChildExit)` here? By my reading, it appears that one could `captureSignal(SIGCHLD)` and then call `onChildExit(...)` and it would work, contradicting the doc comment for `onChildExit()`.

c++/src/kj/async-unix.c++
kentonv: Yeah I don't really care about these details in private classes. :)

c++/src/kj/async-unix.c++
kentonv: indeed

harrishancock:
:+1:

kentonv:
Mac failure is spurious... merging.

PR #592, opened 2017-11-28
Fix wrong poll/epoll timeout after EINTR.

### LINKED ISSUE #580: If poll() or epoll_wait() fails with EINTR, it is restarted with the wrong timeout
The timeout needs to be recalculated since we don't know how much time passed.

kentonv:
Fixes #580


kentonv:
@harrishancock CI tests Mac OS, which is non-epoll.

harrishancock:
Great!

PR #594, opened 2017-12-01
Make decodeBase64() report errors

harrishancock:
This change modifies decodeBase64() to report errors as required by the WHATWG HTML spec's atob() JavaScript function. Notably, it reports errors for non-whitespace characters outside of the valid base64 character range ([+/0-9A-Za-z=]), and performs sanity checks on padding and input length (as calculated after removing whitespace).

I took care to keep the algorithm single-pass, and to support streaming via multiple calls of base64_decode_block(), though we don't currently expose that functionality.


harrishancock:
Reopening under different branch.

PR #595, opened 2017-12-01
Make decodeBase64() report errors

harrishancock:
This change modifies decodeBase64() to report errors as required by the WHATWG HTML spec's atob() JavaScript function. Notably, it reports errors for non-whitespace characters outside of the valid base64 character range ([+/0-9A-Za-z=]), and performs sanity checks on padding and input length (as calculated after removing whitespace).

I took care to keep the algorithm single-pass, and to support streaming via multiple calls of base64_decode_block(), though we don't currently expose that functionality.


harrishancock:
Rebased on master, optimized, and added a `stopOnError` early-bailout flag.

The initial version of error-checking performed all logic in the two layers that bracket `base64_decode_block()`: the calling function `decodeBase64()` and the fragment decoding function `base64_decode_value()`. This version inverts that and performs all logic in `base64_decode_block()`. This made it a little harder to reason about (for me), but ultimately produced a relatively clean, and more optimal, result.

I expanded the fragment decoding table to cover the entire range of an `unsigned char` which helps avoid a branch (though I added plenty more).

I also added a `stopOnError` flag to `decodeBase64()` so that the algorithm can terminate early if the caller knows they're going to discard output on errors (like JS' atob()). As mentioned in the comments, it could be an improvement to make this flag a compile-time function template parameter instead of runtime.

I'm a little hesitant about adding that flag, since it takes up valuable real estate in `decodeBase64()`'s function parameter list. Another option might be to add a `tryDecodeBase64() -> Maybe<Array<byte>>` that early-terminates. OTOH, adding a new function isn't necessarily a wise interface design choice, either.

kentonv:
Nice, this looks a lot cleaner.

I'd rather not add the `stopOnErrors` flag unless we see it being necessary. Error cases should be rare, so optimizing them seems unnecessary, especially if it makes the common case slower (with more branches).

I think if you get rid of the `stopOnErorrs` stuff, the error handling actually becomes branchless. The compiler ought to be able to optimize "x = x || (trivial comparison)" to "x |= (trivial comparison)" -- no branches.

harrishancock:
Sounds good, I'll drop the `stopOnError` commit and re-push.

As an aside, a couple of the error checks have side effects -- `ERROR_IF(fragment < -2 || (fragment == -2 && ++state_in->nPaddingBytesSeen > 1))`, for instance -- which would make the `x = x || y` -> `x |= y` optimization incorrect due to different short-circuiting behavior (the latter would unconditionally increment `nPaddingBytesSeen`, which the former does not).

PR #600, opened 2017-12-13
Add support for Own<void>.

kentonv:
Any `Own<T>` can be converted to `Own<void>` (or `Own<const void>`). The main purpose of this is to give someone a way to invoke an object's disposer without them having to know anything about the type.


c++/src/kj/memory-test.c++
```diff
@@ -136,6 +136,92 @@ TEST(Memory, AttachNested) {
...
+
+    Own<void> voidPtr = kj::mv(basePtr);
+    KJ_EXPECT(voidPtr.get() == static_cast<void*>(addr));
+
+    KJ_EXPECT(!destructorCalled);
+    voidPtr = nullptr;
+    KJ_EXPECT(destructorCalled);
+  }
+}
+
```
harrishancock: I'd recommend adding some trivial test for the `Own<const void>` case, just to prove it compiles.

PR #603, opened 2017-12-18
setNonblocking on android

lcy03406:
This pr is mainly to fix the problem of connectImpl() on Android platform.

When connecting to a valid but unavailable address (eg. "192.168.200.200 while the device has no access to .200 subnet), the Android app freezes 2 minutes before throwing the exception. This cannot be canceled by timeoutAfter(), because the 2 minutes is to waiting the connect() syscall.

I believe in kj-async all fds should be nonblocking if posssible, so it may be better to add setNonblocking() in somewhere low level, rather then wrapConnectingSocketFd(). So I call setNonblocking() as soon as socket() returns, similar as on other platforms.

I also added setCloseOnExec() to sockets, and also setNonblocking() and setCloseOnExec() to pipes. However I didn't test it yet.

### DIFF
--- a/c++/src/kj/async-io-unix.c++
+++ b/c++/src/kj/async-io-unix.c++
@@ -514,8 +514,12 @@ public:
     int result;
 #if __linux__ && !__BIONIC__
     type |= SOCK_NONBLOCK | SOCK_CLOEXEC;
-#endif
     KJ_SYSCALL(result = ::socket(addr.generic.sa_family, type, 0));
+#else
+    KJ_SYSCALL(result = ::socket(addr.generic.sa_family, type, 0));
+    setNonblocking(result);
+	setCloseOnExec(result);
+#endif
 
     if (isStream && (addr.generic.sa_family == AF_INET ||
                      addr.generic.sa_family == AF_INET6)) {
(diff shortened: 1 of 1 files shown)


lcy03406:
wait. I will do more testing if you want to merge.

kentonv:
Wow, this is definitely a bug in wrapConnectingSocketFd(). In fact, there are multiple bugs: if connect() fails, and the TAKE_OWNERSHIP flag was used, it looks like the fd is leaked!

The problem is that the flags aren't interpreted until the AsyncStreamFd is created. We probably need to move that line to the beginning of the function.

I don't think we should modify all the places where we create FDs. I think only wrapConnectingSocketFd() is broken and we should fix that specifically. Usually, AsyncStreamFd's constructor changes the FD to non-blocking, but the constructor runs too late in wrapConnectingSocketFd()'s case.

kentonv:
I've created #604 to more directly address the bugs. Let me know if it fixes things for you.

kentonv:
I believe with #604 merged, this problem should be fixed, so I'm closing this PR. Thanks for reporting this!

PR #606, opened 2017-12-21
Fix capnp_generate_cpp usage of --src-prefix

harrishancock:
This PR:

- Makes `capnp_generate_cpp()` import the src-prefix by default instead of `CMAKE_CURRENT_SOURCE_DIR`.
- Sets corrected values for `CAPNPC_OUTPUT_DIR` and `CAPNPC_SRC_PREFIX` in `c++/src/capnp/CMakeLists.txt`.
- Removes `FindCapnProto.cmake` (rather than also fix its copy).
- Fixes (well... works around) a race condition that caused test.capnp and friends to be compiled twice in a parallel make build.

@pqu, would you mind taking a look at this if/when you have a chance? For context, this fixes an issue that @kentonv discovered in #605.

Note that I retired `FindCapnProto.cmake`, since its functionality is now duplicated, and it contains code that will otherwise atrophy. What do you think? (If you prefer to keep it, I'll remove that commit and apply the fix to that file's copy of `capnp_generate_cpp`.)

@kentonv, the generated capnp command is now:

```
./capnp compile \
    -o /home/harris/src/capnproto/c++/build/src/capnp/capnpc-c++:/home/harris/src/capnproto/c++/build/src/capnp/test_capnp \
    --src-prefix /home/harris/src/capnproto/c++/src \
    -I /home/harris/src/capnproto/c++/src \
    -I /home/harris/src/capnproto/c++/src/capnp/.. \
    /home/harris/src/capnproto/c++/src/capnp/test.capnp
```

Note the duplicate -I flags. It turns out to be difficult to prevent that from happening, without hacks -- I explain why in a commit message and code comment -- but at least they now both resolve to the source prefix.


harrishancock:
Thanks, @pqu, I appreciate it! @kentonv, this should be good to merge now.

PR #607, opened 2017-12-21
KJ_SWITCH_ONEOF should parenthesize its argument

harrishancock:
This allows us to deal with Maybe<OneOf<...>> variables properly inside a KJ_IF_MAYBE block.

### DIFF
--- a/c++/src/kj/one-of-test.c++
+++ b/c++/src/kj/one-of-test.c++
@@ -151,4 +151,23 @@ TEST(OneOf, Switch) {
   }
 }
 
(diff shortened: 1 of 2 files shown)


harrishancock:
Linux/clang failed to build because one of the samples crashed with this change. Bizarre! Will debug tomorrow.

harrishancock:
I guess it was just some transient connect() failure on the CI box.

kentonv:
Ah, yeah, I've seen that a few times...

PR #612, opened 2018-01-05
Various tweaks needed by Sandstorm HTTP changes

kentonv:
These are three tweaks I made that are needed by: https://github.com/sandstorm-io/sandstorm/pull/3007

### DIFF
--- a/c++/src/kj/compat/tls.c++
+++ b/c++/src/kj/compat/tls.c++
@@ -30,6 +30,7 @@
 #include <openssl/x509v3.h>
 #include <openssl/evp.h>
 #include <openssl/conf.h>
+#include <openssl/ssl.h>
 #include <openssl/tls1.h>
 #include <kj/debug.h>
 #include <kj/vector.h>
@@ -489,6 +490,13 @@ TlsContext::Options::Options()
 // Classic DH is arguably obsolete and will only become more so as time passes, so perhaps we'll
 // never bother.
 
+struct TlsContext::SniCallback {
+  // struct SniCallback exists only so that callback() can be declared in the .c++ file, since it
+  // references OpenSSL types.
+
+  static int callback(SSL* ssl, int* ad, void* arg);
+};
+
 TlsContext::TlsContext(Options options) {
   ensureOpenSslInitialized();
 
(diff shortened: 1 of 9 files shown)


kentonv:
Well, I guess deprecating RealmGateway is not "needed", but Sandstorm will no longer use it.

c++/src/kj/compat/tls.c++
```diff
@@ -267,8 +268,9 @@ private:
           return readBuffer.whenReady().then(kj::mvCapture(func,
               [this](Func&& func) mutable { return sslCall(kj::fwd<Func>(func)); }));
         case SSL_ERROR_WANT_WRITE:
+          KJ_DBG("want write");
           return writeBuffer.whenReady().then(kj::mvCapture(func,
-              [this](Func&& func) mutable { return sslCall(kj::fwd<Func>(func)); }));
+              [this](Func&& func) mutable { KJ_DBG("got write"); return sslCall(kj::fwd<Func>(func)); }));
```
harrishancock: Leftover KJ_DBG.

(I peeked at the rest of the PR, but didn't fully review since I lack some context. Nevertheless, 👍 FWIW.)

kentonv: Yikes, thanks for the catch. I forgot about that debug because it never actually triggers in local testing (since the write buffer never fills up).

c++/src/kj/compat/tls.c++
```diff
@@ -489,6 +490,13 @@ TlsContext::Options::Options()
 // Classic DH is arguably obsolete and will only become more so as time passes, so perhaps we'll
 // never bother.
 
+struct TlsContext::SniCallback {
+  // struct SniCallback exists only so that callback() can be declaned in the .c++ file, since it
```
zarvox: s/declaned/declared/

kentonv: Whoa, what brings you here? :)

(fixed)

zarvox: I'm still a watcher on capnproto and sandstorm-io, so I get emails from GitHub for each PR.  Don't always find the time to look over them, and I rarely have anything to contribute, but when the opportunity arises... :)

PR #615, opened 2018-01-12
Use relative path to fix escaping problems

asergunov:
The bug was about having "/../../include" value for CAPNP_INCLUDE_DIRECTORY in CapnProtoConfig.cmake. 

The problem about unescaped "${CMAKE_CURRENT_LIST_DIR}/.." value in c++/configure.ac

I've changed that to be relative path moving ${CMAKE_CURRENT_LIST_DIR} to c++/cmake/CapnProtoConfig.cmake.in


asergunov:
It breaks the cmake building

harrishancock:
Hi Anton,

Could you show me an example project that fails, so I can understand the problem better? Is`${CMAKE_CURRENT_LIST_DIR}` expanded to nothing before CapnProtoConfig.cmake.in is configured from configure.ac?

The `PACKAGE_CMAKE_INSTALL_FULL_INCLUDEDIR` variable name is chosen by the CMake build scripts, specifically in c++/CMakeLists.txt where it calls `configure_package_config_file()`.

We can rename that variable, but we can't change what it means: it needs to be a path to the include directory that will work even if the installation directory is moved. The `configure_package_config_file()` command controls that, so in configure.ac I just tried to simulate the same behavior.

One option to do what you perhaps intend here might be something like:

```cmake
# CapnProtoConfig.cmake.in
# ...

set(full_includedir "@PACKAGE_CMAKE_INSTALL_FULL_INCLUDEDIR@")
set(relative_includedir "@PACKAGE_CMAKE_INSTALL_RELATIVE_INCLUDEDIR@")

if(NOT full_includedir STREQUAL "")
  # We are being configured by a CMake build.
  set(CAPNP_INCLUDE_DIRECTORY "${full_includedir}")
else()
  # We are being configured by an autotools build.
  set(CAPNP_INCLUDE_DIRECTORY "${CMAKE_CURRENT_LIST_DIR}/${relative_includedir}")
endif()
```

That's kind of hacky, though ... perhaps there's a better solution?

asergunov:
Hello Harris!

To reproduce the bug just follow online build instructions.

cd c++
autoreconfigure -i
./configure --prefix=<some local path>
make -j8 install
Check CapnProtoConfig.cmake

All works great if configured using CMake. So it's not a blocker for me.
Just to let you know autoconf build broken

Regards
Anton Sergunov

пт, 12 янв. 2018 г., 22:02 Harris Hancock <notifications@github.com>:

PR #619, opened 2018-01-18
Escape CMake config var in configure.ac

harrishancock:
Previously ${CMAKE_CURRENT_LIST_DIR} was expanded (to an empty string) before CapnProtoConfig.cmake.in was configured, resulting in CapnProtoConfig.cmake setting CAPNP_INCLUDE_DIRECTORY to "/../../include", which is obviously wrong.

Originally reported by @asergunov in PR #615. Anton, could you confirm this solves the problem?


asergunov:
Hello @harrishancock 

I have build error:

```
Compiling Cap'n Proto schema <...>.capnp
<capnproto-install-autoconf>/bin/capnp compile: -I <capnproto-install-autoconf>/lib/cmake/CapnProto/../../include: no such directory
```

So configurations in `<capnproto-install-autoconf>/lib/cmake/CapnProto` on linux. In `<capnproto-install-autoconf>/lib/cmake/CapnProto/CapnProtoConfig.cmake:45` I have 
```
set(CAPNP_INCLUDE_DIRECTORY "${CMAKE_CURRENT_LIST_DIR}/../../include")
```

but should be 
```
set(CAPNP_INCLUDE_DIRECTORY "${CMAKE_CURRENT_LIST_DIR}/../../../include")
```

But keep in mind it can be Linux only issue because on windows cmake configs are in different place.

harrishancock:
Ack, you're right, I got the path wrong. I'm sorry about that -- I forgot to run the right test. :(

@kentonv, I git-bisected the bug being fixed here to e2a9467b5e9 -- the bug existed before that, but that's when it first manifested in build failures. I think ideally we should catch this in CI, so I added a couple commits to that effect.

c++/src/kj/common.h
```diff
@@ -1416,7 +1416,19 @@ To& downcast(From& from) {
   }
 
 #if !KJ_NO_RTTI
+
+#ifdef __GNUC__
+// GCC 6.3 with -Werror=address will fail to build if To and From are the same type.
```
kentonv: What is the warning it generates? Is this a compiler bug? Can we reference the bug number?

Are we ignoring the warning because it breaks *our* build, or other people's builds? Other people should be unaffected because we mark our headers as system headers. If it's our own build (with -Werror) that's affected, maybe we should adjust our build config to disable this warning if it is buggy?

harrishancock: Rebased in a more appropriate fix that removes the superfluous downcast, as discussed in person.

PR #622, opened 2018-01-30
Allow working around the message size limit

asilversempirical:
AFAICT, without this, there's no way to convince capnp to send a message larger than 64MB, even if you know the other side will accept it. This should cause it to work reasonably, assuming both sides set their message limits identically.

### DIFF
--- a/c++/src/capnp/rpc-twoparty.c++
+++ b/c++/src/capnp/rpc-twoparty.c++
@@ -86,10 +86,10 @@ public:
     for (auto& segment: message.getSegmentsForOutput()) {
       size += segment.size();
     }
-    KJ_REQUIRE(size < ReaderOptions().traversalLimitInWords, size,
-               "Trying to send Cap'n Proto message larger than the single-message size limit. The "
-               "other side probably won't accept it and would abort the connection, so I won't "
-               "send it.") {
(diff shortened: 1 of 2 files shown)


c++/src/capnp/rpc-twoparty.c++
```diff
@@ -86,10 +86,11 @@ public:
...
     }
-    KJ_REQUIRE(size < ReaderOptions().traversalLimitInWords, size,
-               "Trying to send Cap'n Proto message larger than the single-message size limit. The "
-               "other side probably won't accept it and would abort the connection, so I won't "
-               "send it.") {
+    KJ_REQUIRE(size < network.receiveOptions.traversalLimitInWords, size,
+               "Trying to send Cap'n Proto message larger than our single-message size limit. The "
+               "other side probably won't accept it (assuming its traversalLimitInWords matches "
+               "ours) and would abort the connection, so I won't send it.")
+    {
```
kentonv: Nitpick: Can you move this brace back to the previous line like it was before?

asilversempirical: Oops, bad editor! Done

kentonv:
Thanks!

PR #624, opened 2018-01-31
Add Array::attach() and ArrayPtr::attach().

kentonv:
Array::attach() is like Own::attach().

ArrayPtr::attach() promotes an ArrayPtr to an Array by attaching other objects to it. (Hopefully one of those objects actually owns the underlying array data.)

### DIFF
--- a/c++/src/kj/array.h
+++ b/c++/src/kj/array.h
@@ -25,7 +25,7 @@
 #pragma GCC system_header
 #endif
 
-#include "common.h"
+#include "memory.h"
 #include <string.h>
 #include <initializer_list>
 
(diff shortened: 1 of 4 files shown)


c++/src/kj/array.h
```diff
@@ -834,4 +838,48 @@ inline Array<Decay<T>> arr(T&& param1, Params&&... params) {
...
+  KJ_IREQUIRE(ptr != nullptr, "cannot attach to null pointer");
+
+  // HACK: If someone accidentally calls .attach() on a null pointer in opt mode, try our best to
+  //   accomplish reasonable behavior: We turn the pointer non-null but still invalid, so that the
+  //   disposer will still be called when the pointer goes out of scope.
+  if (ptrCopy == nullptr) ptrCopy = reinterpret_cast<T*>(1);
+
+  auto bundle = new _::ArrayDisposableOwnedBundle<Attachments...>(
+      kj::fwd<Attachments>(attachments)...);
+  return Array<T>(ptr, size_, *bundle);
```
harrishancock: `ptrCopy` instead of `ptr` here? (Maybe the same in the `KJ_IREQUIRE` above, for symmetry with `ArrayPtr::attach()`.)

PR #627, opened 2018-02-05
Use application/x-www-form-urlencoded for URL query strings

harrishancock:
The main point here is to prevent escaping '+' in a query string as '%2B'. It should instead be interpreted as a space.

Note that application/x-www-form-urlencoded is the format that the WHATWG URL spec uses to serialize query's through its URLSearchParams interface: https://url.spec.whatwg.org/#concept-urlsearchparams-update

### DIFF
--- a/c++/src/kj/encoding-test.c++
+++ b/c++/src/kj/encoding-test.c++
@@ -279,6 +279,9 @@ KJ_TEST("URI encoding/decoding") {
   KJ_EXPECT(encodeUriComponent("\xab\xba") == "%AB%BA");
   KJ_EXPECT(encodeUriComponent(StringPtr("foo\0bar", 7)) == "foo%00bar");
 
+  // Encode characters reserved by application/x-www-form-urlencoded, but not by RFC 2396.
+  KJ_EXPECT(encodeUriComponent("'foo'! (~)") == "'foo'!%20(~)");
+
   expectRes(decodeUriComponent("foo%20bar"), "foo bar");
   expectRes(decodeUriComponent("%ab%BA"), "\xab\xba");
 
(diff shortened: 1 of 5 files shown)


c++/src/kj/encoding.h
```diff
@@ -141,15 +141,20 @@ EncodingResult<String> decodeWwwForm(ArrayPtr<const char> text);
 struct DecodeUriOptions {
   // Parameter to `decodeBinaryUriComponent()`.
 
-  bool nulTerminate = false;
+  // TODO(someday): Remove this constructor once we only support compilers that can deal with
```
kentonv: Specifically this is a C++14ism.

c++/src/kj/encoding.h
```diff
@@ -124,10 +124,34 @@ EncodingResult<Array<byte>> decodeHex(ArrayPtr<const char> text);
...
+
+  bool nulTerminate = false;
+  // Append a terminal NUL byte.
+
+  bool plusToSpace = false;
+  // Convert '+' to ' ' characters before percent decoding. Used to decode
+  // application/x-www-form-urlencoded text, such as query strings.
+};
+EncodingResult<Array<byte>> decodeBinaryUriComponent(
+    ArrayPtr<const char> text, DecodeUriOptions options = {});
```
kentonv: For backwards-compatibility I think DecodeUriOptions needs an implicit conversion from boolean, where the boolean is the value for `nullTerminate`.

harrishancock:
Addressed C++14ism comment, rebased onto master.

c++/src/kj/encoding-test.c++
```diff
@@ -287,8 +290,43 @@ KJ_TEST("URI encoding/decoding") {
...
+    byte bytes[] = {12, 34, 56};
+    KJ_EXPECT(decodeBinaryUriComponent(encodeUriComponent(bytes)).asPtr() == bytes);
+
+    // decodeBinaryUriComponent() takes a DecodeUriOptions struct as its second parameter, but it
+    // once took a single `bool nulTerminate`. Verify that the old behavior still compiles and
+    // works.
+    auto bytesWithNul = decodeBinaryUriComponent(encodeUriComponent(bytes), true);
+    KJ_ASSERT(bytesWithNul.size() == 4);
+    KJ_EXPECT(bytesWithNul[3] == '\0');
+    KJ_EXPECT(bytesWithNul.slice(0, 3) == bytes);
```
harrishancock: Backwards-compatibility test here.

PR #638, opened 2018-02-23
Extend HttpServer to inform the caller when a connection ends cleanly on drain().

kentonv:
This allows an application which calls drain() to potentially pass off HTTP connections to a new HttpServer afterwards. Without this, the application has no way to know if the connections are in an indeterminant state.

This change also makes it OK for an application to fail to read the whole request body. Previously, if an app returned a response without reading the whole request, an exception would eventually be thrown, but potentially not until the client had initiated a new request on the same connection. The client would then get a spurious 500 error.

### DIFF
--- a/c++/src/kj/compat/http-test.c++
+++ b/c++/src/kj/compat/http-test.c++
@@ -1704,10 +1704,6 @@ public:
     } else if (url == "/ws-inline") {
       auto ws = response.acceptWebSocket(responseHeaders);
       return doWebSocket(*ws, "start-inline").attach(kj::mv(ws));
-    } else if (url == "/ws-detached") {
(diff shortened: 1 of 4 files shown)


c++/src/kj/compat/http-test.c++
```diff
@@ -1909,31 +1903,6 @@ KJ_TEST("HttpServer WebSocket handshake") {
   listenTask.wait(io.waitScope);
 }
 
-KJ_TEST("HttpServer WebSocket handshake detached") {
```
kentonv: This test was testing that you could extend a WebSocket's lifetime beyond the request handler promise, which was a dubious feature that I never relied on. That feature is now removed.

PR #641, opened 2018-03-12
Fix bugs in relative URL parser

harrishancock:
The relative URL parser allowed certain hosts which the absolute parser correctly weeded out. This change copies two lines from the absolute parser over to the relative parser.


harrishancock:
Added a second bug fix -- relative-parsing "/foo/../bar" previously resulted in "/foo/bar".

Haven't figured out the build failure yet.

harrishancock:
Rebased on master to fix CI.

PR #645, opened 2018-03-19
When catching an unknown exception type, at least log the type.

kentonv:


harrishancock:
I'm late, but looks good!

kentonv:
@harrishancock Stop doing code reviews while on vacation! :)

PR #650, opened 2018-03-27
Implement various WHATWG URL percent encoding functions

harrishancock:
~DRAFT - do not merge yet. See below about path serialization for my main concern.~

This adds `encodeUriFragment()`, `encodeUriPath()`, and `encodeUriUserInfo()` to kj-encoding, and uses them in `kj::Url::toString()` to become slightly more conformant in the eyes of the WHATWG URL spec.

It's worth noting that the reserved character set of `encodeUriFragment()` is a subset of `encodeUriPath()`, which in turn is a subset of `encodeUriUserInfo()`, owing to their positions in URL strings. All three functions' reserved character sets are subsets of `encodeUriComponent()` (RFC 2396).

~**Path serialization:** I implemented `encodeUrlPath()` the obvious way: by following what the spec says to do. However, the spec says not to encode `/` characters. This means that if we parse a URL with one path component, and we set that one path component to "foo/bar", then if we serialize and reparse it, we will now magically have two path components. I'm not sure if this is a problem or not. For our use case, it may not matter.~

**Implementation:** I think it would be more elegant implementation-wise if we refactored all of these new functions, and `encodeUriComponent()`, into one function parameterized on a `CharGroup_` expressing the unreserved (or reserved) character sets in question. However, I wanted to get your opinion on this draft, first, as I'm unsure if that would change performance characteristics significantly in one way or another. Note that I'm not sure that `encodeWwwForm()` could be rolled into such a refactor, owing to its weird plus-to-space rules.

~**Naming:** I'm open to suggestions. It makes sense to me that `encodeUriComponent()` is named the way it is, because it duplicates the similarly-named JS function. Likewise, it makes sense that these new functions are named `encodeUrl*()`, since they implement functionality standardized by the URL spec. Nevertheless, the resulting inconsistency between Url and Uri is a bit glaring. (Note that I don't think `encodeWwwForm()` should be named `encodeUrlQueryParameter()`, since it's a bit more special than these new functions.):~

### DIFF
--- a/c++/src/kj/encoding.c++
+++ b/c++/src/kj/encoding.c++
@@ -390,9 +390,67 @@ EncodingResult<Array<byte>> decodeHex(ArrayPtr<const char> text) {
 String encodeUriComponent(ArrayPtr<const byte> bytes) {
   Vector<char> result(bytes.size() + 1);
   for (byte b: bytes) {
-    if (('A' <= b && b <= 'Z') || ('a' <= b && b <= 'z') || ('0' <= b && b <= '9') ||
+    if (('A' <= b && b <= 'Z') ||
+        ('a' <= b && b <= 'z') ||
+        ('0' <= b && b <= '9') ||
         b == '-' || b == '_' || b == '.' || b == '!' || b == '~' || b == '*' || b == '\'' ||
         b == '(' || b == ')') {
+      result.add(b);
+    } else {
+      result.add('%');
+      result.add(HEX_DIGITS_URI[b/16]);
+      result.add(HEX_DIGITS_URI[b%16]);
+    }
+  }
+  result.add('\0');
+  return String(result.releaseAsArray());
+}
+
+String encodeUriFragment(ArrayPtr<const byte> bytes) {
+  Vector<char> result(bytes.size() + 1);
+  for (byte b: bytes) {
+    if (('?' <= b && b <= '_') || // covers A-Z
+        ('a' <= b && b <= '~') || // covers a-z
+        ('#' <= b && b <= ';') || // covers 0-9
+        b == '!' || b == '=') {
+      result.add(b);
+    } else {
+      result.add('%');
+      result.add(HEX_DIGITS_URI[b/16]);
+      result.add(HEX_DIGITS_URI[b%16]);
+    }
+  }
+  result.add('\0');
+  return String(result.releaseAsArray());
+}
+
+String encodeUriPath(ArrayPtr<const byte> bytes) {
+  Vector<char> result(bytes.size() + 1);
+  for (byte b: bytes) {
+    if (('@' <= b && b <= '[') || // covers A-Z
+        ('a' <= b && b <= 'z') ||
+        ('0' <= b && b <= ';') || // covers 0-9
+        ('$' <= b && b <= '.') ||
+        b == '_' || b == '!' || b == '=' || b == ']' || b == '^' || b == '|' || b == '~') {
+      result.add(b);
+    } else {
+      result.add('%');
+      result.add(HEX_DIGITS_URI[b/16]);
+      result.add(HEX_DIGITS_URI[b%16]);
+    }
+  }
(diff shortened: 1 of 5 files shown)


kentonv:
> I implemented encodeUrlPath() the obvious way: by following what the spec says to do. However, the spec says not to encode / characters. This means that if we parse a URL with one path component, and we set that one path component to "foo/bar", then if we serialize and reparse it, we will now magically have two path components. I'm not sure if this is a problem or not. For our use case, it may not matter.

This is definitely a problem. We need to encode '/'. The reason we differ from the spec here is because we split the path into components whereas the spec does not.

kentonv:
Regarding the API:
* Parameterizing over `CharGroup_` would be elegant, but I don't think I want to `#include <kj/parse/char.h>` from `encoding.h`... so we still need these separate functions, which could be backed by a common implementation in the `.c++` file, but meh, it's not worth the effort to change -- what you have is fine.
* I would use `encodeUri*` rather than `encodeUrl*` for consistency.

kentonv:
Other than the above, the change looks good.

harrishancock:
> This is definitely a problem. We need to encode '/'. The reason we differ from the spec here is because we split the path into components whereas the spec does not.

I researched this a bit more. The spec actually does require implementations to split the path (to facilitate ".." collapsing and relative-path parsing), so the reason we differ is more subtle: the spec stores the path components in percent-encoded form and hides mutation behind a setter function which splits out and percent-encodes the individual components of the user's pathname (so slashes can never get into individual path components), whereas we store the path components in percent-decoded form in a struct and have no setter.

Regardless, the fix is the same: we need to encode '/', as you said. (And '\\'! The URL spec honors them as separators, ugh.)

Note that our percent-decoding behavior technically breaks spec, though we're in good company. For example, round-tripping "https://foo/%62%61%72" is not supposed to change it at all: Firefox is conformant, while Chrome decodes it to "https://foo/bar".

c++/src/kj/encoding.c++
```diff
@@ -390,7 +390,9 @@ EncodingResult<Array<byte>> decodeHex(ArrayPtr<const char> text) {
 String encodeUriComponent(ArrayPtr<const byte> bytes) {
   Vector<char> result(bytes.size() + 1);
   for (byte b: bytes) {
-    if (('A' <= b && b <= 'Z') || ('a' <= b && b <= 'z') || ('0' <= b && b <= '9') ||
+    if (('A' <= b && b <= 'Z') ||
+        ('a' <= b && b <= 'z') ||
+        ('0' <= b && b <= '9') ||
```
harrishancock: No functional change, just reindented for uniformity with the new functions.

c++/src/kj/encoding.h
```diff
@@ -130,6 +130,33 @@ EncodingResult<String> decodeUriComponent(ArrayPtr<const char> text);
...
+String encodeUriPath(ArrayPtr<const byte> bytes);
+String encodeUriPath(ArrayPtr<const char> bytes);
+// Encode URL path components (not entire paths!) using the path percent encode set defined by the
+// WHATWG URL specification. Use decodeUriComponent() to decode.
+//
+// Quirk: This percent-encodes '/' and '\' characters as well, which are not actually in the set
+//   defined by the WHATWG URL spec. Since a conforming URL implementation will only ever call this
+//   function on individual path components, and never entire paths, augmenting the character set to
+//   include these separators allows this function to be used to implement a URL class that stores
+//   its path components in either percent-encoded OR percent-decoded form.
```
harrishancock: New comment (used to be "NOTE: We pass through '/'").

harrishancock:
This should be good now, PTAL.

kentonv:
Let's add a test that percent-encoded '/' characters in path components remain so, i.e.:

    parseAndCheck("https://capnproto.org/foo%2Fbar/baz", "https://capnproto.org/foo%2Fbar/baz");

PR #662, opened 2018-04-22
Implement newHttpClient(HttpService&).

kentonv:
It turns out wrapping an HttpService in an HttpClient is considerably more complicated than vice versa, due to the need for pipes. This commit adds a WebSocket pipe implementation very similar to the recent byte-stream pipe (though considerably simpler since there's no need to deal with mismatched buffer sizes).

### DIFF
--- a/c++/src/kj/async-io.c++
+++ b/c++/src/kj/async-io.c++
@@ -522,7 +522,26 @@ private:
 
     void abortRead() override {
       canceler.cancel("abortRead() was called");
-      fulfiller.reject(KJ_EXCEPTION(FAILED, "read end of pipe was aborted"));
+
+      // The input might have reached EOF, but we haven't detected it yet because we haven't tried
+      // to read that far. If we had not optimized tryPumpFrom() and instead used the default
+      // pumpTo() implementation, then the input would not have called write() again once it
+      // reached EOF, and therefore the abortRead() on the other end would *not* propagate an
+      // exception! We need the same behavior here. To that end, we need to detect if we're at EOF
+      // by reading one last byte.
+      checkEofTask = kj::evalNow([&]() {
+        static char junk;
+        return input.tryRead(&junk, 1, 1).then([this](uint64_t n) {
+          if (n == 0) {
+            fulfiller.fulfill(kj::cp(pumpedSoFar));
+          } else {
+            fulfiller.reject(KJ_EXCEPTION(FAILED, "read end of pipe was aborted"));
+          }
+        }).eagerlyEvaluate([this](kj::Exception&& e) {
+          fulfiller.reject(kj::mv(e));
+        });
+      });
+
       pipe.endState(*this);
       pipe.abortRead();
     }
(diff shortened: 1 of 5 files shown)


c++/src/kj/compat/http.h
```diff
@@ -634,6 +640,15 @@ kj::Own<WebSocket> newWebSocket(kj::Own<kj::AsyncIoStream> stream,
 // like HTTP requests" in a message as being actual HTTP requests, which could result in cache
 // poisoning. See RFC6455 section 10.3.
 
+struct WebSocketPipe {
+  kj::Own<WebSocket> ends[2];
+};
+
+WebSocketPipe newWebSocketPipe();
+// Create a WebSocket pipe. Messages written to one end of the pipe will be readable from the other
+// end. No buffering occurs -- a message send does not complete under a corresponding receive
```
harrishancock: /under/until/

c++/src/kj/compat/http.c++
```diff
@@ -2406,17 +2406,532 @@ static kj::Promise<void> pumpWebSocketLoop(WebSocket& from, WebSocket& to) {
...
+    kj::Promise<void> disconnect() override {
+      KJ_REQUIRE(canceler.isEmpty(), "already pumping");
+      fulfiller.reject(KJ_EXCEPTION(DISCONNECTED, "WebSocket disconnected"));
+      pipe.endState(*this);
+      return pipe.disconnect();
+    }
+    kj::Maybe<kj::Promise<void>> tryPumpFrom(WebSocket& other) override {
+      KJ_REQUIRE(canceler.isEmpty(), "already pumping");
+      return canceler.wrap(other.receive().then([this,&other](Message message) {
+        fulfiller.fulfill(kj::mv(message));
```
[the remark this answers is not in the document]
kentonv: Yes, good catch.

kentonv:
Added a bugfix for a problem I discovered while using this in Workers: If the person reading from the read end of a pipe knows exactly how much data to expect and stops reading at that point, `tryPumpFrom()` won't get a chance to notice that EOF has been reached and will fail out when the read end is dropped.

c++/src/kj/async-io.c++
```diff
@@ -522,7 +522,26 @@ private:
...
     void abortRead() override {
       canceler.cancel("abortRead() was called");
-      fulfiller.reject(KJ_EXCEPTION(FAILED, "read end of pipe was aborted"));
+
+      // The input might have reached EOF, but we haven't detected it yet because we haven't tried
+      // to read that far. If we had not optimized tryPumpFrom() and instead used the default
+      // pumpTo() implementation, then the input would not have called write() again once it
+      // reached EOF, and therefore the abortRead() on the other end would *not* propagate an
+      // exception! We need the same behavior here. To that end, we need to detect if we're at EOF
+      // by reading one last byte.
```
harrishancock: If we're not at EOF here, and the last, one-byte read would block, wouldn't this prevent the `tryPumpFrom()` promise from fulfilling/rejecting until data becomes available?

kentonv: Yes, but that's what the generic `pumpFrom()` implementation would do as well. It would normally not notice that the output has been aborted until it gets some input which it tries to write, and that write fails.

harrishancock: Ah, I see!

PR #663, opened 2018-04-23
Fix filesystem-disk-test to support platforms with non-4K filesystem block size

secumod:
Some memory-based filesystems, namely tmpfs, select their underlying block size
equal to the configured page size of the platform/OS.

On systems with page size not equal to 4K (for example, arm64 can be configured
to have 16K or 64K page sizes -
https://www.kernel.org/doc/Documentation/arm64/memory.txt) "DiskFile holes"
test fails, when creating a file on a memory-based filesystem, because it
"punches" a 4096 byte hole in the file (using fallocate) and expects the number
of filesystem blocks used by the file to decrease. When the system page size is
greater than 4K, 4096 byte hole is not enough to make the OS release the
underlying blocks.

Need to punch a whole, which is equal to the underlying filesystem block size at
least.

### DIFF
--- a/c++/src/kj/filesystem-disk-test.c++
+++ b/c++/src/kj/filesystem-disk-test.c++
@@ -897,7 +897,14 @@ KJ_TEST("DiskFile holes") {
 #endif
 
   // Try punching a hole with zero().
-  file->zero(1 << 20, 4096);
+#if _WIN32
+  uint64_t blockSize = 4096; // TODO(someday): Actually ask the OS.
+#else
+  struct stat stats;
+  KJ_SYSCALL(fstat(KJ_ASSERT_NONNULL(file->getFd()), &stats));
+  uint64_t blockSize = stats.st_blksize;
+#endif
+  file->zero(1 << 20, blockSize);
   file->datasync();
 #if !_WIN32
   // TODO(someday): This doesn't work on Windows. I don't know why. We're definitely using the


kentonv:
Can we maybe do a simpler change where we don't change the `Metadata` type, but instead alter the test to check the block size directly using the appropriate syscall?

So the test should have code like:

    #if _WIN32
      size_t blockSize = 4096;  // TODO(someday): Actually ask the OS.
    #else
      struct stat stats;
      KJ_SYSCALL(stat(KJ_ASSERT_NONNULL(file->getFd()), &stats));
      size_t blockSize = stats.st_blksize;
    #endif
      file->zero(1 << 20, blockSize);

The equivalent syscall to `stat()` on Windows does not return filesystem block size, so extending `Metadata` in this way implies that the Windows implementation will need to do an additional syscall to fetch data that most applications won't use.

secumod:
This was actually my first fix, but then I was thinking that having "preferred" FS block size in the `Metadata` might be beneficial. Maybe, we just embed `blkSize = 4096` as a fixed value for now into the Windows version of the `Metadata`, which will be semantically the same as the code above?

kentonv:
There's a big difference between hard-coding the value in a test and hard-coding it in a public-facing API. :)

Returning a hard-coded block size on Windows would be incorrect, and then we'd presumably be obliged to "fix" it which would require adding syscalls.

secumod:
I haven't looked/used the code much myself, but I don't quite understand: if it has code to test FS sparse files/hole punching and `Metadata` already exposes `spaceUsed` information, then I would expect it tries to help the caller manage disk space/IO more efficiently. But not having FS block size available makes `spaceUsed` not very useful as the caller has to make guesses (or hope his FS has the popular 4K block size) to actually influence `spaceUsed` in some meaningful way. I would say `blkSize`/`st_blksize` and `spaceUsed`/`st_blocks` usually go hand-in-hand and if the code exposes one it should expose the other.

But this indeed may be the story of another issue/PR/feature request. For now, I'll redo as above.

kentonv:
An application does not actually need to understand the details of block allocation to use sparse files nor to understand the `spaceUsed` member. The API does not require page-aligned reads or writes; the filesystem will automatically round up/down to the nearest block boundaries as needed. `spaceUsed` is useful for reporting purposes, but a normal application would not need to know exactly where `spaceUsed` came from nor how to influence it. The test is unique in that we're explicitly testing that `spaceUsed` is implemented correctly, so we have to figure out how to reliably influence it.

Some complex apps that need to fine-tune their disk I/O may need to understand blocks, but they probably need to understand many other things about the filesystem, too, and this interface is not likely to be satisfying to them in any case.

secumod:
Fixed, as requested above.

kentonv:
Thanks!

PR #666, opened 2018-05-01
Improve HTTP body cancellation handling.

kentonv:


c++/src/kj/compat/http.c++
```diff
@@ -1689,6 +1704,11 @@ private:
   bool inBody = false;
   bool broken = false;
 
+  bool writeInProgress = false;
+  // True if a write method has been called and has not completed successfully. In the case that
+  // a write throws an exception or is canceled, this remains true forever. In these cases, the
+  // underlying steram is in an inconsitent state and cannot be reused.
```
harrishancock: /steram/stream/
/inconsitent/inconsistent/

PR #674, opened 2018-05-14
Add flag to kj::Url::parse() to prevent automatic decoding of URL components

harrishancock:


harrishancock:
Amended to use an Options struct instead.

c++/src/kj/compat/url-test.c++
```diff
@@ -36,6 +36,11 @@ Url parseAndCheck(kj::StringPtr originalText, kj::StringPtr expectedRestringifie
   return url;
 }
 
+static constexpr Url::Options NO_DECODE {
+  /* .percentDecode = */ false,
+  // TODO(cleanup): Use designated initializer once MSVC supports them.
```
kentonv: Designated initializers are not standardized for C++, only C. Some compilers support them in C++ but we shouldn't rely on that until they're actually standardized (maybe C++20, it looks like).

harrishancock:
Addressed designated initializer review comment, moved tests into their own KJ_TEST blocks, and squashed.

PR #677, opened 2018-05-17
Rewrite JSON decoding

RReverser:
- Add support for decoding with Orphanage (as a consequence, allow decoding of non-struct root objects).
 - Add support for custom decoding type and field handlers.
 - Make basic tests verify that they roundtrip to the same string.
 - Simplify implementation of number decoding and, as a bonus, allow overflowing floating numbers to convert to +/-`Infinity` like they do in `JSON.parse`.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -17,6 +17,7 @@ Harris Hancock <vortrab@gmail.com>: MSVC support
 Branislav Katreniak <branislav.katreniak@digitalstrom.com>: JSON decode
 Matthew Maurer <matthew.r.maurer@gmail.com>: Canonicalization Support
 David Renshaw <david@sandstorm.io>: bugfixes and miscellaneous maintenance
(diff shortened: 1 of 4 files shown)


kentonv:
Thanks!

kentonv:
Notes for future work: Ingvar and I found that the interface is somewhat broken in that JsonCodec pre-allocates a struct and then parses into it. This leaves no opportunity for a Handler to allocate a struct that's larger than the schema defines, in order to retain unknown fields. It's not clear if there's a direct use case for this, but a side effect is that you cannot easily copy an existing struct value into the output struct, because Cap'n Proto intentionally doesn't provide a way to copy a struct into an already-allocated destination location due to the fact that this would lose unknown fields.

The correct interface would be for a struct handler to return an Orphan just like list handlers do. But, to make that work, JsonCodec's own interface would also need to return an orphan rather than parse into a pre-allocated struct...

RReverser:
@kentonv I'd suggest to create an issue with this description, as comments in closed PRs are much harder to find in the future.

PR #678, opened 2018-05-21
Implement kj::Table, an alternative to STL maps/sets.

kentonv:
Hash-based (unordered) and tree-based (ordered) indexing are provided.

kj::Table offers advantages over STL:
- A Table can have multiple indexes (allowing lookup by multiple keys). Different indexes can use different algorithms (e.g. hash vs. tree) and have different uniqueness constraints.
- The properties on which a Table is indexed need not be explicit fields -- they can be computed from the table's row type.
- Tables use less memory and make fewer allocations than STL, because rows are stored in a contiguous array.
- The hash indexing implementation uses linear probing rather than chaining, which again means far fewer allocations and more cache-friendliness.
- The tree indexing implementation uses B-trees optimized for cache line size, whereas STL uses cache-unfriendly and allocation-heavy red-black binary trees. (However, STL trees are overall more cache-friendly; see below.)
- Most of the b-tree implementation is not templated. This reduces code bloat, at the cost of some performance due to virtual calls.

On an ad hoc benchmark on large tables, the hash index implementation appears to outperform libc++'s `std::unordered_set` by ~60%. However, libc++'s `std::set` still outperforms the B-tree index by ~70%. It looks like the B-tree implementation suffers in part from the fact that keys are not stored inline in the tree nodes, forcing extra memory indirections. This is a price we pay for lower memory usage overall, and the ability to have multiple indexes on one table. The b-tree implementation also suffers somewhat from not being 100% templates, compared to STL, but I think this is a reasonable trade-off. The most performance-critical use cases will use hash indexes anyway.

### DIFF
--- a/c++/src/kj/map-test.c++
+++ b/c++/src/kj/map-test.c++
@@ -1,0 +1,133 @@
+// Copyright (c) 2018 Kenton Varda and contributors
+// Licensed under the MIT License:
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+#include "map.h"
+#include <kj/test.h>
+
+namespace kj {
+namespace _ {
+
+KJ_TEST("HashMap") {
+  HashMap<String, int> map;
+
+  map.insert(kj::str("foo"), 123);
+  map.insert(kj::str("bar"), 456);
+
+  KJ_EXPECT(KJ_ASSERT_NONNULL(map.find("foo"_kj)) == 123);
+  KJ_EXPECT(KJ_ASSERT_NONNULL(map.find("bar"_kj)) == 456);
+  KJ_EXPECT(map.find("baz"_kj) == nullptr);
+
+  map.upsert(kj::str("foo"), 789, [](int& old, uint newValue) {
+    KJ_EXPECT(old == 123);
+    KJ_EXPECT(newValue == 789);
+    old = 321;
+  });
+
+  KJ_EXPECT(KJ_ASSERT_NONNULL(map.find("foo"_kj)) == 321);
+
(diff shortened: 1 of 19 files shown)


harrishancock:
Build failures! MSVC didn't like the `__asm__` keyword (hehe, good luck) and gcc/clang didn't like a `static_assert()` with no string message (a C++17 feature).

c++/src/kj/table.c++
```diff
@@ -0,0 +1,810 @@
...
+        KJ_ASSERT(f(*leaf.rows[i], *leaf.rows[i + 1]));
+      } else {
+        KJ_ASSERT(maxRow == nullptr || leaf.rows[n-1] == maxRow);
+      }
+    }
+    return n;
+  }
+}
+
+void BTreeImpl::logInconsitency() const {
```
zarvox: nit: `s/Inconsitency/Inconsistency/`

c++/src/kj/table.h
```diff
@@ -0,0 +1,1482 @@
...
+// A Table index based on a B-tree.
+//
+// This allows sorted iteration over rows.
+//
+// The `Callbacks` type defines how to compare rows. It should be defined like:
+//
+//   class Callbacks {
+//   public:
+//     bool isBefore(const Row&, const Row&) const;
+//     // Returns true if the row on te left comes before the row on the right.
```
zarvox: nit: `s/te/the/`

c++/src/kj/table.h
```diff
@@ -0,0 +1,1482 @@
...
+private:
+  struct Link {
+    uint next;
+    uint prev;
+  };
+
+  uint capacity;
+  Link* links;
+  // links[0] is special: links[0].next points to the first link, links[0].prev points to the last.
+  // links[n+1] corresponds no row n.
```
zarvox: nit: `s/no/to/`

kentonv:
I was expecting the `findOrCreate()` change would totally break the MSVC build, since I did some template stuff.

Instead, it apparently broke everything *except* the MSVC build (and very-old Clang).

wat

c++/src/kj/table.h
```diff
@@ -0,0 +1,1391 @@
...
+
+template <typename Callbacks>
+class TreeIndex {
+public:
+  TreeIndex() = default;
+  template <typename... Params>
+  TreeIndex(Params&&... params): cb(kj::fwd<Params>(params)...) {}
+
+  template <typename Row>
+  void verify(kj::ArrayPtr<Row> table) { // KJ_DBG
```
harrishancock: `KJ_DBG`? Just marking that this is a unit test API?

c++/src/kj/table.h
```diff
@@ -293,18 +293,22 @@ class HashIndex;
...
+//     SearchParam& keyForRow(const Row& row);
+//     // Given a row of the table, return the SearchParams that might be passed to the other
+//     // methods to match this row.
 //
-//     uint hashCode(const Row&) const;
-//     // Computes the hash code of the given row. Matching rows (as determined by match()) must
-//     // have the same hash code. Non-matching rows should have different hash codes, to the
-//     // maximum extent possible. Non-matching rows with the same hash code hurt performance.
+//     bool matches(const Row&, SearchParams&&...) const;
+//     // Returns true if the the row on the left matches thes search params on the right.
```
harrishancock: /thes/the/

c++/src/kj/table-test.c++
```diff
@@ -0,0 +1,600 @@
...
+          KJ_ASSERT(*iter++ == i * 5 + 123);
+        }
+      }
+      KJ_ASSERT(iter == range.end());
+    }
+  }
+}
+
+}  // namespace kj
+}  // namespace _
```
harrishancock: Swap order of comments.

c++/src/kj/map-test.c++
```diff
@@ -0,0 +1,133 @@
...
+    auto ordered = KJ_MAP(e, map) -> kj::StringPtr { return e.key; };
+    KJ_ASSERT(ordered.size() == 3);
+    KJ_EXPECT(ordered[0] == "bar");
+    KJ_EXPECT(ordered[1] == "foo");
+    KJ_EXPECT(ordered[2] == "qux");
+  }
+}
+
+}  // namespace kj
+}  // namespace _
```
harrishancock: Swap order of comments.

PR #679, opened 2018-05-22
cmake: drop NO_DEFAULT_PATHS when searching libraries

intelfx:
It looks like pkg-config silently drops -I and -L flags pointing to
default directories:

```
$ cat /usr/lib/pkgconfig/capnp.pc
prefix=/usr
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include

Name: Cap'n Proto
Description: Insanely fast serialization system
Version: 0.7-dev
Libs: -L${libdir} -lcapnp -pthread  -lpthread
Libs.private:  -lpthread
Requires: kj = 0.7-dev
Cflags: -I${includedir} -pthread

$ pkg-config --libs capnp
-lcapnp -pthread -lpthread -lkj -pthread -lpthread
```

Ideally, however, we should use FindPkgConfig.cmake's own facilities
to generate IMPORTED targets from pkg-config files.

### DIFF
--- a/c++/cmake/CapnProtoTargets.cmake
+++ b/c++/cmake/CapnProtoTargets.cmake
@@ -86,7 +86,6 @@ function(_capnp_import_pkg_config_target target)
   find_library(CapnProto_${target}_IMPORTED_LOCATION
     NAMES ${target_name_shared} ${target_name_static}  # prefer libfoo-version.so over libfoo.a
     PATHS ${${target}_LIBRARY_DIRS}
-    NO_DEFAULT_PATH
   )
   if(NOT CapnProto_${target}_IMPORTED_LOCATION)
     # Not an error if the library doesn't exist -- we may have found a lite mode installation.
--- a/c++/cmake/CapnProtoTargets.cmake
+++ b/c++/cmake/CapnProtoTargets.cmake
@@ -86,7 +86,14 @@ function(_capnp_import_pkg_config_target target)
   find_library(CapnProto_${target}_IMPORTED_LOCATION
     NAMES ${target_name_shared} ${target_name_static}  # prefer libfoo-version.so over libfoo.a
     PATHS ${${target}_LIBRARY_DIRS}
+    NO_DEFAULT_PATH
   )
+  # If the installed version of Cap'n Proto is in a system location, pkg-config will not have filled
+  # in ${target}_LIBRARY_DIRS. To account for this, fall back to a regular search.
+  find_library(CapnProto_${target}_IMPORTED_LOCATION
+    NAMES ${target_name_shared} ${target_name_static}  # prefer libfoo-version.so over libfoo.a
+  )
+
   if(NOT CapnProto_${target}_IMPORTED_LOCATION)
     # Not an error if the library doesn't exist -- we may have found a lite mode installation.
     if(CapnProto_DEBUG)


intelfx:
This ultimately leads to inability to use CMake integration at all:

```
$ cmake . -DCapnProto_DEBUG=ON
-- Boost version: 1.66.0
-- Found the following Boost libraries:
--   filesystem
--   program_options
--   system
-- capnp_VERSION = 0.7-dev
-- capnp_LIBRARY_DIRS = 
-- capnp_LIBRARIES = capnp;pthread;kj;pthread
-- capnp_LDFLAGS_OTHER = -pthread
-- capnp_INCLUDE_DIRS = 
-- capnp_CFLAGS_OTHER = -pthread
-- capnp library does not exist
-- capnp-rpc_VERSION = 0.7-dev
-- capnp-rpc_LIBRARY_DIRS = 
-- capnp-rpc_LIBRARIES = capnp-rpc;capnp;pthread;kj-async;pthread;kj;pthread
-- capnp-rpc_LDFLAGS_OTHER = -pthread
-- capnp-rpc_INCLUDE_DIRS = 
-- capnp-rpc_CFLAGS_OTHER = -pthread
-- capnp-rpc library does not exist
-- capnp-json_VERSION = 0.7-dev
-- capnp-json_LIBRARY_DIRS = 
-- capnp-json_LIBRARIES = capnp-json;capnp;pthread;kj;pthread
-- capnp-json_LDFLAGS_OTHER = -pthread
-- capnp-json_INCLUDE_DIRS = 
-- capnp-json_CFLAGS_OTHER = -pthread
-- capnp-json library does not exist
-- kj_VERSION = 0.7-dev
-- kj_LIBRARY_DIRS = 
-- kj_LIBRARIES = kj;pthread
-- kj_LDFLAGS_OTHER = -pthread
-- kj_INCLUDE_DIRS = 
-- kj_CFLAGS_OTHER = -pthread
-- kj library does not exist
-- kj-async_VERSION = 0.7-dev
-- kj-async_LIBRARY_DIRS = 
-- kj-async_LIBRARIES = kj-async;pthread;kj;pthread
-- kj-async_LDFLAGS_OTHER = -pthread
-- kj-async_INCLUDE_DIRS = 
-- kj-async_CFLAGS_OTHER = -pthread
-- kj-async library does not exist
-- kj-http_VERSION = 0.7-dev
-- kj-http_LIBRARY_DIRS = 
-- kj-http_LIBRARIES = kj-http;pthread;kj-async;pthread;kj;pthread
-- kj-http_LDFLAGS_OTHER = -pthread
-- kj-http_INCLUDE_DIRS = 
-- kj-http_CFLAGS_OTHER = -pthread
-- kj-http library does not exist
-- kj-test_VERSION = 0.7-dev
-- kj-test_LIBRARY_DIRS = 
-- kj-test_LIBRARIES = kj-test;pthread;kj;pthread
-- kj-test_LDFLAGS_OTHER = -pthread
-- kj-test_INCLUDE_DIRS = 
-- kj-test_CFLAGS_OTHER = -pthread
-- kj-test library does not exist
CMake Error at src/shared_cache/impls/lmdb_shared_cache/shared_server_proto/CMakeLists.txt:1 (find_package):
  Found package configuration file:

    /usr/lib64/cmake/CapnProto/CapnProtoConfig.cmake

  but it set CapnProto_FOUND to FALSE so package "CapnProto" is considered to
  be NOT FOUND.


-- Configuring incomplete, errors occurred!
See also "/home/intelfx/devel/__mainline/waf/ptaf/cmake-build-debug/CMakeFiles/CMakeOutput.log".
```

harrishancock:
Nice catch!

I think the reason I used `NO_DEFAULT_PATH` originally was because I wanted to make sure default system paths don't override locally installed capnproto distributions (e.g., a /usr prefix taking precedence over /opt/capnproto), since the default paths are checked before the paths specified in `PATHS`. I must never have tested the actual default path case itself.

I think this change might make the `PATHS ${${target}_LIBRARY_DIRS}` parameter redundant. It's the last to be searched by the `find_library()` algorithm, so in the scenario I was originally worried about (one capnproto in /usr, another in /opt/capnproto), I'd expect `find_library()` to search /usr before /opt/capnproto. Except that doesn't happen, because for pkg-config to even see the *.pc files in the first place, we must have set something like `CMAKE_PREFIX_PATH`, which would override both the default paths and the `PATHS` paths, so `find_library()` does the right thing without even searching the `PATHS` argument. @intelfx , maybe we should remove that argument, too?

> Ideally, however, we should use FindPkgConfig.cmake's own facilities
to generate IMPORTED targets from pkg-config files.

I agree. We'd need to upgrade our minimum CMake version required to 3.6 to use that, it looks like -- but that shouldn't be a problem nowadays. We'd also want to ensure that the generated IMPORTED targets match CMake-generated IMPORTED targets closely enough that reasonable CMake-based projects which `find_package(CapnProto)` won't notice the difference. I'm mostly worried about transitive dependencies (e.g. `CapnProto::kj-async` -> `CapnProto::kj`) working correctly. Hopefully we can do this at some point.

c++/cmake/CapnProtoTargets.cmake
```diff
@@ -86,7 +86,6 @@ function(_capnp_import_pkg_config_target target)
   find_library(CapnProto_${target}_IMPORTED_LOCATION
     NAMES ${target_name_shared} ${target_name_static}  # prefer libfoo-version.so over libfoo.a
     PATHS ${${target}_LIBRARY_DIRS}
-    NO_DEFAULT_PATH
```
harrishancock: @intelfx, I thought about this a bit. As written, if the user has capnproto installed in a custom location and relies on the `PKG_CONFIG_PATH` environment variable to find it, pkg-config and CMake's `find_library` will find different installations, because system-default paths are searched before the `PATHS` argument (and the `HINTS` argument, which is the other one we could use). It is difficult to prohibit or even diagnose this situation, so I think we should change this to:

```cmake
find_library(...
  NAMES ...
  PATHS ...
  NO_DEFAULT_PATH
)
find_library(...
  NAMES ...
)
```

That makes the search order explicit, and is a fairly well-recognized CMake idiom.

I'd be happy to update the PR, if that's okay with you?

harrishancock:
Updated to use the NO_DEFAULT_PATH-then-system-dirs CMake idiom, so that `PKG_CONFIG_PATH` is honored correctly for both include/ and lib/.

@pqu, would you take a look when you can?

pqu:
The logic is a little odd, but the change LGTM. Moving to `pkg_check_modules(... IMPORTED_TARGET)` when it becomes feasible would help clean things up.

harrishancock:
Thanks @pqu. I agree regarding `pkg_check_modules(... IMPORTED_TARGET)`. It looks like that was added in CMake 3.6, which is old enough by this point that I think we can switch whenever we want. I'll work on that when I have some time -- probably not in time for the upcoming release, though. :(

PR #686, opened 2018-06-04
Allow hash signs in path and query components in kj::Url request contexts

harrishancock:
I'm not super thrilled about this PR, and wish there were a cleaner solution. My main concern is what happens when a URL is parsed in a request context, then serialized out as a REMOTE_HREF. Since we typically don't percent-decode on URL parsing now in servers, we also don't percent-encode on serialization. Thus, part of the path/query would be lost, instead interpreted as the fragment, unless the calling code anticipated the problem at manually escaped the '#'.

One solution might be to percent-encode such problematic characters during the parsing stage, right before we optionally-percent-decode. This would be similar to what the URL spec does, so we might consider just implementing its behavior.


kentonv:
Yeah I guess...

PR #688, opened 2018-06-07
Chunked body fixes in kj-http

harrishancock:
I ran into these because the userspace pipe's `BlockedWrite` state uses a gather-write in its `pumpTo()` implementation.


kentonv:
Doh.

PR #691, opened 2018-06-11
Add limit to readAllBytes()/readAllText()

harrishancock:
Reading an unbounded amount of data from a stream is a potential DoS vector. To manage this risk, `readAllText()` and `readAllBytes()` now accept a `limit` parameter. For source backwards-compatibility, this limit defaults to `kj::maxValue`.

The presence of this limit might make it seem that `readAllText()`/`readAllBytes()` are now misnamed. If we care, we could add `read*Until()` overloads that accept a mandatory parameter, and deprecate the `readAll*()` versions.

### DIFF
--- a/c++/src/kj/async-io-test.c++
+++ b/c++/src/kj/async-io-test.c++
@@ -681,6 +681,72 @@ kj::Promise<void> expectRead(kj::AsyncInputStream& in, kj::StringPtr expected) {
   }));
 }
 
+class MockAsyncInputStream: public AsyncInputStream {
+public:
+  MockAsyncInputStream(kj::ArrayPtr<const byte> bytes, size_t blockSize)
+      : bytes(bytes), blockSize(blockSize) {}
+
+  kj::Promise<size_t> tryRead(void* buffer, size_t minBytes, size_t maxBytes) override {
+    // Clamp max read to blockSize.
+    size_t n = kj::min(blockSize, maxBytes);
+
+    // Unless that's less than minBytes -- in which case, use minBytes.
+    n = kj::max(n, minBytes);
+
+    // But also don't read more data than we have.
+    n = kj::min(n, bytes.size());
+
+    memcpy(buffer, bytes.begin(), n);
+    bytes = bytes.slice(n, bytes.size());
+    return n;
+  }
+
+private:
+  kj::ArrayPtr<const byte> bytes;
+  size_t blockSize;
+};
+
+KJ_TEST("AsyncInputStream::readAllText() / readAllBytes()") {
+  kj::EventLoop loop;
+  WaitScope ws(loop);
+
+  auto bigText = strArray(kj::repeat("foo bar baz"_kj, 12345), ",");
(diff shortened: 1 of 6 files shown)


c++/src/kj/async-io-test.c++
```diff
@@ -681,6 +681,52 @@ kj::Promise<void> expectRead(kj::AsyncInputStream& in, kj::StringPtr expected) {
...
+  WaitScope ws(loop);
+
+  auto bigText = strArray(kj::repeat("foo bar baz"_kj, 12345), ",");
+  size_t blockSizes[] = { 1, 4, 256, 4096, 8192, bigText.size() };
+
+  for (size_t blockSize: blockSizes) {
+    {
+      KJ_CONTEXT(blockSize);
+      MockAsyncInputStream input(bigText.asBytes(), blockSize);
+      KJ_EXPECT(input.readAllText().wait(ws) == bigText);
```
kentonv: Not testing with a limit here?

Seems like we actually want a two-dimensional test matrix, one to set the stream's block size, and one to choose the read limit? And then test readAllText() and readAllBytes() for each?

harrishancock:
Yep. I had meant to call out the exception-versus-return-the-limit design choice explicitly. Given the naming of the function, you're right that throwing an exception does make more sense. Will fix.

c++/src/kj/async-io-test.c++
harrishancock: That makes sense.

harrishancock:
Updated.

I made it throw an exception if `limit` bytes are read before seeing EOF, rather than throwing an exception if we read more than `limit` bytes. This means `readAllText(0)` will unconditionally throw, and `readAllText(1)` throws if the stream is not at EOF. The alternative requires us to potentially perform an extra read to detect an EOF before finishing the algorithm, or to unconditionally increment the limit by one (making sure to clamp to `kj::maxValue`) before running the algorithm.

Since the only situation I could see where someone would care about the difference in behavior is if they already know in advance how much data they are about to read (and should thus be using `read()` or `tryRead()`), I figured we should prioritize implementation cleanliness.

I also used a three-dimensional test matrix instead of a two-dimensional one: in addition to simulated InputStream block size and read-all limit, I think we need to test when a stream ends at or near a read-all block boundary.

harrishancock:
Cygwin failure looked like a transient connection problem, pushed a new commit to try again.

c++/src/kj/io.h
```diff
@@ -29,6 +29,7 @@
 #include "common.h"
 #include "array.h"
 #include "exception.h"
+#include <cstdint>
```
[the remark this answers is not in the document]
harrishancock: Fixed.

> I believe `<cwhatever>` headers are deprecated.

Other way around actually -- `<whatever.h>` headers were deprecated in C++98. You're right about the namespacing issue ... but we can make a similar argument that we should be wrapping every .h in an `extern "C"` block. Both choices are left to the implementation. :(

PR #696, opened 2018-06-23
Fix an issue with force append -std=gnu++11 to cmake CXX_FLAGS

chronoxor:
target_compile_features(kj PUBLIC cxx_constexpr) force globally append -std=gnu++11 to CXX_FLAGS:

CXX_FLAGS =  -std=gnu++17 -Wall -Werror -O2 -g -DNDEBUG   -std=gnu++11

I replaced it with set(CMAKE_CXX_STANDARD 11)

### DIFF
--- a/c++/src/kj/CMakeLists.txt
+++ b/c++/src/kj/CMakeLists.txt
@@ -66,8 +66,10 @@ set(kj-std_headers
 )
 add_library(kj ${kj_sources})
 add_library(CapnProto::kj ALIAS kj)
-target_compile_features(kj PUBLIC cxx_constexpr)
-# Requiring the cxx_std_11 metafeature would be preferable, but that doesn't exist until CMake 3.8.
+if(NOT CYGWIN)
+  target_compile_features(kj PUBLIC cxx_constexpr)
+  # Requiring the cxx_std_11 metafeature would be preferable, but that doesn't exist until CMake 3.8.
+endif()
 
 if(UNIX AND NOT ANDROID)
   target_link_libraries(kj PUBLIC pthread)


harrishancock:
Hi @chronoxor,

I'm only able to reproduce this issue if the dependent project sets the `-std=gnu++17` flag manually, like one of the following:

```cmake
target_compile_options(target PRIVATE -std=gnu++17)
# or ...
set(CMAKE_CXX_FLAGS -std=gnu++17)
```

If the dependent project itself uses one of the more abstract ways of selecting the C++ standard, like:

```cmake
target_compile_features(target PRIVATE cxx_std_17)
# or ...
set(CMAKE_CXX_STANDARD 17)
# or ...
set_target_property(target PROPERTIES CXX_STANDARD 17)
```

then CMake does the right thing and uses only the flag for the newest required version of C++ requested.

Are you setting the flag manually in your dependent project? If so, can you switch to using one of the abstract methods?

The disadvantage to using `set(CMAKE_CXX_STANDARD 11)` in Cap'n Proto's CMakeLists.txt is that it isn't transitive to dependent projects, so every dependent project must explicitly request at least the version of C++ required by Cap'n Proto, unless their compiler happens to have a modern enough default. This isn't a huge deal, but I'd like to keep things cleanly working out of the box for modern dependent projects if we can.

chronoxor:
I use the following workaround for old cmake in Cygwin (3.6.0) which does not support set(CMAKE_CXX_STANDARD 17):

# Choose C++ standard
if(CMAKE_VERSION VERSION_LESS "3.7")
  if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++17")
  endif()
else()
  set(CMAKE_CXX_STANDARD 17)
endif()

Maybe we can add the following Cygwin workaround instead?

if(NOT CYGWIN)
  target_compile_features(kj PUBLIC cxx_constexpr)
  # Requiring the cxx_std_11 metafeature would be preferable, but that doesn't exist until CMake 3.8.
endif()

Which is also work well for Cygwin build. If you agree I'll update the PR

harrishancock:
> I use the following workaround for old CMake (3.6.0) in Cygwin which does not support set(CMAKE_CXX_STANDARD 17):

Ah, I see now. That's frustrating. I think the ideal answer is "upgrade to CMake 3.7+", but I assume there is a good reason why you have to support 3.6. ;)

What about the following workaround?

```cmake
find_package(CapnProto)

if(CMAKE_VERSION VERSION_LESS "3.7")
  if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
    # Our version of CMake does not have knowledge of C++17, so we need to manually
    # set -std=gnu++17. However, that conflicts with CMake's compile features system.
    # Hack around this by removing CapnProto::kj's compile features.
    set_target_properties(CapnProto::kj PROPERTIES INTERFACE_COMPILE_FEATURES "")
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++17")
  endif()
else()
  set(CMAKE_CXX_STANDARD 17)
endif()
```

If your `find_package(CapnProto)` line is somewhere else, you could move the `set_target_properties()` line after it and it should still work.

If that doesn't work for you, then we can revisit a hack on the Cap'n Proto side.

chronoxor:
> Ah, I see now. That's frustrating. I think the ideal answer is "upgrade to CMake 3.7+", but I assume there is a good reason why you have to support 3.6. ;)

The answer is simple - there is no CMake package greater than 3.6 in Cygwin (https://cygwin.com/ml/cygwin/2018-06/msg00265.html) and native Windows CMake is not suitable for Cygwin.

> What about the following workaround?

Unfortunately, I have CMAKE_CXX_STANDARD global check in the top of lots projects, but Cap'n'proto included as submodule deeper in hierarchy:

```
  # Module subdirectory
  add_subdirectory("capnproto")
```

For this reason I see the minimal evil is to place small Cygwin hack in kj CMakeLists.txt

kentonv:
Doesn't this break the Cygwin build for everyone else, though? Now everyone who uses Cygwin will be required to explicitly specify their C++ standard version rather than rely on `find_package(CapnProto)` doing it automatically.

It seems like you need to find a way to make sure your -std is added to the end of the flag list rather than the beginning.

harrishancock:
> Unfortunately, I have CMAKE_CXX_STANDARD global check in the top of lots projects, but Cap'n'proto included as submodule deeper in hierarchy:

I think that's okay, though. If you set either `CMAKE_CXX_FLAGS` to include `-std=gnu++17` or `CMAKE_CXX_STANDARD` to 17 in the top of your projects, then you can remove the `kj` target's `INTERFACE_COMPILE_FEATURES` property unconditionally right after the `add_subdirectory(capnproto)` command, and it should work with both old and new versions of CMake.

chronoxor:
@harrishancock, your workaround works perfect in my case and does not require any changes in Cap'n'Proto cmake files. I'll close the PR for this reason. Thanks a lot for assistant!

harrishancock:
Glad it works, @chronoxor!

PR #697, opened 2018-06-24
Allow .cmd/.bat files as compiler plugins on Windows.

### LINKED ISSUE #487: Shouldn't send "plugin" to argv[0] on Windows platforms
> // MSVCRT's spawn*() don't correctly escape arguments, which is necessary on Windows
> // since the underlying system call takes a single command line string rather than
> // an arg list. Instead of trying to do the escaping ourselves, we just pass "plugin"
> // for argv[0].

I don't think escaping is necessary here since it'd just be the path. In addition, by doing this, you allow people to create implementations in non-native languages by giving people the opportunity to use .BAT or .CMD files (as the cmd interpreter relies on argv[0] to tell it what batch file to run.)
Alternatively, if it was found on the path, you could pass just the filename.
I made the change locally and it allowed me to write a compiler in LuaJIT via having batch call luajit.exe <script> -- making it so I didn't have to compile a dedicated platform specific binary to do what I needed.

### LINKED ISSUE #695: Can't run capnpc-ts on windows
There is no way to run the capnpc typescript compiler on windows. The reason is that capnp-ts is not a binary (.exe) file but rather a shell script (it's called capnpc-ts.cmd).
It does work on linux because there capnpc-ts has a shebang line, but windows doesn't support this.

The easiest way to work around this would be to spawn "cmd.exe /c capnpc-ts"; this wouldn't break plugins that do provide .exe files.

One might argue that this is a capnp-ts bug and should be reported there -- but there seems to be no reasonable solution that they could implement, since javascript/typescript can't be compiled into a binary.

kentonv:
Fixes #487
Fixes #695

### DIFF
--- a/c++/src/capnp/compiler/capnp.c++
+++ b/c++/src/capnp/compiler/capnp.c++
@@ -540,13 +540,20 @@ public:
 #endif
         }
 
+#if _WIN32
+        // MSVCRT's spawn*() don't correctly escape arguments, which is necessary on Windows
+        // since the underlying system call takes a single command line string rather than
+        // an arg list. We do the escaping ourselves by wrapping the name in quotes. We know
+        // that exeName itself can't contain quotes (since filenames aren't allowed to contain
+        // quotes on Windows), so we don't have to account for those.
+        KJ_ASSERT(exeName.findFirst('\"') == nullptr,
+            "Windows filenames can't contain quotes", exeName);
+        auto escapedExeName = kj::str("\"", exeName, "\"");
+#endif
+
         if (shouldSearchPath) {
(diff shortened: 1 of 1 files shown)


kentonv:
Tested under WINE, not real Windows. But, I was able to reproduce the problem under WINE and verified this fixed it. It would be nice to add some sort of automated test, but I don't understand cmake enough to do it. (In fact, the cmake/Windows build is currently altogether missing anything like capnp-test.sh, which would be the obvious place to add this -- but capnp-test.sh itself would need to be rewritten as a `.cmd` to run on Windows.)

PR #698, opened 2018-06-25
Automatically reduce `Promise<RemotePromise<T>>` -> `RemotePromise<T>`.

### LINKED ISSUE #341: Promise<Promise<T>> to Promise<T> autojoin does not work for RemotePromises.
As per [this documentation](https://github.com/sandstorm-io/capnproto/blob/0fb92820af3c9f31117081b3c08e01066e07ed55/c%2B%2B/src/kj/async.h#L146-L148), I expect the following program to print "in foo()" when it is run:

``` capnp
# test-wait.capnp
@0xdc4d9ce7f7ca0267;

interface Foo {
  foo @0 () -> ();
}
```

``` cpp
// test_wait.c++

#include "capnp/capability.h"
#include "test-wait.capnp.h"
#include <iostream>
#include <kj/async.h>

class FooImpl: public Foo::Server {
  kj::Promise<void> foo(FooContext context) override {
    std::cout << "in foo()\n";
    return kj::READY_NOW;
  }
};

int main() {
  kj::EventLoop loop;
  kj::WaitScope waitScope(loop);
  Foo::Client client(kj::heap<FooImpl>());

  kj::evalLater([&]() {
    return client.fooRequest().send();
  }).wait(waitScope);
}
```

However, the program does not actually print anything, because `foo()` never gets called. To get it to work, we need to wait again!

``` cpp
  kj::evalLater([&]() {
    return client.fooRequest().send();
  }).wait(waitScope).wait(waitScope);
```

kentonv:
Just like `Promise<Promise<T>>` reduces to `Promise<T>`, we should be able to reduce `Promise<RemotePromise<T>>`. Moreover, this reduction should support pipelining, hence the reduced type should itself be `RemotePromise<T>`.

Fixes #341

### DIFF
--- a/c++/src/capnp/capability.h
+++ b/c++/src/capnp/capability.h
@@ -60,6 +60,9 @@ public:
   KJ_DISALLOW_COPY(RemotePromise);
   RemotePromise(RemotePromise&& other) = default;
   RemotePromise& operator=(RemotePromise&& other) = default;
+
+  static RemotePromise<T> reducePromise(kj::Promise<RemotePromise>&& promise);
+  // Hook for KJ so that Promise<RemotePromise<T>> automatically reduces to RemotePromise<T>.
 };
 
 class LocalClient;
@@ -736,6 +739,19 @@ private:
 
 // =======================================================================================
 // Inline implementation details
+
+template <typename T>
+RemotePromise<T> RemotePromise<T>::reducePromise(kj::Promise<RemotePromise>&& promise) {
+  kj::Tuple<kj::Promise<Response<T>>, kj::Promise<kj::Own<PipelineHook>>> splitPromise =
+      promise.then([](RemotePromise&& inner) {
+    return kj::tuple(kj::Promise<Response<T>>(kj::mv(inner)),
+                     PipelineHook::from(kj::mv(inner)));
+  }).split();
+
+  return RemotePromise(kj::mv(kj::get<0>(splitPromise)),
+      typename T::Pipeline(AnyPointer::Pipeline(
+          newLocalPromisePipeline(kj::mv(kj::get<1>(splitPromise))))));
+}
 
 template <typename Params, typename Results>
 RemotePromise<Results> Request<Params, Results>::send() {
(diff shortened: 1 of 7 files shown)


kentonv:
I don't understand MSVC 2015's problem. @harrishancock What do you think about dropping support for VS2015?

harrishancock:
@kentonv sounds great to me!

c++/src/capnp/capability.h
```diff
@@ -737,6 +740,19 @@ struct List<T, Kind::INTERFACE> {
 // =======================================================================================
 // Inline implementation details
 
+template <typename T>
+RemotePromise<T> RemotePromise<T>::reducePromise(kj::Promise<RemotePromise>&& promise) {
+  kj::Tuple<kj::Promise<Response<T>>, kj::Promise<kj::Own<PipelineHook>>> splitPromise =
+      promise.then([](RemotePromise&& inner) {
+    return kj::tuple(kj::Promise<Response<T>>(kj::mv(inner)),
+                     PipelineHook::from(kj::mv(inner)));
```
dwrensha: The double appearance of `kj::mv(inner)` here and on the previous line looks fishy to me. At first glance, I would guess that `inner` is invalid after the first move. Is multiple inheritance guaranteed to do the right thing here? (I.e. first move just takes the `Promise` part, and second move just takes the `Pipeline` part.)

c++/src/capnp/rpc.c++
```diff
@@ -910,15 +910,17 @@ private:
         // ultimately resolves to a local capability. Instead, we'll need to queue the call until
         // the promise resolves.
 
-        auto vpapPromises = fork.addBranch().then(kj::mvCapture(context,
-            [interfaceId,methodId](kj::Own<CallContextHook>&& context,
-                                   kj::Own<ClientHook> resolvedCap) {
+        kj::Tuple<kj::Promise<void>, kj::Promise<kj::Own<PipelineHook>>> vpapPromises =
```
dwrensha: Here and elsewhere you've added type annotations. Does the new implicit conversion make these necessary? Is there a chance that the new conversion will break downstream code (e.g. in Sandstorm)?

kentonv: Oh sorry, please ignore the second commit in this PR. I was trying to figure out why it doesn't build on MSVC 2015, but I think I'm just going to give up on it and require MSVC 2017.

c++/src/capnp/capability.h
harrishancock: I think it's worth a code comment. :) I had to go and read the relevant recipients to convince myself this was okay, too.

PR #700, opened 2018-06-28
Implement annotations to control JSON parsing.

kentonv:
Features:
- Rename any field or enum value for JSON purposes.
- Flatten structs/groups into their parent object, possibly with a prefix.
- Assign a special discriminant field for unions, so that union members can be flattened without ambiguity.
- Also, `JsonValue` (now called `json::Value`) can be used to accept arbitrary JSON somewhere in your object tree.

/cc @RReverser


RReverser:
Nice! I'll review this tomorrow morning.

kentonv:
Doh, forgot to update build files (since I use Ekam).

RReverser:
This looks really good. Just to reconfirm - as far as I understand, `$Json.name` is not checked for uniqueness, so it's even possible to represent unions where variant data is always encoded into the same property with something like:

```capnp
aUnion :union $Json.discriminator("type") {
  foo :group $Json.name("data") {
    fooMember @9 :Text;
  }
  bar :group $Json.name("data") {
    barMember @11 :UInt32;
  }
}
```

and it would encode like

```json
{
  "type": "foo",
  "data": { "fooMember":"whatever" }
}
```

Is that right?

kentonv:
@RReverser To get the JSON you're looking for, you'd need to do this:

    aUnion :union $Json.discriminator("type") {
      foo :group $Json.flatten() {
        data :group { fooMember @9 :Text; }
      }
      bar :group $Json.flatten() {
        data :group { barMember @11 :UInt32; }
      }
    }

I hadn't considered your approach, but it may be worth supporting as well, as it results in better capnp structure. Let me think about it. One problem is that the discriminator value currently uses the JSON name.

kentonv:
I added annotations to tag `Data` fields as base64 or hex.

c++/src/capnp/compat/json.capnp
```diff
@@ -48,45 +48,44 @@ struct JsonValue {
...
-  annotation flatten @0x82d3e852af0336bf (field, group, union): JsonFlattenFlags;
-  # Specifies that an aggregate field should be flattened into its parent.
-  #
-  # In order to flatten a member of a union, the union (or, for an anonymous union, the parent
-  # struct type) must have the $jsonDiscribinator annotation.
+annotation flatten @0x82d3e852af0336bf (field, group, union): FlattenOptions;
+# Specifies that an aggregate field should be flattened into its parent.
+#
+# In order to flatten a member of a union, the union (or, for an anonymous union, the parent
+# struct type) must have the $jsonDiscribinator annotation.
```
harrishancock: /jsonDiscribinator/jsonDiscriminator/

c++/src/capnp/compat/json.c++
```diff
@@ -882,4 +886,430 @@ void JsonCodec::addFieldHandlerImpl(StructSchema::Field field, Type type, Handle
...
+          case JSON_NAME_ANNOTATION_ID:
+            info.name = anno.getValue().getText();
+            break;
+          case JSON_FLATTEN_ANNOTATION_ID:
+            KJ_REQUIRE(type.isStruct(), "only struct types can be flattened", fieldName, typeName);
+            flattened = true;
+            info.prefix = anno.getValue().getStruct().getAs<JsonFlattenFlags>().getPrefix();
+            break;
+          case JSON_DISCRIMINATOR_ANNOTATION_ID:
+            KJ_REQUIRE(fieldProto.isGroup(), "only unions can have discriminator");
```
harrishancock: Does isGroup() being true guarantee that this field is a union, as opposed to a group in a struct? I'm not super familiar with the schema code.

RReverser:
> One problem is that the discriminator value currently uses the JSON name.

Do you think it would be possible to change as part of this PR or as a breaking change in future?

c++/src/capnp/compat/json.c++
kentonv: No but distinguishing a union from a non-union group is not simple so whatever.

c++/src/capnp/compat/json.c++
```diff
@@ -882,4 +886,430 @@ void JsonCodec::addFieldHandlerImpl(StructSchema::Field field, Type type, Handle
...
+        }
+      }
+
+      if (!flattened) {
+        fieldsByName.insert(std::make_pair(info.name, kj::mv(nameInfo)));
+      }
+
+      if (isUnionMember) {
+        unionTagValues.insert(std::make_pair(info.name, field));
+      }
```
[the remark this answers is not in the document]
kentonv: This code got more complicated in my latest update. But in any case it works and I'd rather not refactor further for now.

RReverser:
@kentonv Awesome, thanks!

c++/src/capnp/compat/json.c++
```diff
@@ -882,4 +886,430 @@ void JsonCodec::addFieldHandlerImpl(StructSchema::Field field, Type type, Handle
...
+    auto typeName = schemaProto.getDisplayName();
+
+    if (discriminator == nullptr) {
+      // There are two cases of unions:
+      // * Named unions, which are special cases of named groups. In this case, the union may be
+      //   annotated by annotating the field. In this case, we receive a non-null `discriminator`
+      //   as a constructor parameter, and schemaProto.getAnnotations() must be empty because
+      //   it's not possible to annotate a group's type (becaues the type is anonymous).
+      // * Unnamed unions, of which there can only be one in any particular scope. In this case,
+      //   the parent struct type itself is annotated.
```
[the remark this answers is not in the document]
kentonv: Not really. Oh well.

PR #701, opened 2018-06-30
fix const-correctness on AnyPointer equality checks

antis81:
Stumbled upon this, by writing a gtest like this:

```c++
AnyStruct as1;
AnyStruct as2;
// do stuff with as1 & as2…
EXPECT_EQ(as1, as2); // does not compile because ==/!= operators are expected to be const

// we can workaround this currently by using the "equals" function
EXPECT_EQ(capnp::Equality::Equal, as1.equals(as2));
```

Compiles fine in local build setup (Clang 6.0,Debian 9); Waiting for Travis and Appveyor builds…


kentonv:
Thanks! This looks right to me.

PR #703, opened 2018-07-04
capnp-js-base has moved

sevki:
here https://github.com/capnp-js/capnp-js-base


sevki:
@kentonv does this mean if you merge this do I get a capnproto tshirt?

sevki:
Ha. mate I know how busy you are, don't worry about it :smile: 

:thinking: I think I'll see you all at retreat

PR #706, opened 2018-07-16
Make Promise variadic, and implement joins

kentonv:
Now you can do e.g. `Promise<A, B, C>`.

`.then()` will pass all the values as separate constructor parameters.

A callback passed to `.then()` may return `kj::Tuple<A, B, C>` to produce a `Promise<A, B, C>`.

This is something I've wanted to do for a long time. I really should have done it from the beginning, to be honest. `Promise<void>` could now be expressed as `Promise<>`, and a lot of special-casing of `void` could then go away.

`Promise<A, B, C>::split()` splits into `Tuple<Promise<A>, Promise<B>, Promise<C>>`. (The `split()` method already existed, but previously operated on `Promise<Tuple<A, B, C>>`. Promises of Tuples are no longer permitted, though. Technically this is a breaking change, but `split()` was relatively new and obscure so probably no one knew about it?)

Also, you can join two promises: `Promise<A, B>::join(Promise<C, D>)` returns `Promise<A, B, C, D>`.


kentonv:
Dammit github, why can't you display commits in the order in which they were committed?

PR #708, opened 2018-07-20
Ignore ECHILD in UnixEventPort::ChildSet::checkExits

ohwgiles:
If no child process has changed state, `waitpid()` returns zero. But if there are no more child processes at all, it returns -1 and sets `errno` to `ECHILD`.

Since `checkExits` calls `waitpid()` in a loop, it should probably not throw an exception, and instead ignore `ECHILD`.

+Unit test modified to exhibit the behaviour


kentonv:
Nice catch! Thanks!

(Sorry for slow review; swamped this week.)

PR #709, opened 2018-07-20
Prefer standard attributes and attribute syntax.

ecatmur:
Prefer standard attributes e.g. `[[noreturn]]`, `[[deprecated]]`, etc.

Prefer standard (C++11) attribute syntax for nonstandard attributes e.g.
`[[gnu::noinline]]`, since mixing standard attributes `[[nodiscard]]` with
GNU attribute syntax `__attribute__((noinline))` doesn't always work.

Prefer `[[nodiscard]]` to `__attribute__((warn_unused_result))`, since GCC
allows the former to be suppressed by a cast to `(void)`; see
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66425

### DIFF
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -151,33 +151,78 @@ typedef unsigned char byte;
 #define KJ_ALWAYS_INLINE(...) inline __VA_ARGS__
 // Don't force inline in debug mode.
 #else
+#if defined(__has_cpp_attribute)
+#if __has_cpp_attribute(gnu::always_inline)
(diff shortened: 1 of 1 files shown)


kentonv:
Hi Ed,

I like this change, but it looks like it broke the Clang build.

ecatmur:
Ah, yeah - clang is stricter on placement of the attributes. I'll have to make KJ_WARN_UNUSED_RESULT a wrapper macro, rather than a suffix. Will try to post a fix soonish.

PR #711, opened 2018-07-24
Makefile.am: Remove ldconfig to fix a sandbox violation.

orbea:
During `make install` sandbox will report violations for when ldconfig is run where it will attempt to edit `/etc/ld.so.cache~`.

Modifying the root file system during installation when using `DESTDIR` is highly undesirable and since these commands are not critical I suggest removing them.

More information on sandbox can be found here.

https://wiki.gentoo.org/wiki/Project:Sandbox
```
make[3]: Entering directory '/tmp/capnproto/c++'
ln -sf capnp /tmp/capnproto/usr/local/bin/capnpc
ldconfig < /dev/null > /dev/null 2>&1 || true
 * ACCESS DENIED:  open_wr:      /etc/ld.so.cache~
make[3]: Leaving directory '/tmp/capnproto/c++'
make[2]: Leaving directory '/tmp/capnproto/c++'
make[1]: Leaving directory '/tmp/capnproto/c++'
 * --------------------------- ACCESS VIOLATION SUMMARY ----
 * LOG FILE: "/tmp/sandbox-22967.log"
 *
VERSION 1.0
FORMAT: F - Function called
FORMAT: S - Access Status
FORMAT: P - Path as passed to function
FORMAT: A - Absolute Path (not canonical)
FORMAT: R - Canonical Path
FORMAT: C - Command Line

F: open_wr
S: deny
P: /etc/ld.so.cache~
A: /etc/ld.so.cache~
R: /etc/ld.so.cache~
C: ldconfig
 * ---------------------------------------------------------
```

### DIFF
--- a/c++/Makefile.am
+++ b/c++/Makefile.am
@@ -347,13 +347,8 @@ capnpc_c___SOURCES = src/capnp/compiler/capnpc-c++.c++
 
 # Symlink capnpc -> capnp.  The capnp binary will behave like the old capnpc
 # binary (i.e. like "capnp compile") when invoked via this symlink.
-#
-# Also attempt to run ldconfig, because otherwise users get confused.  If
-# it fails (e.g. because the platform doesn't have it, or because the
-# user doesn't have root privileges), don't worry about it.
 install-exec-hook:
 	ln -sf capnp $(DESTDIR)$(bindir)/capnpc
-	ldconfig < /dev/null > /dev/null 2>&1 || true
 
 uninstall-hook:
 	rm -f $(DESTDIR)$(bindir)/capnpc
@@ -361,7 +356,6 @@ uninstall-hook:
 else LITE_MODE
 
 install-exec-hook:
-	ldconfig < /dev/null > /dev/null 2>&1 || true
 
 endif LITE_MODE
 


orbea:
Yes, there is a way to make sandbox ignore that violation by editing `/etc/sandbox.conf`, however I do not want to do that because its not the correct solution and other programs could then take advantage of this to do more malicious things maybe. If do not want to implement this in capnproto this can be removed easily enough in the distro build process with sed. A configure option could work too, but its kind of silly...

I would suggest adding this to the installation documentation and instruct confused users to read it instead, but its your call.

kentonv:
> I would suggest adding this to the installation documentation and instruct confused users to read it instead, but its your call.

Like I said, it was in the documentation originally (see a86b5f2), but people were still confused.

You don't have to change the sandbox to permit the write, just change it so that it blocks it silently. The makefile doesn't care if `ldconfig` fails -- the overall build/install will still succeed. (Indeed, in your log, it looks like the build finished successfully and the only "problem" is the violation report written thereafter.)

orbea:
I'm not sure sandbox can ignore it, only accept the violation, but yes the build succeeds despite the issue.

kentonv:
OK, I'd prefer not to remove the `ldconfig`, so I'm going to close this. I'd accept a change that somehow avoids the `ldconfig` specifically when this sandbox is in effect and/or when explicitly requested by the user.

Sorry for the inconvenience.

PR #716, opened 2018-08-08
Expose CONNECTION_HEADERS_COUNT in http.h

harrishancock:
I put the constants in the `HttpHeaders` class to avoid polluting the `kj` namespace. This meant that `BuiltinIndices` and `BuiltinIndicesEnum` had to go in `HttpHeaders` as well, though they might be a better logical fit in `HttpHeaderId`. I think it's worth keeping the constants as close to `HttpHeaders::serialize*()` as possible, though.


c++/src/kj/compat/http.h
```diff
@@ -342,6 +342,42 @@ class HttpHeaders {
   // `connectionHeaders`, if provided, contains connection-level headers supplied by the HTTP
   // implementation, in the order specified by the KJ_HTTP_FOR_EACH_BUILTIN_HEADER macro. These
   // headers values override any corresponding header value in the HttpHeaders object.
+  //
+  // For convenience, several constants are defined below which you can use to help keep code up to
```
kentonv: Most people shouldn't be using these methods at all, so I feel like all this extra documentation may be distracting. Maybe just a simple note that "the constants below can help you construct this array" and maybe also preface the whole doc comment with "most applications will not use these methods; they are called by the HTTP client and server implementations".

harrishancock: Done and rebased.

PR #723, opened 2018-08-16
Fix linking during lite builds

### LINKED ISSUE #576: LITE build undefined reference to units.c++
Looks like that the LITE build is now broken. Declaring an object of `MallocMessageBuilder` throws a linker error: 

```
1> libcapnp.lib(layout.obj) : error LNK2019: unresolved external symbol "public: void __thiscall kj::ThrowOverflow::operator()(void)const " (??RThrowOverflow@kj@@QBEXXZ) referenced in function "public: class capnp::Data::Builder __thiscall capnp::_::PointerBuilder::getBlob<struct capnp::Data>(void const *,unsigned int)" (??$getBlob@UData@capnp@@@PointerBuilder@_@capnp@@QAE?AVBuilder@Data@2@PBXI@Z)
```

Edit: It looks like, this is even true for classes derived from `MessageBuilder`. 
Edit 2: Can be fixed by adding `units.c++` to the required build files [here](https://github.com/capnproto/capnproto/blob/master/c%2B%2B/src/kj/CMakeLists.txt#L4)

Crunkle:
Add `units.c++` and `encoding.c++` to allow the linking of `kj::ThrowOverflow::operator()()` and `kj::decodeWideString` respectively. The lite build remains a fair amount smaller than heavy build and has been tested as working on Clang and GCC both.

This is needed in order to fix #576 and other undefined symbols with current CMake setups. I am unsure of the scope of lite builds and their exact exclusions but, without these two sources in the build process, the library seems to have major issues linking to a usable state.

### DIFF
--- a/c++/src/kj/CMakeLists.txt
+++ b/c++/src/kj/CMakeLists.txt
@@ -16,12 +16,12 @@ set(kj_sources_lite
   main.c++
   arena.c++
   test-helpers.c++
+  units.c++
+  encoding.c++
 )
 set(kj_sources_heavy
-  units.c++
   refcount.c++
   string-tree.c++
-  encoding.c++
   time.c++
   filesystem.c++
   filesystem-disk-unix.c++


kentonv:
Thanks!

Looks like @harrishancock is off the hook. :)

harrishancock:
Thanks @Crunkle!

PR #728, opened 2018-08-21
Two fixes for build failure on Android (resubmit)

bazurbat:
Hi! I've noticed that master fails to compile for Android but I needed that to verify some things. Tested with the fixes and it still builds fine on Linux, Mac OS 10.13 and Android using NDK r16b. Linux build still uses aligned_alloc because _ISOC11_SOURCE is actually defined in C++ as well with g++ and clang++ both. Using feature test macros seemed more robust to me than sprinkling `__ANDROID__` everywhere.

### DIFF
--- a/c++/src/kj/filesystem-disk-unix.c++
+++ b/c++/src/kj/filesystem-disk-unix.c++
@@ -72,6 +72,11 @@ namespace {
 #undef SEEK_HOLE
 #endif
 
+#if __ANDROID__
+// no DTTOIF function
+#undef DT_UNKNOWN
+#endif
+
 static void setCloexec(int fd) KJ_UNUSED;
 static void setCloexec(int fd) {
   // Set the O_CLOEXEC flag on the given fd.
--- a/c++/src/kj/filesystem-disk-unix.c++
+++ b/c++/src/kj/filesystem-disk-unix.c++
@@ -72,8 +72,8 @@ namespace {
 #undef SEEK_HOLE
 #endif
 
-#if __ANDROID__
-// no DTTOIF function
+#if __BIONIC__
+// No no DTTOIF function
 #undef DT_UNKNOWN
 #endif
 
--- a/c++/src/kj/table.c++
+++ b/c++/src/kj/table.c++
@@ -288,19 +288,7 @@ void BTreeImpl::growTree(uint minCapacity) {
   // aligned_alloc() function. Unfortunately, many platforms don't implement it. Luckily, there
   // are usually alternatives.
 
-#if _WIN32
-  // Windows lacks aligned_alloc() but has its own _aligned_malloc() (which requires freeing using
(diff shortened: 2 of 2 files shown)


c++/src/kj/filesystem-disk-unix.c++
```diff
@@ -72,6 +72,11 @@ namespace {
 #undef SEEK_HOLE
 #endif
 
+#if __ANDROID__
+// no DTTOIF function
```
harrishancock: Could this be defined if we defined `_GNU_SOURCE` at the top?

c++/src/kj/table.c++
```diff
@@ -288,27 +288,27 @@ void BTreeImpl::growTree(uint minCapacity) {
   // aligned_alloc() function. Unfortunately, many platforms don't implement it. Luckily, there
   // are usually alternatives.
 
-#if __APPLE__
```
kentonv: I'd prefer to add `__ANDROID__` (or perhaps better `__BIONIC__`) here, rather than make posix_memalign() the default. alligned_alloc() is part of the C11 standard, while posix_memalign() is POSIX-specific, so aligned_alloc() is the better default case. If there's a universal way to detect C11, that works too, but _ISOC11_SOURCE seems to be glibc-specific.

c++/src/kj/filesystem-disk-unix.c++
bazurbat: I'm not sure. _GNU_SOURCE has no meaning for Android AFAIK.

bazurbat:
I've found that `__ANDROID__` is defined by toolchain when and can be set by for ex. `-mandroid` option. But `__BIONIC__` is defined in NDK's `sys/cdefs.h`, so it seems to me that it is possible to have `__ANDROID__` without `__BIONIC__` is using upstream GCC for example. Libc will probably be different in this case. For cases in this PR I think `__BIONIC__` is more appropriate.

Actually I think the future-proof way is to have feature testing during configure and use the resulting macros such as `HAVE_POSIX_MEMALIGN` from the generated platform-specific header. Even without configure stage it is more maintainable to write such a header manually and ifdef on feature names in code to avoid specifying a list of OS/libc names for every case. Are you interested in PRs for this?

kentonv:
The problem with configure checks is that autotools is only one of at least three build systems used to build Cap'n Proto. We'd need to replicate the same check to cmake and also figure out what happens to people using ekam. Things are generally a lot easier if we can stick with #ifdefs only.

I don't think there's a lot of value in factoring out platform macros into a separate header unless we end up using posix_memalign or aligned_alloc() somewhere else in the code, forcing us to repeat the same checks. If that happens I'll factor out a header for it at that point.

kentonv:
PR looks good. Thanks!

PR #731, opened 2018-08-26
Fix Android build/test issues.

kentonv:
No real bug fixes here, just:

- Update bitrot in super-test.sh, e.g. GCC is deprecated in favor of Clang on Android.
- Fix warnings due to char being unsigned.
- Fix NaN canonicalization, which only matters for tests doing bitwise comparisons.
- Fix tests using /tmp, which is in a different location on Android.


kentonv:
Just gonna merge again since these are trivial changes that don't really affect other platforms.

PR #735, opened 2018-08-30
Wrap KJ_LOG in braces

bazurbat:
This fixes "dangling else" warnings when KJ_LOG is used after unbraced if statement.

The code such as:
```
if (some_function1())
    KJ_LOG(ERROR, "failed1");
if (some_function2())
    KJ_LOG(ERROR, "failed2");
```
produces warnings about ambiguous or dangling else. I know this form is sometimes frowned upon but I do not think that the library provided macro should impose any particular style on the surrounding code.

### DIFF
--- a/c++/src/kj/debug.h
+++ b/c++/src/kj/debug.h
@@ -131,9 +131,9 @@ namespace kj {
 #define KJ_EXPAND(X) X
 
 #define KJ_LOG(severity, ...) \
-  if (!::kj::_::Debug::shouldLog(::kj::LogSeverity::severity)) {} else \
+{ if (!::kj::_::Debug::shouldLog(::kj::LogSeverity::severity)) {} else \
     ::kj::_::Debug::log(__FILE__, __LINE__, ::kj::LogSeverity::severity, \
-                        "" #__VA_ARGS__, __VA_ARGS__)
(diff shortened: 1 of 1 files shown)


kentonv:
I worry about the fact that this means `KJ_LOG()` lines no longer require a trailing semicolon. If we allow that, then we can't change it later to something that requires a semicolon, because it might break anyone who happened to write a log line without a semicolon in the meantime. That also means we wouldn't be able to extend the logging framework with trailing syntax like `KJ_LOG(...).someMethod();` or `KJ_LOG(...) >> whatever;` in the future, although admittedly I don't have any specific ideas for how we'd use such syntax.

Thoughts?

Maybe we could use a definition like:

    for (bool _kj_shouldLog = ::kj::_::Debug::shouldLog(::kj::LogSeverity::severity); _kj_shouldLog; _kj_shouldLog = false) ::kj::_::Debug::log(...)

This has the same effect as the current if/else-based definition but probably avoids the dangling-else warning?

bazurbat:
I personally always use `;` after statement macros but maybe some people do not like it and try to avoid this if they can. Having compiler error on invalid usage is definitely a win, so I've changed `KJ_LOG` to use for statement as you've suggested.

kentonv:
Sorry for the delay. This looks great, thanks!

PR #738, opened 2018-08-30
Allow immovable types in Maybe

RReverser:
Immovable types can be already wrapped into `Maybe`, which can be constructed as default `nullptr`, set to a constructed value via `.emplace()`, but couldn't be set back to `nullptr` without this fix.

### DIFF
--- a/c++/src/kj/common-test.c++
+++ b/c++/src/kj/common-test.c++
@@ -45,6 +45,11 @@ struct ImplicitToInt {
   }
 };
 
+struct Immovable {
+  Immovable() = default;
+  KJ_DISALLOW_COPY(Immovable);
+};
+
 TEST(Common, Maybe) {
   {
     Maybe<int> m = 123;
@@ -215,6 +220,16 @@ TEST(Common, Maybe) {
     } else {
       ADD_FAILURE();
     }
+  }
+
+  {
+    // Test usage of immovable types.
+    Maybe<Immovable> m;
+    KJ_EXPECT(m == nullptr);
+    m.emplace();
+    KJ_EXPECT(m != nullptr);
+    m.clear();
+    KJ_EXPECT(m == nullptr);
   }
 }
 
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -897,9 +897,7 @@ public:
       noexcept(noexcept(instance<T&>().~T()))
 #endif
   {
-    if (isSet) {
-      dtor(value);
-    }
+    clear();
   }
 
   inline T& operator*() & { return value; }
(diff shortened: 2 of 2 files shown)


c++/src/kj/common.h
```diff
@@ -895,6 +895,7 @@ class NullableValue {
 #endif
   {
     if (isSet) {
+      isSet = false;
```
kentonv: I suspect the compiler may be allowed to optimize away this assignment, since it can observe that once the destructor completes, the object has become uninitialized memory and its contents may as well be anything. So I think the refactoring you did here may not be safe, unfortunately.

Can we reduce this PR to just add the operator=, which should do `dtor(value)` instead of `dtor(*this)`?

c++/src/kj/common-test.c++
```diff
@@ -216,6 +221,22 @@ TEST(Common, Maybe) {
       ADD_FAILURE();
     }
   }
+
+  {
+    // Test usage of immovable types.
+    Maybe<Immovable> m;
+    KJ_IF_MAYBE(v, m) {
```
kentonv: The CI is failing due to `v` being unused and warnings being treated as errors...

c++/src/kj/common.h
RReverser: Ugh. I don't know if you're right, but that sounds like something C++ could do indeed. I'll revert it.

RReverser:
Ok reverted changes. I think it should be actually safe to replace all these `if (isSet) { isSet = false; ... }` with `this = nullptr` instead of `kj::dtor(*this)` which I tried before, but probably won't try again to keep this PR minimal.

RReverser:
What do you want from me this time, GCC...

kentonv:
Note that there's one more specialization of Maybe<Own<T>> in memory.h.

RReverser:
@kentonv Doesn't that one already do the right thing on `nullptr` given that `Own` is movable?

kentonv:
@RReverser Good point, but then why did you update `Maybe<T&>`? :)

RReverser:
@kentonv I didn't? At least not specifically, I updated generic `Maybe` class.

kentonv:
You're right, I misread. Never mind me.

RReverser:
Ok so looks it's something specific to gcc-5 but got fixed in gcc-7. @kentonv any ideas about what is that issue? Or can we drop gcc-5 as a requirement?

RReverser:
Oh, apparently it's occasionally happening on master too and unrelated to my changes: https://travis-ci.org/capnproto/capnproto/builds/426766381. It started with https://github.com/capnproto/capnproto/pull/741 but somewhat flappy, judging by status indicators on commit list: https://github.com/capnproto/capnproto/commits/master

RReverser:
So I added a temporary test and... this looks fun:

```c++
  {
    Maybe<bool> m;
    EXPECT_TRUE(m == nullptr);
    m = false;
    EXPECT_FALSE(m == nullptr); // fails here on GCC 5
  }
```

Apparently `m = false` somehow manages to hit `operator=(decltype(nullptr))` specialisation in GCC 5?

RReverser:
Ohh... It probably requires all the specialisations that constructor has (for `T&&`, `T&`, ...) so that it doesn't choose `nullptr` over conversions. Ugh. I'd really love to just drop it rather than add all that noise...

RReverser:
Superseded by #747.

PR #747, opened 2018-09-26
Overload Maybe::operator=() to match all copy constructors.

kentonv:
Previously we depended on the compiler to implicitly invoke the copy constructor followed by the assignment operator, but this can lead to the compiler making surprising decisions in some cases, such as deciding that the literal 0 is a better match for nullptr than for T. Overloading operator=() to match every copy constructor is the safe thing to do.

### DIFF
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -870,7 +870,7 @@ class NullableValue {
   // boolean flag indicating nullness.
 
 public:
-  inline NullableValue(NullableValue&& other) noexcept(noexcept(T(instance<T&&>())))
+  inline NullableValue(NullableValue&& other)
       : isSet(other.isSet) {
     if (isSet) {
       ctor(value, kj::mv(other.value));
@@ -922,8 +922,8 @@ public:
     return value;
   }
 
-  inline NullableValue() noexcept: isSet(false) {}
-  inline NullableValue(T&& t) noexcept(noexcept(T(instance<T&&>())))
+  inline NullableValue(): isSet(false) {}
+  inline NullableValue(T&& t)
       : isSet(true) {
     ctor(value, kj::mv(t));
   }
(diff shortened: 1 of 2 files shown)


c++/src/kj/common.h
```diff
@@ -1092,10 +1138,36 @@ class Maybe {
...
+  inline Maybe& operator=(T& other) { ptr = other; return *this; }
+  inline Maybe& operator=(const T& other) { ptr = other; return *this; }
+  inline Maybe& operator=(const T* other) { ptr = other; return *this; }
+
   inline Maybe& operator=(Maybe&& other) { ptr = kj::mv(other.ptr); return *this; }
   inline Maybe& operator=(Maybe& other) { ptr = other.ptr; return *this; }
   inline Maybe& operator=(const Maybe& other) { ptr = other.ptr; return *this; }
 
+  template <typename U>
+  Maybe& operator=(Maybe<U>&& other) noexcept(noexcept(T(instance<U&&>()))) {
```
RReverser: Why did you keep `noexcept` just here?

c++/src/kj/common.h
```diff
@@ -1092,10 +1138,36 @@ class Maybe {
...
+      ptr.emplace(kj::mv(*val));
+    } else {
+      ptr = nullptr;
+    }
+    return *this;
+  }
+  template <typename U>
+  Maybe& operator=(const Maybe<U>& other) {
+    KJ_IF_MAYBE(val, kj::mv(other)) {
+      ptr.emplace(other);
```
harrishancock: This calls `NullableValue::emplace(const Maybe<U>&)`, which seems a bit surprising. Also, maybe no point in `kj::mv()`ing the const reference?

kentonv: Indeed, this code is nonsense. Fixed.

PR #751, opened 2018-10-10
Zero-length HTTP responses get no Content-Length header

harrishancock:
There is currently no way to explicitly omit a Content-Length/Transfer-Encoding header on an HTTP response. This is awkward for a proxy, which would ideally pass along responses as-is, even if they have no such headers.

This change allows an author to pass zero as the expected body length to HttpService::Response::send() to mean "do not set any body header." This means that a proxy might strip Content-Length: 0 headers, but will never add a Content-Length header where there was none before.

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -1603,6 +1603,10 @@ kj::Own<kj::AsyncInputStream> HttpInputStreamImpl::getEntityBody(
       kj::Maybe<uint64_t> length;
       KJ_IF_MAYBE(cl, headers.get(HttpHeaderId::CONTENT_LENGTH)) {
         length = strtoull(cl->cStr(), nullptr, 10);
+      } else if (headers.get(HttpHeaderId::TRANSFER_ENCODING) == nullptr) {
+        // HACK: Neither Content-Length nor Transfer-Encoding header in response to HEAD request.
+        //   Propagate this fact with a 0 expected body length.
+        length = uint64_t(0);
       }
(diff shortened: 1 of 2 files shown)


harrishancock:
A more targeted approach we could take might be to only allow this semantic on responses to HEAD requests.

kentonv:
This should definitely only apply to HEAD requests. It's incorrect to omit content-length and transfer-encoding on responses where a body is expected.

c++/src/kj/compat/http.c++
```diff
@@ -4384,8 +4388,14 @@ private:
     if (statusCode == 204 || statusCode == 205 || statusCode == 304) {
       // No entity-body.
     } else KJ_IF_MAYBE(s, expectedBodySize) {
-      lengthStr = kj::str(*s);
-      connectionHeaders[HttpHeaders::BuiltinIndices::CONTENT_LENGTH] = lengthStr;
+      // HACK: We interpret a zero-length expected body length on responses to HEAD requests to mean
+      //   "don't set a Content-Length header at all." This provides a way to omit a body header on
+      //   HEAD responses with non-null-body status codes. This is a hack that *only* makes sense
+      //   for HEAD responses.
+      if (method != HttpMethod::HEAD || *s > 0) {
```
kentonv: Seems like it would be cleaner to edit the condition on line 4407, below, which is already specific to head requests?

kentonv: Eh I guess not really. Whatever, we'll probably want to find a better answer later anyway.

PR #757, opened 2018-10-26
Don't escape forward slashes in JSON output

Timmmm:
Forward slashes don't need to be escaped in JSON.

Not tested but I assume you have some kind of CI system (and it's a trivial change anyway).


kentonv:
Huh, apparently JSON allows but does not require escaping `/`, perhaps so that it can be safely placed in a `<script></script>` block. I agree that this probably shouldn't happen by default.

kentonv:
(Thanks!)

PR #765, opened 2018-11-29
Multi-host HttpClient shouldn't rewrite URL.

kentonv:


kentonv:
This is currently causing the Workers preview service to silently coalesce consecutive `/`s and other minor rewrites on outgoing subrequests, whereas this behavior doesn't actually happen in production (which doesn't use the multi-host HttpClient).

PR #769, opened 2018-12-07
Fix overflow in PackedInputStream.

simon-struk:
The overflow happens e.g. when using an ArrayInputStream over an array larger than max unit32.


kentonv:
Thanks!

PR #771, opened 2018-12-10
Fix cancellation in AsyncTee::PumpSink.

kentonv:


kentonv:
@harrishancock looked at this code last week. Merging.

PR #775, opened 2018-12-11
Check syscall result for sigqueue

RReverser:
Gracefully handle unimplemented `sigqueue` syscall (e.g. on WSL).

### DIFF
--- a/c++/src/kj/async-unix-test.c++
+++ b/c++/src/kj/async-unix-test.c++
@@ -99,12 +99,15 @@ TEST(AsyncUnixTest, SignalWithValue) {
   union sigval value;
   memset(&value, 0, sizeof(value));
   value.sival_int = 123;
-  sigqueue(getpid(), SIGURG, value);
-
-  siginfo_t info = port.onSignal(SIGURG).wait(waitScope);
-  EXPECT_EQ(SIGURG, info.si_signo);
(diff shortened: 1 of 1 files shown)


c++/src/kj/async-unix-test.c++
```diff
@@ -99,12 +99,15 @@ TEST(AsyncUnixTest, SignalWithValue) {
...
-
-  siginfo_t info = port.onSignal(SIGURG).wait(waitScope);
-  EXPECT_EQ(SIGURG, info.si_signo);
-  EXPECT_SI_CODE(SI_QUEUE, info.si_code);
-  EXPECT_EQ(123, info.si_value.sival_int);
+  KJ_SYSCALL_HANDLE_ERRORS(sigqueue(getpid(), SIGURG, value)) {
+    case ENOSYS:
+      break;
+    default:
+      siginfo_t info = port.onSignal(SIGURG).wait(waitScope);
```
kentonv: The `default` case gets executed in case there *was* an error, but it wasn't `ENOSYS`... I'll suggest a change here.

kentonv: Apparently "suggest change" doesn't easily allow for multi-line changes, so I added a commit... :/

RReverser: Thanks!

Issue #458, opened 2017-04-25
Doesn't work under Windows Subsystem for Linux


realuser:
In Windows 10 64-bit I have enabled [Windows Subsystem for Linux (WSL)](https://msdn.microsoft.com/en-us/commandline/wsl/about), which gives me Bash on Ubuntu 14.04 LTS on Windows.

I'm following these steps:
```
curl -O https://capnproto.org/capnproto-c++-0.5.3.1.tar.gz
tar zxf capnproto-c++-0.5.3.1.tar.gz
cd capnproto-c++-0.5.3.1
./configure
make -j6 check
```
It appears to compile fine but then tests seem to fail.

FAIL: capnp-test
PASS: capnp-evolution-test
PASS: src/capnp/compiler/capnp-test.sh

[test-suite.log.txt](https://github.com/sandstorm-io/capnproto/files/954459/test-suite.log.txt)

looking at test-suite.log, it appears the following tests fail:

```
[----------] Global test environment tear-down
[==========] 407 tests from 45 test cases ran. (69 ms total)
[  PASSED  ] 373 tests.
[  FAILED  ] 34 tests, listed below:
[  FAILED  ] AsyncUnixTest.Signals
[  FAILED  ] AsyncUnixTest.SignalWithValue
[  FAILED  ] AsyncUnixTest.SignalWithPointerValue
[  FAILED  ] AsyncUnixTest.SignalsMultiListen
[  FAILED  ] AsyncUnixTest.SignalsMultiReceive
[  FAILED  ] AsyncUnixTest.SignalsAsync
[  FAILED  ] AsyncUnixTest.SignalsNoWait
[  FAILED  ] AsyncUnixTest.ReadObserver
[  FAILED  ] AsyncUnixTest.ReadObserverMultiListen
[  FAILED  ] AsyncUnixTest.ReadObserverMultiReceive
[  FAILED  ] AsyncUnixTest.ReadObserverAsync
[  FAILED  ] AsyncUnixTest.ReadObserverNoWait
[  FAILED  ] AsyncUnixTest.WriteObserver
[  FAILED  ] AsyncUnixTest.SteadyTimers
[  FAILED  ] AsyncUnixTest.Wake
[  FAILED  ] AsyncIo.SimpleNetwork
[  FAILED  ] AsyncIo.AddressParsing
[  FAILED  ] AsyncIo.OneWayPipe
[  FAILED  ] AsyncIo.TwoWayPipe
[...]

PR #777, opened 2018-12-11
Use fstatat instead of faccessat to check if file exists

RReverser:
This is not strictly equivalent, but should be interchangeable for most usecases, and fixes a test on WSL where calling `exists` on a symlink to non-existing directory was reported as existing by `faccessat`.

See (4) in https://github.com/capnproto/capnproto/issues/458#issuecomment-446344395

### DIFF
--- a/c++/src/kj/filesystem-disk-unix.c++
+++ b/c++/src/kj/filesystem-disk-unix.c++
@@ -711,12 +711,13 @@ public:
   }
 
   bool exists(PathPtr path) const {
-    KJ_SYSCALL_HANDLE_ERRORS(faccessat(fd, path.toString().cStr(), F_OK, 0)) {
+    struct stat stats;
+    KJ_SYSCALL_HANDLE_ERRORS(fstatat(fd, path.toString().cStr(), &stats, 0)) {
       case ENOENT:
       case ENOTDIR:
         return false;
       default:
-        KJ_FAIL_SYSCALL("faccessat(fd, path)", error, path) { return false; }
+        KJ_FAIL_SYSCALL("fstatat(fd, path)", error, path) { return false; }
     }
     return true;
   }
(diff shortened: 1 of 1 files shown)


kentonv:
This probably makes the code slower, though, since the filesystem may need to do additional reads to obtain the stats, which we then just throw away...

RReverser:
Do you think `exists` (as opposed to actual read / write) is ever in a path hot enough to notice a difference?

kentonv:
Hmm, I think `faccessat()` doesn't need to check permissions on the target file for `F_OK`, only for `R_OK` and `W_OK`.

I can imagine `exists()` being pretty hot in certain use cases. For example, imagine something like an overlay filesystem which needs to check several layers of overlay looking for a file. (An import/include path search for a compiler is similar but probably not performance-sensitive.)

Maybe I'm being pedantic, though it seems disappointing to make this change just to work around a test case that doesn't really matter, on a platform (WSL) that isn't our main target.

RReverser:
Yeah normally I would just ignore the test if this would be an unsupported feature, but given that it actually reveals an incorrect result on the mentioned platform, it felt right to fix the method instead to work everywhere.

Another fact that led me to this change is that `if [[ -e ... ]]` / `test -e` in Bash are also using `stat` and not `access`. Not sure if they also had platform-specific differences in mind or some other reason, but given that it works on all platforms, it felt right to align with their behaviour.

vlovich:
Not a conclusive test but this microbenchmark (https://quick-bench.com/q/FZzvXcQ1f1eO9_WMCmtBr91QB68) seems to indicate no performance difference on whatever cloud machine it ends up getting run on.

kentonv:
@vlovich Your code has a typo -- you run the faccessat benchmark twice. Fixing it shows fstatat is slightly slower: https://quick-bench.com/q/sfsV5O7ecGkaP0CbJHNhoK_2cIE

It's not a very significant difference. But also, this test will always hit cache, so it wouldn't tell us if fstatat requires additional disk reads.

That said, it _probably_ isn't a big deal. But I didn't like the idea of taking that risk just to make a test case that probably doesn't matter pass in the presence of WSL bugs...

PR #781, opened 2019-01-01
Replace AsyncTee try-catches with kj::evalNow() / add commentary

### LINKED ISSUE #767: Replace try-catches in `kj::newTee()` with `kj::evalNow()`
They won't compile with -fno-exceptions. Also, add commentary explaining why PumpSink::fill() promises must never be rejected.

harrishancock:
Closes #767


harrishancock:
Obviated by @kentonv in https://github.com/capnproto/capnproto/pull/966, which does not segfault. :)

PR #806, opened 2019-03-16
Add HttpClient wrapper for limiting concurrent connections

jclee:
WIP... Seems to work OK, but I still need to make it use PromiseOutputStream instead of the pipe.

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -27,6 +27,7 @@
 #include <stdlib.h>
 #include <kj/encoding.h>
 #include <deque>
+#include <queue>
 #include <map>
 
 namespace kj {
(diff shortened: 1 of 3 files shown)


c++/src/kj/compat/http.c++
```diff
@@ -3951,6 +3952,149 @@ kj::Own<HttpClient> newHttpClient(kj::Timer& timer, HttpHeaderTable& responseHea
...
+  struct ConnectionCounter;
+
+  kj::HttpClient& inner;
+  uint maxConcurrentRequests;
+  uint concurrentRequests = 0;
+
+  std::queue<kj::Own<kj::PromiseFulfiller<kj::Own<ConnectionCounter>>>> pendingRequests;
+  // TODO(someday): want maximum cap on queue size?
+
+  struct ConnectionCounter final {
```
kentonv: I recommend adding:

    KJ_DISALLOW_COPY(ConnectionCounter);

just to be safe.

FWIW, another way to design this would be to make ConnectionCounter a movable type, with `parent` being a pointer that becomes null when moved away. Then it wouldn't need to be allocated on the heap. However, the code would be somewhat more complicated and the performance gain probably insignificant... so meh.

c++/src/kj/compat/http.c++
```diff
@@ -3961,54 +3961,51 @@ public:
...
       return { kj::mv(request.body), kj::mv(promise) };
     }
 
-    auto paf = kj::newPromiseAndFulfiller<kj::Own<ConnectionCounter>>();
+    auto paf = kj::newPromiseAndFulfiller<ConnectionCounter>();
     auto urlCopy = kj::str(url);
     auto headersCopy = headers.clone();
-    // TODO(now): Use PromiseOutputStream instead
-    auto pipe = newOneWayPipe(expectedBodySize);
+    auto outPaf = kj::newPromiseAndFulfiller<kj::Own<AsyncOutputStream>>();
```
kentonv: Ah, so here's a good thing to keep in mind about KJ promises.

A promise/fulfiller pair has a fundamental problem compared to other promises in that it doesn't own the task it is waiting for. This can be problematic in two ways:
1. If whoever owns the task cancels it, the fulfiller will be destroyed without fulfilling anything, which causes the promise at the other end to throw.
2. The caller is not able to cancel the task by destroying it.

For this particular use case, the caller receives Request::response and Request::body, but Request::response owns all related tasks, and Request::body is just an observer. If the caller discards Request::response, then Request::body will unexpectedly also fail. (Admittedly it's not clear why a caller would want to do that, but it's still somewhat unexpected.)

Instead of using a PromiseFulfillerPair here, a better approach is to split() or fork() the promise. This allows both Request::response and Request::body to own the task via reference counting, so that it is only canceled if both are discarded. Check out how PromiseNetworkAddressHttpClient::request() (the other place where PromiseOutputStream is used) does this: it first creates a `Promise<Tuple<T, U>>` and then uses `promise.split()` to split it into two promises that can go separate ways.

jclee: Thanks!  I managed to get it working by following the split pattern you mentioned.  I tried to write a test that would throw an exception when using the stream after releasing the response promise for the old implementation.  However, with the new implementation that test would just block indefinitely, so the best I was able to do was detect the difference via promise.poll().

I think the reason the test was blocking indefinitely was because releasing the response promise still left the corresponding fulfiller in the queue of pending requests, and it wouldn't get cleaned up until enough connections freed up that its entry reached the front of the queue.  This might sort of work in a running system, but it would introduce some amount of unnecessary waiting, depending on how common request cancellation is.

It seems like a better approach would be to somehow detect when the response promise is prematurely released and preemptively remove its entry from the queue.  Does that sound right?

kentonv: Interesting. This sounds like another example of how PromiseFulfillerPairs create complications for cancellation.

In theory the "clean" solution here is to use a PromiseAdapter -- see newAdaptedPromise(). You would create an adapter that registers itself in the queue, but its destructor can remove itself from the queue in the case that it was canceled before starting.

However, I don't think this is worth the effort. Most cases of PromiseFulfillerPair are technically cleaner as PromiseAdapters, but PromiseAdapters are a lot more work to write, so we usually don't bother.

I'm also not sure if this would actually fix the hang you describe. I'm not sure exactly what code you wrote, but my guess is that the test is actually hanging because Request::body was never destroyed, which effectively holds open the request, which then rightly blocks subsequent requests?

jclee: The test was basically:

```
  // Using body stream after releasing blocked response promise throws no exception
  auto req3 = doRequest();
  {
    kj::Own<kj::AsyncOutputStream> req4Body;
    {
      auto req4 = client->request(HttpMethod::GET, kj::str("/", ++i), HttpHeaders(headerTable));
      io.waitScope.poll();
      req4Body = kj::mv(req4.body);
    }
    auto writePromise = req4Body->write("a", 1);
    writePromise.wait(io.waitScope); // Exception with old code, blocks with new code
  }
  req3.wait(io.waitScope);
```

Yeah, in this case the body is still alive, so the request is still open, but I'm not sure how to otherwise exercise the throwing behavior expected when the underlying PromiseAndFulfiller was closed in the old code.  You have to execute some operation on the body and wait for its completion, and to do that the body needs to be held alive for the duration of the operation, right?

kentonv: Hmm, I don't understand why that code would hang, unless the whole request is blocked waiting for a previous request to finish?

jclee: Yes, in this case the service is limited to one open connection, and req3 is already using it, so req4 gets queued.  So req4's body's PromiseOutputStream's promise is not going to be fulfilled until a connection is freed, even though req4's response promise has already been freed, so it blocks.  If this is the expected behavior, I guess we're good. :)

kentonv: Yes, that seems entirely expected to me.

PR #809, opened 2019-03-20
Jlee/fix mingw warnings

jclee:
Just resolving some mingw warnings, since the build otherwise seems to run cleanly.

### DIFF
--- a/c++/src/kj/filesystem-disk-win32.c++
+++ b/c++/src/kj/filesystem-disk-win32.c++
@@ -873,6 +873,7 @@ public:
             // Retry, but make sure we don't try to create the parent again.
             return tryReplaceNode(path, mode - WriteMode::CREATE_PARENT, kj::mv(tryCreate));
           }
+          // fallthrough
         default:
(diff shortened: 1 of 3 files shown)


c++/src/kj/filesystem-disk-win32.c++
```diff
@@ -31,6 +31,10 @@
 
 // Request Vista-level APIs.
 #define WINVER 0x0600
+
+#ifdef _WIN32_WINNT
```
kentonv: Hmm, maybe the #define _WIN32_WINNT should be moved above all the other includes, to make sure no system header defines it first?

jclee: Sure... I was just making the change that would preserve existing behavior, but doing the define earlier might prevent other conflicts.  I wasn't sure if something else was unconditionally defining the value, but it looks like the conflicting definition was in _mingw.h, which seems compatible with moving the definition earlier:

https://github.com/Alexpux/mingw-w64/blob/master/mingw-w64-headers/crt/_mingw.h.in#L225

PR #820, opened 2019-04-18
Fulfill AsyncPipe::tryPumpFrom() on EOF

harrishancock:
Previously if the input stream being pumped from encountered an EOF while the other side of the pipe was being pumped elsewhere, we did not fulfill the tryPumpFrom() promise, leaving it hanging.

### DIFF
--- a/c++/src/kj/async-io-test.c++
+++ b/c++/src/kj/async-io-test.c++
@@ -1375,6 +1375,51 @@ KJ_TEST("Userland pipe pumpFrom EOF on abortRead()") {
   pipe2.out = nullptr;
 }
 
+KJ_TEST("Userland pipe EOF fulfills pumpFrom promise") {
+  kj::EventLoop loop;
+  WaitScope ws(loop);
+
+  auto pipe = newOneWayPipe();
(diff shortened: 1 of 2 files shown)


kentonv:
I don't think this is correct. `pumpTo()` is not supposed to propagate EOF. You can have two input streams and pump each of them, in sequence, to the same output stream.

So the outbound pump from a pipe should not complete just because an inbound pump to the pipe completed. There may be more subsequent inbound pumps coming. The outbound pump should only complete when the inbound side of the pipe is destroyed (which is, unfortunately, how we signal EOF at present).

kentonv:
Actually I may have confused myself.

In your test, `pumpPromise` should complete but `pumpPromise2` should not, AFAICT. But I think your change affects only the first (making it correct), and the second may be a preexisting bug?

harrishancock:
I concur -- it surprised me that `pumpPromise2` fulfilled earlier than I expected. I'm glad you read the test and confirmed.

I amended the commit and force-pushed to address the second bug.

kentonv:
Awesome.

PR #821, opened 2019-04-19
Implement FD passing in Cap'n Proto. 

kentonv:
An endpoint (subclass of `Capability::Server`) may override `kj::Maybe<int> getFd()` to expose an underlying file descriptor.

A remote client may use `Capability::Client::getFd()` on the endpoint's capability to get that FD.

The client and server must explicitly opt into FD passing by passing a max-FDs-per-message limit to TwoPartyVatNetwork and using Unix sockets as the transport. Nothing other than that is needed.

cc @a-robinson


kentonv:
Mac build apparently hangs during tests, which of course means the test output doesn't even get printed because autotools "conveniently" hides it in a log file. Ugh. Might need to dig up a Mac machine to debug...

MinGW build seems to be flaky, unrelated to this change.

c++/src/capnp/capability.h
```diff
@@ -206,6 +206,19 @@ class Capability::Client {
   // Make a request without knowing the types of the params or results. You specify the type ID
   // and method number manually.
 
+  kj::Promise<kj::Maybe<int>> getFd();
+  // If the capability's server implemented Capability::Server::getFd() returning non-null, and all
+  // RPC links between the client and server support FD passing, returns a file descriptor pointing
+  // to the same undelying file description as the server did. Returns null if the server provided
```
dwrensha: s/undelying/underlying/

c++/src/capnp/rpc.c++
```diff
@@ -784,7 +799,8 @@ private:
                    kj::Array<PipelineOp>&& ops)
         : RpcClient(connectionState), questionRef(kj::mv(questionRef)), ops(kj::mv(ops)) {}
 
-   kj::Maybe<ExportId> writeDescriptor(rpc::CapDescriptor::Builder descriptor) override {
+   kj::Maybe<ExportId> writeDescriptor(rpc::CapDescriptor::Builder descriptor,
+                                       kj::Vector<int>& fds) override {
```
a-robinson: I assume it's intentional that these go unused here?

c++/src/capnp/rpc.capnp
```diff
@@ -988,6 +988,63 @@ struct CapDescriptor {
     # Level 1 and 2 implementations that receive a `thirdPartyHosted` may simply send calls to its
     # `vine` instead.
   }
+
+  attachedFd @6 :UInt8 = 0xff;
+  # If the RPC message in which this CapDescriptor was delivered also had file descirptors
```
a-robinson: s/descirptors/descriptors

c++/src/capnp/rpc-twoparty.c++
```diff
@@ -22,12 +22,13 @@
...
 #include "serialize-async.h"
 #include <kj/debug.h>
+#include <kj/io.h>
 
 namespace capnp {
 
 TwoPartyVatNetwork::TwoPartyVatNetwork(kj::AsyncIoStream& stream, rpc::twoparty::Side side,
                                        ReaderOptions receiveOptions)
-    : stream(stream), side(side), peerVatId(4),
+    : stream(&stream), maxFdsPerMesage(0), side(side), peerVatId(4),
```
a-robinson: It seems a little dangerous to save a pointer to something that was only passed in as a reference, no? Given the lack of guarantees around the lifetime of what was passed in.

Is this just because it'd be a lot of work to update all the callers, or is there some other reason for not making this more explicit?

c++/src/kj/async-io.h
```diff
@@ -175,15 +175,57 @@ class AsyncCapabilityStream: public AsyncIoStream {
...
+
+  struct ReadResult {
+    size_t byteCount;
+    size_t capCount;
+  };
+
+  virtual Promise<ReadResult> tryReadWithFds(void* buffer, size_t minBytes, size_t maxBytes,
+                                             AutoCloseFd* fdBuffer, size_t maxFds) = 0;
+  // Read data from the stream that may have file descriptors attached. Any attached descriptors
+  // will be added to `fds`. If multiple bundles of FDs are encountered in the course of reading
```
a-robinson: Added to `fds`, or added to `fdBuffer`?

c++/src/kj/async-io.c++
```diff
@@ -1816,6 +1835,35 @@ Promise<Own<AsyncCapabilityStream>> AsyncCapabilityStream::receiveStream() {
   });
 }
 
+kj::Promise<Maybe<Own<AsyncCapabilityStream>>> AsyncCapabilityStream::tryReceiveStream() {
+  struct ResultHolder {
+    byte b;
+    Own<AsyncCapabilityStream> stream;
+  };
+  auto result = kj::heap<ResultHolder>();
+  auto promise = tryReadWithStreams(&result->b, 1, 1, &result->stream, 1);
```
a-robinson: You don't need to do anything about it, but passing the same un-named, un-commented literal into a function 3 times is making me sad :(

c++/src/kj/async-io-test.c++
```diff
@@ -233,6 +237,187 @@ TEST(AsyncIo, CapabilityPipe) {
...
+  if (n < 0) {
+    KJ_FAIL_ASSERT("out2 was not closed");
+  }
+  KJ_ASSERT(n == 0);
+
+  // Third pipe should have been closed implicitly because we didn't provide space to receive it.
+  KJ_NONBLOCKING_SYSCALL(n = read(in3, buffer, 4));
+  if (n < 0) {
+    KJ_FAIL_ASSERT("out3 was not closed. If ScmRightsTruncatedOdd passed but this test failed "
+        "then your operating system kernel is buggy and leakds file descriptors when an "
```
a-robinson: s/leakds/leaks

c++/src/capnp/rpc.c++
kentonv: Yes, a PipelineClient represents a capability that is expected to be returned by an RPC call which is at present still in-flight. So we clearly can't yet be aware of any file descriptors attached to it.

c++/src/capnp/rpc-twoparty.c++
kentonv: The lifetime requirements haven't changed here. We just changed from storing a reference to storing a pointer (because `kj::OneOf<>` can only store reassignable types -- something I should maybe fix).

It sounds like you might be applying a Google style rule regarding when to use references vs. pointers, but KJ style doesn't buy into the same rules.

KJ almost always prefers references over pointers, because (1) the compiler forces all references to be initialized, and (2) they are not allowed to be set to null. For nullable references, we use `kj::Maybe<T&>`. KJ code only uses raw pointers in cases where the value really must be assignable, but I feel bad about it every time.

Google's style, in contrast, says basically never to use references except for const references used as an optimization to elide a copy. The alleged advantage of Google's style is that the calling code sees an `&` marker highlighting any case where the caller needs to consider lifetime, or where the callee might modify the value. That would indeed be a huge advantage if it worked! Unfortunately, my experience is that in practice, most non-trivial values end up being pointers, which means you _don't_ have to add a `&` when passing it around. So the advantage doesn't really play out in practice.

This is all a matter of opinion, of course.

As for lifetime rules, I try to document them in comments when they diverge from some common-sense baseline rules. The baseline rules are something like:
- If a constructor takes a reference, then the reference must remain valid until the destructor. (This case applies here.)
- If a function or method that returns a promise takes a reference, the reference must remain valid until the promise resolves.
- Otherwise, reference parameters need only remain valid until the function returns, unless otherwise documented.

kentonv: Much to my own surprise, it turns out that these rules are documented:

https://github.com/capnproto/capnproto/blob/master/style-guide.md#pointers-references

https://github.com/capnproto/capnproto/blob/master/style-guide.md#ownership

c++/src/kj/async-io.c++
kentonv: I'm tempted to redo the async I/O interfaces to favor `ArrayPtr<byte>` over `void*, size_t`, which would improve things like this. Main thing holding me back is the amount of code that would need to be updated...

kentonv:
@harrishancock Was wondering why you hadn't commented, then realized I never added you as a reviewer, duh...

harrishancock:
To be fair, it's also because I've been procrastinating! Sorry, I'll review. :)

PR #825, opened 2019-05-24
Add per-object streaming flow control

kentonv:
This can be used on a method to indicate that it is used for "streaming", like:

    write @0 (bytes :Data) -> stream;

A "streaming" method is one which is expected to be called many times to transmit an ordered stream of items. For best throughput, it is often necessary to make multiple overlapping calls, so as not to wait for a round trip for every item. However, to avoid excess buffering, it may be necessary to apply backpressure by having the client limit the total number of overlapping calls. This logic is difficult to get right at the application level, so making it a language feature gives us the opportunity to implement it in the RPC layer.

The approach in this PR is backwards-compatible. In fact, no new features are added to the RPC protocol itself. Streaming simply alters the behavior of the client and server. A streaming call is binary-compatible with a call that returns empty results. Hence, [Sandstorm's ByteStream type](https://github.com/sandstorm-io/sandstorm/blob/92eae9f9673ad71226a5a045287e27662e6c9abb/src/sandstorm/util.capnp#L78-L126) could be updated to utilize streaming while remaining backwards-compatible. Notice how the doc comments of that class describe a method of ad hoc flow control. This PR essentially integrates such flow control into the RPC implementation!

Currently, the method of flow control is to calculate an appropriate "window size" for each capability, analogous to a TCP window, and then apply backpressure if more than that many bytes are "in-flight" (part of a Call that has not received a corresponding Return) at  once time. The window size is, at present, chosen to match the window size of the underlying socket. This is a compromise that works well in a simple two-party scenario but is not ideal in the presence of proxying; see comments in `rpc-twoparty.c++` for discussion and future possibilities.

In fact, this could even have been done entirely on the client side, with servers having no special treatment of streaming methods. However, as a convenience, I decided also to change server-side logic such that streaming calls are automatically queued and delivered one-at-a-time, regardless of what the client does. I suspect most use cases would end up having to do this anyway. For example, an RPC stream wrapping a KJ AsyncOutputStream would have to deal with KJ's requirement not to overlap write() calls; the Cap'n Proto implementation now takes care of that. If it turns out some use cases don't want this, we can add something like CallContext::unblockStream() to let them release the next call.


kentonv:
Duh, forgot to update the pedestrian build systems.

c++/src/capnp/rpc.c++
```diff
@@ -600,6 +600,25 @@ private:
...
+    virtual void adoptFlowController(kj::Own<RpcFlowController> flowController) {
+      // Called when a PromiseClient resolves to another RpcClient. If streaming calls were
+      // outstanding on the old client, we'd like to keep using the same FlowController on the new
+      // client, so as to keep the flow steady.
+
+      if (this->flowController == nullptr) {
+        // We don't have any existing flowController so we can adopt this one, yay!
+        this->flowController = kj::mv(flowController);
+      } else {
+        // Apparently, there is an existing flowController. This is an unusual scenario: Apparenly
```
jclee: "Apparenly"

c++/src/capnp/generated-header-support.h
```diff
@@ -403,3 +403,11 @@ inline constexpr uint sizeInWords() {
       static constexpr ::capnp::_::RawSchema const* schema = &::capnp::schemas::s_##id;
 
 #endif  // CAPNP_LITE, else
+
+namespace capnp {
+namespace schemas {
+CAPNP_DECLARE_SCHEMA(995f9a3377c0b16e);
+// HACK: Forward-declare the RawSchema for StreamResult, from stream.capnp. This allows capnp
+//   files which declare steraming methods to avoid including stream.capnp.h.
```
jclee: "steraming"

kentonv:
@harrishancock Despite being on leave I would like to get this and #829 reviewed rather than let them sit long-term.

harrishancock:
I haven't forgotten about this. I've gotten a start on a review and will finish it up once some other work is out the door.

c++/src/capnp/rpc-twoparty-test.c++
```diff
@@ -639,6 +639,99 @@ KJ_TEST("Streaming over RPC") {
...
+  // Now we can unwrap!
+  KJ_EXPECT(&KJ_ASSERT_NONNULL(unwrapPromise.wait(waitScope)) == &server);
+
+  // But our last stream call still isn't done.
+  KJ_EXPECT(!promise.poll(waitScope));
+
+  // Finish it.
+  KJ_ASSERT_NONNULL(server.fulfiller)->fulfill();
+  promise.wait(waitScope);
+}
```
harrishancock: Nice test.

c++/src/kj/async-io-test.c++
```diff
@@ -1634,6 +1634,27 @@ KJ_TEST("Userland pipe tryPumpFrom to pumpTo for same amount fulfills simultaneo
   KJ_EXPECT(pumpPromise2.wait(ws) == 6);
 }
 
+KJ_TEST("Userland pipe multi-part write doesn't quit early") {
+  kj::EventLoop loop;
+  WaitScope ws(loop);
+
+  auto pipe = newOneWayPipe();
+  auto pipe2 = newOneWayPipe();
+  auto pumpPromise = KJ_ASSERT_NONNULL(pipe2.out->tryPumpFrom(*pipe.in, 6));
```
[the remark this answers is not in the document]
kentonv: Heh, indeed, I think it's actually wrong... `pipe.in` ends up with multiple logical readers, but it doesn't end up causing a problem because the `tryPumpFrom()` stays blocked waiting for a reader on `pipe2.in`, which never comes. Whoops.

c++/src/capnp/capability.c++
```diff
@@ -553,6 +563,106 @@ private:
...
+      // Previous streaming call threw, so everything fails from now on.
+      return kj::cp(*e);
+    }
+
+    auto result = server->dispatchCall(interfaceId, methodId,
+                                       CallContext<AnyPointer, AnyPointer>(context));
+    if (result.isStreaming) {
+      return result.promise
+          .catch_([this](kj::Exception&& e) {
+        brokenException = kj::heap(e);
```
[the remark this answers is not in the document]
kentonv: Yikes, apparently you can set a `Maybe<T>` from a `T*`, and it checks for nullness of the pointer. What was I thinking in 2013? Lemme remove that overload...

c++/src/capnp/rpc.c++
```diff
@@ -2996,4 +3106,146 @@ void RpcSystemBase::baseSetFlowLimit(size_t words) {
...
+            }
+          }
+        }));
+        front.sentFulfiller->fulfill();
+      });
+    }
+
+    KJ_IF_MAYBE(f, emptyFulfiller) {
+      if (queue.empty()) {
+        f->get()->fulfill();
```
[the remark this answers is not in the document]
kentonv: Wow nice catch.

kentonv:
Fixed all issues, PTAL.

c++/src/capnp/rpc.h
```diff
@@ -359,13 +359,13 @@ class RpcFlowController {
   // sends will propagate the exception.
   //
   // Note that messages sent with this method must still be delivered in the same order as if they
-  // had been sent with `message->send()`; they cannot be deloyed until later. This is important
+  // had been sent with `message->send()`; they cannot be deployed until later. This is important
```
harrishancock: This really isn't supposed to be "delayed"?

kentonv: Oh geez, my internal spell correct is no better than your average computer apparently.

c++/src/capnp/stream.capnp
```diff
@@ -0,0 +1,50 @@
...
+  #
+  #     write @0 (bytes :Data) -> import "/capnp/stream.capnp".StreamResult;
+  #
+  # However, implementations that recognize streaming will elide the reference to StreamResult
+  # and instead give write() a different signature appropriate for streaming.
+  #
+  # Streaming methods do not return a result -- that is, they return Promise<void>. This promise
+  # resolves not to indicate that the call was actually delivered, but instead to provide
+  # backpressure. When the previous call's promise resolves, it is time to make another call. On
+  # the client side, the RPC system will resolve promises immediately until an apporpriate number
```
jclee: "apporpriate"

PR #827, opened 2019-05-24
Add support for url-safe base64 encoding

jclee:


kentonv:
We'll probably want to improve the code so that it doesn't have to do a modification pass and a copy/allocation to get rid of the padding... but that can come later, at least this will help clean up all the duplication.

PR #828, opened 2019-05-27
Fix uninitialized byte arrays in encoding-test.

kentonv:
Initializer lists are temporaries. The code apparently works in debug mode but fails when optimized.


kentonv:
(Merging without review since it fixes the build.)

PR #829, opened 2019-05-30
 Define and implement HTTP-over-Cap'n-Proto

kentonv:
~*[NOTE: This PR is targeting the `streaming` branch because it builds on top of that, but should be redirected to `master` once `streaming` has been merged.]*~

This allows an HTTP request/response to be forwarded over Cap'n Proto RPC, multiplexed with arbitrary other RPC transactions.

This could be compared with HTTP/2, which is a binary protocol representation of HTTP that allows multiplexing. HTTP-over-Cap'n-Proto provides the same, but with some advantages inherent in leveraging Cap'n Proto:
- HTTP transactions can be multiplexed with regular Cap'n Proto RPC transactions. (While in theory you could also layer RPC on top of HTTP, as gRPC does, HTTP transactions are much heavier than basic RPC. In my opinion, layering HTTP over RPC makes much more sense because of this.)
- HTTP endpoints are object capabilities. Multiple private endpoints can be multiplexed over the same connection.
- Either end of the connection can act as the client or server, exposing endpoints to each other.
- Cap'n Proto path shortening can kick in. For instance, imagine a request that passes through several proxies, then eventually returns a large streaming response. If the proxies is each a Cap'n Proto server with a level 3 RPC implementation, and the response is simply passed through verbatim, the response stream will automatically be shortened to skip over the middleman servers. At present no Cap'n Proto implementation supports level 3, but path shortening can also apply with only level 1 RPC, if all calls proxy through a central hub process, as is often the case in multi-tenant sandboxing scenarios.

There are also disadvantages vs. HTTP/2:
- HTTP/2 is a standard. This is not.
- This protocol is not as finely optimized for the HTTP use case. It will take somewhat more bandwidth on the wire.
- No mechanism for server push has been defined, although this could be a relatively simple addition to the http-over-capnp interface definitions.
- No mechanism for stream prioritization is defined -- this would likely require new features in the Cap'n Proto RPC implementation itself.
- At present, the backpressure mechanism is naive and its performance will suffer as the network distance increases. I intend to solve this by adding better backpressure mechanisms into Cap'n Proto itself.

Shims are provided for compatibility with the KJ HTTP interfaces.

Note that Sandstorm has its own http-over-capnp protocol: https://github.com/sandstorm-io/sandstorm/blob/master/src/sandstorm/web-session.capnp

Sandstorm's protocol and this new one are intended for very different use cases. Sandstorm implements sandboxing of web applications on both the client and server sides. As a result, it cares deeply about the semantics of HTTP headers and how they affect the browser. This new http-over-capnp protocol is meant to be a dumb bridge that simply passes through all headers verbatim.

### DIFF
--- a/c++/src/capnp/compat/byte-stream.c++
+++ b/c++/src/capnp/compat/byte-stream.c++
@@ -1,0 +1,997 @@
+// Copyright (c) 2019 Cloudflare, Inc. and contributors
+// Licensed under the MIT License:
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+#include "byte-stream.h"
+#include <kj/one-of.h>
+#include <kj/debug.h>
+
+namespace capnp {
+
+class ByteStreamFactory::StreamServerBase: public capnp::ByteStream::Server {
+public:
+  virtual void returnStream(uint64_t written) = 0;
+  // Called after the StreamServerBase's internal kj::AsyncOutputStream has been borrowed, to
+  // indicate that the borrower is done.
+  //
+  // A stream becomes borrowed either when getShortestPath() returns a BorrowedStream, or when
+  // a SubstreamImpl is constructed wrapping an existing stream.
+
+  struct BorrowedStream {
+    // Represents permission to use the StreamServerBase's inner AsyncOutputStream directly, up
+    // to some limit of bytes written.
+
+    StreamServerBase& lender;
+    kj::AsyncOutputStream& stream;
+    uint64_t limit;
+  };
+
+  typedef kj::OneOf<kj::Promise<void>, capnp::ByteStream::Client*, BorrowedStream> ShortestPath;
+
+  virtual ShortestPath getShortestPath() = 0;
+  // Called by KjToCapnpStreamAdapter when it has determined that its inner ByteStream::Client
+  // actually points back to a StreamServerBase in the same process created by the same
+  // ByteStreamFactory. Returns the best shortened path to use, or a promise that resolves when the
+  // shortest path is known.
+
+  virtual void directEnd() = 0;
+  // Called by KjToCapnpStreamAdapter's destructor when it has determined that its inner
+  // ByteStream::Client actually points back to a StreamServerBase in the same process created by
+  // the same ByteStreamFactory. Since destruction of a KJ stream signals EOF, we need to propagate
+  // that by destroying our underlying stream.
+  // TODO(cleanup): When KJ streams evolve an end() method, this can go away.
+};
+
+class ByteStreamFactory::SubstreamImpl final: public StreamServerBase {
+public:
+  SubstreamImpl(ByteStreamFactory& factory,
+                StreamServerBase& parent,
+                capnp::ByteStream::Client ownParent,
+                kj::AsyncOutputStream& stream,
+                capnp::ByteStream::SubstreamCallback::Client callback,
+                uint64_t limit,
+                kj::PromiseFulfillerPair<void> paf = kj::newPromiseAndFulfiller<void>())
+      : factory(factory),
+        state(Streaming {parent, kj::mv(ownParent), stream, kj::mv(callback)}),
+        limit(limit),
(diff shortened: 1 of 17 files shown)


c++/src/capnp/compat/byte-stream.c++
```diff
@@ -0,0 +1,997 @@
...
+          KJ_IF_MAYBE(prober, currentParent->state.tryGet<kj::Own<PathProber>>()) {
+            // Either we didn't find any shorter path at all during probing and faked an EOF
+            // to get out of the probe (see comments in tryRead(), or we DID find a shorter path,
+            // completed a pumpTo() using a substream, and that substream redirected back to us,
+            // and THEN we couldn't find any further shorter paths for subsequent pumps.
+
+            // HACK: If we overwrite the Probing state now, we'll delete ourselves and delete
+            //   this task promise, which is an error... let the event loop do it later by
+            //   detaching.
+            task.attach(kj::mv(*prober)).detach([](kj::Exception&&){});
```
KubaO: Isn't this a pattern typical of GUI frameworks, where an object's deletion is deferred to the event loop? In Qt, it's `QObject::deleteLater()`, in macOS Cocoa, it's the `NSObject -autorelease` instance method. Shouldn't the `kj` framework offer something like this?

kentonv: KJ has such a facility -- the `detach()` mechanism that is being used here.

The problem with `detach()` is that it is prone to creating use-after-free scenarios. If the destructor of an object accesses some other object, then that other object has to still exist. KJ has strong support for cancelling promises by destroying them early. If you start an asynchronous task and hand it a reference to some object Foo, it's generally expected that you can cancel the task at any point by destroying the promise, and once you've done that, then it is safe to destroy Foo. To make this work correctly, it's important to carefully track the ownership hierarchy between promises, so that destroying one promise transitively destroys all the inner promises it depends on.

But when you detach a promise, then it is no longer owned by anyone. It will always run to completion and will hold on to all its references until then. So if it happens to hold any references that it doesn't own, then you have a dangling pointer problem that is hard to solve.

In practice, I've found that almost every place I'm tempted to use `detach()` turns out to be a mistake, introducing use-after-free bugs.

In this specific instance in `byte-stream.c++`, it works out OK because Cap'n Proto capabilities are reference-counted. But I still feel it's bad form to rely on `detach()`. The way to fix this instance is to add a `Promise<void> end()` method to `kj::AsyncOutputStream`, so that terminating a stream is an explicit asynchronous operation rather than something that happens implicitly in the destructor. That, however, requires a major refactoring of all users of `AsyncInputStream`.

KubaO: Is any refactoring like that limited to the scope of capnproto code to be acceptable, or would it require submitting pulls to other projects you might care about (I have no idea, thus I ask).

kentonv: My plan is to make this refactoring part of the 0.8 release. It will be a breaking change for anyone who directly calls or implements `AsyncOutputStream`. We'll have to tell people that they need to update their code when they move from 0.7 to 0.8. It's painful but there's not much other way around it -- this change really has to be made one way or another. I don't think it's feasible for us to go and update dependents' code -- even in the few cases where they are open source.

Once we declare a 1.0 release, I hope that we don't have any more such major backwards-incompatible refactorings. This refactoring is actually the biggest blocker to a 1.0 release currently.

I am actively working on this change BTW.

c++/src/kj/compat/http.c++
```diff
@@ -679,9 +679,16 @@ void HttpHeaders::addNoCheck(kj::StringPtr name, kj::StringPtr value) {
       indexedHeaders[id->id] = value;
     } else {
       // Duplicate HTTP headers are equivalent to the values being separated by a comma.
-      auto concat = kj::str(indexedHeaders[id->id], ", ", value);
-      indexedHeaders[id->id] = concat;
-      ownedStrings.add(concat.releaseArray());
+      if (strcasecmp(name.cStr(), "set-cookie") == 0) {
```
harrishancock: Would a unit test for this branch make sense?

c++/src/capnp/compat/http-over-capnp.c++
```diff
@@ -0,0 +1,713 @@
...
+      results.setBody(factory.streamFactory.kjToCapnp(kj::mv(pipe.out)));
+      state->addTask(pipe.in->pumpTo(*bodyStream)
+          .ignoreResult()
+          .attach(kj::mv(bodyStream), kj::mv(pipe.in)));
+    }
+    return kj::READY_NOW;
+  }
+
+  kj::Promise<void> startWebSocket(StartWebSocketContext context) override {
+    KJ_REQUIRE(!sent, "already called startRespnose() or startWebSocket()");
```
harrishancock: /startRespnose()/startResponse()/

c++/src/capnp/compat/byte-stream.c++
```diff
@@ -0,0 +1,997 @@
...
+    StreamServerBase& lender;
+    kj::AsyncOutputStream& stream;
+    uint64_t limit;
+  };
+
+  typedef kj::OneOf<kj::Promise<void>, capnp::ByteStream::Client*, BorrowedStream> ShortestPath;
+
+  virtual ShortestPath getShortestPath() = 0;
+  // Called by KjToCapnpStreamAdapter when it has determined that its inner ByteStream::Client
+  // actually points back to a StreamServerBase in the same process. Returns the best
```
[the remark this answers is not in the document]
kentonv: Clarified it has to belong to the same ByteStreamFactory.

c++/src/capnp/compat/byte-stream.c++
```diff
@@ -0,0 +1,998 @@
...
+              kjStream.lender.returnStream(kjStream.limit);
+              return write(rest);
+            });
+          } else {
+            // FUUUUUUUU---- we need to split one of the pieces in two.
+            auto left = kj::heapArrayBuilder<kj::ArrayPtr<const byte>>(splitPiece + 1);
+            auto right = kj::heapArrayBuilder<kj::ArrayPtr<const byte>>(pieces.size() - splitPiece);
+
+            for (auto i: kj::zeroTo(splitPiece)) {
+              left[i] = pieces[splitPiece];
```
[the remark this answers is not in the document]
kentonv: nice catch

c++/src/kj/compat/http.c++
kentonv: ohhhh kay.

kentonv:
Looks like 871947c is a winner. @harrishancock is on vacation but told me before leaving that he approves once his comments are addressed, so I'm going to squash fixups and merge now.

PR #831, opened 2019-06-09
Skip test that fails under qemu-user, probably due to a qemu bug.

kentonv:
(This should solve the problem with the Workers CI failure.)


kentonv:
@harrishancock or @jclee feel free to click merge on this yourselves after review and before the next deps roll. I'll most likely be occupied elsewhere... :)

PR #832, opened 2019-06-10
Align HTTP entity-body delimiting rules with RFC 7230.

kentonv:
The main case where the code was wrong is when neither Content-Length nor Transfer-Encoding was provided on a response. In this case the response is delimited by closing the connection, but KJ previously rejected it outright. AFAICT almost no one on the whole internet relies on this anymore... almost.

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -1598,6 +1598,10 @@ static_assert(!fastCaseCmp<'f','O','o','B','1','a'>("FooB1"), "");
 kj::Own<kj::AsyncInputStream> HttpInputStreamImpl::getEntityBody(
     RequestOrResponse type, HttpMethod method, uint statusCode,
     const kj::HttpHeaders& headers) {
+  // Rules to determine how HTTP entity-body is delimited:
+  //   https://tools.ietf.org/html/rfc7230#section-3.3.3
+
+  // #1
   if (type == RESPONSE) {
     if (method == HttpMethod::HEAD) {
       // Body elided.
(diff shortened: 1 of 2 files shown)


kentonv:
As with #831, please feel free to click merge if you like it and want to get it into the deps roll.

c++/src/kj/compat/http.c++
```diff
@@ -1616,35 +1620,68 @@ kj::Own<kj::AsyncInputStream> HttpInputStreamImpl::getEntityBody(
...
+    // NOTE: #3¶3 is ambiguous about what should happen if Transfer-Encoding and Content-Length are
+    //   both present. It says that Transfer-Encoding takes precedence, but also that the request
+    //   "ought to be handled as an error", and that proxies "MUST" drop the Content-Length before
+    //   forwarding. We ignore the vague "ought to" part and implement the other two. (The
+    //   dropping of Content-Length will happen naturally if/when the message is sent back out to
+    //   the network.)
     if (fastCaseCmp<'c','h','u','n','k','e','d'>(te->cStr())) {
+      // #3¶1
       return kj::heap<HttpChunkedEntityReader>(*this);
+    } else if (fastCaseCmp<'i','d','e','n','t','i','t','y'>(te->cStr())) {
```
harrishancock: This is an interesting case. RFC 2616 referred to an "identity" transfer encoding, but then forgot to define it (it referred to a nonexistent "section 3.6.2"). The body length calculations use it, though, so we can determine the length of the body by the connection being closed -- we just don't technically know how to decode the body!

This is functionally equivalent to RFC 7230, which removed the "identity" transfer encoding. So per that RFC we ought to treat identity transfer encoding as any other non-chunked encoding that we don't support: i.e., it ends at connection close, and we don't know how to decode it.

(I'm happy interpreting it with the obvious meaning as you do here; researching this just gave me some chuckles.)

c++/src/kj/compat/http.c++
```diff
@@ -1616,35 +1620,68 @@ kj::Own<kj::AsyncInputStream> HttpInputStreamImpl::getEntityBody(
...
     }
   }
 
+  // #4 and #5
   KJ_IF_MAYBE(cl, headers.get(HttpHeaderId::CONTENT_LENGTH)) {
-    return kj::heap<HttpFixedLengthEntityReader>(*this, strtoull(cl->cStr(), nullptr, 10));
+    // NOTE: By spec, multiple Content-Length values are allowed as long as they are the same, e.g.
+    //   "Content-Length: 5, 5, 5". Hopefully no one actually does that...
+    char* end;
+    uint64_t length = strtoull(cl->cStr(), &end, 10);
```
harrishancock: `strtoull()`'s man page says it is locale-sensitive, i.e., it may do strange things with thousands-separators in certain locales. Is this safe in practice?

kentonv: Well, the code used strtoull() before this change, too, so it's not getting worse.

Our servers all presumably run with the "C" locale which generally behaves unsurprisingly. I suppose there could be a problem when this code runs on desktop machines with a specific locale. Note we use the strtol family of functions all over the place, so if it is a problem, we need to do a lot of fixing.

c++/src/kj/compat/http.c++
kentonv: Doh, didn't notice that "identity" was actually removed in 7230. Oh well, doesn't hurt to recognize it, though I suppose no one uses it since it's... not useful.

c++/src/kj/compat/http.c++
harrishancock: All good points.

PR #834, opened 2019-06-15
Introduce kj::attachVal(), kj::attachRef(), and capnp::clone() utility functions

kentonv:
These have repeatedly come up as something that would be useful.

See individual commits for details.


kentonv:
TFW your build fails on every single CI target _except_ MSVC.

PR #837, opened 2019-06-16
DO NOT MERGE: Use BSD-make-compatible shell expansion syntax.

kentonv:
GNU Make supports this as of 4.0, released in 2013.

Unfortunately, it appears my recently-updated MacOS/XCode ships with GNU Make 3.81, from 2006. WTF Apple.

We may need to either:
1. Extend configure.ac to detect gmake vs. BSD make and adjust the Makefile accordingly. (I think it's able to do this such that the branch it doesn't use is actually commended out entirely.)
2. Detect when people use BSD make and error out early, asking them to please use gmake instead, which I think most BSD users have installed (but they might be offended).

### DIFF
--- a/c++/Makefile.am
+++ b/c++/Makefile.am
@@ -237,7 +237,10 @@ endif
 
 # Don't include security release in soname -- we want to replace old binaries
 # in this case.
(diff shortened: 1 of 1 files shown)


kentonv:
Or maybe I should just calculate SO_VERSION in configure, duh...

kentonv:
Apparently Travis CI's MacOS has a newer make installed. Maybe something's wrong with my machine. But moving the evaluation to `configure` is easy anyway; done in #838. Closing this PR in favor of that.

PR #839, opened 2019-06-17
Implement MutexGuarded::when() (i.e. condvars) on all platforms.

kentonv:
Also, add support for passing a timeout to when().

### DIFF
--- a/c++/src/kj/mutex.c++
+++ b/c++/src/kj/mutex.c++
@@ -511,6 +511,11 @@ void Mutex::lockWhen(Predicate& predicate, Maybe<Duration> timeout) {
   LARGE_INTEGER endTime;
 
   KJ_IF_MAYBE(t, timeout) {
+    // Windows sleeps are inaccurate -- they can be longer *or shorter* than the requested amount.
+    // For many use cases of our API, a too-short sleep would be unacceptable. Experimentally, it
+    // seems like sleeps can be up to half a millisecond short, so we'll add a millisecond.
+    *t += 500 * kj::MICROSECONDS;
+
     // Compute initial sleep time.
     sleepMs = *t / kj::MILLISECONDS;
     if (*t % kj::MILLISECONDS > 0 * kj::SECONDS) {
--- a/c++/src/kj/mutex.c++
+++ b/c++/src/kj/mutex.c++
@@ -280,6 +280,7 @@ void Once::reset() {
 
 #define coercedSrwLock (*reinterpret_cast<SRWLOCK*>(&srwLock))
 #define coercedInitOnce (*reinterpret_cast<INIT_ONCE*>(&initOnce))
+#define coercedCondvar(var) (*reinterpret_cast<CONDITION_VARIABLE*>(&var))
 
 Mutex::Mutex() {
   static_assert(sizeof(SRWLOCK) == sizeof(srwLock), "SRWLOCK is not a pointer?");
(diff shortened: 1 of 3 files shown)


kentonv:
OK, this should be ready now.

It appears that sleeps on Windows (including sleeps waiting for a condvar) are highly inaccurate. You give it a number of milliseconds as the timeout, but it often wakes up up to half a millisecond early, which violates the usual rule that timeouts can wake up late but never early. Amusingly, this issue only shows up when running on actual Windows -- on WINE (which I use for testing), sleeps are accurate, probably because the underlying Linux sleeps are accurate. Hence the extra churn after I originally uploaded this -- the inaccuracy issue only showed up in CI, which runs real Windows.

jclee:
Surprising!  From the API docs, looks like 1ms is the best system clock resolution you're allowed to ask for on Windows.  And if I'm reading the docs correctly, there's no guarantee of even that resolution, unless your program has specifically made API calls asking for it -- you may be just getting the finest clock resolution that any process on the system has requested!

https://docs.microsoft.com/en-us/windows/desktop/api/synchapi/nf-synchapi-sleep

c++/src/kj/mutex.c++
```diff
@@ -688,6 +738,16 @@ void Mutex::lockWhen(Predicate& predicate, Maybe<Duration> timeout) {
     KJ_PTHREAD_CALL(pthread_mutex_unlock(&waiter.stupidMutex));
 
     lock(EXCLUSIVE);
+    currentlyLocked = true;
+
+    KJ_IF_MAYBE(exception, waiter.exception) {
```
harrishancock: Should this go above the `lock()`, since the exception will cause the mutex to be unlocked anyway?

c++/src/kj/mutex.c++
```diff
@@ -511,6 +511,11 @@ void Mutex::lockWhen(Predicate& predicate, Maybe<Duration> timeout) {
   LARGE_INTEGER endTime;
 
   KJ_IF_MAYBE(t, timeout) {
+    // Windows sleeps are inaccurate -- they can be longer *or shorter* than the requested amount.
+    // For many use cases of our API, a too-short sleep would be unacceptable. Experimentally, it
+    // seems like sleeps can be up to half a millisecond short, so we'll add a millisecond.
```
harrishancock: You mean "add half a millisecond"?

c++/src/kj/mutex.c++
kentonv: It can't because the waiter is still in the linked list where it is protected by the lock. In fact it occurs to me that I need to be a little more careful that my calls to `removeWaiter()` happen strictly under lock -- added a commit to ensure that.

c++/src/kj/mutex.c++
```diff
@@ -191,27 +238,80 @@ void Mutex::assertLockedByCaller(Exclusivity exclusivity) {
...
 
-    // Ownership of an exclusive lock was transferred to us. We can continue.
+    // Wait for someone to set our futex to 1.
+    for (;;) {
+      // Note we use FUTEX_WAIT_BITSET_PRIVATE + FUTEX_BITSET_MATCH_ANY to get the same effect as
+      // FUTEX_WAIT_PRIVATE except that the timeout is specified as an absolute time based on
+      // CLOCK_MONOTONIC. Otherwise, FUTEX_WAIT_PRIVATE interprets it as a relative time, forcing
+      // us to recompute the time after every iteration.
+      KJ_SYSCALL_HANDLE_ERRORS(syscall(SYS_futex,
+          &waiter.futex, FUTEX_WAIT_BITSET_PRIVATE, 0, tsp, NULL, FUTEX_BITSET_MATCH_ANY)) {
```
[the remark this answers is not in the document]
kentonv: I wasn't sure how `nullptr` would interact with C-style variadic functions. Did a test and it looks like it works as one would hope, so I changed all the NULLs to nullptr.

c++/src/kj/mutex.c++
harrishancock: Oof, subtle!

harrishancock:
New commits look good.

PR #841, opened 2019-06-20
Fix sendStream() failing if it can't complete immediately.

kentonv:


c++/src/kj/async-io-test.c++
```diff
@@ -238,6 +238,49 @@ TEST(AsyncIo, CapabilityPipe) {
...
+      ++nonBlockedCount;
+      promise.wait(io.waitScope);
+    } else {
+      // Send blocked! Let's continue with this promise then!
+      endpoint1 = kj::mv(pipe2.ends[1]);
+      break;
+    }
+  }
+
+  KJ_DBG(nonBlockedCount);
```
harrishancock: Vestigial KJ_DBG

kentonv: oops, fixed. (It printed ~300 FWIW.)

kentonv: Or 278, to be precise. A weirdly non-round number, and smaller than I would have expected.

harrishancock: Nice, I was curious. I was expecting it to be in the thousands.

kentonv:
@harrishancock Yeah I think that test also failed before I updated the PR, and I re-ran it, and then it passed. Seems it's a bit flaky. :/

c++/src/kj/async-io-unix.c++
```diff
@@ -216,7 +216,7 @@ public:
       return downcast<AsyncStreamFd>(*stream).fd;
     };
     auto promise = writeInternal(data, moreData, fds);
-    return promise.attach(kj::mv(fds));
+    return promise.attach(kj::mv(fds), kj::mv(streams));
```
harrishancock: BTW, I must have read this line of code 5 times in the last 2 days and completely missed this.

jclee:
> Looks transient, though? FYI @jclee

That's unfortunate...  That test is based on the one above it, which was formerly disabled for both Windows and Mac due to flakiness, until I found that adding an extra poll() seemed to make the test pass reliably.  I was hoping the same would be true for this test, but I guess it isn't.  Since two connections are involved, I guess I could try adding a second poll() to see if it makes the test reliable, under the assumption that N connections opening may require up to N trips around the event loop.  Or we could just disable the test entirely on Windows, until we work out what is actually going wrong?

c++/src/kj/async-io-unix.c++
kentonv: I missed it a few times too. Didn't dawn on me until I read the calling code and noticed that ownership of the streams array was being passed in.

PR #842, opened 2019-06-20
Doc fix: Expand the short description of the list pointer.

KubaO:
It helps to list, in one place, both ways the list size is conveyed.


kentonv:
Seems reasonable.

PR #846, opened 2019-06-22
Add basic system clock APIs.

kentonv:
This eliminates the use of `std::chrono`, helps clean up some code in the condvar implementation, and is something I've wanted on several occasions.

### DIFF
--- a/c++/src/kj/time-test.c++
+++ b/c++/src/kj/time-test.c++
@@ -1,0 +1,135 @@
+// Copyright (c) 2019 Cloudflare, Inc. and contributors
+// Licensed under the MIT License:
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+#if _WIN32
+#define WIN32_LEAN_AND_MEAN 1  // lolz
+#endif
+
+#include "time.h"
+#include "debug.h"
+#include <kj/test.h>
+#include <time.h>
+
+#if _WIN32
+#include <windows.h>
+#include "windows-sanity.h"
+#else
+#include <unistd.h>
+#endif
+
+namespace kj {
+namespace {
+
+#if _WIN32
+void delay(kj::Duration d) {
+  Sleep(d / kj::MILLISECONDS);
+}
+#else
+void delay(kj::Duration d) {
+  usleep(d / kj::MICROSECONDS);
+}
+#endif
+
+KJ_TEST("calendar clocks matches unix time") {
+  // Check that the times returned by the calendar clock are within 1s of what time() returns.
+
+  auto& coarse = systemCoarseCalendarClock();
+  auto& precise = systemPreciseCalendarClock();
+
+  Date p = precise.now();
+  Date c = coarse.now();
+  time_t t = time(nullptr);
+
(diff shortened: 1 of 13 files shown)


kentonv:
Decided to give @harrishancock a break and assign this one to @jclee. :)

c++/src/kj/time-test.c++
```diff
@@ -0,0 +1,135 @@
...
+void delay(kj::Duration d) {
+  Sleep(d / kj::MILLISECONDS);
+}
+#else
+void delay(kj::Duration d) {
+  usleep(d / kj::MICROSECONDS);
+}
+#endif
+
+KJ_TEST("calendar clocks matche unix time") {
```
jclee: "matche"

c++/src/kj/time-test.c++
```diff
@@ -0,0 +1,135 @@
...
+        coarseMonoDiff > delayTime - 20 * kj::MILLISECONDS &&
+        coarseMonoDiff < delayTime + 20 * kj::MILLISECONDS &&
+        preciseMonoDiff > delayTime - 20 * kj::MILLISECONDS &&
+        preciseMonoDiff < delayTime + 20 * kj::MILLISECONDS) {
+      // success
+      return;
+    }
+  }
+
+  KJ_FAIL_EXPECT("clocks seem inaccuratae ever after 20 tries",
```
jclee: "inaccuratae" "ever"

c++/src/kj/time.c++
```diff
@@ -35,4 +47,213 @@ const Clock& nullClock() {
...
+class Win32PreciseMonotonicClock: public MonotonicClock {
+  // Precise clock implemented using QueryPerformanceCounter().
+  //
+  // TODO(someday): Windows 10 has QueryUnbiasedInterruptTime() and
+  //   QueryUnbiasedInterruptTimePrecise(), a new API for monotonic timing that isn't as difficult.
+  //   Is there any benefit to dynamically checking for these and using them if available?
+
+public:
+  TimePoint now() const override {
+    static const QpcProperties props;
```
jclee: I guess, strictly speaking, static variable initialization isn't guaranteed to be threadsafe, but the chances of it being a problem seem vanishingly small.

c++/src/kj/time.c++
```diff
@@ -35,4 +47,213 @@ const Clock& nullClock() {
...
+
+    LARGE_INTEGER now;
+    QueryPerformanceCounter(&now);
```
jclee: I'm not sure how broadly supported the Windows code is expected to be, but I vaguely recalled a talk (maybe GDC or cppcon?) about someone having problems profiling with performance counters and multithreading/hyperthreading.  From searching around online, this was probably due to [old, buggy versions of Windows](http://archive.gamedev.net/archive/reference/programming/features/timing/index.html), but there is also more recent [documentation](https://docs.microsoft.com/en-us/windows/desktop/dxtecharts/game-timing-and-multicore-processors) that still recommends only calling QueryPerformanceCounter() from a single thread that's been pinned to a processor:

> While QueryPerformanceCounter and QueryPerformanceFrequency typically adjust for multiple processors, bugs in the BIOS or drivers may result in these routines returning different values as the thread moves from one processor to another.

Windows user beware, I guess.

c++/src/kj/time.c++
```diff
@@ -35,4 +47,213 @@ const Clock& nullClock() {
...
+  }
+
+private:
+  clockid_t clockId;
+};
+
+}  // namespace
+
+// FreeBSD has "_PRECISE", but Linux just defaults to precise.
+#ifndef CLOCK_REALTIME_PRECISE
```
jclee: I don't know enough about the various platforms to know if these constants are guaranteed to always be "#defines" (vs., say, enums), but I assume what we have here is OK if it's currently passing CI on the platforms we care about.

jclee:
LGTM, aside from the typos.

c++/src/kj/time.c++
kentonv: As of C++11, initialization of static locals is in fact thread-safe. See:

https://en.cppreference.com/w/cpp/language/storage_duration#Static_local_variables

c++/src/kj/time.c++
kentonv: This doc seems to have the best modern information about this:

https://docs.microsoft.com/en-us/windows/desktop/SysInfo/acquiring-high-resolution-time-stamps#qpc-support-in-windows-versions

It sounds like the days of QPC being inconsistent ended after XP. Cap'n Proto already requires Vista (for the modern synchronization APIs) so I think we're in good shape.

Windows 8 and 10 apparently finally introduced better APIs for high-precision monotonic clocks (QueryInterruptTimePrecise()), but I'm not comfortable requiring those yet.

c++/src/kj/time.c++
kentonv: I'm not sure exactly what the standards require, but I've noticed glibc and Linux headers are in the habit of doing things like this:

```
enum EPOLL_EVENTS
  {
    EPOLLIN = 0x001,
#define EPOLLIN EPOLLIN
    EPOLLPRI = 0x002,
#define EPOLLPRI EPOLLPRI
    EPOLLOUT = 0x004,
#define EPOLLOUT EPOLLOUT
    EPOLLRDNORM = 0x040,
#define EPOLLRDNORM EPOLLRDNORM
    EPOLLRDBAND = 0x080,
#define EPOLLRDBAND EPOLLRDBAND
```

I guess this strongly implies that they expect people will use `#ifdef` to check whether constants are available.

c++/src/kj/time.c++
jclee: Cool.  Yeah, sounds like the "game timing" documentation might have had an unstated assumption of pre-Vista compatibility.

PR #854, opened 2019-06-27
Do not crash when the source file is empty.

KubaO:
Of course source files shouldn't be empty, but when one is for some reason, then the generic "mmap failed" exception and the ensuing goose chase may waste some time...

Is there some other pattern that would work better when handling errors like these?


kentonv:
This is a bug in the filesystem API layer, not in the compiler. Should be fixed by #855.

PR #855, opened 2019-06-28
Fix filesystem API: Empty mmaps should succeed.

kentonv:
I had a snarky comment in the Windows code being annoyed that mapping an empty file was documented to fail... but it turns out zero-length mappings fail on Linux, too. Also my work-around on Windows didn't work (but was never tested).

This PR fixes both.

This was prompted by @KubaO observing that `capnp compile` on an empty source file failed with a cryptic mmap error.

Closes #854, @KubaO's attempt at fixing this, since the correct fix is in the filesystem API, not in the compiler.


KubaO:
Thank you! Perfect.

Issue #552, opened 2017-09-10
API to access annotations


adeschamps:
I may be missing something, but what's the correct way to access annotations? Does the compiler generate a variable/constant with the ID of each annotation? The only example use of annotations that I can find uses what looks like manually entered constants, in [capnpc-c++.c++][define]:

```c++
static constexpr uint64_t NAMESPACE_ANNOTATION_ID = 0xb9c6f99ebf805f2cull;
static constexpr uint64_t NAME_ANNOTATION_ID = 0xf264a779fef191ceull;
```
That code also looks up annotations by doing a linear search through the list of annotations, which suggests that the order in which they are used can impact performance.

Is there anything analogous to protobuf's `descriptor->options().GetExtension(foo)`? Or is the intention that I run `capnpc -ocapnp foo.capnp`, and grab the IDs from that?

[define]: https://github.com/capnproto/capnproto/blob/1dcabbf81d772069119773025d44c29df19ccd30/c%2B%2B/src/capnp/compiler/capnpc-c%2B%2B.c%2B%2B#L56-L57

KubaO:
Are you writing/modifying a codegen? Annotations are meant to be consumed by the codegen. They are passed to the codegen using the schema protocol (`schema.capnp`), and then it's up to it what to do with them. The C++ codegen interprets both the `name` and `namespace` annotation in obvious ways: the `name` overrides the type/member/constant name in the generated code, and the `namespace` selects the namespace the declarations go into. Namespace is usually used at the file scope, but it can be required to any file-scope type or constant.

kentonv:
@adeschamps Annotations ought to generate some sort of code but I never got around to designing or implementing anything. I've always just copied the annotation number into a constant and scanned the raw schema. It would be neat to generate something that could allow reading annotations in a type-safe way.

The linear search is not likely to be a performance issue unless you have hundreds of annotations on a single declaration. That said, for runtime use cases, I do generally recommend "compiling" annotations into optimized lookup tables upfront, like `JsonCodec` does.

@KubaO Annotations are used for much more than just codegen. E.g. `JsonCodec` uses them at runtime to decide how to convert to/from JSON.

KubaO:
Ahh, yes, the `flatten` annotation. D'ohh.

KubaO:
I have a simple proof of concept that produces something like this (for a type `Struct`):

```
struct Struct {
  // ...
  struct _capnpPrivate {
[...]

PR #858, opened 2019-07-02
Access to annotation nodes.

KubaO:
This is a trivial PoC for #552. It's just to spark discussion, nothing more.


kentonv:
It looks like, given a struct type, you're collecting the set of all annotation types used within that struct type.

That's not what anybody wants, though. When I'm interested in reading an annotation, I already know the annotation type I'm interested in, and what I want that's missing is an easy way to look up the _value_ of that annotation as it pertains to the particular annotated declaration.

PR #861, opened 2019-07-15
Add HttpServerErrorHandler interface to provide visibility and customization of protocol errors

harrishancock:
~~This avoids needing to add another option flag to control whether or not to log protocol errors. It's a bit more flexible, too -- we could use these hooks to log 400 responses, but ignore other errors or record them only as metrics.~~

Updated: separated the hook out into an interface and provided a `Response&` for response customization.

### DIFF
--- a/c++/src/kj/compat/http-test.c++
+++ b/c++/src/kj/compat/http-test.c++
@@ -2428,6 +2428,133 @@ KJ_TEST("HttpServer bad request") {
   KJ_EXPECT(expectedResponse == response, expectedResponse, response);
 }
 
+// Ensure that HttpServerSettings can continue to be constexpr.
+KJ_UNUSED static constexpr HttpServerSettings STATIC_CONSTEXPR_SETTINGS {};
+
+class TestErrorHandler: public HttpServerErrorHandler {
+public:
+  kj::Promise<void> handleClientProtocolError(
+      kj::StringPtr message, kj::HttpService::Response& response) override {
+    return sendError(400, "Bad Request", kj::str("Saw protocol error: ", message), response);
+  }
+
+  kj::Promise<void> handleApplicationError(
+      kj::Exception exception, kj::Maybe<kj::HttpService::Response&> response) override {
+    return sendError(500, "Internal Server Error",
+        kj::str("Saw application error: ", exception.getDescription()), response);
+  }
+
+  kj::Promise<void> handleNoResponse(kj::HttpService::Response& response) override {
+    return sendError(500, "Internal Server Error", kj::str("Saw no response."), response);
+  }
+
+  static TestErrorHandler instance;
+
+private:
+  kj::Promise<void> sendError(uint statusCode, kj::StringPtr statusText, String message,
+      Maybe<HttpService::Response&> response) {
+    KJ_IF_MAYBE(r, response) {
+      HttpHeaderTable headerTable;
+      HttpHeaders headers(headerTable);
+      auto body = r->send(statusCode, statusText, headers, message.size());
+      return body->write(message.begin(), message.size()).attach(kj::mv(body), kj::mv(message));
+    } else {
+      KJ_LOG(ERROR, "Saw an error but too late to report to client.");
+      return kj::READY_NOW;
+    }
+  }
+};
+
+TestErrorHandler TestErrorHandler::instance {};
+
+KJ_TEST("HttpServer no response, custom error handler") {
+  auto PIPELINE_TESTS = pipelineTestCases();
+
+  KJ_HTTP_TEST_SETUP_IO;
+  kj::TimerImpl timer(kj::origin<kj::TimePoint>());
+  auto pipe = KJ_HTTP_TEST_CREATE_2PIPE;
+
+  HttpServerSettings settings {};
+  settings.errorHandler = TestErrorHandler::instance;
+
+  HttpHeaderTable table;
+  BrokenHttpService service;
+  HttpServer server(timer, table, service, settings);
+
+  auto listenTask = server.listenHttp(kj::mv(pipe.ends[0]));
+
+  // Do one request.
+  pipe.ends[1]->write(PIPELINE_TESTS[0].request.raw.begin(), PIPELINE_TESTS[0].request.raw.size())
+      .wait(waitScope);
+  auto text = pipe.ends[1]->readAllText().wait(waitScope);
+
+  KJ_EXPECT(text ==
+      "HTTP/1.1 500 Internal Server Error\r\n"
+      "Connection: close\r\n"
+      "Content-Length: 16\r\n"
+      "\r\n"
+      "Saw no response.", text);
+}
+
+KJ_TEST("HttpServer threw exception, custom error handler") {
+  auto PIPELINE_TESTS = pipelineTestCases();
+
+  KJ_HTTP_TEST_SETUP_IO;
+  kj::TimerImpl timer(kj::origin<kj::TimePoint>());
+  auto pipe = KJ_HTTP_TEST_CREATE_2PIPE;
+
+  HttpServerSettings settings {};
+  settings.errorHandler = TestErrorHandler::instance;
+
+  HttpHeaderTable table;
+  BrokenHttpService service(KJ_EXCEPTION(FAILED, "failed"));
+  HttpServer server(timer, table, service, settings);
+
+  auto listenTask = server.listenHttp(kj::mv(pipe.ends[0]));
+
+  // Do one request.
+  pipe.ends[1]->write(PIPELINE_TESTS[0].request.raw.begin(), PIPELINE_TESTS[0].request.raw.size())
+      .wait(waitScope);
+  auto text = pipe.ends[1]->readAllText().wait(waitScope);
+
+  KJ_EXPECT(text ==
+      "HTTP/1.1 500 Internal Server Error\r\n"
+      "Connection: close\r\n"
+      "Content-Length: 29\r\n"
+      "\r\n"
+      "Saw application error: failed", text);
+}
+
+KJ_TEST("HttpServer bad request, custom error handler") {
+  KJ_HTTP_TEST_SETUP_IO;
+  kj::TimerImpl timer(kj::origin<kj::TimePoint>());
+  auto pipe = KJ_HTTP_TEST_CREATE_2PIPE;
+
+  HttpServerSettings settings {};
+  settings.errorHandler = TestErrorHandler::instance;
+
+  HttpHeaderTable table;
+  BrokenHttpService service;
+  HttpServer server(timer, table, service, settings);
+
+  auto listenTask = server.listenHttp(kj::mv(pipe.ends[0]));
+
+  static constexpr auto request = "bad request\r\n\r\n"_kj;
+  auto writePromise = pipe.ends[1]->write(request.begin(), request.size());
+  auto response = pipe.ends[1]->readAllText().wait(waitScope);
+  KJ_EXPECT(writePromise.poll(waitScope));
+  writePromise.wait(waitScope);
+
+  static constexpr auto expectedResponse =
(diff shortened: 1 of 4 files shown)


c++/src/kj/compat/http.h
```diff
@@ -647,6 +647,18 @@ class HttpService {
...
   // UNIMPLEMENTED.
+
+  virtual void onError(uint statusCode, kj::StringPtr statusText, kj::StringPtr errorMessage) {}
+  virtual void onWebSocketError(
+      uint statusCode, kj::StringPtr statusText, kj::StringPtr errorMessage) {}
+  // Hooks called when HttpServer is about to return an error response outside the control of the
+  // HttpService. This can happen when HttpServer encounters a protocol error, or if
+  // `HttpService::request()` throws an exception before writing a response. The parameters contain
+  // the information that HttpServer will send back in the HTTP response, and are provided for
+  // visibility only only, e.g. to allow applications to log or count such occurrences.
```
jclee: "only only"

harrishancock:
Take two.

The `closeAfterSend` boolean is a hack to send a Connection: close header, the absence of which breaks this test: https://github.com/capnproto/capnproto/blob/master/c%2B%2B/src/kj/compat/http-test.c%2B%2B#L3031

The "HttpService threw exception after generating a partial response ..." message changed a bit in a way I'm not super happy about. The logic at the place that generates that message is now inside of the default implementation of `HttpServerErrorHandler::requestError()` (a consequence of passing a `Maybe<Response&>` to indicate partial responses), meaning it no longer has direct access to the original exception. Instead, it wraps the HttpServer's "suggested" error message, which itself wraps the exception's message.

c++/src/kj/compat/http.h
harrishancock: Rewritten in new version.

c++/src/kj/compat/http.c++
```diff
@@ -5063,20 +5056,23 @@ private:
...
                  kj::newPromiseAndFulfiller<void>()) {}
 
 HttpServer::HttpServer(kj::Timer& timer, HttpHeaderTable& requestHeaderTable,
                        kj::OneOf<HttpService*, HttpServiceFactory> service,
-                       Settings settings, kj::PromiseFulfillerPair<void> paf)
+                       Settings settings, kj::Maybe<HttpServerErrorHandler&> errorHandler,
+                       kj::PromiseFulfillerPair<void> paf)
     : timer(timer), requestHeaderTable(requestHeaderTable), service(kj::mv(service)),
-      settings(settings), onDrain(paf.promise.fork()), drainFulfiller(kj::mv(paf.fulfiller)),
+      settings(settings), errorHandler(errorHandler.orDefault(*this)),
```
jclee: Interesting... I hadn't seen this pattern of defaulting to a privately inherited default implementation before.  I guess it's used here because it simplifies the ownership in the default case?

harrishancock: Yes, exactly. I wouldn't espouse this as a pattern, but I'm lazy and it seemed to make sense. I should probably make `HttpServerErrorHandler` pure abstract anyway, though...

c++/src/kj/compat/http.h
```diff
@@ -769,21 +769,47 @@ struct HttpServerSettings {
...
+  // class) sends just such a response.
+  //
+  // If the HttpService did not already send a partial response, the `response` parameter will be
+  // non-null and available for you to customize the response sent to the client. You could use this
+  // to perform error message sanitization, e.g. to censor stack traces.
+  //
+  // Unlike `HttpService::request()`, when calling `response.send()` in this context, a
+  // "Connection: close" header will be added, and the connection will be closed.
+  //
+  // TODO(cleanup): It might make more sense to pass in the original exception object. But this
```
kentonv: I think we should pass in the exception object, so that the handler can inspect it and decide how to log it. E.g. maybe it wants to treat FAILED, OVERLOADED, and DISCONNECTED differently. Maybe the interface should have multiple methods:

    kj::Promise<void> handleClientProtocolError(kj::StringPtr message, ...);
    kj::Promise<void> handleApplicationError(kj::Exception e, ...);

(Are all the client protocol errors 400's, or are various status codes used?)

c++/src/kj/compat/http.h
```diff
@@ -769,21 +769,47 @@ struct HttpServerSettings {
...
+             Settings settings = Settings(),
+             kj::Maybe<HttpServerErrorHandler&> errorHandler = nullptr);
   // Set up an HttpServer that directs incoming connections to the given service. The service
   // may be a host service or a proxy service depending on whether you are intending to implement
   // an HTTP server or an HTTP proxy.
 
   HttpServer(kj::Timer& timer, HttpHeaderTable& requestHeaderTable,
-             HttpServiceFactory serviceFactory, Settings settings = Settings());
+             HttpServiceFactory serviceFactory, Settings settings = Settings(),
+             kj::Maybe<HttpServerErrorHandler&> errorHandler = nullptr);
```
kentonv: Can `errorHandler` be a member of the `settings` struct?

c++/src/kj/compat/http.h
kentonv: Oh yeah, and if people want to censor stack traces then they definitely want to receive the `Exception` pre-stringification...

harrishancock: I think there's just the one protocol error (bad headers), and it's a 400. In that light, the two function approach sounds great.

c++/src/kj/compat/http.h
harrishancock: Yes, but it will be source-incompatible because the settings objects can no longer be literals. I'm fine with that, but didn't play with it enough to know what all needs to be changed. I'll find out.

kentonv: Not sure what you mean. What kind of existing code would break, exactly?

harrishancock: `constexpr HttpServerSettings LOCAL_CLIENT_SETTINGS = { ... };`

Probably not a big deal -- the place we'd need to modify in Workers we'd need to modify to add the new setting anyway, I just didn't spend any time seeing where else that pattern is used. I agree it'd be preferable as a setting field.

harrishancock: (I'm on the bus right now, so not really explaining well.)

harrishancock:
Okay, I think I addressed all feedback in this rewrite.

I ended up adding a third error handler function, `handleNoResponse()`, though I'm not sure it deserves to be separate from `handleApplicationError()`. I wrote it because the error message that gets generated sounds awkward if we re-use `handleApplicationError()` (something like "service threw exception: service didn't respond").

Also, should we keep the current KJ_LOG behavior in the default error handler?

harrishancock:
Cygwin failure looks transient.

c++/src/kj/compat/http.h
kentonv: Is this just because `Maybe<T&>`'s default constructor is not `constexpr`? Could we make it `constexpr`?

c++/src/kj/compat/http.h
```diff
@@ -767,6 +769,45 @@ struct HttpServerSettings {
...
+
+  kj::Maybe<HttpServerErrorHandler&> errorHandler = nullptr;
+  // Customize how client protocol errors and service application exceptions are handled by the
+  // HttpServer. If null, HttpServerErrorHandler's default implementation will be used.
+};
+
+class HttpServerErrorHandler {
+public:
+  virtual kj::Promise<void> handleClientProtocolError(kj::StringPtr message,
+      kj::HttpHeaderTable& headerTable, kj::HttpService::Response& response);
```
kentonv: I think the header table shouldn't be passed as a method parameter. The normal idiom is to pass it into the constructor. E.g. HttpService::request() doesn't pass the header table; the HttpService is expected to have received it at constructor time.

c++/src/kj/compat/http.c++
```diff
@@ -4968,21 +4947,22 @@ private:
     currentMethod = nullptr;
```
kentonv: It appears that every branch below now either re-assigns currentMethod back to its original value, except the one branch that sets it to GET. Maybe remove this line and all the `currentMethod = method` lines below.

c++/src/kj/compat/http.c++
```diff
@@ -5147,4 +5150,76 @@ void HttpServer::taskFailed(kj::Exception&& exception) {
...
+      errorMessage = kj::str(
+          "ERROR: The server threw an exception. Details:\n\n", exception);
+      body = r->send(500, "Internal Server Error", headers, errorMessage.size());
+    }
+
+    return body->write(errorMessage.begin(), errorMessage.size())
+        .attach(kj::mv(errorMessage), kj::mv(body));
+  }
+
+  // TODO(now): Remove this KJ_LOG? If the application wants to log this stuff, it can do it now.
```
kentonv: Nah I think we should leave it. It's the right default behavior.

c++/src/kj/compat/http-test.c++
```diff
@@ -2622,6 +2622,34 @@ KJ_TEST("HttpServer cancels request when client disconnects") {
   cancelPromise.wait(waitScope);
 }
 
+KJ_TEST("HttpServer responds with 400 Bad Request") {
```
kentonv: No tests that actually set a custom error handler?

c++/src/kj/compat/http.c++
harrishancock: Maybe we should `currentMethod = HttpMethod::GET` here, actually? That would allow us to return an error response correctly if someone tried to open a WebSocket connection with a HEAD request (admittedly unlikely).

c++/src/kj/compat/http-test.c++
harrishancock: Will write one.

c++/src/kj/compat/http.h
harrishancock: I agree, but then we'll have to make these all pure virtual, with no default implementation. If we want to allow users to call a default implementation, we could provide free functions which take a header table as a parameter, perhaps?

kentonv: Could the default implementations use a default empty header table? I don't think there's actually any requirement to use the same header table as other responses. In fact, the HttpHeaderTable passed to HttpService is only meant to be used when parsing requests; I think the application can technically use whatever table it wants for responses.

c++/src/kj/compat/http.c++
kentonv: The very first check below this line verifies that the method is GET, so I think that's redundant. (In fact I guess this means that *all* of the later assignments to `currentMethod` can be deleted if we remove this line that sets it null.)

kentonv: (And note that it would be incorrect to respond to a HEAD request with a bodied response even if the HEAD request is an invalid WebSocket handshake. So no, we should not change HEAD to GET.)

c++/src/kj/compat/http.h
harrishancock: Great idea, that solves it.

c++/src/kj/compat/http.c++
harrishancock: Got it!

harrishancock:
Addressed everything but constexpr Maybes and lack of tests. Will pick those up in a bit.

New commits are intended as fixups, though I didn't explicitly label them such.

harrishancock:
Now msvc has a seemingly transient failure. Will poke it later.

harrishancock:
Squashed existing fixups, added tests, added `constexpr Maybe<T&>` constructors (`Maybe<T>` cannot be, due to its non-trivial destructor). @kentonv, PTAL.

kentonv:
LGTM but there are build failures.

harrishancock:
Build succeeded before fixup squash, got go-ahead from @kentonv via chat to merge.

PR #866, opened 2019-07-31
Report raw HTTP content when handling client protocol errors in kj-http

harrishancock:
This information is typically necessary to debug such protocol errors. However, it requires special handling, since it could contain sensitive information.

The code which has the best view on protocol errors is HttpHeaders::tryParseRequest(), which previously returned an empty Maybe<Request> on failure. This commit changes that function to return a OneOf<Request, ProtocolError>. This required some surgery in various other parts of the code to deal with the OneOf.

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -845,9 +845,11 @@ static char* trimHeaderEnding(kj::ArrayPtr<char> content) {
   return end;
 }
 
-kj::Maybe<HttpHeaders::Request> HttpHeaders::tryParseRequest(kj::ArrayPtr<char> content) {
+HttpHeaders::RequestOrProtocolError HttpHeaders::tryParseRequest(kj::ArrayPtr<char> content) {
   char* end = trimHeaderEnding(content);
-  if (end == nullptr) return nullptr;
+  if (end == nullptr) {
+    return ProtocolError { "ERROR: Request headers have no terminal newline.", content };
+  }
 
   char* ptr = content.begin();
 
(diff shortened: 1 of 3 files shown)


c++/src/kj/compat/http.c++
```diff
@@ -4739,118 +4753,123 @@ public:
         return httpOutput.flush().then([]() { return false; });
       }
 
-      KJ_IF_MAYBE(req, request) {
-        auto& headers = httpInput.getHeaders();
+      KJ_SWITCH_ONEOF(requestOrProtocolError) {
+        KJ_CASE_ONEOF(request, HttpHeaders::Request) {
```
harrishancock: The diff gets screwed up here, but the only substantive change in this code block is the indentation change necessitated by using a `KJ_SWITCH_ONEOF` instead of a `KJ_IF_MAYBE`.

harrishancock:
Note that I only report raw content for header parsing. For WebSocket errors and for timeouts, I don't report any raw content, though I suppose we could.

harrishancock:
The switch from a Maybe to a OneOf caused MSVC some confusion related to a couple `mvCapture()`s, so I replaced them with proper move captures.

harrishancock:
MinGW tests failed with
```
2: kj\compat\http-test.c++:3409: failed: expected req2.poll(io.waitScope)
2: stack: 
2: [ FAIL ] compat\http-test.c++:3330: HttpClient concurrency limiting (3545 μs)
```
which appears unrelated (is it already flakey?). Meanwhile, Cygwin appears hung, though all tests passed.

c++/src/kj/compat/http.c++
jclee: FWIW, diff is easier to read if you click the gear icon in the bar at the top of the diff, then check "hide whitespace changes" and click "apply and reload".  I'm guessing there's a more permanent setting elsewhere.

harrishancock: Oh, yeah, thanks! I remember setting that in the past and had thought it was persistent across PRs. Now I'm less confused.

c++/src/kj/compat/http.h
```diff
@@ -321,7 +321,38 @@ class HttpHeaders {
...
+    // used by HttpServerErrorHandler::handleClientProtocolError().
+    //
+    // TODO(cleanup): Should maybe not live in HttpHeaders? HttpServerErrorHandler::ProtocolError?
+    //   Or HttpProtocolError? Or maybe we need a more general way of attaching sensitive context to
+    //   kj::Exceptions?
+
+    kj::StringPtr description;
+    // An error description safe for all the world to see.
+
+    kj::ArrayPtr<char> rawContent;
```
kentonv: It looks like this is always exactly the `content` parameter passed into `tryParseRequest()`? Going to all the trouble to have a OneOf with a struct in it so that you can return it back seems unnecessary.

I had imagined that `handleClientProtocolError()` would be given a third parameter which is, like, `kj::ArrayPtr<const byte> buffer`, documented as containing whatever bytes the client sent which the server didn't understand.

c++/src/kj/compat/http.h
```diff
@@ -321,7 +321,38 @@ class HttpHeaders {
...
+    // HttpServerErrorHandler::handleClientProtocolError().
+    //
+    // TODO(cleanup): We could make this a self-contained Array<char> for a more intuitive lifetime,
+    //   but this would complicate reporting from within HttpHeaders::tryParseRequest().
+  };
+
+  using RequestOrProtocolError = kj::OneOf<Request, ProtocolError>;
+
+  RequestOrProtocolError tryParseRequest(kj::ArrayPtr<char> content);
   kj::Maybe<Response> tryParseResponse(kj::ArrayPtr<char> content);
```
kentonv: I think if we're updating tryParseRequest() we ought to update tryParseResponse() too. These are technically public interfaces...

c++/src/kj/compat/http.h
harrishancock: I'd prefer that approach, too, but it was less obvious how to do that. We'd need to know the header buffer start and end indices in `HttpServer::Connection::loop()` somehow. We could have `HttpInputStreamImpl::readHeader()` save them when it returns, and allow the most-recently-parsed header slice to be retrieved from a function on that class, I suppose?

harrishancock:
Squashed existing fixups and added a new one to address `tryParseResponse()` returning a `ResponseOrProtocolError`. I don't actually do anything meaningful with it in `HttpClientImpl`, though -- defining a new error handler in `HttpClientSettings` didn't feel like the right approach.

c++/src/kj/compat/http.h
harrishancock: Done.

c++/src/kj/compat/http.h
harrishancock: As discussed, will keep the current approach.

kentonv:
Looks like you broke all the builds.

harrishancock:
D'oh. I had the fix sitting on my disk, uncommitted...

PR #869, opened 2019-08-19
Fix some error propagation bugs in HTTP.

kentonv:


c++/src/kj/compat/http.c++
```diff
@@ -1867,7 +1883,11 @@ public:
     if (length == 0) inner.finishBody();
   }
   ~HttpFixedLengthEntityWriter() noexcept(false) {
-    if (length > 0) inner.abortBody();
+    if (length > 0) {
+      inner.abortBody();
+    } else if (inner.isWriteInProgress()) {
```
harrishancock: Since both branches execute equivalent code, would `if (length > 0 || inner.isWriteInProgress)` make more sense?

kentonv: Maybe. In part I wanted to test here that finishBody() actually detects the problem.

Is it worth starting over the CI to fix this?

kentonv: NM, fixed it.

PR #870, opened 2019-08-20
Return HTTP 501 instead of 400 on unrecognized method.

kentonv:


kentonv:
CI failures bogus, merging...

PR #871, opened 2019-08-21
Add `kj::evalLast()` for running a callback after all other events are done.

kentonv:


c++/src/kj/async-prelude.h
```diff
@@ -220,6 +220,7 @@ void detach(kj::Promise<void>&& promise);
 void waitImpl(Own<_::PromiseNode>&& node, _::ExceptionOrValue& result, WaitScope& waitScope);
 bool pollImpl(_::PromiseNode& node, WaitScope& waitScope);
 Promise<void> yield();
+Promise<void> yieldHarder();
```
jclee: yieldHarder

c++/src/kj/async-test.c++
```diff
@@ -401,12 +403,24 @@ TEST(Async, Ordering) {
       EXPECT_EQ(4, counter++);
     }).then([&]() {
       EXPECT_EQ(5, counter++);
+      promises[8] = kj::evalLast([&]() {
```
jclee: So, I found this test to be kind of a fun brainteaser...  I was trying to work out why the evalLast() doc comment says it fires in LIFO order, while the two evalLast() calls in the test appear to be firing in FIFO order.  As far as I can tell, the armLast() code does indeed seem to queue events in LIFO order, and the reason the evalLast() lambdas run in the opposite order is because the intervening promise structures call armLast() in the opposite order.

Is this a demonstration of the doc comment's rule "If the first callback enqueues new events, then latter callbacks will not execute until those events are drained."?  If so, it might be worth a comment in the test itself...  The other comments in the test were much-valued breadcrumbs in puzzling out how the system is supposed to work.  Or perhaps it might be clearer broken out into a separate test?

Also, I'm guessing that LIFO order of simple evalLast() lambdas is also important to guarantee, in which case we might want to also cover it in a test?

kentonv: Erm. TBH now that you mention it my test code appears to be checking for the wrong ordering, and I don't know why it passes.

kentonv: Oh I bet it's because I didn't do `.eagerlyEvaluate(nullptr)` everywhere and so the ordering turned out to be governed by the ordering of the waits on the promise array.

I really don't know why, when I originally wrote this test, I used a promise array rather than a vector or a TaskSet or something. Maybe TaskSet didn't exist yet... I should rewrite the test to avoid confusing myself.

jclee: Ah, yeah... I saw the lack of ".eagerlyEvaluate", but I assumed it was intentional.

kentonv: Fixed, and rewrote the test to use TaskSet.

c++/src/kj/async-test.c++
```diff
@@ -379,69 +379,73 @@ TEST(Async, Ordering) {
...
-    }).eagerlyEvaluate(nullptr);
-  }).eagerlyEvaluate(nullptr);
+    }));
+  }));
 
-  promises[0] = evalLater([&]() {
+  tasks.add(evalLater([&]() {
     EXPECT_EQ(3, counter++);
 
     // Making this a chain should NOT cause it to preempt promises[1].  (This was a problem at one
```
jclee: I guess this comment now needs to be revised, since with this commit "promises[1]" is no longer a thing. :P

kentonv:
Cygwin seems to have a persistent deadlock problem, will deal with that in a separate PR.

PR #872, opened 2019-08-21
Add a little more detail to invalid response status line errors.

kentonv:


c++/src/kj/compat/http.c++
```diff
@@ -900,18 +900,18 @@ HttpHeaders::ResponseOrProtocolError HttpHeaders::tryParseResponse(kj::ArrayPtr<
...
-        "ERROR: Invalid response status line.", content };
+        "ERROR: Invalid response status line (no spaces).", content };
   }
 
   KJ_IF_MAYBE(code, consumeNumber(ptr)) {
     response.statusCode = *code;
   } else {
     return ProtocolError { 502, "Bad Gateway",
-        "ERROR: Invalid response status code.", content };
+        "ERROR: Invalid response status code (invalid status code).", content };
```
harrishancock: This is actually redundant, though not a big deal.

kentonv: heh, fixed

PR #880, opened 2019-09-09
Fix http-test.c++ to avoid dubious assumptions about gather-writes.

kentonv:
Most (all?) implementations of `write(ArrayPtr<const ArrayPtr<const byte>>)`, if the outer array contains only one inner array, do not use the outer array again after the initial call returns (as opposed to the promise resolving). But, this is not a safe assumption and http-test.c++ should not be relying on it.

(I found this when I tried forcing all writes to complete asynchronously to check if it resulted in any bugs. This is all I found.)

### DIFF
--- a/c++/src/kj/compat/http-test.c++
+++ b/c++/src/kj/compat/http-test.c++
@@ -2118,6 +2118,10 @@ void testWebSocketClient(kj::WaitScope& waitScope, HttpHeaderTable& headerTable,
   }
 }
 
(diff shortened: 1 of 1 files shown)


jclee:
Looks fine...  I assume the reason there isn't a AsyncOutputStream::write(ArrayPtr<const byte>) method for direct use is because it would be easy to forget to attach the backing array to the promise?

kentonv:
@jclee No, it's because the interface is mis-designed. Way back then, copying the `write()` syscall felt like the right answer, but these days it's clear that `ArrayPtr<const byte>` would have been better. I actually plan to fix this in a separate refactoring I've been working on for a while...

PR #883, opened 2019-09-14
Give an example of returning an owning string for KJ_STRINGIFY

ezyang:
I initially didn't read the docs carefully enough and returned a StringPtr to a locally allocated string. Whoops. Make the doc a little clearer about this.

### DIFF
--- a/c++/src/kj/string.h
+++ b/c++/src/kj/string.h
@@ -487,6 +487,8 @@ inline _::Delimited<ArrayPtr<const T>> Stringifier::operator*(const Array<T>& ar
 //
 //    class Foo {...};
 //    inline StringPtr KJ_STRINGIFY(const Foo& foo) { return foo.name(); }
+//      // or perhaps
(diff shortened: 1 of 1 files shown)


c++/src/kj/string.h
```diff
@@ -487,6 +487,8 @@ inline _::Delimited<ArrayPtr<const T>> Stringifier::operator*(const Array<T>& ar
 //
 //    class Foo {...};
 //    inline StringPtr KJ_STRINGIFY(const Foo& foo) { return foo.name(); }
+//      // or perhaps
+//    inline String KJ_STRINGIFY(const Foo& foo) { return kj::str(foo.field1(), ",", foo.field2()); }
```
kentonv: I hate to do this, but it looks like this line is 1 char over our 100-char limit...

One easy fix would be to remove the two instances of `()`; I think it's just as clear either way. Or, you could make it a multi-line function.

kentonv:
Thanks!

PR #885, opened 2019-09-17
Increase arenaSpace for emscripten builds

titrindl:
Building with Emscripten fails with
capnproto\capnproto\c++\src\capnp\message.c++:55:5: error: static_assert failed "arenaSpace is too small to hold a ReaderArena.  Please increase it.  This will break ABI compatibility."

Is constexpr usable? I don't know what C++ standards this project aims to be compatible with.

### DIFF
--- a/c++/src/capnp/message.h
+++ b/c++/src/capnp/message.h
@@ -133,11 +133,17 @@ public:
 private:
   ReaderOptions options;
 
+#if defined(__EMSCRIPTEN__)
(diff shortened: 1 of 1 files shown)


kentonv:
Thanks!

The Cygwin failure is bogus (it has been randomly deadlocking lately) so I'll ignore it and merge this.

(And yes, we target C++14 so constexpr is fine.)

PR #887, opened 2019-10-02
Fix handling of queued RT signals, plus some other crap

kentonv:
For regular (non-RT) POSIX signals, the process can only have at most one instance of each signal queued for delivery at a time. If another copy of the signal arrives before the first is delivered, the new signal is ignored. The idea was that signals are only meant to wake the process up to check some input; the signal itself is not the input.

POSIX RT signals are different. Multiple copies of the same signal can be queued, and each is delivered separately. Each signal may contain some additional information that needs to be processed. The signals themselves are input.

UnixEventPort's `onSignal()` method returns a Promise that resolves the next time the signal is delivered. When the Promise is resolved, the signal is also supposed to be blocked until `onSignal()` can be called again, so that the app cannot miss signals delivered in between.

However, the epoll/signalfd implementation had a bug where it would pull _all_ queued signals off the `signalfd` at once, only delivering the first instance of each signal number and dropping subsequent instances on the floor. That's fine for regular signals, but not RT signals.

This change fixes the bug and adds a test. Incidentally, the poll()-based implementation has been correct all along.

_Also, three other unrelated, tiny commits that were in my repo._

### DIFF
--- a/c++/src/capnp/rpc-twoparty.c++
+++ b/c++/src/capnp/rpc-twoparty.c++
@@ -382,7 +382,9 @@ TwoPartyClient::TwoPartyClient(kj::AsyncCapabilityStream& connection, uint maxFd
       rpcSystem(network, bootstrapInterface) {}
 
 Capability::Client TwoPartyClient::bootstrap() {
-  MallocMessageBuilder message(4);
+  capnp::word scratch[4];
+  memset(&scratch, 0, sizeof(scratch));
+  capnp::MallocMessageBuilder message(scratch);
   auto vatId = message.getRoot<rpc::twoparty::VatId>();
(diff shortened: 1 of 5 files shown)


harrishancock:
Changes look good, but Cygwin protests:
```
[ TEST ] kj/async-unix-test.c++:909: UnixEventPort can receive multiple queued instances of an RT signal
kj/async-unix-test.c++:900: failed: expected promise.poll(waitScope)
stack: 100591383 10059148d 559fe4800
    async-xthread-test.c++:?: returning here
    async-xthread-test.c++:?: returning here
    ??:0: returning here
[ FAIL ] kj/async-unix-test.c++:909: UnixEventPort can receive multiple queued instances of an RT signal (39641 μs)
```

kentonv:
OK, both remaining failures are unrelated to this change.

PR #891, opened 2019-10-07
Expose refcount on AtomicRefcounted

harrishancock:


c++/src/kj/refcount.h
```diff
@@ -134,10 +134,17 @@ class AtomicRefcounted: private kj::Disposer {
...
   inline bool isShared() const {
+    return getRefcount() > 1;
+  }
+
+  inline uint getRefcount() const {
+    // This value can be useful for heuristic purposes, such as estimating the load on a resource.
+
 #if _MSC_VER
-    return KJ_MSVC_INTERLOCKED(Or, acq)(&refcount, 0) > 1;
+    // A static_cast is fine here because sizeof(uint) == sizeof(long) on MSVC.
```
kentonv: Even if it weren't, a static_cast would still be fine because a refcount over 2^32 would be absurd.

Why write the cast at all? We usually let these things be implicit...

jclee: I suppose if you wanted to explicitly enforce this, you could throw in a `static_assert(...)`.

harrishancock: `uint`s can be as short as 16 bits wide, while `long`s can be as long as 64 bits wide, so my brain always screams "narrowing conversion" when I see something like this. If it weren't for the fact that we're specifically targeting MSVC in this code path, then a bug such as a 2^16 refcount (which I would consider grotesquely large, but not improbable) being reported as 0 would be plausible. This comment, and the static_cast, are mostly just to tell the reader "Yes, I thought this through."

harrishancock: True, I'll do that instead of the `static_cast`.

harrishancock:
Amended to remove static_cast and replace comment with static_assert.

c++/src/kj/refcount.h
kentonv: Realistically, almost every single use of `int` and `uint` in this codebase is assuming it is at least 32 bits. A 16-bit platform trying to run this code will crash and burn.

c++/src/kj/refcount.h
```diff
@@ -134,10 +134,17 @@ class AtomicRefcounted: private kj::Disposer {
...
   inline bool isShared() const {
+    return getRefcount() > 1;
+  }
+
+  inline uint getRefcount() const {
+    // This value can be useful for heuristic purposes, such as estimating the load on a resource.
+
 #if _MSC_VER
-    return KJ_MSVC_INTERLOCKED(Or, acq)(&refcount, 0) > 1;
+    static_assert(sizeof(uint) == sizeof(long), "unsafe implicit conversion");
```
kentonv: Sorry but I feel this is worse than what was here before. If there is some platform that MSVC targets where `long` happens to be 64 bits, this will fail to compile for no good reason.

harrishancock: Okay, will remove it. (No such platform exists, though I suppose they might change that someday.)

harrishancock:
Not needed for now, so holding off on it.

PR #894, opened 2019-10-10
Suppress MSVC warnings in headers files

### LINKED ISSUE #893: capnp headers aren't warning safe on MSVC
Our build infra builds with `/wd4200 /wd4996 /wd5030 /wd4244 /wd4267 /wd4305 /wd4324 /W3 /Z7` and capnp generates warnings. That's mostly fine because I explicitly disable warnings for the capnp build itself via `/w` but MSVC (at least 2017) has no facility (unlike GCC/clang) to designate an include path or headers as system headers. This means that all includes of capnp headers need to be wrapped in `__pragma(warning(push, 0))`, `__pragma(warning(pop))` (can't recall if the generated files themselves have a similar issue or if it's all just transitive from the library itself). 

Would it be in-scope to submit a PR fixing this & if so how would you like it fixed? Can either try to address the warnings (primarily unreferenced formal parameter) but there are others that show up in C++17 mode that could be trickier like in hash.h:74 "warning C4127: conditional expression is constant". 

My initial thought is to suppress the header warnings via push/pop as the original reasoning for using the system_header pragma for GCC/clang is to avoid having to deal with supporting the intersection of all the warnings from different compiler versions that users may be using when building the code.

vlovich:
For Visual Studio we have to wrap the headers with push/pop pragmas
at the top and bottom of the file.

This diff was programmatically generated as:

```
cd c++
sed -i -f capnp_header_cleanup.sed $(grep
'defined(CAPNP_HEADER_WARNINGS)' --include=*.h -rl src/capnp/)
sed -i -f kj_header_cleanup.sed $(grep 'KJ_HEADER_WARNINGS' --include=*.h -rl
src/kj/)
```

capnp_header_cleanup.sed:
```
/#if defined(__GNUC__) && !defined(CAPNP_HEADER_WARNINGS)/{
i #ifndef CAPNP_HEADER_WARNINGS
a#ifdef __GNUC__
a#pragma GCC system_header
a#elif defined(_MSC_VER)
a#pragma warning(push, 0)
a#endif  // Compilers
a#endif  // !CAPNP_HEADER_WARNINGS
}

/#if defined(__GNUC__) && !defined(CAPNP_HEADER_WARNINGS)/,/#endif/d

$s/$/\n\n#if defined(_MSC_VER) && !defined(CAPNP_HEADER_WARNINGS)\n#pragma warning(pop)\n#endif/
```

kj_header_cleanup.sed:
```
/#if defined(__GNUC__) && !KJ_HEADER_WARNINGS/{
i #if !defined(KJ_HEADER_WARNINGS) || !KJ_HEADER_WARNINGS
a#ifdef __GNUC__
a#pragma GCC system_header
a#elif defined(_MSC_VER)
a#pragma warning(push, 0))
a#endif  // Compilers
a#endif  // !KJ_HEADER_WARNINGS
}

/#if defined(__GNUC__) && !KJ_HEADER_WARNINGS/,/#endif/d

$s/$/\n\n#if defined(_MSC_VER) && (!defined(KJ_HEADER_WARNINGS) || !KJ_HEADER_WARNINGS)\n#pragma warning(pop)\n#endif/
```

Fixes #893.

### DIFF
--- a/c++/src/capnp/common.h
+++ b/c++/src/capnp/common.h
@@ -25,10 +25,6 @@
 
 #pragma once
 
-#if defined(__GNUC__) && !defined(CAPNP_HEADER_WARNINGS)
-#pragma GCC system_header
-#endif
-
 #include <inttypes.h>
 #include <kj/string.h>
 #include <kj/memory.h>
@@ -37,6 +33,16 @@
 #if CAPNP_DEBUG_TYPES
 #include <kj/units.h>
 #endif
+
+#if !defined(CAPNP_HEADER_WARNINGS) || !CAPNP_HEADER_WARNINGS
+#define CAPNP_BEGIN_HEADER KJ_BEGIN_SYSTEM_HEADER
+#define CAPNP_END_HEADER KJ_END_SYSTEM_HEADER
+#else
+#define CAPNP_BEGIN_HEADER
+#define CAPNP_END_HEADER
+#endif
+
+CAPNP_BEGIN_HEADER
 
 namespace capnp {
 
@@ -741,3 +747,5 @@ inline constexpr kj::ArrayPtr<U> arrayPtr(U* ptr, T size) {
 #endif
 
 }  // namespace capnp
+
+CAPNP_END_HEADER
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -25,9 +25,30 @@
 
 #pragma once
 
-#if defined(__GNUC__) && !KJ_HEADER_WARNINGS
-#pragma GCC system_header
+#ifdef __GNUC__
+#define KJ_BEGIN_SYSTEM_HEADER _Pragma("GCC system_header")
+#elif defined(_MSC_VER)
+#define KJ_BEGIN_SYSTEM_HEADER __pragma(warning(push, 0))
+#define KJ_END_SYSTEM_HEADER __pragma(warning(pop))
 #endif
+
+#ifndef KJ_BEGIN_SYSTEM_HEADER
+#define KJ_BEGIN_SYSTEM_HEADER
+#endif
+
+#ifndef KJ_END_SYSTEM_HEADER
+#define KJ_END_SYSTEM_HEADER
+#endif
+
+#if !defined(KJ_HEADER_WARNINGS) || !KJ_HEADER_WARNINGS
+#define KJ_BEGIN_HEADER KJ_BEGIN_SYSTEM_HEADER
+#define KJ_END_HEADER KJ_END_SYSTEM_HEADER
+#else
+#define KJ_BEGIN_HEADER
+#define KJ_END_HEADER
+#endif
+
+KJ_BEGIN_HEADER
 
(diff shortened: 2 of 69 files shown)


kentonv:
Hmm, I wonder if we can make this less ugly by using `_Pragma` on GCC and `__pragma` on MSVC, both of which are allowed to appear in macro expansions. We could then define two macros KJ_BEGIN_HEADER and KJ_END_HEADER to put at the start and end of each header, rather than several lines of repetitive preprocessor stuff...

kentonv:
(Presumably we'd also have `CAPNP_BEGIN_HEADER` and `CAPNP_END_HEADER` for capnp headers. These macros would be defined in `src/kj/common.h` and `src/capnp/common.h`.)

vlovich:
Yeah, I was a bit concerned that some common.h wouldn't be warning safe creating a slight chicken-egg problem but I'll try posting something along the lines you suggest tomorrow.

vlovich:
Made the change as you requested. I noticed that c++/src/capnp/message.h had #includes above the pragma once. I assume that was accidental so I moved the pragma above those headers.

c++/src/kj/common.h
```diff
@@ -25,10 +25,31 @@
 
 #pragma once
 
-#if defined(__GNUC__) && !KJ_HEADER_WARNINGS
-#pragma GCC system_header
+#ifdef __GNUC__
+#define KJ_SUPPRESS_WARNINGS _Pragma("GCC system_header")
```
kentonv: Seems like the name `KJ_SUPPRESS_WARNINGS` doesn't really describe what the macro does in the GCC case. Maybe `KJ_SUPPRESS_HEADER_WARNINGS`?

c++/src/capnp/common.h
```diff
@@ -38,6 +34,16 @@
 #include <kj/units.h>
 #endif
 
+#if !defined(CAPNP_HEADER_WARNINGS) || CAPNP_HEADER_WARNINGS
```
kentonv: should be: `|| !CAPNP_HEADER_WARNINGS`

c++/src/kj/common.h
vlovich: Yeah, I struggled this one because if you make it HEADER_WARNINGS it's not accurately describing MSVC is doing. Thoughts?

kentonv: How about we name the pair: `KJ_BEGIN_SYSTEM_HEADER`, `KJ_END_SYSTEM_HEADER`

vlovich: Do you have any concerns about confusion with KJ_BEGIN_HEADER which is conditions by KJ_HEADER_WARNINGS?

kentonv: Nope, I'm fine with those as-is.

vlovich:
@kentonv this is another instance where the cygwin build is failing. Did you have any other changes you'd like me to make?

kentonv:
Looks great, thanks!

PR #895, opened 2019-10-11
Add compat/std-iterator.h

### LINKED ISSUE #892: std::copy doesn't work on List types?
```
struct Foo {
    field @0 List(Float32);
}
```

In C++:
```
Foo::Reader r = ...;
if (r.getField().size() != 11) {
    return;
}
std::array<float, 11> extracted;
std::copy(r.getField().begin(), r.getField().end(), extracted.begin());
```
I get a compiler error that there's no matching std::copy/std::copy_n:
```
error: no matching function for call to 'copy_n'
  std::copy(r.getField().begin(), r.getField().end(), extracted.begin());
  ^~~~~~~~~~~
C:/Users/vlovich/AppData/Local/Android/Sdk/ndk-bundle/toolchains/llvm/prebuilt/windows-x86_64/sysroot/usr/include/c++/v1\algorithm:1735:1: note: candidate template ignored: requirement '__is_input_iterator<IndexingIterator<const Reader, float> >::value' was not satisfied [with _InputIterator = capnp::_::IndexingIterator<const capnp::List<float, capnp::Kind::PRIMITIVE>::Reader, float>, _Size = unsigned int, _OutputIterator = float *]
copy_n(_InputIterator __first, _Size __orig_n, _OutputIterator __result)
^
C:/Users/vlovich/AppData/Local/Android/Sdk/ndk-bundle/toolchains/llvm/prebuilt/windows-x86_64/sysroot/usr/include/c++/v1\algorithm:1760:1: note: candidate template ignored: requirement '__is_random_access_iterator<IndexingIterator<const Reader, float> >::value' was not satisfied [with _InputIterator = capnp::_::IndexingIterator<const capnp::List<float, capnp::Kind::PRIMITIVE>::Reader, float>, _Size = unsigned int, _OutputIterator = float *]
copy_n(_InputIterator __first, _Size __orig_n, _OutputIterator __result)
```
Is it an oversight in how IndexingIterator is defined or are capnproto `List` objects incompatible with STL algorithms?

vlovich:
Allow including a header to obtain the iterator traits instead of
needing to carefully define a slightly more obscure macro that someone
might be tempted to define globally.

Fixes #892

### DIFF
--- a/c++/src/capnp/compat/std-iterator.h
+++ b/c++/src/capnp/compat/std-iterator.h
@@ -1,0 +1,39 @@
+// Copyright (c) 2013-2014 Sandstorm Development Group, Inc. and contributors
+// Licensed under the MIT License:
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+#pragma once
+
+// This exposes IndexingIterator as something compatible with std::iterator so that things like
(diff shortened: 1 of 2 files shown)


c++/src/capnp/compat/std-iterator.h
```diff
@@ -0,0 +1,39 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+/**
+ * This exposes IndexingIterator as something compatible with std::iterator so that things like
```
kentonv: Nitpick: We don't usually use C-style comment blocks. I'd also put the comment immediately after the `#pragma once` -- just because that's what I've tended to do for header summary comments elsewhere.

vlovich:
I don't see any reason the previous diff failed. @kentonv do you have any thoughts cygwin complained? https://ci.appveyor.com/project/kentonv/capnproto/builds/28051511

kentonv:
There's some sort of problem with Cygwin where half the time it deadlocks on one of the integration tests. I haven't yet decided whether to try to debug it or just drop Cygwin support because WSL has presumably made it obsolete...

kentonv:
Hmm I merged this but I just realized that you didn't update Makefile.am nor the cmake files so it will fail on "make install" I think -- it won't install the header.

PR #896, opened 2019-10-14
Add std-iterator.h to build files

vlovich:


c++/Makefile.am
```diff
@@ -191,6 +191,7 @@ includekjcompat_HEADERS =                                      \
 includecapnp_HEADERS =                                         \
   src/capnp/c++.capnp.h                                        \
   src/capnp/common.h                                           \
+  src/capnp/compat/std-iterator.h                              \
```
kentonv: Unfortunately, due to automake being terrible, this won't work -- it'll install into `/include/capnp`. You need to put this header under `includecapnpcompat_HEADERS`, below.

vlovich: Gotcha. Thanks for the feedback.

kentonv:
Thanks!

PR #897, opened 2019-10-15
Fixup examples code markup

valery1707:
Currently resulting markup has extra `</div>` tag

### DIFF
--- a/doc/language.md
+++ b/doc/language.md
@@ -403,7 +403,7 @@ Cap'n Proto generics work very similarly to Java generics or C++ templates. Some
   substituting the type parameters manually. For example, `Map(Text, Person)` is encoded exactly
   the same as:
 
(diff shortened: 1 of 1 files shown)


valery1707:
Failed tests:
```
2: [ FAIL ] compat\http-test.c++:3392: HttpClient concurrency limiting (5711 μs)
```
This PR change only documentation.

kentonv:
It looks like there is a deeper problem here. These code snippets are supposed to be within bullet lists, but it looks like the markdown parser is interpreting the code block as ending the bullet list, leading the following text to be indented wrong.

It would seem that Jekyll's markdown parser must have changed at some point after I wrote these docs. Ugh. This makes me scared to update the site ever because who knows which pages might be subtly reformatted?

Maybe I need to redo the whole site with a different generator that isn't so... unreliable.

PR #905, opened 2019-10-31
Make HttpClient adapter preserve exception behavior

jclee:
Ideally, the HttpClient/HttpServer adapters should maintain the invariant that
the behavior of a given client is the same as the behavior of
newHttpClient(newHttpService(client)).  Prior to this change, the HttpClient
wrapper lazily called request(), so a client whose request() eagerly threw an
exception could produce a different exception when called directly versus when
wrapped -- the laziness allowed additional code to run.  This was particularly
evident when making a request with a body, since code using a wrapped client
would be able to set up a subsequent BlockedWrite, eventually resulting in a
"read end of pipe was aborted" exception instead of the actual exception.

This change makes HttpClientAdapter::request() execute the wrapped request()
eagerly.

(Note that it partially undoes 90d48343... Hopefully, it preserves the desired
behavior added there.)

### DIFF
--- a/c++/src/kj/compat/http-test.c++
+++ b/c++/src/kj/compat/http-test.c++
@@ -3032,6 +3032,39 @@ KJ_TEST("newHttpClient from HttpService WebSockets") {
   testWebSocketClient(waitScope, *headerTable, hMyHeader, *client);
 }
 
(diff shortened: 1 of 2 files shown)


kentonv:
Thanks, and nice debugging work.

PR #906, opened 2019-11-01
added tutorial sample

ghbplayer:
This is a draft of the new tutorial sample. I'm currently unable to build it against head, but am able to build against the C++ source from capnproto.org (you can see an artifact of that in the build file 'src ='). I spent hours trying to track down the problem thinking it was my mistake.

This is a draft, I'm interested in feedback. I realize you probably want cmake over meson; cmake isn't something I know how to do. I'll add more comments before the final version. But mostly, I'd like to know what features you think are important to demonstrate in a tutorial.

### DIFF
--- a/c++/samples/tutorial/build_capnp
+++ b/c++/samples/tutorial/build_capnp
@@ -1,0 +1,4 @@
+#!/bin/bash
+pushd "`dirname \"$0\"`" >/dev/null
+capnp compile interface.capnp -oc++
+popd >/dev/null
--- a/c++/samples/tutorial/clean
+++ b/c++/samples/tutorial/clean
@@ -1,0 +1,3 @@
+#!/bin/bash
+rm interface.capnp.c++ interface.capnp.h
+rm -r out .pch
--- a/c++/samples/tutorial/client.c++
+++ b/c++/samples/tutorial/client.c++
@@ -1,0 +1,108 @@
+#include "pch.h"
+
+int main()
+{
+	//std::cout << "starting" << std::endl;
+
+	//kj::AsyncIoContext ioContext(kj::setupAsyncIo());
+	//auto address = ioContext.provider->getNetwork().parseAddress("127.0.0.1", 2000).wait(ioContext.waitScope);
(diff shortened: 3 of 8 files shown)


kentonv:
Hi @ghbplayer,

Thanks for working on this sample program. Having good sample code is really important.

I have a few concerns about this code which may make it difficult to merge in its current incarnation:
- The code is written using very different style from the rest of the Cap'n Proto / KJ codebase. It may seem like an irrelevant aesthetic issue, but if people are copy/pasting from this sample code as well as from actual KJ code, they'll end up with inconsistent and confusing code. I'd prefer that our samples are consistent with our regular style.
- It looks like you're using a build system that we don't use elsewhere in the codebase, and which isn't especially widely used. I know that make, automake, and cmake are all pretty ugly, but for sample code it would be good to stick to something everyone is familiar with and can try out without installing new tools.
- Your sample protocol is abstract, not based on any real-world function. I'm worried that this makes it harder for newcomers to follow what is going on, because they can't anchor it to anything real and familiar that let's them know what to expect. Granted, my own calculator example wasn't particularly well chosen -- it gets confusing by being too "meta", e.g. having an interface that represents a "function". It would be nice to come up with something closer to what people might want to do in the real world, though a simplified version of course.

Thoughts?

ghbplayer:
The build system is something I knew would come up. I don't know cmake or automake but I can write a makefile.

Considering tutorial versus sample, I think capnp and kj are complex enough to justify both. The tutorial demonstrates how to get started and do basic things. A beginner can find a function similar to what they want and see the syntax and required operations. Calculator or some other sample can demonstrate higher concepts.

I know you wrote this software, and you have a background in similar things like E, so capnp and kj probably make intuitive sense to you. For those of us without the right background, it helps to separate learning into smaller steps and learn one step at a time. The syntax and the concepts together are a lot to digest.

As for style, okay. It's a tutorial for your software so it should match your style. I can try to match your style but it's probably better if you tell me what things catch your eye that you'd like changed.

PR #908, opened 2019-11-07
Add getsockname() to ConnectionReceiver.

kentonv:
This is a no-brainer given AsyncIoStream has getsockopt() and getsockname(), whereas ConnectionReceiver has only getsockopt().

### DIFF
--- a/c++/src/kj/async-io-unix.c++
+++ b/c++/src/kj/async-io-unix.c++
@@ -1135,6 +1135,11 @@ public:
   void setsockopt(int level, int option, const void* value, uint length) override {
     KJ_SYSCALL(::setsockopt(fd, level, option, value, length));
   }
+  void getsockname(struct sockaddr* addr, uint* length) override {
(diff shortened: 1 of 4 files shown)


c++/src/kj/async-io-unix.c++
```diff
@@ -1135,6 +1135,11 @@ public:
   void setsockopt(int level, int option, const void* value, uint length) override {
     KJ_SYSCALL(::setsockopt(fd, level, option, value, length));
   }
+  void getsockname(struct sockaddr* addr, uint* length) override {
+    socklen_t socklen = *length;
+    KJ_SYSCALL(::getsockname(fd, addr, &socklen));
+    *length = socklen;
```
jclee: (I assume the temporary value is to ensure that nothing gets written if an exception is thrown?)

kentonv: No, it's because `socklen_t` might not be the same as `uint`.

PR #911, opened 2019-11-14
Fix RPC loopback bootstrap().

kentonv:
When VatNetwork::connect() returns nullptr, it means that the caller is trying to connect to itself.

rpc-test.c++ failed to test this in two ways:
- The test VatNetwork's connect() never returned null.
- There was no test case for loopback connect.

As a result, the code to handle loopback in rpc.c++ had bitrotted. It failed to handle the new bootstrap mechanism introduced in v0.5, and instead only implemented the restorer mechanism from 0.4.

/cc @bretthoerner


bretthoerner:
cherry-picking this fixed my issue with self-connections, thanks!

PR #913, opened 2019-12-04
Implement fibers

kentonv:
Fibers allow code to be written in a synchronous / blocking style while running inside the KJ event loop, by executing the code on an alternate call stack and switching back to the main stack whenever it waits.

We introduce a new function, `kj::startFiber(stackSize, func)`. `func` is executed on the fiber stack. It is passed as its parameter a `WaitScope&`, which can then be passed into the `.wait()` method of any promise in order to wait on the promise in a blocking style. `startFiber()` returns a promise for the eventual value returned by `func()` (much as `evalLater()` and friends do).

WIP DO NOT MERGE: I haven't done Windows yet. But it should be easy since Windows' fibers API is actually really good. Putting up this PR now to get ahead of MacOS failures.

### DIFF
--- a/c++/src/kj/async-test.c++
+++ b/c++/src/kj/async-test.c++
@@ -844,5 +844,93 @@ KJ_TEST("exclusiveJoin both events complete simultaneously") {
   KJ_EXPECT(!joined.poll(waitScope));
 }
 
+KJ_TEST("start a fiber") {
+  EventLoop loop;
+  WaitScope waitScope(loop);
+
+  auto paf = newPromiseAndFulfiller<int>();
+
+  Promise<StringPtr> fiber = startFiber(65536,
+      [promise = kj::mv(paf.promise)](WaitScope& fiberScope) mutable {
+    int i = promise.wait(fiberScope);
+    KJ_EXPECT(i == 123);
+    return "foo"_kj;
+  });
+
+  KJ_EXPECT(!fiber.poll(waitScope));
+
+  paf.fulfiller->fulfill(123);
+
+  KJ_ASSERT(fiber.poll(waitScope));
+  KJ_EXPECT(fiber.wait(waitScope) == "foo");
+}
+
+KJ_TEST("fiber promise chaining") {
+  EventLoop loop;
+  WaitScope waitScope(loop);
+
+  auto paf = newPromiseAndFulfiller<int>();
+  bool ran = false;
+
+  Promise<int> fiber = startFiber(65536,
+      [promise = kj::mv(paf.promise), &ran](WaitScope& fiberScope) mutable {
+    ran = true;
+    return kj::mv(promise);
+  });
+
+  KJ_EXPECT(!ran);
+  KJ_EXPECT(!fiber.poll(waitScope));
+  KJ_EXPECT(ran);
+
+  paf.fulfiller->fulfill(123);
+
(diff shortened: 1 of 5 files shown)


dwrensha:
Fascinating!

I was worried when I saw the title of this pull request, as it calls to mind Meteor fibers, which allow *any* function to yield control back to the event loop. That makes it difficult to implement critical sections; you need to know that all functions you call are not secretly async.

The `WaitScope&` parameter neatly solves this problem. Functions that can yield should have such a parameter, thus statically declaring that they are async.

kentonv:
@dwrensha Indeed. Note this design was introduced six years ago (exactly, as of today), before I'd worked much with Meteor fibers: fecb6086e38d507fe27bc07e943c2780f6e79864 :smirk:

kentonv:
OK, this now passes all CI, yay!

c++/src/kj/async-test.c++
```diff
@@ -844,5 +844,93 @@ KJ_TEST("exclusiveJoin both events complete simultaneously") {
...
+      return "foo"_kj;
+    });
+
+    KJ_EXPECT(!fiber.poll(waitScope));
+    KJ_EXPECT(!exited);
+  }
+
+  KJ_EXPECT(exited);
+}
+
```
harrishancock: What would this do?
```cpp
kj::Promise<void> fiber;
fiber = startFiber(65536, [&](WaitScope& fiberScope) {
  fiber.wait(fiberScope);
});
```
or also
```cpp
kj::Promise<void> fiber;
fiber = startFiber(65536, [&](WaitScope& fiberScope) {
  fiber = nullptr;
});
```

If they crash, that's great. If they throw an exception, then they might be worth a test. Hopefully there's no third option. :)

kentonv: The first one will deadlock. Arguably this is the correct behavior: The fiber is waiting for itself to be done. But it won't be done until it finishes waiting. So it's never done. This sort of thing isn't unique to fibers; you can create promise loops with regular promises too. It's probably impractical to try to detect, especially when you have a complex chain of promises with forks and joins and such.

The second one will call `::abort()` on line 866 of `async.c++`.

unicomp21:
@kentonv 

Why fibers instead of UMS threads?  Fibers are pretty limited when comparing the two.
https://docs.microsoft.com/en-us/windows/win32/procthread/user-mode-scheduling

But then again, maybe the c++20 co_await is where everything should be headed.

Has anyone used capnp w/ emscripten?
https://emscripten.org/docs/api_reference/fiber.h.html

kentonv:
@unicomp21 Thanks for your comment, but I think you've misunderstood what's happening here.

KJ (the C++ toolkit library under Cap'n Proto) has featured an async I/O framework based on Promises for seven years now. It is better than asio.

We introduced fibers as a compatibility hack. Fibers allow us to call into libraries that are designed to operate in a synchronous way -- with synchronous callbacks -- within an async event loop. We explicitly do not want fibers to behave like threads.

We also plan to support async/await with KJ Promises soon (they are a perfect fit). But, while async/await is absolutely beautiful in new code, it wouldn't allow us to call into existing synchronous code the way fibers do.

unicomp21:
Aaah, got it, thanks @kentonv !

PR #915, opened 2019-12-05
remove kj-test from CAPNP_LIBRARIES

faywong:
as kj/test.c++ contains "KJ_MAIN(kj::TestRunner);", this will generate the main entry for a process, so as to hidden the one defined in the main program(which is maybe the most common use case of CAPNP_LIBRARIES)

### DIFF
--- a/c++/cmake/CapnProtoConfig.cmake.in
+++ b/c++/cmake/CapnProtoConfig.cmake.in
@@ -65,7 +65,7 @@ set(CAPNP_INCLUDE_DIRS ${CAPNP_INCLUDE_DIRECTORY})
 # No need to list all libraries, just the leaves of the dependency tree.
 set(CAPNP_LIBRARIES_LITE CapnProto::capnp)
 set(CAPNP_LIBRARIES CapnProto::capnp-rpc CapnProto::capnp-json
-                    CapnProto::kj-http CapnProto::kj-test)
+                    CapnProto::kj-http)
 
 set(CAPNP_DEFINITIONS)
 if(TARGET CapnProto::capnp AND NOT TARGET CapnProto::capnp-rpc)


jclee:
Hmm... I'm not an expert in these variables, but per the comments, it looks like they only exist for use in FindCapnProto.cmake, which was apparently removed in 6b788321.  Might be better to remove these variables entirely, if they are just confusing people (as well as removing their mention in the comment at the top of the file).

kentonv:
Do @harrishancock or @pqu have any thoughts on this?

harrishancock:
Ah, sorry, yes, I meant to weigh in.

Since the variable is both broken and deprecated, @jclee is right, we might as well just remove it (and probably all the other FindCapnProto compatibility variables). Dependent projects really should be using the imported targets directly instead.

On the other hand, merging this requires almost no effort and is a strict improvement over the status quo. :) So I think there are two right answers.

kentonv:
One possible disadvantage of merging this: At present, CAPNP_LIBRARIES cannot possibly be linked into any program without totally breaking it, which means that we can be pretty confident that removing it entirely won't break anyone. But if we merged this now, someone could then possibly grow a real dependency on CAPNP_LIBRARIES...

harrishancock:
I think projects already do depend on CAPNP_LIBRARIES as defined by the old FindCapnProto.cmake find-module (a search on GitHub shows it all over the place). If/when they eventually upgrade to the more modern config-module which provides this broken variable, they'll be broken.

That seems fine, so long as the breakage is clear. Right now, I guess it'd manifest as a linker error, which is pretty terrible. If we removed it entirely, the breakage would be more easily diagnosable.

kentonv:
OK, I get it now. I will merge this because it's a strict improvement and requires minimal effort!

kentonv:
Thanks @faywong!

faywong:
@kentonv You're welcome.

PR #917, opened 2019-12-09
Make newPromisedStream() generally available

harrishancock:
This just moves Promise{Io,Output}Stream from http.c++ to PromisedAsync{Io,Output}Stream in async-io.c++ and exposes them via newPromisedStream() so other code can use them.

No substantive code changes.

### DIFF
--- a/c++/src/kj/async-io.c++
+++ b/c++/src/kj/async-io.c++
@@ -1821,6 +1821,210 @@ Tee newTee(Own<AsyncInputStream> input, uint64_t limit) {
   return { { mv(branch1), mv(branch2) } };
 }
 
+namespace {
+
+class PromisedAsyncIoStream final: public kj::AsyncIoStream, private kj::TaskSet::ErrorHandler {
+  // An AsyncIoStream which waits for a promise to resolve then forwards all calls to the promised
+  // stream.
+
+public:
+  PromisedAsyncIoStream(kj::Promise<kj::Own<AsyncIoStream>> promise)
+      : promise(promise.then([this](kj::Own<AsyncIoStream> result) {
+          stream = kj::mv(result);
+        }).fork()),
+        tasks(*this) {}
+
+  kj::Promise<size_t> read(void* buffer, size_t minBytes, size_t maxBytes) override {
+    KJ_IF_MAYBE(s, stream) {
+      return s->get()->read(buffer, minBytes, maxBytes);
+    } else {
+      return promise.addBranch().then([this,buffer,minBytes,maxBytes]() {
+        return KJ_ASSERT_NONNULL(stream)->read(buffer, minBytes, maxBytes);
+      });
+    }
+  }
+  kj::Promise<size_t> tryRead(void* buffer, size_t minBytes, size_t maxBytes) override {
+    KJ_IF_MAYBE(s, stream) {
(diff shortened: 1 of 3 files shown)


harrishancock:
Mac build failed in the HTTP concurrency limiter test. Since this PR just moves code around, it seems unlikely that my changes could have caused it, so I restarted it.

Meanwhile, MinGW failed with a segfault:
```
4: [ TEST ] rpc-twoparty-test.c++:651: Streaming over RPC then unwrap with CapabilitySet
4/5 Test #4: capnp-heavy-tests-run ............***Exception: SegFault  0.31 sec
```

c++/src/kj/async-io.c++
```diff
@@ -1821,6 +1821,210 @@ Tee newTee(Own<AsyncInputStream> input, uint64_t limit) {
...
+    KJ_IF_MAYBE(s, stream) {
+      return s->get()->abortRead();
+    } else {
+      tasks.add(promise.addBranch().then([this]() {
+        return KJ_ASSERT_NONNULL(stream)->abortRead();
+      }));
+    }
+  }
+
+public:
```
jclee: Maybe meant to be "private" here?

c++/src/kj/async-io.c++
```diff
@@ -1821,6 +1821,210 @@ Tee newTee(Own<AsyncInputStream> input, uint64_t limit) {
...
+        if (e.getType() == kj::Exception::Type::DISCONNECTED) {
+          return kj::READY_NOW;
+        } else {
+          return kj::mv(e);
+        }
+      });
+    }
+  }
+
+public:
```
jclee: Ditto

c++/src/kj/async-io.h
```diff
@@ -281,6 +281,11 @@ Tee newTee(Own<AsyncInputStream> input, uint64_t limit = kj::maxValue);
 //
 // It is recommended that you use a more conservative value for `limit` than the default.
 
+Own<AsyncOutputStream> newPromisedStream(Promise<Own<AsyncOutputStream>> promise);
+Own<AsyncIoStream> newPromisedStream(Promise<Own<AsyncIoStream>> promise);
+// Constructs an Async*Stream which waits for a promise to resolve, then forwards all calls to the
+// promised stream.
```
jclee: Guess there's no immediate need for an AsyncInputStream equivalent.

jclee:
Yeah, I think I've seen the concurrency limiter fail occasionally on the Mac, too...  There are a few places in the http test where we need to insert an `io.waitScope.poll()` to make the test pass consistently on macOS and Windows.  I'm guessing there's a less-common race condition in the limiter test that still needs similar treatment -- or we need to work out why the event loop isn't consistent on these platforms.

c++/src/kj/async-io.c++
harrishancock: Hmm, probably so.

c++/src/kj/async-io.h
harrishancock: Yeah, I just need PromiseOutputStream and didn't want to make another copy, so I didn't put too much effort into this.

c++/src/kj/async-io.c++
harrishancock: Amended, and below too.

Issue #743, opened 2018-09-12
Centos 6 compilation problem


smirnov-vs:
I was not able to compile lib on centos 6 with devtoolset-7.
aligned_alloc() is not supported on this system.
I've added `__unix__` macro to branch with posix_memalign and it worked. (table.c++)
I have no idea is it final resolution or it is only fixing my case, so I'm writing this information here for you :)

smirnov-vs:
test fail and not completed

`
2: [ TEST ] async-io-test.c++:50: legacy test: AsyncIo/SimpleNetwork
2: kj/async-io-unix.c++:522: failed: ::socket(addr.generic.sa_family, type, 0): Address family not supported by protocol
2: stack: 7d2197 5ef67c 615659 612d3b 798b6e 79c2d3 5ef734 5ef7a6 5c94b2 5efa6b 7ee406 7f2525
2:     async-io-unix.c++:?: returning here
2:     async-io-test.c++:?: returning here
2:     async-io-test.c++:?: returning here
2:     async-io-test.c++:?: returning here
2:     async.c++:?: returning here
2:     async.c++:?: returning here
2:     async-io-test.c++:?: returning here
2:     async-io-test.c++:?: returning here
2:     ??:?: returning here
2:     async-io-test.c++:?: returning here
2:     ??:?: returning here
2:     ??:?: returning here
2: [ FAIL ] async-io-test.c++:50: legacy test: AsyncIo/SimpleNetwork (47403 μs)
`

and then stuck here
`
2: [ TEST ] filesystem-disk-test.c++:354: DiskFile::copy()
`

kentonv:
Hmm, sounds like we'll need a patch that actually fixes CentOS then.

timkpaine:
Maybe this helps? https://github.com/Microsoft/CNTK/issues/729#issuecomment-276086884

haata:
I'm unfortunately running into this issue when compiling pycapnp with manylinux2010 (https://github.com/capnproto/pycapnp/issues/197) as it uses CentOS 6 as the base.

For now I'll probably just add a patch when bundling that detects if building with a glibc older than 2.16
```
0000000000000000      DF *UND*  0000000000000000  GLIBC_2.16  aligned_alloc
```

```c
#if __APPLE__ || __BIONIC || (defined(__GLIBC__) && (__GLIBC__ == 2) && (__GLIBC_MINOR__ < 16))
[...]

PR #921, opened 2019-12-27
Make sure to use posix_memalign with glibc 2.15 and older

haata:
- This is needed to handle CentOS 6 builds
  Specifically for manylinux2010 which is used for compiling pypi binary
  packages
- Should resolve Issue #743
- There were some Windows test issues (not 100% sure on why) using only posix_memalign and _aligned_malloc

### DIFF
--- a/c++/src/kj/table.c++
+++ b/c++/src/kj/table.c++
@@ -302,8 +302,9 @@ void BTreeImpl::growTree(uint minCapacity) {
   // aligned_alloc() function. Unfortunately, many platforms don't implement it. Luckily, there
   // are usually alternatives.
 
-#if __APPLE__ || __BIONIC__ || __OpenBSD__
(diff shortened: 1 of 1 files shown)


kentonv:
> There were some Windows test issues (not 100% sure on why) using only posix_memalign and _aligned_malloc

What were the issues? Is there a log of this?

The change shouldn't have had any effect on Windows since we'd always be using `_aligned_malloc` when `_WIN32` is defined...

haata:
I'm a little bit fuzzy on it (before the holidays, haha), but I believe this log shows it.
https://github.com/haata/capnproto/runs/364592418
I'll have to look, but I believe I have the other patch around as well.

kentonv:
It looks like the failed test in that run was in the clock tests, probably due to a random timing hiccup. (It's hard to write fully reliable tests for real-time clocks...)

PR #922, opened 2019-12-27
Initial GitHub Actions

haata:
Should work with Windows, Linux, macOS using the default settings.
Also includes a manylinux2010 compatibility build which is used to build
linux pypi packages (to maximize glibc compatibility of binary
packages). manylinux2010 uses a newer gcc (8+) with an older glibc 2.12

These don't fully replace (yet) the travis and appveyor builds. However, they are much easier to maintain and write.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,28 @@
+name: Build Test
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      max-parallel: 4
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macOS-latest, windows-latest]
+    steps:
+    - uses: actions/checkout@v1
+    - name: CMake Setup
+      run: |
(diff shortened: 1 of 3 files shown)


kentonv:
Thanks for working on this!

Can we change this to use `./super-test.sh -j2 quick` instead of directly executing a build? This is what the existing CI does on all platforms except MSVC.

Note that the official build system on all platforms except MSVC is autotools, not cmake. The super-test script covers both, though.

haata:
Shouldn't be a problem, I'll see what I can put together this week.

kentonv:
Inspired by this, I ended up building out some extensive GitHub actions to cover all the tests I used to run locally for each release: https://github.com/capnproto/capnproto/tree/master/.github/workflows

I guess this PR is obsolete now, but thanks for getting me started!

haata:
Cool! One of my favourite parts about GitHub Actions is how it seamlessly works when you fork a repository for your own local testing.

(Sorry for not getting to this, I've been a bit overwhelmed with real life lately...)

PR #923, opened 2020-01-16
Add vcpkg installation instructions

JackBoosY:
capnproto is available as a port in vcpkg, a C++ library manager that simplifies installation for capnproto and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build capnproto, ready to be included in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64, UWP, ARM) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and [here is what the port script looks like](https://github.com/microsoft/vcpkg/blob/master/ports/capnproto/portfile.cmake). We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -33,3 +33,15 @@ If you have permission, after you've pushed your changes back to github, you can
     ./push-site.sh
 
 Otherwise, send a pull request and let someone else actually push the new site.
+
+# Package Managers
+
+You can download and install capnproto using the [vcpkg](https://github.com/Microsoft/vcpkg) dependency manager:
+
+    git clone https://github.com/Microsoft/vcpkg.git
+    cd vcpkg
+    ./bootstrap-vcpkg.sh
+    ./vcpkg integrate install
+    ./vcpkg install capnproto
+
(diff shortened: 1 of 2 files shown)


JackBoosY:
@harrishancock I've already moved it to https://github.com/capnproto/capnproto/blob/master/doc/install.md , can you review it again?

Thanks.

kentonv:
Hmm... I think it makes sense to add this somewhere, but it feels not quite right to add a top-level section highlighting vcpkg specifically, when other package managers (apt, homebrew) are only mentioned in a subsection of the Unix section.

Maybe we could add a subsection to the existing "Windows" section, as well as a new bullet point in the "Unix > From Package Managers" section?

I also hesitate to provide complete instructions for downloading and installing the package manager itself as part of our web site, since those instructions could change. Better to refer to the package manager's own docs, and only document Cap'n Proto's package name.

JackBoosY:
@kentonv Since vcpkg is a package manager that supports multiple platforms(including Windows, Linux and OSX), I don't think it is a good choice to put it on a certain platform.

kentonv:
@JackBoosY Since the page is organized into separate instructions for Unix vs. Windows, we will need to mention vcpkg in both places. Sorry but I don't think it's worth reorganizing the whole page for this.

JackBoosY:
@kentonv Okay, thanks!

PR #925, opened 2020-01-18
Add test for FdObserver(..., flags=0)::whenWriteDisconnected() support

### LINKED ISSUE #924: Use blocking STDOUT_FILENO with KJ event loop
This is a follow up to https://github.com/capnproto/capnproto/issues/919.

I don't want to make STDOUT_FILENO non-blocking, because it can have side-effect to stdin/stderr, e.g. when they all point to the same tty.
I'm fine with writing to stdout being blocking, so I made an `AsyncOutputStream` where the `write(...)` is just a call to `kj::FdOutputStream(STDOUT_FILENO).write(...)`.

However, there is one thing I'd like to watch when stdout is pollable, it's hangups / `whenWriteDisconnected()`.
So that my application can be stopped when the output stream read-end closes, e.g.:

    ./myapp | grep --max-count=1 SomeEvent

To support this, I used `kj::UnixEventPort::FdObserver::OBSERVE_WRITE`, but truth be told, it wasn't for waiting for writes, just to get an `FdObserver` that signals `whenWriteDisconnected()`.
And I realize now, that this had an unfortunate side-effect.

When running my application under strace, I now get an infinite loop, because strace writes to STDERR, which is shared with STDOUT:

    $ strace ./myapp
    ...
    [pid 15119] epoll_wait(3, [{EPOLLOUT, {u32=23202328, u64=23202328}}], 16, -1) = 1
    [pid 15119] epoll_wait(3, [{EPOLLOUT, {u32=23202328, u64=23202328}}], 16, -1) = 1
    [pid 15119] epoll_wait(3, [{EPOLLOUT, {u32=23202328, u64=23202328}}], 16, -1) = 1

My understanding is that each write to stderr strace make, generates an `epoll_wait()` event (not sure why, but why not?).
Each event generates an strace writes.
And this loop goes on indefinitely.

I found a way to watch only `whenWriteDisconnected()`, by instantiating an `FdObserver` that observes for nothing, value `0`: that is neither `OBSERVE_WRITE`, nor `OBSERVER_READ`, nor `OBSERVER_URGENT`.
This still watches hang-ups, since the epoll man page states that:

       EPOLLHUP
          Hang up happened on the associated file descriptor.
          epoll_wait(2) will always wait for this event; it is not necessary to set it in events.

With the current FdObserver code, this work fine:

https://github.com/capnproto/capnproto/blob/b301982c66f18d8f299537375f4293c5717f2e93/c++/src/kj/async-unix.c++#L341-L350

But I'm wondering if I'm relying on an implementation detail here, or if you think it is safe?
Also asking, in case there is a better method I did not think of.

Sarcasm:
Fixes #924

### DIFF
--- a/c++/src/kj/async-unix-test.c++
+++ b/c++/src/kj/async-unix-test.c++
@@ -856,6 +856,33 @@ KJ_TEST("UnixEventPort whenWriteDisconnected()") {
   hupPromise.wait(waitScope);
 }
 
+KJ_TEST("UnixEventPort FdObserver(..., flags=0)::whenWriteDisconnected()") {
+  // Verifies that given `0' as a `flags' argument,
+  // FdObserver still observes whenWriteDisconnected().
+  //
+  // This can be useful to watch disconnection on a blocking file descriptor.
+  // See discussion: https://github.com/capnproto/capnproto/issues/924
+
+  captureSignals();
+  UnixEventPort port;
+  EventLoop loop(port);
+  WaitScope waitScope(loop);
+
+  int pipefds[2];
+  KJ_SYSCALL(pipe(pipefds));
+  kj::AutoCloseFd infd(pipefds[0]), outfd(pipefds[1]);
+
+  UnixEventPort::FdObserver observer(port, outfd, 0);
+
+  auto hupPromise = observer.whenWriteDisconnected();
+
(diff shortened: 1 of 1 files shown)


kentonv:
Thanks!

PR #928, opened 2020-01-28
Add a 9-way variant for OneOf matching

bretthoerner:
I needed a 9 way switch... :blush:

---

Should I be doing something else at this point? Should we go ahead and add even more variants?


bretthoerner:
The build failures seem unrelated to me?

jclee:
Yeah, probably.  Looks like the message is "Build execution time has reached the maximum allowed time for your plan (60 minutes).".  I guess it's theoretically possible that making the compiler consider more templates could make compilation take longer.  But it seems more likely it would be due to a hiccup in the CI system, or maybe a compiler version upgrade.  I'm not sure if this is something that has come up before, or how to kick off new builds/diagnose.

(The change looks fine to me, FWIW.  I guess if it's important that a certain number of params works, we could add a test for it to ensure it continues to build correctly on all platforms.  But there don't seem to be exhaustive tests for the preexisting templates, so maybe not a big deal?)

PR #930, opened 2020-02-03
Fix method name and dereferencing in parseFromDirectory() comments

vasild:
In the documentation of parseFromDirectory():

* parseFromDirectory() is equivalent to parseFile(...), not
  parseFromDirectory(...).

* Use `.` to dereference instead of `->`. `parser` is not a pointer.


kentonv:
Thanks!

PR #931, opened 2020-02-04
Fix some typos

jclee:
Note that one of the commits fixes typos in public symbols.  The correcting commit fixes them directly, but this will break existing consumers.  I'm not sure if there's a deprecation mechanism that should be used instead or documentation where the change needs to be called out.


kentonv:
The only API typos are in the HTTP settings struct? I guess I'm OK with just changing it. In all likelihood, no one is using them except us.

jclee:
Heh... I don't know if they are the only API typos -- just the only ones I've run across so far while working in kj code. :)

PR #935, opened 2020-02-11
Fix Table::findOrCreate() to be exception-safe.

kentonv:
If the creator callback throws, we shouldn't leave the table broken.


kentonv:
MinGW and Cygwin failures unrelated.

PR #936, opened 2020-02-12
Allow OneOf to be initialized from a subset OneOf.

kentonv:
E.g. it should be possible to initialize `OneOf<T, U, V>` from `OneOf<T, V>`.

This change also makes the error messages nicer when someone tries to access a OneOf member type that it doesn't have. This will now fire a static_assert with an error message rather than produce an obscure error about `TypeIndex_` being incomplete.

### DIFF
--- a/c++/src/kj/one-of.h
+++ b/c++/src/kj/one-of.h
@@ -29,10 +29,32 @@ namespace kj {
 
 namespace _ {  // private
 
-template <uint i, typename Key, typename First, typename... Rest>
-struct TypeIndex_ { static constexpr uint value = TypeIndex_<i + 1, Key, Rest...>::value; };
-template <uint i, typename Key, typename... Rest>
-struct TypeIndex_<i, Key, Key, Rest...> { static constexpr uint value = i; };
+template <uint i, template<uint> class Fail, typename Key, typename... Variants>
+struct TypeIndex_;
+template <uint i, template<uint> class Fail, typename Key, typename First, typename... Rest>
+struct TypeIndex_<i, Fail, Key, First, Rest...> {
+  static constexpr uint value = TypeIndex_<i + 1, Fail, Key, Rest...>::value;
+};
+template <uint i, template<uint> class Fail, typename Key, typename... Rest>
+struct TypeIndex_<i, Fail, Key, Key, Rest...> { static constexpr uint value = i; };
+template <uint i, template<uint> class Fail, typename Key>
+struct TypeIndex_<i, Fail, Key>: public Fail<i> {};
+
+template <uint i>
+struct OneOfFailError_ {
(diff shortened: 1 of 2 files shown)


kentonv:
The first version of this failed on everything _except_ MSVC. Whoa.

(I think it's because I was inadvertently relying on C++17 rule changes.)

c++/src/kj/one-of.h
```diff
@@ -29,10 +29,32 @@ namespace kj {
...
+  static constexpr uint value = TypeIndex_<i + 1, Fail, Key, Rest...>::value;
+};
+template <uint i, template<uint> class Fail, typename Key, typename... Rest>
+struct TypeIndex_<i, Fail, Key, Key, Rest...> { static constexpr uint value = i; };
+template <uint i, template<uint> class Fail, typename Key>
+struct TypeIndex_<i, Fail, Key>: public Fail<i> {};
+
+template <uint i>
+struct OneOfFailError_ {
+  static_assert(i == -1, "type does not match any in OneOf");
```
harrishancock: Nice trick. I guess `static_assert(false, ...)` would trigger an "eager" compile failure?

c++/src/kj/one-of.h
```diff
@@ -67,18 +89,43 @@ using Variants = typename Variants_<i>::Type;
...
   OneOf(OneOf& other) { copyFrom(other); }
   OneOf(OneOf&& other) { moveFrom(other); }
-  template <typename T>
+  // Copy/move from same OneOf type.
+
+  template <typename... OtherVariants, typename = typename HasAll<1, OtherVariants...>::Success>
+  OneOf(const OneOf<OtherVariants...>& other) { copyFromSubset(other); }
+  template <typename... OtherVariants, typename = typename HasAll<1, OtherVariants...>::Success>
+  OneOf(OneOf<OtherVariants...>& other) { copyFromSubset(other); }
+  template <typename... OtherVariants, typename = typename HasAll<1, OtherVariants...>::Success>
```
harrishancock: I'm curious if there's a significance to the `1` constant, versus the `0` on line 123? Since empty OneOfs (i.e. `OneOf<>`) aren't allowed, it seems like we'll always instantiate the `HasAll<i, First, Rest...>` specialization, rather than the `HasAll<i>` specialization, thus ignoring both the `0` and `1`.

c++/src/kj/one-of.h
kentonv: Yeah, I hate that about `static_assert`.

c++/src/kj/one-of.h
kentonv: You thought about it harder than I did. I was just shoving garbage in there.

PR #938, opened 2020-02-24
HttpClient adapter (around HttpService) should propagate exceptions.

kentonv:
A while ago, I adjusted the HttpClient adapter implemnetation so that the response stream on the client side would delay reading EOF until the HttpService's request() promise had completed. This ensured that the promise didn't get prematurely canceled.

But, originally I only had it delay completion, without propagating any exceptions. It turns out to be useful to propagate exceptions here. Although exceptions would not be propagated over a true network connection, propagating them over in-process adapters seems to make errors easier to debug in practice. This also means they will propagate over http-over-capnp.

I also added tests for this functionality, which were previously missing.

### DIFF
--- a/c++/src/kj/compat/http-test.c++
+++ b/c++/src/kj/compat/http-test.c++
@@ -3033,7 +3033,7 @@ KJ_TEST("newHttpClient from HttpService WebSockets") {
 }
 
 KJ_TEST("adapted client/server propagates request exceptions like non-adapted client") {
-  auto io = kj::setupAsyncIo();
+  KJ_HTTP_TEST_SETUP_IO;
 
   HttpHeaderTable table;
   HttpHeaders headers(table);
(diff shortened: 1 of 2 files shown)


c++/src/kj/compat/http-test.c++
```diff
@@ -3065,6 +3065,193 @@ KJ_TEST("adapted client/server propagates request exceptions like non-adapted cl
...
+    KJ_ASSERT(msg.is<kj::String>());
+    KJ_ASSERT(msg.get<kj::String>() == "foo");
+  }
+
+  kj::Promise<void> promise = nullptr;
+  if (closeUpstreamFirst) {
+    // Receiving the close hangs.
+    promise = ws->receive()
+        .then([](WebSocket::Message&& msg) { KJ_EXPECT(msg.is<WebSocket::Close>()); });
+    KJ_EXPECT(!promise.poll(waitScope));
```
harrishancock: This line should be outside of the if-else block to test the other branch correctly, I think.

kentonv: Whoa, nice catch... When I read test code my eyes just glaze over.

harrishancock: It's my favorite pastime. :)

PR #940, opened 2020-02-26
Fix HashMap::eraseAll() to match documentation.

kentonv:
@bretthoerner


bretthoerner:
Thanks, cherry-picked and it's working for me.

PR #941, opened 2020-02-26
rpc-twoparty-test: Keep pipe streams alive until disconnect

jclee:
Fixes invalid memory access error in valgrind, caused by the pipe being destroyed earlier than the event loop that still contains detached promises that reference it.


kentonv:
This seems like it's papering over a deeper problem in the RPC implementation. When the `RpcSystem` instance is destroyed, in theory it's supposed to disconnect any live objects from the underlying `VatNetwork` so that they won't try to use it anymore. So, since `TwoPartyClient` and `TwoPartyServer` are destroyed before the underlying pipes are dropped, there should be no use-after-free here. Let me take a look at how to fix this properly.

jclee:
Ah, I was afraid that might have been the case, but wasn't sure.

kentonv:
Closing in favor of #944.

PR #944, opened 2020-02-27
Cancel every RPC connection's incoming message loop when the RpcSystem is destroyed.

kentonv:
If the RpcSystem is destroyed, the underlying VatNetwork will probably be destroyed too, and reading messages from it may be a use-after-free.

This fixes valgrind errors in rpc-twoparty-test.

Closes #941

### DIFF
--- a/c++/src/capnp/rpc.c++
+++ b/c++/src/capnp/rpc.c++
@@ -400,6 +400,7 @@ public:
           return kj::READY_NOW;
         });
     disconnectFulfiller->fulfill(DisconnectInfo { kj::mv(shutdownPromise) });
+    canceler.cancel(networkException);
     connection.init<Disconnected>(kj::mv(networkException));
   }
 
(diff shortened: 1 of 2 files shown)


kentonv:
I guess what's actually implemented here is cancelling message loops on disconnect. Destroying the RpcSystem, though, disconnects all connections.

kentonv:
Right, `newOutgoingMessage()` doesn't itself start any I/O. Its `send()` method does, but that method does not return a promise -- the promise is queued in the underlying `VatNetwork`. And when the `VatNetwork` is destroyed, it will destroy the whole queue of writes. Also, the `RpcSystem` always checks if the connection is still up immediately before calling `send()`.

With all that said, with the additional 6.5 years of KJ experience I have now, I probably would have designed these things a bit differently...

kentonv:
It seems like cygwin has developed some new warning that is causing compile failures in old code that hasn't changed... ugh.

PR #945, opened 2020-02-28
Fix bug in byte stream path-shortening.

kentonv:
Apparently, I only initiated capability promise resolution in one of LocalClient's two constructors. Ironically, the other constructor is always the one used by ByteStreams created by ByteStreamFactory, because they are always placed in a CapabilityServerSet. The effect is that ByteStreams couldn't be path-shortened in cases that require promise capabilities to resolve before unwrapping can occur, such as over RPC.

This PR fixes the bug and adds a test covering the RPC use case.

### DIFF
--- a/c++/src/capnp/compat/byte-stream-test.c++
+++ b/c++/src/capnp/compat/byte-stream-test.c++
@@ -21,6 +21,7 @@
 
 #include "byte-stream.h"
 #include <kj/test.h>
+#include <capnp/rpc-twoparty.h>
 
 namespace capnp {
 namespace {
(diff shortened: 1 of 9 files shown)


c++/src/capnp/compat/byte-stream-test.c++
```diff
@@ -228,6 +229,77 @@ KJ_TEST("KJ -> ByteStream -> KJ pipe -> ByteStream -> KJ with shortening") {
...
+  auto rpcConnection = kj::newTwoWayPipe();
+  capnp::TwoPartyClient client(*rpcConnection.ends[0],
+      clientFactory.kjToCapnp(kj::mv(backPipe.out)),
+      rpc::twoparty::Side::CLIENT);
+  capnp::TwoPartyClient server(*rpcConnection.ends[1],
+      serverFactory.kjToCapnp(kj::mv(middlePipe.out)),
+      rpc::twoparty::Side::CLIENT);
+
+  auto backWrapped = serverFactory.capnpToKj(server.bootstrap().castAs<ByteStream>());
+  auto midPumpPormise = middlePipe.in->pumpTo(*backWrapped, 3);
```
jclee: "midPumpPormise"

kentonv:
Note: I'm currently debugging a deeper problem that this change uncovers (by making path-shortening more common). It looks like there's a case where CapabilityServerSet will unwrap a capability while calls are still in-flight (in fact, haven't been delivered yet). That _should_ be prevented by embargoes... more debugging to do on Monday.

kentonv:
This now fixes more bugs, but it isn't done. I'm still tracking down one obscure bug observed in Workers Runtime tests.

I also really should write Cap'n Proto unit tests for each of these bugs, rather than rely on Workers Runtime tests.

kentonv:
Wow WTF GitHub lists the commits ordered by timestamp rather than by actual commit history order. Ugh. Oh well, they're fairly independent.

kentonv:
Actually no, they aren't in date order. They're in... some random order.

c++/src/capnp/rpc.c++
```diff
@@ -998,7 +993,9 @@ private:
     }
 
     kj::Maybe<kj::Promise<kj::Own<ClientHook>>> whenMoreResolved() override {
-      return fork.addBranch();
+      return fork.addBranch().then([this](auto&&) {
+        return cap->addRef();
```
harrishancock: This seems to mean that if `resolve()` throws, then `whenMoreResolved()` will return the `initial` constructor parameter. The previous behavior was always to return the `eventual` constructor parameter. Is that intended, or am I reading the code wrong?

kentonv: `cap` is replaced during `resolve()`, so it's the same thing. But I think the reason I put this here was because I was thinking of making `resolve()` return void and then `fork` would be `ForkedPromise<void>`, but it looks like I didn't follow through on that... will fix.

harrishancock: I guess it wouldn't make sense to return `eventual` if `resolve()` threw -- I would intuitively expect a broken capability, though.

harrishancock: > cap is replaced during resolve(), so it's the same thing.

My point is that if `resolve()` throws, that's not necessarily true, since `cap` is replaced at the very end.

harrishancock: At any rate, LGTM otherwise!

kentonv: If resolve() throws then this `.then()` will be skipped...

harrishancock: By my reading, an exception in `resolve()` would be caught by the `.catch_()` on line 890, which would then pass `newBrokenCap()` to this `.then()`. I.e., it would not be skipped.

kentonv: Ah... yeah I don't think that's what I intended. Fixup: eb44b761

PR #947, opened 2020-03-03
Make FixedArray constexpr

harrishancock:


c++/src/kj/array.h
```diff
@@ -518,21 +518,21 @@ class FixedArray {
...
     return arrayPtr(content, fixedSize);
   }
-  inline operator ArrayPtr<const T>() const {
+  inline constexpr operator ArrayPtr<const T>() const {
     return arrayPtr(content, fixedSize);
   }
 
-  inline T& operator[](size_t index) { return content[index]; }
-  inline const T& operator[](size_t index) const { return content[index]; }
+  inline constexpr T& operator[](size_t index) { return content[index]; }
```
harrishancock: This might look strange (the corresponding accessor in kj::ArrayPtr seems intentionally non-constexpr), but is required to be able to effectively use FixedArray in a constexpr function.

jclee:
Looks good to me, although gcc looks unhappy.

harrishancock:
gcc on Windows failed with
```
In file included from ./src/kj/memory.h:24,
                 from ./src/kj/array.h:24,
                 from ./src/kj/string.h:25,
                 from ./src/capnp/common.h:29,
                 from ./src/capnp/raw-schema.h:24,
                 from ./src/capnp/generated-header-support.h:26,
                 from ./src/capnp/compiler/lexer.capnp.h:6,
                 from src/capnp/compiler/lexer.h:24,
                 from src/capnp/compiler/capnp.c++:26:
./src/kj/common.h: In instantiation of ‘constexpr kj::ArrayPtr<T>::ArrayPtr(std::initializer_list<typename kj::RemoveConstOrDisable_<T>::Type>) [with T = const kj::ArrayPtr<const unsigned char>; typename kj::RemoveConstOrDisable_<T>::Type = kj::ArrayPtr<const unsigned char>]’:
src/capnp/compiler/capnp.c++:1111:69:   required from here
./src/kj/common.h:1391:45: error: initializing ‘kj::ArrayPtr<const kj::ArrayPtr<const unsigned char> >::ptr’ from ‘std::initializer_list<kj::ArrayPtr<const unsigned char> >::begin’ does not extend the lifetime of the underlying array [-Werror=init-list-lifetime]
 1391 |       : ptr(init.begin()), size_(init.size()) {}
      |                                             ^
cc1plus: all warnings being treated as errors
```
But I don't see how this has anything to do with the `FixedArray` change. :/

harrishancock:
Looks like this isn't the only PR with the gcc failure: https://github.com/capnproto/capnproto/pull/945

kentonv:
It's a new warning in GCC10, which isn't released yet but Cygwin is apparently already shipping.

kentonv:
On the bright side it means Cygwin fails quickly instead of being killed after hanging for an hour.

jclee:
Ah, true.  I saw "constexpr ArrayPtr" in the error and assumed.  Looks like we're missing a number of older CI data on PRs, but although the cygwin build has been marked as an error for a while, the older ones seem to generally contain success messages...  not sure if they're getting cut off, or just failing due to taking too long.  The last one lacking the constexpr error is 936.  The first one with the constexpr error after that is 941, with several missing builds in between.

Re: gcc10: Ah.  Good to know.  Looks like 941 contains an installation message for gcc 9.2.0-2, which presumably carries the new warning.

PR #951, opened 2020-03-09
Do not override CMake variables from target

### LINKED ISSUE #950: Compile fails in cross-compiling scenarios
If you include the capnproto library into the build e.g. via CMake's `  FetchContent_MakeAvailable(CapnProto)`, building with cross toolchain fails due to the code generation tools not found correctly.

The reason for this is that `find_package` correctly finds the built executable, possibly for another architecture, or uses the information from the `capnp_tool` target. If you want to override this, there is no possibility at all, since the following code in `CapnProtoMacros.cmake` prevents this:
```cmake
31    if(TARGET capnp_tool)
32      set(CAPNP_EXECUTABLE capnp_tool)
33      GET_TARGET_PROPERTY(CAPNPC_CXX_EXECUTABLE capnpc_cpp CAPNPC_CXX_EXECUTABLE)
34      GET_TARGET_PROPERTY(CAPNP_INCLUDE_DIRECTORY capnp_tool CAPNP_INCLUDE_DIRECTORY)
35      list(APPEND tool_depends capnp_tool capnpc_cpp)
36    endif()
```
* The target capnp_tool is checked, if true, the CAPNP_EXECUTABLE is set to `capnp_tool`, which is not existent anymore (should use `set(CAPNP_EXECUTABLE $<TARGET_FILE:CapnProto::capnp_tool>` instead)
* Overriding the variables should be prevented: A check if the variables are initial should be added to enable the caller to override it with values appropriate for his build system:
  ```
  if(NOT CAPNP_EXECUTABLE)
    set(CAPNP_EXECUTABLE $<TARGET_FILE:CapnProto::capnp_tool>
  endif()
   ```

svenihoney:
In case of externally set variables for CMAKE_CXX_EXECUTABLE etc., they
are overridden again if the capnp_tool target has been found. This
change prevents this by checking if the variables are already set and
sets them from target only if they are empty.

Fixes #950.

Acked-by: Sven Fischer <sven@leiderfischer.de>


kentonv:
Thanks!

PR #969, opened 2020-04-11
Double check endTime in Mutex::wait() on Windows

harrishancock:
This fixes a failure I observed on my Windows VM at mutex-test.c++:281, in which value.when() was timing out up to 6ms too early.

I preserved the existing behavior in the event endTime is null and the error is ERROR_TIMEOUT. I'm not sure it's correct, though.

### DIFF
--- a/c++/src/kj/mutex.c++
+++ b/c++/src/kj/mutex.c++
@@ -547,8 +547,8 @@ void Mutex::wait(Predicate& predicate, Maybe<Duration> timeout) {
     } else {
       DWORD error = GetLastError();
       if (error == ERROR_TIMEOUT) {
-        // Timed out. Skip predicate check.
-        return;
+        // Windows may have woken us up too early, so don't return yet. Instead, proceed through the
+        // loop and rely on our sleep time recalculation to detect if we timed out.
       } else {
         KJ_FAIL_WIN32("SleepConditionVariableSRW()", error);
       }


harrishancock:
Amusingly, I did this before seeing your adventures in #968.

c++/src/kj/mutex.c++
```diff
@@ -547,8 +547,16 @@ void Mutex::wait(Predicate& predicate, Maybe<Duration> timeout) {
...
-        // Timed out. Skip predicate check.
-        return;
+        // As mentioned above, Windows sleeps are inaccurate, so we must double check.
+        KJ_IF_MAYBE(e, endTime) {
+          if (*e <= clock->now()) {
+            // Actually timed out. Skip predicate check.
+            return;
+          }
+        } else {
+          // Unexpected ERROR_TIMEOUT. Should we KJ_FAIL_WIN32() instead?
```
kentonv, line 557: Maybe the `KJ_IF_MAYBE(e, endTime)` can be `KJ_ASSERT_NONNULL(endTime)`?

c++/src/kj/mutex.c++
```diff
@@ -547,8 +547,16 @@ void Mutex::wait(Predicate& predicate, Maybe<Duration> timeout) {
...
       DWORD error = GetLastError();
       if (error == ERROR_TIMEOUT) {
-        // Timed out. Skip predicate check.
-        return;
+        // As mentioned above, Windows sleeps are inaccurate, so we must double check.
+        KJ_IF_MAYBE(e, endTime) {
+          if (*e <= clock->now()) {
+            // Actually timed out. Skip predicate check.
+            return;
+          }
```
kentonv, line 555: For the case where we woke up too early, maybe comment to explain that we expect the predicate is still false, but we're going to proceed around the loop anyway just for the sake of setting up a new sleep.

c++/src/kj/mutex.c++
harrishancock: Done.

c++/src/kj/mutex.c++
harrishancock: And done.

c++/src/kj/mutex.c++
```diff
@@ -547,8 +547,15 @@ void Mutex::wait(Predicate& predicate, Maybe<Duration> timeout) {
     } else {
       DWORD error = GetLastError();
       if (error == ERROR_TIMEOUT) {
-        // Timed out. Skip predicate check.
-        return;
+        // As mentioned above, Windows sleeps are inaccurate, so we must double check endTime.
+        auto e = KJ_ASSERT_NONNULL(endTime, "unexpected timeout from SleepConditionVariableSRW()");
+        if (e <= clock->now()) {
```
kentonv, line 552: It occurs to me that we're reading `clock->now()` here and then immediately reading it again below if we continue around the loop. Is that worth avoiding?

harrishancock: Maybe if there's a reasonably readable way. I suppose we could remove the check entirely and do something like:

```cpp
if (error == ERROR_TIMEOUT) {
  // We may have been woken up early, so don't return yet. Instead, proceed through
  // the loop and rely on our endTime recalculation to detect if we timed out.
} else {
  KJ_FAIL_WIN32(...);
}
```

It'd be a minor behavior change (we'd check for exceptions where before we didn't), but that seems fine.

harrishancock: Amended as such, open to suggestions though.

kentonv: Oh! Yes! That's fine and much better than what I was thinking.

PR #971, opened 2020-04-14
Skip 'HttpClient concurrency limiting' test on Mac and Windows, where it is flaky

kentonv:
NOTE: Depends on #968. Only the last three commits are new.

I got a little carried away with this and ended up extending userland pipes to support capability-passing so that I could use them in the HTTP connection management tests, meaning we _can_ still run a non-flaky version of the tests on Mac and Windows, we just can't run the version that depends on kernel event delivery being immediate.


kentonv:
what is this I don't even
Screenshot from 2020-04-14 15-02-11

jclee:
Heh... annoying, yeah, but I could see it guarding against legit bugs in other contexts, if someone used a different brace style, like:
```
KJ_IF_MAYBE(e, exception);
{
  abort();
}
```
Kind of annoying that they couldn't or wouldn't notice the separating blank line, though.

PR #972, opened 2020-04-14
WIP enable extended release tests when requested

kentonv:

PR #975, opened 2020-04-18
Fix bugs and add release tests as GitHub Action

kentonv:
The release tests won't run on this PR, but to see the actual results, check out #972, which I think I'm just going to leave up for posterity.


c++/src/kj/async-io.c++
```diff
@@ -917,7 +917,7 @@ private:
     Promise<void> writeWithFds(ArrayPtr<const byte> data,
                                ArrayPtr<const ArrayPtr<const byte>> moreData,
                                ArrayPtr<const int> fds) override {
-#if __GNUC__ && !__clang__ && __GNUC__ >= 8
+#if __GNUC__ && !__clang__ && __GNUC__ >= 7
 // GCC 8 decides the open-brace below is "misleadingly indented" as if it were guarded by the `for`
```
harrishancock, line 921: Could update the comment, too, up to you.

kentonv: Fixed and merged.

Issue #313, opened 2016-04-13
Reading an unaligned message triggers a KJ_DASSERT() failure, but only sometimes.


dwrensha:
Currently, if you try to read an unaligned message (on an architecture that allows such reads), you can hit [this error](https://github.com/sandstorm-io/capnproto/blob/f6a31613efce413795a200ccc194c2c05d463da2/c%2B%2B/src/capnp/layout.c%2B%2B#L2731-L2733) if your code was compiled in debug mode. The error is not triggered on all messages, so tracing down its cause can be confusing. See for example the mailing list discussion [here](https://groups.google.com/forum/#!topic/capnproto/RPqMlhkSdeo).

It would be better if capnproto had a more uniform policy on alignment errors. We should either:
1. Always throw a `KJ_DASSERT()` failure on unaligned reads, with a more helpful error message explaining the need for alignment. 
2. Never throw an exception due to an unaligned read. Justification: the programmer, by explicitly casting the data to `capnp::word*`, has taken responsibility for any dangers that lurk.

kentonv:
Problem with (2) is that we've had a few conversations with people where they seemed unaware that there was any problem with such a cast.

Maybe what we want is a very upfront check (e.g. in FlatArrayMessageReader's constructor), but with the ability to call a special version of the method that acknowledges that the caller doesn't care about alignment (for people who are on x86 and would rather have unaligned reads than do a copy).

DemiMarie:
@kentonv Unfortunately at least LLVM considers unaligned reads to be undefined behavior.  Though I wonder if there is any realistic way this could cause a problem.

PR #977, opened 2020-04-20
Remove `UnalignedFlatArrayMessageReader`.

kentonv:
This reverts commit 3aa2b2aa02edb1c160b154ad74c08c929a02512a.

This is not safe, even on x86. Unaligned access is undefined behavior in C/C++. Compilers can and do optimize under the assumption that access is aligned. For example, GCC may use SIMD instructions in optimizations, and these do in fact require alignment. In principle, a compiler could also choose to mask away the lowest bits of a pointer that should always be zero when aligned.

I considered deprecating this class instead of deleting it, but a Google search came up with no mentions of this class anywhere except our own code and issues, suggesting it is not used in any open source code. If it's used in any closed source codebases, then the code's own developers will be the first to see the build breakage and will be well-positioned to fix it. I decided it's better to force them to fix it now than to let them run into inscrutable bugs from unexpected compiler optimizations later.

I considered adding more explicit checks for alignment within the implementation but figured that would have a higher risk of breaking some open source code out there whose users may not easily be able to fix it...

### DIFF
--- a/c++/src/capnp/serialize.h
+++ b/c++/src/capnp/serialize.h
@@ -47,30 +47,7 @@ CAPNP_BEGIN_HEADER
 
 namespace capnp {
 
-class UnalignedFlatArrayMessageReader: public MessageReader {
-  // Like FlatArrayMessageReader, but skips checking that the array is properly-aligned.
-  //
-  // WARNING: This only works on architectures that support unaligned reads, like x86/x64 and
-  //   modern ARM. Unaligned access may incur a performance penalty on these platforms. On many
-  //   other platforms, the program will simply crash on unaligned reads. Also note that unaligned
-  //   data access may be considered undefined behavior by compilers; use at your own risk. If at
-  //   all possible, try to ensure your data ends up in aligned buffers rather than rely on this
-  //   class.
-
-public:
-  UnalignedFlatArrayMessageReader(
-      kj::ArrayPtr<const word> array, ReaderOptions options = ReaderOptions());
-  kj::ArrayPtr<const word> getSegment(uint id) override;
-  const word* getEnd() const { return end; }
-
-private:
-  // Optimize for single-segment case.
-  kj::ArrayPtr<const word> segment0;
-  kj::Array<kj::ArrayPtr<const word>> moreSegments;
-  const word* end;
-};
-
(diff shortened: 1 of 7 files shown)


c++/src/capnp/serialize.h
```diff
@@ -231,14 +213,6 @@ void writeMessageToFd(int fd, kj::ArrayPtr<const kj::ArrayPtr<const word>> segme
 // =======================================================================================
 // inline stuff
 
-inline FlatArrayMessageReader::FlatArrayMessageReader(
-    kj::ArrayPtr<const word> array, ReaderOptions options)
-#ifdef KJ_DEBUG
-    : UnalignedFlatArrayMessageReader(checkAlignment(array), options) {}
```
dwrensha, line 237: Can we keep this alignment check in the constructor of `FlatArrayMessageReader`? Otherwise, client code that's sloppy about alignment will not get any indication of its error unless it also happens to call `ListPointer::getStructElement()` and hits the`KJ_DASSERT` there. (I'm assuming that `KJ_DEBUG` is enabled exactly when `KJ_DASSERT` is enabled.)

I still agree with what I wrote in #313. If unaligned buffers are not okay, then we should make the error message for them as clear and as reliable as possible.

kentonv:
OK, I added a commit to enforce alignment strictly, throwing an exception if any segment is unaligned.

kentonv:
Added small fixup needed to fix, specifically, clang with `-fno-exceptions`.

```diff
diff --git a/c++/src/capnp/message-test.c++ b/c++/src/capnp/message-test.c++
index e70a7c7b..4d7362e3 100644
--- a/c++/src/capnp/message-test.c++
+++ b/c++/src/capnp/message-test.c++
@@ -186,7 +186,8 @@ KJ_TEST("disallow unaligned") {
 
   auto unaligned = kj::arrayPtr(reinterpret_cast<word*>(buffer + 1), 1);
 
-  SegmentArrayMessageReader message({unaligned});
+  kj::ArrayPtr<const word> segments[1] = {unaligned};
+  SegmentArrayMessageReader message(segments);
   KJ_EXPECT_THROW_RECOVERABLE_MESSAGE("unaligned", message.getRoot<TestAllTypes>());
 }
 #endif
```

PR #979, opened 2020-04-21
Fixing CI badges to use new GitHub Actions.

haata:
Since you aren't using Travis-CI and Appveyor anymore, might as well show GitHub Action badges.

One limitation of the GitHub Action badges is that you can only show status at a workflow level (each yaml file).


kentonv:
Hmm, what branch's state to these badges reflect? `master`? But the release tests only run on branches matching `release-*`. Do the badges just show the status of the last build? But what if it's a branch that hasn't been merged?

haata:
There's a bit more information here: https://help.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow#adding-a-workflow-status-badge-to-your-repository

I just used the default badge which uses the last event I believe.

Another way to do it would be to attach it to a branch
```
https://github.com/capnproto/capnproto/workflows/Quick%20Tests/badge.svg?branch=master
```


There is also a pretty large list of event types (?event=) if you go https://github.com/capnproto/capnproto/actions?query=workflow%3A%22Release+Tests%22 and click on the "Create status Badge" drop-down.

kentonv:
Hmm I guess currently none of our tests actually run at master... they either run on a PR (pre-merge) or on a release branch.

How about this:
* We update both workflows so that they also run on "push" to the master branch.
* We have the badges report the status of the master branch.

kentonv:
I went ahead and made #982 based on this.

PR #982, opened 2020-04-22
Make actions run at master and power CI badges

kentonv:
Based on @haata's #979, but I changed the badges to show specifically the master branch, and made the actions actually run at master -- including the release tests. It'll be nice to find out when the release tests break without having to explicitly ask for them, but also without having to run them on every PR commit.


kentonv:
Gonna merge without review since this is a pretty trivial, non-code change.

PR #987, opened 2020-04-26
expand comments for Brand.Scope.inherit

dwrensha:
The first time I learned about Cap'n Proto generics, and again this weekend while trying to relearn them, I spent a long amount of time trying to understand the meaning of `Brand.Scope.inherit`. I found the existing doc comment to be unhelpful: what exactly is a "reference point" and what does it mean to inherit from one?

What finally made things click was to think of `Brand`s as type substitutions, and to think of `inherit` as an identity substitution -- it just replaces the parameters with themselves.

This pull request expands the documentation and adds an example that should make things more clear.

### DIFF
--- a/c++/src/capnp/schema.capnp
+++ b/c++/src/capnp/schema.capnp
@@ -407,8 +407,8 @@ struct Brand {
       #     struct Inner {
       #       value @0 :T;
       #     }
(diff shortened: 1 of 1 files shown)


kentonv:
Thanks!

PR #991, opened 2020-04-28
Add a seperate CanceledException type for fiber cancelation.

ObsidianMinor:
This may not be suitable since it requires exceptions to be enabled to use fibers as the new exception type isn't a kj exception.

### DIFF
--- a/c++/src/kj/async-inl.h
+++ b/c++/src/kj/async-inl.h
@@ -1093,13 +1093,17 @@ inline PromiseForResult<Func, void> evalNow(Func&& func) {
 
 template <typename Func>
 inline PromiseForResult<Func, WaitScope&> startFiber(size_t stackSize, Func&& func) {
-  typedef _::FixVoid<_::ReturnType<Func, WaitScope&>> ResultT;
+  #if KJ_NO_EXCEPTIONS
+    KJ_UNIMPLEMENTED("Fibers require exceptions");
+  #else
+    typedef _::FixVoid<_::ReturnType<Func, WaitScope&>> ResultT;
 
-  Own<_::FiberBase> intermediate = kj::heap<_::Fiber<Func>>(stackSize, kj::fwd<Func>(func));
-  intermediate->start();
-  auto result = _::PromiseNode::to<_::ChainPromises<_::ReturnType<Func, WaitScope&>>>(
-      _::maybeChain(kj::mv(intermediate), implicitCast<ResultT*>(nullptr)));
-  return _::maybeReduce(kj::mv(result), false);
+    Own<_::FiberBase> intermediate = kj::heap<_::Fiber<Func>>(stackSize, kj::fwd<Func>(func));
+    intermediate->start();
+    auto result = _::PromiseNode::to<_::ChainPromises<_::ReturnType<Func, WaitScope&>>>(
+        _::maybeChain(kj::mv(intermediate), implicitCast<ResultT*>(nullptr)));
+    return _::maybeReduce(kj::mv(result), false);
+  #endif
 }
 
 template <typename T>
--- a/c++/src/kj/async-inl.h
+++ b/c++/src/kj/async-inl.h
@@ -1093,17 +1093,13 @@ inline PromiseForResult<Func, void> evalNow(Func&& func) {
 
 template <typename Func>
 inline PromiseForResult<Func, WaitScope&> startFiber(size_t stackSize, Func&& func) {
-  #if KJ_NO_EXCEPTIONS
-    KJ_UNIMPLEMENTED("Fibers require exceptions");
-  #else
-    typedef _::FixVoid<_::ReturnType<Func, WaitScope&>> ResultT;
+  typedef _::FixVoid<_::ReturnType<Func, WaitScope&>> ResultT;
 
-    Own<_::FiberBase> intermediate = kj::heap<_::Fiber<Func>>(stackSize, kj::fwd<Func>(func));
-    intermediate->start();
-    auto result = _::PromiseNode::to<_::ChainPromises<_::ReturnType<Func, WaitScope&>>>(
-        _::maybeChain(kj::mv(intermediate), implicitCast<ResultT*>(nullptr)));
-    return _::maybeReduce(kj::mv(result), false);
-  #endif
+  Own<_::FiberBase> intermediate = kj::heap<_::Fiber<Func>>(stackSize, kj::fwd<Func>(func));
+  intermediate->start();
+  auto result = _::PromiseNode::to<_::ChainPromises<_::ReturnType<Func, WaitScope&>>>(
+      _::maybeChain(kj::mv(intermediate), implicitCast<ResultT*>(nullptr)));
+  return _::maybeReduce(kj::mv(result), false);
 }
 
 template <typename T>
--- a/c++/src/kj/async-test.c++
+++ b/c++/src/kj/async-test.c++
@@ -854,7 +854,7 @@ KJ_TEST("exclusiveJoin both events complete simultaneously") {
   KJ_EXPECT(!joined.poll(waitScope));
 }
 
-#if !__BIONIC__
+#if !__BIONIC__ && !KJ_NO_EXCEPTIONS
 KJ_TEST("start a fiber") {
   EventLoop loop;
   WaitScope waitScope(loop);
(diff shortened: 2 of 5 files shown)


c++/src/kj/async-test.c++
```diff
@@ -925,33 +923,31 @@ KJ_TEST("cancel a fiber") {
...
       KJ_DEFER(exited = true);
-#if KJ_NO_EXCEPTIONS
-      promise.wait(fiberScope);
-#else
-      int i = promise.wait(fiberScope);
-      KJ_EXPECT(i == 123);
-#endif
+      try {
+        promise.wait(fiberScope);
+      } catch (kj::CanceledException) {
```
kentonv, line 938: This shouldn't be caught in the test code, it should be caught by the fiber infrastructure. In fact, I think the fiber infrastructure should require that if a fiber was canceled, the fiber exits by throwing CanceledException, not by returning normally.

kentonv:
Yeah, I think I'm fine with not supporting fibers when KJ_NO_EXCEPTIONS is defined. In fact it occurs to me that the entire `ExceptionCallback` mechanism (used especially when exceptions are disabled) is currently incompatible with fibers since it relies on thread-locals. So there's a bunch of work that would need to be done to make this work reasonably and I don't really care that much about supporting exotic features for people who disable exceptions.

c++/src/kj/async-inl.h
```diff
@@ -1093,13 +1093,17 @@ inline PromiseForResult<Func, void> evalNow(Func&& func) {
...
 template <typename Func>
 inline PromiseForResult<Func, WaitScope&> startFiber(size_t stackSize, Func&& func) {
-  typedef _::FixVoid<_::ReturnType<Func, WaitScope&>> ResultT;
-
-  Own<_::FiberBase> intermediate = kj::heap<_::Fiber<Func>>(stackSize, kj::fwd<Func>(func));
-  intermediate->start();
-  auto result = _::PromiseNode::to<_::ChainPromises<_::ReturnType<Func, WaitScope&>>>(
-      _::maybeChain(kj::mv(intermediate), implicitCast<ResultT*>(nullptr)));
-  return _::maybeReduce(kj::mv(result), false);
+  #if KJ_NO_EXCEPTIONS
```
kentonv, line 1096: Maybe this check could be in `FiberBase::FiberBase`, along-side the `__BIONIC__` check there that also calls `KJ_UNIMPLEMENTED`?

You'll also need to `#ifdef` out any code that actually uses `try`, `catch`, or `throw` when `KJ_NO_EXCEPTIONS` is on.

c++/src/kj/exception.h
```diff
@@ -132,8 +132,15 @@ class Exception {
   friend class ExceptionImpl;
 };
 
+struct CanceledException { };
+  // This exception is thrown to force-unwind a stack in order to immediately cancel whatever that
+  // stack was doing. It is used in the implementation of fibers in praticular. Application code
```
jclee, line 137: s/praticular/particular

c++/src/kj/async-inl.h
ObsidianMinor: I `#ifdef`ed out the fiber handling code in `waitImpl` which should cover all of it.

kentonv: You added a `throw` statement to `fiberCanceledException()`. That'll produce a compilation failure if exceptions are disabled with `-fno-exceptions`.

And per my other comment I think you need to add a `try`/`catch` in the fiber start function, since we don't want fibers to be required to catch `CanceledException` themselves... so that'll also need to be #ifdef'd out if exceptions are disabled.

ObsidianMinor: The start function being `armDepthFirst`? Also there is no `throw` statement in `fiberCanceledException`.

kentonv: Sorry, I had missed your latest changes when I wrote my last comment. It looks like you've got it mostly right. But `FiberBase::run()` looks like it won't compile if `-fno-exceptions` is enabled, so it needs to be #ifdef'd out. (That's what I meant by "start function".)

ObsidianMinor: Ah ok. :+1:

c++/src/kj/async.c++
```diff
@@ -914,16 +917,31 @@ void FiberBase::switchToMain() {
...
+        throw;
+      }
+      caughtCanceled = true;
+    }
   })) {
     result.addException(kj::mv(*exception));
   }
 
+  if (state == CANCELED && !caughtCanceled) {
+    // if the fiber caught the canceled exception, throw it again
```
kentonv, line 942: Throwing here will terminate the process, which is a little too harsh. I was thinking something like:

    KJ_LOG(ERROR, "Canceled fiber apparently caught CanceledException and didn't rethrow it. "
        "Generally, applications should not catch CanceledException, but if they do, they must always rethrow.");

c++/src/kj/async.c++
```diff
@@ -1154,14 +1170,14 @@ void waitImpl(Own<_::PromiseNode>&& node, _::ExceptionOrValue& result, WaitScope
 
     // The main stack switched back to us, meaning either the event we registered with
     // node->onReady() fired, or we are being canceled by FiberBase's destructor.
-
+
```
kentonv, line 1175: Nitpick: Added trailing whitespace. Can you configure your editor to automatically remove this on save?

c++/src/kj/exception.h
```diff
@@ -132,8 +132,15 @@ class Exception {
   friend class ExceptionImpl;
 };
 
+struct CanceledException { };
+  // This exception is thrown to force-unwind a stack in order to immediately cancel whatever that
```
kentonv, line 136: Nitpick: The doc comment should only be indented like this if it's inside the `{}` block. Since it isn't, reduce indentation by one level.

c++/src/kj/exception.c++
```diff
@@ -1133,6 +1137,8 @@ Maybe<Exception> runCatchingExceptions(Runnable& runnable) noexcept {
   } catch (Exception& e) {
     e.truncateCommonTrace();
     return kj::mv(e);
+  } catch (CanceledException) {
```
kentonv, line 1140: We should have a test case that verifies that `runCatchingExceptions()` doesn't catch `CanceledException`.

Or alternatively you could change the nesting in `FiberBase::run()` so that the `try/catch` is outside the `runCatchingExceptions()`... that would prove that `CanceledException` is rethrown.

c++/src/kj/exception.c++
```diff
@@ -733,6 +733,10 @@ String KJ_STRINGIFY(const Exception& e) {
              stringifyStackTrace(e.getStackTrace()));
 }
 
+StringPtr KJ_STRINGIFY(CanceledException e) {
```
kentonv, line 736: Eh, I don't think a stringifier is needed for this type, since it has only one possible value.

c++/src/kj/async-inl.h
```diff
@@ -1093,13 +1093,17 @@ inline PromiseForResult<Func, void> evalNow(Func&& func) {
...
+    KJ_UNIMPLEMENTED("Fibers require exceptions");
```
kentonv, line 1097: I'd still like to see this moved into FiberBase's constructor. A couple issues with having it here:

* It's a little messy to have two different places in the code where we're using #ifdef's to check whether fibers are supported and throwing if not. Might as well consolidate that.
* Generally `kj/debug.h` isn't supposed to be included from other headers, so this macro shouldn't be available here... apparently it must have been included by mistake. The motivation behind that rule is that I worry that these macros expand into a surprising amount of code and therefore putting them into inlined methods and templates could contribute to excessive code bloat.

kentonv:
Nice work!

kentonv:
@ObsidianMinor Looks like the release tests failed: https://github.com/capnproto/capnproto/actions/runs/91563969

Looks like the `-fno-exceptions` build generates a new warning:

    src/kj/async.c++:1144:30: error: unused function 'fiberCanceledException' [-Werror,-Wunused-function]
    static kj::CanceledException fiberCanceledException() {

You should be able to reproduce this locally by building Cap'n Proto like:

    sudo apt install automake autoconf libtool
    cd c++
    autoreconf -i
    mkdir build
    cd build
    ../configure --disable-shared CXXFLAGS=-O2 -DDEBUG -Wall -Wextra -Werror -Wno-strict-aliasing -Wno-sign-compare -Wno-unused-parameter -DCAPNP_EXPENSIVE_TESTS=1 -Wno-error=unused-command-line-argument -fno-rtti -fno-exceptions

PR #997, opened 2020-05-01
Skip test that ticles MinGW 32-bit compiler bug.

kentonv:
cc @ObsidianMinor


kentonv:
Gonna go ahead and merge since the release tests won't run until it hits master anyway.

PR #998, opened 2020-05-01
CI: Try using MSVC projects in 64-bit mode.

kentonv:


kentonv:
Apparently the MSVC build consistently segfaults but ONLY on 64-bit builds with MSVC project files on Visual Studio 2015... 32-bit works, ninja works, newer VS works... blergh

kentonv:
Oh geez, the reason it works with Ninja is because NINJA IS USING MINGW TO BUILD, not MSVC, why did I not notice this? :man_facepalming:

Issue #815, opened 2019-04-08
clang 8 build falied


chronoxor:
clang 8 build failed with error:
```
[ 15%] Building CXX object modules/capnproto/c++/src/kj/CMakeFiles/kj.dir/encoding.c++.o
/home/travis/build/chronoxor/CppSerialization/modules/capnproto/c++/src/kj/encoding.c++:242:19: error: no matching conversion for functional-style cast from 'const char8_t [4]' to 'kj::StringPtr'
    result.addAll(StringPtr(u8"\ufffd"));
                  ^~~~~~~~~~~~~~~~~~~~
/home/travis/build/chronoxor/CppSerialization/modules/capnproto/c++/src/kj/string.h:70:7: note: candidate constructor (the implicit copy constructor) not viable: no known conversion from 'const char8_t [4]' to 'const kj::StringPtr' for 1st argument
class StringPtr {
      ^
/home/travis/build/chronoxor/CppSerialization/modules/capnproto/c++/src/kj/string.h:70:7: note: candidate constructor (the implicit move constructor) not viable: no known conversion from 'const char8_t [4]' to 'kj::StringPtr' for 1st argument
/home/travis/build/chronoxor/CppSerialization/modules/capnproto/c++/src/kj/string.h:73:10: note: candidate constructor not viable: no known conversion from 'const char8_t [4]' to 'decltype(nullptr)' (aka 'nullptr_t') for 1st argument
  inline StringPtr(decltype(nullptr)): content("", 1) {}
         ^
/home/travis/build/chronoxor/CppSerialization/modules/capnproto/c++/src/kj/string.h:74:10: note: candidate constructor not viable: no known conversion from 'const char8_t [4]' to 'const char *' for 1st argument
  inline StringPtr(const char* value): content(value, strlen(value) + 1) {}
         ^
/home/travis/build/chronoxor/CppSerialization/modules/capnproto/c++/src/kj/string.h:141:20: note: candidate constructor not viable: no known conversion from 'const char8_t [4]' to 'ArrayPtr<const char>' for 1st argument
  inline constexpr StringPtr(ArrayPtr<const char> content): content(content) {}
                   ^
/home/travis/build/chronoxor/CppSerialization/modules/capnproto/c++/src/kj/string.h:499:19: note: candidate constructor not viable: no known conversion from 'const char8_t [4]' to 'const kj::String' for 1st argument
inline StringPtr::StringPtr(const String& value): content(value.cStr(), value.size() + 1) {}
[...]

PR #1002, opened 2020-05-12
Fix build for C++20 via reinterpret_cast

swamp-agr:
Closes https://github.com/capnproto/capnproto/issues/815.

According to [**char8_t backward compatibility remediation**](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1423r0.html) I tried to implement `reintrerpret_cast` case:

Screenshot 2020-05-12 at 18 16 44

File [encoding.c++:242](https://github.com/capnproto/capnproto/blob/master/c%2B%2B/src/kj/encoding.c%2B%2B#L242) was changed:
```
-    result.addAll(StringPtr(u8"\ufffd"));
+    result.addAll(StringPtr(reinterpret_cast<const char *>(u8"\ufffd")));
```

It works for me!

### DIFF
--- a/c++/src/kj/encoding.c++
+++ b/c++/src/kj/encoding.c++
@@ -239,7 +239,7 @@ EncodingResult<String> decodeUtf32(ArrayPtr<const char32_t> utf16) {
     }
 
   error:
-    result.addAll(StringPtr(u8"\ufffd"));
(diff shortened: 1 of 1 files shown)


swamp-agr:
Travis is happy! @kentonv please go ahead and review PR!

kentonv:
Thanks!

PR #1003, opened 2020-05-12
Add freelisting fibers API

ObsidianMinor:
This abstracts the fiber stack implementation into a separate `FiberStack` class, allowing stacks to be saved and freelisted via `FiberPool`.

### DIFF
--- a/c++/src/kj/async-inl.h
+++ b/c++/src/kj/async-inl.h
@@ -877,7 +877,8 @@ class FiberBase: public PromiseNode, private Event {
   // Base class for the outer PromiseNode representing a fiber.
 
 public:
-  FiberBase(size_t stackSize, _::ExceptionOrValue& result);
+  explicit FiberBase(size_t stackSize, _::ExceptionOrValue& result);
+  explicit FiberBase(FiberPool& pool, _::ExceptionOrValue& result);
   ~FiberBase() noexcept(false);
 
   void start() { armDepthFirst(); }
@@ -895,31 +896,20 @@ protected:
 private:
   enum { WAITING, RUNNING, CANCELED, FINISHED } state;
 
-  size_t stackSize;
-
-#if _WIN32 || __CYGWIN__
-  void* osFiber;
-#elif !__BIONIC__
-  struct Impl;
-  Impl& impl;
-#endif
-
   _::PromiseNode* currentInner = nullptr;
   OnReadyEvent onReadyEvent;
+  Own<FiberStack> stack;
+  Maybe<FiberPool&> pool;
   _::ExceptionOrValue& result;
 
   void run();
   virtual void runImpl(WaitScope& waitScope) = 0;
 
-  struct StartRoutine;
-
-  void switchToFiber();
-  void switchToMain();
-
   Maybe<Own<Event>> fire() override;
   // Implements Event. Each time the event is fired, switchToFiber() is called.
 
   friend class WaitScope;
+  friend class FiberStack;
   friend void _::waitImpl(Own<_::PromiseNode>&& node, _::ExceptionOrValue& result,
                           WaitScope& waitScope);
   friend bool _::pollImpl(_::PromiseNode& node, WaitScope& waitScope);
@@ -928,7 +918,8 @@ private:
 template <typename Func>
 class Fiber final: public FiberBase {
 public:
-  Fiber(size_t stackSize, Func&& func): FiberBase(stackSize, result), func(kj::fwd<Func>(func)) {}
+  explicit Fiber(size_t stackSize, Func&& func): FiberBase(stackSize, result), func(kj::fwd<Func>(func)) {}
+  explicit Fiber(FiberPool& pool, Func&& func): FiberBase(pool, result), func(kj::fwd<Func>(func)) {}
   ~Fiber() noexcept(false) { destroy(); }
 
   typedef FixVoid<decltype(kj::instance<Func&>()(kj::instance<WaitScope&>()))> ResultType;
@@ -1096,6 +1087,17 @@ inline PromiseForResult<Func, WaitScope&> startFiber(size_t stackSize, Func&& fu
   typedef _::FixVoid<_::ReturnType<Func, WaitScope&>> ResultT;
 
   Own<_::FiberBase> intermediate = kj::heap<_::Fiber<Func>>(stackSize, kj::fwd<Func>(func));
+  intermediate->start();
+  auto result = _::PromiseNode::to<_::ChainPromises<_::ReturnType<Func, WaitScope&>>>(
+      _::maybeChain(kj::mv(intermediate), implicitCast<ResultT*>(nullptr)));
+  return _::maybeReduce(kj::mv(result), false);
+}
+
+template <typename Func>
+inline PromiseForResult<Func, WaitScope&> FiberPool::startFiber(Func&& func) {
+  typedef _::FixVoid<_::ReturnType<Func, WaitScope&>> ResultT;
+
+  Own<_::FiberBase> intermediate = kj::heap<_::Fiber<Func>>(*this, kj::fwd<Func>(func));
   intermediate->start();
   auto result = _::PromiseNode::to<_::ChainPromises<_::ReturnType<Func, WaitScope&>>>(
       _::maybeChain(kj::mv(intermediate), implicitCast<ResultT*>(nullptr)));
--- a/c++/src/kj/async.c++
+++ b/c++/src/kj/async.c++
@@ -291,6 +291,73 @@ Promise<void> TaskSet::onEmpty() {
   }
 }
 
+namespace _ {
+
+class FiberStack final {
+  // A class containing a fiber stack impl. This is seperate from fiber
+  // promises since it lets us move the stack itself around and reuse it.
+
+public:
+  FiberStack(size_t stackSize);
+  ~FiberStack() noexcept(false);
+
+  void initialize(FiberBase& fiber);
+
+  void reset() {
+    fiber = nullptr;
+  }
+
+  void switchToFiber();
+  void switchToMain();
+private:
+  size_t stackSize;
+  Maybe<FiberBase&> fiber;
+  Maybe<Own<FiberStack>> freelistNext;
+
+  friend class FiberBase;
+  friend struct FiberPool::Impl;
+
+  struct StartRoutine;
+
+#if _WIN32 || __CYGWIN__
+  void* osFiber;
+#elif !__BIONIC__
+  struct Impl;
+  Impl& impl;
+#endif
+};
+
+}
+
+struct FiberPool::Impl {
+  Impl(size_t stackSize): stackSize(stackSize) {}
+
+  size_t stackSize;
+  MutexGuarded<Maybe<Own<_::FiberStack>>> start;
+
+  void returnStack(Own<_::FiberStack> stack) {
+    stack->reset();
+    auto lock = start.lockExclusive();
+    stack->freelistNext = kj::mv(*lock);
+    *lock = kj::mv(stack);
+  }
+
+  Own<_::FiberStack> takeStack() {
+    auto lock = start.lockExclusive();
+    auto& value = *lock;
+    KJ_IF_MAYBE(current, value) {
+      auto moved = kj::mv(*current);
+      *lock = kj::mv(moved->freelistNext);
+      return kj::mv(moved);
+    } else {
+      return kj::heap<_::FiberStack>(stackSize);
+    }
+  }
+};
+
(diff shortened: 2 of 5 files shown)


c++/src/kj/async-inl.h
```diff
@@ -1095,7 +1114,33 @@ template <typename Func>
 inline PromiseForResult<Func, WaitScope&> startFiber(size_t stackSize, Func&& func) {
   typedef _::FixVoid<_::ReturnType<Func, WaitScope&>> ResultT;
 
-  Own<_::FiberBase> intermediate = kj::heap<_::Fiber<Func>>(stackSize, kj::fwd<Func>(func));
+  Own<_::FiberStack> stack = kj::heap<_::FiberStack>(stackSize);
```
kentonv, line 1117: Ideally, we want to keep the amount of code in a template function as minimal as possible, since the compiler generates a separate copy of this whole function for each instantiation (which, in this function's case, is each place where it is called, since the `Func` is different every time!).

Instead of adding all this stack-allocating code here, what if `FiberBase`'s constructor accepted `size_t stackSize, Maybe<FiberPool&>`, and then did the appropriate allocation depending on whether a FiberPool was provided?

I think this would also allow you to move `FiberStack`'s definition entirely to the `.c++` file, which is always nice since the header file needs to be re-parsed for every single dependent file we compile.

c++/src/kj/async-inl.h
```diff
@@ -1095,7 +1114,33 @@ template <typename Func>
...
+  Own<_::FiberBase> intermediate = kj::heap<_::Fiber<Func>>(*stack, kj::fwd<Func>(func));
+  stack->initialize(*intermediate);
+  intermediate = intermediate.attach(kj::mv(stack));
+  intermediate->start();
+  auto result = _::PromiseNode::to<_::ChainPromises<_::ReturnType<Func, WaitScope&>>>(
+      _::maybeChain(kj::mv(intermediate), implicitCast<ResultT*>(nullptr)));
+  return _::maybeReduce(kj::mv(result), false);
+}
+
+struct FiberPool::Node final: public _::FiberStack {
```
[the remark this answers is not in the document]
kentonv: In fact, I think the use of inheritance here may be overkill. I think you could have `FiberStack` itself include `Maybe<Own<FiberStack>> freelistNext;`, which would just remain null for stacks that aren't part of a freelist. It's not as elegant but the code would be a bit easier to follow. These classes are all private implementation details anyway.

c++/src/kj/async-inl.h
ObsidianMinor: Or if it was a separate constructor, yeah I could see that.

c++/src/kj/async.c++
```diff
@@ -853,14 +920,52 @@ FiberBase::FiberBase(size_t stackSizeParam, _::ExceptionOrValue& result)
...
+FiberStack::~FiberStack() noexcept(false) {
 #if _WIN32 || __CYGWIN__
   DeleteFiber(osFiber);
 #elif !__BIONIC__
   Impl::free(impl, stackSize);
 #endif
 }
 
+void FiberStack::initialize(FiberBase& fiber) {
+  KJ_ASSERT(this->fiber == nullptr);
```
kentonv, line 932: Nitpick: Use `KJ_REQUIRE` to check preconditions (failure = there is a bug in the caller). Use `KJ_ASSERT` is to test assumptions (failure = there's a bug in this code).

(In practice, they do exactly the same thing, but I like to use one or the other for self-documentation... admittedly I might eventually give up on this since even I mix them up fairly regularly.)

c++/src/kj/async.c++
```diff
@@ -853,14 +920,52 @@ FiberBase::FiberBase(size_t stackSizeParam, _::ExceptionOrValue& result)
...
+  this->fiber = fiber;
+}
+
+FiberBase::FiberBase(size_t stackSize, _::ExceptionOrValue& result)
+    : state(WAITING), stack(kj::heap<FiberStack>(stackSize)), result(result) {
+    stack->initialize(*this);
```
kentonv, line 938: Nitpick: too much indentation on this line.

c++/src/kj/async.c++
```diff
@@ -291,31 +291,77 @@ Promise<void> TaskSet::onEmpty() {
...
-  auto& value = *lock;
-  KJ_IF_MAYBE(current, value) {
-    node->next = kj::mv(*current);
-    *current = kj::mv(node);
-  } else {
-    *lock = kj::mv(node);
+namespace _ {
+
+class FiberStack final {
+  // A class containing a fiber stack impl. This is seperate from fiber
```
jclee, line 297: s/seperate/separate

ObsidianMinor:
Ready for re-review. Added in some optimizations and fixed stacks so they can be reused properly. Test has also been updated to make sure fibers are reused.

ObsidianMinor:
Ping @kentonv

c++/src/kj/async.c++
```diff
@@ -19,6 +19,8 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
+#undef _FORTIFY_SOURCE // prevent longjmp from doing checks in fibers
```
kentonv, line 22: Probably could use a clearer comment, like:

> If _FORTIFY_SOURCE is defined, longjmp() will complain when it detects the stack pointer is moving in the "wrong direction", thinking that you're jumping to a stack frame that no longer exists. However, we use longjmp() to jump between entirely different stacks to implement fibers. So, this debug check is not appropriate for us.

c++/src/kj/async.c++
```diff
@@ -829,11 +835,13 @@ struct FiberStack::Impl {
...
-    KJ_SYSCALL(getcontext(&impl.fiberContext));
-    impl.fiberContext.uc_stack.ss_size = allocSize - sizeof(Impl);
-    impl.fiberContext.uc_stack.ss_sp = reinterpret_cast<char*>(stack);
-    impl.fiberContext.uc_stack.ss_flags = 0;
-    impl.fiberContext.uc_link = &impl.originalContext;
+    KJ_SYSCALL(getcontext(&impl.fiberContext.context));
+    impl.fiberContext.context.uc_stack.ss_size = allocSize - sizeof(Impl);
+    impl.fiberContext.context.uc_stack.ss_sp = reinterpret_cast<char*>(stack);
+    impl.fiberContext.context.uc_stack.ss_flags = 0;
+     // to reuse fibers the start fibers runs forever, so this doesn't
```
kentonv, line 842: nitpick: Comment formatted weirdly. Extra space of indentation and wraps too early.

c++/src/kj/async.c++
```diff
@@ -878,6 +886,11 @@ struct FiberStack::StartRoutine {
     ptr |= static_cast<uintptr_t>(static_cast<uint>(arg2)) << (sizeof(ptr) * 4);
     auto stack = reinterpret_cast<FiberStack*>(ptr);
 
+    /// Jump back into the fiber stack constructor
```
kentonv, line 889: nitpick: three slashes?

c++/src/kj/async.c++
```diff
@@ -878,6 +886,11 @@ struct FiberStack::StartRoutine {
...
+    if (setjmp(&stack->impl.fiberContext.jumpBuf[0]) == 0) {
```
kentonv, line 890: `&jumpBuf[0]` is redundant (kind of like saying `&*jumpBuf`). You should be able to write: `setjmp(stack->impl.originalContext.jumpBuf)`

Similarly for all the other setjmp/longjmp calls below.

c++/src/kj/async.c++
kentonv: The comment is also a little obtuse, maybe we could phrase it as: "We don't use `uc_link` because our fiber start routine runs forever in a loop to allow for reuse. When we're done with the fiber, we just destroy it, without switching back to its stack. This is safe since the start routine doesn't allocate any memory or RAII objects before looping."

c++/src/kj/async.c++
```diff
@@ -878,6 +886,11 @@ struct FiberStack::StartRoutine {
...
+      swapcontext(&stack->impl.fiberContext.context, &stack->impl.originalContext.context);
```
kentonv, line 891: This can be `setcontext()` rather than `swapcontext()` as we don't use `fiberContext.context` anymore after this point.

Though better yet, I think you could be using `longjmp()` here too to return to the calling fiber, if `FiberStack`'s constructor were to call `setjmp()`. And then you could use `setcontext()` rather than `swapcontext()` there.

Then you can go another step further: You only ever really need a `ucontext_t` when you are calling `makecontext()` to create the fiber initially. You could allocate that `ucontext_t` on the stack inside `FiberStack`'s constructor, rather than put it into `FastContext`. Then you could remove `FastContext::context` entirely; you only need `jumpBuf` there.

Let me know if that makes sense...

c++/src/kj/async.c++
ObsidianMinor: Ah that makes a lot more sense than what the errors were telling me.

c++/src/kj/async.c++
ObsidianMinor: Set context stack info, make the context, setjmp, set the context to start, longjmp back. Got it.

c++/src/kj/async.c++
ObsidianMinor: The Rustdoc Way:tm: and a force of habit :)

ObsidianMinor:
Ping2 @kentonv

c++/src/kj/async.c++
```diff
@@ -831,24 +830,26 @@ struct FiberStack::Impl {
...
 
-    KJ_SYSCALL(getcontext(&impl.fiberContext.context));
-    impl.fiberContext.context.uc_stack.ss_size = allocSize - sizeof(Impl);
-    impl.fiberContext.context.uc_stack.ss_sp = reinterpret_cast<char*>(stack);
-    impl.fiberContext.context.uc_stack.ss_flags = 0;
-     // to reuse fibers the start fibers runs forever, so this doesn't
-     // need to point to the main context
-    impl.fiberContext.context.uc_link = 0;
+    KJ_SYSCALL(getcontext(context));
+    context->uc_stack.ss_size = allocSize - sizeof(Impl);
```
kentonv, line 838: Hmm, I think it would be clearer to move this initialization of `context` to happen inside `FiberStack`'s constructor, since it's no longer a member of `Impl`.

And then you don't have to change `impl` to be a pointer...

ObsidianMinor: It'll require doing the pointer math for the stack allocation again inside the `FiberStack` ctor as well as `Impl::alloc`.

ObsidianMinor: Done

kentonv: You are right, the way you did it made more sense. Sorry about that. I guess you can remove the last commit.

Can you rebase the PR to make the commit history cleaner? I think it would make sense to squash down to two commits:

1. Implement freelist of fiber stacks.
2. Switch to setjmp/longjmp.

I think this should be just a matter of `git rebase -i master` and then squash the first sequence of commits and then the second sequence, and update the resulting commit descriptions -- probably no need to reorder anything.

ObsidianMinor:
Rebased and squashed @kentonv

c++/src/kj/async.c++
```diff
@@ -1047,15 +1056,20 @@ FiberStack::FiberStack(size_t stackSizeParam)
...
   // POSIX says the arguments are ints, not pointers. So we split our pointer in half in order to
   // work correctly on 64-bit machines. Gross.
   uintptr_t ptr = reinterpret_cast<uintptr_t>(this);
   int arg1 = ptr & ((uintptr_t(1) << (sizeof(ptr) * 4)) - 1);
   int arg2 = ptr >> (sizeof(ptr) * 4);
 
-  makecontext(&impl.fiberContext, reinterpret_cast<void(*)()>(&StartRoutine::run), 2, arg1, arg2);
+  makecontext(&context, reinterpret_cast<void(*)()>(&StartRoutine::run), 2, arg1, arg2);
 
+  if (setjmp(impl->originalJmpBuf) == 0) {
```
kentonv, line 1070: Oops, should be `_setjmp`?

kentonv: Sorry for the run-around. Feel free to use `--amend` for this one fix, then should be good to merge.

ObsidianMinor: Hmm how did that get there.

ObsidianMinor: Fixed.

PR #1005, opened 2020-05-13
 Extend Executor to gracefully handle remote event loop exits.

kentonv:
In practice, we've found that Executor is hard to use correctly if it's possible that the target event loop might exit while other threads are still trying to queue things to its executor. This change makes handling that case more graceful. Now, if an event loop exits without completing a cross-thread task, the requesting thread gets a DISCONNECTED exception (instead of hanging forever or crashing). It's also now possible to hold on to an Executor reference beyond the lifetime of the destination thread, and to query whether its event loop still lives.

### DIFF
--- a/c++/src/kj/async.c++
+++ b/c++/src/kj/async.c++
@@ -392,7 +392,7 @@ struct Executor::Impl {
         cancel.erase(event);
 
         if (event.promiseNode == nullptr) {
-          event.state = _::XThreadEvent::DONE;
+          event.setDoneState();
         } else {
           // We can't destroy the promiseNode while the mutex is locked, because we don't know
           // what the destructor might do. But, we *must* destroy it before acknowledging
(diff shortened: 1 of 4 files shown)


c++/src/kj/async.c++
```diff
@@ -421,39 +434,99 @@ struct Executor::Impl {
...
-      event->state = _::XThreadEvent::DONE;
+      event->setDoneState();
     }
   }
-};
+
+  void disconnect() {
+    state.lockExclusive()->loop = nullptr;
+
+    // Now that `loop` is set null is `state`, other threads will no longer try to manipulate our
```
harrishancock, line 444: Nit: "Now that `loop` is set null is `state`" sounds strange. Delete "is `state`"?

kentonv: Typo, s/is/in/

PR #1008, opened 2020-06-01
Add const modifier to FiberPool::startFiber

ObsidianMinor:
Should've been done originally since `FiberPool` is thread-safe.


ObsidianMinor:
cc: @kentonv

kentonv:
Thanks, and sorry I missed this earlier!

PR #1009, opened 2020-06-05
Fix tail call optimization of AsyncTee::pull() loop.

kentonv:


kentonv:
> Is there a way to articulate a rule that we should follow when writing async recursive functions in order to avoid breaking TCO?

It''s not quite as complicated as it looks. You have to make sure the tail call is the last thing that happens in each loop iteration (otherwise, it's not a tail call).

So this is a tail call:

    Promise<void> loop() {
      foo().then([]() { return loop(); });
    }

But this isn't:

    Promise<void> loop() {
      foo().then([]() { return loop(); })
          .then([]() { something(); });
    }

Here, you can see that the continuation that executes `something()` would run _after_ the invocation of `loop()`. So if you loop 1000 times, then you've actually queued up 1000 calls to `something()` that won't run until the loop decides to stop looping and actually return a result. KJ needs to remember that all these calls are queued up, hence the memory leak...

kentonv:
I guess where this gets weird is when you have added an error handler to the chain, where the error handler actually handles the error (doesn't pass it through). Then it looks like this error handler should only run once. But KJ doesn't _know_ that the error handler won't itself throw, so it still builds a list of all the error handlers you've queued up, just in case they all do need to execute.

`.attach()` can also seem innocuous because we don't think of it as adding more code to run. But, it does in fact queue up execution of a destructor, which will need to run in reverse order...

PR #1010, opened 2020-06-05
Extend event loop to support switching stacks while running callbacks. 

kentonv:
The event loop will use the main stack while sleeping waiting for I/O, but will switch to a FiberPool stack while running event callbacks. Programs with many threads that spend most of their time sleeping can use this to save memory by sharing a small number of "big stacks" when actually executing.

/cc @ObsidianMinor (would set you as a reviewer but GitHub won't let me...)

### DIFF
--- a/c++/src/kj/async-test.c++
+++ b/c++/src/kj/async-test.c++
@@ -1053,6 +1053,115 @@ KJ_TEST("fiber pool runSynchronously()") {
   KJ_EXPECT_THROW_MESSAGE("test exception",
       pool.runSynchronously([&]() { KJ_FAIL_ASSERT("test exception"); }));
 }
+
+KJ_TEST("run event loop on freelisted stacks") {
+  FiberPool pool(65536);
+
+  class MockEventPort: public EventPort {
+  public:
+    bool wait() override {
+      char c;
+      waitStack = &c;
+      KJ_IF_MAYBE(f, fulfiller) {
+        f->get()->fulfill();
+        fulfiller = nullptr;
+      }
+      return false;
+    }
+    bool poll() override {
+      char c;
+      pollStack = &c;
+      KJ_IF_MAYBE(f, fulfiller) {
+        f->get()->fulfill();
+        fulfiller = nullptr;
+      }
+      return false;
+    }
+
+    char* waitStack = nullptr;
+    char* pollStack = nullptr;
+
+    kj::Maybe<kj::Own<PromiseFulfiller<void>>> fulfiller;
+  };
+
+  MockEventPort port;
+  EventLoop loop(port);
+  WaitScope waitScope(loop);
+  waitScope.runEventCallbacksOnStackPool(pool);
+
+  {
+    auto paf = newPromiseAndFulfiller<void>();
+    port.fulfiller = kj::mv(paf.fulfiller);
+
+    char* ptr1 = nullptr;
+    char* ptr2 = nullptr;
+    kj::evalLater([&]() {
+      char c;
+      ptr1 = &c;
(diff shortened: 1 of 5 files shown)


kentonv:
These test failures are nonsense. Google says that `Exit code 0xc0000139` is Windowsese for "couldn't find DLL", but what the heck DLL is it talking about? Both MSVC and MinGW fail in the same way? Feels like the runner broke, but tests running at master still seem to pass. WTF.

kentonv:
OK, I think it's because I used `thread_local`. I guess that... doesn't work right except with MSVC 2019. Ironically it worked fine on MinGW+WINE running locally. Ugh.

kentonv:
Oh duh my local build didn't use DLLs.

c++/src/kj/async.c++
```diff
@@ -374,13 +435,65 @@ private:
   size_t stackSize;
   MutexGuarded<kj::Vector<_::FiberStack*>> freelist;
 
+#if USE_CORE_LOCAL_FREELISTS
+  struct CoreLocalFreelist {
+    union {
+      _::FiberStack* stacks[2];
+      // For now, we don't try to freelist more than 2 stacks per core. If you have three or more
+      // threads interleaved on a core, chances are you have bigger problems...
```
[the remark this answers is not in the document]
kentonv: I mean if the number of _runnable_ threads (not blocked on anything) is much larger than the core count.

c++/src/kj/async-test.c++
```diff
@@ -1053,6 +1053,115 @@ KJ_TEST("fiber pool runSynchronously()") {
...
+        f->get()->fulfill();
+        fulfiller = nullptr;
+      }
+      return false;
+    }
+
+    char* waitStack = nullptr;
+    char* pollStack = nullptr;
+
+    kj::Maybe<kj::Own<PromiseFulfiller<void>>> fulfiller;
```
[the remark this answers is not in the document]
kentonv: No, it's definitely used.

Note that this is declaring a class member, not a local variable.

c++/src/kj/async.c++
```diff
@@ -332,39 +370,213 @@ private:
...
+
+  bool isReset() { return main == nullptr; }
 };
 
-}
+}  // namespace _
+
+#if __linux__
+// TODO(someday): Support core-local freelists on OSs other than Linux. The only tricky part is
+//   finding what to use instead of sched_getcpu() to get the current CPU ID.
```
ObsidianMinor, line 383: On Windows you'd use `GetCurrentProcessorNumberEx` (idk about OSX) but we could probably leave that for another PR later.

kentonv: Oh nice, that even identifies physical cores vs. hyperthreads. For our use case we'd care about physical cores...

c++/src/kj/async-test.c++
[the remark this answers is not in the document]
harrishancock: Ah, I missed the `port.fulfiller = ...` lines.

c++/src/kj/async.c++
ObsidianMinor: I don't think so? Windows [processor groups](https://docs.microsoft.com/en-us/windows/win32/procthread/processor-groups) are static groups of 64 logical processors used by Windows for scheduling, not physical cores.

kentonv: Oh, darn... I saw them use the term "logical processor" which is commonly used as a generic term for "hyperthread", and assumed that distinguishing hyperthreads from cores was the whole point, but it appears not. Darn.

PR #1014, opened 2020-06-18
Consistent checking of __GNUC__ define.

ScatteredRay:
The kj/common.h header sometimes checks for if __GNUC__ is defined, and sometimes if it's true, this is changing it to be consistent in checking if it's true.

Specifically, MSVC classifies checking the value of an undefined define as a warning, this header can be included in other projects that have warnings as errors turned on.

My approach to solving this was to manually define these defines as 0, such as ___GNUC__, __CLANG__, __NO_INLINE__, KJ_USE_FUTEX, etc. However doing such for __GNUC__ caused additional problems due to this inconcinstency when it is defined, but set to 0.

This could alternativelly be fixed by consistently checking for all the macro definitions being undefined.

### DIFF
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -25,7 +25,7 @@
 
 #pragma once
 
-#ifdef __GNUC__
+#if __GNUC__
 #define KJ_BEGIN_SYSTEM_HEADER _Pragma("GCC system_header")
 #elif defined(_MSC_VER)
 #define KJ_BEGIN_SYSTEM_HEADER __pragma(warning(push, 0))
@@ -55,13 +55,13 @@ KJ_BEGIN_HEADER
 // had it defined to 201300L even with -std=c++14.
 #if __cplusplus < 201300L && !__CDT_PARSER__ && !_MSC_VER
   #error "This code requires C++14. Either your compiler does not support it or it is not enabled."
-  #ifdef __GNUC__
+  #if __GNUC__
     // Compiler claims compatibility with GCC, so presumably supports -std.
     #error "Pass -std=c++14 on the compiler command line to enable C++14."
   #endif
 #endif
 
-#ifdef __GNUC__
+#if __GNUC__
   #if __clang__
     #if __clang_major__ < 5
       #warning "This library requires at least Clang 5.0."
@@ -117,7 +117,7 @@ typedef unsigned char byte;
 // Detect whether RTTI and exceptions are enabled, assuming they are unless we have specific
 // evidence to the contrary.  Clients can always define KJ_NO_RTTI or KJ_NO_EXCEPTIONS explicitly
 // to override these checks.
-#ifdef __GNUC__
+#if __GNUC__
   #if !defined(KJ_NO_RTTI) && !__GXX_RTTI
     #define KJ_NO_RTTI 1
   #endif
@@ -152,7 +152,7 @@ typedef unsigned char byte;
   classname& operator=(const classname&) = delete
 // Deletes the implicit copy constructor and assignment operator.
 
-#ifdef __GNUC__
+#if __GNUC__
 #define KJ_LIKELY(condition) __builtin_expect(condition, true)
 #define KJ_UNLIKELY(condition) __builtin_expect(condition, false)
 // Branch prediction macros.  Evaluates to the condition given, but also tells the compiler that we


kentonv:
Hmm, we wrap every one of our header files with KJ_BEGIN_HEADER and KJ_END_HEADER, whose job is specifically to disable compiler warnings when the header is being included in other projects. They are defined as:

```
#ifdef __GNUC__
#define KJ_BEGIN_SYSTEM_HEADER _Pragma("GCC system_header")
#elif defined(_MSC_VER)
#define KJ_BEGIN_SYSTEM_HEADER __pragma(warning(push, 0))
#define KJ_END_SYSTEM_HEADER __pragma(warning(pop))
#endif
```

Is this mechanism not working correctly for you?

Unfortunately if you change that very first `#ifdef` to `#if`, then I think will be that you are creating a new MSVC warning for most people which sneaks in _before_ we manage to disable warnings.

ScatteredRay:
Hmm no, that mechanism isn't working for me. I'll try to diagnose it next week.

ScatteredRay:
@kentonv So, I finally got back to this (some other things forced me to look into it.) I'm not entirely certain yet why I'm getting this, but here are my errors:

C:\work\Fables2\Fables2\SceneFlow\external\capnproto\c++\src\kj/common.h(28): error C4668: '__GNUC__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  C:\work\Fables2\Fables2\SceneFlow\external\capnproto\c++\src\kj/common.h(92): error C4668: '__linux__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  C:\work\Fables2\Fables2\SceneFlow\external\capnproto\c++\src\kj/common.h(120): error C4668: '__GNUC__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  C:\work\Fables2\Fables2\SceneFlow\external\capnproto\c++\src\kj/common.h(155): error C4668: '__GNUC__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  C:\work\Fables2\Fables2\SceneFlow\external\capnproto\c++\src\kj/common.h(166): error C4668: '__NO_INLINE__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  C:\work\Fables2\Fables2\SceneFlow\external\capnproto\c++\src\kj/common.h(184): error C4668: '__clang__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  C:\work\Fables2\Fables2\SceneFlow\external\capnproto\c++\src\kj/common.h(197): error C4668: '__clang__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  C:\work\Fables2\Fables2\SceneFlow\external\capnproto\c++\src\kj/common.h(205): error C4668: '__clang__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  C:\work\Fables2\Fables2\SceneFlow\external\capnproto\c++\src\kj/common.h(210): error C4668: '__GNUC__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  C:\work\Fables2\Fables2\SceneFlow\external\capnproto\c++\src\kj/common.h(263): error C4668: '__clang__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  C:\work\Fables2\Fables2\SceneFlow\external\capnproto\c++\src\kj/common.h(275): error C4668: '__GNUC__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  C:\work\Fables2\Fables2\SceneFlow\external\capnproto\c++\src\kj/common.h(501): error C4668: '__GNUC__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  C:\work\Fables2\Fables2\SceneFlow\external\capnproto\c++\src\kj/common.h(690): error C4668: '__GNUC__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  C:\work\Fables2\Fables2\SceneFlow\external\capnproto\c++\src\kj/common.h(690): error C4668: '__clang__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  C:\work\Fables2\Fables2\SceneFlow\external\capnproto\c++\src\kj/common.h(1391): error C4668: '__GNUC__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  C:\work\Fables2\Fables2\SceneFlow\external\capnproto\c++\src\kj/common.h(1415): error C4668: '__GNUC__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  c:\work\fables2\fables2\sceneflow\external\capnproto\c++\src\kj\string.h(57): error C4668: '__clang__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  c:\work\fables2\fables2\sceneflow\external\capnproto\c++\src\kj\string.h(57): error C4668: '__GNUC__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  C:\work\Fables2\Fables2\SceneFlow\external\capnproto\c++\src\kj/mutex.h(30): error C4668: '__linux__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  c:\work\fables2\fables2\sceneflow\external\capnproto\c++\src\capnp\common.h(359): error C4668: '__GNUC__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  c:\work\fables2\fables2\sceneflow\external\capnproto\c++\src\capnp\raw-schema.h(148): error C4668: '__GNUC__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  c:\work\fables2\fables2\sceneflow\external\capnproto\c++\src\capnp\raw-schema.h(214): error C4668: '__GNUC__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'

It's possible maybe that warnings as errors is causing this to become a problem first?

kentonv:
Hi @ScatteredRay,

It looks like that error log comes from compiling your modified code. What errors do you see if your changes are not applied?

What's supposed to happen is:

* On line 51 of `common.h`, we use `KJ_BEGIN_HEADER`.
* On MSVC (which I think is what you're using?), that macro _should_ expand to `__pragma(warning(push, 0))`.
* That is _supposed to_ disable all warnings within the header file, up until `KJ_END_HEADER` at the end, which expands to `__pragma(warning(pop))`, restoring the warnings settings.

It seems like this is not working for you, but I don't understand why. Could you maybe examine the first 50 lines of `common.h` and figure out why they don't result in `KJ_BEGIN_HEADER` working correctly in your environment?

The reason we have this system is so that we don't have to worry about fixing every possible warning on every possible configuration of every possible compiler...

ScatteredRay:
Yeah, actually doing an upgrade, will report back what happens when I build with newest without my changes applied.

ScatteredRay:
Ignore that updated branch, As a quick update, removing my changes and updates results in similar errors:

C:\work\Fables2\Fables2\SceneFlow\external\capnproto\c++\src\kj/common.h(92): error C4668: '__linux__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  C:\work\Fables2\Fables2\SceneFlow\external\capnproto\c++\src\kj/common.h(178): error C4668: '__NO_INLINE__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  C:\work\Fables2\Fables2\SceneFlow\external\capnproto\c++\src\kj/common.h(196): error C4668: '__clang__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  C:\work\Fables2\Fables2\SceneFlow\external\capnproto\c++\src\kj/common.h(209): error C4668: '__clang__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  C:\work\Fables2\Fables2\SceneFlow\external\capnproto\c++\src\kj/common.h(217): error C4668: '__clang__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  C:\work\Fables2\Fables2\SceneFlow\external\capnproto\c++\src\kj/common.h(222): error C4668: '__GNUC__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  C:\work\Fables2\Fables2\SceneFlow\external\capnproto\c++\src\kj/common.h(275): error C4668: '__clang__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  C:\work\Fables2\Fables2\SceneFlow\external\capnproto\c++\src\kj/common.h(281): error C4668: '__clang__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  C:\work\Fables2\Fables2\SceneFlow\external\capnproto\c++\src\kj/common.h(301): error C4668: '__GNUC__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  C:\work\Fables2\Fables2\SceneFlow\external\capnproto\c++\src\kj/common.h(519): error C4668: '__GNUC__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  C:\work\Fables2\Fables2\SceneFlow\external\capnproto\c++\src\kj/common.h(708): error C4668: '__GNUC__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  C:\work\Fables2\Fables2\SceneFlow\external\capnproto\c++\src\kj/common.h(708): error C4668: '__clang__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  C:\work\Fables2\Fables2\SceneFlow\external\capnproto\c++\src\kj/common.h(1397): error C4668: '__GNUC__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  C:\work\Fables2\Fables2\SceneFlow\external\capnproto\c++\src\kj/common.h(1421): error C4668: '__GNUC__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  c:\work\fables2\fables2\sceneflow\external\capnproto\c++\src\kj\string.h(57): error C4668: '__clang__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  c:\work\fables2\fables2\sceneflow\external\capnproto\c++\src\kj\string.h(57): error C4668: '__GNUC__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  C:\work\Fables2\Fables2\SceneFlow\external\capnproto\c++\src\kj/mutex.h(30): error C4668: '__linux__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  c:\work\fables2\fables2\sceneflow\external\capnproto\c++\src\kj\exception.h(398): error C4668: 'KJ_NO_EXCEPTIONS' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  C:\work\Fables2\Fables2\SceneFlow\external\capnproto\c++\src\kj/debug.h(608): error C4668: '__clang__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  c:\work\fables2\fables2\sceneflow\external\capnproto\c++\src\capnp\common.h(359): error C4668: '__GNUC__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  c:\work\fables2\fables2\sceneflow\external\capnproto\c++\src\capnp\raw-schema.h(148): error C4668: '__GNUC__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
  c:\work\fables2\fables2\sceneflow\external\capnproto\c++\src\capnp\raw-schema.h(214): error C4668: '__GNUC__' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'

I'm trying to diagnose this further, but for some reason, right now it's only seeming to manifest in the middle of a 6 hour long automated build, so it's going a bit slowly.

ScatteredRay:
This seems related: https://social.msdn.microsoft.com/Forums/sqlserver/en-US/891a02d2-d0cf-495a-bcea-41001cf599de/pragma-warningpush-0-cant-disable-wall-warning-on-visual-studio-2017

They seem to be saying that /Wall supersedes #pragma warning(push, 0) by design, and the individual warnings need to be suppressed, which is a bit of a pain as I don't really controller compiler flags in calling libraries.

kentonv:
This is unfortunate. Based on what you say, it seems that MSVC's `/Wall /Werror` makes it effectively impossible to consume libraries that aren't themselves `/Wall`-clean. That seems like a surprising oversight in MSVC's design, which makes me wonder if there's something else we're missing here. But if not, then my opinion is that MSVC's `/Wall` is fundamentally broken and should not be used.

I'm afraid that trying to keep Cap'n Proto `/Wall` client is just not feasible for us. MSVC implements many warnings which we feel are not helpful. Also, we do most development on Linux, so MSVC issues only come up in CI, where chasing them down is already burdensome -- having to deal with fixing spurious warnings at that stage would be too much work.

So, you're going to need to figure out how to drop that `/Wall` flag. Sorry.

(That said, if we can find some way to make `KJ_BEGIN_HEADER`/`KJ_END_HEADER` work as intended even with `/Wall`, I'm open to that.)

ScatteredRay:
Oddly enough, this is the only error that seems to cause problems. (Perhaps because it's included in generated capnp headers, it's the only one that I have to build with /wall)

So specifically suppressing the error itself in the macro is able to solve my problem

__pragma warning(push)
__pragma warning(disable : 4668)

kentonv:
OK, I'd be willing to accept extending the macro to specifically disable that warning. (But if we start seeing a lot of requests to disable various warnings, at some point I'll decide it's too much...)

kentonv:
Oh! I bet I know why only that one warning is affected: It's a preprocessor warning, and the pragmas won't actually be parsed until after preprocessing, since we're using the `__pragma` form instead of `#pragma` (and we're doing that because we want to specify the pragmas in a macro).

Unfortunately, if I'm right, then your solution won't work.

If this is the case, then I think we should just use `#pragma` to disable the warning once at the top of `kj/common.h`, without making it part of the KJ_BEGIN_HEADER/KJ_END_HEADER macros. People will just have to deal with the fact that including capnp/kj headers disables that warning.

englercj:
I'm hitting this as well, I compile with `/we4668` on MSVC and `-Wundef` on Clang. I disable warnings when building capnproto so I don't particularly care about the library being /Wall clean, but keeping the user-facing *headers* `/Wall` clean seems like a reasonable request? Capnp disabling my warning settings if I include a header is undesirable.

Would you be open to a PR that cleans up the usage here so that undefined symbols aren't checked?

kentonv:
> keeping the user-facing headers /Wall clean seems like a reasonable request?

In general, no, I don't agree. C++ headers are very complicated (with inlines and templates and whatnot), and compiler writers have managed to come up with some very pedantic and stylistic warnings that I'm not interested in working around. ​Not to mention, most capnp development is done on Linux with no ability to test against MSVC until CI, and it's be extremely frustrating for us to hit pedantic warnings at the CI stage.

This is why we created the `KJ_BEGIN_SYSTEM_HEADER` system, so that we'd stop having people complain about warnings in our headers that we weren't interested in fixing.

With that said, it does seem that `KJ_BEGIN_SYSTEM_HEADER` is not getting the job done when it comes to preprocessor warnings, and that's unfortunate. I would like for this system to work as designed.

But the question of undefined preprocessor symbols is unfortunately very stylistic. KJ style has always preferred relying on undefined symbols being falsy. I find it easier to read and write that way, so that I don't have to type out extra `defined(...)` checks all over. I certainly understand why people might disagree, but it's a style issue, we should all be able to do what we want here.

What I really don't like is the idea that I need to change my style to match someone else's preferences because they configured their compiler to make my style an error.

GCC and Clang nicely solve this issue with the concept of "system headers" which prevents warnings from being reported at all in the header. Unfortunately MSVC doesn't have this, and our attempt to implement something similar is incomplete, and it's not clear if we can do much about it. :(

englercj:
I was able to work around this with the `/external:` compiler switches for msvc. Not an ideal solution but I was able to make 4668 a level 4 warning, and disable warnings in capnp headers by marking them external. Looks like `/W4 /WX /w44668 /external:W0 /external:Icapnp/include/path`

PR #1015, opened 2020-06-19
Doc: fix typo in style guide

parkerduckworth:
Typo in 'No Singletons' section. 

Here is the change, since the diff is multi-line:
`Singleton interfaces ofter turn` -> `Singleton interfaces often turn`


parkerduckworth:
Hmm, interesting that the MinGW test failed and Cygwin test is cancelled (timed out?), as the proposed commit only affects the style guide.

More interesting is that the checks for PR #1014 are unsuccessful for the same reasons as those in this PR.

kentonv:
Indeed, it appears that MinGW and Cygwin have had updates lately that break our build. :(

Thanks for the typo fix!

PR #1016, opened 2020-06-19
Adding getAllNested and getNodeName to ParsedSchema

teqdruid:


teqdruid:
It was the node name I needed!

kentonv:
Can we make this not require allocation by using the same trick as `StructSchema::getFields()` and the like in `schema.h`?

https://github.com/capnproto/capnproto/blob/7b0a59c6fc72c33f54cb6c74dbd6718c30375580/c++/src/capnp/schema.h#L335-L354

teqdruid:
Yep!

c++/src/capnp/schema-parser.c++
```diff
@@ -328,10 +328,33 @@ ParsedSchema ParsedSchema::getNested(kj::StringPtr nestedName) const {
...
 
+ParsedSchema::ParsedSchemaList ParsedSchema::getAllNested() const {
+  return ParsedSchemaList(*this, getProto().getNestedNodes());
+}
+
 schema::Node::SourceInfo::Reader ParsedSchema::getSourceInfo() const {
   return KJ_ASSERT_NONNULL(parser->getSourceInfo(*this));
 }
 
+const kj::StringPtr ParsedSchema::getNodeName() const {
```
kentonv, line 339: I think this function could just be:

    auto proto = getProto();
    return proto.getDisplayName().slice(proto.getDisplayNamePrefixLength());

And it could be a method of `Schema`... I'd call it `Schema::getUnqualifiedName()`.

c++/src/capnp/schema-parser.c++
```diff
@@ -328,10 +328,33 @@ ParsedSchema ParsedSchema::getNested(kj::StringPtr nestedName) const {
...
+  }
+  return nullptr;
+}
+
+// -------------------------------------------------------------------
+
+ParsedSchema ParsedSchema::ParsedSchemaList::operator[](uint index) const {
+  return ParsedSchema(
+    parent.parser->impl->compiler.getLoader().get(
+      parent.getProto().getNestedNodes()[index].getId()),
```
kentonv, line 354: Replace `parent.getProto().getNestedNodes()` with `list`?

teqdruid: D'oh!

c++/src/capnp/schema-parser.c++
teqdruid: I don't think it can be in Schema. I think I need access to a loader to get the scope node from the scope node ID. Am I correct?

Thinking about the name, I'd prefer findUnqualifiedName() to indicate that it is a linear search rather than a constant-time property get.

kentonv: The code snippet I gave above is an alternate implementation of this function that does not require inspecting any other nodes nor doing any sort of linear lookup...

teqdruid: Sorry, for some reason I didn't see that snippet.

I thought it was bad practice to rely on the display name...? ([Do not attempt to parse](https://github.com/capnproto/capnproto/blob/88f9da4a48e57265b5e5bc642a202f5f949265bd/c%2B%2B/src/capnp/schema.capnp#L34).) I really do need the actual struct name, so I'd rather not rely on something that's not recommended.

kentonv: Well, the point of that comment is really that I don't want people relying on textual names when they shouldn't be -- especially, I don't want people expecting that these work as globally unique identifiers. But I'd say relying on _unqualified_ names is generally unlikely to be a problem, and in any case getting the name by searching the parent's children is not any better than extracting it from the display name.

teqdruid: Done

kentonv:
Thanks!

teqdruid:
Thanks for the merge!

PR #1018, opened 2020-06-21
Try to fix new MinGW and Cygwin build failures

kentonv:
Both MinGW and Cygwin builds spontaneously broke in recent weeks for different reasons -- not due to any change to Cap'n Proto, but some external factors.

MinGW seems to have picked up an errant include of `windows.h` which is shitting all over the global namespace and breaking the build.

Cygwin has started hanging after running the calculator demo, just like it used to do on Travis before we switched to GitHub Actions.

### DIFF
--- a/.github/workflows/quick-test.yml
+++ b/.github/workflows/quick-test.yml
@@ -83,6 +83,9 @@ jobs:
       - name: Build and test
         shell: cmd
         run: |
+            echo "Deleting broken Postgres install until https://github.com/actions/virtual-environments/issues/1089 is fixed..."
+            rmdir /s /q C:\PROGRA~1\POSTGR~1
+
             echo "Building Cap'n Proto with MinGW"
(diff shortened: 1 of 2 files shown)


kentonv:
Whelp... I tracked down the MinGW problem to a regression in GitHub Actions' Windows image. It appears that they installed Postgres, Postgres happily placed itself in the global include path, and Postgres' include directory contains its own `pthread.h` which is incompatible with MinGW's. Among other things, it includes `windows.h` which breaks our build, but even if it didn't, I bet it's not ABI-compatible so would cause deeper isuses.

I filed: https://github.com/actions/virtual-environments/issues/1089

kentonv:
FFS, Cygwin just _won't kill_ that damn calculator-server process no matter what I try.

kentonv:
Wow that worked, I'm just gonna merge this because it's CI-only and I want to rebase other things.

PR #1017, opened 2020-06-21
Utilities for handling network disconnects

kentonv:
Sandstorm has had hackier versions of these for a long time, but lots of Cap'n Proto applications will need these, so I went and wrote clean versions in capnp itself.

@a-robinson

PR #1021, opened 2020-06-26
Make retryOnDisconnect call the provided func asynchronously both times

a-robinson:
Rather than synchronously the first time and async the second, which
makes it tougher to deal with locks.

Minor follow-up to #1017

### DIFF
--- a/c++/src/kj/async-inl.h
+++ b/c++/src/kj/async-inl.h
@@ -1084,13 +1084,15 @@ inline PromiseForResult<Func, void> evalNow(Func&& func) {
 template <typename Func>
 struct RetryOnDisconnect_ {
   static inline PromiseForResult<Func, void> apply(Func&& func) {
-    auto promise = evalNow(func);
-    return promise.catch_([func = kj::mv(func)](kj::Exception&& e) -> PromiseForResult<Func, void> {
-      if (e.getType() == kj::Exception::Type::DISCONNECTED) {
-        return func();
-      } else {
-        return kj::mv(e);
-      }
+    return evalLater([func = kj::mv(func)]() mutable -> PromiseForResult<Func, void> {
+      auto promise = evalNow(func);
+      return promise.catch_([func = kj::mv(func)](kj::Exception&& e) mutable -> PromiseForResult<Func, void> {
+        if (e.getType() == kj::Exception::Type::DISCONNECTED) {
+          return func();
+        } else {
+          return kj::mv(e);
+        }
+      });
     });
   }
 };
@@ -1099,7 +1101,7 @@ struct RetryOnDisconnect_<Func&> {
   // Specialization for references. Needed because the syntax for capturing references in a
   // lambda is different. :(
   static inline PromiseForResult<Func, void> apply(Func& func) {
-    auto promise = evalNow(func);
+    auto promise = evalLater(func);
     return promise.catch_([&func](kj::Exception&& e) -> PromiseForResult<Func, void> {
       if (e.getType() == kj::Exception::Type::DISCONNECTED) {
         return func();
--- a/c++/src/kj/async-test.c++
+++ b/c++/src/kj/async-test.c++
@@ -1240,7 +1240,7 @@ KJ_TEST("retryOnDisconnect") {
       i++;
       return 123;
     });
-    KJ_EXPECT(i == 1);
+    KJ_EXPECT(i == 0);
     KJ_EXPECT(promise.wait(waitScope) == 123);
     KJ_EXPECT(i == 1);
   }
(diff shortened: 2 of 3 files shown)


c++/src/kj/async-inl.h
```diff
@@ -1084,7 +1084,7 @@ inline PromiseForResult<Func, void> evalNow(Func&& func) {
 template <typename Func>
 struct RetryOnDisconnect_ {
   static inline PromiseForResult<Func, void> apply(Func&& func) {
-    auto promise = evalNow(func);
+    auto promise = evalLater(func);
     return promise.catch_([func = kj::mv(func)](kj::Exception&& e) -> PromiseForResult<Func, void> {
```
a-robinson, line 1088: Looks like this may need to be made `mutable` to handle the case where ` func` is mutable... I'll change that after looking into the g++ test failures and get everything working using this internally.

kentonv:
I wonder if we need to do a similar thing for `autoReconnect`? But it's sadder there because it means any calls made before the initial run will have to be queued...

c++/src/kj/async-inl.h
a-robinson: We not only needed to make this mutable, but also needed to better handle the lifetime of `func`. We can't blindly pass it to `evalLater` since we were also `kj::mv`ing it into the catch. Fixed in the force push.

a-robinson:
The current use of `autoReconnect` is unaffected, although I suppose in theory the same concern applies.

Even here it's kind of sad that starting the async work gets delayed.

a-robinson:
Hm, when I run `./super-test.sh quick g++-8` locally it passes. Any tips on getting more info on what's happening here? Do I just have to start pushing debugging statements to my branch and looking through the CI output?

kentonv:
I'd try running under valgrind.

    valgrind ./capnp-test -f kj/async-test.c++

c++/src/kj/async-inl.h
```diff
@@ -1084,7 +1084,7 @@ inline PromiseForResult<Func, void> evalNow(Func&& func) {
 template <typename Func>
 struct RetryOnDisconnect_ {
   static inline PromiseForResult<Func, void> apply(Func&& func) {
-    auto promise = evalNow(func);
+    auto promise = evalLater(func);
```
kentonv, line 1087: Oh, the problem is probably that `evalLater()` here is capturing `func` by lvalue reference, so by the time it is executed, the reference is invalid.

You probably can't just copy `func` -- many lambdas are not copyable. I think you will have to wrap the entire body of `apply()` into an `evalLater`, rather than just change `evalNow` to `evalLater`.

a-robinson: Wait, my changes aren't showing here... I must have messed up the push, since I already handled this issue locally (as mentioned in my comment 27 minutes ago). Let's try this again.

kentonv: To clarify, `evalLater()` and many other KJ APIs that accept functions use a convention that if you give it an rvalue, it takes ownership of that rvalue by move, but if you give it an lvalue, it takes a reference, on the assumption that you're naming a long-lived object that you want to call.

But neither one of those would be correct here.

c++/src/kj/async-inl.h
```diff
@@ -1084,13 +1084,14 @@ inline PromiseForResult<Func, void> evalNow(Func&& func) {
...
   static inline PromiseForResult<Func, void> apply(Func&& func) {
-    auto promise = evalNow(func);
-    return promise.catch_([func = kj::mv(func)](kj::Exception&& e) -> PromiseForResult<Func, void> {
-      if (e.getType() == kj::Exception::Type::DISCONNECTED) {
-        return func();
-      } else {
-        return kj::mv(e);
-      }
+    return evalLater([func = kj::mv(func)]() mutable -> PromiseForResult<Func, void> {
+      return func().catch_([func = kj::mv(func)](kj::Exception&& e) mutable -> PromiseForResult<Func, void> {
```
kentonv, line 1088: I think we still want `evalNow()` around `func()` here, so that any synchronously-thrown exceptions are caught too.

kentonv: Also note that the reason I used two separate lines before, like:

    auto promise = evalNow(func);
    return promise.catch_(...)

Rather than:

    return evalNow(func).catch_(...)

Is because, technically, in the one-line version, it's undefined whether lambda passed to `catch_` is evaluated before or after `evalNow(func)`. The lambda captures `func` by move, though, so if the lambda is evaluated first, then `func` is moved away and `evalNow(func)` will not work.

I have actually observed GCC using this unexpected ordering of evaluation in the past, though Clang seems to follow intuitive order.

a-robinson: Yeah, good call. Added.

a-robinson: Interesting, thanks. I had assumed that was well-defined. Changed back to separate lines.

Issue #1022, opened 2020-06-26
Inhibition of NVRO


vlovich:
There's a general pattern of `return kj::mv`. A lot of time these are regular l-values for which the `kj::mv` would artificially inhibit `NVRO`.

vlovich:
No easy way to actually benchmark any impact here so abandoning.

PR #1023, opened 2020-06-26
Get rid of superfluous kj::mv

vlovich:
Returning an l-value automatically applies NVRO when possible and
otherwise falls back to calling the move constructor if available. I
tried to be conservative here and only remove the kj::mv that were
obviously superfluous (variables declared & returned within the
function and l-value function parameters returned).

Fixes issue #1022

### DIFF
--- a/c++/src/capnp/compat/http-over-capnp-test.c++
+++ b/c++/src/capnp/compat/http-over-capnp-test.c++
@@ -327,7 +327,7 @@ public:
   kj::Promise<kj::Own<kj::AsyncIoStream>> connect() override {
     auto result = KJ_ASSERT_NONNULL(kj::mv(stream));
     stream = nullptr;
-    return kj::mv(result);
+    return result;
   }
 
   kj::Own<kj::ConnectionReceiver> listen() override { KJ_UNIMPLEMENTED("test"); }
--- a/c++/src/capnp/compat/http-over-capnp.c++
+++ b/c++/src/capnp/compat/http-over-capnp.c++
@@ -198,6 +198,7 @@ public:
         return kj::READY_NOW;
       } else {
         // Some other error... propagate it.
+        // TODO(optimization): Will removing kj::mv still degrade to a move if NVRO isn't possible?
         return kj::mv(e);
       }
     });
@@ -296,6 +297,7 @@ public:
       // to propagate the problem back to the client. If the error came from the client, then
       // .abort() probably is a noop.
       webSocket.abort();
+      // TODO(optimization): Will removing kj::mv still degrade to a move if NVRO isn't possible?
       return kj::mv(e);
     }));
 
--- a/c++/src/capnp/compat/json-rpc.c++
+++ b/c++/src/capnp/compat/json-rpc.c++
@@ -81,7 +81,7 @@ public:
       auto sproto = context.getResultsType().getProto().getStruct();
       MessageSize size { sproto.getDataWordCount(), sproto.getPointerCount() };
       context.initResults(size);
-      return kj::mv(writePromise);
+      return writePromise;
     } else {
       auto paf = kj::newPromiseAndFulfiller<void>();
       parent.awaitedResponses.insert(callId, AwaitedResponse { context, kj::mv(paf.fulfiller) });
--- a/c++/src/capnp/compat/json.c++
+++ b/c++/src/capnp/compat/json.c++
@@ -484,7 +484,7 @@ Orphan<DynamicValue> JsonCodec::decode(
             KJ_REQUIRE(byte(x) == x, "Number in byte array is not an integer in [0, 255]");
             data[i] = x;
           }
-          return kj::mv(orphan);
+          return orphan;
         }
         default:
           KJ_FAIL_REQUIRE("Expected data value");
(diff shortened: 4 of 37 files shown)


vlovich:
Not sure how to benchmark the impact of this diff.

kentonv:
Hi @vlovich,

Unfortunately, many of your changes now invoke the copy constructor. In cases where there is no copy constructor, this results in a compiler failure, which you can see in the CI logs (did you try compiling this before submitting?). In cases where there _is_ a copy constructor, the code is likely now slower.

> Returning an l-value automatically applies NVRO when possible and
> otherwise falls back to calling the move constructor if available.

It's more strict than that. The value being returned must be a verbatim non-reference local variable. It _cannot_ be a function argument, class member, lambda capture, rvalue reference, etc.

> Not sure how to benchmark the impact of this diff.

It's almost certainly nil. Move constructors are almost always nearly free, especially with a competent optimizing compiler.

In order to accept this change, we'd need to go through all the diffs and carefully figure out which ones have caused copy constructors to be invoked, and revert those. But since the performance difference is likely negligible, I'm not sure if it's worth the effort...

kentonv:
Hmm, curious, though, that this change compiles with MSVC 2019. Did a recent C++ standard change make it possible to use automatic move semantics in more cases? In any case we still target C++14, so can't rely on that...

vlovich:
This is a guaranteed part of C++11. https://stackoverflow.com/questions/17473753/c11-return-value-optimization-or-move

vlovich:
It's possible that newer versions of MSVC  have a bug rather than this being a new part of the standard. Generally I tried to stick to converting move-only types OR types that clearly would get NVRO applied.

I disagree that a sufficiently smart compiler would convert the move into NVRO. That kind of optimization would require LTO unless the code is being inlined.

kentonv:
> This is a guaranteed part of C++11.

Yes, I'm familiar with automatic move of returned locals. I try to use it when it's obvious, although if I have to spend more than a second thinking about it, I tend to just type `kj::mv` instead.

Here's a better link:

https://en.cppreference.com/w/cpp/language/return#Notes

And here we discover why your code compiles on MSVC but not GCC/Clang:

> If expression is a (possibly parenthesized) id-expression that names a variable whose type is either
>
> a non-volatile object type or
> **a non-volatile rvalue reference to object type (since C++20)**

So it was only in C++20 that this rule began to apply when returning a variable whose type is itself an rvalue reference. There are a few diffs in your PR that depend on this new rule, hence the CI errors. MSVC always enables the latest version of the standard by default, but on GCC and Clang we explicitly choose C++14.

Note, though, that NRVO definitely _cannot_ occur in these cases:
* The named value is a parameter. (Parameters are allocated by the caller, and the caller cannot know that it should allocate the parameter directly into the return slot.)
* The named value's type does not exactly match the return type. (In this case a move is necessary to convert types.)
    * This includes when the named value is an rvalue reference.

So although "automatic move from local variables and parameters" is supported in those cases, it doesn't have any benefit other than avoiding typing `kj::mv()`. Meanwhile it has the slight disadvantage of potentially being confusing since it looks like a copy.

> That kind of optimization would require LTO unless the code is being inlined.

Move constructors, as well as destructors for movable types, are pretty much always inline. So, I think the compiler should be able to optimize these pretty well.

vlovich:
That's unfortunately not how RVO/NRVO works even if the compiler can inline the move.

https://quick-bench.com/q/GebSpQjifvI5Mn5c4zA6QcFJIPY

The std::move inhibits the ability of the compiler to eliminate the temporary. This is admittedly a pathological case where the move is enormously expensive but it does demonstrate that moving the result out can have a cost even if the compiler is optimizing as best it can. Similarly you are subject to how smart the compiler is whereas NVRO is a guaranteed optimization all conforming compilers must implement (it's unlikely you'll have a compiler that can intelligently elide the std::move but doesn't apply NVRO whereas the reverse is clearly possible).

Additionally, even if the compiler is able to generate identical code for `mv` vs not, that does require optimizations to be enabled so there can be an increased cost due to these in debug builds:

https://godbolt.org/z/Brz6BE

That being said that's also a corner case as it seems like only GCC is following the standard & applying NVRO even in debug builds. Both clang & MSVC appear to generate equivalently bad code at O0 for the NVRO path which is technically incorrect behavior since NVRO has to be a standard-conformance thing due to the violation of the as-if rule.

Anyway, I'm happy to update the diff with just the obvious NVRO cases if that's the ask. If the feedback is that you don't want it even for the obvious cases then I'm happy to abandon it. Let me know.

kentonv:
If you could submit a version that is C++14-compatible and doesn't add any TODO comments, I'll accept it. I'm just saying I am skeptical that this will have any measurable performance impact in practice. I don't find your benchmark compelling because it uses a move constructor that is intentionally slow, while real-world move constructors are generally trivial.

vlovich:
There's no easy way to benchmark any of this so going to abandon it for now.

kentonv:
For the record, I would have accepted a change without benchmark proof as long as the change was C++14-compatible.

vlovich:
Oh I know. My concern stemmed from innocuous places where it should have been safe to return complaining that the copy was deleted. That meant there was some subtlety that could cause it to be a copy which made me nervous (& without benchmarks I could easily introduce a regression without knowing it). Likely the issue was places where there was anonymous returns & named returns so the compiler ended up not being able to apply RVO/NRVO but that in theory should have still been a move.

Anyway, I was looking at the time I spent on it & how much more time I'd need to spend to feel good about submitting a change like that. The cost-benefit didn't seem there given there's no benchmark coverage I can easily rely on & real risk of making a mistake.

kentonv:
Makes sense!

PR #1025, opened 2020-06-29
Don't use Ninja on Windows.

kentonv:
It turns out Ninja builds with some version of GCC (MinGW, probably), NOT with MSVC, but we specifically want to test with MSVC.

### DIFF
--- a/.github/workflows/quick-test.yml
+++ b/.github/workflows/quick-test.yml
@@ -48,13 +48,16 @@ jobs:
             target: 'Visual Studio 15 2017'
             arch:
             install_deps: false
+            cmake_args:
           - os: windows-latest
             target: 'Visual Studio 16 2019'
             arch: -A x64
             install_deps: true
+            cmake_args: -DCMAKE_PREFIX_PATH="%CONDA_PREFIX%"
     steps:
       - uses: actions/checkout@v2
       - name: Include $CONDA in $PATH
+        if: ${{ matrix.install_deps }}
         run: |
           echo "::add-path::$Env:CONDA\condabin"
       - name: Install dependencies via Conda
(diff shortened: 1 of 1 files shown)


kentonv:
So now we're back to the problem where the 32-bit build can't compile against OpenSSL and the 64-bit build segfaults randomly. Time to download a 20GB virtual machine image...

harrishancock:
D'oh! There must be a way to make Ninja use VS, though I suppose that wouldn't actually solve the problems...

kentonv:
WTF why does it skip one but not the other!

Screenshot from 2020-07-13 11-48-04


Screenshot from 2020-07-13 11-47-07

.github/workflows/quick-test.yml
```diff
@@ -45,27 +45,33 @@ jobs:
         os: ['windows-2016', 'windows-latest']
         include:
           - os: windows-2016
-            target: 'Ninja'
+            target: 'Visual Studio 15 2017'
             arch:
```
ScatteredRay, line 49: Missing arch `-A x64`

kentonv: I think this was intentional. I tried it both ways. See my earlier comments in this thread. I was hoping to do a 32-bit build in order to get 32-bit Windows coverage without having to further expand the build matrix. But, I couldn't figure out how to get OpenSSL. With the 64-bit build, meanwhile, I was encountering random segfaults (which don't happen under VS 2019).

nyanpasu64:
> It turns out Ninja builds with some version of GCC (MinGW, probably), NOT with MSVC

This isn't strictly true. When I load a CMake project (like Cap'n Proto) in Qt Creator (even using `-GNinja`), it lets you pick any "kit" (compiler, debugger, etc.) you want. Qt Creator then passes `-DCMAKE_C_COMPILER:STRING=... -DCMAKE_CXX_COMPILER:STRING=` into the CMake invocation. So it's possible to use Ninja to build using MSVC.

Another way (which works on Windows and probably other OSes too) is to define the `CC` and `CXX` environment variables when running CMake. I use it in my [CMake project's CI](https://gitlab.com/exotracker/exotracker-cpp/-/blob/dev/.appveyor.yml#L44).

I don't know if Ninja or Visual Studio projects is preferable (builds faster, more reliable, tested, etc). Note that the VS2019 IDE now has native support for CMakeLists projects, which is accomplished behind the scenes by calling CMake with the Ninja generator, *not* creating temporary .sln files. And writing a CMakeLists.txt which can switch debug/release flags *after* generation (eg. in Visual Studio projects, or Ninja Multi-Config) requires using generator expressions rather than branching on CMake variables, and I'm honestly not proficient with generator expressions.

ScatteredRay:
Yeah, CMake with the ninja generator just does best-effort to find the compiler environment... When run in a VS command prompt, it'll generally build with the VS compiler, I think the problem is on GHA runners, they may have multiple compilers in the environment.

kentonv:
> > It turns out Ninja builds with some version of GCC (MinGW, probably), NOT with MSVC
>
> This isn't strictly true.

I meant specifically in the context of our CI. It's true that an alternate solution could have been to fix the Ninja config.

Anyway, closing this PR as #1210 fixed the problem.

PR #1026, opened 2020-06-30
Fix Windows version friendliness with external build

vlovich:
External build systems may override the minimum windows version when
building to be even newer than Vista. Fix the conflict so that if the
macro is a baked-in parameter for the compiler invocation that
everything still builds.

### DIFF
--- a/c++/src/capnp/compiler/capnp.c++
+++ b/c++/src/capnp/compiler/capnp.c++
@@ -23,6 +23,10 @@
 #define _GNU_SOURCE
 #endif
 
+#if _WIN32
+#include <kj/win32-api-version.h>
+#endif
+
 #include "lexer.h"
 #include "parser.h"
 #include "compiler.h"
@@ -50,10 +54,8 @@
 
 #if _WIN32
 #include <process.h>
-#define WIN32_LEAN_AND_MEAN  // ::eyeroll::
 #include <windows.h>
 #include <kj/windows-sanity.h>
-#undef VOID
 #undef CONST
 #else
 #include <sys/wait.h>
--- a/c++/src/kj/async-win32.h
+++ b/c++/src/kj/async-win32.h
@@ -25,18 +25,16 @@
 #error "This file is Windows-specific. On Unix, include async-unix.h instead."
 #endif
 
+// Include windows.h as lean as possible. (If you need more of the Windows API for your app,
+// #include windows.h yourself before including this header.)
+#include "win32-api-version.h"
+
 #include "async.h"
 #include "timer.h"
 #include "io.h"
 #include <atomic>
 #include <inttypes.h>
 
-// Include windows.h as lean as possible. (If you need more of the Windows API for your app,
-// #include windows.h yourself before including this header.)
-#define WIN32_LEAN_AND_MEAN 1
-#define NOSERVICE 1
-#define NOMCX 1
-#define NOIME 1
 #include <windows.h>
 #include "windows-sanity.h"
 
--- a/c++/src/kj/compat/tls-test.c++
+++ b/c++/src/kj/compat/tls-test.c++
@@ -21,6 +21,10 @@
 
 #if KJ_HAS_OPENSSL
 
+#if _WIN32
+#include <kj/win32-api-version.h>
+#endif
+
 #include "tls.h"
 #include <kj/test.h>
 #include <kj/async-io.h>
@@ -28,7 +32,6 @@
 #include <openssl/opensslv.h>
 
 #if _WIN32
-#define WIN32_LEAN_AND_MEAN
 #include <winsock2.h>
 #include <kj/windows-sanity.h>
 #else
--- a/c++/src/capnp/compiler/capnpc-c++.c++
+++ b/c++/src/capnp/compiler/capnpc-c++.c++
@@ -21,6 +21,10 @@
 
 // This program is a code generator plugin for `capnp compile` which generates C++ code.
 
+#if _WIN32
+#include <kj/win32-api-version.h>
+#endif
+
 #include <capnp/schema.capnp.h>
 #include "../serialize.h"
 #include <kj/debug.h>
@@ -40,10 +44,8 @@
 #include <capnp/stream.capnp.h>
 
 #if _WIN32
-#define WIN32_LEAN_AND_MEAN  // ::eyeroll::
 #include <windows.h>
 #include <kj/windows-sanity.h>
-#undef VOID
 #undef CONST
 #else
 #include <sys/time.h>
--- a/c++/src/capnp/compiler/parser.c++
+++ b/c++/src/capnp/compiler/parser.c++
@@ -19,6 +19,10 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
+#if _WIN32
+#include <kj/win32-api-version.h>
+#endif
+
 #include "parser.h"
 #include "type-id.h"
 #include <capnp/dynamic.h>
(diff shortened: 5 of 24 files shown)


c++/src/capnp/compiler/capnp.c++
```diff
@@ -50,7 +50,9 @@
 
 #if _WIN32
 #include <process.h>
+#ifndef WIN32_LEAN_AND_MEAN
```
kentonv, line 53: I understand `WINVER` and `_WIN32_WINNT`, but do we expect a build system might define `WIN32_LEAN_AND_MEAN`, too? This seems wrong, since this macro actually hides APIs. What if some module in KJ needs to use an API that is not part of the "lean and mean" set?

vlovich: For better or worse it's defined globally for all Windows builds @ Oculus. If some module in KJ actually needed that API there are 3 options:
1. Put code into kj to `#ifndef/#undef` up-front
2. Make it the responsibility of the external build system to exclude it for that 1 file

2 would be my preference and generally compilers will always take the last definition so that seems like a tractable solution assuming it's a built-time issue. If you wanted to be extra clear with the compiler error message you could `#ifndef WIN32_LEAN_AND_MEAN`/`#error`.

That being said, I *think* the conflict really comes from the few spots where `WIN32_LEAN_AND_MEAN` is defined as `1` instead of just defined at all. It would solve my problem if we removed those to avoid the preprocessor conflict. That being said, it wouldn't be unreasonable for a build maintainer in the general case to have defined `WIN32_LEAN_AND_MEAN` to 1 (clearly even within kj it was done both ways).

I don't have any strong opinions here. Let me know your preference.

kentonv: I would argue that it's incorrect to define this globally when building other people's code.

That said, if all of KJ builds correctly with this defined, then great. How about we move the whole ifdef/define into `win32-api-version.h`, and make sure to include that in all the places that currently define WIN32_LEAN_AND_MEAN? That would reduce the amount of boilerplate overall.

vlovich: Will do. I don't disagree with your point. I think it will be challenging though to figure out the best way to enable that policy as I'm sure there are good reasons it was done. The build team here is super sharp. The build macros do know if the code is third party or not so maybe keying off of that to add `/UWIN32_LEAN_AND_MEAN` might be a generic way to do that. On the other hand I suspect it could be a policy thing to reduce build capacity given how limited our Windows build farm is.

I'll apply your recommendations to this diff hopefully today.

c++/src/kj/async-win32.h
```diff
@@ -33,7 +33,7 @@
 
 // Include windows.h as lean as possible. (If you need more of the Windows API for your app,
 // #include windows.h yourself before including this header.)
-#define WIN32_LEAN_AND_MEAN 1
+#include "win32-api-version.h"
 #define NOSERVICE 1
```
kentonv, line 37: Maybe we should throw these into win32-api-version.h as well?

vlovich: NOMCX & NOIME too?

vlovich: Found a few more similar symbols since it seems like maybe you would prefer it all be centralized.

c++/src/kj/compat/tls-test.c++
```diff
@@ -28,7 +28,7 @@
 #include <openssl/opensslv.h>
 
 #if _WIN32
-#define WIN32_LEAN_AND_MEAN
+#include "win32-api-version.h"
```
kentonv, line 31: This one needs to be in `<>` -- this is breaking the CI build:

D:\a\capnproto\capnproto\c++\src\kj\compat\tls-test.c++(31,10): fatal error C1083: Cannot open include file: 'win32-api-version.h': No such file or directory [D:\a\capnproto\capnproto\build-output\src\kj\kj-heavy-tests.vcxproj]

kentonv:
MinGW is failing claiming that _WIN32_WINNT is somehow already defined to be older than vista:

```
2020-07-09T03:56:09.8239817Z In file included from D:\a\capnproto\capnproto\c++\src\kj\debug.c++:32:
2020-07-09T03:56:09.8243449Z D:\a\capnproto\capnproto\c++\src\kj\win32-api-version.h:34:2: error: #error "_WIN32_WINNT defined but older than Vista"
2020-07-09T03:56:09.8251957Z  #error "_WIN32_WINNT defined but older than Vista"
2020-07-09T03:56:09.8255304Z   ^~~~~
```

This is odd. I wonder if the MinGW C library headers are defining `_WIN32_WINNT` themselves. We might need to go through and make sure `win32_api-version.h` is the first include in every file.

Sorry this is dragging on. I didn't notice the build failures before.

vlovich:
Yeah, I didn't notice it either. Hopefully this latest set of diffs resolves it.

vlovich:
@kentonv I'm having trouble finishing off the diff. Moving the headers above now fails because CONST is improperly defined somewhere. Would appreciate your advice on what I need to do to fix this.

vlovich:
Ok. Maybe I got it working? @kentonv is the MacOS failure a failure caused by this diff or a latent issue?

```
[ FAIL ] kj/async-unix-test.c++:634: legacy test: AsyncUnixTest/InterruptedTimer (2254281 μs)
stack: 103783c9b 1039bcf93 1045a4a7c 1045cbfca 1045a33ab 1045a3121 1046043ab 1045cbfca 104601d1e 1045a298d 7fff704accc8
[ TEST ] kj/async-unix-test.c++:673: legacy test: AsyncUnixTest/Wake
    kj::(anonymous namespace)::TestCase634::run() (in capnp-test) + 467: returning here
    kj::_::runCatchingExceptions(kj::_::Runnable&) (in libkj-0.9-dev.dylib) + 26: returning here
    kj::TestRunner::run() (in libkj-test-0.9-dev.dylib) + 635: returning here
    kj::Function<kj::MainBuilder::Validity ()>::Impl<kj::_::BoundMethod<kj::TestRunner&, kj::TestRunner::getMain()::'lambda3'(auto&, auto&&...), kj::TestRunner::getMain()::'lambda4'(auto&, auto&&...)> >::operator()() (in libkj-test-0.9-dev.dylib) + 17: returning here
    kj::MainBuilder::MainImpl::operator()(kj::StringPtr, kj::ArrayPtr<kj::StringPtr const>) (in libkj-0.9-dev.dylib) + 3499: returning here
    kj::_::runCatchingExceptions(kj::_::Runnable&) (in libkj-0.9-dev.dylib) + 26: returning here
    kj::runMainAndExit(kj::ProcessContext&, kj::Function<void (kj::StringPtr, kj::ArrayPtr<kj::StringPtr const>)>&&, int, char**) (in libkj-0.9-dev.dylib) + 606: returning here
    main (in libkj-test-0.9-dev.dylib) + 109: returning here
    start (in libdyld.dylib) + 0: returning here
```

kentonv:
So, I really hate to bounce this back again, but... it looks like this now breaks a KJ style rule in a bunch of files:

https://github.com/capnproto/capnproto/blob/master/style-guide.md#file-templates

> Note that in both the source and test files, you should always include the corresponding header first, in order to ensure that it is self-contained (does not secretly require including some other header before it).

Clearly `win32-api-version.h` itself has to break this rule, due to MinGW's overriding _WIN32_WINNT in its own headers if we don't set it first. However, in a lot of cases you're also moving up a bunch of other headers that happened to be in the same `#if _WIN32`/`#else` block. We need to keep those where the were, and only move up the include of `win32-api-version.h` (adding a new `#if _WIN32` block around it, of course).

kentonv:
Thanks, looks good now!

PR #1027, opened 2020-06-30
Fix ignored qualifier

vlovich:
clang-cl rightly points out that the const in its current spot doesn't
do anything as it's applying the const to the function type. Instead
make the pointer itself const which is what's intended.

### DIFF
--- a/c++/src/kj/time.c++
+++ b/c++/src/kj/time.c++
@@ -72,7 +72,7 @@ class Win32PreciseClock: public Clock {
   typedef VOID WINAPI GetSystemTimePreciseAsFileTimeFunc(LPFILETIME);
 public:
   Date now() const override {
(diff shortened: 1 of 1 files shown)


c++/src/kj/time.c++
```diff
@@ -86,7 +86,7 @@ public:
   }
 
 private:
-  static GetSystemTimePreciseAsFileTimeFunc* getGetSystemTimePreciseAsFileTime() {
+  static GetSystemTimePreciseAsFileTimeFunc* const getGetSystemTimePreciseAsFileTime() {
```
kentonv, line 89: I don't understand. According to the diff, there wasn't any `const` qualifier present previously.

Did you maybe intend to modify line 75 instead?

vlovich: I did. My bad. I copied the change over by hand. Let me update this.

PR #1028, opened 2020-06-30
Fix unqualified friend declaration

vlovich:
clang-cl warns about this as a Microsoft extension:
```
async-inl.h(922,16): warning: unqualified friend declaration referring
to type outside of the nearest enclosing namespace is a Microsoft
extension; add a nested name specifier [-Wmicrosoft-unq
ualified-friend]
  friend class WaitScope;
               ^
               ::kj::
```


kentonv:
It looks like this friend declaration actually isn't necessary. Clang and GCC interpret the line correctly, which means friendship was being granted to a class that doesn't exist. So I think we can just remove the line...

kentonv:
Thanks!

PR #1029, opened 2020-06-30
Add lock timeout support to MutexGuarded

ObsidianMinor:
Not supported on Windows since slim rw locks don't support timeouts.

### DIFF
--- a/c++/src/kj/mutex-test.c++
+++ b/c++/src/kj/mutex-test.c++
@@ -59,7 +59,7 @@ TEST(Mutex, MutexGuarded) {
     EXPECT_EQ(123u, *lock);
     EXPECT_EQ(123u, value.getAlreadyLockedExclusive());
 
-#if !(_WIN32 || __CYGWIN__)
+#if KJ_USE_FUTEX
     EXPECT_TRUE(value.lockExclusiveWithTimeout(MILLISECONDS * 50) == nullptr);
     EXPECT_TRUE(value.lockSharedWithTimeout(MILLISECONDS * 50) == nullptr);
 #endif
@@ -76,7 +76,7 @@ TEST(Mutex, MutexGuarded) {
     auto earlyRelease = kj::mv(lock);
   }
 
-#if !(_WIN32 || __CYGWIN__)
+#if KJ_USE_FUTEX
   EXPECT_TRUE(value.lockExclusiveWithTimeout(MILLISECONDS * 50) != nullptr);
   EXPECT_TRUE(value.lockSharedWithTimeout(MILLISECONDS * 50) != nullptr);
 #endif
--- a/c++/src/kj/mutex-test.c++
+++ b/c++/src/kj/mutex-test.c++
@@ -59,6 +59,9 @@ TEST(Mutex, MutexGuarded) {
     EXPECT_EQ(123u, *lock);
     EXPECT_EQ(123u, value.getAlreadyLockedExclusive());
 
+    EXPECT_TRUE(value.lockExclusiveWithTimeout(MILLISECONDS * 50) == nullptr);
+    EXPECT_TRUE(value.lockSharedWithTimeout(MILLISECONDS * 50) == nullptr);
+
     Thread thread([&]() {
       Locked<uint> threadLock = value.lockExclusive();
       EXPECT_EQ(456u, *threadLock);
@@ -71,6 +74,8 @@ TEST(Mutex, MutexGuarded) {
     auto earlyRelease = kj::mv(lock);
   }
 
+  EXPECT_TRUE(value.lockExclusiveWithTimeout(MILLISECONDS * 50) != nullptr);
+  EXPECT_TRUE(value.lockSharedWithTimeout(MILLISECONDS * 50) != nullptr);
   EXPECT_EQ(789u, *value.lockExclusive());
 
   {
--- a/c++/src/kj/mutex-test.c++
+++ b/c++/src/kj/mutex-test.c++
@@ -59,8 +59,10 @@ TEST(Mutex, MutexGuarded) {
     EXPECT_EQ(123u, *lock);
     EXPECT_EQ(123u, value.getAlreadyLockedExclusive());
 
+#if !(_WIN32 || __CYGWIN__)
     EXPECT_TRUE(value.lockExclusiveWithTimeout(MILLISECONDS * 50) == nullptr);
     EXPECT_TRUE(value.lockSharedWithTimeout(MILLISECONDS * 50) == nullptr);
+#endif
 
     Thread thread([&]() {
       Locked<uint> threadLock = value.lockExclusive();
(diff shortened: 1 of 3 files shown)


c++/src/kj/mutex.c++
```diff
@@ -172,6 +172,65 @@ void Mutex::lock(Exclusivity exclusivity) {
...
+          break;
+        }
+
+        // The mutex is exclusively locked by another thread.  Since we incremented the counter
+        // already, we just have to wait for it to be unlocked.
+        auto result = syscall(SYS_futex, &futex, FUTEX_WAIT_PRIVATE, state, &spec, nullptr, 0);
+        if (result < 0) {
+          // If we timeout though, we need to signal that we're not waiting anymore.
+          if (errno == ETIMEDOUT) {
+            state = __atomic_sub_fetch(&futex, 1, __ATOMIC_ACQUIRE);
```
ObsidianMinor, line 222: Is this the correct ordering? I'm still not 100% sure on what orderings should be used when.

ObsidianMinor:
So apparently `pthread_rwlock_timedrdlock` and `pthread_rwlock_timedwrlock` just doesn't exist on MacOS? Do we have to kill timeout support on that platform too? Feels weird to only support one platform for lock timeouts...

ObsidianMinor:
I believe this is ready for review @kentonv

c++/src/kj/mutex.c++
```diff
@@ -172,6 +172,65 @@ void Mutex::lock(Exclusivity exclusivity) {
...
+          }
+
+          state |= EXCLUSIVE_REQUESTED;
+        }
+
+        auto result = syscall(SYS_futex, &futex, FUTEX_WAIT_PRIVATE, state, &spec, nullptr, 0);
+        if (result < 0) {
+          // We timed out, remove exclusive requested flag.
+          if (errno == ETIMEDOUT) {
+            __atomic_fetch_and(&futex, ~EXCLUSIVE_REQUESTED, __ATOMIC_RELEASE);
```
kentonv, line 202: Hmm, I don't think you can just remove the exclusive flag here, because there could be more than one thread waiting for an exclusive lock. I think it's safe to leave it as-is even if no one is waiting, as the only down-side is that the next thread to unlock the mutex might perform a redundant futex wakeup.

c++/src/kj/mutex.c++
kentonv: You can use `__ATOMIC_RELAXED` here, because the calling thread hasn't touched and will not touch the protected object. Generally, `__ATOMIC_ACQUIRE` is needed if the calling thread will be accessing the protected object after the call (but was not permitted to do so before), and `__ATOMIC_RELEASE` is needed if it accessed the protected object before the call (but won't be permitted to after).

c++/src/kj/mutex.c++
```diff
@@ -172,6 +172,65 @@ void Mutex::lock(Exclusivity exclusivity) {
   }
 }
 
+bool Mutex::lockWithTimeout(Exclusivity exclusivity, Duration timeout) {
```
kentonv, line 175: Instead of introducing a new method, could we extend the existing `lock()` method with a `Maybe<Duration> timeout` parameter? It looks like most of the (rather complicated) logic is repeated, and the extra logic to handle timeouts shouldn't cost much when the timeout is null.

c++/src/kj/mutex.c++
ObsidianMinor: Yeah, that's what I though. Good to know it'll work the other way.

c++/src/kj/mutex.c++
```diff
@@ -151,7 +157,14 @@ void Mutex::lock(Exclusivity exclusivity) {
           state |= EXCLUSIVE_REQUESTED;
         }
 
-        syscall(SYS_futex, &futex, FUTEX_WAIT_PRIVATE, state, nullptr, nullptr, 0);
+        auto result = syscall(SYS_futex, &futex, FUTEX_WAIT_PRIVATE, state, specp, nullptr, 0);
+        if (result < 0) {
+          // We timed out, we can't remove the exclusive request flag (since others might be waiting)
```
kentonv, line 162: Nitpick: Seems like this comment belongs inside the below `if` block.

c++/src/kj/mutex.c++
```diff
@@ -164,12 +177,20 @@ void Mutex::lock(Exclusivity exclusivity) {
 
         // The mutex is exclusively locked by another thread.  Since we incremented the counter
         // already, we just have to wait for it to be unlocked.
-        syscall(SYS_futex, &futex, FUTEX_WAIT_PRIVATE, state, nullptr, nullptr, 0);
+        auto result = syscall(SYS_futex, &futex, FUTEX_WAIT_PRIVATE, state, specp, nullptr, 0);
+        if (result < 0) {
+          // If we timeout though, we need to signal that we're not waiting anymore.
+          if (errno == ETIMEDOUT) {
+            state = __atomic_sub_fetch(&futex, 1, __ATOMIC_RELAXED);
```
kentonv, line 184: Hmmmmmm... what if, between timing out, and calling this to decrement the reader count, the mutex was unlocked, and also this decrement sets the reader count to 0, and also another thread in the meantime was waiting for an exclusive lock?

I think we need to perform the same sequence as in unlock() after the `__atomic_sub_fetch`, to make sure we send a wake signal if needed.

c++/src/kj/mutex-test.c++
```diff
@@ -71,8 +76,14 @@ TEST(Mutex, MutexGuarded) {
...
   }
 
+#if KJ_USE_FUTEX
+  EXPECT_TRUE(value.lockExclusiveWithTimeout(MILLISECONDS * 50) != nullptr);
+  EXPECT_TRUE(value.lockSharedWithTimeout(MILLISECONDS * 50) != nullptr);
+#endif
+
   EXPECT_EQ(789u, *value.lockExclusive());
 
+
```
kentonv, line 86: Nitpick: Random added newline...

c++/src/kj/mutex-test.c++
```diff
@@ -59,6 +59,11 @@ TEST(Mutex, MutexGuarded) {
     EXPECT_EQ(123u, *lock);
     EXPECT_EQ(123u, value.getAlreadyLockedExclusive());
 
+#if KJ_USE_FUTEX
+    EXPECT_TRUE(value.lockExclusiveWithTimeout(MILLISECONDS * 50) == nullptr);
```
kentonv, line 63: We should probably expand these tests a bit:

* Verify that, after a timeout, the expected amount of time has actually passed (wall time clock has advanced by at least the duration indicated).
* For the cases where the lock is acquired, actually do something with the lock.
* Maybe try to test for the bug where if two threads tried to take exclusive locks at the same time and one thread timed out, the other thread's exclusive lock would never complete due to having unset the flag?
* I guess the other bug (failing to send wakeup after reader timeout) is probably too hard to test (since the timing window is so narrow), so you can skip that.

PR #1035, opened 2020-07-21
Add Response autoEncode

xtuc:


kentonv:
Auto-encoding in Cloudflare Workers is implemented in the workers runtime, not KJ, so it doesn't make sense to add this here.

PR #1036, opened 2020-07-27
Improve clang cl support

vlovich:
clang-cl defines a really old version of _MSC_VER which trips up some of
the version detection and also would trigger unnecessary workarounds
that clang is able to handle.

Additionally remove a MSVC 2015 workaround since 2017 is the minimum
version anyway.

Audit all _MSC_VER usage throughout codebase to try to handle it
correctly. There are still places that access _MSC_VER when they
probably mean to check if building for Windows - those places are
untouched.

Not sure how to integrate this into automated build support. Will be testing it
on my own build farm but I don't run the capnp unit tests so there could be a potential
gap there.

### DIFF
--- a/c++/src/capnp/compiler/capnpc-c++.c++
+++ b/c++/src/capnp/compiler/capnpc-c++.c++
@@ -1785,7 +1785,7 @@ private:
             "      ::capnp::bounded<", offset, ">() * ::capnp::POINTERS), kj::mv(value));\n"
             "}\n",
             COND(type.hasDisambiguatedTemplate(),
-                "#ifndef _MSC_VER\n"
+                "#if !defined(_MSC_VER) || defined(__clang__)\n"
                 "// Excluded under MSVC because bugs may make it unable to compile this method.\n"),
             templateContext.allDecls(),
             "inline ::capnp::Orphan<", type, "> ", scope, "Builder::disown", titleCase, "() {\n",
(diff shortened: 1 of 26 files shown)


c++/src/capnp/compiler/capnpc-c++.c++
```diff
@@ -1785,15 +1785,15 @@ private:
             "      ::capnp::bounded<", offset, ">() * ::capnp::POINTERS), kj::mv(value));\n"
             "}\n",
             COND(type.hasDisambiguatedTemplate(),
-                "#ifndef _MSC_VER\n"
+                "#if !defined(_MSC_VER) || defined(__clang__)\n"
```
harrishancock, line 1788: Since this touches code generation, you'll want to run `./regenerate-bootstraps.sh` in capnproto/c++, then check in the regenerated code in a second commit.

kentonv: Hmm we have a unit test that should have failed if the bootstraps weren't regenerated. Is it possible this line doesn't appear in any of the bootstraps?

vlovich: I did also change all the generated code by hand. However, I did notice that when I reran `regenerate-bootstrap.sh`  I got a bunch of changes so maybe there's a bug in that somewhere?

```
-  ::capnp::Capability::Server::DispatchCallResult dispatchCall(
-      uint64_t interfaceId, uint16_t methodId,
+  ::kj::Promise<void> dispatchCall(uint64_t interfaceId, uint16_t methodId,
```

This is in addition to all the CAPNP_VERSION checks that switch from `!= 7000` to `!= 9000`

kentonv:
Thanks! It's exciting that Clang is finally a viable replacement for cl.exe.

kentonv:
BTW if you want to make sure we don't randomly break clang-cl, it'd help to add an Action that tests it.

vlovich:
Happy to do that. How would I do that?

kentonv:
@vlovich Extend https://github.com/capnproto/capnproto/blob/master/.github/workflows/quick-test.yml with a new job that installs clang-cl and builds with it.

I don't know much about the Windows ecosystem but GitHub's Windows runner has a package manager called "chocolatey" pre-installed and it looks like they have [a clang-cl package](https://chocolatey.org/packages/llvm).

PR #1037, opened 2020-07-29
Add option to reject decoding of JSON messages with unrecognized fields

mologie:
Hi kentonv,

I would like to propose the attached change to libcapnp-json. It adds an opt-in flag to make JsonCodec's decoding fail when encountering unknown fields.

My use case, which I believe requires this change, involves a tool for preserving/archiving an inbound JSON data stream in Cap'n Proto's wire format. A failure scenario exists where a client is changed to provide data not yet known to the archival tool (and thus not yet approved by the archive owner). Given my philosophy of "fail early", I would like the tool to reject messages unless it is sure that no data is lost in the conversion process. Forward compatibility is of insignificant importance compared to a risk of data loss or maintenance cost of storing inbound data in multiple formats.

An alternative to this change would involve checking inbound data using tools such as json-schema prior to processing it with libcapnp. However, this would require that I maintain two schema files and formats with basically the same logical content, which is not an attractive option.

Thank you for your consideration and time to look into this change!

Best

Oliver

### DIFF
--- a/c++/src/capnp/compat/json.c++
+++ b/c++/src/capnp/compat/json.c++
@@ -37,6 +37,7 @@ struct JsonCodec::Impl {
   bool prettyPrint = false;
   HasMode hasMode = HasMode::NON_NULL;
   size_t maxNestingDepth = 64;
+  bool ignoreUnrecognizedFields = true;
 
   kj::HashMap<Type, HandlerBase*> typeHandlers;
   kj::HashMap<StructSchema::Field, HandlerBase*> fieldHandlers;
@@ -184,6 +185,8 @@ void JsonCodec::setMaxNestingDepth(size_t maxNestingDepth) {
 }
 
 void JsonCodec::setHasMode(HasMode mode) { impl->hasMode = mode; }
+
+void JsonCodec::setIgnoreUnrecognizedFields(bool ignore) { impl->ignoreUnrecognizedFields = ignore; }
 
 kj::String JsonCodec::encode(DynamicValue::Reader value, Type type) const {
   MallocMessageBuilder message;
@@ -386,7 +389,7 @@ void JsonCodec::decodeObject(JsonValue::Reader input, StructSchema type, Orphana
     KJ_IF_MAYBE(fieldSchema, type.findFieldByName(field.getName())) {
       decodeField(*fieldSchema, field.getValue(), orphanage, output);
     } else {
-      // Unknown json fields are ignored to allow schema evolution
+      KJ_REQUIRE(impl->ignoreUnrecognizedFields, "Unrecognized field.", field.getName());
     }
   }
 }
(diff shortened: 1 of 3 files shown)


c++/src/capnp/compat/json.h
```diff
@@ -85,6 +83,11 @@ class JsonCodec {
   // setHasMode(HasMode::NON_DEFAULT) to specify that default-valued primitive fields should be
   // omitted as well.
 
+  void setIgnoreUnrecognizedFields(bool ignore);
```
kentonv, line 86: Hmm default-true options can be a bit confusing. What do you think of inverting this to `setRejectUnknownFields()`?

Also, maybe bikeshedding, but there is precedent (going back to protobuf) for calling these "unknown fields" rather than "unrecognized fields".

c++/src/capnp/compat/json.c++
```diff
@@ -386,7 +389,7 @@ void JsonCodec::decodeObject(JsonValue::Reader input, StructSchema type, Orphana
     KJ_IF_MAYBE(fieldSchema, type.findFieldByName(field.getName())) {
       decodeField(*fieldSchema, field.getValue(), orphanage, output);
     } else {
-      // Unknown json fields are ignored to allow schema evolution
+      KJ_REQUIRE(impl->ignoreUnrecognizedFields, "Unrecognized field.", field.getName());
```
kentonv, line 392: This is super-nitpicky but I'd remove the `.` from this string. It's not a complete sentence anyway, and KJ error output is going to add a `;` after it which would look weird.

kentonv:
Other than those nitpicks, I'm happy to accept this this feature, thanks!

c++/src/capnp/compat/json.c++
mologie: Guilty as charged, done.

c++/src/capnp/compat/json.h
mologie: Yes, I agree with both points, done.

The reason why this was named `setAccept...` in the first place was due to building it inside-out, where I avoided a double negative at the actual check ("require that not reject is true" sounded wrong). Naturally, I ended up with a boolean that defaulted to true, and hence the option as wrapper for it. From an API perspective, I agree that the default value false/0 should take precedence to avoid confusion.

Regarding unrecognized vs. unknown: I followed the wording in a comment of the original author. However, *unknown* seems to be the preferred term in most of Cap'n Protos codebase, and given its precedent in Protobuf, switching out the adverb seems to be the more natural solution here.

mologie:
Awesome, I have updated the API, exception, and tests like discussed. This was a rather small change with a big impact, and I'm glad that you do not object to (ab)using Cap'n Proto as schema checker ;-). It is rather ironic that Cap'n Proto's JSON interop features make it a pretty good (statically and dynamically typed!) JSON library.

kentonv:
Thanks!

PR #1038, opened 2020-07-31
Remove obsolete strtod includes from json.c++

mologie:
I noticed that json.c++ references unneeded headers for strtod, which is no longer used directly. This merge request removes the obsolete includes.


kentonv:
Thanks!

PR #1045, opened 2020-08-13
Check if fufiller is waiting before asserting that the fulfiller exists

ObsidianMinor:
This can prevent issues where `drain` is called and the promise is destroyed, preventing further `drain`s from being done.

### DIFF
--- a/c++/src/kj/async.c++
+++ b/c++/src/kj/async.c++
@@ -290,7 +290,11 @@ kj::String TaskSet::trace() {
 }
 
 Promise<void> TaskSet::onEmpty() {
-  KJ_REQUIRE(emptyFulfiller == nullptr, "onEmpty() can only be called once at a time");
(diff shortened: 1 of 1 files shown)


c++/src/kj/async.c++
```diff
@@ -290,7 +290,11 @@ kj::String TaskSet::trace() {
 }
 
 Promise<void> TaskSet::onEmpty() {
-  KJ_REQUIRE(emptyFulfiller == nullptr, "onEmpty() can only be called once at a time");
+  KJ_IF_MAYBE(fulfiller, emptyFulfiller) {
+    if (fulfiller->get()->isWaiting()) {
+      KJ_REQUIRE(emptyFulfiller == nullptr, "onEmpty() can only be called once at a time");
```
kentonv, line 295: Since we already know the fulfiller is non-null, I'd change this to:

    KJ_FAIL_REQUIRE("onEmpty() can only be called once at a time");

PR #1048, opened 2020-08-23
Fix JsonCodec::setRejectUnknownFields() to work with annotated types.

kentonv:
/cc @mologie

### DIFF
--- a/c++/src/capnp/compat/json.c++
+++ b/c++/src/capnp/compat/json.c++
@@ -1316,7 +1316,8 @@ private:
 
       KJ_UNREACHABLE;
     } else {
-      // Ignore undefined field.
+      // Ignore undefined field -- unless the flag is set to reject them.
+      KJ_REQUIRE(!codec.impl->rejectUnknownFields, "Unknown field", name);
       return true;
     }
   }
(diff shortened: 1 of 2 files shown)


c++/src/capnp/compat/json.c++
```diff
@@ -1317,6 +1317,7 @@ private:
       KJ_UNREACHABLE;
     } else {
       // Ignore undefined field.
+      KJ_REQUIRE(!codec.impl->rejectUnknownFields, "Unknown field", name);
```
mologie, line 1320: The comment in the previous line is no longer up-to-date with this change.

mologie (changes requested):
I had a patch for this ready, but you were faster than me ;-). LGTM!

Slightly related: I also noticed that null fields are skipped, even when field names are unknown. I have not looked into the issue in detail yet because it is not really an issue for my use case, but I'd prefer to enhance/complete my previous pull request to avoid unexpected/surprising behavior with null fields.

mologie:
Addendum: Local testing based on your branch confirmed that the behavior with null values that I described was related to missing checks for annotations, which your changes here fix.

PR #1050, opened 2020-08-26
Emscripten fibers, pieces and parts

unicomp21:
Hi @kentonv 

My hope is this dockerfile can be a "start" point for the journey of adding emscripten fiber support.  The intention w/ this PR is to get a "game plan" going for moving forward (ie requirements, expectations, ideas).

Thanks,
JD

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,5 +1,5 @@
 ###
-FROM ubuntu:19.10 as filament
+FROM ubuntu:19.10
 
 RUN apt-get update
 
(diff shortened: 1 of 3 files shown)


kentonv:
Hmm I don't really understand. Why does emscripten call for Docker, when we've never used it for any of our other build targets? What would people use this Dockerfile for exactly?

unicomp21:
I clearly don't understand how tests are run for this repo.  Given they don't run in docker, do they run in vm's?  If I automate tests w/ node.js or headless chrome, what will it all run in?

kentonv:
In CI, we use GitHub Actions. See: https://github.com/capnproto/capnproto/tree/master/.github/workflows

In theory we could use Docker on top of Actions, but the Actions runners come with a whole lot of stuff already installed and it's nice to leverage that so that we don't have to download lots of extra stuff. E.g. Chrome is already installed. See: https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md

When developing, I run the tests on my own machine, not in any VM or container. I install all the tools I need globally on my system. (I use a Linux desktop.)

unicomp21:
Got it, thanks @kentonv

PR #1055, opened 2020-09-01
Fix use-after-free in reconnect-test.

kentonv:
This is a bug in the test, not in the code it is testing.


kentonv:
MSVC ICE'd. Gross. But not on this file, so must not be new.

PR #1060, opened 2020-09-14
Adds CAPNP_BEGIN_HEADER and CAPNP_END_HEADER around schema headers

ThatGeoGuy:
In reference to an email I sent to the mailing list: https://groups.google.com/d/msg/capnproto/NN5LXfR5ngw/Jf8ACrWzBAAJ

This should bound any instance of namespaces for `capnp::schemas` with `CAPNP_BEGIN_HEADER` and `CAPNP_END_HEADER` like so:

```
CAPNP_BEGIN_HEADER

namespace capnp {
namespace schemas {

} // namespace schemas
} // namespace capnp

CAPNP_END_HEADER

```


c++/src/capnp/compiler/capnpc-c++.c++
```diff
@@ -3045,13 +3045,17 @@ private:
...
+          "\n"
           "namespace capnp {\n"
           "namespace schemas {\n"
           "\n",
           KJ_MAP(n, nodeTexts) { return kj::mv(n.capnpSchemaDecls); },
           "\n"
           "}  // namespace schemas\n"
           "}  // namespace capnp\n"
+          "\n"
+          "CAPNP_END_HEADER\n"
```
kentonv, line 3058: Looks like you've placed this at the end of the schema block, but the main body of the header actually appears after that. The last part of the header is generated at line 3067 below, so you'll want to insert this after that.

kentonv:
You'll need to run `regenerate-bootstraps.sh` to regenerate all the generated code (like `schema.capnp.h`) using the modified compiler.

c++/src/capnp/compiler/capnpc-c++.c++
ThatGeoGuy: Should be resolved as of 4df555139a3feaaf9746740cefecf0beadc01ba9

ThatGeoGuy:
Resolved the above feedback and re-ran `regenerate-bootstraps.sh`. 

Do let me know what else may be needed!

kentonv:
Ugh, looks like our MSVC and Cygwin tests have both bitrotted -- not your fault.

Looks good, thanks for doing this!

PR #1069, opened 2020-09-22
Combine small writes together in TLS compatibility layer

a-robinson:
This is a bit hacky, but it's a quick fix that makes a noticeable
difference in the number of small packets we send over the wire,
especially now that TCP_NODELAY is enabled on the server side.

If you'd prefer a different approach, let me know.

I also added a tiny little second commit that's mostly unrelated, which I can remove if you don't care about it.

### DIFF
--- a/c++/src/kj/compat/readiness-io.h
+++ b/c++/src/kj/compat/readiness-io.h
@@ -71,19 +71,54 @@ public:
   kj::Promise<void> whenReady();
   // Returns a promise that resolves when write() will return non-null.
 
+  class Cork;
+  // An object that, when destructed, will uncork its parent stream.
+
+  Cork cork();
+  // After calling, data won't be pumped until either the internal buffer fills up or the returned
+  // object is destructed. Use this if you know multiple small write() calls will be happening in
+  // the near future and want to flush them all at once.
+  // Once the returned object is destructed, behavior goes back to normal. The returned object
+  // must be destructed before the ReadyOutputStreamWrapper.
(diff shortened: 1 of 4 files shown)


c++/src/kj/compat/readiness-io.h
```diff
@@ -71,11 +71,20 @@ class ReadyOutputStreamWrapper {
   kj::Promise<void> whenReady();
   // Returns a promise that resolves when write() will return non-null.
 
+  _::Deferred<Function<void()>> disableAutomaticPumping();
```
jclee, line 74: Maybe "disableEagerPumping"?

a-robinson: Seems a bit clearer, sure. Done.

c++/src/kj/compat/readiness-io.h
```diff
@@ -71,11 +71,21 @@ class ReadyOutputStreamWrapper {
   kj::Promise<void> whenReady();
   // Returns a promise that resolves when write() will return non-null.
 
+  _::Deferred<Function<void()>> disableEagerPumping();
```
kentonv, line 74: `Function` requires allocations and virtual calls, and `_::Deferred` is an internal type which makes it not great for interface use. Maybe better to introduce a new hand-written type here?

Also in the bikeshedding department, you could call this `cork()`, and the returned type could be `Cork`, after Linux's similarly-purposed `TCP_CORK`. (Or ignore me if you don't like that idea...)

a-robinson: Yeah, seems wise. And `cork` is good enough for me 👍

Issue #1062, opened 2020-09-16
How to add search directories for capnp?


Biswa96:
In MSYS2/MinGW, the include directory is `/mingw64/include/` for x86 and `/mingw64/include/` for x86_64. But according to the [docs][1], I have to add `-I` option in command line to include the search directory. Is it possible to add search directory path at build time? See the discussion at https://github.com/msys2/MINGW-packages/pull/6910.

[1]: https://capnproto.org/cxx.html

kentonv:
Hi @Biswa96,

It sounds like in this environment, you should tell the build system to install to a different prefix than the default `/usr/local`. You probably want to invoke `configure` like so:

    ./configure --prefix=/mingw64

If you do this, the resulting `capnp` binary will have `/mingw64/include` in its default include path. Also, setting `--prefix` this way will cause `make install` to install Cap'n Proto into `/mingw64` in general, which seems like what you want.

If you aren't using the autotools build, you can define the preprocessor symbol `CAPNP_INCLUDE_DIR` to specify an additional include path. This macro should expand to a string literal. So, you could use a compiler flag like `'-DCAPNP_INCLUDE_DIR="/mingw64/include"'` -- note that the outer single-quotes are necessary so that the inner double-quotes don't get removed by the shell.

Biswa96:
Here are something I have tried:

----

```sh
CFLAGS+=' -DCAPNP_INCLUDE_DIR=\"/mingw64/include\"' \
CXXFLAGS+=' -DCAPNP_INCLUDE_DIR=\"/mingw64/include\"' \
cmake.exe
// other options here
```

This compiles but the capnp.exe contains full path of `/mingw64/include` directory. This causes problem because my installation may not be same with others.

----

```sh
CFLAGS+=' -DCAPNP_INCLUDE_DIR="/mingw64/include"' \
CXXFLAGS+=' -DCAPNP_INCLUDE_DIR="/mingw64/include"' \
cmake.exe
// other options here
```

This causes compiler error.

----

Tried to replace `/usr/include` with `/mingw64/include` in `src/capnp/compiler/capnp.c++` file. Then compiled it. Compilation went fine. capnp.exe contains `/mingw64/include` string. But capnp compile fails with `error: Import failed: /capnp/c++.capnp`.

kentonv:
I see, so a compile-time constant is not sufficient here, because you want to be able to redistribute the binary and other users may use different installation locations.

[...]

PR #1071, opened 2020-09-28
cmake: set CAPNP_INCLUDE_DIR

p4l1ly:
when installing capnp to a non-standard location (out of `/usr` or `/usr/local`), the `capnp compile` must be issued with the `-I` option even for the standard `/capnp/c++.capnp` import. Some projects that use `capnp compile` directly, not via cmake (e.g. [capnproto-java](https://github.com/capnproto/capnproto-java/blob/b60bc1e7148893ad584e369b7bd3c1717f1d2cb0/compiler/pom.xml#L87)) cannot be built in this configuration.

### DIFF
--- a/c++/src/capnp/CMakeLists.txt
+++ b/c++/src/capnp/CMakeLists.txt
@@ -168,6 +168,9 @@ if(NOT CAPNP_LITE)
   set_target_properties(capnp_tool PROPERTIES CAPNP_INCLUDE_DIRECTORY
     $<JOIN:$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/..>,$<INSTALL_INTERFACE:${CMAKE_INSTALL_BINDIR}/..>>
   )
+  set_target_properties(capnp_tool PROPERTIES
+    COMPILE_DEFINITIONS CAPNP_INCLUDE_DIR=\"${CMAKE_INSTALL_FULL_INCLUDEDIR}\"
+  )
 
   add_executable(capnpc_cpp
     compiler/capnpc-c++.c++


p4l1ly:
The windows CI checks fail for some reason that seems unrelated (test executables don't get built) and I have too little experience to be able to debug it

c++/src/capnp/CMakeLists.txt
```diff
@@ -168,6 +168,7 @@ if(NOT CAPNP_LITE)
   set_target_properties(capnp_tool PROPERTIES CAPNP_INCLUDE_DIRECTORY
     $<JOIN:$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/..>,$<INSTALL_INTERFACE:${CMAKE_INSTALL_BINDIR}/..>>
   )
+set_target_properties(capnp_tool PROPERTIES COMPILE_FLAGS -DCAPNP_INCLUDE_DIR='"${CMAKE_INSTALL_FULL_INCLUDEDIR}"')
```
kentonv, line 171: nitpick: Seems like this should be indented to match the surrounding lines?

kentonv:
The Cygwin test failure is expected (it has... bitrotted), but the other failures seem to be caused by this change.

c++/src/capnp/CMakeLists.txt
p4l1ly: ahh ugly, sorry, am I blind or did vim betray me few minutes before push?

harrishancock (approved):
Makes sense!

kentonv:
Can you please make this change without changing the format of the macro? Otherwise this breaks people who are already customizing the macro in their own builds.

p4l1ly:
The problem is that I cannot figure out how to pass the double quotes to `COMPILE_FLAGS` in a Windows/Linux compatible way.

Anyway, I think that requiring double quotes around the argument is quite weird and as it moreover violates platform independency, it might be a good idea to remove it. The version is still unstable...

p4l1ly:
ok, the escaping `"\\\"${CMAKE_INSTALL_FULL_INCLUDEDIR}\\\""`, however ugly, seems to work.

c++/src/capnp/CMakeLists.txt
```diff
@@ -168,6 +168,7 @@ if(NOT CAPNP_LITE)
   set_target_properties(capnp_tool PROPERTIES CAPNP_INCLUDE_DIRECTORY
     $<JOIN:$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/..>,$<INSTALL_INTERFACE:${CMAKE_INSTALL_BINDIR}/..>>
   )
+  set_target_properties(capnp_tool PROPERTIES COMPILE_FLAGS -DCAPNP_INCLUDE_DIR="\\\"${CMAKE_INSTALL_FULL_INCLUDEDIR}\\\"")
```
Sarcasm, line 171: Any reason to use `COMPILE_FLAGS` over `COMPILE_DEFINITIONS`?
The latter seems more appropriate, and I think the escaping syntax will be lighter:

```cmake
set_target_properties(capnp_tool PROPERTIES
  COMPILE_DEFINITIONS CAPNP_INCLUDE_DIR=\"${CMAKE_INSTALL_FULL_INCLUDEDIR}\")
```

I don't know if the capnproto installation is relocatable today, but using `CMAKE_INSTALL_FULL_XXX` in the binary may harm relocatability of the package.
Unfortunately, I think the solution to this is more complex.
To do so, the binary should be able to locate it's own installation prefix at runtime, which is platform specific.

p4l1ly: Thank you for the hint, I didn't know about `COMPILE_DEFINITIONS`.
As for the relocatability, the installation was not relocatable before in the context of default include paths. Actually, the default include paths were hardcoded to `/usr/include` and `/usr/local/include`. The proposed change is not the ideal solution but is a bit more dynamic than before. The installation cannot be relocated but it can be at least installed locally. It was not possible before if using `cmake`.

kentonv: Yeah, this makes the cmake build match what the autotools build has done for a long time, which seems like an improvement.

Making the binary relocatable was discussed a bit in #1062, but it's not completely obvious what the right approach is.

kentonv:
@harrishancock Can you verify this still looks good to you and, if so, feel free to merge?

PR #1072, opened 2020-09-29
Disable clang unreachable code warnings for known-unreachable code

jclee:
This change cleans up compiler output when building capnproto in a project that is using clang with `-Wunreachable-code`.  I'm not sure if there is a cleaner way to do this, or one that could be applied to more compilers.

(Since these are `#pragma`s, I assume there's no way to substitute them via macros, so I assume we'd otherwise have to resort to tiny `#include` files, which seemed kind of messy?)

### DIFF
--- a/c++/src/kj/async-io-unix.c++
+++ b/c++/src/kj/async-io-unix.c++
@@ -1110,6 +1110,7 @@ public:
         KJ_SYSCALL_HANDLE_ERRORS(::setsockopt(
               ownFd.get(), IPPROTO_TCP, TCP_NODELAY, (char*)&one, sizeof(one))) {
           case EOPNOTSUPP:
+          case ENOPROTOOPT: // (returned for AF_UNIX in cygwin)
             break;
           default:
             KJ_FAIL_SYSCALL("setsocketopt(IPPROTO_TCP, TCP_NODELAY)", error);
--- a/c++/src/kj/common.c++
+++ b/c++/src/kj/common.c++
@@ -44,7 +44,7 @@ void unreachable() {
   KJ_FAIL_ASSERT("Supposedly-unreachable branch executed.");
 
   // Really make sure we abort.
-  abort();
+  KJ_KNOWN_UNREACHABLE(abort());
 }
 
 }  // namespace _ (private)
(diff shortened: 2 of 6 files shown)


jclee:
Ugh... g++ doesn't like clang `#pragma`s, since `-Werror=unknown-pragmas` is enabled.  I guess the tiny-include route is required -- unless we disable that error, which is similarly unappealing.  :(

kentonv:
You can invoke pragmas using a non-preprocessor syntax:

    _Pragma("clang diagnostic ignored \"-Wunreachable-code\"")

Which you can in turn place inside a macro. So maybe define a macro like `KJ_KNOWN_UNREACHABLE(code)` that resolves to `code` with the appropriate pragmas around it to suppress the warning?

jclee:
Oh, nice, will try that.

c++/src/kj/async-io-unix.c++
```diff
@@ -1101,19 +1101,21 @@ public:
...
+	if (addr.ss_family != AF_UNIX) {
+	  // TODO(perf):  As a hack for the 0.4 release we are always setting
+	  //   TCP_NODELAY because Nagle's algorithm pretty much kills Cap'n Proto's
+	  //   RPC protocol.  Later, we should extend the interface to provide more
+	  //   control over this.  Perhaps write() should have a flag which
+	  //   specifies whether to pass MSG_MORE.
+	  int one = 1;
+	  KJ_SYSCALL_HANDLE_ERRORS(::setsockopt(
+		ownFd.get(), IPPROTO_TCP, TCP_NODELAY, (char*)&one, sizeof(one))) {
+	    case EOPNOTSUPP:
```
jclee, line 1113: Oh, I guess another possible explanation for why the tests were passing on non-cygwin is that implementations return EOPNOTSUPP for unix domain sockets?  In which case, the cygwin error message ("Protocol not available") suggests that it might have been returning EPROTONOSUPPORT instead.  So an alternative solution would be to just add that to the list of ignored cases.

kentonv: Yeah that's probably a better solution, in case anyone ever uses this code with families other than IP and Unix.

c++/src/kj/compat/http.c++
```diff
@@ -1613,6 +1613,15 @@ private:
 template <char...>
 struct FastCaseCmp;
 
+// Disable tautological compare warning/error seen on gcc 10.  It triggers on patterns like
```
kentonv, line 1616: An alternative solution that I like a little better, if it works: https://github.com/ecatmur/capnproto/commit/c38629e5a4b8ce9561b81cb23ea5fc39cbd93eb7

@ecatmur Were you planning to make a PR for that commit?

jclee: Nice!

PR #1076, opened 2020-10-08
Add -ocapnp example to encoding doc

jclee:
Adds an example of the "-ocapnp" command line invocation.  I'm not sure how valuable this is... It is already mentioned a couple other places in the docs, and it looks like the encoding doc is trying to avoid getting into command-line specifics.  But on the other hand, the feature seems to be commonly overlooked, so an additional in-context mention might help?

Also, drive-by fixes of a couple typos...  Was also thinking of changing "echos" to "echoes", but both are apparently accepted usage.


doc/encoding.md
```diff
@@ -117,7 +117,7 @@ Fields are positioned within the struct according to an algorithm with the follo
 
 Field offsets are computed by the Cap'n Proto compiler.  The precise algorithm is too complicated
 to describe here, but you need not implement it yourself, as the compiler can produce a compiled
-schema format which includes offset information.
+schema format which includes offset information, e.g. by running `capnp compile -ocapnp myschema.capnp`.
```
kentonv, line 120: Hmm, this is somewhat awkward in that the paragraph is really talking about programmatic access to the offsets via the compiled schemas / CodeGeneratorRequest, but `-ocapnp` produces human-readable debug output. I'm not sure this is the right place to mention `-ocapnp`, but if we do want to mention it, we should also describe in more detail how to access the compiled schemas programmatically.

jclee: Ah, right.  I wasn't sure whether the "compiled schemas" referred to the "-ocapnp" output or something more general.  I updated the text.

doc/encoding.md
```diff
@@ -117,7 +117,10 @@ Fields are positioned within the struct according to an algorithm with the follo
 
 Field offsets are computed by the Cap'n Proto compiler.  The precise algorithm is too complicated
 to describe here, but you need not implement it yourself, as the compiler can produce a compiled
-schema format which includes offset information.
+schema format which includes offset information, which you can then use in a
+[compiler plugin](otherlang.html#how-to-write-compiler-plugins).  The included
```
kentonv, line 121: Well... Compiled schemas are used by more than just compiler plugins. You can access them at runtime, use them with the Dynamic API, etc...

https://capnproto.org/cxx.html#dynamic-reflection might be worth linking?

Issue #1079, opened 2020-10-09
UBSAN violation in kj::Vector/ArrayBuilder? Also uninitialized variable in ArrayBuilder::disposer


vlovich:
The following lines of code:
```
      kj::Vector<int> fds;
      auto exports = connectionState->writeDescriptors(
          capTable.getTable(), callBuilder.getParams(), fds);
      message->setFds(fds.releaseAsArray());
```

in RpcRequest (rpc.c++) seem to trigger a violation but the problem appears to be in kj::Vector::releaseAsArray on an empty array (which is possible, for one example, if the cap table passed to `writeDescriptors` is empty).

```
capnproto/c++/src/kj/array.h:472:59: runtime error: reference binding to null pointer of type 'const kj::ArrayDisposer'
    #0 0x922f7a in kj::ArrayBuilder<int>::finish() capnproto/c++/src/kj/array.h:472:14
    #1 0x916a65 in kj::Vector<int>::releaseAsArray() capnproto/c++/src/kj/vector.h:70:20
    #2 0x891acb in capnp::_::(anonymous namespace)::RpcConnectionState::RpcRequest::setupSend(bool) capnproto/c++/src/capnp/rpc.c++:1828:27
    #3 0x89099b in capnp::_::(anonymous namespace)::RpcConnectionState::RpcRequest::sendInternal(bool) capnproto/c++/src/capnp/rpc.c++:1849:21
    #4 0x88f069 in capnp::_::(anonymous namespace)::RpcConnectionState::RpcRequest::send() capnproto/c++/src/capnp/rpc.c++:1718:27
```

I also noticed that on Windows disposer isn't NULL but that's because it's an uninitialized variable containing a garbage value. Namely the `ArrayBuilder()` and `ArrayBuilder(decltype(nullptr))` constructors don't initialize `disposer` to `nullptr`. If I explicitly default the disposer member variable to a default of `nullptr` then Windows has the same problem dereferencing a null pointer.

Stylistic curiosity. Why `decltype(nullptr)` vs `std::nullptr` from `cstdef`?

dwrensha:
Related: #549, which was reverted in #578.

kentonv:
I guess an easy fix would be to initialize the disposer pointer to `&NullArrayDisposer::instance`.

> Stylistic curiosity. Why decltype(nullptr) vs std::nullptr from cstdef?

KJ has ambitions to be independent of the C++ standard library, so when it's easy to avoid including a standard header I like to avoid it.

PR #1081, opened 2020-10-09
Fixes to make sanitizers happy

vlovich:
Address TSAN discussion from https://groups.google.com/g/capnproto/c/634juhn5ap0/m/pyRiwWl1AAAJ
Address UBSAN discussion in #1079 per (at least as I understood) @kentonv's suggestion.

### DIFF
--- a/c++/src/capnp/layout.c++
+++ b/c++/src/capnp/layout.c++
@@ -34,7 +34,7 @@ namespace capnp {
 namespace _ {  // private
 
 #if !CAPNP_LITE
-static BrokenCapFactory* brokenCapFactory = nullptr;
+static BrokenCapFactory* globalBrokenCapFactory = nullptr;
 // Horrible hack:  We need to be able to construct broken caps without any capability context,
 // but we can't have a link-time dependency on libcapnp-rpc.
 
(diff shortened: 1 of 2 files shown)


c++/src/capnp/layout.c++
```diff
@@ -30,26 +30,47 @@
...
+#endif
+#endif
+
+
 namespace capnp {
 namespace _ {  // private
 
 #if !CAPNP_LITE
-static BrokenCapFactory* brokenCapFactory = nullptr;
+static BrokenCapFactory* gBrokenCapFactory = nullptr;
```
kentonv, line 48: This `g` prefix isn't idiomatic KJ, but I'd accept `globalBrokenCapFactory`.

c++/src/capnp/layout.c++
```diff
@@ -30,26 +30,47 @@
...
 #elif _MSC_VER
-  *static_cast<BrokenCapFactory* volatile*>(&brokenCapFactory) = &factory;
+  *static_cast<BrokenCapFactory* volatile*>(&gBrokenCapFactory) = &factory;
 #else
 #error "Platform not supported"
 #endif
 }
 
+static BrokenCapFactory* readGlobalBrokenCapFactoryForLayoutCpp() {
+#if IS_BUILDING_WITH_TSAN()
```
kentonv, line 65: Let's just have it always be atomic and skip the TSAN check. No need to comment about TSAN either -- it's reasonable enough that if the global is atomically written, it ought to be atomically read.

c++/src/capnp/layout.c++
vlovich: Sorry about that. Switching between different codebases.  Definitely prefer KJ's style.

c++/src/capnp/layout.c++
vlovich: Good call.

kentonv:
Thanks.

PR #1083, opened 2020-10-15
Fix canceler.isEmpty() check in http-over-capnp

bretthoerner:
Additionally, the documentation for the bool result from Canceler.isEmpty was reversed.

---

All credit to @harrishancock


kentonv:
Wow, nice work.

Issue #1063, opened 2020-09-16
rpc over websockets


zenhack:
I was looking at the Sandstorm app-index's implementation, and rather than using capnp rpc for submissions it uses an ad-hoc protocol, with a comment indicating that this is blocked on capnp-rpc over the internet.

I'd like to clean this up by just speaking capnp rpc over a websocket, which seems like it ought to be straightforward enough. The first step would be an implementation of AsyncIoStream on top of WebSocket, which probably belongs in kj.

There's a slight impedence mismatch here, since websockets is message oriented while AsyncIoStream is a streaming interface, but for capnproto's purposes it's fine for the API to obscure this detail.

kentonv:
Hmm, I think it's pretty important that capnp-over-WebSocket actually respects the WebSocket framing, i.e. encoding one capnp message in one WebSocket message. So, it might be necessary to implement an alternative `VatNetwork` that uses `WebSocket` rather than `AsyncIoStream`...

zenhack:
Sounds reasonable to me; changed the title accordingly.

But it seems like most of the VatNetwork implementation would be the same as for TwoPartyVatNetwork; perhaps instead we could just factor out the streaming vs. datagram difference, and add a new constructor for this case (so we'd have one that takes AsyncIoStream, one that takes CapabilityStream, and then some new interface (unless there is already something suitable) that is datagram-oriented).

kentonv:
That makes sense to me.

Probably the right abstraction is basically to virtualize the `serialize-async.h` interfaces, like:

```
class MessageSocket {
public:
  virtual kj::Promise<MessageReaderAndFds> readMessage(
      kj::ArrayPtr<kj::AutoCloseFd> fdSpace, ReaderOptions options = ReaderOptions()) = 0;
  virtual kj::Promise<void> writeMessage(kj::ArrayPtr<const int> fds, MessageBuilder& builder) = 0;
};
```

JonasVautherin:
@zenhack: you did resolve this with https://github.com/capnproto/capnproto/pull/1204, and hence it can be closed, right?

I ended up here because I was curious about why WebSockets are using `MessageStream` instead of `AsyncIoStream` directly, and Kenton said (above):

> Hmm, I think it's pretty important that capnp-over-WebSocket actually respects the WebSocket framing, i.e. encoding one capnp message in one WebSocket message.

[...]

PR #1095, opened 2020-10-30
Abstract out the interfaces in serialize-async.h, use the abstraction in TwoPartyVatNetwork

zenhack:
This defines a `MessageTransport` interface class, per discussion in #1063. It includes implementations which wrap `AsyncIoStream` and `AsyncCapabilityStream`, and a subsequent PR can add a websocket implementation.


c++/src/capnp/serialize-async.h
```diff
@@ -76,6 +76,107 @@ kj::Promise<void> writeMessage(kj::AsyncCapabilityStream& output, kj::ArrayPtr<c
...
+      kj::ArrayPtr<kj::AutoCloseFd> fdSpace,
+      ReaderOptions options = ReaderOptions(), kj::ArrayPtr<word> scratchSpace = nullptr) override;
+  kj::Promise<void> writeMessage(
+      kj::ArrayPtr<const int> fds,
+      kj::ArrayPtr<const kj::ArrayPtr<const word>> segments) override;
+private:
+  kj::AsyncCapabilityStream& stream;
+};
+
+kj::Promise<kj::Own<MessageReader>> readMessage(
```
kentonv, line 156: I think these freestanding functions should be methods of `MessageTransport` instead. This is consistent with e.g. `AsyncInputStream` which has `read()` defined as a method on the base class, which calls `tryRead()` and converts EOF into an exception.

The reason the existing functions in this file are free-standing is because they don't really belong on any type in particular, but `MessageTransport` could clearly host these new functions.

c++/src/capnp/serialize-async.h
```diff
@@ -76,6 +76,107 @@ kj::Promise<void> writeMessage(kj::AsyncCapabilityStream& output, kj::ArrayPtr<c
     KJ_WARN_UNUSED_RESULT;
 // Write a message with FDs attached, e.g. to a Unix socket with SCM_RIGHTS.
 
+// -----------------------------------------------------------------------------
+// Versions virtualized over an interface.
+
+class MessageTransport {
```
kentonv, line 82: `MessageStream` might be a clearer name. "Transport" is more abstract and I sometimes use it to describe a network rather than an individual stream.

I wonder if it makes any sense to have separate `MessageInputStream` and `MessageOutputStream`. But I guess it doesn't matter for the immediate use case, and it's possible to refactor later if we want.

c++/src/capnp/serialize-async.h
```diff
@@ -101,6 +91,29 @@ class MessageTransport {
...
+  // Get the size of the underlying send buffer, if applicable. The RPC
+  // system uses this as a hint for flow control purposes; see:
+  //
+  // https://capnproto.org/news/#multi-stream-flow-control
+  //
+  // ...for a more thorough explanation of how this is used. Implementations
+  // may return nullptr if they do not have access to this information, or if
+  // the underlying transport does not use a congestion window.
+
+  virtual kj::Promise<void> shutdownWrite() = 0;
```
kentonv, line 114: Let's call this `end()`. I'm working on a refactor of `AsyncOutputStream` which replaces `shutdownWrite()` with `end()` (returning a promise, like you do here).

c++/src/capnp/rpc-twoparty.h
```diff
@@ -83,7 +89,20 @@ class TwoPartyVatNetwork: public TwoPartyVatNetworkBase,
...
   class IncomingMessageImpl;
 
-  kj::OneOf<kj::AsyncIoStream*, kj::AsyncCapabilityStream*> stream;
+  typedef kj::OneOf<
+      AsyncIoTransport,
+      AsyncCapabilityTransport,
+      MessageTransport*
+  > TransportUnion;
+
+  TransportUnion transport;
```
kentonv, line 98: I'd go with:

    MesasgeTransport& transport;
    kj::Own<MessageTransport> ownTransport;  // only if we created it

Saving an allocation here is not important as this is a long-lived object and many, many allocations will be performed while using it. (Though if you really wanted to save an allocation, you could use `kj::SpaceFor`... but I wouldn't.)

Or you could even reduce this to a single `kj::Own<MesasgeTransport> transport;` by using `NullDisposer` in the case that you don't actually own it... but that's a little hacky.

c++/src/capnp/rpc-twoparty.c++
```diff
@@ -248,52 +246,31 @@ kj::Own<OutgoingRpcMessage> TwoPartyVatNetwork::newOutgoingMessage(uint firstSeg
...
-                  kj::heap<IncomingMessageImpl>(kj::mv(*m), kj::mv(fdSpace)));
-            } else {
-              return kj::Own<IncomingRpcMessage>(kj::heap<IncomingMessageImpl>(kj::mv(m->reader)));
-            }
-          } else {
-            return nullptr;
-          }
-        });
+  return kj::evalLater([this]() -> kj::Promise<kj::Maybe<kj::Own<IncomingRpcMessage>>> {
+    auto fdSpace = (maxFdsPerMessage == 0)
```
kentonv, line 250: Type deduction through ternary operators is a weird corner of C++ that I've grown to distrust, so I would probably have written this as:

```
kj::Array<kj::AutoCloseFd> fdSpace = nullptr;
if (maxFdsPerMessage > 0) {
  fdSpace = kj::heapArray<kj::AutoCloseFd>(maxFdsPerMessage);
}
```

I realize that probably looks pretty awful to a Haskell programmer... :)

The alternative that I trust is to make sure both branches resolve to the same type, but that'll end up even more verbose in this case.

kentonv:
I could have sworn I was adding all those comments as part of a review, [but apparently not](https://twitter.com/KentonVarda/status/1333522927690657807).

Anyway, sorry it took so long to review this, I've literally not gotten close to inbox zero in months... but finally getting close now... :/

zenhack:
Ok, I believe I've addressed all of your comments.

zenhack:
Ack, looks like there are some test failures. I'll dig into those tomorrow.

I also merged back in changes from master and fixed a link in the docs (see the last commit), which I noticed while making the changes.

c++/src/capnp/rpc-twoparty.c++
```diff
@@ -19,30 +19,53 @@
...
+  disconnectPromise = paf.promise.fork();
+  disconnectFulfiller.fulfiller = kj::mv(paf.fulfiller);
+}
+
+TwoPartyVatNetwork::TwoPartyVatNetwork(
+      kj::Own<capnp::MessageStream>&& stream,
+      uint maxFdsPerMessage,
+      rpc::twoparty::Side side,
+      ReaderOptions receiveOptions)
+  : TwoPartyVatNetwork(*stream, kj::mv(stream), maxFdsPerMessage, side, receiveOptions) {}
```
kentonv, line 57: I bet that different compilers make difference decisions about whether `*stream` or `kj::mv(stream)` should be evaluated first. If the `Own<>` is moved away too early then `*stream` will dereference a null pointer.

This is... kind of a pain to solve.

I think one hacky solution would be to have yet another intermediate constructor that takes `MessageStream&, kj::Own<MessageStream>&&` (without `Maybe` on the second param yet). This way, `stream` does not actually get moved away on this line; rather, an rvalue reference to `stream` is passed down. The new constructor can, in turn, forward to the constructor that takes `MessageStream&, kj::Maybe<kj::Own<MessageStream>>&&`.

Otherwise, the only alternative may be to live with two copies of the constructor.

zenhack: Pushed a slightly different solution, which goes back to using a kj::OneOf; I think this is easier to follow while still not duplicating code.

kentonv: Hmm, I guess this is bikeshedding at this point but to me having to call `getStream()` which does a switch on every usage seems harder to follow... but ok, it doesn't really matter I guess.

PR #1096, opened 2020-10-30
Add lazy option to autoReconnect to avoid connecting until used

a-robinson:
It still reconnects eagerly, but not establishing the initial connection
until the first time the capability is used is nice for capabilities
that are expensive to set up but often not used.

### DIFF
--- a/c++/src/capnp/reconnect.c++
+++ b/c++/src/capnp/reconnect.c++
@@ -27,13 +27,13 @@ namespace {
 
 class ReconnectHook final: public ClientHook, public kj::Refcounted {
 public:
-  ReconnectHook(kj::Function<Capability::Client()> connectParam)
+  ReconnectHook(kj::Function<Capability::Client()> connectParam, bool lazy = false)
       : connect(kj::mv(connectParam)),
-        current(ClientHook::from(connect())) {}
+        current(lazy ? kj::Maybe<kj::Own<ClientHook>>() : ClientHook::from(connect())) {}
 
   Request<AnyPointer, AnyPointer> newCall(
       uint64_t interfaceId, uint16_t methodId, kj::Maybe<MessageSize> sizeHint) override {
-    auto result = current->newCall(interfaceId, methodId, sizeHint);
+    auto result = getCurrent().newCall(interfaceId, methodId, sizeHint);
     AnyPointer::Builder builder = result;
     auto hook = kj::heap<RequestImpl>(kj::addRef(*this), RequestHook::from(kj::mv(result)));
     return { builder, kj::mv(hook) };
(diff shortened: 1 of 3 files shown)


c++/src/capnp/reconnect.c++
```diff
@@ -27,23 +27,33 @@ namespace {
...
-    auto hook = kj::heap<RequestImpl>(kj::addRef(*this), RequestHook::from(kj::mv(result)));
-    return { builder, kj::mv(hook) };
+    KJ_IF_MAYBE(c, current) {
+      auto result = (*c)->newCall(interfaceId, methodId, sizeHint);
+      AnyPointer::Builder builder = result;
+      auto hook = kj::heap<RequestImpl>(kj::addRef(*this), RequestHook::from(kj::mv(result)));
+      return { builder, kj::mv(hook) };
+    } else {
+      current = ClientHook::from(connect());
+      return newCall(interfaceId, methodId, sizeHint);
```
kentonv, line 43: This recursion strikes me as a little odd. What about having a helper method like:

    ClientHook& getCurrent() {
      KJ_IF_MAYBE(c, current) {
        return **c;
      } else {
        return current.emplace(ClientHook::from(connect()));
      }
    }

Then call that in the other methods.

c++/src/capnp/reconnect.h
```diff
@@ -52,13 +52,26 @@ auto autoReconnect(ConnectFunc&& connect);
...
-Capability::Client autoReconnect(kj::Function<Capability::Client()> connect);
+Capability::Client autoReconnect(kj::Function<Capability::Client()> connect, bool lazy = false);
 template <typename ConnectFunc>
 auto autoReconnect(ConnectFunc&& connect) {
-  return autoReconnect(kj::Function<Capability::Client()>(kj::fwd<ConnectFunc>(connect)))
+  return autoReconnect(kj::Function<Capability::Client()>(kj::fwd<ConnectFunc>(connect)), false)
+      .castAs<FromClient<kj::Decay<decltype(connect())>>>();
+}
+
+Capability::Client lazyAutoReconnect(kj::Function<Capability::Client()> connect);
```
kentonv, line 71: Were you intending to define this in the `.c++` file? It looks like you ended up adding a `bool lazy` parameter to `autoReconnect()` instead.

I think I'd prefer having the two named declarations rather than the boolean parameter visible in the header.

c++/src/capnp/reconnect.c++
a-robinson: Agreed, done.

c++/src/capnp/reconnect.h
a-robinson: Looks like I did a little of column A and a little of column B then forgot to clean one of them up. Switched over to the separate declarations.

bretthoerner (approved):

a-robinson:
Assuming that this is ok to merge based on having resolved the comments here and the prior agreement on this general solution.

PR #1097, opened 2020-11-02
Add support for List(Capability)

### LINKED ISSUE #1089: Support for List(Capability)?
I was trying to use a `List(Capability)` in a schema, and ended up with the following error:

```
error: 'List(AnyPointer)' is not supported.
```

I found [an e-mail from 2017](https://www.mail-archive.com/capnproto@googlegroups.com/msg00599.html) where you say:

> These days we also have AnyStruct, AnyList, and Capability. I think
> List(AnyList) and List(Capability) ought to be made legal if they aren't
> already (can't remember). Or in some use cases, it makes sense to replace
> List(AnyPointer) with AnyList -- thus supporting lists of any type, as long
> as all the elements are the same type.

Looking at [schema.c++](https://github.com/capnproto/capnproto/blob/b4f0a7116a02146592e6d4445346695977d394ae/c%2B%2B/src/capnp/schema.c%2B%2B#L742), it seems like the error comes from that file:

```
case schema::Type::ANY_POINTER:
      KJ_FAIL_REQUIRE("List(AnyPointer) not supported.");
```

However, I can create a struct wrapping the Capability, say `CapabilityHolder`, and successfully use a `List(CapabilityHodler)`, which works as a workaround.

Is it tricky to enable the use of `List(Capability)` (or, as you mention, "lists of any type, as long as all the elements are the same type"), or could it qualify for a first contribution in the codebase? With some pointers, I would be happy to have a look at this.

BTW, I am new to capabilities, but I'm quickly becoming a big fan (and I love that capnproto's documentation gives reference to capabilities-related publications, that is invaluable). Same for capnproto, thanks a lot for all the work there!

JonasVautherin:
This is hoping to solve #1089. This time I managed to get to a point where it works for me, with a small example using a list of capabilities defined as:

```
fetchCapabilities @0 () -> (caps :List(Capability));
```

With those changes, I can pass a list of capabilities through `fetchCapabilities` by modifying the calculator example, and it works. I finally realized that I should handle the `AnyPointer` type further in the pipeline. I was first trying to do that in `schema.h`, but I think that it did not work because an `Unconstrained::Capability` is not an `InterfaceSchema`, though both are `INTERFACE`s in the end.

Resolves #1089.

### DIFF
--- a/c++/src/capnp/compiler/generics.c++
+++ b/c++/src/capnp/compiler/generics.c++
@@ -131,10 +131,19 @@ bool BrandedDecl::compileAsType(
         }
 
         if (elementType.isAnyPointer()) {
-          addError(errorReporter, "'List(AnyPointer)' is not supported.");
-          // Seeing List(AnyPointer) later can mess things up, so change the type to Void.
-          elementType.setVoid();
-          return false;
+          auto unconstrained = elementType.getAnyPointer().getUnconstrained();
+
+          if (unconstrained.isAnyKind()) {
+            addError(errorReporter, "'List(AnyPointer)' is not supported.");
+            // Seeing List(AnyPointer) later can mess things up, so change the type to Void.
+            elementType.setVoid();
+            return false;
+          } else if (unconstrained.isStruct()) {
+            addError(errorReporter, "'List(AnyStruct)' is not supported.");
+            // Seeing List(AnyStruct) later can mess things up, so change the type to Void.
+            elementType.setVoid();
+            return false;
+          }
         }
 
         return true;
--- a/c++/src/capnp/capability-test.c++
+++ b/c++/src/capnp/capability-test.c++
@@ -80,6 +80,39 @@ TEST(Capability, Basic) {
   EXPECT_TRUE(barFailed);
 }
 
+TEST(Capability, CapabilityList) {
+  kj::EventLoop loop;
+  kj::WaitScope waitScope(loop);
+
+  MallocMessageBuilder builder;
+  auto root = builder.initRoot<test::TestListOfAny>();
+  auto initCapList = root.initCapList(2);
+
+  int callCount0 = 0;
+  int callCount1 = 0;
+  initCapList.set(0, kj::heap<TestInterfaceImpl>(callCount0));
+  initCapList.set(1, kj::heap<TestInterfaceImpl>(callCount1));
+
+  auto capList = root.getCapList();
+  auto cap0 = capList[0].castAs<test::TestInterface>();
+  auto cap1 = capList[1].castAs<test::TestInterface>();
+
+  EXPECT_EQ(2u, root.getCapList().size());
+
+  auto request0 = cap0.fooRequest();
+  request0.setI(123);
+  request0.setJ(true);
+  EXPECT_EQ("foo", request0.send().wait(waitScope).getX());
+
+  auto request1 = cap1.fooRequest();
+  request1.setI(123);
+  request1.setJ(true);
+  EXPECT_EQ("foo", request1.send().wait(waitScope).getX());
+
+  EXPECT_EQ(1, callCount0);
+  EXPECT_EQ(1, callCount1);
+}
+
 TEST(Capability, Inheritance) {
   kj::EventLoop loop;
   kj::WaitScope waitScope(loop);
--- a/c++/src/capnp/compiler/capnpc-c++.c++
+++ b/c++/src/capnp/compiler/capnpc-c++.c++
@@ -304,7 +304,7 @@ kj::String KJ_STRINGIFY(const CppTypeName& typeName) {
   }
 }
 
-CppTypeName whichKind(schema::Type::Which which) {
+CppTypeName whichKind(Type type) {
   // Make a CppTypeName representing the capnp::Kind value for the given schema type. This makes
   // CppTypeName conflate types and values, but this is all just a hack for MSVC's benefit. Its
   // primary use is as a non-type template parameter to `capnp::List<T, K>` -- normally the Kind K
@@ -312,7 +312,7 @@ CppTypeName whichKind(schema::Type::Which which) {
   // of `capnp::List<T, K>` is the return type of a function, and the element type T is a template
   // instantiation.
 
-  switch (which) {
+  switch (type.which()) {
       case schema::Type::VOID: return CppTypeName::makePrimitive(" ::capnp::Kind::PRIMITIVE");
 
       case schema::Type::BOOL: return CppTypeName::makePrimitive(" ::capnp::Kind::PRIMITIVE");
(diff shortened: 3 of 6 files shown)


c++/src/capnp/compiler/generics.c++
```diff
@@ -131,10 +131,19 @@ bool BrandedDecl::compileAsType(
...
 
         if (elementType.isAnyPointer()) {
-          addError(errorReporter, "'List(AnyPointer)' is not supported.");
-          // Seeing List(AnyPointer) later can mess things up, so change the type to Void.
-          elementType.setVoid();
-          return false;
+          auto unconstrained = elementType.getAnyPointer().getUnconstrained();
+
+          if (unconstrained.isAnyKind()) {
+            addError(errorReporter, "'List(AnyKind)' is not supported.");
```
kentonv, line 137: Let's change the error message back to say `List(AnyPointer)`, since that's what someone would actually type in their code to trigger this error. There is no type called `AnyKind`.

c++/src/capnp/compiler/generics.c++
```diff
@@ -131,10 +131,19 @@ bool BrandedDecl::compileAsType(
...
+            // Seeing List(AnyKind) later can mess things up, so change the type to Void.
+            elementType.setVoid();
+            return false;
+          } else if (unconstrained.isStruct()) {
+            addError(errorReporter, "'List(Struct)' is not supported.");
```
kentonv, line 142: Similarly here, change `List(Struct)` to `List(AnyStruct)`, since that's what you'd actually have to write in code to trigger this error.

kentonv (commented):
Nice! I have some nitpicks about the error message below.

Also, we should add tests. I think the errors test should be extended to verify that `List(AnyStruct)` is an error (same as how it tests `List(AnyPointer)` currently), while `test.capnp` should test the allowed cases with a type like:

    struct TestListOfAny {
      listList @0 :List(AnyList);
      capList @1 :List(Capability);
    }

Then maybe write some very simple test case in `encoding-test.c++` (or maybe `capability-test.c++`?) which exercises this to make sure it works. Just verifying that you can allocate each type of list, fill it in, and then read it back, should be good enough.

JonasVautherin:
I tried to add tests for `List(AnyList)`, but I think I am missing a few more steps. More specifically, I think that `any.h` should define `template<> struct List<AnyList, Kind::OTHER>`, but it doesn't. I started working on that, but I struggled on some operator[].

@kentonv: would it be fine to keep this PR as just `List(Capability)`, and have me create a new PR later when I get `List(AnyList)` working?

JonasVautherin:
Note: I am not convinced that the macOS failure is related to this PR, is it?

```
[ TEST ] kj/mutex-test.c++:380: legacy test: Mutex/WhenWithTimeoutPreciseTiming
Failed to start process notifications for pid 17102 (19)
kj/mutex-test.c++:401: failed: expected t >= 100 * kj::MILLISECONDS
[ FAIL ] kj/mutex-test.c++:380: legacy test: Mutex/WhenWithTimeoutPreciseTiming (380257 μs)
stack: 1039bcf9b 1039c5ed0 104795adc 1047c9cca 104794436 104794211 10480219f 1047c9cca 1047ffcfb 104793a6d 7fff6909fcc8 0
    kj::(anonymous namespace)::TestCase380::run() (in capnp-test) + 272: returning here
    kj::_::runCatchingExceptions(kj::_::Runnable&) (in libkj-0.9-dev.dylib) + 26: returning here
    kj::TestRunner::run() (in libkj-test-0.9-dev.dylib) + 534: returning here
    kj::Function<kj::MainBuilder::Validity ()>::Impl<kj::_::BoundMethod<kj::TestRunner&, kj::TestRunner::getMain()::'lambda3'(auto&, auto&&...), kj::TestRunner::getMain()::'lambda4'(auto&, auto&&...)> >::operator()() (in libkj-test-0.9-dev.dylib) + 17: returning here
    kj::MainBuilder::MainImpl::operator()(kj::StringPtr, kj::ArrayPtr<kj::StringPtr const>) (in libkj-0.9-dev.dylib) + 2799: returning here
    kj::_::runCatchingExceptions(kj::_::Runnable&) (in libkj-0.9-dev.dylib) + 26: returning here
    kj::runMainAndExit(kj::ProcessContext&, kj::Function<void (kj::StringPtr, kj::ArrayPtr<kj::StringPtr const>)>&&, int, char**) (in libkj-0.9-dev.dylib) + 347: returning here
    main (in libkj-test-0.9-dev.dylib) + 109: returning here
    start (in libdyld.dylib) + 0: returning here
    0: returning here
```

c++/src/capnp/compiler/generics.c++
```diff
@@ -131,10 +131,24 @@ bool BrandedDecl::compileAsType(
...
+            addError(errorReporter, "'List(AnyPointer)' is not supported.");
+            // Seeing List(AnyKind) later can mess things up, so change the type to Void.
+            elementType.setVoid();
+            return false;
+          } else if (unconstrained.isStruct()) {
+            addError(errorReporter, "'List(AnyStruct)' is not supported.");
+            // Seeing List(Struct) later can mess things up, so change the type to Void.
+            elementType.setVoid();
+            return false;
+          } else if (unconstrained.isList()) { // TODO This should be supported
```
kentonv, line 146: I think I'd leave this as supported in the compiler for now, even though the C++ library currently mishandles it. Other implementations might handle it just fine.

In `test.capnp` I'd add the `List(AnyList)` line but comment it out and put `# TODO(soon): Make List(AnyList) work correctly in C++ generated code.`

c++/src/capnp/compiler/generics.c++
```diff
@@ -131,10 +131,24 @@ bool BrandedDecl::compileAsType(
...
         if (elementType.isAnyPointer()) {
-          addError(errorReporter, "'List(AnyPointer)' is not supported.");
-          // Seeing List(AnyPointer) later can mess things up, so change the type to Void.
-          elementType.setVoid();
-          return false;
+          auto unconstrained = elementType.getAnyPointer().getUnconstrained();
+
+          if (unconstrained.isAnyKind()) {
+            addError(errorReporter, "'List(AnyPointer)' is not supported.");
+            // Seeing List(AnyKind) later can mess things up, so change the type to Void.
```
kentonv, line 138: Can you update these comments too?

kentonv (commented):
Thanks for keeping the commit series so nicely organized!

There's a few follow-ups we should do related to this, but I won't make you do them (unless you want to):
* Fix `List(AnyList)` in C++ generated code.
* Verify that dynamic API works with these types.
* Verify that constants of these types work correctly.

But I'm happy to merge this PR after a couple more small comments are addressed:

kentonv:
And yeah, the mac failure appears to be a random timing issue. Unfortunately those tests are literally testing timing-related stuff so it's impossible to make them completely reliable. :/

JonasVautherin:
Thanks for the reviews! About the follow-up tasks:

> Fix List(AnyList) in C++ generated code.

I started working on that, trying to write the implementation of `template<> struct List<AnyList, Kind::OTHER>` in `any.h`, but I'm not sure yet if that's the right way to go. I may open a new PR to get feedback on that, but I'll spend more time on it before.

> Verify that dynamic API works with these types.

I don't know yet how the dynamic API works, so maybe later, when I get to know it :slightly_smiling_face:.

> Verify that constants of these types work correctly.

I am not sure what that means. You mean checking that a `List(const Capability)` (if that is even a thing?) keeps it const?

kentonv:
Thanks, this looks great!

> I am not sure what that means. You mean checking that a `List(const Capability)` (if that is even a thing?) keeps it const?

No I mean like:

```capnp
const list1 :List(Text) = ["foo"];
const list2 :List(TestAllTyps) = [()];
const listAnyList :List(AnyList) = [list1, list2];
```

I guess that constant capabilities aren't a thing, so this can only really be tested once `List(AnyList)` is fully supported.

PR #1098, opened 2020-11-04
Fix for `make check` in an out-of-source build directory

Dr-Emann:
src/capnp/compiler/capnp-test.sh was failing `make check` for me when building in an out-of-source build. This fixes that.


c++/src/capnp/compiler/capnp-test.sh
```diff
@@ -37,10 +39,10 @@ else
   CAPNP=${CAPNP:-capnp}
 fi
 
-SCHEMA=`dirname "$0"`/../test.capnp
-JSON_SCHEMA=`dirname "$0"`/../compat/json-test.capnp
-TESTDATA=`dirname "$0"`/../testdata
-SRCDIR=`dirname "$0"`/../..
+SCHEMA=$BASE/../test.capnp
```
kentonv, line 42: This part of the change looks unnecessary -- any particular reason for it? If not would you mind reverting it so that it's clear what's actually being fixed?

kentonv (commented):
Thanks! It looks like I broke this recently when I added the "branded alias" test cases. FWIW, this would have been caught before release (indeed, our extended release tests are currently failing in CI).

kentonv:
Thanks!

PR #1100, opened 2020-11-06
Cleanup: rewrite c++1y to c++14

haneefmubarak:
IIRC c++1y should hit some draft of C++14, perhaps the relatively trivial fix is worthwhile?


kentonv:
Thanks. They are equivalent, but it makes sense to use the official name now that it's known. Will merge once tests pass.

PR #1101, opened 2020-11-08
CMake: append cmake folder into CMAKE_MODULE_PATH instead of set

SpaceIm:
Paths defined in `CMAKE_MODULE_PATH` by callers are lost with current CMakeLists. Best practice is to always append new values in CMAKE_MODULE_PATH. For conan package manager, it leads to issues because it injects specific paths in this variable.


harrishancock (approved):
Makes sense to me.

kentonv:
Sorry, this fell off my radar.

PR #1104, opened 2020-11-16
Add Coroutines TS support

harrishancock:
This adds support for the Coroutines TS to libkj. Specifically, you can now co_await kj::Promise<T> objects from inside functions which themselves return kj::Promise<U> objects.

The integration is header-only, so the only difference in a Cap'n Proto build is that we might conditionally compile the coroutine tests, if the compiler and standard library both support the Coroutines TS.

I tested this with Clang/libc++ 12 and the latest VS2019.

* For Clang with autotools / CMake it should be sufficient to configure the toolchain like so, then configure and build like normal:
  export CC=clang
  export CXX=clang++
  export CXXFLAGS="-std=gnu++17 -fcoroutines-ts -stdlib=libc++"
  export LDFLAGS="-fcoroutines-ts -stdlib=libc++"
  ./configure && make check

* For VS2019, add -DCMAKE_CXX_FLAGS="/std:c++17 /await /EHsc" to your regular CMake command line. (The /EHsc shuts up a compiler warning...)

VS2019 and GCC both apparently support C++20-standardized coroutines, but VS2019's version seems ICE-y, and I'm too lazy to figure out how to install GCC 10 on Debian Buster.

### DIFF
--- a/c++/src/kj/async-coroutine-test.c++
+++ b/c++/src/kj/async-coroutine-test.c++
@@ -1,0 +1,496 @@
+// Copyright (c) 2020 Cloudflare, Inc. and contributors
+// Licensed under the MIT License:
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+#include <kj/async.h>
+#include <kj/array.h>
+#include <kj/compat/http.h>
+#include <kj/debug.h>
+#include <kj/test.h>
+
+namespace kj {
+namespace {
+
+#ifdef KJ_HAS_COROUTINE
+
+template <typename T>
+Promise<kj::Decay<T>> identity(T&& value) {
+  co_return kj::fwd<T>(value);
+}
+// Work around a bonkers MSVC ICE with a separate overload.
+Promise<const char*> identity(const char* value) {
+  co_return value;
+}
+
+KJ_TEST("Identity coroutine") {
+  EventLoop loop;
+  WaitScope waitScope(loop);
+
+  KJ_EXPECT(identity(123).wait(waitScope) == 123);
+  KJ_EXPECT(*identity(kj::heap(456)).wait(waitScope) == 456);
+
+  {
+    auto p = identity("we can cancel the coroutine");
+  }
+}
+
+template <typename T>
+Promise<T> simpleCoroutine(kj::Promise<T> result, kj::Promise<bool> dontThrow = true) {
+  KJ_ASSERT(co_await dontThrow);
+  co_return co_await result;
+}
+
+KJ_TEST("Simple coroutine test") {
+  EventLoop loop;
+  WaitScope waitScope(loop);
+
+  simpleCoroutine(kj::Promise<void>(kj::READY_NOW)).wait(waitScope);
+
+  KJ_EXPECT(simpleCoroutine(kj::Promise<int>(123)).wait(waitScope) == 123);
+}
+
+struct Counter {
+  size_t& count;
+  Counter(size_t& count): count(count) {}
+  ~Counter() { ++count; }
+  KJ_DISALLOW_COPY(Counter);
+};
+
+kj::Promise<void> countDtorsAroundAwait(size_t& count, kj::Promise<void> promise) {
+  Counter counter1(count);
+  co_await promise;
+  Counter counter2(count);
+  co_return;
+};
+
+KJ_TEST("co_awaiting an immediate promise does not suspend if the event loop is empty and running") {
+  // The coroutine PromiseNode implementation contains an optimization which allows us to avoid
+  // suspending the coroutine and instead immediately call PromiseNode::get() and proceed with
+  // execution.
+
+  EventLoop loop;
+  WaitScope waitScope(loop);
+
+  // The immediate-execution optimization is only enabled when the event loop is running, so use an
+  // eagerly-evaluated evalLater() to perform the test from within the event loop. (If we didn't
+  // eagerly-evaluate the promise, the result would be extracted after the loop finished.)
+  kj::evalLater([&]() {
+    size_t count = 0;
+
+    auto promise = kj::Promise<void>(kj::READY_NOW);
+    auto coroPromise = countDtorsAroundAwait(count, kj::READY_NOW);
+
+    // `coro` has not been destroyed yet, but it completed and unwound its frame.
+    KJ_EXPECT(count == 2);
+  }).eagerlyEvaluate(nullptr).wait(waitScope);
+
+  kj::evalLater([&]() {
+    // If there are no background tasks in the queue, coroutines execute through an evalLater()
+    // without suspending.
+
+    size_t count = 0;
+    bool evalLaterRan = false;
+
+    auto promise = kj::evalLater([&]() { evalLaterRan = true; });
+    auto coroPromise = countDtorsAroundAwait(count, kj::mv(promise));
+
+    KJ_EXPECT(evalLaterRan == true);
+    KJ_EXPECT(count == 2);
+  }).eagerlyEvaluate(nullptr).wait(waitScope);
+}
+
+KJ_TEST("co_awaiting an immediate promise suspends if the event loop is not running") {
+  // We only want to enable the immediate-execution optimization if the event loop is running, or
+  // else a whole bunch of RPC tests break, because some .then()s get evaluated on promise
+  // construction, before any .wait() call.
+
+  EventLoop loop;
+  WaitScope waitScope(loop);
+
+  size_t count = 0;
+
+  auto promise = kj::Promise<void>(kj::READY_NOW);
+  auto coroPromise = countDtorsAroundAwait(count, kj::READY_NOW);
+
+  // In the previous test, this exact same code executed immediately because the event loop was
+  // running.
+  KJ_EXPECT(count == 0);
+}
+
+KJ_TEST("co_awaiting immediate promises suspends if the event loop is not empty") {
+  // We want to make sure that we can still return to the event loop when we need to.
+
+  EventLoop loop;
+  WaitScope waitScope(loop);
+
+  // The immediate-execution optimization is only enabled when the event loop is running, so use an
+  // eagerly-evaluated evalLater() to perform the test from within the event loop. (If we didn't
+  // eagerly-evaluate the promise, the result would be extracted after the loop finished.)
+  kj::evalLater([&]() {
+    size_t count = 0;
+
+    // We need to enqueue an Event on the event loop to inhibit the immediate-execution
+    // optimization. Creating and then immediately fulfilling an EagerPromiseNode is a convenient
+    // way to do so.
+    auto paf = newPromiseAndFulfiller<void>();
+    paf.promise = paf.promise.eagerlyEvaluate(nullptr);
+    paf.fulfiller->fulfill();
+
+    auto promise = kj::Promise<void>(kj::READY_NOW);
+    auto coroPromise = countDtorsAroundAwait(count, kj::READY_NOW);
+
+    // We didn't immediately extract the READY_NOW.
+    KJ_EXPECT(count == 0);
+  }).eagerlyEvaluate(nullptr).wait(waitScope);
+
+  kj::evalLater([&]() {
+    size_t count = 0;
+    bool evalLaterRan = false;
+
+    // We need to enqueue an Event on the event loop to inhibit the immediate-execution
+    // optimization. Creating and then immediately fulfilling an EagerPromiseNode is a convenient
+    // way to do so.
+    auto paf = newPromiseAndFulfiller<void>();
+    paf.promise = paf.promise.eagerlyEvaluate(nullptr);
+    paf.fulfiller->fulfill();
+
+    auto promise = kj::evalLater([&]() { evalLaterRan = true; });
+    auto coroPromise = countDtorsAroundAwait(count, kj::mv(promise));
+
+    // We didn't continue through the evalLater() promise, because the background promise's
+    // continuation was next in the event loop's queue.
+    KJ_EXPECT(evalLaterRan == false);
+    // No Counter destructor has run.
+    KJ_EXPECT(count == 0, count);
+  }).eagerlyEvaluate(nullptr).wait(waitScope);
+}
+
+KJ_TEST("Exceptions propagate through layered coroutines") {
+  EventLoop loop;
+  WaitScope waitScope(loop);
+
+  auto throwy = simpleCoroutine(kj::Promise<int>(kj::NEVER_DONE), false);
+
+  KJ_EXPECT_THROW_RECOVERABLE(FAILED, simpleCoroutine(kj::mv(throwy)).wait(waitScope));
+}
+
+KJ_TEST("Exceptions before the first co_await don't escape, but reject the promise") {
+  EventLoop loop;
+  WaitScope waitScope(loop);
+
+  auto throwEarly = []() -> Promise<void> {
+    KJ_FAIL_ASSERT("test exception");
+#ifdef __GNUC__
+// Yes, this `co_return` is unreachable. But without it, this function is no longer a coroutine.
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunreachable-code"
+#endif  // __GNUC__
+    co_return;
+#ifdef __GNUC__
+#pragma GCC diagnostic pop
+#endif  // __GNUC__
+  };
+
+  auto throwy = throwEarly();
+
+  KJ_EXPECT_THROW_RECOVERABLE(FAILED, throwy.wait(waitScope));
+}
+
+KJ_TEST("Coroutines can catch exceptions from co_await") {
+  EventLoop loop;
+  WaitScope waitScope(loop);
+
+  kj::String description;
+
+  auto tryCatch = [&](kj::Promise<void> promise) -> kj::Promise<kj::String> {
+    try {
+      co_await promise;
+    } catch (const kj::Exception& exception) {
+      co_return kj::str(exception.getDescription());
+    }
+    KJ_FAIL_EXPECT("should have thrown");
+    KJ_UNREACHABLE;
+  };
+
+  {
+    // Immediately ready case.
+    auto promise = kj::Promise<void>(KJ_EXCEPTION(FAILED, "catch me"));
+    KJ_EXPECT(tryCatch(kj::mv(promise)).wait(waitScope) == "catch me");
+  }
+
+  {
+    // Ready later case.
+    auto promise = kj::evalLater([]() -> kj::Promise<void> {
+      return KJ_EXCEPTION(FAILED, "catch me");
+    });
+    KJ_EXPECT(tryCatch(kj::mv(promise)).wait(waitScope) == "catch me");
+  }
+}
+
+KJ_TEST("Coroutines can be canceled while suspended") {
+  EventLoop loop;
+  WaitScope waitScope(loop);
+
+  size_t count = 0;
+
+  auto coro = [&](kj::Promise<int> promise) -> kj::Promise<void> {
+    Counter counter1(count);
+    co_await kj::evalLater([](){});
+    Counter counter2(count);
+    co_await promise;
+  };
+
+  {
+    auto neverDone = kj::Promise<int>(kj::NEVER_DONE);
+    neverDone = neverDone.attach(kj::heap<Counter>(count));
+    auto promise = coro(kj::mv(neverDone));
+    KJ_EXPECT(!promise.poll(waitScope));
+  }
+
+  // Stack variables on both sides of a co_await, plus coroutine arguments are destroyed.
+  KJ_EXPECT(count == 3);
+}
+
+kj::Promise<void> deferredThrowCoroutine(kj::Promise<void> awaitMe) {
+  KJ_DEFER(kj::throwFatalException(KJ_EXCEPTION(FAILED, "thrown during unwind")));
+  co_await awaitMe;
+  co_return;
+};
+
+KJ_TEST("Exceptions during suspended coroutine frame-unwind propagate via destructor") {
+  EventLoop loop;
+  WaitScope waitScope(loop);
+
+  auto exception = KJ_ASSERT_NONNULL(kj::runCatchingExceptions([&]() {
+    deferredThrowCoroutine(kj::NEVER_DONE);
+  }));
+
+  KJ_EXPECT(exception.getDescription() == "thrown during unwind");
+};
+
+KJ_TEST("Exceptions during suspended coroutine frame-unwind do not cause a memory leak") {
+  EventLoop loop;
+  WaitScope waitScope(loop);
+
+  // We can't easily test for memory leaks without hooking operator new and delete. However, we can
+  // arrange for the test to crash on failure, by having the coroutine suspend at a promise that we
+  // later fulfill, thus arming the Coroutine's Event. If we fail to destroy the coroutine in this
+  // state, EventLoop will throw on destruction because it can still see the Event in its list.
+
+  auto exception = KJ_ASSERT_NONNULL(kj::runCatchingExceptions([&]() {
+    auto paf = kj::newPromiseAndFulfiller<void>();
+
+    auto coroPromise = deferredThrowCoroutine(kj::mv(paf.promise));
+
+    // Arm the Coroutine's Event.
+    paf.fulfiller->fulfill();
+
+    // If destroying `coroPromise` does not run ~Event(), then ~EventLoop() will crash later.
+  }));
+
+  KJ_EXPECT(exception.getDescription() == "thrown during unwind");
+};
+
+KJ_TEST("Exceptions during completed coroutine frame-unwind propagate via returned Promise") {
+  EventLoop loop;
+  WaitScope waitScope(loop);
+
+  {
+    // First, prove that exceptions don't escape the destructor of a completed coroutine.
+    auto promise = deferredThrowCoroutine(kj::READY_NOW);
+    KJ_EXPECT(promise.poll(waitScope));
+  }
+
+  {
+    // Next, prove that they show up via the returned Promise.
+    auto promise = deferredThrowCoroutine(kj::READY_NOW);
+    KJ_EXPECT_THROW_RECOVERABLE_MESSAGE("thrown during unwind", promise.wait(waitScope));
+  }
+}
+
+KJ_TEST("Coroutine destruction exceptions are ignored if there is another exception in flight") {
+  EventLoop loop;
+  WaitScope waitScope(loop);
+
+  auto exception = KJ_ASSERT_NONNULL(kj::runCatchingExceptions([&]() {
+    auto promise = deferredThrowCoroutine(kj::NEVER_DONE);
+    kj::throwFatalException(KJ_EXCEPTION(FAILED, "thrown before destroying throwy promise"));
+  }));
+
+  KJ_EXPECT(exception.getDescription() == "thrown before destroying throwy promise");
+}
+
+KJ_TEST("co_await only sees coroutine destruction exceptions if promise was not rejected") {
+  EventLoop loop;
+  WaitScope waitScope(loop);
+
+  // throwyDtorPromise is an immediate void promise that will throw when it's destroyed, which
+  // we expect to be able to catch from a coroutine which co_awaits it.
+  auto throwyDtorPromise = kj::Promise<void>(kj::READY_NOW)
+      .attach(kj::defer([]() {
+    kj::throwFatalException(KJ_EXCEPTION(FAILED, "thrown during unwind"));
+  }));
+
+  // rejectedThrowyDtorPromise is a rejected promise. When co_awaited in a coroutine,
+  // Awaiter::await_resume() will throw that exception for us to catch, but before we can catch it,
+  // the temporary promise will be destroyed. The exception it throws during unwind will be ignored,
+  // and the caller of the coroutine will see only the "thrown during execution" exception.
+  auto rejectedThrowyDtorPromise = kj::evalNow([&]() -> kj::Promise<void> {
+    kj::throwFatalException(KJ_EXCEPTION(FAILED, "thrown during execution"));
+  }).attach(kj::defer([]() {
+    kj::throwFatalException(KJ_EXCEPTION(FAILED, "thrown during unwind"));
+  }));
+
+  auto awaitPromise = [](kj::Promise<void> promise) -> kj::Promise<void> {
+    co_await promise;
+  };
+
+  KJ_EXPECT_THROW_MESSAGE("thrown during unwind",
+      awaitPromise(kj::mv(throwyDtorPromise)).wait(waitScope));
+
+  KJ_EXPECT_THROW_MESSAGE("thrown during execution",
+      awaitPromise(kj::mv(rejectedThrowyDtorPromise)).wait(waitScope));
+}
+
+uint countLines(StringPtr s) {
+  uint lines = 0;
+  for (char c: s) {
+    lines += c == '\n';
+  }
+  return lines;
+}
+
+#if !_MSC_VER || defined(__clang__)
+// TODO(msvc): This test relies on GetFunctorStartAddress, which is not supported on MSVC currently,
+//   so skip the test.
+KJ_TEST("Can trace through coroutines") {
+  // This verifies that async traces, generated either from promises or from events, can see through
+  // coroutines.
+  //
+  // This test may be a bit brittle because it depends on specific trace counts.
+
+  // Enable stack traces, even in release mode.
+  class EnableFullStackTrace: public ExceptionCallback {
+  public:
+    StackTraceMode stackTraceMode() override { return StackTraceMode::FULL; }
+  };
+  EnableFullStackTrace exceptionCallback;
+
+  EventLoop loop;
+  WaitScope waitScope(loop);
+
+  auto paf = newPromiseAndFulfiller<void>();
+
+  // Get an async trace when the promise is fulfilled. We eagerlyEvaluate() to make sure the
+  // continuation executes while the event loop is running.
+  paf.promise = paf.promise.then([]() {
+    auto trace = getAsyncTrace();
+    // We expect one entry for waitImpl(), one for the coroutine, and one for this continuation.
+    // When building in debug mode with CMake, I observed this count can be 2. The missing frame is
+    // probably this continuation. Let's just expect a range.
+    auto count = countLines(trace);
+    KJ_EXPECT(0 < count && count <= 3);
+  }).eagerlyEvaluate(nullptr);
+
+  auto coroPromise = [&]() -> kj::Promise<void> {
+    co_await paf.promise;
+  }();
+
+  {
+    auto trace = coroPromise.trace();
+    // One for the Coroutine PromiseNode, one for paf.promise.
+    KJ_EXPECT(countLines(trace) >= 2);
+  }
+
+  paf.fulfiller->fulfill();
+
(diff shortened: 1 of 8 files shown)


c++/src/kj/async-inl.h
```diff
@@ -1470,4 +1470,290 @@ PromiseForResult<Func, void> Executor::executeAsync(Func&& func) const {
...
+    // not have quite the same semantics.
+
+    // Check for exceptions. We could let Awaiter::await_resume() do it, but then we'd resume the
+    // coroutine just for it to rethrow, which seems pointless.
+    KJ_IF_MAYBE(exception, result->exception) {
+      nullifyPointers();
+      fulfiller.reject(kj::mv(*exception));
+      // The coroutine threw an exception. Our only choice is to destroy it.
+      //
+      // TODO(now): Should we ignore exceptions from destroy()?
```
harrishancock, line 1699: I'm not quite sure about this. The situation would arise with a coroutine like this:

```c++
kj::Promise<void> coro() {
  KJ_DEFER(KJ_FAIL_ASSERT("I'll throw from coroutine.destroy() (second)"));
  co_await kj::Promise<void>(KJ_EXCEPTION(FAILED, "I'll reject the returned promise (first)"));
  KJ_UNREACHABLE;
}
```

I suppose the equivalent .then() based code can't do anything particularly special to avoid the situation, and the deferred assertion would end up getting thrown.

harrishancock:
Amended with just a couple minor changes to async-coroutine-test.c++ (copyright notice, KJ_FAIL_ASSERT instead of KJ_ASSERT(false)).

c++/src/kj/async-prelude.h
```diff
@@ -27,6 +27,20 @@
 #include "exception.h"
 #include "tuple.h"
 
+// Detect whether or not we should enable kj::Promise<T> coroutine integration.
+#ifdef __has_include
+// For now, we only support the Coroutines TS.
+//
+// TODO(someday): Also support standardized C++20 Coroutines. The latest VS2019 and GCC 10 both have
+//   support, though MSVC hides it behind /std:c++latest, which brings an ICE with it.
+#if __has_include(<experimental/coroutine>) && defined(__cpp_coroutines) && __cpp_coroutines
```
kentonv, line 36: Maybe put the `__has_include` last since it presumably hits the filesystem which makes it a lot slower than the other checks.

Also, nitpick, in KJ we don't generally check `defined(X)` if we're going to then check the truthiness of `X`, since the preprocessor will default it to 0 anyawy.

c++/src/kj/async-inl.h
```diff
@@ -1470,4 +1470,290 @@ PromiseForResult<Func, void> Executor::executeAsync(Func&& func) const {
...
+  // These adjust the suspension behavior of coroutines immediately upon initiation, and immediately
+  // after completion.
+  //
+  // The initial suspension point could allow us to defer the initial execution of a coroutine with
+  // evalLater(), for instance.
+  //
+  // The final suspension point could be useful to delay deallocation of the coroutine frame to
+  // match the lifetime of the enclosing promise.
+
+  void unhandled_exception() {
```
kentonv, line 1577: I guess we should wrap this in `#if !KJ_NO_EXCEPTIONS`? (Or otherwise wrap the entire coro implementation in it.)

c++/src/kj/async-coroutine-test.c++
```diff
@@ -0,0 +1,183 @@
...
+  struct Counter {
+    size_t& count;
+    Counter(size_t& count): count(count) {}
+    ~Counter() { ++count; }
+    KJ_DISALLOW_COPY(Counter);
+  };
+
+  {
+    auto neverDone = kj::Promise<int>(kj::NEVER_DONE);
+    neverDone = neverDone.attach(kj::heap<Counter>(count));
```
kentonv, line 108: This verifies that the `neverDone` promise is itself canceled. Great!

It'd also be nice to verify that local variables declared in the coroutine are destroyed on cancellation.

c++/src/kj/async-inl.h
```diff
@@ -1470,4 +1470,290 @@ PromiseForResult<Func, void> Executor::executeAsync(Func&& func) const {
...
+    // the node's result in `result`.
+
+    // We'll need to save these references for fire(). Note that Awaiter is movable, and `result` is
+    // its subobject, but it is safe to store a reference to `result` in Adapter, because by the
+    // time `await_suspend()` is called, Awaiter has already been returned from `await_transform()`
+    // and stored in its final resting place in the coroutine's frame.
+    node = &nodeParam;
+    result = &resultParam;
+
+    // We use the promise adapter as an Event, because we must be able to destroy the coroutine from
```
kentonv, line 1665: Technically `fire()` callbacks _do_ have the capability to delete themselves, by returning an `Own<Event>` pointing to themselves, which the caller drops shortly thereafter. I suppose you'd need a custom disposer but it looks like it should work.

Would there be any advantage to doing it that way?

c++/src/kj/async-inl.h
```diff
@@ -1470,4 +1470,290 @@ PromiseForResult<Func, void> Executor::executeAsync(Func&& func) const {
...
+    static_cast<Self*>(this)->adapter->fulfiller.fulfill();
+  }
+};
+// The Coroutines spec has no `_::FixVoid<T>` equivalent to unify valueful and valueless co_return
+// statements, and programs are ill-formed if the coroutine implementation object (Coroutine<T>) has
+// both a `return_value()` and `return_void()`. No amount of EnableIffery can get around it, so
+// these return_* functions live in a CRTP mixin.
+
+template <typename T>
+class Coroutine<T>::Adapter: private Event {
```
kentonv, line 1631: Instead of using promise adapters, it may make sense to implement `PromiseNode` directly. All `newAdaptedPromise()` really does is wrap your class in an `AdapterPromiseNode` which is just some boilerplate to make it a friendlier public interface.

In fact, I think then you could possibly even make `Coroutine<T>` implement `PromiseNode` directly and avoid having to allocate the adapter separately.

c++/src/kj/async-inl.h
```diff
@@ -1470,4 +1470,290 @@ PromiseForResult<Func, void> Executor::executeAsync(Func&& func) const {
...
+
+  PromiseFulfiller<T>& fulfiller;
+  // Used to fulfill/reject the enclosing coroutine's kj::Promise<T>.
+
+private:
+  Maybe<Own<Event>> fire() override {
+    // The promise this coroutine is currently waiting on is ready. Extract its result and check for
+    // exceptions.
+
+    // Destroying or resuming the coroutine will leave these pointers dangling, so we call this
```
kentonv, line 1679: I guess this is just a safety measure? The pointers won't actually be used for anything, right?

I am conflicted. On one hand it's just two stores, on the other hand any code here is going to be specialized many times driving a lot of code bloat.

c++/src/kj/async-inl.h
kentonv: Actually, the equivalent `.then()` code does explicitly handle the situation where a PromiseNode both yielded an exception and threw from its destructor. The yielded exception takes priority, and the destructor exception is ignored. See `TransformPromiseNodeBase::getDepResult()`.

I think we should do the same here.

Essentially what we have here is a destructor-threw-during-unwind situation. Of course, in regular sync code, this results in std::terminate(), C++'s single worst design decision. I think, in general, these exceptions-thrown-by-destructors-during-unwind are usually "secondary" exceptions that are side effects of the "primary" exception, where the primary exception is the one that initiated the unwind. If we have to pick one to propagate, it should be the primary exception.

(Also, throwing from `fire()` will actually unwind out of the event loop, which we certainly don't want...)

Hmm... but I guess an interesting question: does `coroutine.destroy()` actually propagate exceptions from destructors? Considering that C++11 made `noexcept` the default for destructors, I could certainly imagine that they defined `destroy()` to be noexcept as well, which would be very very sad.

c++/src/kj/async-inl.h
```diff
@@ -1470,4 +1470,290 @@ PromiseForResult<Func, void> Executor::executeAsync(Func&& func) const {
...
+      result = nullptr;
+    };
+
+    node->get(*result);
+
+    // Note: the Promise::wait() implementation throws a recoverable exception if both a value and
+    // an exception are set. We only have visibility on the exception here, so the following might
+    // not have quite the same semantics.
+
+    // Check for exceptions. We could let Awaiter::await_resume() do it, but then we'd resume the
```
kentonv, line 1692: What if the coroutine contains a try/catch wrapping this await?

c++/src/kj/async-inl.h
```diff
@@ -1470,4 +1470,290 @@ PromiseForResult<Func, void> Executor::executeAsync(Func&& func) const {
...
+  // is settled. Once that happens, the enclosing coroutine's promise adapter resumes the coroutine,
+  // which transitively calls await_resume() to unwrap the awaited promise result.
+
+public:
+  explicit Awaiter(Promise<U> promise): promise(kj::mv(promise)) {}
+
+  bool await_ready() const { return false; }
+  // This could return "`promise.node.get()` is safe to call" instead, which would make suspension-
+  // less co_awaits possible for immediately-fulfilled promises. However, that's probably unwise,
+  // since it is explicitly not how .then() works.
```
kentonv, line 1734: Hmm, I think this situation might be genuinely different from `.then()`. With `.then()`, you have this situation:

    promise = thing.then([]() { foo(); });
    bar();

And the question is: could `foo()` sometimes run before `bar()`. We want the answer to be consistent, and the only possible consistent answer is "no".

But with coroutines there is no possibility of a `bar()`. Once you invoke `await`, no further code is executed. So we don't have this scenario.

So the only question here is whether an `await` should always yield to the event loop. But KJ already goes out of its way to run most events in a LIFO-ish way, such that an immediate `.then()` actually _won't_ allow anything else to run in between, because we don't want to context switch too much. So in that case, it seems like enabling immediate `await`s is just a performance optimization and we might as well do it.

c++/src/kj/async-inl.h
harrishancock: True, a custom disposer didn't occur to me. I think it'd work, too. I think the main advantage would be that it would get rid of the pointer twiddling I'm doing here (obviating the source of my nullifyPointers() paranoia). The main disadvantage would be that the coroutine would run in a destructor, which would make some stack traces aesthetically irritating.

To be honest, I could go either way. I think the code would be a little simpler, the stack trace would be uglier, and I don't think there'd be any performance difference.

c++/src/kj/async-inl.h
harrishancock: Yeah, it's purely a safety measure. I think it'd be reasonable to remove it, or hide it behind KJ_DEBUG.

c++/src/kj/async-inl.h
kentonv: Actually yeah, I think handling exceptions here is explicitly wrong. You always want to resume, and then in the exception case `await_resume()` should throw the exception, so that it can potentially be caught by a try/catch?

harrishancock: Aw, crap.

kentonv: Admittedly, if we could somehow know that the coroutine _doesn't_ contain a try/catch -- which we'd expect to be the case the vast majority of the time -- then what you have here is probably a nice optimization, given how atrociously slow C++ exception unwinding is. I'm almost inclined to prohibit try/catch in our coroutines for this reason.

Though, then again, it's very possible that a premature `destroy()` is literally implemented using the same unwind mechanism as exceptions, in which case, maybe it doesn't make so much difference.

kentonv:
Thanks for the clean code and very thorough comments! I can't wait to use this.

c++/src/kj/async-inl.h
kentonv: Oh I see, `.resume()` can end up destroying the coroutine with no way to stop it, huh? That creates two problems:

1. We definitely can't be executing the coroutine in the destructor of the event returned by `fire()`. It's important that application code runs _during_ `fire()`.
2. This implies that my comment below about reducing the allocations from 2 to 1 by having `Coroutine` implement `PromiseNode` won't work.

Is this where it might make sense to change `final_suspend()` to `suspend_always()`? In that case I suppose we gain full control over exactly when the coroutine is destroyed? And then we'd have it happen strictly when the caller drops the `PromiseNode`, which seems like it simplifies a lot.

c++/src/kj/async-inl.h
harrishancock: I was thinking about this, and with -fno-exceptions, this incorrect optimization would be the required code path. But yeah, I think for the common case I'll need to move exception handling to throw from await_resume().

c++/src/kj/async-inl.h
harrishancock: > Is this where it might make sense to change final_suspend() to suspend_always()?

Yup, I think that will be required to support a PromiseNode implementation. I like the idea a lot -- the Promise will just be a pointer to an object on the coroutine frame. It also makes our terminology and the standard's terminology make more sense together ("a kj::Promise<T>'s promise_type is PromiseNode").

> We definitely can't be executing the coroutine in the destructor of the event returned by fire()

I think this means that the promise adapter (soon the PromiseNode) will need to remain the Event, rather than the Awaiters. While we can prevent `coroutine.resume()` from destroying the coroutine, we can't prevent it from destroying the Awaiter, which is essentially a temporary object kept alive for the duration of the suspension.

c++/src/kj/async-inl.h
harrishancock: Fortunately, `coroutine.destroy()` is not noexcept. Destruction semantics aren't explicitly called out, though: http://eel.is/c++draft/dcl.fct.def.coroutine#11. I'll test to see what happens if an object on the frame throws.

I suppose the code we're commenting on here won't exist soon: since we're changing to a PromiseNode implementation, I _think_ we'll be calling coroutine.destroy() exclusively from the PromiseNode's disposer, rather than here, which might change what the right answer is.

c++/src/kj/async-inl.h
harrishancock: Makes sense to me! And it would certainly be a nice optimization.

c++/src/kj/async-inl.h
kentonv: Ah, yeah, I didn't ever imagine the awaiter would be the event... I imagined the `Coroutine<T>` would be the event. I think that's the plan now, right? `Coroutine<T>` will become the `PromiseNode` (subsuming the adapter's role) and therefore also the `Event`?

harrishancock: Oh, I get it! Haha, yes, that's the plan now.

c++/src/kj/async-inl.h
kentonv: True. But the issue might still apply on the awaiter side -- if an awaited promise resolves to an exception, and then destroying the PromiseNode also throws, you want to discard the latter exception and propagate the former.

c++/src/kj/async-prelude.h
harrishancock: Done: https://github.com/capnproto/capnproto/pull/1104/commits/58e0eeffdfe32d033d9617cc9ea5f6cd29902431

c++/src/kj/async-inl.h
harrishancock: Removed: https://github.com/capnproto/capnproto/pull/1104/commits/8b8850f96cb9a43bb7e098578abd3a528975aca8
And then completely obviated when I gave up trying to support -fno-exceptions: https://github.com/capnproto/capnproto/pull/1104/commits/23721723d988ea9a523e10e3543b6f5a3d7f0f18

c++/src/kj/async-inl.h
harrishancock: Fixed: https://github.com/capnproto/capnproto/pull/1104/commits/5a9b13e4f9cd3312ad5f4ddad7c5f894d4ac5797

c++/src/kj/async-coroutine-test.c++
harrishancock: Test here: https://github.com/capnproto/capnproto/pull/1104/commits/9407ece3275c5fafb2561f62283155e0484bcd07

c++/src/kj/async-inl.h
harrishancock: Converted to PromiseNode: https://github.com/capnproto/capnproto/pull/1104/commits/9428a2532df76491831ed96af54b18b577e88813
Made Coroutine<T> the PromiseNode: https://github.com/capnproto/capnproto/pull/1104/commits/99c2cd5c86b4d4660cc2133f2c8c9a701bcc4264

c++/src/kj/async-inl.h
harrishancock: Coroutine destruction exceptions handled here: https://github.com/capnproto/capnproto/pull/1104/commits/0e7c329e0d815d76d4d7815a538ce8adb741400f
And awaited promise destruction exceptions handled here: https://github.com/capnproto/capnproto/pull/1104/commits/a1df89a3a276ac2c763cd0bddf9e9622df3ee531

c++/src/kj/async-inl.h
harrishancock: Optimized: https://github.com/capnproto/capnproto/pull/1104/commits/b3c38109b0037493798ec11c24b5b6718669da64

However, I think that has a bug: `evalLater()` promises turn out to be immediately ready, so we breeze right through them. Since we frequently use `evalLater()` to mean "I really do want to yield to the event loop here," I figure we should come up with a solution to that.

c++/src/kj/async-inl.h
harrishancock: I tried to make coroutines work with -fno-exceptions, but got confused and tired and gave up: https://github.com/capnproto/capnproto/pull/1104/commits/23721723d988ea9a523e10e3543b6f5a3d7f0f18

I just gated the entire KJ_HAS_COROUTINE define behind !KJ_NO_EXCEPTIONS.

c++/src/kj/async-coroutine-test.c++
```diff
@@ -65,6 +66,70 @@ KJ_TEST("Simple coroutine test") {
...
+  WaitScope waitScope(loop);
+
+  size_t count = 0;
+  bool evalLaterRan = false;
+
+  auto promise = kj::evalLater([&]() { evalLaterRan = true; });
+  auto coroPromise = countDtorsAroundAwait(count, kj::mv(promise));
+
+  KJ_EXPECT(evalLaterRan == false);
+  KJ_EXPECT(count == 1, count);
```
harrishancock, line 130: These are true and 2, respectively, because evalLater() promises are actually immediately ready. I'm not sure what to do  about it, though.

harrishancock:
Ready for re-review, PTAL.

c++/src/kj/async-coroutine-test.c++
```diff
@@ -143,13 +143,82 @@ KJ_TEST("Coroutines can be canceled while suspended") {
...
+  KJ_EXPECT(exception.getDescription() == "thrown during unwind");
+};
+
+KJ_TEST("Exceptions during suspended coroutine frame-unwind do not cause a memory leak") {
+  EventLoop loop;
+  WaitScope waitScope(loop);
+
+  // We can't easily test for memory leaks without hooking operator new and delete. However, we can
+  // arrange for the test to crash on failure, by having the coroutine suspend at a promise that we
+  // later fulfill, thus arming the Coroutine's Event. If we destroy the coroutine in this state,
```
harrishancock, line 234: "If we *fail* to destroy the coroutine ..." I mean.

c++/src/kj/async.c++
```diff
@@ -2313,6 +2317,15 @@ ArrayJoinPromiseNodeBase::ArrayJoinPromiseNodeBase(
   }
   branches = builder.finish();
 
+  // Optimization: if all of our branches are ready, then we're ready, too. We can make this happen
+  // by firing off any already-armed branches right away.
```
harrishancock, line 2321: I suppose it might be better to do this in Branch's constructor.

c++/src/kj/async-inl.h
kentonv: Hmm, I think you could solve this by changing `isArmed()` to `isNext()`, which would return true if the event is armed _and_ is next in line to be fired. In that case there's certainly no semantic difference.

c++/src/kj/async.c++
```diff
@@ -2313,6 +2317,15 @@ ArrayJoinPromiseNodeBase::ArrayJoinPromiseNodeBase(
   }
   branches = builder.finish();
 
+  // Optimization: if all of our branches are ready, then we're ready, too. We can make this happen
```
kentonv, line 2320: If you're looking for a place to apply the optimization inside the existing KJ promise logic, a good place to try would be `ChainPromiseNode` (which handles reducing `Promise<Promise<T>>` to `Promise<T>`). Though you'll definitely need to switch to `isNext()` instead of `isArmed()` per my other comment.

Optimizing there might actually be a significant win... would be interesting to test.

Meanwhile trying to optimize `ArrayJoinPromiseNode` probably doesn't make sense anymore with `isNext()`.

c++/src/kj/async-inl.h
```diff
@@ -1470,4 +1474,413 @@ PromiseForResult<Func, void> Executor::executeAsync(Func&& func) const {
...
+
+  void disposeImpl(void* pointer) const override {
+    KJ_IASSERT(pointer == this);
+
+    // const_cast okay -- every Own<PromiseNode> that we build in get_return_object() uses itself
+    // as the disposer, thus every disposer is unique and there are no thread-safety concerns.
+    const_cast<Coroutine&>(*this).destroy();
+  }
+
+  void destroy() {
```
kentonv, line 1581: There's a lot of logic here that doesn't seem like it needs to be specialized for every promise type. Should we factor out a non-template base class to put these in? Also for `Awaiter`.

c++/src/kj/async-inl.h
```diff
@@ -1470,4 +1474,413 @@ PromiseForResult<Func, void> Executor::executeAsync(Func&& func) const {
...
+  }
+};
+// The Coroutines spec has no `_::FixVoid<T>` equivalent to unify valueful and valueless co_return
+// statements, and programs are ill-formed if the coroutine implementation object (Coroutine<T>) has
+// both a `return_value()` and `return_void()`. No amount of EnableIffery can get around it, so
+// these return_* functions live in a CRTP mixin.
+
+static stdcoro::coroutine_handle<void> noopCoroutine() noexcept;
+// TODO(msvc): Coroutines TS implementation doesn't have std::experimental::noop_coroutine(), so
+//   we'll roll our own below. This function is static so we can keep libkj coroutines header-only.
```
kentonv, line 1781: Hmm, why do we want it to be header-only?

c++/src/kj/async-inl.h
```diff
@@ -1470,4 +1474,413 @@ PromiseForResult<Func, void> Executor::executeAsync(Func&& func) const {
...
+    if (coroutineEvent.isArmed()) {
+      // The result is immediately ready! Let's cancel our event.
+      coroutineEvent.disarm();
+
+      // We can resume ourselves by returning our own handle. This accomplishes the same thing as if
+      // we had extracted returned true from await_ready().
+      return coroutine;
+    } else {
+      // Otherwise, we must suspend, so we return a dummy coroutine_handle.
+      return noopCoroutine();
```
kentonv, line 1841: Reading the docs, it looks like you could instead define `await_suspend` to return `bool`, and in this case `return true` to return control to the place where the coroutine was last resumed. (Whereas the other branch can `return false` to resume the coroutine immediately.)

Then I guess you don't need `noopCoroutine()`?

c++/src/kj/async-inl.h
harrishancock: My thinking here was that it'd be nice if a dependent project can use coroutines without worrying about whether their capnproto package was built with or without `-fcoroutines-ts`. But I think your comment about `await_suspend()` returning bool might obviate the concern (for now).

c++/src/kj/async.c++
harrishancock: I don't quite follow here. Consider a simplified version of the test case I wrote:
```c++
auto immediatePromise = kj::Promise<void>(kj::READY_NOW);
auto joinPromise = kj::joinPromises(kj::arr(kj::mv(immediatePromise)));
co_await joinPromise;
```
Ideally, we should see that the immediate promise is ready, and thus the join promise is also ready. However, no `ChainPromiseNode` is involved here -- only an `ImmediatePromiseNode` and an `ArrayJoinPromiseNode`. It might make sense to optimize `ChainPromiseNode` as well, but I'm not sure how we can get away from modifying `ArrayJoinPromiseNode`.

kentonv: I wasn't saying that optimizing `ChainPromiseNode` would have any effect on `ArrayJoinPromiseNode`. What I was trying to say is that `ArrayJoinProimseNode` doesn't seem like a very interesting thing to optimize as it's rarely used anyway, whereas `ChainPromiseNode` is used everywhere.

c++/src/kj/async-inl.h
```diff
@@ -1470,4 +1474,413 @@ PromiseForResult<Func, void> Executor::executeAsync(Func&& func) const {
...
+    } while (!disposalResults.destructorRan);
+
+    // WARNING: `this` is now a dangling pointer.
+
+    KJ_IF_MAYBE(exception, disposalResults.exception) {
+      if (shouldRethrow) {
+        kj::throwFatalException(kj::mv(*exception));
+      } else {
+        // An exception is already unwinding the stack, so throwing this secondary exception would
+        // call std::terminate().
```
vlovich, line 1619: Would using C++ nested exceptions or adding extra machinery into KJ for that be helpful?

kentonv: Sadly, no.

C++ nested exceptions are not intended for this. They are intended to express an exception that is "caused by" another exception. You have to have first _caught_ the "cause" exception, and then you can construct a new exception which has the "cause" as a nested exception. Here, we've detected that an exception is in flight, but it hasn't been caught yet.

I would love for KJ to find a way to magically resolve the throw-during-destructor-during-unwind problem, as the C++ default behavior is awful. But my research indicates that the only way to do this is to modify the C++ exception machinery itself, i.e. change libgcc or whatever implements the C++ ABI. This is a bit hard for a library to do.

vlovich:
Yeah, that was my intuition too but I wasn’t 100% certain. Maybe we should
push the standards body to let you replace the exception in-flight. How
hard could that be to implement :P.

kentonv:
> Maybe we should push the standards body to let you replace the exception in-flight

That's actually almost never what you want. Usually, if an exception is thrown during a destructor that is running during unwind due to another exception... the new exception is just a side-effect of the original, and might as well be discarded. The original exception (that caused the unwind that ran the destructor) is the real error that you want reported.

vlovich:
By "replace" I didn't mean replace with the new exception. I meant replace with a nested exception (still pointing to the original exception) so that you don't lose the exceptions generated by unwind.

c++/src/kj/async-inl.h
harrishancock: Fixed: https://github.com/capnproto/capnproto/pull/1104/commits/3cd44f37fb74b71014a7ba994577439679f95947

c++/src/kj/async-coroutine-test.c++
harrishancock: https://github.com/capnproto/capnproto/pull/1104/commits/c3e2dcec30de0f07500ed49faee855d137e0dfbf

c++/src/kj/async.c++
harrishancock: I eventually figured out how to add this to ChainPromiseNode (and a few others): https://github.com/capnproto/capnproto/pull/1104/commits/c03e519c9c1221443361c2db351ad2f2e4e48b32

One of the keys was to change `isNext()` to return false if the event loop wasn't running, otherwise a bunch of RPC tests broke in mysterious ways.

c++/src/kj/async-inl.h
```diff
@@ -2126,8 +2126,24 @@ class Coroutine<T>::Awaiter {
     return U(kj::mv(*value));
   }
 
-  void await_suspend(Coroutine::Handle coroutine) {
-    coroutine.promise().awaitSuspend(PromiseNode::from(promise), result);
+  bool await_suspend(Coroutine::Handle coroutine) {
+    auto& node = PromiseNode::from(promise);
+    auto& coroutineEvent = coroutine.promise();
+
+  // TODO(now): Should we call node.setSelfPointer() here?
```
harrishancock, line 2162: @kentonv I noticed most other places call setSelfPointer() immediately before onReady(). Should we do that here?

harrishancock:
I cleaned this up and addressed all review comments except for https://github.com/capnproto/capnproto/pull/1104#discussion_r532965137. I also added implementations for `tracePromise()`/`traceEvent()` which I guess were added after I first made this PR.

In addition to the last outstanding review comment, there are two outstanding TODO(now)s:
- a setSelfPointer() call in await_suspend()?
- async stack trace unit test

c++/src/kj/async.c++
harrishancock: This caused a stack overflow in the MSVC build in a Async/DeepChain4, which creates a 1000-long chain of promises. Presumably onReady() is calling fire() is calling onReady(), etc.

harrishancock: Optimizing ChainPromiseNode is difficult, and I'm not really sure what else to try here. I could imagine `co_await` running a little microtask queue or something, but this seems rather involved. I'm going to just leave a TODO(perf) for now.

harrishancock:
> * a setSelfPointer() call in await_suspend()?
> * async stack trace unit test

`setSelfPointer()`: https://github.com/capnproto/capnproto/pull/1104/commits/e6ea70f4a6d6ca089aac08e0064e2cf7c530bbee
async stack trace unit test: https://github.com/capnproto/capnproto/pull/1104/commits/f90f3758d105c3a7546a02237ff0f1725dcbf0f1

c++/src/kj/async.c++
harrishancock: Removed here: https://github.com/capnproto/capnproto/pull/1104/commits/a38a1c04aebca6cfb01726f502258ece23c53ea0

c++/src/kj/async-inl.h
harrishancock: https://github.com/capnproto/capnproto/pull/1104/commits/f97587b67a1423e4d188d1b6adb7658d8c906aa5, https://github.com/capnproto/capnproto/pull/1104/commits/d37e57cfa6a3b3ccddffe89d0a483f2b32c65904, and https://github.com/capnproto/capnproto/pull/1104/commits/d8fd23da384819150dcb6337e6e2d9a1adf865db

harrishancock:
@kentonv I think this is ready for re-review.

c++/src/kj/async.c++
```diff
@@ -2696,4 +2742,202 @@ namespace _ {  // (private)
...
+    }
+  }
+}
+
+CoroutineBase::AwaiterBase::AwaiterBase(Own<PromiseNode> node): node(kj::mv(node)) {}
+CoroutineBase::AwaiterBase::AwaiterBase(AwaiterBase&&) = default;
+CoroutineBase::AwaiterBase::~AwaiterBase() noexcept(false) {
+  // Make sure it's safe to generate an async stack trace between now and when the Coroutine is
+  // destroyed.
+  KJ_IF_MAYBE(coroutineEvent, maybeCoroutineEvent) {
```
harrishancock, line 2900: I realized I forgot to check if `this` was in a moved-from state here. But, I guess the Maybe move constructor's new behavior in leaving the old object empty makes this fine?

c++/src/kj/async-inl.h
```diff
@@ -166,6 +166,15 @@ class Event {
   // Enqueues this event to happen after all other events have run to completion and there is
   // really nothing left to do except wait for I/O.
 
+  bool isNext();
```
kentonv, line 169: I think there's an additional rule around `isNext()` which some of your usages and "TODO"s are missing: It is not safe to rely on `isNext()` unless the code is about to return to the event loop anyway. In particular, it's never safe to rely on `isNext()` while application code is on the stack, because:

1. The application code might manipulate the event loop further before returning to it.
2. If an `if (isNext()) fire();` results in executing some application callback, the application may not be expecting this callback to run synchronously.

Essentially, you can only do `if (isNext()) fire();` inside some other `fire()` callback, or something directly called by some other `fire()` without entering application code.

I believe your optimization for coroutines satisfies this, since if you suspended the coroutine, control would in fact be passed back to the `fire()` function which would then return to the event loop.

For `ChainPromiseNode`, the place you should be applying the optimization is inside its `fire()` method, not inside the constructor nor `onReady()`, both of which are potentially invoked with application code on the stack.

Then again, this means the optimization's usefulness is pretty narrow: If you can only really use it inside `fire()`, but returning from `fire()` is already going to immediately loop to the next event, then how much are you really saving? In the case of coroutines, we're saving the expense of suspending the coroutine, which might be worth it (I have no idea how expensive that is...).

vlovich:
I tried building this with G++ 11 & ran into an ICE which cvise reduced to: https://godbolt.org/z/M5nfEPMKj
When I build the reduced test case with clang: https://godbolt.org/z/KjcW94oKM
it says
```
<source>:57:13: error: this function cannot be a coroutine: 'std::experimental::coroutine_traits<PromiseBase>' has no member named 'promise_type'
PromiseBase httpClient() {
```

This does build with clang proper against libc++ so I'm not sure what the problem with the reduced test case might be...

vlovich:
Given that MSVC also craps out with an ICE, I'm wondering if there isn't a bug somewhere here that just happens to work in clang's TS implementation, but I don't know the code well enough. @harrishancock does the error message that clang gives on the reduced test case perhaps ringing a bell for you on what the fix might be?

vlovich:
So in release mode the test fails.
```
kj/async-coroutine-test.c++:484: failed: expected countLines(trace) >= 2 [0 >= 2]
stack: 471271 466b31 6b9649 6be5e4 6b8dad 6b8b8c 6cb85f 6be5e4 6c9f12 6b80ee 7f240f831b24 40c16d
kj/async-coroutine-test.c++:474: failed: expected countLines(trace) == 3 [0 == 3]
stack: 471271 46722b 467525 6695b1 6be5e4 664968 6676d4 6678cc 6633c7 466b90 6b9649 6be5e4 6b8dad 6b8b8c 6cb85f 6be5e4 6c9f12 6b80ee 7f240f831b24 40c16d
```

harrishancock:
> This does build with clang proper against libc++ so I'm not sure what the problem with the reduced test case might be...

The reduced test case is mixing C++20 coroutines (`std` namespace) and Coroutines TS coroutines (`std::experimental` namespace). It includes <experimental/coroutine>, but refers to `std::coroutine_handle<>`. Also, `coroutine_traits` is defined (in the wrong namespace), whereas it should be explicitly specialized, e.g.

```
namespace std::experimental {
template <class... Args> struct coroutine_traits<PromiseBase, Args...> {
  using promise_type = kj::_::Coroutine<int>;
};
} // namespace std
```

(The variadic part might not be strictly necessary.)

With the namespaces changed and `coroutine_traits` specialized, clang starts giving more recognizable errors, like "you need to define `return_value()`/`return_void()`", which you'd need to work through.

> Given that MSVC also craps out with an ICE

Using C++20 coroutines, or Coroutines TS? I developed this against MSVC and observed that its Coroutines TS implementation worked, and its C++20 implementation did not (see the last paragraph in the PR description).

> Didn't want to take over the PR as that felt rude.

Feel free to do so, or to submit a separate PR (and I'll close this one). Realistically I won't have much bandwidth to work on this until October.

harrishancock:
> So in release mode the test fails.

Those particular parts of the test might only be exercising things that exist in debug mode. `Promise::trace()` has a comment sort of to that effect, though I would have expected `getAsyncTrace()` to work.

vlovich:
Thanks for the feedback. I wonder why the TS is being mixed with the real coroutines... likely a bug in libstdc++. What's weird is that it doesn't have the TS header for import which is why I had to try real coroutines.

I'll absorb my last commit, fix the test and upload over your branch then since this seems like the last thing. I want ma coroutines now and you've already done all the hard work :).

kentonv:
I would prefer to leave this in @harrishancock's hands and this is a very complicated PR that he and I have spent a lot of time on, and I'm not excited about repeating a lot of work to bring a new person up to speed on it.

c++/src/kj/async-inl.h
harrishancock: Took me a bit to wrap my head around all this, but I agree. I backed out the optimization attempt here: https://github.com/capnproto/capnproto/pull/1104/commits/029e11883fd5e4fce2e795ce9c45be9f3eef0acc

harrishancock:
Okay, I picked this up and I think I got it into a mergeable state.

- I removed the optimization attempts, including most of the optimistic comments -- it seems like the optimization only makes sense in coroutine suspension, and that's really about it. @vlovich I ended up not using your commit, sorry. :( I thank you for the impetus to get out of the doldrums, though.
- Fixed the `getAsyncTrace()` test in release mode, which @vlovich discovered.
- `super-test.sh clang` now tests `-fcoroutines-ts` separately, so we should have at least some CI coverage.
- Added a little documentation in kjdoc/tour.md.

I'm going to re-review everything now. :P

harrishancock:
Okay, I re-reviewed everything, added one last commit to tweak comments, and am happy with how it turned out.

If someone has a handy Windows VM to test this on, that'd be great; otherwise, I think we can squash and merge.

kentonv:
As discussed, let's make super-test.sh always enable coroutines-ts when using clang, rather than do a separate build. That way it is tested with quick-tests and also combined with all the other flags.

harrishancock:
> As discussed, let's make super-test.sh always enable coroutines-ts when using clang, rather than do a separate build. That way it is tested with quick-tests and also combined with all the other flags.

Done: https://github.com/capnproto/capnproto/pull/1104/commits/db2070a6c85408cae796311a8066a3fb2baea13c Only our clang-10 CI run worked out of the box, so I limited the logic to clang 10+.

I'm going to squash up.

harrishancock:
Squashed into a single commit. None of the intermediate commits seemed worth keeping in the history, since it was mostly incremental development.

PR #1105, opened 2020-11-16
Change Canceler's default exception type to DISCONNECTED

bretthoerner:
I wasn't sure what the message should actually say...

### DIFF
--- a/c++/src/kj/async.c++
+++ b/c++/src/kj/async.c++
@@ -136,7 +136,7 @@ Canceler::~Canceler() noexcept(false) {
 
 void Canceler::cancel(StringPtr cancelReason) {
   if (isEmpty()) return;
-  cancel(Exception(Exception::Type::FAILED, __FILE__, __LINE__, kj::str(cancelReason)));
+  cancel(Exception(Exception::Type::DISCONNECTED, __FILE__, __LINE__, kj::str(cancelReason)));
 }
 
 void Canceler::cancel(const Exception& exception) {


a-robinson (approved):

c++/src/kj/async-io.c++
```diff
@@ -1123,6 +1123,7 @@ private:
     }
 
     ~BlockedPumpTo() noexcept(false) {
+      canceler.cancel(KJ_EXCEPTION(DISCONNECTED, "BlockedPumpTo was destroyed"));
```
kentonv, line 1126: There's a lot of other cancelers in use in this file that present similar problems.

Maybe we should just make DISCONNECTED the default exception type for `Canceler` in general?

c++/src/kj/async.c++
```diff
@@ -131,7 +131,7 @@ public:
 // =======================================================================================
 
 Canceler::~Canceler() noexcept(false) {
-  cancel("operation canceled");
+  cancel(KJ_EXCEPTION(DISCONNECTED, "operation canceled"));
```
kentonv, line 134: Since this line does allocation now, it should probably be guarded by `if (!isEmpty())` so that it doesn't allocate when `cancel()` would be a no-op anyway.

c++/src/kj/async.c++
```diff
@@ -131,7 +131,7 @@ public:
...
 }
 
 void Canceler::cancel(StringPtr cancelReason) {
```
bretthoerner, line 137: Hmm, I wonder if the exception type here should change too?

kentonv: Actually... Yeah, probably we should not change the destructor, but should instead change the implementation of `cancel(StringPtr)`...

bretthoerner: What's the thinking behind them being different? Is it that developers should explicitly `cancel(StringPtr)` and canceling via the destructor is considered a bug or?

And we'd be back to picking a string to use in `BlockedPumpTo`'s `cancel` (which is fine).

PR #1108, opened 2020-11-19
Fix a cross-thread event bug and add some asserts

kentonv:
See individual commits.


harrishancock (approved):

harrishancock:
MSVC failed because https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

PR #1109, opened 2020-11-19
Split very large byte-stream writes into smaller writes

bretthoerner:


c++/src/capnp/compat/byte-stream-test.c++
```diff
@@ -58,9 +58,13 @@ KJ_TEST("KJ -> ByteStream -> KJ without shortening") {
   auto wrapped = factory1.capnpToKj(factory2.kjToCapnp(kj::mv(pipe.out)));
 
   {
-    auto promise = wrapped->write("foo", 3);
```
a-robinson, line 61: Is a non-split write tested anywhere else? If not I'd leave this old case in too.

c++/src/capnp/compat/byte-stream.c++
```diff
@@ -730,10 +730,20 @@ public:
         }
       }
       KJ_CASE_ONEOF(capnpStream, capnp::ByteStream::Client*) {
-        // TODO(perf): Split very large writes into many smaller writes.
-        auto req = capnpStream->writeRequest(MessageSize { 8 + size / sizeof(word), 0 });
-        req.setBytes(kj::arrayPtr(reinterpret_cast<const byte*>(buffer), size));
-        return req.send();
+        static uint maxBytesPerWrite = 1 << 16;
```
a-robinson, line 733: Nit, but this should be marked `const` as well.

a-robinson (approved):

c++/src/capnp/compat/byte-stream-test.c++
bretthoerner: Well, the final step of a split write is a non-split write.

xortive (approved):

c++/src/capnp/compat/byte-stream-test.c++
kentonv: I think it would be better to add a new test case rather than edit an existing one.

c++/src/capnp/compat/byte-stream.c++
```diff
@@ -730,10 +730,20 @@ public:
         }
       }
       KJ_CASE_ONEOF(capnpStream, capnp::ByteStream::Client*) {
-        // TODO(perf): Split very large writes into many smaller writes.
-        auto req = capnpStream->writeRequest(MessageSize { 8 + size / sizeof(word), 0 });
-        req.setBytes(kj::arrayPtr(reinterpret_cast<const byte*>(buffer), size));
-        return req.send();
+        static const uint maxBytesPerWrite = 1 << 16;
```
kentonv, line 733: Use uppercase snake for constants: `MAX_BYTES_PER_WRITE`

kentonv:
Probably need to do `write(kj::ArrayPtr<const kj::ArrayPtr<const byte>> pieces)` too...

bretthoerner:
OK, I borrowed from the existing pieces splitting code and added tests for splitting large writes (as part of arrays) up.

c++/src/capnp/compat/byte-stream.c++
```diff
@@ -795,20 +805,61 @@ public:
...
+        };
+
+        static const uint MAX_BYTES_PER_WRITE = 1 << 16;
         size_t size = 0;
         for (auto& piece: pieces) size += piece.size();
-        auto req = capnpStream->writeRequest(MessageSize { 8 + size / sizeof(word), 0 });
+        if (size <= MAX_BYTES_PER_WRITE) {
+          return writePieces(pieces);
+        } else {
+          // ughhhhhhhhhh, we need to split the pieces.
```
kentonv, line 828: Since this code is nearly identical to above, and indeed is doing pretty much the same thing, could we factor out a lambda or something?

kentonv: Or a helper method?

bretthoerner: I put up a fixup, I'm not sure if it's what you had in mind.

But I'm really confused/stuck by why I'm getting a complaint about the deleted copy constructor in my `attach` of any `Array` here... https://github.com/capnproto/capnproto/blob/fd4fea1b37e5ffb82d6fb61be559c47505405969/c%2B%2B/src/capnp/compat/byte-stream.c%2B%2B#L774

```
    In file included from /ekam-provider/canonical/capnp/compat/byte-stream.c++:22:
    In file included from /ekam-provider/canonical/capnp/compat/byte-stream.h:25:
    In file included from /ekam-provider/c++header/capnp/compat/byte-stream.capnp.h:9:
    In file included from /ekam-provider/c++header/capnp/capability.h:28:
    In file included from /ekam-provider/c++header/kj/async.h:24:
    In file included from /ekam-provider/c++header/kj/async-prelude.h:28:
    /ekam-provider/c++header/kj/tuple.h:314:12: error: call to deleted constructor of 'Decay<const kj
      ::Array<kj::ArrayPtr<const unsigned char>>>' (aka 'kj::Array<kj::ArrayPtr<const unsigned char>>
      ')
    /ekam-provider/c++header/kj/tuple.h:257:12: note: in instantiation of function template 
      specialization 'kj::_::MakeTupleFunc::operator()<const kj::Array<kj::ArrayPtr<const unsigned 
      char>>>' requested here
    /ekam-provider/c++header/kj/tuple.h:245:10: note: in instantiation of function template 
      specialization 'kj::_::ExpandAndApplyFunc<kj::_::MakeTupleFunc, const kj::Array<kj::ArrayPtr<
      const unsigned char>>>::operator()<>' requested here
    /ekam-provider/c++header/kj/tuple.h:265:10: note: in instantiation of function template 
      specialization 'kj::_::expandAndApply<kj::_::ExpandAndApplyFunc<kj::_::MakeTupleFunc, const kj:
      :Array<kj::ArrayPtr<const unsigned char>>>>' requested here
    /ekam-provider/c++header/kj/tuple.h:348:13: note: in instantiation of function template 
      specialization 'kj::_::expandAndApply<kj::_::MakeTupleFunc, const kj::Array<kj::ArrayPtr<const 
      unsigned char>>>' requested here
    /ekam-provider/c++header/kj/async-inl.h:1040:25: note: in instantiation of function template 
      specialization 'kj::tuple<const kj::Array<kj::ArrayPtr<const unsigned char>>>' requested here
    /ekam-provider/canonical/capnp/compat/byte-stream.c++:774:32: note: in instantiation of function 
      template specialization 'kj::Promise<void>::attach<const kj::Array<kj::ArrayPtr<const unsigned 
      char>>>' requested here
    /ekam-provider/c++header/kj/array.h:146:20: note: 'Array' has been explicitly marked deleted here
    In file included from /ekam-provider/canonical/capnp/compat/byte-stream.c++:22:
    In file included from /ekam-provider/canonical/capnp/compat/byte-stream.h:25:
    In file included from /ekam-provider/c++header/capnp/compat/byte-stream.capnp.h:9:
```

a-robinson: You probably just need to mark the lambda as mutable, since right now you're trying to `mv` a const `Array`.

bretthoerner: lol, sigh, yup. Error message fatigue. :(

c++/src/capnp/compat/byte-stream.c++
```diff
@@ -986,6 +981,45 @@ private:
...
+    size_t splitPiece = 0;
+    while (pieces[splitPiece].size() <= splitByte) {
+      splitByte -= pieces[splitPiece].size();
+      ++splitPiece;
+    }
+
+    if (splitByte == 0) {
+      // Oh thank god, the split is between two pieces.
+      return SplitResult {
+        .piecesToWrite = kj::heapArray(pieces.slice(0, splitPiece)),
```
kentonv, line 1001: This case didn't used to do allocation.

I was thinking something like:

```cpp
template <typename WritePieces>
kj::Promise<void> splitAndWrite(kj::ArrayPtr<const kj::ArrayPtr<const byte>> pieces, size_t limit, WritePieces&& writeFirstPieces)
```

This would then call `writeFirstPieces(firstPieces)`, and then recurse to `write()` for the rest, and would handle all of the attaching of arrays to promises that is needed.

But if you're getting tired of this PR I guess put it back to the last version that had the duplicated code and we'll clean in up later.

bretthoerner: I think that makes a lot more sense, I pushed a fixup but could use some eyes.

c++/src/capnp/compat/byte-stream-test.c++
```diff
@@ -64,6 +70,38 @@ KJ_TEST("KJ -> ByteStream -> KJ without shortening") {
...
+        str.begin() + (1 << 16)), 1 << 16));
+
+    // One 2^17 piece will be split.
+    pieces.add(kj::ArrayPtr(reinterpret_cast<kj::byte*>(
+        str.begin() + (1 << 16)), str.size() - (1 << 17)));
+
+    auto promise = wrapped->write(pieces);
+    KJ_EXPECT(!promise.poll(waitScope));
+    expectRead(*pipe.in, str).wait(waitScope);
+    promise.wait(waitScope);
```
bretthoerner, line 102: With my latest change this wait results in

```
kj/async.c++:1593: failed: Nothing to wait for; this thread would hang forever.
```

I've read and re-read the code. I put in prints and it the splitting follows the path I expect the correct number of times. Maybe someone can spot what I'm missing? I'm not feeling well today, I'm sure it's something dumb.

c++/src/capnp/compat/byte-stream.c++
```diff
@@ -1014,10 +1005,10 @@ private:
       left.back() = pieces[splitPiece].slice(0, splitByte);
       right.front() = pieces[splitPiece].slice(splitByte, pieces[splitPiece].size());
 
-      return SplitResult {
-        .piecesToWrite = kj::mv(left),
-        .rest = kj::mv(right)
-      };
+      return writeFirstPieces(left).attach(kj::mv(left))
+          .then([this,right=kj::mv(right),limit,writeFirstPieces]() mutable {
+        return splitAndWrite(right, limit, writeFirstPieces).attach(kj::mv(right));
```
kentonv, line 1010: I think this should call `write()`, not `splitAndWrite()`. In the case of a borrowed stream, the limit was a one-time thing...

c++/src/capnp/compat/byte-stream.c++
```diff
@@ -756,59 +766,37 @@ public:
           });
         } else {
```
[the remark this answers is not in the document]
bretthoerner: The different behavior here being that if it's `MAX_BYTES_PER_WRITE`, it should write N times (up to `kjStream.limit`) before returning the stream?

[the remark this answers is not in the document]
kentonv: It wouldn't return the borrowed stream after the first write. But it would recurse to `write()` which would eventually release the borrowed stream once that limit is reached, yes.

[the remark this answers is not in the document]
bretthoerner: Oh boy... let me see here... :upside_down_face:

c++/src/capnp/compat/byte-stream.c++
bretthoerner: Yep, good catch, thanks.

c++/src/capnp/compat/byte-stream.c++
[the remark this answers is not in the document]
bretthoerner: Er, I just pushed something but now on re-reading `returnStream` it looks like it might not be doing what I expected (on my first read I thought it wasn't actually doing the return until `completed` bytes were written.

https://github.com/capnproto/capnproto/blob/29cc8acf67e28504d60f12d169507ece58528ce0/c%2B%2B/src/capnp/compat/byte-stream.c%2B%2B#L80-L89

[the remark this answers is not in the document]
bretthoerner: I should have said I updated this and think it's ready. I'm not in love with the `written` field on the `BorrowedStream` but I don't know where else it could live that'd be cleaner. My only other thought was to make `BorrowedStream` itself an `AsyncOutputStream` that would implement the writes so it could handle things, but I think that'd be moving a lot of stuff around for little gain (and probably has other issues I haven't thought of yet).

Could be missing something simpler, though. :)

c++/src/capnp/compat/byte-stream.c++
```diff
@@ -986,6 +986,43 @@ private:
...
+      splitByte -= pieces[splitPiece].size();
+      ++splitPiece;
+    }
+
+    if (splitByte == 0) {
+      // Oh thank god, the split is between two pieces.
+      auto rest = pieces.slice(splitPiece, pieces.size());
+      return writeFirstPieces(pieces.slice(0, splitPiece))
+          .then([this,rest,limit,writeFirstPieces]() mutable {
+        return splitAndWrite(rest, limit, writeFirstPieces);
```
kentonv, line 1005: Change this to `write(rest)`.

c++/src/capnp/compat/byte-stream.c++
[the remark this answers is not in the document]
kentonv: Heh, I just spent like 15 minutes being horribly confused by my own code here, but I think I finally figured it out.

So what's actually happening is, when `write()` first starts and calls `getShortestPath()`, that is what actually borrows the destination stream. Hence, you always need to return it at the end of the operation, and _before_ recursing. When you recurse into `write()` again, it's going to borrow the stream again, and it can't double-borrow.

I think your approach of adding `written` doesn't actually work then, because `kjStream` is actually a temporary so it's just getting thrown away when you recurse. But you don't need to accumulate. You should just always call `returnStream()` telling it exactly how much was written, before recursing.

But I guess that means `splitAndWrite()` can't actually directly recurse to `write()` because it doesn't know if it needs to call `returnStream()` first... But I think what you can do, actually, is undo the whole `splitAndWrite()` refactoring. All you really have to do is change the original code to use `kj::min(kjStream.limit, MAX_BYTES_PER_WRITE)` instead of `kjStream.limit` -- and then pass the actual amount written to `returnStream()`. If it's less that `kjStream.limit`, then the next `getShortestPath()` will borrow the same stream again and everything works fine.

Arguably it would be better to avoid returning and re-borrowing the same stream repeatedly, but for large writes I think it won't be a big deal.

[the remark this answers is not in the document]
kentonv: Sorry, my comment yesterday is totally wrong. I forgot that you're using `splitAndWrite()` in the `capnp::ByteStream::Client*` case, which didn't do any splitting before.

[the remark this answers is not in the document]
bretthoerner: Done. I definitely pushed my fixup (and can see as much in my shell history) but I'm failing to find it in the GitHub UI after my force push...

```diff
d73ebb75 - (origin/bhoerner/http-over-capnp-split-writes) fixup! Split very large byte-stream writes into smaller writes (4 minutes ago) <Brett Hoerner>
diff --git a/c++/src/capnp/compat/byte-stream.c++ b/c++/src/capnp/compat/byte-stream.c++
index ce77b408..c7b709dc 100644
--- a/c++/src/capnp/compat/byte-stream.c++
+++ b/c++/src/capnp/compat/byte-stream.c++
@@ -43,7 +43,6 @@ public:
     StreamServerBase& lender;
     kj::AsyncOutputStream& stream;
     uint64_t limit;
-    uint64_t written = 0;
   };
 
   typedef kj::OneOf<kj::Promise<void>, capnp::ByteStream::Client*, BorrowedStream> ShortestPath;
@@ -718,20 +717,16 @@ public:
         });
       }
       KJ_CASE_ONEOF(kjStream, StreamServerBase::BorrowedStream) {
-        auto limit = kj::min(kjStream.limit - kjStream.written, MAX_BYTES_PER_WRITE);
+        auto limit = kj::min(kjStream.limit, MAX_BYTES_PER_WRITE);
         if (size <= limit) {
           auto promise = kjStream.stream.write(buffer, size);
           return promise.then([kjStream,size]() mutable {
-            kjStream.written += size;
-            kjStream.lender.returnStream(kjStream.written);
+            kjStream.lender.returnStream(size);
           });
         } else {
           auto promise = kjStream.stream.write(buffer, limit);
           return promise.then([this,kjStream,buffer,size,limit]() mutable {
-            kjStream.written += limit;
-            if (kjStream.written == kjStream.limit) {
-              kjStream.lender.returnStream(kjStream.written);
-            }
+            kjStream.lender.returnStream(limit);
             return write(reinterpret_cast<const byte*>(buffer) + limit,
                          size - limit);
           });
@@ -770,18 +765,14 @@ public:
         if (size <= limit) {
           auto promise = kjStream.stream.write(pieces);
           return promise.then([kjStream,size]() mutable {
-            kjStream.written += size;
-            kjStream.lender.returnStream(kjStream.written);
+            kjStream.lender.returnStream(size);
           });
         } else {
           // ughhhhhhhhhh, we need to split the pieces.
           return splitAndWrite(pieces, kjStream.limit,
               [kjStream,limit](kj::ArrayPtr<const kj::ArrayPtr<const byte>> pieces) mutable {
             return kjStream.stream.write(pieces).then([kjStream,limit]() mutable {
-              kjStream.written += limit;
-              if (kjStream.written == kjStream.limit) {
-                kjStream.lender.returnStream(kjStream.written);
-              }
+              kjStream.lender.returnStream(limit);
             });
           });
         }
@@ -1001,8 +992,8 @@ private:
       // Oh thank god, the split is between two pieces.
       auto rest = pieces.slice(splitPiece, pieces.size());
       return writeFirstPieces(pieces.slice(0, splitPiece))
-          .then([this,rest,limit,writeFirstPieces]() mutable {
-        return splitAndWrite(rest, limit, writeFirstPieces);
+          .then([this,rest]() mutable {
+        return write(rest);
       });
     } else {
       // FUUUUUUUU---- we need to split one of the pieces in two.
```

kentonv (approved):

PR #1110, opened 2020-11-19
Add some utilities for debugging exceptions

kentonv:
See individual commits.

### DIFF
--- a/c++/src/kj/exception.c++
+++ b/c++/src/kj/exception.c++
@@ -1146,6 +1146,51 @@ kj::String getCaughtExceptionType() {
 }
 #endif
 
+namespace {
+
+size_t sharedSuffixLength(kj::ArrayPtr<void* const> a, kj::ArrayPtr<void* const> b) {
+  size_t result = 0;
+  while (a.size() > 0 && b.size() > 0 && a.back() == b.back())  {
+    ++result;
+    a = a.slice(0, a.size() - 1);
+    b = b.slice(0, b.size() - 1);
+  }
+  return result;
+}
+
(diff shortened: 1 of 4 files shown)


c++/src/kj/exception.c++
```diff
@@ -830,24 +830,69 @@ void Exception::addTrace(void* ptr) {
...
+    // Look for ourselves in the list.
+    for (auto* ptr = &currentException; *ptr != nullptr; ptr = &(*ptr)->nextCurrentException) {
+      if (*ptr == this) {
+        *ptr = nextCurrentException;
+        return;
+      }
+    }
+
+    // Possibly the ExceptionImpl was destroyed on a different thread than created it? That's
+    // pretty bad, we'd better abort.
```
harrishancock, line 860: This seems like a surprising constraint to me. I could imagine needing be able to move exceptions across threads... but I suppose the way to do that is to return one from an executeAsync() callback, which will make sure the objects are constructed and destroyed on their respective threads?

harrishancock (approved):

c++/src/kj/exception.c++
kentonv: Note that `Exception`s can still be moved between threads just fine. Only `ExceptionImpl` cannot. Instances of `ExceptionImpl` are only constructed for the purpose of throwing an exception, and I'm reasonably certain the C++ exception-throwing machinery won't destroy a thrown object on a different thread than the one that threw it.

kentonv:
Looks like GitHub broke their Windows action runner again.

Issue #1107, opened 2020-11-19
Initializing an AnyList


JonasVautherin:
I am trying to add support for `List<AnyList>`. In my tests, I want to create `AnyList` objects that I can later `set(index, object)` into my `List<AnyList>`. But I can't seem to initialize an `AnyList`.

I have tried to use `messageBuilder.initRoot<AnyList>()`, but that won't work because an `AnyList` initialization needs more information than the usual `initRoot` (namely the element size and list size). I thought about extending `AnyList` in order to allow creating one with something like `AnyList::Builder myBuilder(elementSize, size)`, but it feels like low-level for the API. Alternatively, I thought about passing a `List<T>` to my `List<AnyList>`, like:

```c++
  List<uint16_t>::Builder myListBuilder;
  myListBuilder.set(0, 8);

  auto root = builder.initRoot<test::TestListOfAny>();
  auto anyListList = root.initListList(2);
  anyListList.set(0, myListBuilder);
```

But `myListBuilder` has the wrong type here, and I am not sure how I would cast it.

Any suggestion?

kentonv:
Note that the ability to declare `List(AnyList)` at all (without a compiler error) was only added two weeks ago in #1097, and it was noticed at the time that the C++ generated code didn't work (but the author of that PR was mainly interested in `List(Capability)`, which did work).

PR #1112, opened 2020-11-20
Add support for List(AnyList)

JonasVautherin:
This adds support for `List(AnyList)`. It is lacking a test exercising it, because I don't know how to actually create an `AnyList` in a test (see https://github.com/capnproto/capnproto/issues/1107).

### DIFF
--- a/c++/src/capnp/any.h
+++ b/c++/src/capnp/any.h
@@ -48,6 +48,14 @@ struct AnyList {
   class Builder;
 };
 
+template<>
+struct List<AnyList, Kind::LIST> {
+    List() = delete;
+
+    class Reader;
+    class Builder;
+};
+
 struct AnyStruct {
   AnyStruct() = delete;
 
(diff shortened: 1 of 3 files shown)


kentonv:
Thanks for working on this! Unfortunately this code doesn't look like it works... Your implementation of `operator[]` looks like it always returns empty lists, and the `Builder` is missing important methods like `init(i, size)`, `adopt(i, orphan)`, and `disown(i)`. These are kind of complicated to implement correctly -- e.g. `init(i, size)` probably needs to be templated so that you can specify what type of list you want to initialize at the given index, and there probably also need to be versions that accept dynamically-specified element sizes too. Since this is getting pretty far into the weeds it's probably best for me to work on it myself... but when I poked at this a little bit today it exceeded my time box. :/

Out of curiosity, what is your use case?

JonasVautherin:
> Since this is getting pretty far into the weeds it's probably best for me to work on it myself...

Right, makes total sense in that case! Note: I don't need it :sweat_smile:.

> Out of curiosity, what is your use case?

Back in #1097, you mentioned a few follow-ups:

* Fix List(AnyList) in C++ generated code.
* Verify that dynamic API works with these types.
* Verify that constants of these types work correctly.

I thought that if `List(AnyList)` was the same kind of scope as `List(Capability)`, maybe I could contribute it. As a way to help while learning the codebase, I guess. But it turns out to be more difficult than `List(Capability)` :sweat_smile:. For instance, creating a `Capability` was very simple in a test, but I am not completely sure how to create an `AnyList` and why one would do that.

JonasVautherin:
I think this can be closed, I probably won't manage to do it myself and I don't need it (I was just hoping to help).

PR #1116, opened 2020-11-25
Use archive of "All Concurrency Models Suck"

ObsidianMinor:
RIP Google+


kentonv:
Oof... I'm actually kind of annoyed that my Google+ history has been archived without my consent. When I deleted my G+ account (before G+ shut down) I had intended for my social media history to disappear.

I really should find another home for this particular text, though... hmm....

PR #1117, opened 2020-11-26
Redesign async trace code

kentonv:
The old Event::trace() in particular was nonsense -- it didn't actually trace back to the next event and so on, instead it would trace _forwards_ through `PromiseNode`s.

Additionally, the output of the old traces was not particularly useful, as it would tell you the types of the promises but not line numbers.

The new tracing code produces a list of instruction addresses, just like regular stack traces. It can be passed into all the regular stack tracing tools.

Additionally, a new global function `kj::getAsyncTrace()` gets a trace of the promises leading up to the currently-executing code. Previously, such traces were generated when you threw an exception and it actually passed through the promise chain, but there was no way to get a meaningful async trace e.g. in a crash handler. Now there is.

### DIFF
--- a/c++/src/kj/async.c++
+++ b/c++/src/kj/async.c++
@@ -86,14 +86,29 @@ EventLoop& currentEventLoop() {
   return *loop;
 }
 
-class BoolEvent: public _::Event {
+class RootEvent: public _::Event {
 public:
+  RootEvent(_::PromiseNode* node, void* traceAddr): node(node), traceAddr(traceAddr) {}
+
   bool fired = false;
 
   Maybe<Own<_::Event>> fire() override {
     fired = true;
     return nullptr;
   }
+
+  void traceEvent(_::TraceBuilder& builder) override {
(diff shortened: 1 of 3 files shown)


c++/src/kj/async.c++
```diff
@@ -216,6 +240,13 @@ public:
   Maybe<Own<Task>> next;
   Maybe<Own<Task>>* prev = nullptr;
 
+  kj::String trace() {
```
harrishancock, line 243: I couldn't find where this, and PromiseNode::trace(), are called. Are they used?

harrishancock (approved):

c++/src/kj/async.c++
kentonv: `TaskSet::Task::trace()` is called by `TaskSet::trace()`. `PromiseNode::trace()` is called by `Promise::trace()`. `TaskSet::trace()` and `Promise::trace()` are public interfaces.

PR #1118, opened 2020-11-26
Improve exceptions when dropping the fulfiller of a PromiseFulfillerPair

kentonv:
See individual commits...

### DIFF
--- a/c++/src/kj/async.c++
+++ b/c++/src/kj/async.c++
@@ -2569,6 +2569,19 @@ void AdapterPromiseNodeBase::tracePromise(TraceBuilder& builder, bool stopAtNext
   builder.add(getMethodStartAddress(implicitCast<PromiseNode&>(*this), &PromiseNode::get));
 }
 
+void WeakFulfillerBase::disposeImpl(void* pointer) const {
+  if (inner == nullptr) {
+    // Already detached.
+    delete this;
+  } else {
+    if (inner->isWaiting()) {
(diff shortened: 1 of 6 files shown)


harrishancock (approved):

kentonv:
(force-push to rebase on master and resolve conflicts, which were all of the form "both branches added new code in the same location")

c++/src/kj/async.c++
```diff
@@ -2575,6 +2575,17 @@ void WeakFulfillerBase::disposeImpl(void* pointer) const {
...
+#if !KJ_NO_EXCEPTIONS
+      InFlightExceptionIterator iter;
+      KJ_IF_MAYBE(e, iter.next()) {
+        auto copy = kj::cp(*e);
+        copy.truncateCommonTrace();
+        inner->reject(kj::mv(copy));
+      }
+#endif
+      // Darn, use a generic exception.
       inner->reject(kj::Exception(kj::Exception::Type::FAILED, __FILE__, __LINE__,
```
a-robinson: Not a big deal since this is a pretty small allocation and it looks like duplicate `reject` calls are noops, but just FYI in case you do care this is potentially calling `reject` a second time.

a-robinson (approved):

c++/src/kj/async.c++
kentonv: Whoops, yes, that's a bug -- and the next commit actually makes this considerably more expensive than an allocation by adding a backtrace. Thanks.

PR #1120, opened 2020-11-28
Implement magic assert stringification.

kentonv:
When `KJ_ASSERT(foo == bar)` fails, `foo` and `bar`'s actual values will be stringified in the error message. How does it work? We use template magic and operator precedence. The assertion actually evaluates something like this:

    if (auto _kjCondition = kj::_::MAGIC_ASSERT << foo == bar)

`<<` has operator precedence slightly above `==`, so `kj::_::MAGIC_ASSERT << foo` gets evaluated first. This wraps `foo` in a little wrapper that captures the comparison operators and keeps enough information around to be able to stringify the left and right sides of the comparison independently. As always, the stringification only actually occurs if the assert fails.

You might ask why we use operator `<<` and not e.g. operator `<=`, since operators of the same precedence are evaluated left-to-right. The answer is that some compilers trigger all sorts of warnings when you seem to be using a comparison as the input to another comparison. The particular warning GCC produces is its general "-Wparentheses" warning which is broadly useful, so we don't want to disable it. `<<` also produces some warnings, but only on Clang and the specific warning is one we're comfortable disabling (see below). This does mean that we have to explicitly overload `operator<<` ourselves to make sure using it in an assert still works.

You might also ask, if we're using operator `<<` anyway, why not start it from the right, in which case it would bind after computing any `<<` operators that were actually in the user's code? I tried this, but it resulted in a somewhat broader warning from clang that I felt worse about disabling (a warning about `<<` precedence not applying specifically to overloads) and also created ambiguous overload errors in the KJ units code.

### DIFF
--- a/c++/src/capnp/compat/http-over-capnp-test.c++
+++ b/c++/src/capnp/compat/http-over-capnp-test.c++
@@ -391,7 +391,7 @@ void runEndToEndTests(kj::Timer& timer, kj::HttpHeaderTable& headerTable,
   clientPipe.ends[0] = nullptr;
   auto lastRead = serverPipe.ends[1]->readAllText();
   KJ_ASSERT(lastRead.poll(waitScope), "last read hung");
-  KJ_EXPECT(lastRead.wait(waitScope) == 0);
+  KJ_EXPECT(lastRead.wait(waitScope) == nullptr);
 }
 
 KJ_TEST("HTTP-over-Cap'n-Proto E2E, no path shortening") {
--- a/c++/src/kj/debug-test.c++
+++ b/c++/src/kj/debug-test.c++
@@ -430,6 +430,71 @@ TEST(Debug, Context) {
   }
 }
 
+KJ_TEST("magic assert stringification") {
+  {
+    auto exception = KJ_ASSERT_NONNULL(kj::runCatchingExceptions([&]() {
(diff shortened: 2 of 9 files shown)


kentonv:
Note: I also tried coalescing from the right, i.e.:

    if (auto _kjCondition = foo == bar << kj::_::MAGIC_ASSERT)

This did not go well:
1. It triggered much broader warnings about operator precedence, not specific to overloaded operators, that I felt less comfortable disabling.
2. It caused at least one KJ file to fail to compile because of ambiguous operator overloads.

Whereas the coalesce-from-left version has created fewer issues.

kentonv:
Oh DUH, I just realized that because comparison operators of same precedence are evaluated left-to-right, I should be overloading a comparison operator rather than `<<` to kick off the coalescing.

kentonv:
Note: This wasn't entirely my idea. Well over 10 years ago, at Google, @zhanyong-wan proposed doing something like this in gtest. But, he concluded that it was too crazy. My recollection of reading between the lines was that Zhanyong actually thought it was a good idea but concluded that _other people_ would think it was too crazy.

It's not crazy. I've wished we had this many, many times over the last 7 years.

harrishancock (approved):
Sorry for my slow review; this looks great!

PR #1126, opened 2020-12-03
Do not use fallocate on Android API < 21

elelel:
Android got both fallocate and posix_allocate only in API 21. For this reason there is no way to get os-asisted file zeroing.

### DIFF
--- a/c++/src/kj/filesystem-disk-unix.c++
+++ b/c++/src/kj/filesystem-disk-unix.c++
@@ -383,11 +383,11 @@ public:
   }
 
   void zero(uint64_t offset, uint64_t size) const {
-// If FALLOC_FL_PUNCH_HOLE is defined, use it to efficiently zero the area.
-//
-// A fallocate() wrapper was only added to Android's Bionic C library as of API level 21,
-// but FALLOC_FL_PUNCH_HOLE is apparently defined in the headers before that, so we'll
-// have to explicitly test for that case.
+    // If FALLOC_FL_PUNCH_HOLE is defined, use it to efficiently zero the area.
+    //
+    // A fallocate() wrapper was only added to Android's Bionic C library as of API level 21,
+    // but FALLOC_FL_PUNCH_HOLE is apparently defined in the headers before that, so we'll
+    // have to explicitly test for that case.
 #if defined(FALLOC_FL_PUNCH_HOLE) && !(__ANDROID__ && __BIONIC__ && __ANDROID_API__ < 21)
     KJ_SYSCALL_HANDLE_ERRORS(
         fallocate(fd, FALLOC_FL_PUNCH_HOLE | FALLOC_FL_KEEP_SIZE, offset, size)) {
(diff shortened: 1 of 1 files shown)


c++/src/kj/filesystem-disk-unix.c++
```diff
@@ -383,7 +383,7 @@ public:
   }
 
   void zero(uint64_t offset, uint64_t size) const {
-#ifdef FALLOC_FL_PUNCH_HOLE
+#if (FALLOC_FL_PUNCH_HOLE && (!(defined(__ANDROID__) && __ANDROID_API__ < 21)))
```
kentonv, line 386:
```suggestion
// If FALLOC_FL_PUNCH_HOLE is defined, use it to efficiently zero the area.
//
// A fallocate() wrapper was only added to Android's Bionic C library as of API level 21,
// but FALLOC_FL_PUNCH_HOLE is defined in the Linux headers before that, so we'll
// have to explicitly test for that case.
#if defined(FALLOC_FL_PUNCH_HOLE) && !(__ANDROID__ && __BIONIC__ && __ANDROID_API__ < 21)
```

kentonv: I suppose in theory we could probably use `syscall(__NR_fallocate)` since the kernel supports it, but perhaps that's trying too hard, heh.

elelel: In case of Android would not rely on Linux syscalls not in wide use there. Since the kernel there is only a reference thing for end-user hardware producers, there always be some variant of kernel with a particular thing messed up by their patches.

c++/src/kj/filesystem-disk-unix.c++
```diff
@@ -383,7 +383,12 @@ public:
   }
 
   void zero(uint64_t offset, uint64_t size) const {
-#ifdef FALLOC_FL_PUNCH_HOLE
+// If FALLOC_FL_PUNCH_HOLE is defined, use it to efficiently zero the area.
+//
+// A fallocate() wrapper was only added to Android's Bionic C library as of API level 21,
+// but FALLOC_FL_PUNCH_HOLE is apparently defined in the headers before that, so we'll
+// have to explicitly test for that case.
+#if defined(FALLOC_FL_PUNCH_HOLE) && !(__ANDROID__ && __BIONIC__ && __ANDROID_API__ < 21)
```
kentonv, line 391: Sorry, only realized after seeing this applied that I mis-indented the comments.

```suggestion
    // If FALLOC_FL_PUNCH_HOLE is defined, use it to efficiently zero the area.
    //
    // A fallocate() wrapper was only added to Android's Bionic C library as of API level 21,
    // but FALLOC_FL_PUNCH_HOLE is apparently defined in the headers before that, so we'll
    // have to explicitly test for that case.
#if defined(FALLOC_FL_PUNCH_HOLE) && !(__ANDROID__ && __BIONIC__ && __ANDROID_API__ < 21)
```

kentonv:
Thanks!

PR #1127, opened 2020-12-03
Provide compat layer for atfile libc calls

elelel:
Atfile libc calls became available on Android since API21. This patch provides a best effort compatbility layer for older Android versions by declaring proxy functions with same interface in kj namespace. Additionally this compatibility layer is activated by glibc defines, if provided with.

### DIFF
--- a/c++/src/kj/CMakeLists.txt
+++ b/c++/src/kj/CMakeLists.txt
@@ -27,6 +27,7 @@ set(kj_sources_heavy
   filesystem-disk-unix.c++
   filesystem-disk-win32.c++
   parse/char.c++
+  compat/atfile.c++
 )
 if(NOT CAPNP_LITE)
   set(kj_sources ${kj_sources_lite} ${kj_sources_heavy})
--- a/c++/src/kj/compat/atfile.c++
+++ b/c++/src/kj/compat/atfile.c++
@@ -1,0 +1,98 @@
+#include "atfile.h"
+
+#ifdef __linux
+
+#include <linux/limits.h> // PATH_MAX
+#include <errno.h>        // EBADF and other error constants
+#include <cstdio>         // snprintf (emulate libc only in terms of libc)
+#include <unistd.h>       // non-*at versions of emulated functions, e.g. readlink
+
+// Attempt to get path from file descriptor
+inline static int path_from_fd(int fd, char *buf, size_t buf_sz) {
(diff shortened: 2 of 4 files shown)


elelel:
Sorry for second PR including the previous one. Just happened to make it in parallel.

kentonv:
Hmm, I'm not sure KJ is a good place for this code.

Android API 21 (5.0) is more than six years; older versions have [~3% market share and falling](https://www.appbrain.com/stats/top-android-sdk-versions). The Cap'n Proto maintainers (mostly me) just don't have the time to spend maintaining code targeting such niches.

Here are some alternatives I think would work well:

1. I believe the KJ filesystem code is not actually used anywhere in KJ or capnp except for the capnp compiler. You probably aren't running the compiler on the device itself, so you probably don't actually need this code at runtime. So, perhaps it would be better to add a new `configure` flag that omits the filesystem code from the library. This would be similar to the existing `--disable-reflection` flag; the new flag could be `--disable-filesystem`. Similar to `--disable-reflection`, this flag would also avoid compiling `capnp`, and would need to be paired with `--with-external-capnp` (which you're probably already using anyway, for cross-compiling). (Or if you aren't using the autotools build, then make the equivalent change to whatever build system you are using...)

2. Another solution would be for you to maintain your `openat` compatibility layer as a separate library. You could force libkj to link against this library by passing appropriate `CXXFLAGS` and `LIBS` to `configure`, so no changes to the capnproto repository would be needed. This could be a generally useful library independent of KJ.

3. I would be more willing to accept a more narrow compatibility layer that just defines these wrappers in terms of `syscall()`. I'm pretty sure that every version of Android ever released actually has these calls in the kernel, and it's unlikely that vendors would have messed with such core VFS calls that have no device-specific implications.

elelel:
I think option 1 would the optiomal variant. I'm not using --with-external-capnp, but the cmake macros from c++/cmake directory. I thought it behaved equivalently, and assumed that no parts used in tooling were included. So this pull request may be closed, if it's no longer needed.

kentonv:
Right, the filesystem code is normally part of libkj, but it isn't actually used by anything else in libkj or libcapnp. I had considered putting it in a separate library like `libkj-file` or something, but it felt like too many librariaes.

Anyway, I'd be happy to accept a change to the cmake files to add a flag which causes it to skip building the filesystem code (and also skip building `capnp`).

PR #1129, opened 2020-12-09
Minor speedup for miniposix::iovMax

vlovich:
On platforms that need iovMax to be queried via sysconf, iovMax
does an unnecessary syscall on every call to `iovMax`. This value isn't
going to change during runtime though. Move this to a constant that's
initialized only once and remove the `count` argument into iovMax since
it's actually superfluous to the API. The more correct action is to just return
`std::numeric_limits<size_t>::max()` and let the caller figure out what
the value is.

Open question is whether this should be just an implementation detail of
`iovMax` (i.e. hide the lazy initialization of the runtime value there) or to
make it the responsibility of the caller. My personal preference is to hide
it within `iovMax` but figured that may be more controversial.

### DIFF
--- a/c++/src/kj/miniposix.c++
+++ b/c++/src/kj/miniposix.c++
@@ -34,7 +34,7 @@ size_t iovMax() {
     if (iovmax == -1) {
       if (errno == 0) {
         // The -1 return value was the actual value, not an error. This means there's no limit.
-        return kj::MaxValue;
+        return kj::maxValue;
       } else {
         return _XOPEN_IOV_MAX;
       }
--- a/c++/src/kj/miniposix.c++
+++ b/c++/src/kj/miniposix.c++
@@ -32,8 +32,12 @@ size_t iovMax() {
     errno = 0;
     iovmax = sysconf(_SC_IOV_MAX);
     if (iovmax == -1) {
-      // Assume iovmax == -1 && errno == 0 means "unbounded".
-      return errno ? _XOPEN_IOV_MAX : kj::maxValue;
+      if (errno == 0) {
+        // The -1 return value was the actual value, not an error. This means there's no limit.
+        return kj::MaxValue;
+      } else {
+        return _XOPEN_IOV_MAX;
+      }
     }
 
     return (size_t) iovmax;
--- a/c++/src/kj/miniposix.c++
+++ b/c++/src/kj/miniposix.c++
@@ -1,0 +1,45 @@
+// Copyright (c) 2013-2014 Sandstorm Development Group, Inc. and contributors
+// Licensed under the MIT License:
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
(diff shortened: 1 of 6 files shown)


kentonv:
Huh I'm kind of surprised I wrote the code that way. Maybe it was one of those days when I just couldn't bring myself to care about non-Linux targets, heh.

I guess in the case that `iovMax()` isn't `constexpr`, each of these static consts is going to end up generating some code to ensure thread-safe one-time initialization. Given that, it may in fact make sense to do the caching inside `iovMax()` (and move `iovMax()` out-of-line, since you wouldn't want that sort of stuff regenerated in every translation unit, and it'll be too much code for the compiler to want to inline it anyway).

vlovich:
Should be fixed now.

c++/src/kj/miniposix.c++
```diff
@@ -0,0 +1,45 @@
...
+#if !defined(_WIN32) && !defined(IOV_MAX) && !defined(UIO_MAX_IOV)
+size_t iovMax() {
+  // Thread-safe & lazily initialized only on first use.
+  static const size_t KJ_IOV_MAX = [] {
+    long iovmax;
+
+    errno = 0;
+    iovmax = sysconf(_SC_IOV_MAX);
+    if (iovmax == -1 && errno) {
+      // assume iovmax == -1 && errno == 0 means "unbounded"
```
kentonv, line 35: I just noticed that this logic has changed. The comment is now confusing because it describes a case that is _not_ handled by this block.

It looks like the case that the comment describes is now handled by returning `(size_t) iovmax`, the same code that executes in the case that `sysconf()` returns a positive value.

This looks like it works but only because `-1` is being converted to an unsigned value, and thus effectively becomes `ULONG_MAX`, which effectively expresses "unbounded". But it's pretty confusing to rely on a signed->unsigned conversion just coincidentally producing the right result for two logically distinct cases.

Could you maybe handle this case explicitly and return `kj::maxValue` to make things clearer? (Also make sure to declare the lambda's return type explicitly.)

vlovich: How does it look now?

c++/src/kj/miniposix.c++
```diff
@@ -0,0 +1,45 @@
...
+size_t iovMax() {
+  // Thread-safe & lazily initialized only on first use.
+  static const size_t KJ_IOV_MAX = [] () -> size_t {
+    long iovmax;
+
+    errno = 0;
+    iovmax = sysconf(_SC_IOV_MAX);
+    if (iovmax == -1) {
+      // Assume iovmax == -1 && errno == 0 means "unbounded".
+      return errno ? _XOPEN_IOV_MAX : kj::maxValue;
```
kentonv, line 36: This is the right idea, but I think you'll want to cast both branches to `size_t` otherwise I suspect C++ will choose their types to be `int`, and thus `kj::maxValue` will end up being `MAX_INT` instead of `MAX_SIZE`.

Ternary operators are really tricky when the branches aren't the same type, which is why I tend to avoid them.

kentonv, line 36:
```suggestion
    if (iovmax == -1) {
      if (errno == 0) {
        // The -1 return value was the actual value, not an error. This means there's no limit.
        return kj::MaxValue;
      } else {
        return _XOPEN_IOV_MAX;
      }
```

c++/src/kj/miniposix.c++
```diff
@@ -0,0 +1,49 @@
...
+  // Thread-safe & lazily initialized only on first use.
+  static const size_t KJ_IOV_MAX = [] () -> size_t {
+    long iovmax;
+
+    errno = 0;
+    iovmax = sysconf(_SC_IOV_MAX);
+    if (iovmax == -1) {
+      if (errno == 0) {
+        // The -1 return value was the actual value, not an error. This means there's no limit.
+        return kj::MaxValue;
```
kentonv, line 37:
```suggestion
        return kj::maxValue;
```

PR #1131, opened 2020-12-10
Introduce CallConetxt::setPipeline().

kentonv:
This makes it possible, when an RPC forwards to another RPC, to start forward pipelined calls on the first to the second, before the first RPC actually completes.

### DIFF
--- a/c++/src/capnp/capability.h
+++ b/c++/src/capnp/capability.h
@@ -310,6 +310,44 @@ public:
   // should not be included in the size.  So, if you are simply going to copy some existing message
   // directly into the results, just call `.totalSize()` and pass that in.
 
+  void setPipeline(typename Results::Pipeline&& pipeline);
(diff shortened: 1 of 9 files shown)


c++/src/capnp/capability.h
```diff
@@ -330,19 +331,32 @@ class CallContext: public kj::DisallowConstCopy {
   //
   // To solve this problem, Bob takes the pipeline object from the `bar()` call, transforms it into
   // an appropriate pipeline for a `foo()` call, and passes that to `setPipeline()`. This allows
-  // Alice's pipelined `baz()` call to flow through immediately. If `foo()` and `bar()` actually
-  // return the same type (requiring no transformation), this code looks like:
+  // Alice's pipelined `baz()` call to flow through immediately. The code look like:
```
jclee, line 334: s/look/looks

bretthoerner (approved):
This is great, thanks!

PR #1134, opened 2020-12-17
In rpc-twoparty, propagate write failures to read failures.

kentonv:
Historically we assumed that any write failure on the underlying socket would probably manifest as a read failure at the same time, but this is not necessarily true in practice, especially with exotic socket types like kTLS.

### DIFF
--- a/c++/src/capnp/rpc-twoparty-test.c++
+++ b/c++/src/capnp/rpc-twoparty-test.c++
@@ -763,7 +763,7 @@ KJ_TEST("promise cap resolves between starting request and sending it") {
   auto pipe = kj::newTwoWayPipe();
 
   // Client exports TestCallOrderImpl as its bootstrap.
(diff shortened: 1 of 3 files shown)


c++/src/capnp/rpc-twoparty-test.c++
```diff
@@ -789,6 +791,52 @@ KJ_TEST("promise cap resolves between starting request and sending it") {
   KJ_EXPECT(n2 == 1, n2);
 }
 
+KJ_TEST("read error propagates to write error") {
```
a-robinson, line 794: Is this test name backwards? The error on the write is being propagated to the read.

Obviously not very important though, it's just a test name.

a-robinson (approved):

c++/src/capnp/rpc-twoparty-test.c++
kentonv: You're right. And getting things like this right is important, to avoid confusing people later.

Issue #1138, opened 2020-12-22
msys64 gcc compilation error


DamienLeBars:
06e471e introduced a compilation error with windows cygwin/msys gcc
```
capnproto/c++/src/kj/miniposix.c++:39:16: error: ‘_XOPEN_IOV_MAX’ was not declared in this scope
   39 |         return _XOPEN_IOV_MAX;
      |                ^~~~~~~~~~~~~~
```
(from [surelog compilation error in capnproto](https://github.com/alainmarcel/Surelog/runs/1583733628))

DamienLeBars:
I found strange that the compilation configuration is not consistent for miniposix.c++ between:
capnproto/c++/Makefile.am (miniposix.c++ is missing)
and
capnproto/c++/src/kj/CMakeLists.txt

kentonv:
`miniposix.c++` was added in #1129. Looks like @vlovich forgot to update `Makefile.am`, but this slipped past the tests because `miniposix.c++` isn't actually needed by any of the platforms we test on (it's a fallback for POSIX variants that don't define an appropriate constant).

But I am confused by your error report, because we actually do test on Cygwin and MinGW. In our test environment, either `IOV_MAX` or `UIO_MAX_IOV` (not sure which) is defined on Cygwin, and hence `miniposix.c++` is a no-op. On MinGW, `_WIN32` is defined, which also makes `miniposix.c++` a no-op. So how is it that this code is not being skipped in your case?

Are you actually targeting Cygwin, or MinGW? You say "cygwin/msys gcc", which is confusing -- my understanding is that MSYS is an alternative to Cygwin. MSYS is essentially a minimalist Unix shell environment intended to be used with MinGW GCC. So are you saying you're using MinGW GCC? But it's really weird that `_WIN32` isn't defined, then. If you're actually targeting Cygwin, then it makes sense that `_WIN32` is not defined, but one of the other constants should have been -- or is so in our test environment, at least.

I think possibly we should just delete the non-inline definition of `iovMax()` since we're not aware of any target systems that need it. I think I copy/pasted the code from somewhere long ago but I'm not sure if it has ever been exercised as part of this codebase. But we need to figure out why it's being compiled for you...

DamienLeBars:
we target msys2 gcc (that is the same than cygwin), we use cmake and so capnproto/c++/src/kj/CMakeLists.txt is used (and so miniposix.c++ is compiled => with error)
[...]

PR #1140, opened 2020-12-23
Delete sysconf-based iovMax() fallback.

kentonv:
Apparently, no platform that we target today needs this, and as a result it's totally untested, which probably means it doesn't work anyway. I think I copy/pasted this fallback code from a stack overflow or something a very long time ago.

If, in the future, we encounter a platform that doesn't have one of the constants defined, then we should probably either figure out if that platform provides some other constant, or just determine a number that works and hard-code it.

See also #1138, which should go away with this PR (although not really "fixed", there's still apparently a missing include that we haven't figured out?).

cc @vlovich


vlovich:
LGTM

PR #1141, opened 2020-12-29
KJ Documentation: Maybe/OneOf/Own

haneefmubarak:
In progress. Tasklist: see #1144

Please feel free to add as necessary @kentonv @harrishancock

### DIFF
--- a/kj-doc/wip/kj.md
+++ b/kj-doc/wip/kj.md
@@ -1,0 +1,55 @@
+---
+layout: page
+title: KJ: A Performance and Safety Library for C++
+---
+
+# KJ: A Performance and Safety Library for C++
+
+KJ provides a variety of types, templates, and macros to ease the writing of correct,
+high performance C++ code. Similarities are highlighted when and where KJ features and
+STL features resemble each other, so if you know of an STL feature and are looking for
+the equivalent KJ feature, searching for the STL feature by name should work here.
+
+## Primitives
+
+### Maybe (optional)
+
+`Maybe<T>` is a template for an optionally contained value, similar to (but predating)
+[`std::optional<T>`](https://en.cppreference.com/w/cpp/utility/optional). It has the advantage of
+ensuring that the empty case is handled, thus preventing null pointer dereferences and/or access to
+invalid objects at runtime.
+
+You can construct a `Maybe<T>` using either a value of type `T` or using `nullptr`:
+
+```c++
+#include <kj/common.h>
+
+auto number = kj::Maybe<double>(1.0);
+auto empty = kj::Maybe<double>(nullptr);
+```
+
+You can also use it easily as part of a class, in which case it is default initialized to be empty
+until you initialize it otherwise:
+
+```c++
+struct Attic {
+  ...
+
+  Attic(Ghost ghost) : ghost(ghost) {};
+
+  ...
+
+  kj::Maybe<Ghost> ghost;
+};
+```
+
+Afterwards, you can simultaneously check for and access the contents of a `Maybe<T>` using the
+`KJ_IF_MAYBE()` macro as follows:
+
+```c++
+KJ_IF_MAYBE(pointerToValue, maybeValue) {
+  doSomething(*pointerToValue);
+} else {
+  handleEmptyCase();
+}
+```
--- a/kj-doc/wip/kj.md
+++ b/kj-doc/wip/kj.md
@@ -99,3 +99,69 @@ struct Empty {};
 
 kj::OneOf<Empty, double, kj::String> foo;
 ```
+
+### Own (unique_ptr)
+
+`Own<T>` is a template for a smart pointer that ensures that only a single owner holds a given
+pointer at a given point in time, similar to
+[`std::unique_ptr`](https://en.cppreference.com/w/cpp/memory/unique_ptr). It has the advantage
+of not allowing manual `release()` of the owned memory and only allowing automatic memory
+management, thus preventing accidental double-frees and facilitating the trivial implementation
+of custom memory allocators if and when needed.
+
+For the general case, you can construct an `Own<T>` using `kj::heap<T>(Args ...)`, which allocates
+memory on the heap and constructs an instance of `T` by calling the constructor matching the
+provided arguments:
+
+```c++
+#include <kj/memory.h>
+
+class Thing {
+  ...
+
+  Thing(unsigned quantity, double difficulty);
+
+  ...
+};
+
+auto ownedThing = kj::heap<Thing>(1, 7.5);
+```
+
+You can also use it as part of a class, in which case it's default initialized to `nullptr` until
+you initialize it otherwise:
+
+```c++
+class Continent {
+  ...
+
+  Continent(kj::Own<Thing> creature) : creature(kj::mv(creature)) {};
+
+  ...
+
+  kj::Own<Thing> creature;
+
+  ...
+};
+
+Continent antarctica(kj::heap<Thing>(3, 999));
+```
+
+Since an `Own<T>` is fundamentally a wrapper around a pointer, you can use it in the same manner as
+a raw pointer when attempting to access its internals:
+
+```c++
+if (ownedThing) {
+  ownedThing->morph();
+}
+```
+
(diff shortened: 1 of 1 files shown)


Dr-Emann:
I think it might be worth mentioning `KJ_{ASSERT/REQUIRE}_NONNULL` for getting the value out of a `Maybe<T>` if you know it's non-empty

doc/kj.md
```diff
@@ -36,3 +36,49 @@ KJ_IF_MAYBE(pointer_to_value, maybe_value) {
...
+    doSomethingWithString(s);
+  }
+  KJ_CASE_ONEOF_DEFAULT {
+    doSomethingElse();
+  }
+}
+```
+
+`OneOf<T>` is non-nullable by default; using a nullish value (`nullptr`, etc) to make it nullable
+is inadvisable because it may intefere with implementation details. Instead, if you need an empty
```
jclee, line 77: s/intefere/interfere

doc/kj.md
```diff
@@ -0,0 +1,84 @@
+---
+layout:	page
```
bretthoerner, line 2: Odd tab here.

doc/kj.md
```diff
@@ -0,0 +1,84 @@
...
+auto number = kj::Maybe<double>(1.0);
+auto empty = kj::Maybe<double>(nullptr);
+```
+
+Afterwards, you can simultaneously check for and access the contents of a `Maybe<T>` using the
+`KJ_IF_MAYBE()` macro as follows:
+
+```c++
+KJ_IF_MAYBE(pointer_to_value, maybe_value) {
+  doSomething(*pointer_to_value);
```
bretthoerner, line 34: nit: KJ style would be pointerToValue

doc/kj.md
```diff
@@ -0,0 +1,84 @@
...
+ensuring that the empty case is handled, thus preventing null pointer dereferences and/or access to
+invalid objects at runtime.
+
+You can construct a `Maybe<T>` using either a value of type `T` or using `nullptr`:
+
+```c++
+#include <kj/common.h>
+
+auto number = kj::Maybe<double>(1.0);
+auto empty = kj::Maybe<double>(nullptr);
```
bretthoerner, line 26: This is of course valid, but I think more typical code looks like defining a `kj::Maybe<double> foo;` that is empty in advance, or as part of a class field and then later filling it in (or not!).

doc/kj.md
```diff
@@ -0,0 +1,84 @@
...
+unhandled cases using a familiar `switch`/`case` approach.
+
+You can construct a `OneOf<T>` using a value valid for any of the types:
+
+```c++
+#include <kj/one-of.h>
+
+auto numberInt = kj::OneOf<int, double, kj::String, bignum>(8);
+auto numberFloat = kj::OneOf<int, double, kj::String, bignum>(8.0);
+auto numberString = kj::OneOf<int, double, kj::String, bignum>("8");
```
bretthoerner, line 54: Same deal as with the `Maybe`, it's more natural in KJ (I think) to define a `OneOf` type that you later use, which might be useful to show here rather than `auto`.

bretthoerner (changes requested):

doc/kj.md
haneefmubarak: WRT this and [the above](https://github.com/capnproto/capnproto/pull/1141#discussion_r549881686), I think my main intent there is that reading-wise I always these kinds of one-liners to be the most succinctly informative.

I always like more examples though, so I'll add class/struct examples too that illustrate what you're talking about.

haneefmubarak:
Gonna throw fixups for the changes and then squash before ready-for-merge so I don't have to keep force pushing

haneefmubarak:
> I think it might be worth mentioning `KJ_{ASSERT/REQUIRE}_NONNULL` for getting the value out of a `Maybe<T>` if you know it's non-empty

I just realized I didn't even know this was an option, so yes, absolutely.

doc/kj.md
```diff
@@ -82,3 +82,69 @@ struct Empty {};
...
+```
+
+You can also use it as part of a class, in which case it's default initialized to `nullptr` until
+you initialize it otherwise:
+
+```c++
+class Continent {
+  ...
+
+  Continent(Thing creature) : creature(creature) {};
```
jclee, line 146: Pretty sure this needs to be more like:
```
  Continent(kj::Own<Thing> creature) : creature(kj::mv(creature)) {};
```

doc/kj.md
```diff
@@ -82,3 +82,69 @@ struct Empty {};
...
+  kj::Own<Thing> creature;
+
+  ...
+};
+
+Continent antarctica(kj::heap<Thing>(3, 999));
+```
+
+Since an `Own<T>` is fundamentally a wrapper around a pointer, you can use it in the same manner as
+a raw pointer when attempting to access its' internals:
```
jclee, line 159: s/its'/its

doc/kj.md
```diff
@@ -82,3 +82,69 @@ struct Empty {};
...
+Meanwhile, if you'd like to transfer ownership of the underlying pointer, you can simply move your
+existing `Own<T>` into another, which will set the original (existing) `Own<T>` to `nullptr`:
+
+```c++
+transport(kj::mv(ownedThing))
+```
+
+When the `Own<T>` goes out of scope or when an object containing an `Own<T>` is destructed, the
+destructor for the object pointed to by the `Own<T>` is called and then the underlying memory is
+released automatically.
```
jclee, line 176: That's true most of the time, because usually the object in the `kj::Own` has been allocated with `kj::heap`, which uses a disposer function that calls the destructor.  But there are other ways to create `kj::Own`s that use disposers that don't call the destructor.  For example, `kj::Own`s returned by `kj::refcount()/kj::addRef()` only call the destructor when the reference count drops to zero.  And `kj::Own`s returned by `kj::attachRef()` do nothing when destructed, with the assumption that something else will be taking care of it.  Might be better to just say something like "When an `Own<T>` is destructed, it performs any necessary destruction or deallocation of the object it points to."?

haneefmubarak:
> I think the usage is actually more like:
> 
> ```
> doSomething(KJ_ASSERT_NONNULL(maybeValue));
> ```
> 
> Not sure if it also might be worth pointing out the distinction of intent between "ASSERT" and "REQUIRE"... guess it depends on whether the discussion of KJ_ASSERT/KJ_REQUIRE are nearby. Likewise, could be worth mentioning that the macros would otherwise throw an exception for null values, if it's not otherwise apparent.
> 
> Also, no apostrophe needed for "its".

Gotcha, will fix usage. As for the explanation between ASSERT v REQUIRE, I was planning to add a paragraph about that in the yet to be written debugging macro section.

doc/kj.md
haneefmubarak: excellent catch

doc/kj.md
```diff
@@ -143,7 +141,7 @@ you initialize it otherwise:
 class Continent {
   ...
 
-  Continent(Thing creature) : creature(creature) {};
+  Continent(kj::Own<Thing> creature) : creature(creature) {};
```
jclee, line 144: Haven't tested, but I suspect the initializing value needs to be `kj::mv()`ed into the member for the constructor to compile without errors.  Also, extra semicolon.

haneefmubarak: You would absolutely be right and good catch. Fixing.

haneefmubarak:
As a few of us discussed elsewhere, gonna shift the checklist to an issue and move to merge this so we have somewhere to start off.

haneefmubarak:
Squashing for merge.

bretthoerner (approved):

kentonv:
FWIW I dug up the docs I was working on a year ago and there's a lot of overlap. I'm aiming to get them into a PR this week.

kentonv:
So it turns out I had written a whole bunch of docs a year ago and then sort of forgot about it as Covid hit and everything went crazy. Finally uploaded those in #1149. The docs I wrote already cover the stuff in your PR. Really sorry I didn't upload those earlier! I am kicking myself.

haneefmubarak:
Closing in favor of #1149.

PR #1143, opened 2021-01-05
Pretty-print binary blobs (aka Data) properly.

sobomax:
Adjust encodeCEscape() to generate encoding that is compatible with utf-8 which is used in python3 when presented with a binary blob, which means escaping upper half of the code space. Without this attempt to make printable representation of the valid otherwise binary data in python would sometimes fail with something like this:

```
Traceback (most recent call last):
  File "./tools/lib/logreader.py", line 147, in <module>
    print(msg)
  File "capnp/lib/capnp.pyx", line 1092, in capnp.lib.capnp._DynamicStructReader.__str__
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x9d in position 118: invalid start byte
```
The structure in question is:
```
struct CanData {
  address @0 :UInt32;
  busTime @1 :UInt16;
  dat     @2 :Data;
  src     @3 :UInt8;
}
```
The data:
```
( logMonoTime = 166125304492147,
  sendcan = [
    ( address = 832,
      busTime = 0,
      dat = "\004\000\000$d\000\235\021",
      src = 0 ) ],
  valid = true )
```

I've also tried to fix it at pycapnp level [here](https://github.com/sobomax/pycapnp/commit/18401754222bb3de2840029f6503558f2f04c4da) using "backslashreplace" mechanism, however that produces awfully inconsistently escaped output, so I think this is the best place to fix it.

```
        ( key = "PandaFirmware",
          value = "\021\001[\026|GSe\xd5\032+\x94\xff\f|" ),
```
(note both octal and hex escaping).

### DIFF
--- a/c++/src/kj/encoding.h
+++ b/c++/src/kj/encoding.h
@@ -246,6 +246,8 @@ const T* readMaybe(const EncodingResult<T>& value) {
   }
 }
 
+String encodeCEscapeImpl(ArrayPtr<const byte> bytes, bool isBinary);
+
 }  // namespace _ (private)
 
 inline String encodeUriComponent(ArrayPtr<const char> text) {
@@ -276,8 +278,13 @@ inline EncodingResult<String> decodeWwwForm(ArrayPtr<const char> text) {
 }
 
 inline String encodeCEscape(ArrayPtr<const char> text) {
-  return encodeCEscape(text.asBytes());
+  return _::encodeCEscapeImpl(text.asBytes(), false);
 }
+
+inline String encodeCEscape(ArrayPtr<const byte> bytes) {
+  return _::encodeCEscapeImpl(bytes, true);
+}
+
 inline EncodingResult<String> decodeCEscape(ArrayPtr<const char> text) {
   auto result = decodeBinaryCEscape(text, true);
   return { String(result.releaseAsChars()), result.hadErrors };
--- a/c++/src/capnp/stringify.c++
+++ b/c++/src/capnp/stringify.c++
@@ -139,17 +139,14 @@ static kj::StringTree print(const DynamicValue::Reader& value,
       } else {
         return kj::strTree(value.as<double>());
       }
-    case DynamicValue::TEXT:
+    case DynamicValue::TEXT: {
+      kj::ArrayPtr<const char> chars = value.as<Text>();
+      return kj::strTree('"', kj::encodeCEscape(chars), '"');
+    }
     case DynamicValue::DATA: {
       // TODO(someday): Maybe data should be printed as binary literal.
-      kj::ArrayPtr<const char> chars;
-      if (value.getType() == DynamicValue::DATA) {
-        chars = value.as<Data>().asChars();
-      } else {
-        chars = value.as<Text>();
-      }
-
-      return kj::strTree('"', kj::encodeCEscape(chars), '"');
+      kj::ArrayPtr<const byte> bytes = value.as<Data>().asBytes();
+      return kj::strTree('"', kj::encodeCEscape(bytes), '"');
     }
     case DynamicValue::LIST: {
       auto listValue = value.as<DynamicList>();
--- a/c++/src/kj/encoding-test.c++
+++ b/c++/src/kj/encoding-test.c++
@@ -362,10 +362,12 @@ KJ_TEST("application/x-www-form-urlencoded encoding/decoding") {
 }
 
 KJ_TEST("C escape encoding/decoding") {
-  KJ_EXPECT(encodeCEscape("fooo\a\b\f\n\r\t\v\'\"\\bar") ==
-      "fooo\\a\\b\\f\\n\\r\\t\\v\\\'\\\"\\\\bar");
+  KJ_EXPECT(encodeCEscape("fooo\a\b\f\n\r\t\v\'\"\\barПривет, Мир! Ж=О") ==
+      "fooo\\a\\b\\f\\n\\r\\t\\v\\\'\\\"\\\\bar\xd0\x9f\xd1\x80\xd0\xb8\xd0\xb2\xd0\xb5\xd1\x82\x2c\x20\xd0\x9c\xd0\xb8\xd1\x80\x21\x20\xd0\x96\x3d\xd0\x9e");
   KJ_EXPECT(encodeCEscape("foo\x01\x7fxxx") ==
       "foo\\001\\177xxx");
+  byte bytes[] = {'f', 'o', 'o', 0, '\x01', '\x7f', 'x', 'x', 'x', 128, 254, 255};
+  KJ_EXPECT(encodeCEscape(bytes) == "foo\\000\\001\\177xxx\\200\\376\\377");
 
   expectRes(decodeCEscape("fooo\\a\\b\\f\\n\\r\\t\\v\\\'\\\"\\\\bar"),
       "fooo\a\b\f\n\r\t\v\'\"\\bar");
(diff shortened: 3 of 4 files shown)


kentonv:
Hi @sobomax,

Thanks for contributing! Unfortunately, this changes behavior in a way that many users would find undesirable: Now, if you pass in a valid UTF-8 string (which `kj::StringPtr`s are supposed to be, after all), the non-ASCII code points will be escaped as a sequence of ugly hex escapes. In many use cases (such as when generating actual C/C++ code, which is what the function claims to be for), it would be preferable to keep the UTF-8 sequences unescaped.

The only way we could support this is as some sort of option flag, which is off by default. But, this function is not very complicated. It might make more sense for you to copy/paste it into your own codebase and modify it there to meet your needs. Otherwise I worry the options here could easily proliferate -- maybe some people only want invalid sequences escaped, maybe some want `\u` escapes, maybe someone wants to disable the `\a` escape sequence since Java/JavaScript don't support it, etc. Probably better for people to just copy and customize...

sobomax:
@kentonv thanks for a feedback, just skimming through python module code that routine is called zillion times as part of the prettyPrint() macro expansion. Which means it will also need to be replicated. :(

What's the use case for the escape logic to escape 0x7f but keep upper 128 characters verbatim in C code and why escaping it would upset some people? It's kinda not here not there. It frankly looked like a typo for me.

kentonv:
> What's the use case for the escape logic to escape 0x7f but keep upper 128 characters verbatim in C code

What I said in my last comment. Character codes above 0x7f are UTF-8 sequences. Both the input and the output of this function are UTF-8 strings, so there's no need to escape UTF-8.

kentonv:
Ah I see what might be a point of confusion: The input of this function is declared as `ArrayPtr<const byte>`. However, there is an overload that takes `ArrayPtr<const char>` and just reinterprets the chars to bytes and calls the other function. The version taking chars is, I think, the more common use case. So the input is usually text, and the output is also text, so there's no need to escape UTF-8 in the text.

sobomax:
@kentonv I think the call site leading to this issue is here:

```
static kj::StringTree print(const DynamicValue::Reader& value,
                            schema::Type::Which which, Indent indent,
                            PrintMode mode) {
  switch (value.getType()) {
[...]
    case DynamicValue::TEXT:
    case DynamicValue::DATA: {
      // TODO(someday): Maybe data should be printed as binary literal.
      kj::ArrayPtr<const char> chars;
      if (value.getType() == DynamicValue::DATA) {
        chars = value.as<Data>().asChars();
      } else {
        chars = value.as<Text>();
      }

      return kj::strTree('"', kj::encodeCEscape(chars), '"');
    }
```

Are you saying perhaps we should somehow pass (value.getType() == DynamicValue::DATA) into encodeCEscape() as a bool argument and handle upper 128 bytes accordingly? Would that elevate your concern? As you can see the fix on upper layer is not that trivial, since print() method is quite elaborate and we are not even calling that directly...

```
cdef extern from "<capnp/pretty-print.h>" namespace " ::capnp":
    StringTree printStructReader" ::capnp::prettyPrint"(C_DynamicStruct.Reader) except +reraise_kj_exception
    StringTree printStructBuilder" ::capnp::prettyPrint"(DynamicStruct_Builder) except +reraise_kj_exception
    StringTree printRequest" ::capnp::prettyPrint"(Request &) except +reraise_kj_exception
    StringTree printListReader" ::capnp::prettyPrint"(C_DynamicList.Reader) except +reraise_kj_exception
    StringTree printListBuilder" ::capnp::prettyPrint"(C_DynamicList.Builder) except +reraise_kj_exception
```

kentonv:
Ahhhhh I see. What you really want to change is how `Data` fields are stringified -- I didn't get that before.

OK, I do agree that, at the very least, invalid UTF-8 sequences should be hex-escaped here. _Valid_ UTF-8, I think, can still be allowed through, right?

Of course, if the data is not intended to be text at all, then trying to represent it as escaped text seems non-ideal. It turns out the parser for our text format actually supports an alternate, hex encoding for data, which looks like:

```
dataField = 0x"abcd1234"
```

Perhaps we could decide on a heuristic at which point we decide to to switch to this format. Maybe if more than some percentage of bytes end up hex-escaped, we just switch to hex format?

sobomax:
@kentonv that's my hypothesis. I am not 100% sure the value in question is DATA or TEXT, but I suspect the former, at least it makes the most sense. I have concocted the following patch, which I hope will fix it: https://github.com/sobomax/capnproto/commit/1f000d2af1956a09dede5fb326ef9824720f653f. Maybe if you can take a look and let me know if there are any issues. I am trying to verify that it works now and will update PR accordingly.

sobomax:
@kentonv thanks again for pushing me into the right direction. Further testing revealed the issue is actually on both sides. The app in question on one hand abused capnp by pushing binary blobs into Text (https://github.com/commaai/cereal/pull/109, https://github.com/commaai/openpilot/pull/19693) but also it stored some legit binary blobs that were failing to pretty-print. As such, I've updated this PR to only fix capnp's side of things and adjusted description to match. As far as I can tell, this new approach should be safe for other users of the `encodeCEscape`, as the only place where `ArrayPtr<const char> bytes` used [is here](https://github.com/capnproto/capnproto/pull/1143/files#diff-fc1034567ff85e298eb13a83eb187e0176ee65f22ba93080cad7f1cc48896731R149). Please take a look and let me know how it looks. Thanks!

kentonv:
> as the only place where ArrayPtr<const char> bytes being passed is here.

This is a public API. Anyone who uses KJ could be calling it. For example, the Cloudflare Workers Runtime codebase contains four calls to this function (all passing in bytes). We don't know what other codebases out there might be calling it.

Looking at the Workers Runtime use cases:
* Three of them are cases in tests where some data read off a stream did not match the expected value, and we want to log what was read, but we have no idea if it is binary or text. If it is text, we want it to be legible as text, but if it is binary data, we don't want it to mess up the terminal. I guess that because the team works in English, when our test data is text, it's likely to be ASCII text, so won't be harmed. And when it is binary, the new behavior is likely a win.
* The other call site is also in test code, and is constructing capnp text format in order to be parsed again by another component. No human ever reads the string so it doesn't matter how we escape it.

So... I guess in our use cases, this behavior change would be OK.

It's not clear if we have enough data to extrapolate how this is likely to affect other users. But I guess I can be convinced that it's unlikely to outright break anyone, and it's probably generally preferable to escape high bytes when the input was explicitly bytes and not chars. People can always use `.asChars()` if they really wanted to treat the array as text.

But in any case, for future reference, KJ public APIs are used by more than just the Cap'n Proto project, so we need to be cautious about these kinds of changes.

c++/src/kj/encoding.h
```diff
@@ -197,6 +197,7 @@ EncodingResult<Array<byte>> decodeBinaryUriComponent(
 // Decode URI components using % escapes. This is a lower-level interface used to implement both
 // `decodeUriComponent()` and `decodeWwwForm()`
 
+String encodeCEscapeImpl(ArrayPtr<const byte> bytes, bool isBinary);
```
kentonv, line 200: This should be under `namespace _ {  // private` below.

sobomax: Adjusted & updated PR, thanks! I've also extended encodeCEscape(chars) with UTF-8 text and added small test for the encodeCEscape(bytes).

sobomax:
@kentonv the code has been tidied up wrt using private namespace. I've also extended test case to exercise new functionality and perhaps give API users some idea of what to expect when bytes are passed in. I hope it makes this ready to land. Thanks!

kentonv:
Thanks!

PR #1145, opened 2021-01-07
Add writeMessages for serializing multiple messages to an output stream

a-robinson:
In order to support batching up some of the small writes that capnproto
tends to do in practice when you're sending lots of pipelined RPCs.

---

This could use unit tests, and perhaps should also have versions of the methods added onto the `MessageStream` type, but I wanted to get your feedback on the core approach before going the rest of the way.

### DIFF
--- a/c++/src/capnp/serialize-async.c++
+++ b/c++/src/capnp/serialize-async.c++
@@ -252,37 +252,82 @@ struct WriteArrays {
   kj::Array<kj::ArrayPtr<const byte>> pieces;
 };
 
+inline size_t tableSizeForSegments(size_t segmentsSize) {
+  return (segmentsSize + 2) & ~size_t(1);
+}
+
+// Helper function that allocates and fills the pointed-to table with info about the segments and
+// populates the pieces array with pointers to the segments.
+void fillWriteArraysWithMessage(kj::ArrayPtr<const kj::ArrayPtr<const word>> segments,
+                                kj::ArrayPtr<_::WireValue<uint32_t>> table,
+                                kj::ArrayPtr<kj::ArrayPtr<const byte>> pieces) {
+  KJ_REQUIRE(segments.size() > 0, "Tried to serialize uninitialized message.");
+
+  // We write the segment count - 1 because this makes the first word zero for single-segment
+  // messages, improving compression.  We don't bother doing this with segment sizes because
+  // one-word segments are rare anyway.
+  table[0].set(segments.size() - 1);
+  for (uint i = 0; i < segments.size(); i++) {
+    table[i + 1].set(segments[i].size());
+  }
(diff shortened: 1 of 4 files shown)


a-robinson:
Sorry, it looks like I should have waited for the build/tests to pass after rebasing onto master instead of just assuming the rebase didn't mess anything up 🤦 Looking into the build failures.

kentonv:
Hmm, seems like this should be a method of `MessageStream`. I know our internal `VatNetwork` implementation doesn't use `MessageStream` yet since it was forked from `TwoPartyVatNetwork` before `MessageStream` existed, but probably we should change that.

c++/src/capnp/serialize-async.h
```diff
@@ -197,6 +209,15 @@ inline kj::Promise<void> MessageStream::writeMessage(
   return writeMessage(fds, builder.getSegmentsForOutput());
 }
 
+inline kj::Promise<void> writeMessages(
+    kj::AsyncOutputStream& output, kj::ArrayPtr<MessageBuilder*> builders) {
+  auto messages = kj::heapArray<kj::ArrayPtr<const kj::ArrayPtr<const word>>>(builders.size());
```
kentonv, line 214: This function is too heavy to be inlined.

c++/src/capnp/serialize-async.c++
```diff
@@ -252,39 +252,77 @@ struct WriteArrays {
...
 }
 
+template <typename WriteFunc>
+kj::Promise<void> writeMessagesImpl(
+    kj::ArrayPtr<kj::ArrayPtr<const kj::ArrayPtr<const word>>> messages, WriteFunc&& writeFunc) {
+  KJ_REQUIRE(messages.size() > 0, "Tried to serialize zero messages.");
+
+  // Construct separate tables for each message, but one shared array of pieces that all the
+  // messages get combined into.
+  auto tables = kj::heapArray<kj::Array<_::WireValue<uint32_t>>>(messages.size());
```
kentonv, line 305: I suppose technically it would be possible to allocate one big array containing all the tables, then chop it up... but I guess that'd make the code considerably more complicated for not an enormous benefit.

a-robinson: Eh, it does make the code a little tougher to read, but it's not too bad. I'll throw that in.

a-robinson:
> Hmm, seems like this should be a method of `MessageStream`. I know our internal `VatNetwork` implementation doesn't use `MessageStream` yet since it was forked from `TwoPartyVatNetwork` before `MessageStream` existed, but probably we should change that.

Yup, I've added that in now.

The three fixups are:
* Un-inline the non-trivial function and add `MessageStream` versions of the methods
* Use one large array for the multiple tables
* Add a basic unit test

PR #1146, opened 2021-01-11
Optimise WebSocket pumps between raw streams to a raw stream pump.

kentonv:
This way we don't have to parse each frame that passes through.

### DIFF
--- a/c++/src/kj/compat/http-test.c++
+++ b/c++/src/kj/compat/http-test.c++
@@ -516,6 +516,15 @@ kj::Promise<void> expectRead(kj::AsyncInputStream& in, kj::ArrayPtr<const byte> 
   }));
 }
 
+kj::Promise<void> expectEnd(kj::AsyncInputStream& in) {
+  static char buffer;
+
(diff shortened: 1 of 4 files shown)


c++/src/kj/compat/http-test.c++
```diff
@@ -516,6 +516,15 @@ kj::Promise<void> expectRead(kj::AsyncInputStream& in, kj::ArrayPtr<const byte>
   }));
 }
 
+kj::Promise<void> expectEnd(kj::AsyncInputStream& in) {
+  static char buffer;
+
+  auto promise = in.tryRead(&buffer, 1, 1);
+  return promise.then([&in](size_t amount) {
```
harrishancock, line 523: The compiler wanted me to tell you about this very important error.
```
error: lambda capture 'in' is not used [-Werror,-Wunused-lambda-capture]
```

harrishancock (commented):
Looks good, just a simple build failure.

(Feel free to merge after fixing; I didn't approve only so you wouldn't accidentally merge right away.)

harrishancock (approved):

PR #1147, opened 2021-01-12
Fix link to Google protobuf

Omustardo:
Based on archive.org, this link has been broken since ~2017.

The latest archive.org link with a redirect that I found directs to https://code.google.com/p/protobuf/ which in turn redirects to http://github.com/google/protobuf

### DIFF
--- a/doc/index.md
+++ b/doc/index.md
@@ -8,7 +8,7 @@ title: Introduction
 <img src='images/infinity-times-faster.png' style='width:334px; height:306px; float: right;'>
 
 Cap'n Proto is an insanely fast data interchange format and capability-based RPC system. Think
-JSON, except binary. Or think [Protocol Buffers](http://protobuf.googlecode.com), except faster.
+JSON, except binary. Or think [Protocol Buffers](https://github.com/google/protobuf), except faster.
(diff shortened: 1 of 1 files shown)


kentonv:
Hmm I guess github.com/google/protobuf further redirects to github.com/protocolbuffers/protobuf, maybe we should link directly to that?

jclee:
Another possibility is to just link to https://en.wikipedia.org/wiki/Protocol_Buffers .

ocdtrekkie:
IMHO, linking from a technical repo to another technical repo is the right call. Wikipedia is generally intended for a little bit different audience.

kentonv:
To be fair this is the front page of capnproto.org we're talking about. Wikipedia _could_ make sense there, but I think github is probably still the way to go.

Omustardo:
Updated the link to github.com/protocolbuffers/protobuf.

Personally I would find the github page much more useful than the wiki.

Another option is https://developers.google.com/protocol-buffers/ which is linked from the protocolbuffers/protobuf github page.

kentonv:
Thanks!

(Unfortunately I've made the process for publishing site updates overcomplicated, so I haven't publishing this yet, but hopefully will at some point...)

PR #1152, opened 2021-01-14
Allow limiting WebSocket message size.

kentonv:
(@harrishancock, you already reviewed this, and it's already deployed.)


kentonv:
Since this was already reviewed previously, merging.

harrishancock:
Ah yes, I saw this, but I was still in bed, and my phone wasn't logged into GitHub, and then I forgot about it.

Issue #963, opened 2020-04-01
Should use const references in for loops in dynamic.c++


brooksmoses:
Hello!

(Apologies for sending this as a bug rather than a pull request, but I don't have a git checkout handy.)

We have a copy of capnproto in our internal codebase, and a recent automated Clang-tidy check found a couple of cases of range-based for loops in `dynamic.c++` that should probably use const references.

Specifically, on lines 1017 and 1419, 
```
    for (auto element: value) {
```
should probably instead be
```
    for (const auto& element : value) {
```
to avoid unnecessary copying of the `element` objects.

kentonv:
Happy to accept a PR.

PR #1161, opened 2021-01-20
Fix range-based for loops

Illumaria:
#963 
* Changed iteration type in range-based "for" loops from value to const reference where needed.
* Changed postfix to prefix increment in common "for" loops where needed.


kentonv:
This PR doesn't even compile. Please do not submit PRs without compiling the code first.

Many of these range-based `for`s actually _have to_ be by value, due to the way the Cap'n Proto API works. When iterating over a capnp list Reader, the `Reader` for each element is actually created upon dereferencing the iterator, hence there is nothing to form a reference to.

PR #1162, opened 2021-01-21
KJ doc: fix link to style guide

shamrin:


a-robinson (approved):
Heh, I'd have thought github would be smart enough to follow the symlink, but it looks like it doesn't. Thanks for the fix.

kentonv:
While this fixes browsing in github itself, it probably beaks eventual generation of a web site from these files as sources, so isn't really the right fix.

Meanwhile I didn't move `style-guide.md` into `kjdoc` because lots of links exist around the internet to that file on github.

Not sure what's the right thing to do here.

PR #1163, opened 2021-01-21
Document additional valid backwards-compatible schema transformations

jviotti:
I have been running a set of schema evolution experiments using Cap'n
Proto and I believe that the schema transformations I'm documenting in
this commit are all backwards-compatible, at least as far as my
experiments went.

Please let me know your thoughts! I can double check from my side (and provide examples) if you think some of these transformations sound odd. I'm also not sure if you want to mention the enumeration constant to integer transformation given that the documentation explicitly asks the reader to not think about enumeration constants as numeric types.

Signed-off-by: Juan Cruz Viotti <jv@jviotti.com>

### DIFF
--- a/doc/language.md
+++ b/doc/language.md
@@ -801,6 +801,14 @@ this list, but most apps can safely use them:
   As a special exception to this rule, `List(Bool)` may **not** be upgraded to a list of structs,
   because implementing this for bit lists has proven unreasonably expensive.
 
+* A integer field may be promoted to an integer field of larger width while keeping the sign
+* A boolean field may be promoted to an integer field
+* A `Text` field may be promoted to `Data`
+* A `Data` field may be promoted to `List(UInt8)`
+* An enumeration may be promoted to `Uint16`, `UInt32`, or `UInt64`
+* An unsigned integer may be promoted to a signed integer if the positive part of the signed integer 
+  is wide enough to contain the unsigned integer
+
 Any change not listed above should be assumed NOT to be safe.  In particular:
 
 * You cannot change a field, method, or enumerant's number.


jviotti:
Additionally, the spec mentions that:

> You cannot move an existing field into or out of an existing union, nor can you form a new union containing more than one existing field.

However moving an optional field into an existing union is a forwards-compatible change. In this case, the new schema has i.e. 3 choices inside a single union and the older schema has 2 choices inside the union and an additional field outside the union. The older schema can always parse data produced by the newer schema and either the union field or the additional field will be unset.

Do you think it'd be worth mentioning this?

kentonv:
Hi @jviotti,

Thanks, but unfortunately, most of these are not backwards-compatible. They may appear to work in your tests because you are testing a struct containing only the one field. However, if your struct contained other fields _after_ the one you are changing, you'd find that those later fields are broken by many of these changes.

Keep in mind that fields in a Cap'n Proto struct are identified by their offset from the beginning of the struct, not by any sort of tag like in Protobuf. So imagine if you have:

```capnp
struct Foo {
  i @0 :UInt8;
  j @1 :UInt8;
}
```

In this case, `i` is located in the first byte of the struct, and `j` is the second byte. Now if you change `i` to be `UInt16` instead, then `j` is pushed to be the third byte. This is not backwards-compatible.

In order to decide what is and is not compatible, you really need to read the encoding spec. You can't rely on tests for this, as your tests can easily miss cases.

doc/language.md
```diff
@@ -801,6 +801,14 @@ this list, but most apps can safely use them:
   As a special exception to this rule, `List(Bool)` may **not** be upgraded to a list of structs,
   because implementing this for bit lists has proven unreasonably expensive.
 
+* A integer field may be promoted to an integer field of larger width while keeping the sign
+* A boolean field may be promoted to an integer field
```
kentonv, line 805: Not compatible: Similar to integers. If you have two boolean fields, they will be packed in adjacent bits. If you change the first one to an integer, the second one will move.

doc/language.md
```diff
@@ -801,6 +801,14 @@ this list, but most apps can safely use them:
...
+* A `Text` field may be promoted to `Data`
+* A `Data` field may be promoted to `List(UInt8)`
```
kentonv, line 807: This is actually compatible.

doc/language.md
```diff
@@ -801,6 +801,14 @@ this list, but most apps can safely use them:
...
+* An enumeration may be promoted to `Uint16`, `UInt32`, or `UInt64`
```
kentonv, line 808: An enumeration can safely be changed to `UInt16`, because they are the same size. It cannot be changed to any larger type, for the same reason integers cannot be changed to larger types.

doc/language.md
```diff
@@ -801,6 +801,14 @@ this list, but most apps can safely use them:
...
+* An unsigned integer may be promoted to a signed integer if the positive part of the signed integer
```
kentonv, line 809: As long as the integers are the same size, then changing signedness is compatible, with the obvious two's complement semantics.

kentonv (commented):
> moving an optional field into an existing union is a forwards-compatible change.

Not true: Fields in a union will occupy the same offset in the struct if possible. Fields not in a union must have different offsets. So moving a field into an existing union will likely change the offsets of the later-numbered field, and all subsequent fields.

doc/language.md
kentonv: However, there is a caveat: Normally a `List(UInt8)` can be promoted to `List(SomeStruct)` where `SomeStruct`'s first field has type `UInt8`. If this upgrade occurs, then programs expecting a `List(UInt8)` will be able to read new data just fine, but programs expecting `Data` will not. Hence I don't document this as compatible because it interacts badly with other compatibility rules.

jviotti:
> Thanks, but unfortunately, most of these are not backwards-compatible. They may appear to work in your tests because you are testing a struct containing only the one field. However, if your struct contained other fields after the one you are changing, you'd find that those later fields are broken by many of these changes.

I see. Thanks a lot for the explanations. My tests, as you indicate, as very simple and might be showing false positives. I'll see if I can make them more robust!

doc/language.md
```diff
@@ -801,6 +801,14 @@ this list, but most apps can safely use them:
   As a special exception to this rule, `List(Bool)` may **not** be upgraded to a list of structs,
   because implementing this for bit lists has proven unreasonably expensive.
 
+* A integer field may be promoted to an integer field of larger width while keeping the sign
+* A boolean field may be promoted to an integer field
+* A `Text` field may be promoted to `Data`
```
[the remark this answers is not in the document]
jviotti: I might be missing something, but a `Text` field includes a NUL-delimiter. When transforming the `Text` into a `Data`, the whole byte-string is preserved exactly as it is, including the `NUL-delimiter` (I'll double check my test now), which means that there is no lost information, right?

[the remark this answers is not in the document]
kentonv: That's correct. The point is, the NUL terminator is not normally considered to be part of the text itself (e.g. text.size() doesn't include it), but it will be considered part of the data (data.size() _will_ include it), which may be confusing or lead to bugs. But it's not hard to adjust the program code to account for this, in which case the change is acceptable.

Relatedly, if you want old programs to be able to understand messages written by new programs, then the new program has to always append a NUL terminator to the data, so that it can be interpreted as text by old programs.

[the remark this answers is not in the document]
jviotti: That's a good point. Thanks again for the clarification!

PR #1164, opened 2021-01-21
`kj::defer`: allow lvalue lambda reference

haneefmubarak:
This allows for the usage of named / reusable lambdas with `kj::defer()`, which is especially helpful in testing functions.

### DIFF
--- a/c++/src/kj/common-test.c++
+++ b/c++/src/kj/common-test.c++
@@ -596,5 +596,33 @@ KJ_TEST("kj::range()") {
   KJ_EXPECT(expected == 8);
 }
 
+KJ_TEST("kj::defer()") {
+  bool executed;
+
+  // rvalue reference
+  {
+    executed = false;
+    auto deferred = kj::defer([&executed]() {
+      executed = true;
+    });
+    KJ_EXPECT(!executed);
+  }
+
+  KJ_EXPECT(executed);
+
+  // lvalue reference
+  auto executor = [&executed]() {
+    executed = true;
+  };
+
+  {
+    executed = false;
+    auto deferred = kj::defer(executor);
+    KJ_EXPECT(!executed);
+  }
+
+  KJ_EXPECT(executed);
+}
+
 }  // namespace
 }  // namespace kj
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -1604,7 +1604,7 @@ public:
   KJ_DISALLOW_COPY(Deferred);
 
   // This move constructor is usually optimized away by the compiler.
-  inline Deferred(Deferred&& other): func(kj::mv(other.func)), canceled(false) {
+  inline Deferred(Deferred&& other): func(kj::fwd<Func>(other.func)), canceled(false) {
     other.canceled = true;
   }
 private:


c++/src/kj/common.h
```diff
@@ -1603,8 +1603,7 @@ class Deferred {
   inline ~Deferred() noexcept(false) { if (!canceled) func(); }
   KJ_DISALLOW_COPY(Deferred);
 
-  // This move constructor is usually optimized away by the compiler.
```
kentonv, line 1606: The comment is still accurate (it's referring to the constructor, not the use of `kj::mv`).

kentonv (commented):
Can we add a unit test?

c++/src/kj/common.h
haneefmubarak: gotcha, fixing

kentonv:
Still needs unit test.

haneefmubarak:
Apologies, missed that in the email notif; added.

c++/src/kj/common-test.c++
```diff
@@ -596,5 +596,31 @@ KJ_TEST("kj::range()") {
...
+KJ_TEST("kj::defer()") {
+  bool executed;
+
+  // rvalue reference
+  {
+    executed = false;
+    kj::defer([&executed]() {
+      executed = true;
+    });
+  }
```
kentonv, line 608: Maybe add `KJ_EXPECT(!executed)` before leaving the block, to prove that `defer` isn't executing synchronously.

(As written right now, it is executing synchronously since you didn't assign the defer object to a variable. :) )

haneefmubarak:
I guess GitHub doesn't support fixup-then-rebase in a visible manner...

c++/src/kj/common-test.c++
haneefmubarak: Good catch! Done.

c++/src/kj/common-test.c++
```diff
@@ -596,5 +596,33 @@ KJ_TEST("kj::range()") {
...
+KJ_TEST("kj::defer()") {
+  bool executed;
+
+  // rvalue reference
+  {
+    executed = false;
+    kj::defer([&executed]() {
+      KJ_EXPECT(!executed);
+      executed = true;
+    });
```
kentonv, line 608: This is what I meant:

```suggestion
    executed = false;
    auto deferred = kj::defer([&executed]() {
      executed = true;
    });
    KJ_EXPECT(!executed);
```

c++/src/kj/common-test.c++
```diff
@@ -596,5 +596,33 @@ KJ_TEST("kj::range()") {
...
+
+  // lvalue reference
+  auto executor = [&executed]() {
+    KJ_EXPECT(!executed);
+    executed = true;
+  };
+
+  {
+    executed = false;
+    kj::defer(executor);
```
kentonv, line 621:
```suggestion
    auto deferred = kj::defer(executor);
    KJ_EXPECT(!executed);
```

haneefmubarak:
Ahhh my bad, I see what you mean. Fixing and squashing.

PR #1165, opened 2021-01-22
Fill in KJ 'tour' documentation of Promises and System I/O.

kentonv:
/cc @haneefmubarak @thomcc @xortive

### DIFF
--- a/kjdoc/tour.md
+++ b/kjdoc/tour.md
@@ -1,0 +1,556 @@
+---
+title: A tour of KJ
+---
+
+This page is a tour through the functionality provided by KJ. It is intended for developers new to KJ who want to learn the ropes.
+
+**This page is not an API reference.** KJ's reference documentation is provided by comments in the headers themselves. Keeping reference docs in the headers makes it easy to find them using your editor's "jump to declaration" hotkey. It also ensures that the documentation is never out-of-sync with the version of KJ you are using.
+
+Core Programming
+======================================================================
+
+This section covers core KJ features used throughout nearly all KJ-based code.
+
+Every KJ developer should familiarize themselves at least with this section.
+
+## Core Utility Functions
+
+### Argument-passing: move, copy, forward
+
+`kj::mv` has exactly the same semantics as `std::move`, but takes fewer keystrokes to type. Since this is used extraordinarily often, saving a few keystrokes really makes a legitimate difference. If you aren't familiar with `std::move`, I recommend reading up on [C++11 move semantics](https://stackoverflow.com/questions/3106110/what-is-move-semantics).
+
+`kj::cp` is invoked in a similar way to `kj::mv`, but explicitly invokes the copy constructor of its argument, returning said copy. This is occasionally useful when invoking a function that wants an rvalue reference as a parameter, which normally requires pass-by-move, but you really want to pass it a copy.
+
+`kj::fwd`, is equivalent to `std::forward`. It is used to implement [perfect forwarding](https://en.cppreference.com/w/cpp/utility/forward), that is, forwarding arbitrary arguments from a template function into some other function without understanding their types.
+
+### Deferring code to scope exit
+
+This macro declares some code which must execute when exiting the current scope (whether normally or by exception). It is essentially a shortcut for declaring a class with a destructor containing said code, and instantiating that destructor. Example:
+
+```c++
+void processFile() {
+  int fd = open("file.txt", O_RDONLY);
+  KJ_ASSERT(fd >= 0);
+
+  // Make sure file is closed on return.
+  KJ_DEFER(close(fd));
+
+  // ... do something with the file ...
+}
+```
+
+You can also pass a multi-line block (in curly braces) as the argument to `KJ_DEFER`.
+
+There is also a non-macro version, `kj::defer`, which takes a lambda as its argument, and returns an object that invokes that lambda on destruction. The returned object has move semantics. This is convenient when the scope of the deferral isn't necessarily exactly function scope, such as when capturing context in a callback. Example:
+
+```c++
+kj::Function<void(int arg)> processFile() {
+  int fd = open("file.txt", O_RDONLY);
+  KJ_ASSERT(fd >= 0);
+
+  // Make sure file is closed when the returned function
+  // is eventually destroyed.
+  auto deferredClose = kj::defer([fd]() { close(fd); });
+
+  return [fd, deferredClose = kj::mv(deferredClose)]
+         (int arg) {
+    // ... do something with fd and arg ...
+  }
+}
+```
+
+Sometimes, you want a deferred action to occur only when the scope exits normally via `return`, or only when it exits due to an exception. For those purposes, `KJ_ON_SCOPE_SUCCESS` and `KJ_ON_SCOPE_FAILURE` may be used, with the same syntax as `KJ_DEFER`.
+
+### Size and range helpers
+
+`kj::size()` accepts a built-in array or a container as an argument, and returns the number of elements. In the case of a container, the container must implement a `.size()` method. The idea is that you can use this to find out how many iterations a range-based `for` loop on that container would execute. That said, in practice `kj::size` is most commonly used with arrays, as a shortcut for something like `sizeof(array) / sizeof(array[0])`.
+
+```c++
+int arr[15];
+KJ_ASSERT(kj::size(arr) == 15);
+```
+
+`kj::range(i, j)` returns an iterable that contains all integers from `i` to `j` (including `i`, but not including `j`). This is typically used in `for` loops:
+
+```c++
+for (auto i: kj::range(5, 10)) {
+  KJ_ASSERT(i >= 5 && i < 10);
+}
+```
+
+In the very-common case of iterating from zero, `kj::zeroTo(i)` should be used instead of `kj::range(0, i)`, in order to avoid ambiguity about what type of integer should be generated.
+
+`kj::indices(container)` is equivalent to `kj::zeroTo(kj::size(container))`. This is extremely convenient when iterating over parallel arrays.
+
+```c++
+KJ_ASSERT(foo.size() == bar.size());
+for (auto i: kj::indices(foo)) {
+  foo[i] = bar[i];
+}
+```
+
+`kj::repeat(value, n)` returns an iterable that acts like an array of size `n` where every element is `value`. This is not often used, but can be convenient for string formatting as well as generating test data.
+
+### Casting helpers
+
+`kj::implicitCast<T>(value)` is equivalent to `static_cast<T>(value)`, but will generate a compiler error if `value` cannot be implicitly cast to `T`. For example, `static_cast` can be used for both upcasts (derived type to base type) and downcasts (base type to derived type), but `implicitCast` can only be used for the former.
+
+`kj::downcast<T>(value)` is equivalent to `static_cast<T>(value)`, except that when compiled in debug mode with RTTI available, a runtime check (`dynamic_cast`) will be performed to verify that `value` really has type `T`. Use this in cases where you are casting a base type to a derived type, and you are confident that the object is actually an instance of the derived type. The debug-mode check will help you catch bugs.
+
+`kj::dynamicDowncastIfAvailable<T>(value)` is like `dynamic_cast<T*>(value)` with two differences. First, it returns `kj::Maybe<T&>` instead of `T*`. Second, if the program is compiled without RTTI enabled, the function always returns null. This function is intended to be used to implement optimizations, where the code can do something smarter if `value` happens to be of some specific type -- but if RTTI is not available, it is safe to skip the optimization. See [KJ idiomatic use of dynamic_cast](style-guide.md#dynamic_cast) for more background.
+
+### Min/max, numeric limits, and special floats
+
+`kj::min()` and `kj::max()` return the minimum and maximum of the input arguments, automatically choosing the appropriate return type even if the inputs are of different types.
+
+`kj::minValue` and `kj::maxValue` are special constants that, when cast to an integer type, become the minimum or maximum value of the respective type. For example:
+
+```c++
+int16_t i = kj::maxValue;
+KJ_ASSERT(i == 32767);
+```
+
+`kj::inf()` evaluates to floating-point infinity, while `kj::nan()` evaluates to floating-point NaN. `kj::isNaN()` returns true if its argument is NaN.
+
+### Explicit construction and destruction
+
+`kj::ctor()` and `kj::dtor()` explicitly invoke a constructor or destructor in a way that is readable and convenient. The first argument is a reference to memory where the object should live.
+
+These functions should almost never be used in high-level code. They are intended for use in custom memory management, or occasionally with unions that contain non-trivial types (but consider using `kj::OneOf` instead). You must understand C++ memory aliasing rules to use these correctly.
+
+## Ownership and memory management
+
+KJ style makes heavy use of [RAII](style-guide.md#raii-resource-acquisition-is-initialization). KJ-based code should never use `new` and `delete` directly. Instead, use the utilities in this section to manage memory in a RAII way.
+
+### Owned pointers, heap allocation, and disposers
+
+`kj::Own<T>` is a pointer to a value of type `T` which is "owned" by the holder. When a `kj::Own` goes out-of-scope, the value it points to will (typically) be destroyed and freed.
+
+`kj::Own` has move semantics. Thus, when used as a function parameter or return type, `kj::Own` indicates that ownership of the object is being transferred.
+
+`kj::heap<T>(args...)` allocates an object of type `T` on the heap, passing `args...` to its constructor, and returns a `kj::Own<T>`. This is the most common way to create owned objects.
+
+However, a `kj::Own` does not necessarily refer to a heap object. A `kj::Own` is actually implemented as a pair of a pointer to the object, and a pointer to a `kj::Disposer` object that knows how to destroy it; `kj::Own`'s destructor invokes the disposer. `kj::Disposer` is an abstract interface with many implementations. `kj::heap` uses an implementation that invokes the object's destructor then frees its underlying space from the heap (like `delete` does), but other implementations exist. Alternative disposers allow an application to control memory allocation more precisely when desired.
+
+Some example uses of disposers include:
+
+* `kj::fakeOwn(ref)` returns a `kj::Own` that points to `ref` but doesn't actually destroy it. This is useful when you know for sure that `ref` will outlive the scope of the `kj::Own`, and therefore heap allocation is unnecessary. This is common in cases where, for example, the `kj::Own` is being passed into an object which itself will be destroyed before `ref` becomes invalid. It also makes sense when `ref` is actually a static value or global that lives forever.
+* `kj::refcounted<T>(args...)` allocates a `T` which uses reference counting. It returns a `kj::Own<T>` that represents one reference to the object. Additional references can be created by calling `kj::addRef(*ptr)`. The object is destroyed when no more `kj::Own`s exist pointing at it. Note that `T` must be a subclass of `kj::Refcounted`. If references may be shared across threads, then atomic refcounting must be used; use `kj::atomicRefcounted<T>(args...)` and inherit `kj::AtomicRefcounted`. Reference counting should be using sparingly; see [KJ idioms around reference counting](style-guide.md#reference-counting) for a discussion of when it should be used and why it is designed the way it is.
+* `kj::attachRef(ref, args...)` returns a `kj::Own` pointing to `ref` that actually owns `args...`, so that when the `kj::Own` goes out-of-scope, the other arguments are destroyed. Typically these arguments are themselves `kj::Own`s or other pass-by-move values that themselves own the object referenced by `ref`. `kj::attachVal(value, args...)` is similar, where `value` is a pass-by-move value rather than a reference; a copy of it will be allocated on the heap. Finally, `ownPtr.attach(args...)` returns a new `kj::Own` pointing to the same value that `ownPtr` pointed to, but such that `args...` are owned as well and will be destroyed together. Attachments are always destroyed after the thing they are attached to.
+* `kj::SpaceFor<T>` contains enough space for a value of type `T`, but does not construct the value until its `construct(args...)` method is called. That method returns an `kj::Own<T>`, whose disposer destroys the value. `kj::SpaceFor` is thus a safer way to perform manual construction compared to invoking `kj::ctor()` and `kj::dtor()`.
+
+These disposers cover most use cases, but you can also implement your own if desired. `kj::Own` features a constructor overload that lets you pass an arbitrary disposer.
+
+### Arrays
+
+`kj::Array<T>` is similar to `kj::Own<T>`, but points to (and owns) an array of `T`s.
+
+A `kj::Array<T>` can be allocated with `kj::heapArray<T>(size)`, if `T` can be default-constructed. Otherwise, you will need to use a `kj::ArrayBuilder<T>` to build the array. First call `kj::heapArrayBuilder<T>(size)`, then invoke the builder's `add(value)` method to add each element, then finally call its `finish()` method to obtain the completed `kj::Array<T>`. `ArrayBuilder` requires that you know the final size before you start; if you don't, you may want to use `kj::Vector<T>` instead.
+
+Passing a `kj::Array<T>` implies an ownership transfer. If you merely want to pass a pointer to an array, without transferring ownership, use `kj::ArrayPtr<T>`. This type essentially encapsulates a pointer to the beginning of the array, plus its size. Note that a `kj::ArrayPtr` points to _the underlying memory_ backing a `kj::Array`, not to the `kj::Array` itself; thus, moving a `kj::Array` does NOT invalidate any `kj::ArrayPtr`s already pointing at it. You can also construct a `kj::ArrayPtr` pointing to any C-style array (doesn't have to be a `kj::Array`) using `kj::arrayPtr(ptr, size)` or `kj::arrayPtr(beginPtr, endPtr)`.
+
+Both `kj::Array` and `kj::ArrayPtr` contain a number of useful methods, like `slice()`. Be sure to check out the class definitions for more details.
+
+## Strings
+
+A `kj::String` is a segment of text. By convention, this text is expected to be Unicode encoded in UTF-8. But, `kj::String` itself is not Unicode-aware; it is merely an array of `char`s.
+
+NUL characters (`'\0'`) are allowed to appear anywhere in a string and do not terminate the string. However, as a convenience, the buffer backing a `kj::String` always has an additional NUL character appended to the end (but not counted in the size). This allows the text in a `kj::String` to be passed to legacy C APIs that use NUL-terminated strings without an extra copy; use the `.cStr()` method to get a `const char*` for such cases. (Of course, keep in mind that if the string contains NUL characters other than at the end, legacy C APIs will interpret the string as truncated at that point.)
+
+`kj::StringPtr` represents a pointer to a `kj::String`. Similar to `kj::ArrayPtr`, `kj::StringPtr` does not point at the `kj::String` object itself, but at its backing array. Thus, moving a `kj::String` does not invalidate any `kj::StringPtr`s. This is a major difference from `std::string`! Moving an `std::string` invalidates all pointers into its backing buffer (including `std::string_view`s), because `std::string` inlines small strings as an optimization. This optimization may seem clever, but means that `std::string` cannot safely be used as a way to hold and transfer ownership of a text buffer. Doing so can lead to subtle, data-dependent bugs; a program might work fine until someone gives it an unusually small input, at which point it segfaults. `kj::String` foregoes this optimization for simplicity.
+
+Also similar to `kj::ArrayPtr`, a `kj::StringPtr` does not have to point at a `kj::String`. It can be initialized from a string literal or any C-style NUL-terminated `const char*` without making a copy. Also, KJ defines the special literal suffix `_kj` to write a string literal whose type is implicitly `kj::StringPtr`.
+
+```c++
+// It's OK to initialize a StringPtr from a classic literal.
+// No copy is performed; the StringPtr points directly at
+// constant memory.
+kj::StringPtr foo = "foo";
+
+// But if you add the _kj suffix, then you don't even need
+// to declare the type. `bar` will implicitly have type
+// kj::StringPtr. Also, this version can be declared
+// `constexpr`.
+constexpr auto bar = "bar"_kj;
+```
+
+### Stringification
+
+To allocate and construct a `kj::String`, use `kj::str(args...)`. Each argument is stringified and the results are concatenated to form the final string. (You can also allocate an uninitialized string buffer with `kj::heapString(size)`.)
+
+```c++
+kj::String makeGreeting(kj::StringPtr name) {
+  return kj::str("Hello, ", name, "!");
+}
+```
+
+KJ knows how to stringify most primitive types as well as many KJ types automatically. Note that integers will be stringified in base 10; if you want hexadecimal, use `kj::hex(i)` as the parameter to `kj::str()`.
+
+You can additionally extend `kj::str()` to work with your own types by declaring a stringification method using `KJ_STRINGIFY`, like so:
+
+```c++
+enum MyType { A, B, C };
+kj::StringPtr KJ_STRINGIFY(MyType value) {
+  switch (value) {
+    case A: return "A"_kj;
+    case B: return "B"_kj;
+    case C: return "C"_kj;
+  }
+  KJ_UNREACHABLE;
+}
+```
+
+The `KJ_STRINGIFY` declaration should appear either in the same namespace where the type is defined, or in the global scope. The function can return any random-access iterable sequence of `char`, such as a `kj::String`, `kj::StringPtr`, `kj::ArrayPtr<char>`, etc. As an alternative to `KJ_STRINGIFY`, you can also declare a `toString()` method on your type, with the same return type semantics.
+
+When constructing very large, complex strings -- for example, when writing a code generator -- consider using `kj::StringTree`, which maintains a tree of strings and only concatenates them at the very end. For example, `kj::strTree(foo, kj::strTree(bar, baz)).flatten()` only performs one concatenation, whereas `kj::str(foo, kj::str(bar, baz))` would perform a redundant intermediate concatenation.
+
+## Core Utility Types
+
+### Maybes
+
+`kj::Maybe<T>` is either `nullptr`, or contains a `T`. In KJ-based code, nullable values should always be expressed using `kj::Maybe`. Primitive pointers should never be null. Use `kj::Maybe<T&>` instead of `T*` to express that the pointer/reference can be null.
+
+In order to dereference a `kj::Maybe`, you must use the `KJ_IF_MAYBE` macro, which behaves like an `if` statement.
+
+```c++
+kj::Maybe<int> maybeI = 123;
+kj::Maybe<int> maybeJ = nullptr;
+
+KJ_IF_MAYBE(i, maybeI) {
+  // This block will execute, with `i` being a
+  // pointer into `maybeI`'s value. In a better world,
+  // `i` would be a reference rather than a pointer,
+  // but we couldn't find a way to trick the compiler
+  // into that.
+  KJ_ASSERT(*i == 123);
+} else {
+  KJ_FAIL_ASSERT("can't get here");
+}
+
+KJ_IF_MAYBE(j, maybeJ) {
+  KJ_FAIL_ASSERT("can't get here");
+} else {
+  // This block will execute.
+}
+```
+
+Note that `KJ_IF_MAYBE` forces you to think about the null case. This differs from `std::optional`, which can be dereferenced using `*`, resulting in undefined behavior if the value is null.
+
+Performance nuts will be interested to know that `kj::Maybe<T&>` and `kj::Maybe<Own<T>>` are both optimized such that they take no more space than their underlying pointer type, using a literal null pointer to indicate nullness. For other types of `T`, `kj::Maybe<T>` must maintain an extra boolean and so is somewhat larger than `T`.
+
+### Variant types
+
+`kj::OneOf<T, U, V>` is a variant type that can be assigned to exactly one of the input types. To unpack the variant, use `KJ_SWITCH_ONEOF`:
+
+```c++
+void handle(kj::OneOf<int, kj::String> value) {
+  KJ_SWITCH_ONEOF(value) {
+    KJ_CASE_ONEOF(i, int) {
+      // Note that `i` is an lvalue reference to the content
+      // of the OneOf. This differs from `KJ_IF_MAYBE` where
+      // the variable is a pointer.
+      handleInt(i);
+    }
+    KJ_CASE_ONEOF(s, kj::String) {
+      handleString(s);
+    }
+  }
+}
+```
+
+Often, in real-world usage, the type of each variant in a `kj::OneOf` is not sufficient to understand its meaning; sometimes two different variants end up having the same type used for different purposes. In these cases, it would be useful to assign a name to each variant. A common way to do this is to define a custom `struct` type for each variant, and then declare the `kj::OneOf` using those:
+
+```c++
+struct NotStarted {
+  kj::String filename;
+};
+struct Running {
+  kj::Own<File> file;
+};
+struct Done {
+  kj::String result;
+};
+
+typedef kj::OneOf<NotStarted, Running, Done> State;
+```
+
+### Functions
+
+`kj::Function<ReturnType(ParamTypes...)>` represents a callable function with the given signature. A `kj::Function` can be initialized from any callable object, such as a lambda, function pointer, or anything with `operator()`. `kj::Function` is useful when you want to write an API that accepts a lambda callback, without defining the API itself as a template. `kj::Function` supports move semantics.
+
+`kj::ConstFunction` is like `kj::Function`, but is used to indicate that the function should be safe to call from multiple threads. (See [KJ idioms around constness and thread-safety](style-guide.md#constness).)
+
+A special optimization type, `kj::FunctionParam`, is like `kj::Function` but designed to be used specifically as the type of a callback parameter to some other function where that callback is only called synchronously; i.e., the callback won't be called anymore after the outer function returns. Unlike `kj::Function`, a `kj::FunctionParam` can be constructed entirely on the stack, with no heap allocation.
+
+### Vectors (growable arrays)
+
+Like `std::vector`, `kj::Vector` is an array that supports appending an element in amortized O(1) time. When the underlying backing array is full, an array of twice the size is allocated and all elements moved.
+
+### Hash/tree maps/sets... and tables
+
+`kj::HashMap`, `kj::HashSet`, `kj::TreeMap`, and `kj::TreeSet` do what you'd expect, with modern lambda-oriented interfaces that are less awkward than the corresponding STL types.
+
+All of these types are actually specific instances of the more-general `kj::Table`. A `kj::Table` can have any number of columns (whereas "sets" have exactly 1 and "maps" have exactly 2), and can maintain indexes on multiple columns at once. Each index can be hash-based, tree-based, or a custom index type that you provide.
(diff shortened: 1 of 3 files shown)


kjdoc/tour.md
```diff
@@ -538,13 +538,441 @@ This section describes KJ APIs that control process execution and low-level inte
...
+* On Unix, disk `kj::Directory` objects are backed by open file descriptors and use the `openat()` family of system calls.
+* Makes it easy to use atomic replacement when writing new files -- and even whole directories.
+* Symlinks, hard links, listing directories, recursive delete, recursive create parents, recursive copy directory, memory mapping, and unnamed temporary files are all exposed and easy to use.
+* Sparse files ("hole punching"), copy-on-write file cloning (`FICLONE`, `FICLONERANGE`), `sendfile()`-based copying, `renameat2()` atomic replacements, and more will automatically be used when available.
+
+See the API reference (header file) for details.
+
+### Clocks and time
+
+KJ provides a time library in `kj/time.h` whicth uses the type system to enforce unit safety.
```
Dr-Emann, line 963: Misspelling of which

kjdoc/tour.md
```diff
@@ -538,13 +538,441 @@ This section describes KJ APIs that control process execution and low-level inte
...
+    kj::throwFatalException(kj::mv(exception));
+  })
+}
+
+kj::Promise<void> boopEvery5SecondsLoop(kj::Timer& timer) {
+  // No memory leaks now!
+  timer.afterDelay(5 * kj::SECONDS).then([&timer]() {
+    KJ_LOG(WARNING, "boop");
+    // Loop by recursing.
+    return logEvery5Seconds(timer);
```
harrishancock, line 849:
```suggestion
    return boopEvery5SecondsLoop(timer);
```

kjdoc/tour.md
```diff
@@ -538,13 +538,441 @@ This section describes KJ APIs that control process execution and low-level inte
...
+      .catch_([](kj::Exception&& exception) {
+    // Oh no, an error! Log it and end the loop.
+    KJ_LOG(ERROR, exception);
+    kj::throwFatalException(kj::mv(exception));
+  })
+}
+
+kj::Promise<void> boopEvery5SecondsLoop(kj::Timer& timer) {
+  // No memory leaks now!
+  timer.afterDelay(5 * kj::SECONDS).then([&timer]() {
```
harrishancock, line 846:
```suggestion
  return timer.afterDelay(5 * kj::SECONDS).then([&timer]() {
```

kjdoc/tour.md
```diff
@@ -538,13 +538,441 @@ This section describes KJ APIs that control process execution and low-level inte
...
+  });
+}
+```
+
+Another possible fix would be to make sure the recursive continuation and then error handler are passed to the same `.then()` invocation:
+
+```c++
+kj::Promise<void> boopEvery5Seconds(kj::Timer& timer) {
+  // WARNING! MEMORY LEAK!
+  timer.afterDelay(5 * kj::SECONDS).then([&timer]() {
```
harrishancock, line 859:
```suggestion
  return timer.afterDelay(5 * kj::SECONDS).then([&timer]() {
```

kjdoc/tour.md
```diff
@@ -538,13 +538,441 @@ This section describes KJ APIs that control process execution and low-level inte
...
+    return logEvery5Seconds(timer);
+  });
+}
+```
+
+Another possible fix would be to make sure the recursive continuation and then error handler are passed to the same `.then()` invocation:
+
+```c++
+kj::Promise<void> boopEvery5Seconds(kj::Timer& timer) {
+  // WARNING! MEMORY LEAK!
```
harrishancock, line 858: Vestigial comment?
```suggestion
  // No more memory leaks, but hard to reason about.
```

kjdoc/tour.md
```diff
@@ -538,13 +538,441 @@ This section describes KJ APIs that control process execution and low-level inte
...
+    return boopEvery5Seconds(timer);
+  }).catch_([](kj::Exception&& exception) {
+    // Oh no, an error! Log it and end the loop.
+    KJ_LOG(ERROR, exception);
+    kj::throwFatalException(kj::mv(exception));
+  });
+}
+```
+
+The problem in this example is that the recursive call is _not_ a tail call, due to the `.catch_()` appended to the end. Every time around the loop, a new `.catch_()` is added to the promise chain. If an exception were thrown, that exception would end up being logged many times -- once for each time the loop has repeated so far.
```
harrishancock, line 830: I think it'd be good to drive the memory leak point home here. Would this scenario also risk a stack overflow on destruction?

kjdoc/tour.md
```diff
@@ -538,13 +538,441 @@ This section describes KJ APIs that control process execution and low-level inte
...
+    // Oh no, an error! Log it and end the loop.
+    KJ_LOG(ERROR, exception);
+    kj::throwFatalException(kj::mv(exception));
+  })
+}
+
+kj::Promise<void> boopEvery5SecondsLoop(kj::Timer& timer) {
+  // No memory leaks now!
+  timer.afterDelay(5 * kj::SECONDS).then([&timer]() {
+    KJ_LOG(WARNING, "boop");
```
harrishancock, line 847: For symmetry with the other snippets.
```suggestion
    boop();
```

kjdoc/tour.md
```diff
@@ -538,13 +538,441 @@ This section describes KJ APIs that control process execution and low-level inte
...
+kj::Promise<int> promise = ...;
+kj::ForkedPromise<int> forked = promise.fork();
+kj::Promise<int> branch1 = promise.addBranch();
+kj::Promise<int> branch2 = promise.addBranch();
+kj::Promise<int> branch3 = promise.addBranch();
+```
+
+A forked promise can have any number of "branches" which represent different consumers waiting for the same result.
+
+Forked promises use reference counting. The `ForkedPromise` itself, and each branch created from it, each represents a reference to the original promise. The original promise will only be canceled if all branches are canceled and the `ForkedPromise` itself is destroyed.
```
harrishancock, line 888: Nit: grammar -- I think _The `ForkedPromise` itself, and each branch created from it, each represents ..._ has a plural subject, despite the coordinate making it plural also being a parenthetical expression ("and each branch created from it"). Honestly, though, it seems like it's a judgment call. See this: https://english.stackexchange.com/questions/56210/singular-or-plural-verb-form-where-subject-includes-a-parenthetical-element

My brain can't come up with a better way to word it; feel free to ignore.

kjdoc/tour.md
```diff
@@ -538,13 +538,441 @@ This section describes KJ APIs that control process execution and low-level inte
...
+Sometimes, the copyable requirement of `.fork()` can be burdensome and unnecessary. If the result type has multiple components, and each branch really only needs one of the components, then being able to copy (or refcount) is unnecessary. In these cases, you can use `.split()` instead. `.split()` converts a promise for a `kj::Tuple` into a `kj::Tuple` of promises. That is:
+
+```c++
+kj::Promise<kj::Tuple<kj::Own<Foo>, kj::String>> promise = ...;
+kj::Tuple<kj::Promise<kj::Own<Foo>>, kj::Promise<kj::String>> promises = promise.split();
+```
+
+### Threads
+
+The KJ async framework is designed around single-threaded event loops. However, you can multiple threads, with each running its own loop.
```
harrishancock, line 901: Nit: can haz
```suggestion
The KJ async framework is designed around single-threaded event loops. However, you can have multiple threads, with each running its own loop.
```

kjdoc/tour.md
```diff
@@ -538,13 +538,441 @@ This section describes KJ APIs that control process execution and low-level inte
...
+
+All KJ async objects, unless specifically documented otherwise, are intrinsically tied to the thread and event loop on which they were created. These objects must not be accessed from any other thread.
+
+To communicate between threads, you may use `kj::Executor`. Each thread (that has an event loop) may call `kj::getCurrentThreadExecutor()` to get a reference to its own `Executor`. That reference may then be shared with other threads. The other threads can use the methods of `Executor` to queue functions to execute on the owning thread's event loop.
```
harrishancock, line 905: Maybe clarify that it's a "strong" reference (`kj::Own<T>`), rather than a literal C++ reference.

kjdoc/tour.md
```diff
@@ -538,13 +538,441 @@ This section describes KJ APIs that control process execution and low-level inte
...
+
+The threads which call an `Executor` do not have to have KJ event loops themselves. Thus, you can use an `Executor` to signal a KJ event loop thread from a non-KJ thread.
```
harrishancock, line 907: Only if you use `executeSync()`, right? I thought `executeAsync()` required a local event loop.

kjdoc/tour.md
```diff
@@ -538,13 +538,441 @@ This section describes KJ APIs that control process execution and low-level inte
...
+
+Because of this, fibers should not be used just to make code look nice (C++20's `co_await`, which KJ will soon support, is a better way to do that). Instead, the main use case for fibers is to be able to call into existing libraries that are not designed to operate in an asynchronous way. For example, say you find a library that performs stream I/O, and lets you provide your own `read()`/`write()` implementations, but expects those implementations to operate in a blocking fashion. With fibers, you can use such a library within the asynchronous KJ event loop.
```
harrishancock, line 925:
```suggestion
Because of this, fibers should not be used just to make code look nice (C++20's `co_await`, which KJ will soon support, is a better way to do that.) Instead, the main use case for fibers is to be able to call into existing libraries that are not designed to operate in an asynchronous way. For example, say you find a library that performs stream I/O, and lets you provide your own `read()`/`write()` implementations, but expects those implementations to operate in a blocking fashion. With fibers, you can use such a library within the asynchronous KJ event loop.
```

kjdoc/tour.md
```diff
@@ -538,13 +538,441 @@ This section describes KJ APIs that control process execution and low-level inte
...
 
 ## System I/O
 
-TODO: async-io, etc.
+TODO: OS handles, sync IO, async IO, filesystem, network, time
```
harrishancock, line 929: Looks like you wrote some of these sections. Do you plan to expand them, or can this line be removed? Or maybe just network is left.
```suggestion
TODO: network
```

kjdoc/tour.md
```diff
@@ -538,13 +538,441 @@ This section describes KJ APIs that control process execution and low-level inte
...
+`kj/io.h` provides some more basic, synchronous streaming interfaces, like `kj::InputStream` and `kj::OutputStream`. Implementations are provided on top of file descriptors and Windows `HANDLE`s.
+
+Additionally, the important utility class `kj::AutoCloseFd` (and `kj::AutoCloseHandle` for Windows) can be found here. This is an RAII wrapper around a file descriptor (or `HANDLE`), which you will likely want to use any time you are manipulating raw file descriptors (or `HANDLE`s) in KJ code.
+
+### Filesystem
+
+KJ provides an advanced, cross-platform filesystem API in `kj/filesystem.h`. Features include:
+
+* Paths represented using `kj::Path`. In addition to providing common-sense path parsing and manipulation functions, this class is designed to defend against path injection attacks.
+* All interfaces are abstract, allowing mulitple implementations.
```
harrishancock, line 952:
```suggestion
* All interfaces are abstract, allowing multiple implementations.
```

kjdoc/tour.md
harrishancock:
```suggestion
KJ provides a time library in `kj/time.h` which uses the type system to enforce unit safety.
```

@Dr-Emann you can use the button to the left of the style buttons to add a suggestion, which makes it super easy for the PR author to incorporate the change. It's the button highlighted in blue in this screenshot:

Screenshot from 2021-01-22 09-50-34

kjdoc/tour.md
```diff
@@ -538,13 +538,441 @@ This section describes KJ APIs that control process execution and low-level inte
...
+
+### Clocks and time
+
+KJ provides a time library in `kj/time.h` whicth uses the type system to enforce unit safety.
+
+`kj::Duration` represents a length of time, such as a number of seconds. Multiply an integer by `kj::SECONDS`, `kj::MINUTES`, `kj::NANOSECONDS`, etc. to get a `kj::Duration` value. Divide by the appropriate constant to get an integer.
+
+`kj::Date` represents a point in time in the real world. `kj::UNIX_EPOCH` represents January 1st, 1970, 00:00 UTC. Other dates can be constructed by adding a `kj::Duraction` to `kj::UNIX_EPOCH`. Taking the difference between to `kj::Date`s produces a `kj::Duration`.
+
+`kj::TimePoint` represens a time point measured against an unspecified origin time. This is typically used with monotonic clocks that don't necessarily reflect calendar time. Unlike `kj::Date`, there is no implicit guarantee that two `kj::TimePoint`s are measured against the same origin and are therefore comparable; it is up to the application to track which clock any particluar `kj::TimePoint` came from.
```
harrishancock, line 969:
```suggestion
`kj::TimePoint` represents a time point measured against an unspecified origin time. This is typically used with monotonic clocks that don't necessarily reflect calendar time. Unlike `kj::Date`, there is no implicit guarantee that two `kj::TimePoint`s are measured against the same origin and are therefore comparable; it is up to the application to track which clock any particluar `kj::TimePoint` came from.
```

kjdoc/tour.md
```diff
@@ -538,13 +538,441 @@ This section describes KJ APIs that control process execution and low-level inte
...
+
+On top of KJ's async framework (described earlier), KJ provides asynchronous APIs for byte streams, networking, and timers.
+
+As mentioned previously, `kj::setupAsyncIo()` allocates an appropriate OS-specific event queue (such as `epoll` on Linux), returning implementations of `kj::AsyncIoProvider` and `kj::LowLevelAsyncIoProvider` implemented in terms of that queue. `kj::AsyncIoProvider` provides an OS-independent API for byte streams, networking, and timers. `kj::LowLevelAsyncIoProvider` allows native OS handles (file descriptors on Unix, `HANDLE`s on Windows) to be wrapped in KJ byte stream APIs, like `kj::AsyncIoStream`.
+
+Please refer to the API reference (the header files) for details on these APIs.
+
+### Synchronous I/O
+
+Although most complex KJ applications use Async I/O, sometimes you want something a little simpler.
```
harrishancock, line 941: Nit: you lowercase "async" elsewhere
```suggestion
Although most complex KJ applications use async I/O, sometimes you want something a little simpler.
```

harrishancock (approved):
This is great! I left some comments of varying value, feel free to pick and choose. :)

kjdoc/tour.md
```diff
@@ -538,13 +538,441 @@ This section describes KJ APIs that control process execution and low-level inte
...
+    boop();
+    // Loop by recursing.
+    return boopEvery5Seconds(timer);
+  });
+}
+```
+
+KJ promises include "tail call optimization" for loops like the one above, so that the promise chain length remains finite no matter how many times the loop iterates.
+
+**WARNING!** It is very easy to accidentally break tail call optimization, creating a memory leak. Consider the following:
```
Dr-Emann, line 813: I'm not sure if this is the best place to discuss this, but just to clarify, does this also work with non-directly recursive functions, like a loop between `fn_a -> fn_b -> fn_c -> fn_a`, all of which return different promises? Is the cause of the memory leak only due to the buildup of catches (either from `_catch()`, or `.then(..., catcher)`, so I can be sure I'm safe if the recursion has none, or are there other cases where tail call optimization can be broken?

kentonv: Yes, like classical "tail call optimization", the optimization works any time the last thing you do is `return someFunction()`. It doesn't matter whether `someFunction()` is actually a recursive call.

To put it another way, when a continuation passed to `.then()` itself returns a promise, the new promise node _replaces_ the `.then()` in the chain, rather than adding on to it.

But if you're adding another `.catch_()` or `then()` after the one that made the "tail call", that one can't be eliminated from the chain, and so the chain will become longer with each iteration.

kjdoc/tour.md
```diff
@@ -538,13 +538,441 @@ This section describes KJ APIs that control process execution and low-level inte
...
+});
+```
+
+As with `.then()` continuations, the lambda passed to these functions may itself return a `Promise`.
+
+`kj::evalNow()` executes the lambda immediately -- before `evalNow()` even returns. The purpose of `evalNow()` is to catch any exceptions thrown and turn them into a rejected promise. This is often a good idea when you don't want the caller to have to handle both synchronous and asynchronous exceptions -- wrapping your whole function in `kj::evalNow()` ensures that all exceptions are delivered asynchronously.
+
+`kj::evalLater()` executes the lambda on a future turn of the event loop. This is equivalent to `kj::Promise<void>().then()`.
+
+`kj::evalLast()` arranges for the lambda to be called only once all other work queued to the event loop has completed (but before querying the OS for new I/O events). This can often be useful, for example, for batching. For example, if a program tends to make many small write()s to a socket in rapid succession, you might want to add a layer that collects the writes into a batch, then sends the whole batch in a single write from an `evalLast()`. This way, none of the bytes are significantly delayed, but they can still be coalesced.
```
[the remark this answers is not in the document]
kentonv: Torn on this one. It does sound sort of weird, but the described example is just one specific example of "batching".

kjdoc/tour.md
kentonv: > Would this scenario also risk a stack overflow on destruction?

Yup.

kjdoc/tour.md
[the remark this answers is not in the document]
harrishancock: Yeah, both instances of the phrase do have a purpose. Could change the first instance to `e.g.`, but maybe it's better to just leave as-is.

kjdoc/tour.md
kentonv: I actually did mean C++ reference. Using a strong reference is not strictly required if you have some other way to guarantee the event loop hasn't exited. I figured people will see `addRef()` when they look at the API reference.

kjdoc/tour.md
kentonv: That's correct, but I haven't introduced the API details here at all, people can read the header to learn the details.

FWIW when we generate a web site out of this, I intend for all of the class name references to be links to header listings.

kjdoc/tour.md
kentonv: Hmm, really? I know that it's said that when a sentence ends with a quoted phrase, the period is supposed to be inside the quote. As a programmer, I can't deal with that rule and usually ignore it. But I thought that when a sentence ends with a parenthetical, it's normal for the period to be outside the parenthetical?

Or are you suggesting that the parenthetical should be separated from the preceding sentence entirely? In that case I suppose we need to add a period before the parenthetical?

kjdoc/tour.md
kentonv:
```suggestion
```

I actually did mention networking under Async I/O, so this TODO is now complete.

kjdoc/tour.md
harrishancock: The answers to all four of your questions is: I made a mistake. :(

kentonv: Haha well you're the linguist so I trusted you!

kjdoc/tour.md
harrishancock: Ah, got it. I'd forgotten that function actually did return a reference.

kjdoc/tour.md
harrishancock: Works for me.

kjdoc/tour.md
```diff
@@ -538,13 +538,440 @@ This section describes KJ APIs that control process execution and low-level inte
...
+    networkAddress.connect();
+kj::Promise<kj::String> textPromise = connectPromise
+    .then([](kj::Own<kj::AsyncIoStream> stream) {
+  // We must attach the stream so that it remains alive until `readAllText()`
+  // is done. The stream will then be discarded.
+  return stream->readAllText().attach(kj::mv(stream));
+});
+```
+
+Using `.attach()` is semantically equivalent to using `.then()`, passing an identify function as the continuation, while having that function capture ownership of the attached object, i.e.:
```
jclee, line 700:
```suggestion
Using `.attach()` is semantically equivalent to using `.then()`, passing an identity function as the continuation, while having that function capture ownership of the attached object, i.e.:
```

kjdoc/tour.md
```diff
@@ -538,13 +538,440 @@ This section describes KJ APIs that control process execution and low-level inte
 
 KJ makes asynchronous programming manageable using an API modeled on E-style Promises. E-style Promises were also the inspiration for JavaScript Promises, so modern JavaScript programmers should find KJ Promises familiar, although there are some important differences.
 
-A `kj::Promise<T>` represents an asynchronous background task that, upon completion, will either "resolve" to a value of type `T`, or "reject" with an exception.
+A `kj::Promise<T>` represents an asynchronous background task that, upon completion, either "resolves" to a value of type `T`, or "rejects" with an exception.
```
jclee, line 541: I don't know if it might be worth calling out the third possibility of promise cancellation here?

kjdoc/tour.md
```diff
@@ -538,13 +538,440 @@ This section describes KJ APIs that control process execution and low-level inte
...
+
+In order to implement all this, KJ will set up the appropriate OS-specific constructs to handle I/O events on the host platform. For example, on Linux, KJ will use `epoll`, whereas on Windows, it will set up an I/O Completion Port.
+
+Sometimes, you may need KJ promises to cooperate with some existing event loop, rather than set up its own. For example, you might be using libuv, or Boost.Asio. Usually, a thread can only have one event loop, because it can only wait on one OS event queue (e.g. `epoll`) at a time. To accommodate this, it is possible (though not easy) to adapt KJ to run on top of some other event loop, by creating a custom implementation of `kj::EventPort`. The details of how to do this are beyond the scope of this document.
+
+Sometimes, you may find that you don't really need to perform operating system I/O at all. For example, a unit test might only need to call some asynchronous functions using mock I/O interfaces, and a thread in a multi-threaded program may only need to exchange events with other threads and not the OS. In these cases, you can create a simple event loop instead:
```
jclee, line 586:
```suggestion
Sometimes, you may find that you don't really need to perform operating system I/O at all. For example, a unit test might only need to call some asynchronous functions using mock I/O interfaces, or a thread in a multi-threaded program may only need to exchange events with other threads and not the OS. In these cases, you can create a simple event loop instead:
```

kjdoc/tour.md
```diff
@@ -538,13 +538,440 @@ This section describes KJ APIs that control process execution and low-level inte
...
+timer.wait(waitScope);  // returns after 5 seconds' delay
+```
+
+`promise.wait()` will run the thread's event loop until the promise completes. It will then return the `Promise`'s result (or throw the `Promise`'s exception). `.wait()` consumes the `Promise`, as if the `Promise` has been moved away.
+
+Synchronous waits cannot be nested -- i.e. a `.then()` callback (see below) that is called by the event loop itself cannot execute another level of synchronous waits. Hence, synchronous waits generally can only be used at the top level of the thread. The API requires passing a `kj::WaitScope` to `.wait()` as a way to demonstrate statically that the caller is allowed to perform synchronous waits. Any function which wishes to perform synchronous waits must take a `kj::WaitScope&` as a parameter to indicate that it does this.
+
+Synchronous waits often make sense to use in "client" programs that only have one task to complete before they exit. On the other end of the spectrum, server programs that handle many clients generally must do everything asynchronously. At the top level of a server program, you will typically instruct the event loop to run forever like so:
```
jclee, line 607:
```suggestion
Synchronous waits often make sense to use in "client" programs that only have one task to complete before they exit. On the other end of the spectrum, server programs that handle many clients generally must do everything asynchronously. At the top level of a server program, you will typically instruct the event loop to run forever, like so:
```

kjdoc/tour.md
kentonv: Eh, I think that's better introduced separately.

kjdoc/tour.md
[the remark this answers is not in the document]
kentonv:
```suggestion
`kj::evalLast()` arranges for the lambda to be called only once all other work queued to the event loop has completed (but before querying the OS for new I/O events). This can often be useful e.g. for batching. For example, if a program tends to make many small write()s to a socket in rapid succession, you might want to add a layer that collects the writes into a batch, then sends the whole batch in a single write from an `evalLast()`. This way, none of the bytes are significantly delayed, but they can still be coalesced.
```

kjdoc/tour.md
```diff
@@ -538,13 +538,440 @@ This section describes KJ APIs that control process execution and low-level inte
...
+});
+```
+
+As with `.then()` continuations, the lambda passed to these functions may itself return a `Promise`.
+
+`kj::evalNow()` executes the lambda immediately -- before `evalNow()` even returns. The purpose of `evalNow()` is to catch any exceptions thrown and turn them into a rejected promise. This is often a good idea when you don't want the caller to have to handle both synchronous and asynchronous exceptions -- wrapping your whole function in `kj::evalNow()` ensures that all exceptions are delivered asynchronously.
+
+`kj::evalLater()` executes the lambda on a future turn of the event loop. This is equivalent to `kj::Promise<void>().then()`.
+
+`kj::evalLast()` arranges for the lambda to be called only once all other work queued to the event loop has completed (but before querying the OS for new I/O events). This can often be useful, for example, for batching. For example, if a program tends to make many small write()s to a socket in rapid succession, you might want to add a layer that collects the writes into a batch, then sends the whole batch in a single write from an `evalLast()`. This way, none of the bytes are significantly delayed, but they can still be coalesced.
```
jclee:
```suggestion
`kj::evalLast()` arranges for the lambda to be called only after all other work queued to the event loop has completed (but before querying the OS for new I/O events). This can often be useful, for example, for batching. For example, if a program tends to make many small write()s to a socket in rapid succession, you might want to add a layer that collects the writes into a batch, then sends the whole batch in a single write from an `evalLast()`. This way, none of the bytes are significantly delayed, but they can still be coalesced.
```
Might be a little clearer? At first glance, I incorrectly saw "`kj::evalLast()` arranges for the lambda to be called only once" as its own clause.

kjdoc/tour.md
```diff
@@ -538,13 +538,440 @@ This section describes KJ APIs that control process execution and low-level inte
...
+
+If multiple `evalLast()`s exist at the same time, they will execute in last-in-first-out order. If the first one out schedules more work on the event loop, that work will be completed before the next `evalLast()` executes, and so on.
+
+### Attachments
+
+Often, a task represented by a `Promise` will require that some object remains live until the `Promise` completes. In particular, under KJ conventions, unless documented otherwise, any class method which returns a `Promise` inherently expects that the caller will ensure that the object it was called on will remain live until the `Promise` completes (or is canceled). Put another way, member function implementations may assume their `this` pointer is valid as long as their returned `Promise` is alive.
```
jclee, line 685:
```suggestion
Often, a task represented by a `Promise` will require that some object remains alive until the `Promise` completes. In particular, under KJ conventions, unless documented otherwise, any class method which returns a `Promise` inherently expects that the caller will ensure that the object it was called on will remain alive until the `Promise` completes (or is canceled). Put another way, member function implementations may assume their `this` pointer is valid as long as their returned `Promise` is alive.
```

kjdoc/tour.md
```diff
@@ -538,13 +538,440 @@ This section describes KJ APIs that control process execution and low-level inte
...
+  return stream->readAllText().attach(kj::mv(stream));
+});
+```
+
+Using `.attach()` is semantically equivalent to using `.then()`, passing an identify function as the continuation, while having that function capture ownership of the attached object, i.e.:
+
+```c++
+// This...
+promise.attach(kj::mv(attachment));
+// ... is equivalent to this...
```
jclee, line 705:
```suggestion
// ...is equivalent to this...
```

kjdoc/tour.md
```diff
@@ -538,13 +538,440 @@ This section describes KJ APIs that control process execution and low-level inte
...
+```c++
+promise = promise.attach(kj::defer([]() {
+  // This code will execute when the promise completes or is canceled.
+}));
+```
+
+
+### Background tasks
+
+If you construct a `Promise` and then just leave it be without calling `.then()` or `.wait()` to consume it, the task it represents will execute in the background. You can call `.then()` or `.wait()` later on, when you're ready. This makes it possible to run multiple concurrent tasks at once.
```
jclee, line 720: Might be worth clarifying what "execute in the background" means here -- particularly that the work gets done as part of event loop processing, not in some background thread?

kjdoc/tour.md
```diff
@@ -538,13 +538,440 @@ This section describes KJ APIs that control process execution and low-level inte
...
+
+```c++
+kj::Promise<void> timer.afterDelay(5 * kj::SECONDS)
+    .then([]() {
+  // This log will correctly be written after 5 seconds.
+  KJ_LOG(WARNING, "It has been 5 seconds!!!");
+}).eagerlyEvaluate([](kj::Exception&& exception) {
+  KJ_LOG(ERROR, exception);
+});
+kj::NEVER_DONE.wait(waitScope);
```
jclee, line 744: I suppose it might be possible for someone to skim these examples and conclude that `.eagerlyEvaluate()` somehow makes the promise run synchronously.  Not sure if there's an easy way to dispel this.

kjdoc/tour.md
```diff
@@ -538,13 +538,440 @@ This section describes KJ APIs that control process execution and low-level inte
...
+```
+
+`.eagerlyEvaluate()` takes an error handler callback as its parameter, with the same semantics as `.catch_()` or the second parameter to `.then()`. This is required because otherwise, it is very easy to forget to install an error handler on background tasks, resulting in errors being silently discarded. However, if you are certain that errors will be properly handled elsewhere, you may pass `nullptr` as the parameter to skip error checking -- this is equivalent to passing a callback that merely re-throws the exception.
+
+### Cancellation
+
+If you destroy a `Promise` before it has completed, any incomplete work will be immediately canceled.
```
jclee, line 751: Don't know if it might be worth pointing out, somewhere in this section, that one can cancel a promise in a variable prior to running its destructor, by assigning `nullptr` to it.

kjdoc/tour.md
```diff
@@ -538,13 +538,440 @@ This section describes KJ APIs that control process execution and low-level inte
...
+
+For these situations, you may use a `Promise`-`Fulfiller` pair.
+
+```c++
+kj::PromiseFulfillerPair<int> paf = kj::newPromiseAndFulfiller<int>();
+
+// Consumer waits for the promise.
+paf.promise.then([](int i) { ... });
+
+// Producer calls the fulfiller to fulfill he promise.
```
jclee, line 779:
```suggestion
// Producer calls the fulfiller to fulfill the promise.
```

kjdoc/tour.md
```diff
@@ -538,13 +538,440 @@ This section describes KJ APIs that control process execution and low-level inte
...
+  // No memory leaks now!
+  return timer.afterDelay(5 * kj::SECONDS).then([&timer]() {
+    boop();
+    // Loop by recursing.
+    return boopEvery5SecondsLoop(timer);
+  });
+}
+```
+
+Another possible fix would be to make sure the recursive continuation and then error handler are passed to the same `.then()` invocation:
```
jclee, line 854:
```suggestion
Another possible fix would be to make sure the recursive continuation and the error handler are passed to the same `.then()` invocation:
```

kjdoc/tour.md
```diff
@@ -538,13 +538,440 @@ This section describes KJ APIs that control process execution and low-level inte
...
+    kj::throwFatalException(kj::mv(exception));
+  });
+}
+```
+
+Notice that in this second case, the error handler is scoped so that it does _not_ catch exceptions thrown by the recursive call; it only catches exceptions from `boop()`. This solves the problem, but it's a bit trickier to understand and to ensure that exceptions can't accidentally slip past the error handler.
+
+### Forking and splitting promises
+
+As mentioned above, `.then()` and similar functions consume the promise on which they are called, so can only be called once. But, what if you want to start multiple tasks using the result of a promise? You could solve this in a convoluted way using adapted promises, but KJ has a built-in solution: `.fork()`
```
jclee, line 876:
```suggestion
As mentioned above, `.then()` and similar functions consume the promise on which they are called, so they can only be called once. But, what if you want to start multiple tasks using the result of a promise? You could solve this in a convoluted way using adapted promises, but KJ has a built-in solution: `.fork()`
```

kjdoc/tour.md
```diff
@@ -538,13 +538,440 @@ This section describes KJ APIs that control process execution and low-level inte
...
+constexpr size_t STACK_SIZE = 65536;
+kj::Promise<int> promise =
+    kj::startFiber(STACK_SIZE, [](kj::WaitScope& waitScope) {
+  int i = someAsyncFunc().wait(waitScope);
+  i += anotherAsyncFunc().wait(waitScope);
+  return i;
+});
+```
+
+**CAUTION:** Fibers produce attractive-looking code, but have serious drawbacks. Every fiber must allocate a new call stack, which is typically rather large. The above example allocates a 64kb stack, which is the _minimum_ supported size. Some programs and libraries expect to be able to allocate megabytes of data on the stack. On modern Linux systems, a default stack size of 8MB is typical. Stacks space is allocated lazily on page faults, but just setting up the memory mapping is much more expensive than a typical `malloc()`. If you create lots of fibers, you should use `kj::FiberPool` to reduce allocation costs -- but while this reduces allocation overhead, it will increase memory usage.
```
jclee, line 923:
```suggestion
**CAUTION:** Fibers produce attractive-looking code, but have serious drawbacks. Every fiber must allocate a new call stack, which is typically rather large. The above example allocates a 64kb stack, which is the _minimum_ supported size. Some programs and libraries expect to be able to allocate megabytes of data on the stack. On modern Linux systems, a default stack size of 8MB is typical. Stack space is allocated lazily on page faults, but just setting up the memory mapping is much more expensive than a typical `malloc()`. If you create lots of fibers, you should use `kj::FiberPool` to reduce allocation costs -- but while this reduces allocation overhead, it will increase memory usage.
```

kjdoc/tour.md
```diff
@@ -538,13 +538,440 @@ This section describes KJ APIs that control process execution and low-level inte
...
+
+See the API reference (header file) for details.
+
+### Clocks and time
+
+KJ provides a time library in `kj/time.h` which uses the type system to enforce unit safety.
+
+`kj::Duration` represents a length of time, such as a number of seconds. Multiply an integer by `kj::SECONDS`, `kj::MINUTES`, `kj::NANOSECONDS`, etc. to get a `kj::Duration` value. Divide by the appropriate constant to get an integer.
+
+`kj::Date` represents a point in time in the real world. `kj::UNIX_EPOCH` represents January 1st, 1970, 00:00 UTC. Other dates can be constructed by adding a `kj::Duraction` to `kj::UNIX_EPOCH`. Taking the difference between to `kj::Date`s produces a `kj::Duration`.
```
jclee, line 966:
```suggestion
`kj::Date` represents a point in time in the real world. `kj::UNIX_EPOCH` represents January 1st, 1970, 00:00 UTC. Other dates can be constructed by adding a `kj::Duration` to `kj::UNIX_EPOCH`. Taking the difference between to `kj::Date`s produces a `kj::Duration`.
```

kjdoc/tour.md
```diff
@@ -538,13 +538,440 @@ This section describes KJ APIs that control process execution and low-level inte
...
+
+`kj::TimePoint` represents a time point measured against an unspecified origin time. This is typically used with monotonic clocks that don't necessarily reflect calendar time. Unlike `kj::Date`, there is no implicit guarantee that two `kj::TimePoint`s are measured against the same origin and are therefore comparable; it is up to the application to track which clock any particluar `kj::TimePoint` came from.
```
jclee, line 968:
```suggestion
`kj::TimePoint` represents a time point measured against an unspecified origin time. This is typically used with monotonic clocks that don't necessarily reflect calendar time. Unlike `kj::Date`, there is no implicit guarantee that two `kj::TimePoint`s are measured against the same origin and are therefore comparable; it is up to the application to track which clock any particular `kj::TimePoint` came from.
```

kjdoc/tour.md
```diff
@@ -538,13 +538,440 @@ This section describes KJ APIs that control process execution and low-level inte
...
+```
+
+### Synchronous waits
+
+In the top level of your program (or thread), the program is allowed to synchronously wait on a promise using the `kj::WaitScope` (see above).
+
+```
+kj::Timer& timer = io.provider->getTimer();
+kj::Promise<void> promise = timer.afterDelay(5 * kj::SECONDS);
+timer.wait(waitScope);  // returns after 5 seconds' delay
```
a-robinson, line 600:
```suggestion
promise.wait(waitScope);  // returns after 5 seconds' delay
```

kjdoc/tour.md
```diff
@@ -538,13 +538,440 @@ This section describes KJ APIs that control process execution and low-level inte
...
+// Producer calls the fulfiller to fulfill he promise.
+paf.fulfiller->fulfill(123);
+
+// Producer can also reject the promise.
+paf.fulfiller->reject(KJ_EXCEPTION(FAILED, "something went wrong"));
+```
+
+**WARNING! DANGER!** When using promise-fulfiller pairs, it is very easy to forget about both exception propagation and, more importantly, cancellation-safety.
+
+* **Exception-safety:** If your code stops early due to an exception, it may forget to invoke the fulfiller. Upon destroying the fulfiller, the consumer end will receive a generic, unhelpful exception, merely saying that the fulfiller was destroyed unfulfilled. To aid in debugging, you should make sure to catch exceptions and call `reject()` to propagate them.
```
a-robinson, line 788:
```suggestion
* **Exception-safety:** If your code stops early due to an exception, it may forget to invoke the fulfiller. Upon destroying the fulfiller, the consumer end will receive a generic, unhelpful exception, merely saying that the fulfiller was destroyed unfulfilled. To aid in debugging, you should make sure to catch exceptions and call `fulfiller->reject()` to propagate them.
```

kjdoc/tour.md
```diff
@@ -538,13 +538,440 @@ This section describes KJ APIs that control process execution and low-level inte
...
+Forked promises require that the result type has a copy constructor, so that it can be copied to each branch. (Regular promises only require the result type to be movable, not copyable.) Or, alternatively, if the result type is `kj::Own<T>` -- which is never copyable -- then `T` must have a method `kj::Own<T> T::addRef()`; this method will be invoked to create each branch. Typically, `addRef()` would be implemented using reference counting.
+
+Sometimes, the copyable requirement of `.fork()` can be burdensome and unnecessary. If the result type has multiple components, and each branch really only needs one of the components, then being able to copy (or refcount) is unnecessary. In these cases, you can use `.split()` instead. `.split()` converts a promise for a `kj::Tuple` into a `kj::Tuple` of promises. That is:
+
+```c++
+kj::Promise<kj::Tuple<kj::Own<Foo>, kj::String>> promise = ...;
+kj::Tuple<kj::Promise<kj::Own<Foo>>, kj::Promise<kj::String>> promises = promise.split();
+```
+
+### Threads
```
a-robinson, line 899: I don't think this doc needs to go into any detail on it since the header is pretty concise, but it seems worth mentioning that `kj::Thread` provides a primitive for creating a new thread.

a-robinson (approved):
This will be incredibly helpful to anyone starting with kj 👍

Sarcasm:
Nice to see this overview of KJ async patterns. :+1:
I had to discover some of them scrutinizing kj/capnproto/sandstorm.

Forking and splitting promises is described but not joining isn't.
I think it makes sense to document it for completeness.
I use it regularly in my application by means of `kj::Vector` + `joinPromises()`.

kjdoc/tour.md
```diff
@@ -538,13 +538,440 @@ This section describes KJ APIs that control process execution and low-level inte
...
+
+
+### Background tasks
+
+If you construct a `Promise` and then just leave it be without calling `.then()` or `.wait()` to consume it, the task it represents will execute in the background. You can call `.then()` or `.wait()` later on, when you're ready. This makes it possible to run multiple concurrent tasks at once.
+
+Note that, when possible, KJ evaluates continuations lazily. Continuations which merely transform the result (without returning a new `Promise` that might require more waiting) are only evaluated when the final result is actually needed. This is an optimization which allows a long chain of `.then()`s to be executed all at once, rather than turning the event loop for each one. However, it can lead to some confusion when storing an unconsumed `Promise`. For example:
+
+```c++
+kj::Promise<void> timer.afterDelay(5 * kj::SECONDS)
```
harrishancock, line 725:
```suggestion
kj::Promise<void> promise = timer.afterDelay(5 * kj::SECONDS)
```

kjdoc/tour.md
kentonv: Hmm, I sort of like "live" better here, but don't have any particularly good argument.

bretthoerner (approved):
This is awesome. The big warnings sections are especially helpful, I imagine I'll come back to this often.

kjdoc/tour.md
jclee: Not a big deal... I was thinking to myself that "alive" was a little better because:
- "live" is more ambiguous, in that it can mean either "alive" or "in active use"
- the paragraph already also uses "alive" to refer to the promise, which might lead someone to question the distinction.

kentonv:
Note to self: Add documentation of `promise.poll()` for use in unit tests.

jclee:
Also, something I missed from the previous PR: two instances of "philosohpy".

kjdoc/tour.md
kentonv: This suggestion doesn't seem to appear in the code view, and I'm not allowed to apply it from the thread view for some reason. Guess I have to do it manually.

kjdoc/tour.md
kentonv: Yeah, I can't think of anything to do about this... people will have to read the surrounding text.

kjdoc/tour.md
kentonv: Nah, I think that's an overly minor API detail.

kjdoc/tour.md
kentonv: `kj::Thread` is discussed earlier in the doc. It's not really directly related to promises. If I call it out specifically here, then I also have to explain that there's nothing special about `kj::Thread`-created threads and you can use some other threading library if you prefer, but then that all seems too verbose... so I think I just won't name it here.

kjdoc/tour.md
```diff
@@ -896,6 +897,18 @@ kj::Promise<kj::Tuple<kj::Own<Foo>, kj::String>> promise = ...;
 kj::Tuple<kj::Promise<kj::Own<Foo>>, kj::Promise<kj::String>> promises = promise.split();
 ```
 
+### Joining promises
+
+The opposite of forking promises is joining promises. There are two types of joins:
+* **Exclusive** joins wait for any one input promise to complete, then cancel the rest, returning the result of the promise that completed.
+* **Inclusive** joins wait for all input promises to complete, and render all of the results.
+
+For an exclusive join, use `promise.join(kj::mv(otherPromise))`. The two promises must return the same type. The result is a promise that returns whichever result is produced first, and cancels the other promise at that time. (To exclusively join more than two promises, call `.exclusiveJoin()` multiple times in a chain.)
```
Sarcasm, line 906: `promise.join` -> `promise.exclusiveJoin`

kentonv: thanks, fixed

kjdoc/tour.md
```diff
@@ -538,13 +538,476 @@ This section describes KJ APIs that control process execution and low-level inte
...
+  // is waiting on the final result of the promise.
+  KJ_LOG(WARNING, "It has been 5 seconds!!!");
+});
+kj::NEVER_DONE.wait(waitScope);
+```
+
+To solve this, use `.eagerlyEvaluate()`:
+
+```c++
+kj::Promise<void> timer.afterDelay(5 * kj::SECONDS)
```
harrishancock, line 736:
```suggestion
kj::Promise<void> promise = timer.afterDelay(5 * kj::SECONDS)
```
(I spotted this in the above snippet, but missed this instance of it.)

kjdoc/tour.md
```diff
@@ -538,13 +538,476 @@ This section describes KJ APIs that control process execution and low-level inte
...
+```
+
+KJ promises include "tail call optimization" for loops like the one above, so that the promise chain length remains finite no matter how many times the loop iterates.
+
+**WARNING!** It is very easy to accidentally break tail call optimization, creating a memory leak. Consider the following:
+
+```c++
+kj::Promise<void> boopEvery5Seconds(kj::Timer& timer) {
+  // WARNING! MEMORY LEAK!
+  timer.afterDelay(5 * kj::SECONDS).then([&timer]() {
```
harrishancock, line 819:
```suggestion
  return timer.afterDelay(5 * kj::SECONDS).then([&timer]() {
```

PR #1168, opened 2021-01-28
Fix two obscure RPC bugs

kentonv:
I'm not aware of either of these being an issue in practice. I noticed the first bug when reading the code while I was thinking about intentionally throwing exceptions from `send()` as a way to reject calls when overloaded, but I ended up changing my mind about doing that. I noticed the second bug while writing a unit test for the first bug.


c++/src/capnp/rpc-test.c++
```diff
@@ -1437,6 +1437,90 @@ KJ_TEST("when OutgoingRpcMessage::send() throws, we don't leak exports") {
   }
 }
 
+KJ_TEST("export the same promise twice") {
+  TestContext context;
+
+  bool exportExportIsPromise;
```
harrishancock, line 1443: `exportExport...`?

harrishancock (approved):

c++/src/capnp/rpc-test.c++
kentonv: o/\` I like to exportExport
I like to exportExport o/`

PR #1170, opened 2021-01-29
Fix build with C++20.

kentonv:
The biggest issue is that C++20 greatly expanded the overload search space for comparison operators. In particular, when evaluating `a == b`, C++20 will now look for overloads for both `a == b` and `b == a`. If it finds matches for both, with no reason to think one is "better" than the other, it will pronounce them ambiguous and error out!

This turns out to play very poorly with implicit conversions. If you have two types that can implicitly convert to each other (including two specializations of the same template), and both of them implement `operator==`, then you can no longer compare type A against type B, because the compiler can't decide whether to convert A to B or B to A. (Previously, it was always the right-hand side of the comparison that would be converted.)

It turns out this is even true when one of the conversions is private and inaccessible! The overload is still ambiguous, even though one of the choices would not compile.

Another class of issues is the handling of UTF-8 literals. The literal `u8"foo"` used to be type `const char*` but is now type `const char8_t*`, which is not compatible. Ugh. Since KJ always uses UTF-8, I added some automatic conversions.

### DIFF
--- a/c++/src/kj/compat/url.h
+++ b/c++/src/kj/compat/url.h
@@ -103,6 +103,15 @@ struct Url {
   ~Url() noexcept(false);
   Url& operator=(Url&&) = default;
 
+  inline Url(String&& scheme, Maybe<UserInfo>&& userInfo, String&& host, Vector<String>&& path,
+             bool hasTrailingSlash, Vector<QueryParam>&& query, Maybe<String>&& fragment,
+             UrlOptions options)
+      : scheme(kj::mv(scheme)), userInfo(kj::mv(userInfo)), host(kj::mv(host)), path(kj::mv(path)),
+        hasTrailingSlash(hasTrailingSlash), query(kj::mv(query)), fragment(kj::mv(fragment)),
+        options(options) {}
+  // This constructor makes brace initialization work in C++11 and C++20 -- but is technically not
+  // needed in C++14 nor C++17. Go figure.
+
   Url clone() const;
 
   enum Context {
--- a/c++/src/kj/filesystem.h
+++ b/c++/src/kj/filesystem.h
@@ -157,6 +157,13 @@ public:
   bool operator>=(PathPtr other) const;
   // Compare path components lexically.
 
+  bool operator==(const Path& other) const;
+  bool operator!=(const Path& other) const;
+  bool operator< (const Path& other) const;
+  bool operator> (const Path& other) const;
+  bool operator<=(const Path& other) const;
+  bool operator>=(const Path& other) const;
+
   uint hashCode() const;
   // Can use in HashMap.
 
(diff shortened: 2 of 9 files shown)


c++/src/kj/compat/url.h
```diff
@@ -103,6 +103,15 @@ struct Url {
...
   Url& operator=(Url&&) = default;
 
+  inline Url(String&& scheme, Maybe<UserInfo>&& userInfo, String&& host, Vector<String>&& path,
+             bool hasTrailingSlash, Vector<QueryParam>&& query, Maybe<String>&& fragment,
+             UrlOptions options)
+      : scheme(kj::mv(scheme)), userInfo(kj::mv(userInfo)), host(kj::mv(host)), path(kj::mv(path)),
+        hasTrailingSlash(hasTrailingSlash), query(kj::mv(query)), fragment(kj::mv(fragment)),
+        options(options) {}
+  // This constructor makes brace initialization work in C++11 and C++20 -- but is technically not
+  // needed in C++14 nor C++17. Go figure.
```
harrishancock, line 113: wat

Do you know the reason? Presumably this hasn't anything to do with the `operator==` or `u8` issues...

c++/src/kj/filesystem.h
```diff
@@ -157,6 +157,13 @@ class Path {
   bool operator>=(PathPtr other) const;
   // Compare path components lexically.
 
+  bool operator==(const Path& other) const;
+  bool operator!=(const Path& other) const;
+  bool operator< (const Path& other) const;
+  bool operator> (const Path& other) const;
+  bool operator<=(const Path& other) const;
+  bool operator>=(const Path& other) const;
```
harrishancock, line 165: Could this be simplified with the new spaceship operator? https://devblogs.microsoft.com/cppblog/simplify-your-code-with-rocket-science-c20s-spaceship-operator/

I suppose that would require a new #include, and more #iffery, so might not actually save many lines.

(Side note: I always wanted to be a spaceship operator.)

c++/src/kj/string.h
```diff
@@ -214,6 +225,16 @@ class String {
   inline bool operator<=(const StringPtr& other) const { return StringPtr(*this) <= other; }
   inline bool operator>=(const StringPtr& other) const { return StringPtr(*this) >= other; }
 
+  inline bool operator==(const String& other) const { return StringPtr(*this) == StringPtr(other); }
+  inline bool operator!=(const String& other) const { return StringPtr(*this) != StringPtr(other); }
+  inline bool operator< (const String& other) const { return StringPtr(*this) <  StringPtr(other); }
+  inline bool operator> (const String& other) const { return StringPtr(*this) >  StringPtr(other); }
+  inline bool operator<=(const String& other) const { return StringPtr(*this) <= StringPtr(other); }
+  inline bool operator>=(const String& other) const { return StringPtr(*this) >= StringPtr(other); }
```
harrishancock, line 233: Same spaceship operator comment here.

harrishancock (approved):

c++/src/kj/units-test.c++
```diff
@@ -349,6 +349,7 @@ TEST(UnitMeasure, BoundedMinMax) {
   assertTypeAndValue(boundedValue<4,t2>(3), kj::min(bounded<5>(), boundedValue<4,t2>(3)));
 }
 
+#if !_MSC_VER  // MSVC barfs on this test and I just don't have time to care.
```
harrishancock, line 352: I'm curious, was this an ICE, or just weird errors?

c++/src/kj/compat/url.h
kentonv: I didn't look into it, but my guess is that it doesn't like the non-default destructor -- which I only have so that it can be out-of-line since there's a lot of stuff to destroy.

c++/src/kj/filesystem.h
kentonv: I suppose we could delete everything except `==` and `<` since I think C++ will derive the rest from those.

But I want nothing to do with the spaceship operator. `strcmp()` has always hurt my brain trying to remember which result means what.

c++/src/kj/units-test.c++
kentonv: It looked like the magic assert logic was trying to stringify the operands to the comparisons, but there was no stringification defined. But earlier SFINAE should have prevented it from going down that path. I didn't dig in further.

c++/src/kj/string.h
```diff
@@ -136,15 +145,17 @@ class StringPtr {
...
-  inline constexpr StringPtr(ArrayPtr<const char> content): content(content) {}
+  inline explicit constexpr StringPtr(ArrayPtr<const char> content): content(content) {}
 
   ArrayPtr<const char> content;
 
   friend constexpr kj::StringPtr (::operator "" _kj)(const char* str, size_t n);
 };
 
+#if __cplusplus < 202000L
 inline bool operator==(const char* a, const StringPtr& b) { return b == a; }
```
ryanofsky, line 156: Note: this change does not just fix a compile-time error. It also fixes an infinite recursion bug if this overload is ever called by c++20 code.

Specifically, with Cap'n Proto 0.8.1 and earlier, if code is compiled against this header with `-std=c++20` and uses a `var == "literal"` or `"literal" == var` comparison where `var` is `StringPtr`, the code will either crash or loop infinitely without this fix.

If the same code is compiled with `-std=c++17` there is no problem because instead of this function calling itself recursively, it will implicitly convert `a` to a `StringPtr` call the StringPtr operator== method instead.

This problem was encountered at https://github.com/bitcoin/bitcoin/issues/33176 on ubuntu 22.04 which uses v0.8.0.

PR #1171, opened 2021-01-29
Fix TLS handshake errors being leaked from accept

bretthoerner:
Before this change a TLS wrapped network used in an accept loop would
have to handle errors that leaked during TLS handshake. This change brings
TLS wrapped networks in line with raw networks, in that only errors
during the underlying raw network accept() call will be thrown since
that indicates an issue with the underlying socket.

---

I guess the log could be more descriptive. I'm open to ideas. The actual function being called is `SSL_accept`.

### DIFF
--- a/c++/src/kj/compat/tls.c++
+++ b/c++/src/kj/compat/tls.c++
@@ -408,13 +408,21 @@ public:
 
   Promise<Own<AsyncIoStream>> accept() override {
     return inner->accept().then([this](kj::Own<AsyncIoStream> stream) {
-      return tls.wrapServer(kj::mv(stream));
+      return kj::evalNow([&] { return tls.wrapServer(kj::mv(stream)); })
+          .catch_([this](kj::Exception&& e) {
+        KJ_LOG(ERROR, "error accepting tls connection", e);
+        return accept();
+      });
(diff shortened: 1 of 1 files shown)


c++/src/kj/compat/tls.c++
```diff
@@ -408,13 +408,21 @@ public:
 
   Promise<Own<AsyncIoStream>> accept() override {
     return inner->accept().then([this](kj::Own<AsyncIoStream> stream) {
-      return tls.wrapServer(kj::mv(stream));
+      return tls.wrapServer(kj::mv(stream))
```
kentonv, line 411: Might be worth wrapping in `kj::evalNow()`, in case of synchronous exceptions.

kentonv:
Fine first step, but we're probably going to want to go further and set up some sort of concurrent accept loop so that if two connections arrive at once, we don't wait for the first handshake before starting on the second... which is gonna be pretty ugly. I guess this concurrent loop would need to place accepted connections into a producer/consumer queue that the TLS network's `accept()` pulls from... with some heuristic to stop accepting raw connections if there are unclaimed fully-ready connections in the queue.

PR #1173, opened 2021-01-29
Expose RpcSystemBase's accept loop promise

bretthoerner:
This allows users of the RPC system to detect when the accept loop stops
for some reason.

---

An alternative would be to have people manually start the RPC system and own the actual promise. It was a bit surprising to me (but not necessarily bad) that the system runs itself on construction.

### DIFF
--- a/c++/src/capnp/rpc-prelude.h
+++ b/c++/src/capnp/rpc-prelude.h
@@ -87,6 +87,8 @@ public:
 
   void setTraceEncoder(kj::Function<kj::String(const kj::Exception&)> func);
 
+  kj::Promise<void> run();
+
 private:
   class Impl;
   kj::Own<Impl> impl;
--- a/c++/src/capnp/rpc.c++
+++ b/c++/src/capnp/rpc.c++
@@ -3081,15 +3081,15 @@ public:
   Impl(VatNetworkBase& network, kj::Maybe<Capability::Client> bootstrapInterface)
       : network(network), bootstrapInterface(kj::mv(bootstrapInterface)),
         bootstrapFactory(*this), tasks(*this) {
-    tasks.add(acceptLoop());
+    acceptLoopPromise = acceptLoop().eagerlyEvaluate([](kj::Exception&& e) { KJ_LOG(ERROR, e); });
   }
   Impl(VatNetworkBase& network, BootstrapFactoryBase& bootstrapFactory)
       : network(network), bootstrapFactory(bootstrapFactory), tasks(*this) {
-    tasks.add(acceptLoop());
+    acceptLoopPromise = acceptLoop().eagerlyEvaluate([](kj::Exception&& e) { KJ_LOG(ERROR, e); });
   }
   Impl(VatNetworkBase& network, SturdyRefRestorerBase& restorer)
       : network(network), bootstrapFactory(*this), restorer(restorer), tasks(*this) {
-    tasks.add(acceptLoop());
+    acceptLoopPromise = acceptLoop().eagerlyEvaluate([](kj::Exception&& e) { KJ_LOG(ERROR, e); });
   }
 
   ~Impl() noexcept(false) {
@@ -3142,6 +3142,8 @@ public:
     traceEncoder = kj::mv(func);
   }
 
+  kj::Promise<void> run() { return kj::mv(acceptLoopPromise); }
+
 private:
   VatNetworkBase& network;
   kj::Maybe<Capability::Client> bootstrapInterface;
@@ -3149,6 +3151,7 @@ private:
   kj::Maybe<SturdyRefRestorerBase&> restorer;
   size_t flowLimit = kj::maxValue;
   kj::Maybe<kj::Function<kj::String(const kj::Exception&)>> traceEncoder;
+  kj::Promise<void> acceptLoopPromise = nullptr;
   kj::TaskSet tasks;
 
   typedef std::unordered_map<VatNetworkBase::Connection*, kj::Own<RpcConnectionState>>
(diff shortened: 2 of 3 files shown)


c++/src/capnp/rpc.c++
```diff
@@ -3179,16 +3182,10 @@ private:
...
+    return network.baseAccept().then(
         [this](kj::Own<VatNetworkBase::Connection>&& connection) {
       getConnectionState(kj::mv(connection));
-    });
-    return receive.then([this]() {
-      // No exceptions; continue loop.
-      //
-      // (We do this in a separate continuation to handle the case where exceptions are
-      // disabled.)
-      tasks.add(acceptLoop());
```
bretthoerner, line 3191: It's entirely possible that I'm missing something about this continuation, I don't know much about exceptions being disabled.

Also possible that I'm missing something about the previous use of `tasks` here that differs from storing a promise loop?

kentonv: The `tasks.add(acceptLoop())`thing is because this code actually predates promise tail call optimization, so I had to do it that way to avoid a memory leak. `return acceptLoop()` is strictly better and we should have changed that a long time ago.

As for the exceptions issue: If `getConnectionState()` throws a recoverable exception, but the library is compiled with `-fno-exceptions`, then `getConnectionState()` will still return normally. Once the contituation completes, the event loop will notice the exception was raised, and will discard the continuation result in favor of the exception. But if the continuation calls `tasks.add(acceptLoop())` immediately after `getConnectionState()`, this will be executed even when `getConnectionState()` threw, and hence the loop would continue despite the exception. However, now that you've changed it to `return acceptLoop()`, I think that problem is solved. (Technically `acceptLoop()` will be invoked an extra time and then immediately canceled, but that's not the kind of thing I worry too much about when it comes to supporting no-exceptions.)

So I think your changes here are fine.

c++/src/capnp/rpc.c++
```diff
@@ -3142,13 +3142,16 @@ public:
...
+  kj::Promise<void> getAcceptLoopPromise() { return acceptLoopPromise.addBranch(); }
+
 private:
   VatNetworkBase& network;
   kj::Maybe<Capability::Client> bootstrapInterface;
   BootstrapFactoryBase& bootstrapFactory;
   kj::Maybe<SturdyRefRestorerBase&> restorer;
   size_t flowLimit = kj::maxValue;
   kj::Maybe<kj::Function<kj::String(const kj::Exception&)>> traceEncoder;
+  kj::ForkedPromise<void> acceptLoopPromise = nullptr;
```
kentonv, line 3154: Instead of a forked promise, how about we tell people that they can only call `getAcceptLoopPromise()` once? That's something I commonly do in KJ code in order to avoid unnecessary promise forking.

c++/src/capnp/rpc.c++
```diff
@@ -3081,15 +3081,15 @@ public:
   Impl(VatNetworkBase& network, kj::Maybe<Capability::Client> bootstrapInterface)
       : network(network), bootstrapInterface(kj::mv(bootstrapInterface)),
         bootstrapFactory(*this), tasks(*this) {
-    tasks.add(acceptLoop());
+    acceptLoopPromise = acceptLoop().fork();
```
kentonv, line 3084: We should add an `.eagerlyEvaluate()` with a callback that logs the exception... in order to maintain the existing behavior for anyone who doesn't update their code to call `getAcceptLoopPromise()`.

c++/src/capnp/rpc.c++
bretthoerner: Ahh, makes sense.

c++/src/capnp/rpc-prelude.h
```diff
@@ -87,6 +87,8 @@ class RpcSystemBase {
 
   void setTraceEncoder(kj::Function<kj::String(const kj::Exception&)> func);
 
+  kj::Promise<void> getAcceptLoopPromise();
```
kentonv, line 90: Let's call this `run()` -- which is what I would have called it if it were required to call -- and document it as:

> Listens for incoming RPC connections and handles them. Never returns normally, but could throw an exception if the system becomes unable to accept new connections (e.g. because the underlying listen socket becomes broken somehow).
>
> For historical reasons, the `RpcSystem` will actually run itself even if you do not call this. However, if an exception is thrown, the `RpcSystem` will log the exception to the console and then cease accepting new connections. In this case, your server may be in a broken state, but without restarting. All servers should therefore call `run()` and handle failures in some way.

c++/src/capnp/rpc.c++
bretthoerner: Nice, thanks for the explanation!

kentonv:
I guess `TwoPartyVatNetwork` is not really affected by this problem because it implements a mostly-fake `accept()` method that only returns a result once... `TwoPartyServer` actually implements its own accept loop and construct a new `RpcSystem` for each connection... So probably all users of this library that weren't written by me are not affected by this problem, heh.

c++/src/capnp/rpc-prelude.h
```diff
@@ -87,6 +87,16 @@ class RpcSystemBase {
 
   void setTraceEncoder(kj::Function<kj::String(const kj::Exception&)> func);
 
+  kj::Promise<void> run();
+  // Listens for incoming RPC connections and handles them. Never returns normally, but could throw
```
kentonv, line 91: So, `rpc-prelude.h` is supposed to be private implementation details. This doc comment should go in `rpc.h` instead. I guess you can declare `run()` there with inline implementation `{ return RpcSystemBase::run(); }`, just for the sake of putting a doc comment on it...

bretthoerner:
Tests:

```
E: Package 'clang-5.0' has no installation candidate
```

```
E: Package 'g++-5' has no installation candidate
```

:thinking:

kentonv:
> E: Package 'clang-5.0' has no installation candidate

Ugh GitHub actions are so unstable. Ignoring for now.

PR #1175, opened 2021-02-05
Try to fix GCC 5 / Clang 5 CI builds by pinning to older Ubuntu.

kentonv:


kentonv:
Seems like that worked. But I'm going to tack this onto #1174 instead of submitting it separately, so that I can actually verify those tests pass...

PR #1180, opened 2021-02-12
[WIP] fix `KJ_CLEAN_SHUTDOWN` behavior

haneefmubarak:
Prior to these changes, using KJ_CLEAN_SHUTDOWN results in the following error:

```
failed: unknown non-KJ exception of type:
kj::TopLevelProcessContext::CleanShutdownException
```

### DIFF
--- a/c++/src/kj/exception.c++
+++ b/c++/src/kj/exception.c++
@@ -40,6 +40,7 @@
 #include "threadlocal.h"
 #include "miniposix.h"
 #include "function.h"
+#include "main.h"
 #include <stdlib.h>
 #include <exception>
 #include <new>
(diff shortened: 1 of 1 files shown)


mologie:
Tried this locally on macOS: I can confirm that it fixes using Clang's [Source-based Code Coverage](https://clang.llvm.org/docs/SourceBasedCodeCoverage.html) with `kj/main.h`. Otherwise, the `profraw` file generated by LLVM would be empty. Thanks!

haneefmubarak:
Awesome! I've tested this on Linux as part of a larger application test suite that uses KJ heavily; when I get a moment I'm going to add a test case to the Cap'n Proto KJ test suite and then hopefully it should be ready for merge.

kentonv:
Heh, I forgot this existed, apparently it dates all the way back to the first version of kj::Main and I haven't thought about it since 2013. Thanks for fixing it.

PR #1184, opened 2021-02-14
Leave Maybe<T> empty after its contents are taken by another Maybe

bretthoerner:
This has come up before as something people (me) didn't expect. I _think_ the consensus was:

* You shouldn't try to use a moved-out-of-object
* ...but leaving the moved-out-of Maybe in a valid state for things like `int` is a foot gun for people who expect it be left "empty" after a move

It's a bit scary since this is technically backward incompatible...

### DIFF
--- a/c++/src/kj/common-test.c++
+++ b/c++/src/kj/common-test.c++
@@ -53,7 +53,7 @@ struct Immovable {
 struct CopyOrMove {
   // Type that detects the difference between copy and move.
   CopyOrMove(int i): i(i) {}
-  CopyOrMove(CopyOrMove&& other): i(other.i) { i = -1; }
+  CopyOrMove(CopyOrMove&& other): i(other.i) { other.i = -1; }
   CopyOrMove(const CopyOrMove&) = default;
 
   int i;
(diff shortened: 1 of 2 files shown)


c++/src/kj/common-test.c++
```diff
@@ -285,9 +285,40 @@ TEST(Common, Maybe) {
     CopyOrMove x(123);
     Maybe<CopyOrMove&> m(x);
     Maybe<CopyOrMove> m2 = kj::mv(m);
-    KJ_EXPECT(KJ_ASSERT_NONNULL(m).i == 123);
+    KJ_EXPECT(m == nullptr);  // m is still moved out of and cleared
```
bretthoerner, line 288: Not sure if I'm understanding this test -- I would expect `m` to be `nullptr` here, but it's a bit weird considering the comment says "Test that initializing Maybe<T> from Maybe<T&>&& does a copy, not a move."

kentonv: Hmm... it looks like what it really meant to test was:

    KJ_EXPECT(x.i == 123);

That is, even though `m` was moved, the thing it pointed to, `x`, was copied, not moved.

kentonv: See #805

bretthoerner: > That is, even though m was moved, the thing it pointed to, x, was copied, not moved.

Ahhh, got it. But even if a move was performed, wouldn't `x.i` remain 123? I think this is verified from the other end,

https://github.com/capnproto/capnproto/blob/7a31fa889d13286f02a19444e7073d12a6e773c7/c%2B%2B/src/kj/common-test.c%2B%2B#L53-L60

Because if a move was performed, the next line would `== 122`?

kentonv: Wow uh I think that move constructor also has a bug. It should be setting `other.i = -1`. Hence you can tell if the value was moved away because it would be overwritten with -1.

Must have been drunk when I wrote this.

bretthoerner: Heh, I also kept misreading it as `-= 1`.

Fixed, I think.

PR #1185, opened 2021-02-15
JSON: Fix inflexibilities with unions

mologie:
This pull request addresses two minor shortcomings of the current JSON implementation that I stumbled over recently. This first part is probably an oversight and trivial, the second part is a feature that removes a limitation. All changes are backwards-compatible.

### 1. Flattening anonymous unions

The following construct was so far rejected by the JSON en-/decoder:

```capnp
struct TestAnonymousUnion $Json.flatten() $Json.discriminator(name = "type") {
  union {
    foo @0 :Foo $Json.flatten();
    ref @1 :Bar $Json.flatten();
  }
}
```

However, the following was always legal:

```capnp
struct TestUnion {
  u :union $Json.flatten() $Json.discriminator(name = "type") {
    foo @0 :Foo $Json.flatten();
    bar @1 :Bar $Json.flatten();
  }
}
```

I believe that's totally unfair, anonymous unions deserve love too, and thus this merge request makes the first one legal too. No code changes are required apart from modifying the schema. I however added a test for it.

### 2. Allow referencing union tags from flattened member structs

A union tag, defined via `Json.discriminator`, configures which union member the JSON en-/decoder should pick. It is registered as internal field that so far could no longer appear in the structs below it if they were flattened. For example, the following was illegal:

```capnp
struct TestTagReference $Json.flatten() $Json.discriminator(name = "type") {
  union {
    foo @0 :Foo $Json.flatten();
    ref @1 :Ref $Json.flatten();
  }
}

struct Bar {
  type @0 :Text; #<- error: name already reserved by discriminator field
  value @1 :Text;
}
```

This pull request makes the JSON implementation accept the above construct and do the obvious: The discriminator field's value is simply copied into Bar. Thus, Bar's "type" will always equal "bar" when decoded from JSON as union member. This adds a good deal of flexibility. For example, it allows Bar to be reused in the schema where a "type" field must still explicitly appear. The encoding of Bar to JSON also happens to be a valid TestTagReference. 

My use case for both changes is being able to use Cap'n Proto's JSON decoder as ingress for messages that come in a fixed, legacy format that I have no control over. They can now be parsed seamlessly to Cap'n Proto structs and forwarded via RPC.

### DIFF
--- a/c++/src/capnp/compat/json-test.c++
+++ b/c++/src/capnp/compat/json-test.c++
@@ -835,7 +835,7 @@ public:
 };
 
 KJ_TEST("register capability handler") {
-  // This test currently only checks that this compiles, which at one point wasn't the caes.
+  // This test currently only checks that this compiles, which at one point wasn't the case.
   // TODO(test): Actually run some code here.
 
   TestCapabilityHandler handler;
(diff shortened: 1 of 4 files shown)


mologie:
I see there is one test failure that I did not experience locally, will look into it next morning. It's getting late here :-).

kentonv:
Hmm, what does it mean to flatten an anonymous union? Isn't an anonymous union is already flattened by definition? Does the test still pass if you remove the annotation?

For the second issue, would it work to do something like:

```
struct Bar {
  value @0 :Text;
}

struct BarWithType {
  type @0 :Text;
  value @1 :Bar $Json.flatten();
}
```

Then use the appropriate one depending on whether the context already defines `type`? Though it's still kind of weird to me that you would need to manually set the `type` field to a constant value, that seems wrong. I admit without seeing the real use case I'm a little lost here.

mologie:
> Hmm, what does it mean to flatten an anonymous union? Isn't an anonymous union is already flattened by definition? Does the test still pass if you remove the annotation?

It is indeed! Not sure what I was thinking. This was a placebo solution, and the flatten annotation within the anonymous union are sufficient already. I have updated the pull request to drop the commit.

> For the second issue, would it work to do something like [...]

Thank you for the input, I will have to think about whether I can restructure the whole thing to fit into what you suggested. However it looks vague familiar to what I had a few months ago and dropped in favour of this patch. I will get back to you early in the morning with a real-world use case.

mologie:
You are right @kentonv, I was able to restructure my schema and application to use a type wrapper struct depending on whether `type` is expected or not, and unwrap it when passing structs to backends.

I do not believe that this PR is necessary anymore. I could not find an example where the workaround you suggested would not yield the desired result. Thus, this PR would add unnecessary complexity.

Happy to be proven wrong! Closing in favor of me not writing stupid code and schemas to deal with legacy input ;-).

PR #1188, opened 2021-02-23
Adds fuzz targets for oss-fuzz integration

catenacyber:
@kentonv this follows our discussion in https://github.com/google/oss-fuzz/pull/5202

I removed `ParseErrorCatcher` and `TopLevelProcessContext ` from the fuzz targets
I added the compilation scheme with auto tools for the run fuzz target.

As for the afl test case, to compile it with autotools, you need to
```
./configure --disable-shared
make
make fuzz-capnp-run
```
It uses `LIB_FUZZING_ENGINE if defined as an environment variable, and a dummy fuzz driver otherwise

Do you have remarks about the style and formatting ?


catenacyber:
Friendly ping @kentonv

c++/Makefile.am
```diff
@@ -576,6 +576,14 @@ capnp_evolution_test_SOURCES = src/capnp/compiler/evolution-test.c++
 
 capnp_afl_testcase_LDADD = libcapnp-test.a libcapnp-rpc.la libcapnp.la libkj.la libkj-async.la
 capnp_afl_testcase_SOURCES = src/capnp/afl-testcase.c++
+
+fuzz_capnp_run_LDADD = libcapnp-test.a libcapnp-rpc.la libcapnp.la libkj.la libkj-async.la
+fuzz_capnp_run_SOURCES = src/capnp/fuzz-run.c++
+if HAS_NO_FUZZING_ENGINE
+    fuzz_capnp_run_SOURCES += src/capnp/fuzz-driver.c++
```
kentonv, line 583: What's the purpose of this fuzz driver exactly? Maybe we should just not build the executable at all if a fuzz driver wasn't specified?

c++/Makefile.am
```diff
@@ -481,7 +481,7 @@ capnp_test_LDADD = libcapnp-test.a libcapnp.la libkj-test.la libkj.la
 
 else !LITE_MODE
 
-check_PROGRAMS = capnp-test capnp-evolution-test capnp-afl-testcase
+check_PROGRAMS = capnp-test capnp-evolution-test capnp-afl-testcase fuzz-capnp-run
```
kentonv, line 484: `fuzz-capnp-run` seems like a strange name to me. Maybe `capnp-llvm-fuzzer-testcase`?

c++/src/capnp/fuzz-convert.c++
```diff
@@ -0,0 +1,29 @@
...
+    capnp::ReaderOptions options;
+
+    //TODO use readOneAndConvert(input, output) ?
+    //with int fd = open("/dev/null", O_WRONLY);
+    //and kj::FdOutputStream output(fd);
+
+    KJ_IF_MAYBE(e, kj::runCatchingExceptions([&]() {
+        kj::BufferedInputStreamWrapper input(ais);
+        capnp::InputStreamMessageReader message(input, options);
+        capnp::PackedMessageReader messagep(input, options);
```
kentonv, line 23: This test case seems kind of uninteresting as it isn't actually invoking any capnp parsing code -- it's just setting up a `MessageReader` and then throwing it away. Since it's a `PackedMessageReader`, I suppose it does exercise some of the "packed" format decoder, but if that's the goal it might make sense to target `PackedInputStream` more directly. Otherwise, I think I'd just delete `fuzz-convert.c++`.

c++/src/capnp/fuzz-run.c++
```diff
@@ -0,0 +1,21 @@
+#include "test-util.h"
+#include <kj/main.h>
+#include "serialize.h"
+#include <capnp/test.capnp.h>
+#include <unistd.h>
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t* Data, size_t Size)
+{
```
kentonv, line 8: Can we follow KJ/capnp formatting conventions here, like putting opening braces at the end of the preceding line, and using two-space indents?

c++/src/capnp/fuzz-run.c++
```diff
@@ -0,0 +1,21 @@
+#include "test-util.h"
```
kentonv, line 1: Can we add some comments to this file explaining what it's for and how to use it?

kentonv (commented):
Sorry for the delay, this has been sitting in my inbox but I've been really slammed in the last couple weeks, having trouble keeping up.

Thanks for working on this! I'm excited to have more fuzzing of Cap'n Proto.

c++/Makefile.am
catenacyber: The main purpose is for CI to check compilation of the fuzz targets, without requiring a real/complex fuzz driver

c++/Makefile.am
catenacyber: ok

c++/src/capnp/fuzz-run.c++
catenacyber: ok, let me know if there is more to do

c++/src/capnp/fuzz-run.c++
catenacyber: Adding something

c++/src/capnp/fuzz-convert.c++
catenacyber: Feel free to delete it.

This was inspired by `readOneAndConvert` which does
```
      case Format::BINARY: {
        capnp::InputStreamMessageReader message(input, options);
        return writeConversion(message.getRoot<AnyStruct>(), output);
      }
      case Format::PACKED: {
        capnp::PackedMessageReader message(input, options);
        return writeConversion(message.getRoot<AnyStruct>(), output);
      }
```

c++/Makefile.am
kentonv: Why not actually install the LLVM fuzzer in CI?

c++/src/capnp/fuzz-convert.c++
kentonv: Yes, `readOneAndConvert()`, when implementing `capnp convert packed:binary`, does not actually do very much.

Let's remove `fuzz-convert.c++`.

catenacyber: done

super-test.sh
```diff
@@ -97,6 +97,9 @@ while [ $# -gt 0 ]; do
       ;;
     clang* )
       export CXX=clang++${1#clang}
+      if [ "$1" -eq clang-9]
```
kentonv: I suppose this is because clang-5.0 didn't support this flag? Let's do `-ne clang-5.0` instead of `-eq clang-9` so we don't accidentally stop using this if/when we upgrade to newer clangs in CI.

c++/Makefile.am
```diff
@@ -576,6 +576,14 @@ capnp_evolution_test_SOURCES = src/capnp/compiler/evolution-test.c++
 
 capnp_afl_testcase_LDADD = libcapnp-test.a libcapnp-rpc.la libcapnp.la libkj.la libkj-async.la
 capnp_afl_testcase_SOURCES = src/capnp/afl-testcase.c++
+
+capnp_llvm_fuzzer_testcase_LDADD = libcapnp-test.a libcapnp-rpc.la libcapnp.la libkj.la libkj-async.la
+capnp_llvm_fuzzer_testcase_SOURCES = src/capnp/fuzz-run.c++
+if HAS_NO_FUZZING_ENGINE
+    capnp_llvm_fuzzer_testcase_SOURCES += src/capnp/fuzz-driver.c++
```
kentonv, line 583: Let's remove `fuzz-driver.c++` since we'll test using the actual clang fuzzer in CI.

c++/Makefile.am
catenacyber: Trying that with the latest commit

c++/Makefile.am
catenacyber: Ok, so no fuzzing driver, no fuzzing target get compiled

super-test.sh
catenacyber: I think it started at clang-6 indeed

c++/Makefile.am
```diff
@@ -482,6 +482,9 @@ capnp_test_LDADD = libcapnp-test.a libcapnp.la libkj-test.la libkj.la
 else !LITE_MODE
 
 check_PROGRAMS = capnp-test capnp-evolution-test capnp-afl-testcase
+if !HAS_NO_FUZZING_ENGINE
```
kentonv, line 485: Sorry, this is nitpicky but I think it's the last thing... This is a double-negative. I think it would be easier to read as a positive, i.e. the variable should be `HAS_FUZZING_ENGINE`.

catenacyber: Thought the same
No problem with being nitpicky
Looks like this double-negative build succeeded in building the fuzz target

c++/src/capnp/CMakeLists.txt
```diff
@@ -291,3 +291,8 @@ if(BUILD_TESTING)
     add_test(NAME capnp-evolution-tests-run COMMAND capnp-evolution-tests)
   endif()  # NOT CAPNP_LITE
 endif()  # BUILD_TESTING
+
+if(DEFINED ENV{LIB_FUZZING_ENGINE})
+  add_executable(fuzz_run fuzz-run.c++ test-util.c++ test-util.h ${test_capnp_cpp_files} ${test_capnp_h_files})
```
kentonv, line 296: Ugh sorry, just noticed the cmake file is still calling this fuzz_run, can we make it capnp_llvm_fuzzer_testcase here too? And rename the source file to llvm-fuzzer-testcase.

catenacyber: Done.
Do you want me to squash the commits ?

kentonv: I was also hoping the source file could be renamed to `llvm-fuzzer-testcase.c++`, but it looks like it's still `fuzz-run.c++` right now.

I can tell GitHub to squash when I merge.

catenacyber: Renamed

kentonv:
Thanks for working on this!

kentonv:
(And sorry for all the back-and-forth...)

catenacyber:
Thank you @kentonv 
I just set https://github.com/google/oss-fuzz/pull/5202 ready for review.
You can also ack there

PR #1190, opened 2021-02-25
Add ability to check if a promise is valid.

vlovich:
Useful as an assert if using a cross-thread promise fulfiller pair if we want to enforce that we only take the promise once.


kentonv:
Hmm, I'd say if you have a situation where it's not obvious from lexical scope alone whether or not a promise has been consumed, then you should probably wrap it in a `Maybe`.

PR #1191, opened 2021-03-02
Apply some clang-tidy performance-unnecessary-value-param fixes

MaskRay:


kentonv:
Is there a measurable benefit from these changes? If not I'd rather leave the code as it is. We've seen analyzer-generated changes like this break things in the past.

MaskRay:
Beside minor performance improvement, consistency.

For instance, `KJ_STRINGIFY` accepts `const &` so the forwarder `stringify` would better using `const &` as well.

kentonv:
OK, thanks but I'd rather leave the code as it is unless you're seeing this measurably improve performance in some benchmark.

PR #1194, opened 2021-03-09
Add page to support using capnproto.org for the Go implementation

zenhack:
Per https://github.com/capnproto/go-capnproto2/issues/166

### DIFF
--- a/doc/go/capnp.html
+++ b/doc/go/capnp.html
@@ -1,0 +1,18 @@
+---
+layout: none
+---
+<!doctype html>
+<html>
+  <!--
+    This page exists to set up an import path for the Go implementation; see:
+
+    * https://golang.org/cmd/go/#hdr-Remote_import_paths
+    * https://github.com/capnproto/go-capnproto2/issues/166
+  -->
+  <head>
+    <meta name="go-import" content="capnproto.org/go/capnp git https://github.com/capnproto/go-capnproto2" />
+    <meta http-equiv="Refresh" content="0; url='https://github.com/capnproto/go-capnproto2'" />
+  </head>
+  <body>
+  </body>
+</html>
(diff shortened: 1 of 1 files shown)


doc/go/capnp.html
```diff
@@ -0,0 +1,18 @@
...
+<!doctype html>
+<html>
+  <!--
+    This page exists to set up an import path for the Go implementation; see:
+
+    * https://golang.org/cmd/go/#hdr-Remote_import_paths
+    * https://github.com/capnproto/go-capnproto2/issues/166
+  -->
+  <head>
+    <meta name="go-import" content="capnproto.org/go/capnp git https://github.com/capnproto/go-capnproto2" />
```
zombiezen, line 13: You'll also want to add `go-source` to help direct godoc.org (and I think pkg.go.dev as well?) to show the data correctly.

```suggestion
    <meta name="go-import" content="capnproto.org/go/capnp git https://github.com/capnproto/go-capnproto2.git">
    <meta name="go-source" content="capnproto.org/go/capnp https://github.com/capnproto/go-capnproto2 https://github.com/capnproto/go-capnproto2/tree/master{/dir} https://github.com/capnproto/go-capnproto2/blob/master{/dir}/{file}#L{line}">
```

zenhack: Applied, though the `.git` suffix is not strictly necessary.

kentonv:
Sorry for the delay on this.

So, I deployed this, but I think it's wrong. This serves from the URL `capnproto.org/go/capnp.html`, but I think we actually want to serve from `capnproto.org/go/capnp/`, right? (Note the trailing slash -- this is not just a matter of removing the `.html` extension.)

kentonv:
I fixed this with: b5ab41ea033874cb3c144b9ad87e3442a297ed55

This is now deployed.

Let me know if that was the wrong thing to do. But it seems to match what @zombiezen's site does.

zenhack:
Makes sense to me.

PR #1197, opened 2021-03-31
Make sigset safe to use in memcmp

vlovich:
sigemptyset isn't guaranteed to clear the structure to a level where a
byte-level comparison is safe (e.g. padding, unused fields, etc).
Explicitly 0-initialize the structs before comparison.

This was found by running async-unix-test under valgrind 3.16.1 on Arch
running 5.10.23-1-lts and glibc 2.33. Didn't happen always, but like
80-90% of the time. After this change confirmed that valgrind doesn't
complain about use of uninitialized memory.

PR #1198, opened 2021-04-01
Skip RT signal test on aarch64 to work around possible qemu-user bug.

kentonv:


harrishancock (approved):
https://github.com/capnproto/capnproto/pull/1197 didn't help? :(

kentonv:
@harrishancock It didn't. Which doesn't seem that surprising in retrospect -- that issue only reproduced on Arch, not Debian.

PR #1200, opened 2021-04-05
Fix logging tests, context filename trimming

jclee:
It seems that the mechanism for capturing logs in the logging tests also ate
the exceptions that the asserts rely upon, so many of the test assertions had
no effect.


kentonv (approved):
Doh. I suspect the test worked correctly when originally written using googletest. The bug would have been introduced when I created KJ test and shimmed out the googletest API on top of that. KJ test is based on KJ logging, so then there's a cycle here.

PR #1204, opened 2021-04-06
Add a WebSocket implementation of MessageStream.

zenhack:


c++/src/capnp/compat/websocket-rpc-test.c++
```diff
@@ -0,0 +1,73 @@
+
+#include "websocket-rpc.h"
+#include <kj/test.h>
+
+#include "json.capnp.h" // Arbitrary schema, just so we have something to play with
```
kentonv, line 5: Usually we'd use `test.capnp`, especially `TestAllTypes`. See also `test-util.h` which has a bunch of methods for filling in the content and checking it later.

c++/src/capnp/compat/websocket-rpc-test.c++
```diff
@@ -0,0 +1,73 @@
...
+  class FailErrorHandler final : public kj::TaskSet::ErrorHandler {
+    public:
+      void taskFailed(kj::Exception&& exception) override {
+        KJ_FAIL_ASSERT(exception);
+      }
+  };
+};
+
+KJ_TEST("WebSocketMessageStream") {
+  auto io = kj::setupAsyncIo();
```
kentonv, line 17: I think this test doesn't actually use OS I/O, so you can do:

```suggestion
  kj::EventLoop loop;
  kj::WaitScope waitScope(loop);
```

c++/src/capnp/compat/websocket-rpc-test.c++
```diff
@@ -0,0 +1,73 @@
...
+
+  auto originalSegments = originalMsg.getSegmentsForOutput();
+
+  capnp::_::WebSocketMessageStream::FailErrorHandler errorHandler;
+  kj::TaskSet tasks(errorHandler);
+
+  // Send the message across the websocket, make sure it comes out unharmed.
+  tasks.add(msgStreamA.writeMessage(nullptr, originalSegments));
+  tasks.add(msgStreamB.tryReadMessage(nullptr)
+    .then([&](auto maybeResult) -> kj::Promise<void> {
```
kentonv, line 40: This is cosmetic, but usually I'd write this like:

```suggestion
  kj::Promise<void> writePromise = msgStreamA.writeMessage(nullptr, originalSegments);
  auto maybeResult = msgStreamB.readMessage(nullptr).wait(waitScope);
  writePromise.wait(waitScope);
```

Then proceed with the checking of the result.

I think it comes out a little easier to read, and you can avoid the need to define `FailErrorHandler` (which, incidentally, as written, might not be correct -- I don't think you can just rethrow the exception in an error handler, since the purpose of the error handler was to handle an uncaught exception).

c++/src/capnp/compat/websocket-rpc.c++
```diff
@@ -0,0 +1,107 @@
...
+  : socket(socket),
+    maxMessageReceiveSize(maxMessageReceiveSize)
+  {};
+
+kj::Promise<kj::Maybe<MessageReaderAndFds>> WebSocketMessageStream::tryReadMessage(
+    kj::ArrayPtr<kj::AutoCloseFd> fdSpace,
+    ReaderOptions options, kj::ArrayPtr<word> scratchSpace) {
+  return socket.receive(maxMessageReceiveSize)
+    .then([options, scratchSpace](auto msg) -> kj::Promise<kj::Maybe<MessageReaderAndFds>> {
+      KJ_SWITCH_ONEOF(msg) {
```
kentonv, line 17: We normally intend line continuations by 4 spaces, but we indent the body of a lambda by two spaces past the parent block. So this should be:

```suggestion
      .then([options, scratchSpace](auto msg) -> kj::Promise<kj::Maybe<MessageReaderAndFds>> {
    KJ_SWITCH_ONEOF(msg) {
```

And then rest of the lambda body should be un-indented two spaces. It looks like several other uses of lambdas in this PR should be updated similarly.

c++/src/capnp/compat/websocket-rpc.c++
```diff
@@ -0,0 +1,107 @@
...
+        KJ_CASE_ONEOF(closeMsg, kj::WebSocket::Close) {
+          return kj::Maybe<MessageReaderAndFds>();
+        }
+        KJ_CASE_ONEOF(str, kj::String) {
+          KJ_FAIL_REQUIRE(
+            "Unexpected websocket text message; expected only binary messages."
+          );
```
kentonv, line 24:
```suggestion
              "Unexpected websocket text message; expected only binary messages.");
```

c++/src/capnp/compat/websocket-rpc.c++
```diff
@@ -0,0 +1,107 @@
...
+          break;
+        }
+        KJ_CASE_ONEOF(bytes, kj::Array<byte>) {
+          kj::ArrayPtr<byte> ptr(bytes);
+          if(reinterpret_cast<uintptr_t>(&bytes[0]) % alignof(word) == 0
```
kentonv, line 29: Nitpick: We put a space between a control keyword (`if`, `for`, ...) at a `(`:

```suggestion
          if (reinterpret_cast<uintptr_t>(&bytes[0]) % alignof(word) == 0
```

I don't have a good reason for this rule... it comes from Google style and I never unlearned it.

(Several other instances of this in the PR.)

kentonv, line 29: Use `bytes.begin()` instead of `&bytes[0]` to avoid an out-of-bounds array access exception when the array is empty.

```suggestion
          if(reinterpret_cast<uintptr_t>(bytes.begin()) % alignof(word) == 0
```

c++/src/capnp/compat/websocket-rpc.c++
```diff
@@ -0,0 +1,107 @@
...
+          return kj::Maybe<MessageReaderAndFds>();
+        }
+        KJ_CASE_ONEOF(str, kj::String) {
+          KJ_FAIL_REQUIRE(
+            "Unexpected websocket text message; expected only binary messages."
+          );
+          break;
+        }
+        KJ_CASE_ONEOF(bytes, kj::Array<byte>) {
+          kj::ArrayPtr<byte> ptr(bytes);
```
kentonv, line 28: Unused?

```suggestion
```

c++/src/capnp/compat/websocket-rpc.c++
```diff
@@ -0,0 +1,107 @@
...
+          if(reinterpret_cast<uintptr_t>(&bytes[0]) % alignof(word) == 0
+              && bytes.size() % sizeof(word) == 0) {
+            return kj::Maybe(MessageReaderAndFds {
+              .reader = kj::heap<FlatArrayMessageReader>(
+                  kj::arrayPtr(
+                    reinterpret_cast<word *>(&bytes[0]),
```
kentonv, line 34:
```suggestion
                    reinterpret_cast<word *>(bytes.begin()),
```

c++/src/capnp/compat/websocket-rpc.c++
```diff
@@ -0,0 +1,107 @@
...
+                  ),
+                  options
+                ).attach(kj::mv(bytes)),
+              .fds = nullptr
+            });
+          } else {
+            // The array is misaligned or not a whole number of words, so we
+            // need to copy it.
+            auto stream = kj::heap<kj::ArrayInputStream>(bytes);
+            auto reader = kj::heap<InputStreamMessageReader>(*stream, options, scratchSpace);
```
kentonv, line 45: I think it would be cheaper to make an aligned copy of the whole buffer rather than to use `InputStreamMessageReader` around `ArrayInputStream`. These classes will end up copying all the underlying bytes anyway and will do more allocations overall.

c++/src/capnp/compat/websocket-rpc.c++
```diff
@@ -0,0 +1,107 @@
...
+  // TODO(perf): Right now the WebSocket interface only supports send() for
+  // contiguous arrays, so we need to copy the whole message into a new buffer
+  // in order to send it, whereas ideally we could just write each segment
+  // (and the segment table) in sequence. Perhaps we should extend the WebSocket
+  // interface to be able to send an ArrayPtr<ArrayPtr<byte>> as one binary
+  // message, and then use that to avoid an extra copy here.
+
+  auto stream = kj::heap<kj::VectorOutputStream>(
+    computeSerializedSizeInWords(segments) * sizeof(word)
+  );
```
kentonv, line 69:
```suggestion
      computeSerializedSizeInWords(segments) * sizeof(word));
```

c++/src/capnp/compat/websocket-rpc.c++
```diff
@@ -0,0 +1,107 @@
...
+
+kj::Promise<void> WebSocketMessageStream::writeMessages(
+    kj::ArrayPtr<kj::ArrayPtr<const kj::ArrayPtr<const word>>> messages) {
+  if(messages.size() == 0) {
+    return kj::READY_NOW;
+  }
+  return writeMessage(nullptr, messages[0])
+    .then([this, messages = messages.slice(1, messages.size())]() -> kj::Promise<void> {
+      return writeMessages(messages);
+    });
```
kentonv, line 83:
```suggestion
      .then([this, messages = messages.slice(1, messages.size())]() -> kj::Promise<void> {
    return writeMessages(messages);
  });
```

c++/src/capnp/compat/websocket-rpc.c++
```diff
@@ -0,0 +1,107 @@
...
+    computeSerializedSizeInWords(segments) * sizeof(word)
+  );
+  capnp::writeMessage(*stream, segments);
+  auto arrayPtr = stream->getArray();
+  return socket.send(arrayPtr).attach(kj::mv(stream));
+}
+
+kj::Promise<void> WebSocketMessageStream::writeMessages(
+    kj::ArrayPtr<kj::ArrayPtr<const kj::ArrayPtr<const word>>> messages) {
+  if(messages.size() == 0) {
```
kentonv, line 77:
```suggestion
  // TODO(perf): Extend WebSocket interface with a way to write multiple messages at once.

  if(messages.size() == 0) {
```

kentonv (commented):
Nice!

We'll need to add this to the makefiles in order for CI to actually build it.

c++/src/capnp/compat/websocket-rpc-test.c++
zenhack: This does appear to work as you suggest, but it wasn't obvious to me from the docs that you can rely on the write end resolving if nothing is waiting on the read end -- the comment for newWebSocketPipe() says:

> No buffering occurs -- a message send does not complete until a corresponding receive accepts the message.

...which I interpreted as suggesting that the version you describe could deadlock. Is this relying on undocumented implementation details?

Really what I want here is to be able to join two promises and reject if either of them rejects, and let the event loop handle the scheduling -- the TaskSet with the error handler that just rethrows was ultimately a verbose way of getting that, since I couldn't find something relevant elsewhere in the library.

zenhack:
@kentonv, I believe I've addressed all of your comments, except the one I replied to. I found one spot in the Makefiles that seemed like it made sense to add the header, but I don't fully understand the non-ekam build system so if I've missed something let me know.

c++/src/capnp/compat/websocket-rpc-test.c++
kentonv: > it wasn't obvious to me from the docs that you can rely on the write end resolving if nothing is waiting on the read end

The write end won't resolve until a read is performed on the other end. That's why my code creates the write promise but doesn't wait on it, then performs the read, and only after that verifies that the write completed.

> Really what I want here is to be able to join two promises and reject if either of them rejects

Admittedly if the write rejected immediately (without actually writing) then this code would end up deadlocking waiting for the read end, and you'd have to poke the code a bit to figure out what happened. I've been meaning to add a way to join promises with nice syntax and fail-fast behavior... but haven't gotten around to it yet.

If you were really worried about it, you could check `readPromise.poll()` to verify the read actually completes before you `wait()` it -- and if it doesn't complete, then go back and check on `writePromise`. But I don't think it's worth it.

kentonv:
If we want to build this in CI then you'll need to add the `.c++` file to the makefiles somewhere. I guess it'll have to be Yet Another Library since it depends on KJ-HTTP.  You can follow the pattern for `capnp-json` to create `capnp-websocket`. I think you'll need to update `c++/Makefile.am` and `c++/src/capnp/CMakeLists.txt`.

zenhack:
Ok, after a few iterations it's building in CI.

kentonv:
Sorry for the delay. Looks good.

I notice your new files don't have copyright notices at the top, which feels weird since everything else does. Want to copy the notice used in other files, but change the first line to "Copyright 2021 Ian Denhardt and contributors"?

zenhack:
Quoting Kenton Varda (2021-04-19 16:47:29)

> I notice your new files don't have copyright notices at the top, which
> feels weird since everything else does. Want to copy the notice used
> in other files, but change the first line to "Copyright 2021 Ian
> Denhardt and contributors"?

Done.

c++/src/capnp/compat/websocket-rpc-test.c++
```diff
@@ -0,0 +1,59 @@
+
```
kentonv, line 1: Missed this file.

zenhack:
Quoting Kenton Varda (2021-04-19 19:34:28)

> Missed this file.

Fixed.

c++/src/capnp/compat/websocket-rpc.c++
```diff
@@ -0,0 +1,128 @@
...
+namespace capnp {
+
+WebSocketMessageStream::WebSocketMessageStream(kj::WebSocket& socket)
+  : socket(socket)
+  {};
+
+kj::Promise<kj::Maybe<MessageReaderAndFds>> WebSocketMessageStream::tryReadMessage(
+    kj::ArrayPtr<kj::AutoCloseFd> fdSpace,
+    ReaderOptions options, kj::ArrayPtr<word> scratchSpace) {
+  return socket.receive(options.traversalLimitInWords)
```
kentonv, line 35: Whoops, just noticed this...

```suggestion
  return socket.receive(options.traversalLimitInWords * sizeof(word))
```

zenhack:
Quoting Kenton Varda (2021-04-19 21:54:33)

>    Whoops, just noticed this...
>    �� Suggested change
> -  return socket.receive(options.traversalLimitInWords)
> +  return socket.receive(options.traversalLimitInWords * sizeof(word))

Good catch, fixed.

alaindargelas:
You introduced a dependency to libssl and other libs that comes to a surprise, breaking my build:
https://github.com/alainmarcel/UHDM

with errors like:
```
-- Found OpenSSL: /usr/lib/x86_64-linux-gnu/libcrypto.so (found version "1.1.1f") found components: Crypto SSL 
-- Found ZLIB: /usr/lib/x86_64-linux-gnu/libz.so (found version "1.2.11") 
CMake Error: File /home/alain/UHDM/third_party/capnproto/c++/pkgconfig/capnp-websocket.pc.in does not exist.
CMake Error at third_party/capnproto/c++/CMakeLists.txt:172 (configure_file):
  configure_file Problem configuring file


-- Configuring incomplete, errors occurred!
See also "/home/alain/UHDM/build/CMakeFiles/CMakeOutput.log".
make: *** [Makefile:45: build] Error 1
```

How can I proceed?

zenhack:
There's no new dependency (the openssl dependency was already there, and
it looks like your build found it just fine), but it looks like I missed
adding a pkg-config file for the new lib -- apparently CI doesn't catch
this.

Quoting alaindargelas (2021-04-22 02:13:19)

kentonv:
@alaindargelas I really recommend against pulling dependencies directly from their master branch. If you want to track git (rather than use a release version), at the very least you need to pin to a known-good commit ID, and update that pin periodically on your own schedule. If an update fails, you can hold back to the previous known-good commit until the issue is fixed. This is also important so that you can see what version of a dependency each past version of your own code was built against.

My own projects that use Cap'n Proto don't rely on master being always-working. Here's how Sandstorm manages dependencies, for example: https://github.com/sandstorm-io/sandstorm/tree/master/deps

I then update dependencies once a month with `make update-deps`:

https://github.com/sandstorm-io/sandstorm/blob/981ffa0dc4c88d1ed53da0a5f2560eaf416ceebd/Makefile#L206-L213

The Cloudflare Workers build (not open source) does the same thing, updating weekly.

This does appear to a bug and we will fix it (thanks for reporting!), but you should not expect any urgency from our end.

alaindargelas:
Ssl was also a problem and I had to use this statement to get around, then found the 2nd error
set(WITH_OPENSSL OFF CACHE BOOL "No ssl")  <<<
add_subdirectory(third_party/capnproto

alaindargelas:
I'll try your recommendation for pulling a stable version.

PR #1205, opened 2021-04-07
Fix TSAN unhappiness

vlovich:
Under a certain workload TSAN will complain about ReadLimiter which is a
false-positive signal per the class comments. Switch to relaxed atomics
because they should make TSAN happy while not altering the performance
characteristics of what we were doing.

### DIFF
--- a/c++/src/capnp/arena.h
+++ b/c++/src/capnp/arena.h
@@ -91,11 +91,28 @@ public:
 
 private:
   volatile uint64_t limit;
-  // Current limit, decremented each time catRead() is called.  Volatile because multiple threads
-  // could be trying to modify it at once.  (This is not real thread-safety, but good enough for
-  // the purpose of this class.  See class comment.)
+  // Current limit, decremented each time catRead() is called. We modify this variable using atomics
+  // with "relaxed" thread safety to make TSAN happy (on ARM & x86 this is no different from a
+  // regular read/write of the variable). See the class comment for why this is OK (previously we
+  // used a regular volatile variable - this is just to make ASAN happy).
 
   KJ_DISALLOW_COPY(ReadLimiter);
+
+  KJ_ALWAYS_INLINE(void setLimit(uint64_t newLimit)) {
+#if defined(__GNUC__) || defined(__clang__)
+    __atomic_store_n(&limit, newLimit, __ATOMIC_RELAXED);
+#else
+    limit = newLimit;
+#endif
+  }
+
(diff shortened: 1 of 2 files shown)


c++/src/capnp/arena.h
```diff
@@ -91,11 +99,35 @@ class ReadLimiter {
...
+  // previously we used a volatile variable).
 
   KJ_DISALLOW_COPY(ReadLimiter);
+
+  KJ_ALWAYS_INLINE(void setLimit(uint64_t newLimit)) {
+#if _MSC_VER && !defined(__clang__)
+#if _M_ARM
+    _InterlockedExchange64_nf(reinterpret_cast<volatile int64_t*>(&limit), newLimit);
+#else
+    _InterlockedExchange64(reinterpret_cast<volatile int64_t*>(&limit), newLimit);
```
kentonv, line 113: `_InterlockedExchange64` does an atomic exchange, not just a store, so is much more expensive than a store. (The function returns the old value of the target.)

Can we just leave it as a regular assignment on MSVC?

c++/src/capnp/arena.h
```diff
@@ -91,11 +99,35 @@ class ReadLimiter {
...
+    __atomic_store_n(&limit, newLimit, __ATOMIC_RELAXED);
+#endif
+  }
+
+  KJ_ALWAYS_INLINE(uint64_t readLimit() const) {
+#if _MSC_VER && !defined(__clang__)
+#if _M_ARM
+    return _InterlockedOr64_nf(reinterpret_cast<volatile int64_t*>(&limit), 0);
+#else
+    return _InterlockedOr64(reinterpret_cast<volatile int64_t*>(&limit), 0);
```
kentonv, line 125: Similarly, `_InterlockedOr64` actually stores the result back to the target variable, making this a much more expensive atomic operation than a simple load.

kentonv:
FWIW I would guess that the reason there are no "Interlocked" load or store operations is because any load or store on an integer is already atomic by default on all platforms MSVC targets. The MSVC team probably figured that making atomics for these was pointless. The GCC people decided to be consistent and make atomics for these operations even though they compile down to a simple load or store.

Keep in mind our only goal here is to make TSAN happy. The reason using atomics on GCC/Clang makes TSAN happy is because it tells TSAN that we're aware that the variable is being accessed by multiple threads and we're OK with the data race in this case. TSAN is not flagging any real problem, it is just flagging something where it looks like there's a problem and it is unaware that we've already considered the problem and decided it's fine.

Issue #1209, opened 2021-04-10
Fresh build of master on VS 2017  cannot convert from 'void *' to 'kj::_::XThreadPaf *'


ScatteredRay:
Compiling latest I get
`..\c++\src\kj\async.c++(1150):` error C2440: '=': cannot convert from 'void *' to 'kj::_::XThreadPaf *'`

with a terminal with git, cmake, and visual studio 2017 command line tools I was able to get it like so:

`
git clone https://github.com/capnproto/capnproto.git
cd .\capnproto\
mkdir build
cd .\build\
cmake -GNinja ..
ninja
`

kentonv:
Seems to be passing in CI, which uses VS 2019: https://github.com/capnproto/capnproto/runs/2292060151?check_suite_focus=true

We used to test VS 2017 in CI but currently the build that claims to be testing 2017 is actually using MinGW by accident. There's a PR to fix it: https://github.com/capnproto/capnproto/pull/1025 However, as mentioned in that PR, we can't seem to get that build working. I don't have time.

I'm happy to accept a PR that fixes our VS 2017 build, in which case we'll continue to try to keep things building on 2017. Otherwise, I think right now we're likely to drop support for VS 2017 and support only VS 2019. We just don't have any way to debug issues with VS 2017 so it's been impractical for us to deal with... :(

ScatteredRay:
I'll give the CI a look.

ScatteredRay:
Ok, I made a clone locally, and tried a similar change: https://github.com/TelltaleGames/capnproto/commit/44554281df51546344f36c75593f231b0147df6b

I noticed for your change in #1024 you were missing `-A x64`

Running the CI on my branch also failed: https://github.com/TelltaleGames/capnproto/runs/2314616524?check_suite_focus=true

We seem to get test failures: e.g. `Could not find executable D:/a/capnproto/capnproto/build-output/src/kj/Debug/kj-heavy-tests.exe`

But then when we scroll back through the errors, we see the one I posted:
`  Building Custom Rule D:/a/capnproto/capnproto/c++/src/kj/CMakeLists.txt
  async.c++
D:\a\capnproto\capnproto\c++\src\kj\async.c++(1150): error C2440: '=': cannot convert from 'void *' to 'kj::_::XThreadPaf *' [D:\a\capnproto\capnproto\build-output\src\kj\kj-async.vcxproj]
[...]

PR #1210, opened 2021-04-10
Fix for VS2017 Builds and CI.

ScatteredRay:
Fix for #1209, and also CI issue identified in #1025.

VS2017 apparently doesn't properly coerce from nullptr, and it was difficult to solve this in the __atomic_exchange_n macro since the VS2017 template engine didn't seem capable of template deduction either.


kentonv:
Huh, I guess whatever segfault issue I had been seeing before was fixed in the meantime. Cool.

Thanks for figuring this out!

PR #1217, opened 2021-04-22
Prohibit assigning to a StringPtr from a String rvalue

xortive:
This keeps you from doing things like

```
struct CopyableConfigStruct {
   kj::StringPtr someStr;
}
// .. lots of code in between, potentially a different file
config.someStr = kj::str("hello", "world");
```


c++/src/kj/string.h
```diff
@@ -75,6 +75,10 @@ class StringPtr {
   }
   inline StringPtr(const char* begin, const char* end): StringPtr(begin, end - begin) {}
   inline StringPtr(const String& value);
+  void operator=(String&& value) = delete;
+  inline void operator=(decltype(nullptr)) {
```
xortive, line 79: required to resolve an operator ambiguity since `String` can be implicitly constructed from `nullptr`, and we actually do want to allow `StringPtr s = nullptr`

xortive:
I wonder if we could do this for ArrayPtr too

kentonv:
> I wonder if we could do this for ArrayPtr too

Yes I think that would be a good idea, want to add it to this PR?

Also happy to merge this as-is.

xortive:
> > I wonder if we could do this for ArrayPtr too
> 
> Yes I think that would be a good idea, want to add it to this PR?
> 
> Also happy to merge this as-is.

I took a look but not sure how to do it -- It looks like `Array` -> `ArrayPtr` is implemented in `Array` as a cast operator

kentonv:
I suspect you can still define `operator=(Array<T>&&) = delete` on `ArrayPtr`, and C++ will prefer that over the cast operators. Try it and see what happens...

xortive:
Sweet yeah that worked, needs the nullptr_t specialization to avoid ambiguity just like the String one.

c++/src/kj/common.h
```diff
@@ -1396,6 +1396,11 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
   inline constexpr ArrayPtr(decltype(nullptr)): ptr(nullptr), size_(0) {}
   inline constexpr ArrayPtr(T* ptr, size_t size): ptr(ptr), size_(size) {}
   inline constexpr ArrayPtr(T* begin, T* end): ptr(begin), size_(end - begin) {}
+  ArrayPtr<T> operator=(Array<T>&&) = delete;
+  ArrayPtr<T> operator=(nullptr_t) {
+    ptr = nullptr;
+    size_ = 0;
+  }
```
kentonv, line 1403: Missing return statement.

Note that by convention, `operator=` should always return a reference to `*this`, so the return type here needs to be `ArrayPtr<T>&`.

In `StringPtr`, you had `operator=` returning `void`, but that's not correct either -- it should return `StringPtr&`.

c++/src/kj/common.h
```diff
@@ -1396,6 +1396,12 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
   inline constexpr ArrayPtr(decltype(nullptr)): ptr(nullptr), size_(0) {}
   inline constexpr ArrayPtr(T* ptr, size_t size): ptr(ptr), size_(size) {}
   inline constexpr ArrayPtr(T* begin, T* end): ptr(begin), size_(end - begin) {}
+  ArrayPtr<T>& operator=(Array<T>&&) = delete;
+  ArrayPtr<T>& operator=(nullptr_t) {
```
kentonv, line 1400: `nullptr_t` is a member of the `std` namespace and normally not defined if you haven't included the right header. You will need to write `decltype(nullptr)` instead (which is what `nullptr_t` is typedef'd as in the std headers).

I'm a little surprised this compiles on GCC and MSVC...

PR #1220, opened 2021-04-26
Add MappedIterator/MappedIterable to common.h

xortive:


kentonv:
This just moves code, right? No semantic changes?

I'm a little concerned about bloat in `common.h`, since literally everyone has to include that header. But I'm not sure if there's any better place. At some point it might be necessary to split `common.h` into several headers and fix everything else to include what it actually uses...

xortive:
> This just moves code, right? No semantic changes?

Yes, no changes (aside from removing forward declarations and `_::` as necessary in `table.h`)

> I'm a little concerned about bloat in `common.h`, since literally everyone has to include that header. But I'm not sure if there's any better place. At some point it might be necessary to split `common.h` into several headers and fix everything else to include what it actually uses...

Yep :( I was thinking of making an `iterator.h`, but then I remembered the other fake containers/iterators were in `common.h`

PR #1221, opened 2021-04-26
Add pkgconfig/capnp-websocket.pc.in

harrishancock:
I didn't test the new capnp-websocket.pc.in, just copied and modified capnp-json.pc.in. Note that the CMake build enumerates over all its targets to find .pc.in files, meaning adding a target without a corresponding .pc.in file breaks the CMake build, which is how I noticed.

### DIFF
--- a/c++/configure.ac
+++ b/c++/configure.ac
@@ -120,6 +120,7 @@ AC_DEFUN([CAPNP_PKG_CONFIG_FILES], [ \
   pkgconfig/capnpc.pc \
   pkgconfig/capnp-rpc.pc \
   pkgconfig/capnp-json.pc \
+  pkgconfig/capnp-websocket.pc \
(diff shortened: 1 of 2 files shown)


harrishancock:
Derp, I didn't add the regular .pc file. Hang on...

kentonv:
FWIW, this came up in the thread for #1204.

Something I noticed is that the MinGW CI was failing on this -- but not actually reporting the build as having failed. So it was just skipping the build and reporting success. :(

cc @zenhack

harrishancock:
> Derp, I didn't add the regular .pc file. Hang on...

Ah, last night I hadn't realized these were also autogenerated in the autotools build.

kentonv:
Whoops, I read your "Hang on...", but then I came back to this thread again and forgot about it and went ahead and merged.

PR #1225, opened 2021-04-28
add hashcode implementation for enums

xortive:
I assumed we wanted this done with compiler intrinsics rather than using `type_traits`?


kentonv:
Yes, that's perfect, thanks.

PR #1228, opened 2021-04-30
Mutex tracking

vlovich:
This adds two tracking operations:

* Async signal safe TLS variables to track when a thread is blocked on a lock primitive.
* Per-mutex contextual information pointing to the locked mutex.

These should be sufficient for deadlock detection with the notable exception that I don't know what to do about shared mutexes. By their very nature, I can only track one random acquisition point. If a shared mutex is part of a deadlock chain, it can't be found. To try to track that data seems cost-prohibitive: needs some kind of array per mutex & maybe hope that lock acquisition/release can be reflected with atomics (otherwise it also implies an acquisition of an exclusive mutex to record the TID holding the shared lock).

I added a `kj::SourceLocation` object when building against c++20 since all compilers support it but not all STL provide `source_location` yet (+ more convenient when everything is a `kj::StringPtr`).

Putting up the PR to get some early feedback on what to do about shared mutexes.

### DIFF
--- a/.github/workflows/quick-test.yml
+++ b/.github/workflows/quick-test.yml
@@ -25,6 +25,23 @@ jobs:
       - name: super-test
         run: |
             ./super-test.sh quick ${{ matrix.compiler }}
+  Linux-lock-tracking:
+    runs-on: ubuntu-18.04
+    strategy:
+      fail-fast: false
+      matrix:
+        compiler: [clang-9]
+        features: ["-DKJ_TRACK_LOCK_BLOCKING=1 -DKJ_SAVE_ACQUIRED_LOCK_INFO=1"]
+    steps:
+      - uses: actions/checkout@v2
+      - name: install dependencies
+        run: |
+            export DEBIAN_FRONTEND=noninteractive
+            sudo apt-get install -y build-essential git zlib1g-dev cmake libssl-dev ${{ matrix.compiler }}
+      - name: super-test
+        run: |
+            # librt is used for timer_create in the unit tests for lock tracking (mutex-test.c++).
+            ./super-test.sh quick ${{ matrix.compiler }} cpp-features "${{matrix.features}}" extra-libs "-lrt"
   MacOS:
     runs-on: macos-latest
     strategy:
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -214,6 +214,28 @@ typedef unsigned char byte;
 #define KJ_UNUSED_MEMBER
 #endif
 
+#if __cplusplus > 201703L || (__clang__  && __clang_major__ >= 9 && __cplusplus >= 201103L)
+// Technically this was only added to C++20 but Clang allows it for >= C++11 and spelunking the
+// attributes manual indicates it first came in with Clang 9.
+#define KJ_NO_UNIQUE_ADDRESS [[no_unique_address]]
+#else
+#define KJ_NO_UNIQUE_ADDRESS
+#endif
+
+#if defined(__has_feature)
+#if __has_feature(thread_sanitizer)
+#define KJ_SANITIZE_THREAD 1
+#endif
+#elif __SANITIZE_THREAD__
+#define KJ_SANITIZE_THREAD 1
+#endif
+
+#if KJ_SANITIZE_THREAD
+#define KJ_DISABLE_TSAN __attribute__((no_sanitize("thread"), noinline))
+#else
+#define KJ_DISABLE_TSAN
+#endif
+
 #if __clang__
 #define KJ_DEPRECATED(reason) \
     __attribute__((deprecated(reason)))
--- a/c++/src/kj/mutex.c++
+++ b/c++/src/kj/mutex.c++
@@ -139,7 +139,7 @@ Mutex::~Mutex() {
   KJ_ASSERT(futex == 0, "Mutex destroyed while locked.") { break; }
 }
 
-bool Mutex::lock(Exclusivity exclusivity, Maybe<Duration> timeout) {
+bool Mutex::lock(Exclusivity exclusivity, Maybe<Duration> timeout, LockSourceLocationArg location) {
   auto spec = timeout.map([](Duration d) { return toRelativeTimespec(d); });
   struct timespec* specp = nullptr;
   KJ_IF_MAYBE(s, spec) {
@@ -315,7 +315,7 @@ void Mutex::assertLockedByCaller(Exclusivity exclusivity) {
   }
 }
 
-void Mutex::wait(Predicate& predicate, Maybe<Duration> timeout) {
+void Mutex::wait(Predicate& predicate, Maybe<Duration> timeout, const LockSourceLocation& location) {
   // Add waiter to list.
   Waiter waiter { nullptr, waitersTail, predicate, nullptr, 0, timeout != nullptr };
   addWaiter(waiter);
@@ -362,7 +362,7 @@ void Mutex::wait(Predicate& predicate, Maybe<Duration> timeout) {
                                           __ATOMIC_ACQUIRE, __ATOMIC_ACQUIRE)) {
             // OK, we set our own futex to 1. That means no other thread will, and so we won't be
             // receiving a mutex ownership transfer. We have to lock the mutex ourselves.
-            lock(EXCLUSIVE);
+            lock(EXCLUSIVE, nullptr, location);
             currentlyLocked = true;
             return;
           } else {
@@ -412,7 +412,7 @@ void Mutex::induceSpuriousWakeupForTest() {
   }
 }
 
-void Once::runOnce(Initializer& init) {
+void Once::runOnce(Initializer& init, const LockSourceLocation& location) {
 startOver:
   uint state = UNINITIALIZED;
   if (__atomic_compare_exchange_n(&futex, &state, INITIALIZING, false,
@@ -486,7 +486,7 @@ Mutex::Mutex() {
 }
 Mutex::~Mutex() {}
 
-bool Mutex::lock(Exclusivity exclusivity, Maybe<Duration> timeout) {
+bool Mutex::lock(Exclusivity exclusivity, Maybe<Duration> timeout, NoopSourceLocation) {
   if (timeout != nullptr) {
     KJ_UNIMPLEMENTED("Locking a mutex with a timeout is only supported on Linux.");
   }
@@ -555,7 +555,7 @@ void Mutex::assertLockedByCaller(Exclusivity exclusivity) {
   // held for debug purposes anyway, we just don't bother.
 }
 
-void Mutex::wait(Predicate& predicate, Maybe<Duration> timeout) {
+void Mutex::wait(Predicate& predicate, Maybe<Duration> timeout, NoopSourceLocation) {
   // Add waiter to list.
   Waiter waiter { nullptr, waitersTail, predicate, nullptr, 0 };
   static_assert(sizeof(waiter.condvar) == sizeof(CONDITION_VARIABLE),
@@ -662,7 +662,7 @@ Once::Once(bool startInitialized) {
 }
 Once::~Once() {}
 
-void Once::runOnce(Initializer& init) {
+void Once::runOnce(Initializer& init, NoopSourceLocation) {
   BOOL needInit;
   while (!InitOnceBeginInitialize(&coercedInitOnce, 0, &needInit, nullptr)) {
     // Init was occurring in another thread, but then failed with an exception. Retry.
@@ -712,7 +712,7 @@ Mutex::~Mutex() {
   KJ_PTHREAD_CLEANUP(pthread_rwlock_destroy(&mutex));
 }
 
-bool Mutex::lock(Exclusivity exclusivity, Maybe<Duration> timeout) {
+bool Mutex::lock(Exclusivity exclusivity, Maybe<Duration> timeout, NoopSourceLocation) {
   if (timeout != nullptr) {
     KJ_UNIMPLEMENTED("Locking a mutex with a timeout is only supported on Linux.");
   }
@@ -781,7 +781,7 @@ void Mutex::assertLockedByCaller(Exclusivity exclusivity) {
   }
 }
 
-void Mutex::wait(Predicate& predicate, Maybe<Duration> timeout) {
+void Mutex::wait(Predicate& predicate, Maybe<Duration> timeout, NoopSourceLocation) {
   // Add waiter to list.
   Waiter waiter {
     nullptr, waitersTail, predicate, nullptr,
@@ -901,7 +901,7 @@ Once::~Once() {
   KJ_PTHREAD_CLEANUP(pthread_mutex_destroy(&mutex));
 }
 
-void Once::runOnce(Initializer& init) {
+void Once::runOnce(Initializer& init, NoopSourceLocation) {
   KJ_PTHREAD_CALL(pthread_mutex_lock(&mutex));
   KJ_DEFER(KJ_PTHREAD_CALL(pthread_mutex_unlock(&mutex)));
 
--- a/c++/src/kj/mutex.c++
+++ b/c++/src/kj/mutex.c++
@@ -179,6 +179,34 @@ struct timespec toAbsoluteTimespec(TimePoint time) {
 // =======================================================================================
 // Futex-based implementation (Linux-only)
 
+#if KJ_SAVE_ACQUIRED_LOCK_INFO
+#if !__GLIBC_PREREQ(2, 30)
+#ifndef SYS_gettid
+#error SYS_gettid is unavailable on this system
+#endif
+
+#define gettid() ((pid_t)syscall(SYS_gettid))
+#endif
+
+static thread_local pid_t tlsTid = gettid();
+#define TRACK_ACQUIRED_TID() tlsTid
+
+Mutex::AcquiredMetadata Mutex::lockedInfo() const {
+  auto state = __atomic_load_n(&futex, __ATOMIC_RELAXED);
+  auto tid = lockedExclusivelyByThread;
+  auto location = lockAcquiredLocation;
+
+  if (state & EXCLUSIVE_HELD) {
+    return HoldingExclusively{tid, location};
+  } else {
+    return HoldingShared{location};
+  }
+}
+
+#else
+#define TRACK_ACQUIRED_TID() 0
+#endif
+
 Mutex::Mutex(): futex(0) {}
 Mutex::~Mutex() {
   // This will crash anyway, might as well crash with a nice error message.
@@ -201,6 +229,7 @@ bool Mutex::lock(Exclusivity exclusivity, Maybe<Duration> timeout, LockSourceLoc
         uint state = 0;
         if (KJ_LIKELY(__atomic_compare_exchange_n(&futex, &state, EXCLUSIVE_HELD, false,
                                                   __ATOMIC_ACQUIRE, __ATOMIC_RELAXED))) {
+
           // Acquired.
           break;
         }
@@ -228,9 +257,11 @@ bool Mutex::lock(Exclusivity exclusivity, Maybe<Duration> timeout, LockSourceLoc
           }
         }
       }
+      acquiredExclusive(TRACK_ACQUIRED_TID(), location);
       break;
     case SHARED: {
       uint state = __atomic_add_fetch(&futex, 1, __ATOMIC_ACQUIRE);
+
       for (;;) {
         if (KJ_LIKELY((state & EXCLUSIVE_HELD) == 0)) {
           // Acquired.
@@ -263,6 +294,19 @@ bool Mutex::lock(Exclusivity exclusivity, Maybe<Duration> timeout, LockSourceLoc
         }
         state = __atomic_load_n(&futex, __ATOMIC_ACQUIRE);
       }
+
+      // We just want to record the lock being acquired somewhere but the specific location doesn't
+      // matter. This does mean that race conditions could occur where a thread might read this
+      // inconsistently (e.g. filename from 1 lock & function from another). This currently is just
+      // meant to be a debugging aid for manual analysis so it's OK for that purpose. If it's ever
+      // required for this to be used for anything else, then this should probably be changed to
+      // use an additional atomic variable that can ensure only one writer updates this. Or use the
+      // futex variable to ensure that this is only done for the first one to acquire the lock,
+      // although there may be thundering herd problems with that whereby there's a long wallclock
+      // time between when the lock is acquired and when the location is updated (since the first
+      // locker isn't really guaranteed to be the first one unlocked).
+      acquiredShared(location);
+
       break;
     }
   }
@@ -273,6 +317,10 @@ void Mutex::unlock(Exclusivity exclusivity, Waiter* waiterToSkip) {
   switch (exclusivity) {
     case EXCLUSIVE: {
       KJ_DASSERT(futex & EXCLUSIVE_HELD, "Unlocked a mutex that wasn't locked.");
+
+#ifdef KJ_CONTENTION_WARNING_THRESHOLD
+      auto acquiredLocation = releasingExclusive();
+#endif
 
       // First check if there are any conditional waiters. Note we only do this when unlocking an
       // exclusive lock since under a shared lock the state couldn't have changed.
@@ -331,7 +379,7 @@ void Mutex::unlock(Exclusivity exclusivity, Waiter* waiterToSkip) {
         uint readerCount = oldState & SHARED_COUNT_MASK;
         if (readerCount >= KJ_CONTENTION_WARNING_THRESHOLD) {
           KJ_LOG(WARNING, "excessively many readers were waiting on this lock", readerCount,
-              kj::getStackTrace());
+              kj::getStackTrace(), acquiredLocation);
         }
 #endif
       }
@@ -370,7 +418,7 @@ void Mutex::assertLockedByCaller(Exclusivity exclusivity) {
   }
 }
 
-void Mutex::wait(Predicate& predicate, Maybe<Duration> timeout, const LockSourceLocation& location) {
+void Mutex::wait(Predicate& predicate, Maybe<Duration> timeout, LockSourceLocationArg location) {
   // Add waiter to list.
   Waiter waiter { nullptr, waitersTail, predicate, nullptr, 0, timeout != nullptr };
   addWaiter(waiter);
@@ -382,7 +430,10 @@ void Mutex::wait(Predicate& predicate, Maybe<Duration> timeout, const LockSource
   // currently.
   bool currentlyLocked = true;
   KJ_DEFER({
-    if (!currentlyLocked) lock(EXCLUSIVE);
+    // Infinite timeout for re-obtaining the lock is on purpose because the post-condition for this
+    // function has to be that the lock state hasn't changed (& we have to be locked when we enter
+    // since that's how condvars work).
+    if (!currentlyLocked) lock(EXCLUSIVE, nullptr, location);
     removeWaiter(waiter);
   });
 
@@ -475,7 +526,7 @@ void Mutex::induceSpuriousWakeupForTest() {
   }
 }
 
-void Once::runOnce(Initializer& init, const LockSourceLocation& location) {
+void Once::runOnce(Initializer& init, LockSourceLocationArg location) {
 startOver:
   uint state = UNINITIALIZED;
   if (__atomic_compare_exchange_n(&futex, &state, INITIALIZING, false,
@@ -860,7 +911,7 @@ void Mutex::wait(Predicate& predicate, Maybe<Duration> timeout, NoopSourceLocati
   // currently.
   bool currentlyLocked = true;
   KJ_DEFER({
-    if (!currentlyLocked) lock(EXCLUSIVE, nullptr);
+    if (!currentlyLocked) lock(EXCLUSIVE, nullptr, NoopSourceLocation::current());
     removeWaiter(waiter);
 
     // Destroy pthread objects.
@@ -928,7 +979,7 @@ void Mutex::wait(Predicate& predicate, Maybe<Duration> timeout, NoopSourceLocati
     // because we've already been signaled.
     KJ_PTHREAD_CALL(pthread_mutex_unlock(&waiter.stupidMutex));
 
-    lock(EXCLUSIVE);
+    lock(EXCLUSIVE, nullptr, NoopSourceLocation::current());
     currentlyLocked = true;
 
     KJ_IF_MAYBE(exception, waiter.exception) {
--- a/c++/src/kj/mutex.c++
+++ b/c++/src/kj/mutex.c++
@@ -59,6 +59,52 @@
 #endif
 
 namespace kj {
+#if KJ_TRACK_LOCK_BLOCKING
+static thread_local const BlockedOnReason* tlsBlockReason __attribute((tls_model("initial-exec")));
+// The initial-exec model ensures that even if this code is part of a shared library built PIC, then
+// we still place this variable in the appropriate ELF section so that __tls_get_addr is avoided.
+// It's unclear if __tls_get_addr is still not async signal safe in glibc. The only negative
+// downside of this approach is that a shared library built with kj & lock tracking will fail if
+// dlopen'ed which isn't an intended use-case for the initial implementation.
+
+Maybe<const BlockedOnReason&> blockedReason() noexcept {
+  if (tlsBlockReason == nullptr) {
+    return nullptr;
+  }
+  return *tlsBlockReason;
+}
+
+static void setCurrentThreadIsWaitingFor(const BlockedOnReason* meta) {
+  tlsBlockReason = meta;
+}
+
+static void setCurrentThreadIsNoLongerWaiting() {
+  tlsBlockReason = nullptr;
+}
+#elif KJ_USE_FUTEX
+struct BlockedOnMutexAcquisition {
+  constexpr BlockedOnMutexAcquisition(const _::Mutex& mutex, LockSourceLocationArg) {}
+};
+
+struct BlockedOnCondVarWait {
+  constexpr BlockedOnCondVarWait(const _::Mutex& mutex, const void *waiter,
+      LockSourceLocationArg) {}
+};
+
+struct BlockedOnOnceInit {
+  constexpr BlockedOnOnceInit(const _::Once& once, LockSourceLocationArg) {}
+};
+
+struct BlockedOnReason {
+  constexpr BlockedOnReason(const BlockedOnMutexAcquisition&) {}
+  constexpr BlockedOnReason(const BlockedOnCondVarWait&) {}
+  constexpr BlockedOnReason(const BlockedOnOnceInit&) {}
+};
+
+static void setCurrentThreadIsWaitingFor(const BlockedOnReason* meta) {}
+static void setCurrentThreadIsNoLongerWaiting() {}
+#endif
+
 namespace _ {  // private
 
 #if KJ_USE_FUTEX
@@ -140,6 +186,9 @@ Mutex::~Mutex() {
 }
 
 bool Mutex::lock(Exclusivity exclusivity, Maybe<Duration> timeout, LockSourceLocationArg location) {
+  BlockedOnReason blockReason = BlockedOnMutexAcquisition{*this, location};
+  KJ_DEFER(setCurrentThreadIsNoLongerWaiting());
+
   auto spec = timeout.map([](Duration d) { return toRelativeTimespec(d); });
   struct timespec* specp = nullptr;
   KJ_IF_MAYBE(s, spec) {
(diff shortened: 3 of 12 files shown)


c++/src/kj/mutex.c++
```diff
@@ -58,7 +59,100 @@
...
+      auto origin = reinterpret_cast<const SourceLocation*>(pthread_getspecific(waitingOriginKey));
+      KJ_DASSERT(origin != nullptr);
+      auto o = reinterpret_cast<const _::Once*>(pthread_getspecific(waitingObjectKey));
+      return BlockOnReason{BlockOnOnceInit{.once = o, .origin = origin}};
+    }
+  }
+}
+
+static void setCurrentThreadIsWaitingFor(const _::Mutex* mutex, const SourceLocation* origin) {
+  pthread_setspecific(waitingTypeKey, reinterpret_cast<void *>(BlockState::Dirty));
```
kentonv, line 121: You mention the importance of async signal safety in the commit description, but setting four variables one by one doesn't seem very signal safe. I guess that's the purpose of the "dirty" state here, so the signal handler can see that this was currently being modified.

Another solution would be to have a single thread-local pointer which points to a struct allocated on the stack. That way, updating the pointer is atomic. It does require that this info is always scoped to a stack frame, but `origin` already is so-scoped, so that seems fine.

On another note, it would be nice to use RAII to set and unset these values, to avoid any possibility of forgetting to unset the state on the way out.

c++/src/kj/mutex.c++
```diff
@@ -58,7 +59,100 @@
...
 #endif
 
+#if KJ_TRACK_LOCK_BLOCKING
+#include <pthread.h>
+#endif
+
 namespace kj {
+
+#if KJ_TRACK_LOCK_BLOCKING
+static pthread_key_t waitingTypeKey;
```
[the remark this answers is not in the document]
vlovich: I agree I too would strongly prefer thread_local. I've read too much conflicting information that I don't think there's a clear answer & basically amounts to "It depends. YOLO". If TLS space is lazily allocated, then you may need to acquire a lock to allocate the storage. Some times it depends what libc you link against. What I read seemed to indicate that glibc's `pthread_getspecific` specifically is async-signal safe which is why I went this route.

c++/src/kj/source-location.h
```diff
@@ -0,0 +1,75 @@
...
+      uint32_t line = __builtin_LINE(), uint32_t column = KJ_CALLER_COLUMN()) {
+    // When invoked with no arguments returns a SourceLocation that points at the caller of the
+    // function.
+    return SourceLocation(file, function, line, column);
+  }
+
+  constexpr SourceLocation() : SourceLocation("??"_kj, "??"_kj, 0, 0) {}
+  // Constructs a dummy source location that's not pointing at anything.
+
+  constexpr kj::StringPtr file_name() const { return this->filename; }
```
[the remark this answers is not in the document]
vlovich: I was maintaining consistency with `std::source_location`.

c++/src/kj/mutex.h
```diff
@@ -162,7 +215,7 @@ class Once {
     virtual void run() = 0;
   };
 
-  void runOnce(Initializer& init);
+  void runOnce(Initializer& init KJ_LOCK_ACQUIRE_ADD_CALLER_LOCATION_ARG(location));
```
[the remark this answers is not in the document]
vlovich: I agree. My motivation was to  to have this feature be 0 cost when compiled out.

c++/src/kj/mutex.c++
```diff
@@ -19,6 +19,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
+#include <cstdint>
```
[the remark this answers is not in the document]
vlovich: Oversight. May have been auto-added by VSCode.

c++/src/kj/mutex.c++
vlovich: Yeah, what the hell was I thinking. I think at some point I had realized this & promptly forgot it?

I don't know that I can use RAII since these lifetimes aren't strictly tied to lexical scope (you could, but then you'll be superfluously saying you're not blocked for a period of time if you got a spurious wake & failed to acquire the lock)..

vlovich:
@kentonv did you have any thoughts on the reader mutex situation & how we can detect problematic locks/deadlocks there? Specifically SIGINT received while waiting on exclusive lock held by reader lock which is either taking too long. Or one of the threads holding the reader lock is waiting on an exclusive lock that you're holding. 

My preliminary thought is one of several approaches:

* Live with the current best-effort this PR will provide. If we have a SIGINT on a thread that's waiting on an exclusive lock that's currently held by 1 or more readers, we just blame a random location where the reader had been acquired (may not even be held anymore). Something like "this thread was waiting on an exclusive lock acquired at <file>:<line> at <function>. one reader lock location was %s:%s at %s" (since the reader lock location is "random", we don't want it part of the fingerprint).
* Have heavy-weight reader-lock tracking for a subset of mutex operations within EW only that keep track of all the threads that are holding the reader lock. This would bloat the size of a mutex & the cost of acquiring a reader lock would be a bit more pricey for doing the book-keeping. If we can't use a lock-free or wait-free data structure, we'll need to acquire an exclusive mutex around the book-keeping. Since within EW we know how many threads we will have, it's not the end of the world to just allocate 5000 counters per mutex indexed by thread idx that specifies the TID + count (the TID can be amortized if we have one such data structure locally that supports a finite set of mutexes).
* When we encounter this situation, do several stack trace of all threads over several ms & save all of them. Then post-process them to reduce them to the set of threads that are making forward progress. If we could somehow figure out how to have per-mutex TLS variables that are async-safe, we could even filter out the threads that aren't holding the relevant mutex (we have the mutex* & we just forward it along when requesting a stack trace of each thread which can then retrieve whether or not it's holding a lock).

Option 3 seems like the most robust but also the most R&D option. Option 1 feels kind of like the best MVP option to ship & then see if we need to invest in anything more fancy. Option 2 seems like a middle ground we could reasonably implement but would only capture the locks we explicitly control for.

c++/src/kj/mutex.c++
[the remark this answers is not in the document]
kentonv: https://gcc.godbolt.org/z/3dTYsWxWj

There's no lazy allocation. The compiler puts thread-local definitions into a special ELF binary section that is initialized by the binary loader. `thread_local` is pretty much ideal on ELF platforms.

It might be different on Mac, but this code is Linux-specific anyway, so I wouldn't worry about that.

c++/src/kj/mutex.c++
kentonv: Hmm if they're not tied to a lexical scope, how do you make sure the `origin` pointer stays valid?

It looks to me like `setCurrentTheradIsWaitingFor()`/`setCurrentThreadIsNoLongerWaiting()` are confined to a scope, though...

c++/src/kj/source-location.h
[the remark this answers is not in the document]
kentonv: Why?

c++/src/kj/mutex.h
```diff
@@ -172,6 +220,44 @@ class Mutex {
...
 #endif
 
+#if KJ_SAVE_ACQUIRED_LOCK_INFO
+  uint32_t lockedExclusivelyByThread = 0;
+  SourceLocation lockAcquiredLocation;
+
+  volatile bool updatedSharedMetadata = false;
+
+  void acquiredExclusive(uint32_t tid, const SourceLocation& location) noexcept {
+    __atomic_store_n(&lockedExclusivelyByThread, tid, __ATOMIC_RELEASE);
```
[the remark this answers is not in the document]
vlovich: My understanding is that the rough mapping for atomics is that read => acquire, write => release.Since this is a write operation, isn't release the right operation?

c++/src/kj/source-location.h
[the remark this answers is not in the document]
vlovich: No good reason. I'll fix it up.

c++/src/kj/mutex.c++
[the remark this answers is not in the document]
vlovich: I think that may not be true for code loaded from a shared library. So maybe we just document it as unsafe for dynamic linkage?

c++/src/kj/mutex.h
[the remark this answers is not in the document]
kentonv: No, load vs. store and acquire vs. release are separate things. Acquire-load and release-store are the two combinations that are "free" on x86, but that doesn't meant all loads should be acquires and all releases stores.

Acquire vs. release are about whether the "protected" data access happens before or after this operation. In this case, you perform this store before accessing the locked data. So if anything, it's an acquire.

That said, since this particular flag is used only for debugging, it might be worth using `__ATOMIC_RELAXED` instead, to avoid invoking the processor's locking protocol, which is expensive. A relaxed store is just a regular store, whereas an acquire-store requires instructing the processor to perform a locking instruction. The down side of using relaxed is that it might take longer for another core to pick up this change, but that's probably not a big deal in practice because we don't start trying to trace locks until they are held for a long time. And anyway, since you're not actually setting this value atomically at the same time as taking the lock, such a race condition exists either way.

So I'd go with `__ATOMIC_RELAXED` here.

For `releasingExclusive()` below, a release-store is "free", but I'd still probably change it to `__ATOMIC_RELAXED` for consistency.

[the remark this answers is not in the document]
Dr-Emann: I don't think acquire semantics have any meaning for stores, nor does release semantics have any meaning for a load.

The [documentation](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html) for `__atomic_store_n` says: The valid memory order variants are __ATOMIC_RELAXED, __ATOMIC_SEQ_CST, and __ATOMIC_RELEASE. There are similar rules for c++'s  `std::atomic_store_explicit` and `std::atomic<T>::store` (which specifies undefined behavior if it's not one of those three memory orderings).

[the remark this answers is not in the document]
kentonv: @Dr-Emann That's odd that the docs say that, when the compiler does in fact implement unique behavior for acquire-store that is different from any of the models that the document says are valid.

That said, it's true that acquire semantics on a pure-store operation is almost never useful. This is not because they inherently don't make sense, but rather because there's no way for another thread to correctly synchronize with it. The other thread would need some sort of "compare-and-store" operation, where it can atomically check the value of one location and, based on its value, conditionally store to another location. Generally, CPUs don't implement atomic ops that transact over multiple locations at once, as that would be incredibly complicated. But if they did, then acquire-store could theoretically make sense.

(This point is essentially the same point as I made above when I pointed out that using acquire semantics here wouldn't really accomplish much as the potential race condition still exists.)

However, that doesn't mean that "stores should always have release semantics". Acquire semantics make sense on atomic operations that are both load-and-store (e.g. atomic increment/decrement, compare-and-swap, exchange, etc.) -- and that's not just because "acquires are for loads", it's usually important in these cases that the "store" part has acquire semantics as well.

ANYWAY

Release semantics don't make sense here, because nothing is being released. Acquire semantics don't make sense, because they don't really solve any particular race condition. So relaxed is the way to go.

PS. Out of curiosity, @Dr-Emann, who are you and why do you read all our PR threads?

c++/src/kj/mutex.c++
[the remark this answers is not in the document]
kentonv: Hmm you are right! If I add `-fPIC` to the godbolt example, it starts calling a function `__tls_get_addr`. Luckily, `-fPIE` does not do this. It's also apparently possible to use `__attribute((tls_model("initial-exec")))` on the thread-local variable, which avoids the call to `__tls_get_addr` with the caveat that the library cannot be dynamically `dlopen()`ed at runtime; it must be loaded strictly at program start.

Apparently, `__tls_get_addr` itself _could_ be signal-safe -- [apparently it is in musl libc](https://sourceware.org/legacy-ml/libc-alpha/2013-09/msg00569.html). But it seems glibc's version has not always been signal-safe, and while I can find evidence of people wanting to fix that, I can't find any definitive answer as to whether they did.

For our purposes, we don't care about `-fPIC`. But, you could throw `__attribute((tls_model("initial-exec")))` on the thread-locals anyway, so that if someone does try to turn on this feature with `-fPIC`, it'll probably still work. If they try to `dlopen()` the library, then the'll get an error. This feature isn't on by default anyway, so that doesn't seem like a big deal.

kentonv:
Regarding reader locks, I think we should go with option 1. The other options seem to require excessive bookkeeping. I think if reader locks are a common source problems, it's more likely to be a starvation situation where no particular lock is held for long but  new threads keep coming in and taking the lock... in that case it probably won't be terribly helpful to see stack traces of all the readers.

c++/src/kj/mutex.h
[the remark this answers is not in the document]
Dr-Emann: I found capnproto looking for an RPC framework for work, and it was the only RPC system that really fit what I wanted (bidirectional data, especially typed "server" initiated RPC calls). I ended up getting pretty into it, it's a really elegant system. I set myself as a watcher for issues/PRs just to keep up with new features, in case there were any which looked so nice I'd upgrade before the next release, or what to look forward to in the next release (really excited for cross thread fulfillers and addRef on kj::Executor).

I hope you don't mind.

c++/src/kj/mutex.c++
vlovich: The lifetime of the data is contained to lexical scope but whether or not the thread is blocking isn't tied to lexical scope. A thread may wake (spuriously or otherwise), fail to acquire the lock, & go back to being blocked. If I used an RAII object to do this, then a spurious wake would introduce a race condition (that a SIGINT could conceivably hit) whereby the lock is blocked but we've cleared the metadata from the TLS variables spuriously.

I think a `KJ_DEFER` is probably a better solution here to provide the guarantee that we always unset on the way out since at that point it's safe to unset multiple times on our way out.

c++/src/kj/mutex.c++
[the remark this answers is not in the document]
vlovich: Yup. That's what I had come across as well. `thread_local` + the attribute I think is the much better approach.

c++/src/kj/mutex.h
```diff
@@ -172,6 +220,44 @@ class Mutex {
   mutable pthread_rwlock_t mutex;
 #endif
 
+#if KJ_SAVE_ACQUIRED_LOCK_INFO
+  uint32_t lockedExclusivelyByThread = 0;
```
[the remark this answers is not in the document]
vlovich: tid_t

c++/src/kj/mutex.h
```diff
@@ -172,6 +220,44 @@ class Mutex {
...
+  SourceLocation lockAcquiredLocation;
+
+  volatile bool updatedSharedMetadata = false;
+
+  void acquiredExclusive(uint32_t tid, const SourceLocation& location) noexcept {
+    __atomic_store_n(&lockedExclusivelyByThread, tid, __ATOMIC_RELEASE);
+    lockAcquiredLocation = location;
```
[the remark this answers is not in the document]
vlovich: TLDR: Yes in theory, but how? Do I use `__atomic_thread_fence` with `__ATOMIC_RELAXED`? Assuming yes, my reading of the GCC man page would indicate that `__atomic_thread_fence` only impacts memory that was mutated by atomics per the C11 memory model (which I think is consistent with C++, right?).

This is a multi-word composite object so there's not really any atomics that would work (& this must be a copy unlike the blocking TLS metadata since this is "global" for the instance of the mutex). I think that's why I had non-relaxed semantics to try to control the visibility but that probably wouldn't have worked anyway as written (probably would need to carefully maintain a sentinel value).

I had actually thought about the line of reasoning you expressed above:

> because we don't start trying to trace locks until they are held for a long time

but my unease with that line of thinking was that the watchdog is racing ongoing lock/unlock operations so there is actually good reason to want this to be consistent & timely or we'll end up with superfluous Sentry issues (due to fingerprinting separating it from the underlying issue) that will also be missing critical data to make it easy to triage (or mislead the triager down the wrong analysis path). That being said, I think I can convince myself that this is a marginal scenario not worth worrying about for now.

c++/src/kj/mutex.h
```diff
@@ -607,6 +615,42 @@ inline const T& Lazy<T>::get(Func&& init, const LockSourceLocation& location) co
...
+  const SourceLocation& origin;
+  // Where did the blocking operation originate from.
+};
+
+using BlockedOnReason = OneOf<BlockedOnMutexAcquisition, BlockedOnCondVarWait, BlockedOnOnceInit>;
+
+Maybe<BlockedOnReason&> blockedReason() noexcept;
+// Returns the information about the reason the current thread is blocked synchronously on KJ
+// operations. Returns nullptr if the current thread is not currently waiting synchronously on a
+// reason.
```
kentonv, line 731: Maybe add: "This is intended to be called from a signal handler."

Otherwise it sounds like it's impossible to return non-null. :)

c++/src/kj/mutex.c++
```diff
@@ -59,6 +59,51 @@
 #endif
 
 namespace kj {
+#if KJ_TRACK_LOCK_BLOCKING
+static thread_local BlockedOnReason* tlsBlockReason __attribute((tls_model("initial-exec")));
```
kentonv, line 63: The commit message could use an update.

c++/src/kj/mutex.h
```diff
@@ -45,11 +46,24 @@ KJ_BEGIN_HEADER
 #include <pthread.h>
 #endif
 
+// There are 3 macros controlling lock tracking:
```
kentonv, line 49: Why three instead of just one that enables everything? Who would want to enable just one without the others? Is the code actually tested in every configuration?

c++/src/kj/mutex.c++
```diff
@@ -263,6 +288,10 @@ bool Mutex::lock(Exclusivity exclusivity, Maybe<Duration> timeout,
         }
         state = __atomic_load_n(&futex, __ATOMIC_ACQUIRE);
       }
+
+      // First reader wins in terms of saving the location.
```
kentonv, line 292: Seems like last reader wins, since every reader will overwrite? You could check if the increment brought the reader count to 1, in order to decide whether this is the "first" reader.

c++/src/kj/mutex.c++
```diff
@@ -342,6 +375,10 @@ void Mutex::unlock(Exclusivity exclusivity, Waiter* waiterToSkip) {
       KJ_DASSERT(futex & SHARED_COUNT_MASK, "Unshared a mutex that wasn't shared.");
       uint state = __atomic_sub_fetch(&futex, 1, __ATOMIC_RELEASE);
 
+      if ((state & SHARED_COUNT_MASK) == 0) {
+         releasingShared();
```
kentonv, line 379: Technically the reader lock is _already_ released at this point, if the count reached zero, and some other thread could could already be writing a new location (maybe even an exclusive one), making this a bit of a race. Since this is just for debugging I suppose it doesn't matter, but maybe worth a comment. Or maybe you shouldn't bother trying to erase the location info when releasing a shared lock...

c++/src/kj/mutex.h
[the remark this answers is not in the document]
kentonv: I suppose you could use `tid` to protect it:

* In this function, write `tid` _after_ `lockAcquiredLocation`, with `__ATOMIC_RELEASE`. This ensures that `lockAquiredLocation` is fully written before `tid`.
* When releasing, set `tid` back to zero _before_ modifying `lockAcquiredLocation`, with `__ATOMIC_ACQUIRE`. This ensures the subsequent modification of `lockAcquiredLocation` happens only after `tid` is already zero.
* In `lockedInfo()`, read `tid` with `__ATOMIC_ACQUIRE` before accessing `lockAcquiredLocation`. If `tid` is zero, assume `lockAcquiredLocation` is invalid.

I guess in `lockedInfo()` you currently assume `tid` of zero means a shared lock, but that assumption seems sketchy -- `tid` might be zero just because the locking thread is already in the process of unlocking.

I'm not sure if there's a good way to extend this to shared locks, though.

Admittedly this protection doesn't accomplish very much. But if we're not really rigorously protecting `location` then I would say there's no point is using atomics on `tid` either.

c++/src/kj/source-location.c++
```diff
@@ -0,0 +1,32 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+#if __cplusplus > 201703L
+
+#include "source-location.h"
+
+namespace kj {
+kj::String KJ_STRINGIFY(const SourceLocation& l) {
```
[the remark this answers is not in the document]
vlovich: That's a good point. I'll clean this up.

c++/src/kj/mutex.h
```diff
@@ -67,7 +80,8 @@ class Mutex {
     SHARED
   };
 
-  bool lock(Exclusivity exclusivity, Maybe<Duration> timeout = nullptr);
+  bool lock(Exclusivity exclusivity, Maybe<Duration> timeout = nullptr
+     , const LockSourceLocation& location = LockSourceLocation::current());
```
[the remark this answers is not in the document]
vlovich: Search & replace. I'll fix it.

c++/src/kj/mutex.h
```diff
@@ -405,12 +423,17 @@ class ExternalMutexGuarded {
   //   - The value has been moved away.
   // - If ExternalMutexGuarded<T> is ever moved, then T must have a move constructor and move
   //   assignment operator that do not follow any pointers, therefore do not need to take a lock.
-
+  //
+  // Inherits from LockSourceLocation to perform an empty base class optimization when lock tracking
```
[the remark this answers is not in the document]
vlovich: Ok. I'll move this to be a member variable annotated with `[[no_unique_address]]` in C++20.

c++/src/kj/mutex.h
```diff
@@ -80,7 +94,8 @@ class Mutex {
     virtual bool check() = 0;
   };
 
-  void wait(Predicate& predicate, Maybe<Duration> timeout = nullptr);
+  void wait(Predicate& predicate, Maybe<Duration> timeout = nullptr
+     , const LockSourceLocation& location = LockSourceLocation::current());
```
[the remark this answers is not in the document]
vlovich: Yeah, I was a little unsure here & had similar thoughts as you. I was really just using the cppreference code as the template. The counter though is what is the compiler's behavior when it's an actual `SourceLocation`. It turns out that my back & forth is justified. By value is optimal for `NoopSourceLocation` & const ref is optimal for `SourceLocation` (with clang being a bit worse than GCC for SourceLocation). Interestingly on ARM by value is universally optimal so there's something about x86's calling convention perhaps.

Let me know if I should pessimize the SourceLocation case or if I should use a macro to make this optimal for x86.

https://godbolt.org/z/6abEn3hs3

c++/src/kj/mutex.h
vlovich: Technically there's only 2 macros (should update the doc to be clearer) - tracking the location the thread is blocked on & tracking the location that's holding the lock. I tested that it builds in the 4 permutations.

I figure there could be a desire to want to be able to track which mutex something is bocked on without increasing the size of the Mutex. It's probably safe to reduce it to 3 permutations by implying tracking block location if you are requesting to track the location holding the lock. That being said, since we're the only initial customer & we'll want both, I could also be convinced to just simplify further & only support 1 configuration here. Let me know.

c++/src/kj/mutex.c++
vlovich: "First reader" here means "first reader to hit this". I originally had this check the count but I was uneasy that there would be a non-controllable amount of time where the lock is held by a reader without the state being updated (e.g. unlock with 5k threads blocked on a reader lock, the first lock may be the last one out & take a while). This approach felt safer in that it shortens this time window to be more bounded since the locks trying to set the state will be the batch that's winning the thundering herd.

c++/src/kj/mutex.c++
vlovich: Hmm... that's a good point. I like your suggestion of not bothering to erase the location info.

c++/src/kj/mutex.h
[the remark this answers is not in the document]
vlovich: Figured out an elegant solution by having a typedef to `const SourceLocation&` vs `NoopLocation` when using it as an argument: https://godbolt.org/z/q6G8ee5a3

c++/src/kj/mutex.c++
vlovich: Will leave this open so I can remember to do it once I squash the fixups.

c++/src/kj/source-location.c++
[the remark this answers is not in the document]
vlovich: Also it turns out that support for this can be relaxed to C++20 or clang/gcc (v9 & v5 respectively).

c++/src/kj/mutex.h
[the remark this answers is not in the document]
vlovich: I've rewritten the `lockedInfo` function to just read atomically from state instead of trying to guess at it from variables that are impossible to synchronize. At least this way while the data may be read stale, at least we'll get the right type of lock.

c++/src/kj/mutex.c++
kentonv: > "First reader" here means "first reader to hit this".

OK, but again, it looks like _every_ reader will overwrite `lockAcquiredLocation`, not just the first reader, so I don't understand how the first one "wins". In fact they could be writing to `lockAcquiredLocation` at the same time so the value could end up inconsistent (like a filename from one thread and a line number from another). That's all fine if this is just for debugging anyway but I think the comment could be clearer.

c++/src/kj/source-location.h
```diff
@@ -0,0 +1,100 @@
...
+  // makes it integrate a little more nicely into KJ.
+public:
+#if KJ_COMPILER_SUPPORTS_SOURCE_LOCATION
+  static constexpr SourceLocation current(
+      const char *file = __builtin_FILE(), const char *function = __builtin_FUNCTION(),
+      uint line = __builtin_LINE(), uint column = KJ_CALLER_COLUMN()) {
+    // When invoked with no arguments returns a SourceLocation that points at the caller of the
+    // function.
+#if !__clang__ && __GNUC__
+    // GCC is very unhappy with constexpr StringPtr from const char* construction. Seems like Clang
```
kentonv, line 59: I wonder if `SourceLocation` should just store `const char*` instead of `StringPtr`. It would make `SourceLocation` smaller, would avoid some possibly-crashy inconsistencies from non-atomic initialization, and the only cost would be that if and when the strings are actually _used_ for something, they need to be dynamically `strlen()`ed... but that would only be in crash reporting anyway.

`kj::Exception` uses `const char*` for the source file name, FWIW.

c++/src/kj/common.h
```diff
@@ -214,6 +214,28 @@ typedef unsigned char byte;
...
+#if __cplusplus > 201703L || (__clang__  && __clang_major__ >= 9 && __cplusplus >= 201103L)
+// Technically this was only added to C++20 but Clang allows it for >= C++11 and spelunking the
+// attributes manual indicates it first came in with Clang 9.
+#define KJ_NO_UNIQUE_ADDRESS [[no_unique_address]]
+#else
+#define KJ_NO_UNIQUE_ADDRESS
+#endif
+
+#if defined(__has_feature)
+#if __has_feature(thread_sanitizzer)
```
kentonv, line 226:
```suggestion
#if __has_feature(thread_sanitizer)
```

c++/src/kj/mutex.c++
vlovich: Unfortunately github doesn't make it easy to view diffs between patchsets. I think the comment was from when I was checking for reader count to 1. I thought I had changed it to an atomic that protected the first one but I guess I had changed my mind part-way through. Sorry for the confusion. I'll update the comment.

c++/src/kj/source-location.h
vlovich: Not a bad suggestion. I would like a convenience method that returns the `kj::StringPtr`. Any objection?

c++/src/kj/mutex.h
vlovich: That brings up a good case. How do I make sure the new config is covered in CI to avoid accidental bitrot before a roll-deps lands this in EW?

c++/src/kj/source-location.h
kentonv: Why though? `const char*` can implicitly cast to `StringPtr` already...

c++/src/kj/mutex.c++
```diff
@@ -262,6 +285,19 @@ bool Mutex::lock(Exclusivity exclusivity, Maybe<Duration> timeout, LockSourceLoc
...
         state = __atomic_load_n(&futex, __ATOMIC_ACQUIRE);
       }
+
+      // We just want to record the lock being acquired somewhere but the specific location doesn't
+      // matter. This does mean that race conditions could occur where a thread might read this
+      // inconsistently (e.g. filename from 1 lock & function from another). This currently is just
+      // meant to be a debugging aid for manual analysis so it's OK for that purpose. If it's ever
+      // required for this to be used for anything else, then this should probably be changed to
+      // use an additional atomic variable that can ensure only one writer updates this. Or use the
+      // futex variable to ensures that this is only done for the first one to acquire the lock,
```
jclee, line 295: s/ensures/ensure

c++/src/kj/mutex.h
```diff
@@ -339,27 +422,27 @@ class MutexGuarded {
...
 
-  Locked<T> lockExclusive(const LockSourceLocation& location = LockSourceLocation::current()) const;
+  Locked<T> lockExclusive(LockSourceLocationArg location = LockSourceLocation::current()) const;
   // Exclusively locks the object and returns it.  The returned `Locked<T>` can be passed by
   // move, similar to `Own<T>`.
   //
   // This method is declared `const` in accordance with KJ style rules which say that constness
   // should be used to indicate thread-safety.  It is safe to share a const pointer between threads,
-  // but it is not safe to share a mutable pointer.  Since the whole point of MutexGuarded is to
+  // but it is not safe to share a mutable pointer.  Since the whole point of MutexGuarded is to/sign
```
jclee, line 431: stray edit?

c++/src/kj/mutex.h
vlovich: I think I figured out how to add coverage. Was conservative for now in case you were concerned about build time but I can add coverage for the mix'n'match combination of options as well.

c++/src/kj/mutex.c++
```diff
@@ -59,6 +59,51 @@
 #endif
 
 namespace kj {
+#if KJ_TRACK_LOCK_BLOCKING
+static thread_local BlockedOnReason* tlsBlockReason __attribute((tls_model("initial-exec")));
+// The initial-exec model ensures that even if this code is part of a shared library built PIC, then
+// we still place this variable in the appropriate ELF section so that __tls_get_addr is avoided.
+// It's unclear if __tls_get_addr is still async signal safe in glibc. The only negative downside
```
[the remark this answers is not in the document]
kentonv: ^ typo still present in this comment

.github/workflows/quick-test.yml
```diff
@@ -25,6 +25,22 @@ jobs:
...
+        features: ["-DKJ_TRACK_LOCK_BLOCKING=1 -DKJ_SAVE_ACQUIRED_LOCK_INFO=1"]
+    steps:
+      - uses: actions/checkout@v2
+      - name: install dependencies
+        run: |
+            export DEBIAN_FRONTEND=noninteractive
+            sudo apt-get install -y build-essential git zlib1g-dev cmake libssl-dev ${{ matrix.compiler }}
+      - name: super-test
+        run: |
+            ./super-test.sh quick ${{ matrix.compiler }} cpp-features "${{matrix.features}}" extra-libs "-lrt"
```
kentonv, line 43: What is librt needed for?

kentonv:
FYI there are some open comments that GitHub is "conveniently" hiding behind this little box:

Screenshot from 2021-05-12 23-15-53

It looks like they haven't been addressed.

.github/workflows/quick-test.yml
vlovich: `timer_X` functions in the unit test. I was using it as a way to generate a signal that could run during a deadlock so I can test the TLS stuff.

kentonv: Fair enough, maybe worth a comment.

c++/src/kj/mutex.h
```diff
@@ -162,7 +177,7 @@ class Once {
     virtual void run() = 0;
   };
 
-  void runOnce(Initializer& init);
+  void runOnce(Initializer& init, const LockSourceLocation& location = LockSourceLocation::current());
```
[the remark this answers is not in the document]
vlovich: It's a good idea but unfortunately the only place that can actually work is `runOnce`. At the other call-sites it comes after default variables & since there's no longer macros I can't move it to be a non-default parameter without breaking a lot of existing code. This feels like a change that would make sense if it could be applied in all 3 places, but just for `runOnce` this feels perhaps not worth the effort?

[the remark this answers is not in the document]
kentonv: > At the other call-sites it comes after default variables

But clearly those defaults are no longer relied upon either, so you might as well make them not defaulted?

c++/src/kj/mutex.h
vlovich: Clarified this section wording further & put a note on why this feature isn't available outside of a signal handler (& how an enhancement could be made that way for anyone who needs it).

vlovich:
Oof. That's an unfortunate UI flaw. Is there some UI setting I can set to never hide these?

c++/src/kj/mutex.h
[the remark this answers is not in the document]
vlovich: Ah, that's a good point I hadn't considered. Let me try that.

vlovich:
I forgot there were outstanding issues and we were in an indefinite impasse :). Should be resolved now.

c++/src/kj/source-location.h
```diff
@@ -0,0 +1,91 @@
...
+  static constexpr SourceLocation current(
+      const char* file = __builtin_FILE(), const char* function = __builtin_FUNCTION(),
+      uint line = __builtin_LINE(), uint column = KJ_CALLER_COLUMN()) {
+    // When invoked with no arguments returns a SourceLocation that points at the caller of the
+    // function.
+    return SourceLocation(file, function, line, column);
+  }
+#endif
+
+  constexpr SourceLocation() : SourceLocation("??", "??", 0, 0) {}
```
kentonv, line 62: Is there a good reason not to have SourceLocation's constructor default to initializing it to the line that called it? Then the separate `current()` call wouldn't be needed, and you could use `{}` as the default initializer which is much more concise.

I ask because I wrote some code yesterday that took advantage of __builtin_FILE() / __builtin_LINE() and I did it this way and thought it worked out pretty nicely.

.github/workflows/quick-test.yml
```diff
@@ -25,6 +25,24 @@ jobs:
...
+        features: ["-DKJ_TRACK_LOCK_BLOCKING=1 -DKJ_SAVE_ACQUIRED_LOCK_INFO=1"]
+    steps:
+      - uses: actions/checkout@v2
+      - name: install dependencies
+        run: |
+            export DEBIAN_FRONTEND=noninteractive
+            sudo apt-get install -y build-essential git zlib1g-dev cmake libssl-dev ${{ matrix.compiler }}
+      - name: super-test
+        run: |
+            # librt is used for creating timers via (timer_X) functions so that we can simulate a deadlock
```
kentonv, line 43: I was thinking more along the lines of: "librt is needed for timer_create in the unit tests for lock tracking"

The goal is not to fully explain the reasoning, but just to give people a hint where to look if they're trying to figure out why this random extra flag is here. Sometimes less information is better. :)

c++/src/kj/source-location.h
vlovich: Not that I can think of. As I mentioned, I was just following the spec for `std::source_location` & I figured there must be some reason why it was a separate function. I'll change it to be automatic for the default constructor.

vlovich: I guess the one weird thing is that you can't have an "explicit" on the constructor. That means that if you pass in a string in the wrong spot you'll "corrupt" the SourceLocation with no warning from the compiler.

```
void foo(SourceLocation = {});

foo(); // OK - this is what we want.
foo("abc"); // BAD - no warning since the constructor can't be explicit or `{}` wouldn't work
```

I think that's the reason the C++ standard went with an explicit function as it avoids these problems. Let me know what you'd like (generally I'm paranoid about default arguments being eager to swallow accidents).

kentonv: Interesting point. It looks like the following construction is impossible to call with non-default arguments:

```
class SourceLocation {
private:
  struct Badge { explicit Badge() {} };
public:
  SourceLocation(Badge = Badge(), const char* file = __builtin_FILE(), uint line = __builtin_LINE())
      : file(file), line(line) {}
  ...
};
```

vlovich: Clever. Applied.

vlovich:
Test failure appears sporadic & unrelated:

```
[ TEST ] kj/time-test.c++:95: all clocks advance in real time
Failed to start process notifications for pid 15348 (19)
kj/time-test.c++:142: clocks seem inaccurate even after 20 tries; coarseCalDiff / kj::MICROSECONDS = 187390; preciseCalDiff / kj::MICROSECONDS = 187390; coarseMonoDiff / kj::MICROSECONDS = 187391; preciseMonoDiff / kj::MICROSECONDS = 187391
[ FAIL ] kj/time-test.c++:95: all clocks advance in real time (4577185 μs)
[ TEST ] kj/threadlocal-test.c++:33: legacy test: ThreadLocal/Basic
```

PR #1229, opened 2021-05-05
Add methods to WebSocket to return bytes sent and received

bthwaites:
~~The functions send(), close(), disconnect(), pumpTo() and tryPumpFrom() are modified to return a promise for the number of bytes sent by the respective operations.~~

~~One possible problem with this approach is that if the promise after bytes, we'll lose the count. I wonder if it might be better to store the bytes sent as some field on the WebSocket object itself.~~

Updated the PR to add new methods to the WebSocket interface for returning the number of bytes sent and received. This is still quite a work in progress. Issues:

1) The accounting in::receive() isn't right. It looks at the payload length in the header, but doesn't work if the sender disconnects before sending the entire payload. This is why the new test is currently failing. I haven't figured out exactly how to count properly here yet.
2) The implementation in WebSocketPipeImpl is not tested, and might still be incomplete.

### DIFF
--- a/c++/src/capnp/compat/http-over-capnp.c++
+++ b/c++/src/capnp/compat/http-over-capnp.c++
@@ -157,6 +157,7 @@ public:
     auto req = KJ_REQUIRE_NONNULL(out, "already called disconnect()").sendDataRequest(
         MessageSize { 8 + message.size() / sizeof(word), 0 });
     req.setData(message);
+    sentBytes += message.size();
     return req.send();
   }
 
@@ -164,6 +165,7 @@ public:
     auto req = KJ_REQUIRE_NONNULL(out, "already called disconnect()").sendTextRequest(
         MessageSize { 8 + message.size() / sizeof(word), 0 });
     memcpy(req.initText(message.size()).begin(), message.begin(), message.size());
+    sentBytes += message.size();
     return req.send();
   }
 
@@ -171,6 +173,7 @@ public:
     auto req = KJ_REQUIRE_NONNULL(out, "already called disconnect()").closeRequest();
     req.setCode(code);
     req.setReason(reason);
+    sentBytes += reason.size() + 2;
     return req.send().ignoreResult();
   }
 
@@ -223,10 +226,14 @@ public:
     }
   }
 
+  uint64_t sentByteCount() override { return sentBytes; }
+  uint64_t receivedByteCount() override { return KJ_ASSERT_NONNULL(in)->receivedByteCount(); }
+
 private:
   kj::Maybe<kj::Own<kj::WebSocket>> in;   // One end of a WebSocketPipe, used only for receiving.
   kj::Maybe<capnp::WebSocket::Client> out;  // Used only for sending.
   kj::Own<kj::PromiseFulfiller<kj::Promise<Capability::Client>>> shorteningFulfiller;
+  uint64_t sentBytes = 0;
 };
 
 // =======================================================================================
--- a/c++/src/kj/compat/http-test.c++
+++ b/c++/src/kj/compat/http-test.c++
@@ -1687,7 +1687,8 @@ KJ_TEST("WebSocket core protocol") {
       .then([&]() { return client->send(mediumString); })
       .then([&]() { return client->send(bigString); })
       .then([&]() { return client->send(kj::StringPtr("world").asBytes()); })
-      .then([&]() { return client->close(1234, "bored"); });
+      .then([&]() { return client->close(1234, "bored"); })
+      .then([&]() { KJ_EXPECT(client->sentByteCount() == 90307)});
 
   {
     auto message = server->receive().wait(waitScope);
@@ -1718,6 +1719,7 @@ KJ_TEST("WebSocket core protocol") {
     KJ_ASSERT(message.is<WebSocket::Close>());
     KJ_EXPECT(message.get<WebSocket::Close>().code == 1234);
     KJ_EXPECT(message.get<WebSocket::Close>().reason == "bored");
+    KJ_EXPECT(server->receivedByteCount() == 90307);
   }
 
   auto serverTask = server->close(4321, "whatever");
@@ -1727,6 +1729,7 @@ KJ_TEST("WebSocket core protocol") {
     KJ_ASSERT(message.is<WebSocket::Close>());
     KJ_EXPECT(message.get<WebSocket::Close>().code == 4321);
     KJ_EXPECT(message.get<WebSocket::Close>().reason == "whatever");
+    KJ_EXPECT(client->receivedByteCount() == 12);
   }
 
   clientTask.wait(waitScope);
@@ -2012,6 +2015,35 @@ KJ_TEST("WebSocket ping received during pong send") {
   expectRead(*client, EXPECTED2).wait(waitScope);
 
   clientTask.wait(waitScope);
+}
+
+KJ_TEST("WebSocket pump byte counting") {
+  KJ_HTTP_TEST_SETUP_IO;
+  auto pipe1 = KJ_HTTP_TEST_CREATE_2PIPE;
+  auto pipe2 = KJ_HTTP_TEST_CREATE_2PIPE;
+
+  FakeEntropySource maskGenerator;
+  auto server1 = newWebSocket(kj::mv(pipe1.ends[1]), nullptr);
+  auto client2 = newWebSocket(kj::mv(pipe2.ends[0]), maskGenerator);
+  auto server2 = newWebSocket(kj::mv(pipe2.ends[1]), nullptr);
+
+  auto pumpTask = server1->pumpTo(*client2);
+  auto receiveTask = server2->receive();
+
+  // Client sends three bytes of a valid message then disconnects.
+  const char DATA[] = {0x01, 0x06, 'h'};
+  pipe1.ends[0]->write(DATA, 3).wait(waitScope);
+  pipe1.ends[0] = nullptr;
+
+  // The pump completes successfully, forwarding the disconnect.
+  pumpTask.wait(waitScope);
+
+  // The eventual receiver gets a disconnect execption.
+  KJ_EXPECT_THROW(DISCONNECTED, receiveTask.wait(waitScope));
+
+  KJ_EXPECT(server1->receivedByteCount() == 3);
+  KJ_EXPECT(client2->sentByteCount() == 3);
+  KJ_EXPECT(server2->receivedByteCount() == 3);
 }
 
 KJ_TEST("WebSocket pump disconnect on send") {
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -2194,6 +2194,7 @@ public:
 
       return stream->tryRead(recvData.end(), 1, recvBuffer.end() - recvData.end())
           .then([this,maxSize](size_t actual) -> kj::Promise<Message> {
+        receivedBytes += actual;
         if (actual == 0) {
           if (recvData.size() > 0) {
             return KJ_EXCEPTION(DISCONNECTED, "WebSocket EOF in frame header");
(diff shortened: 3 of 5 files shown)


c++/src/kj/compat/http-test.c++
```diff
@@ -1720,7 +1720,7 @@ KJ_TEST("WebSocket core protocol") {
     KJ_EXPECT(message.get<WebSocket::Close>().reason == "bored");
   }
 
-  auto serverTask = server->close(4321, "whatever");
+  auto serverTask = server->close(4321, "whatever").then([](uint64_t s) { KJ_EXPECT(s == 10); });
```
jclee, line 1723: Is the assertion run?  Naively, I'd expect s to be 8 for the length of "whatever", but maybe I'm missing something.

bthwaites: The close() function is adding two extra bytes here: https://github.com/capnproto/capnproto/blob/master/c%2B%2B/src/kj/compat/http.c%2B%2B#L2132

I am not sure what the purpose for this is, I'd be interested to know.

jclee: Ah, looks like it packs the 16-bit code in the first two bytes.

kentonv:
Hmmm.... I hate to say this after all this work was done, but I am wondering if it would be better to just add some sort of `getTotalBandwidth()` method that returns the number of bytes sent and received so far. Otherwise I'm a bit concerned that this change seems to be a breaking API change that could be a pain for some users. Plus I bet it would be a lot easier to count totals.

c++/src/kj/compat/http.c++
```diff
@@ -2857,6 +2875,9 @@ private:
 
   kj::Own<WebSocket> ownState;
 
+  uint64_t sentBytes = 0;
```
kentonv, line 2878: So, `WebSocketPipeImpl` is a little weird. Normally the `WebSocket` class represents one end of a WebSocket. Messages written to it are received by some other `WebSocket` instance somewhere else. But `WebSocketPipeImpl` is actually a reflector -- messages you send to it with `send()` are then received by calling `receive()` _on the same object_.

When you create a WebSocket pipe, under the hood, two `WebSocketPipeImpl`s are constructed to represent messages flowing in the two directions. And then, as a layer on top of that, two `WebSocketPipeEnd`s are constructed to represent the two ends of the pipe. Each end holds references to both underlying `WebSocketPipeImpl`, one to satisfy sends and one for receives. The other end holds the same references in the opposite order.

What all this means is that `WebSocketPipeImpl` doesn't need to have separate `sentBytes` and `receivedBytes`, because the bytes it sends are exactly the same as the bytes it receives. It can just have a single `transferredBytes`.

Moreover, I think the right way to implement this count is not by editing all the individual state types like `BlockedSend` and `BlockedReceive`, but instead modify the methods of `WebSocketPipeImpl` itself. Note that `send()` and `tryPumpFrom()` both "sending" operations, while `receive()` and `pumpTo()` are both "receiving" operations; you should choose either the sending side or the receiving side to do the counting (but not both). Then you can revert all your changes to the `BlockedWhatever` state classes as they don't need to do any counting of their own.

For pump operations, I think the easiest way to determine how many bytes were pumped is to ask the other WebSocket for its counts before and after the operation, and take the difference. E.g. if you count in `pumpTo()`, check `other.sentBytesCount()` before and after the call (since `pumpTo()` receives from this object and sends to `other`).

Note that for `send()`/`receive()`, you should only count the bytes after the operation completes (without being canceled or erroring out), since otherwise no bytes were actually transferred. But for `pumpTo()`, a partial pump is possible. The strategy mentioned above of checking the other WebSocket's counts before and after should therefore be applied whether the pump succeeds, fails, or is canceled. To cover all three cases, use an attached defer, like:

    auto before = other.receivedBytesCount();
    return pumpTo(other).attach(kj::defer([this,before]() {
      transferredBytes += other.receivedBytesCount() - before;
    }));

The attached-defer will be executed on success, failure, and cancellation.

c++/src/kj/compat/http.h
```diff
@@ -572,6 +572,9 @@ class WebSocket {
   // if this WebSocket implementation is able to perform the pump in an optimized way, better than
   // the default implementation of pumpTo(). The default implementation of pumpTo() always tries
   // calling this first, and the default implementation of tryPumpFrom() always returns null.
+
+  virtual uint64_t sentBytesCount();
```
kentonv, line 576: Nitpick: Normally I'd use either a plural word or a "-Count" suffix but not both. So I'd make this `sentByteCount()` (non-plural "byte").

c++/src/kj/compat/http.c++
bthwaites: I appreciate this explanation, I was getting turned around in this part. Do you think it makes sense for both `sentBytesCount()` and `receivedBytesCount()` to return the `transferredBytes`?

kentonv: That would be fine, or you could have them throw and define a new method. All that matters is that `WebSocketPipeEnd` knows which method to call, since it is the only thing that gets direct access to the underlying `WebSocketPipeImpl`s.

c++/src/capnp/compat/http-over-capnp.c++
```diff
@@ -223,6 +223,9 @@ public:
     }
   }
 
+  uint64_t sentBytesCount() override { return KJ_ASSERT_NONNULL(in)->sentBytesCount(); }
```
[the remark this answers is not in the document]
bthwaites: Updated to keep using `in` for receiving and to count bytes sent in `send()`.

c++/src/kj/compat/http.c++
```diff
@@ -2305,6 +2305,7 @@ public:
       }
     });
 
+    receivedBytes += payloadLen;
```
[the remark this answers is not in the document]
bthwaites: Added to the calls of `stream->tryRead()`. I noticed that this changes the byte counts a little bit in my test. I think it is because now we're counting the full message size including headers, not just the payload size. I'd say this behavior is probably better.

c++/src/kj/compat/http.c++
```diff
@@ -2645,11 +2654,13 @@ private:
...
     }
 
-    if (recvData.size() > 0) {
+    auto size = recvData.size();
+    if (size > 0) {
       // We have some data buffered. Write it first.
-      return other.stream->write(recvData.begin(), recvData.size())
-          .then([this, &other]() {
+      return other.stream->write(recvData.begin(), size)
+          .then([this, &other, size]() {
```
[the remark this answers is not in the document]
bthwaites: Switched to this syntax.

c++/src/kj/compat/http.c++
```diff
@@ -2745,6 +2757,9 @@ kj::Maybe<kj::Promise<void>> WebSocket::tryPumpFrom(WebSocket& other) {
   return nullptr;
 }
 
+uint64_t WebSocket::sentBytesCount() { return 0; }
```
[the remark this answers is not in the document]
bthwaites: Switched to using pure-virtual methods.

c++/src/kj/compat/http.c++
bthwaites: We need to be able to call `receivedBytesCount()` on `other` so I have both methods returning `transferredBytes` instead of adding a new function.

c++/src/kj/compat/http.h
bthwaites: Switched to `sentByteCount()`

bthwaites:
I think I'm still missing some test coverage such as using `pumpTo()` on a `WebSocketPipe`, working on that now.

c++/src/kj/compat/http.c++
```diff
@@ -2649,8 +2659,9 @@ private:
     if (recvData.size() > 0) {
       // We have some data buffered. Write it first.
       return other.stream->write(recvData.begin(), recvData.size())
-          .then([this, &other]() {
+          .then([this, &other, size = recvData.size()]() {
         recvData = nullptr;
+        sentBytes += size;
```
bthwaites, line 2664: I'm thinking I got this backwards. `a.pumpTo(b)` should only increment `receivedBytes` on `a`. Eventually we'll call `b.send()` which will increment `sentBytes` on `b`. Is that right?

kentonv: On this line specifically, I think you want to increment `other.sentBytes`. I think that `this->receivedBytes` has already accounted for bytes in the `recvData` buffer.

c++/src/kj/compat/http.c++
```diff
@@ -2663,10 +2674,11 @@ private:
     });
 
     // There's no buffered incoming data, so start pumping stream now.
-    return stream->pumpTo(*other.stream).then([&other](size_t) -> kj::Promise<void> {
+    return stream->pumpTo(*other.stream).then([this, &other](size_t s) -> kj::Promise<void> {
       // WebSocket pumps are expected to include end-of-stream.
       other.disconnected = true;
       other.stream->shutdownWrite();
+      sentBytes += s;
```
kentonv, line 2681: Down here, you need to increment both `this->receivedBytes` as well as `other.sentBytes` to account for the bytes that were pumped directly from one underlying stream to the other.

bthwaites:
Fixed http-test to go with the fixed logic

c++/src/capnp/compat/http-over-capnp.c++
```diff
@@ -157,21 +157,21 @@ public:
     auto req = KJ_REQUIRE_NONNULL(out, "already called disconnect()").sendDataRequest(
         MessageSize { 8 + message.size() / sizeof(word), 0 });
     req.setData(message);
-    return req.send();
+    return req.send().then([this, size = message.size()](){ sentBytes += size; });
```
kentonv, line 160: I think you can count the `sentBytes` before actually calling `send()`, to avoid the need to use a `.then()`.

Technically, the promise returned by `send()` doesn't really tell you when the bytes have "been sent" anyway. It resolves as soon as it's OK to send more. The bytes may still be in a buffer somewhere, and any number of errors could still prevent the bytes from ever being delivered.

bthwaites: So I suppose there's no way to exclude the bytes that were not delivered without modifying the underlying `StreamingRequest` to do some more accounting. Do you think it's worth the effort here? Or maybe we actually prefer to count the bytes that we attempted to send.

bthwaites:
> I think I'm still missing some test coverage such as using `pumpTo()` on a `WebSocketPipe`, working on that now.

I modified websocket-rpc-test to use two WebSocketPipes with one pumping to the other. If you'd prefer not to change the existing test I could also make this a totally new test.

c++/src/capnp/compat/http-over-capnp.c++
kentonv: I mean, theoretically speaking, if you _really_ want to track what was actually delivered, you'd have to query the operating system's TCP layer to find out which packets were ACKed. And that assumes there's no proxies between you and the final destination; if there are, then you need to ask those proxies to report what they managed to deliver, etc...

I don't think any of this is worth it. If we tried to send the thing, then count it. I don't think the error rate will be significant here.

kentonv:
> I modified websocket-rpc-test to use two WebSocketPipes with one pumping to the other. If you'd prefer not to change the existing test I could also make this a totally new test.

Let's clone the test and apply your changes to the copy, just to be sure we haven't lost any coverage of any code path the old test was exercising. You can remove any lines from the copy that aren't directly relevant to what you're testing, to cut down the noise a bit.

kentonv:
Sorry this was delayed while I was on vacation (and then catching up).

Is this ready to be merged now? I can't quite remember (or figure out from the history) whether there were some issues outstanding. It looks like my most recent comments were addressed so if you think this is ready to merge, please squash the fixups, then I'll merge.

bthwaites:
I think all the issues have been addressed, I squashed the fixups.

PR #1234, opened 2021-05-11
Fix kj/parse/common.h not compiling in MSVC's C++20 mode

### LINKED ISSUE #1233: kj/parse/common.h fails to compile on MSVC C++20
kj/parse/common.h currently special-cases MSVC during compilation, but the special case fails to build:

https://github.com/capnproto/capnproto/blob/6b5bcc2c6e954bc6e167ac581eb628e5a462a469/c%2B%2B/src/kj/parse/common.h#L100-L111

- result_of_t was removed in C++20, causing a build failure when you try compiling with MSVC in C++20 mode.
- The non-MSVC branch compiles properly in the latest MSVC shipping with VS2019, and no longer results in a C2064 error. But I don't know if removing the MSVC branch will cause the decltype() expression to fail to build on older MSVC versions. I removed the MSVC branch locally, but I'm not sure if that's an acceptable bugfix to upstream.

nyanpasu64:
"kj/parse/common.h" currently special-cases MSVC to use `std::result_of_t`, but it was removed in C++20. Consequently, if you try compiling with MSVC 2019 in C++20 mode, the build fails because the variable is not found. However, I can't easily test if `result_of_t` was removed, because on MSVC, `__cplusplus` is always `199711L`, and the removal has no corresponding feature flag.

This PR removes the special case for MSVC. The resulting code compiles successfully in the latest MSVC shipping with VS2019, and no longer results in a C2064 error. Testing older MSVC versions on Godbolt, it also successfully compiles on MSVC 19.10 (VS2017, the earliest supported version) and newer. Nonetheless it's possible that that specific older versions of VS2017 not on Godbolt will fail to build.

Fixes #1233. This reverts commit 34031907426f9e8595d764834bb328eada03a6b1.

### DIFF
--- a/c++/src/kj/parse/common.h
+++ b/c++/src/kj/parse/common.h
@@ -40,7 +40,12 @@
 #include "../array.h"
 #include "../tuple.h"
 #include "../vector.h"
-#if _MSC_VER && !__clang__
+
+#if _MSC_VER && _MSC_VER < 1920 && !__clang__
+#define KJ_MSVC_BROKEN_DECLTYPE 1
+#endif
+
+#if KJ_MSVC_BROKEN_DECLTYPE
 #include <type_traits>  // result_of_t
 #endif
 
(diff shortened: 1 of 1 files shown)


kentonv:
Looks like VS 2017 (aka version 15) doesn't like it. VS 2019 (version 16) is happy. So we'll need an _MSC_VER check in between, and I guess we won't be able to support C++20 compilation with VS 2017.

nyanpasu64:
The tests seem to pass now, but IDK if the #define is a good idea.

kentonv:
I would at least prefix it with `KJ_`. Otherwise seems fine.

nyanpasu64:
Also the errors that showed up during CI are *no longer* C2064. There are multiple such errors, but no C2064 in sight. Is it worth updating the comments? (Unfortunately I can no longer pull up the GitHub Actions logs for the older broken build.)

kentonv:
You can update the comment if you want. Personally I don't care that much.

kentonv:
Thanks!

nyanpasu64:
Thanks for reviewing!

PR #1235, opened 2021-05-11
Fix String/StringPtr compilation on MinGW/MSVC C++20

### LINKED ISSUE #1232: Cap'n Proto/kj fails to build in CMake C++20 (-std:c++latest, -std=gnu++2a) due to incorrect __cplusplus value
I'm writing a CMake project with C++20 features turned on, using `set(CMAKE_CXX_STANDARD 20)`. However this causes Cap'n Proto to frequently fail to compile, because the MSVC headers are built in C++20 mode (using various internal compiler-specific flags like `_HAS_DEPRECATED_RESULT_OF` and `_HAS_CXX17`), but Cap'n Proto only tests `__cplusplus` (which is often inaccurate) for conditionally compiling based on the C++ version.

To test, I tried compiling my project in `set(CMAKE_CXX_STANDARD 20)` mode under various compilers. I added the following code to my project, which prints the value of `__cplusplus` as an error message:

```
char f(int);
char test[__cplusplus]={f(&test)};
```

To my surprise, when I set CMake to compile in C++20 mode, `__cplusplus` doesn't reflect C++20, but actually varies greatly between compilers. My compilers are all fairly recent within the last year or so.

- MSVC from VS 16.9.4
	- -std:c++latest, __cplusplus = 199711L
- Clang MSVC, 12.0.0
	- -std=gnu++20, __cplusplus = 202002L
- Clang MinGW, 11.0.0 (msys2 doesn't have Clang 12 yet)
	- -std=gnu++20, __cplusplus = 202002L (CMake treats Clang MSVC/MinGW identically, even though they take different linker flags)
- GCC MinGW, 10.3.0
	- -std=gnu++2a, __cplusplus = 201709L
	- Note that 201709L isn't the standardized __cplusplus value corresponding to C++17 (201703L).

I don't know if it's CMake's "fault" for passing -std=gnu++2a or -std:c++latest, or the compilers' "fault" for defining __cplusplus incorrectly. Nonetheless I had to fix my vendored copy of Cap'n Proto to build under these circumstances. (Clang already gets gnu++20, GCC supports gnu++20 but CMake sends gnu++2a, and MSVC doesn't support c++20 yet.)

Solution
---------

Instead of branching on `__cplusplus`, I think it's reasonable to branch on [individual feature flags](https://en.cppreference.com/w/cpp/feature_test) based on the actual features requiring C++20.

For example, when fixing [the StringPtr(char8_t*) conversions](https://github.com/capnproto/capnproto/commit/4db91df614be91db82833ec1b7cef18ef97711cb#diff-23505eba9e8acaabd5d165daafed66e65dc9714186d7f0a6c459923d954239f2R79-R87),  I replaced `#if __cplusplus >= 202000L` with `#if __cpp_char8_t`. On all compilers I've tested (MSVC, Clang MSVC/MinGW, GCC MinGW), this constant is undefined with `set(CMAKE_CXX_STANDARD 17)`, and has the expected value with `set(CMAKE_CXX_STANDARD 20)`.

I also replaced the `operator==` conditional compilation in string.h with `#if !__cpp_impl_three_way_comparison`. This was enough to make my code build and run properly, but there are probably more usages of `__cplusplus` which may fail under some circumstances, which can be located through grep or search.

nyanpasu64:
When CMake passes MinGW `-std=gnu++2a`, then MinGW GCC defines `__cplusplus` as `201709L` even though C++20 features are active. This causes Cap'n Proto's conditional compilation to think the compiler is in C++17 mode, despite the compiler running in C++20 mode, resulting in build errors (in the case of `char8_t`) or miscompilations and infinite recursion (in the case of `operator==`).

MSVC is even worse, since it unconditionally reports __cplusplus = 199711L, unless you pass in /Zc:__cplusplus in addition to /std:c++xyz ([source](https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/)).

This pull request feature-tests C++20 features by name (rather by compiler version) when conditionally compiling code based on those features. This PR fixes "kj/string.h" and makes Cap'n Proto's lite mode and generated headers compile, but I haven't looked at other usages of `__cplusplus` in the rest of kj. (There are no occurrences of `__cplusplus ` in "c++/src/capnp".)

TBH it's probably a good idea to add a `/std:c++latest` MSVC build to CI, to flush out the remaining issues in C++20 mode.

Fixes #1232.

### DIFF
--- a/c++/src/kj/string.h
+++ b/c++/src/kj/string.h
@@ -76,7 +76,7 @@ public:
   inline StringPtr(const char* begin, const char* end): StringPtr(begin, end - begin) {}
   inline StringPtr(const String& value);
 
-#if __cplusplus >= 202000L
+#if __cpp_char8_t
   inline StringPtr(const char8_t* value): StringPtr(reinterpret_cast<const char*>(value)) {}
   inline StringPtr(const char8_t* value, size_t size)
       : StringPtr(reinterpret_cast<const char*>(value), size) {}
@@ -152,7 +152,7 @@ private:
   friend constexpr kj::StringPtr (::operator "" _kj)(const char* str, size_t n);
 };
 
-#if __cplusplus < 202000L
+#if !__cpp_impl_three_way_comparison
 inline bool operator==(const char* a, const StringPtr& b) { return b == a; }
 inline bool operator!=(const char* a, const StringPtr& b) { return b != a; }
 #endif
@@ -254,7 +254,7 @@ private:
   Array<char> content;
 };
 
-#if __cplusplus < 202000L
+#if !__cpp_impl_three_way_comparison
 inline bool operator==(const char* a, const String& b) { return b == a; }
 inline bool operator!=(const char* a, const String& b) { return b != a; }
 #endif
@@ -366,7 +366,7 @@ struct Stringifier {
   template<size_t n>
   inline ArrayPtr<const char> operator*(const FixedArray<char, n>& s) const { return s; }
   inline ArrayPtr<const char> operator*(const char* s) const { return arrayPtr(s, strlen(s)); }
-#if __cplusplus >= 202000L
+#if __cpp_char8_t
   inline ArrayPtr<const char> operator*(const char8_t* s) const {
     return operator*(reinterpret_cast<const char*>(s));
   }


vlovich:
@kentonv LGTM.

kentonv:
Thanks!

PR #1239, opened 2021-05-12
Make Maybe<Own<T>> compatible with ForkedPromise

vlovich:
Add copyOrAddRef specialization for `Maybe<Own<T>>` to make this compatible with `ForkedPromise`.
An open question is whether `addRef` should be allowed to change the returned type.

I don't think that's true for `Own<T>` so I kept it consistent, but I thought I read some documentation somewhere
where `addRef` is supposed to allow to transmute to `Own<U>` (in which case there's a bug here).

### DIFF
--- a/c++/src/kj/async-test.c++
+++ b/c++/src/kj/async-test.c++
@@ -542,6 +542,34 @@ TEST(Async, ForkRef) {
   EXPECT_EQ(789, branch2.wait(waitScope));
 }
 
+TEST(Async, ForkMaybeRef) {
+  EventLoop loop;
+  WaitScope waitScope(loop);
+
+  Promise<Maybe<Own<RefcountedInt>>> promise = evalLater([&]() {
+    return Maybe<Own<RefcountedInt>>(refcounted<RefcountedInt>(123));
+  });
+
+  auto fork = promise.fork();
+
+  auto branch1 = fork.addBranch().then([](Maybe<Own<RefcountedInt>>&& i) {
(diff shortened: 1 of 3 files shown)


c++/src/kj/async-test.c++
```diff
@@ -509,6 +509,47 @@ TEST(Async, Fork) {
   EXPECT_EQ(789, branch2.wait(waitScope));
 }
 
+TEST(Async, ForkMaybeOwn) {
+  struct MyType {
+    int value = 123;
+
+    Own<MyType> addRef() const {
+      return kj::heap<MyType>(MyType{ .value = 456, });
```
kentonv, line 517: Build failure because MSVC doesn't support designated initializers until C++20.

```suggestion
      return kj::heap<MyType>(MyType{456});
```

c++/src/kj/async-test.c++
```diff
@@ -509,6 +509,47 @@ TEST(Async, Fork) {
...
+    }
+  };
+  struct MyType2 {
```
kentonv, line 520: Unused type?

FWIW the next test below tests the addRef() behavior of forked promises and defines `RefcountedInt` for that purpose. Might have been easier to copy/paste that test as a starting point.

vlovich: Yeah, much better suggestion. Thanks!

c++/src/kj/async-test.c++
vlovich: Resolved by switching to use ForkRef as the base test.

PR #1240, opened 2021-05-13
Turn create_symlink failure on Windows into a warning

ihnorton:
CMake versions >= 3.13 make use of Windows support for symlinks;
previously `cmake -E create_symlink` was a no-op. However, symlinks
are only available in Windows 10 > build 1607 with developer mode
enabled. Otherwise, the install step fails with:

```
EXEC : CMake error : failed to create symbolic link 'C:/bld/capnproto-build/install/bin/capnpc.exe': operation not permitted
```

This commit checks the error status of the create_symlink install
command, and reports failure as a warning but does not error out.


harrishancock (approved):

harrishancock:
Thanks!

PR #1241, opened 2021-05-14
Expose isValidHeaderValue in HttpHeaders

vlovich:
It's useful for the caller to be able to validate whether it's a valid
header value in case they want to handle such a situation a different
way.

### DIFF
--- a/c++/src/kj/compat/http.h
+++ b/c++/src/kj/compat/http.h
@@ -249,6 +249,17 @@ class HttpHeaders {
 public:
   explicit HttpHeaders(const HttpHeaderTable& table);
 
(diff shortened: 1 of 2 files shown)


c++/src/kj/compat/http.h
```diff
@@ -249,6 +249,8 @@ class HttpHeaders {
 public:
   explicit HttpHeaders(const HttpHeaderTable& table);
 
+  static bool isValidHeaderValue(kj::StringPtr value);
```
harrishancock, line 252: Doc comment on what this checks and when it's useful would be good.

harrishancock (approved):

PR #1244, opened 2021-05-17
Add 'acceptTimeout' option to TLS

bretthoerner:
This adds a new required constructor argument (a `Timer`) to `TlsContext`... if we want to prevent breaking existing code I could get cute and make it optional and throw an error if you try to use `acceptTimeout` without it?

### DIFF
--- a/c++/src/kj/memory.h
+++ b/c++/src/kj/memory.h
@@ -435,8 +435,17 @@ public:
   static const HeapDisposer instance;
 };
 
+#if _MSC_VER && _MSC_VER < 1920 && !defined(__clang__)
+template <typename T>
+__declspec(selectany) const HeapDisposer<T> HeapDisposer<T>::instance = HeapDisposer<T>();
+// On MSVC 2017 we suddenly started seeing a linker error on one specific specialization of
+// `HeapDisposer::instance` when seemingly-unrelated code was modified. Explicitly specifying
+// `__declspec(selectany)` seems to fix it. But why? Shouldn't template members have `selectany`
+// behavior by default? We don't know. It works and we're moving on.
+#else
 template <typename T>
 const HeapDisposer<T> HeapDisposer<T>::instance = HeapDisposer<T>();
+#endif
 
 }  // namespace _ (private)
 
--- a/c++/src/kj/compat/tls.c++
+++ b/c++/src/kj/compat/tls.c++
@@ -644,6 +644,12 @@ TlsContext::TlsContext(Options options) {
     SSL_CTX_set_tlsext_servername_arg(ctx, sni);
   }
 
+  KJ_IF_MAYBE(timeout, options.acceptTimeout) {
+    this->timer = KJ_REQUIRE_NONNULL(options.timer,
+        "acceptTimeout option requires that a timer is also provided");
+    this->acceptTimeout = *timeout;
+  }
+
   this->ctx = ctx;
 }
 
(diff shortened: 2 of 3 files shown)


a-robinson (approved):
This looks fine to me, I don't have strong feelings on whether to make it optional.

It still seems less than ideal to only wait on one incoming connection's TLS handshake than to allow multiple handshakes to operate in parallel, but this is at least a temporary improvement over possibly hanging the accept loop altogether.

kentonv:
Hmm, I'm hesitant to add this to the interface when it isn't really the right long-term solution. Seems like we need to maintain an internal loop that accepts connections and then runs them through their handshake and only resolves the external accept() call once the handshake completes... so that multiple accepts can be happening at the same time. Kind of like the kernel's listen backlog but in userspace.

kentonv:
Well... I guess you'd still want a timeout even with a backlog, huh? Otherwise it's easy to fill the backlog.

bretthoerner:
Yeah, I think we'll still want a timeouts even when we fix the larger issue.

kentonv:
I don't think the backwards-incompatible interface change will fly, though. I suspect some people other than us use this class. Can you put `kj::Maybe<kj::Timer&>` into the options struct, instead of making it a required argument? Require it only when a timeout is specified.

Hopefully that also makes the MSVC 2017 error go away. It looks like a bizarre compiler bug to me.

kentonv:
MSVC 2017 is still behaving bizarrely. The error is clearly wrong -- it's complaining that a template specialization (of `HeapDisposer<...>::instance`) had duplicate definitions, but template specializations are supposed to be tolerant of this. Also the particular specialization involves a lambda, how could that possibly be duplicated? Is this some sort of hash collision? How the heck did this PR cause it?

If you want to merge this PR now for the release and then circle back to debug later, I'm OK with that.

bretthoerner:
I think we're going to hold it to see if the previous network change does anything, so I'll check out the error.

bretthoerner:
Ah, I guess `inline` vars are C++17, have to figure out how to move this specialization into `memory.c++` then.

c++/src/kj/memory.h
```diff
@@ -435,8 +435,13 @@ class HeapDisposer final: public Disposer {
   static const HeapDisposer instance;
 };
 
+#if defined(_MSC_VER) && !defined(__clang__)
+template <typename T>
+__declspec(selectany) const HeapDisposer<T> HeapDisposer<T>::instance = HeapDisposer<T>();
```
bretthoerner, line 440: My reading list was:
* https://developercommunity.visualstudio.com/t/already-defined-symbol-for-static-template-member/311664
* https://developercommunity.visualstudio.com/t/linker-failing-because-of-multiple-definitions-of/224597
* https://docs.microsoft.com/en-us/cpp/error-messages/tool-errors/linker-tools-error-lnk2005?view=msvc-160

Not sure how we prefer to gate these things, I could do this _only_ for MSVC 2017?

kentonv: Yeah, `_MSC_VER` should evaluate to some sort of version number which you can compare against. It looks values less than 1920  would correspond to MSVC 2017? https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros

c++/src/kj/memory.h
```diff
@@ -435,8 +435,13 @@ class HeapDisposer final: public Disposer {
   static const HeapDisposer instance;
 };
 
+#if _MSC_VER && _MSC_VER < 1920 && !defined(__clang__)
```
kentonv, line 438: Maybe could use a comment...

> On MSVC 2017 we suddenly started seeing a linker error on one specific specialization of `HeapDisposer::instance` when seemingly-unrelated code was modified. Explicitly specifying `__declspec(selectany)` seems to fix it. But why? Shouldn't template members have `selectany` behavior by default? We don't know. It works and we're moving on.

bretthoerner: :+1:, done.

PR #1245, opened 2021-05-19
Implement variadic heterogenous joinPromises

vlovich:
It's pretty convenient to be able to join a heterogenous set of promises
into a single promise for the combined result (for hopefully obvious
reasons, this really only makes sense for inclusive promises).

### DIFF
--- a/c++/src/kj/async-inl.h
+++ b/c++/src/kj/async-inl.h
@@ -787,6 +787,73 @@ private:
 
 // -------------------------------------------------------------------
 
+#if __cpp_fold_expressions
+class TupleJoinPromiseNodeBase: public PromiseNode {
+public:
+  TupleJoinPromiseNodeBase(Array<Own<PromiseNode>> promises,
+                           ExceptionOrValue* resultParts, size_t maxPartSize);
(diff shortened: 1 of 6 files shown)


kentonv:
See previous attempt: https://github.com/capnproto/capnproto/pull/706

At the time I gave up on getting MSVC to work, but MSVC has improved since then so it might be worth another try. (I think back then we were still supporting MSVC 2015, but these days the minimum is MSVC 2017, and I'd even be OK with raising the requirement to MSVC 2019 if needed.)

In that PR, I chose not to expose the use of `Tuple` to the user, but instead make `Promise` itself be variadic, so you could write `kj::Promise<T, U, V>`, and then the three values would be passed as separate parameters to `.then()`. I prefer that approach because tuples are really ugly to use directly. (The implementation still used tuples though.)

Side note: I haven't looked closely at what you implemented here, but any new replacement for joinPromises should probably have fail-fast behavior, meaning that if any one promise throws, the others are canceled. The existing `kj::joinPromises` waits until all promises are settled even when one throws, and in practice it turns out that is almost never the behavior we actually want.

PR #1246, opened 2021-05-19
Annotate lifetimebound

vlovich:
Annotate `Array`/`ArrayBuilder`/`Vector`/`String` with `[[clang::lifetimebound]]` so that the compiler will warn if the lifetime of the used variable exceeds the lifetime of the underlying storage.

Unfortunately I don't think there's a way to track the lifetime bound through multiple objects (e.g. kj::Array -> kj::ArrayPtr -> slice into a new kj::ArrayPtr).

### DIFF
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -206,6 +206,35 @@ typedef unsigned char byte;
 #define KJ_WARN_UNUSED_RESULT __attribute__((warn_unused_result))
 #endif
 
+#ifdef __has_cpp_attribute
+#define KJ_HAS_CPP_ATTRIBUTE(x) __has_cpp_attribute(x)
+#else
+#define KJ_HAS_CPP_ATTRIBUTE(x) 0
+#endif
+
+#if KJ_HAS_CPP_ATTRIBUTE(clang::lifetimebound)
+// If this is generating too many false-positives, the user is responsible for disabling the
+// problematic warning at the compiler switch level or by suppressing the place where the
+// false-positive is reported through compiler-specific pragmas if available.
+#define KJ_LIFETIMEBOUND [[clang::lifetimebound]]
+#else
+#define KJ_LIFETIMEBOUND
+#endif
+// Annotation that indicates the returned value is referencing a resource owned by this type (e.g.
+// cStr() on a std::string). Unfortunately this lifetime can only be superficial currently & cannot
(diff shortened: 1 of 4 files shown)


kentonv:
This seems pretty cool, but I'm curious how it behaves in a case like:

```
kj::String str = ...;
auto ptr = str.asPtr();
someVectorMember.add(kj::mv(str));
return ptr;
```

Does the compiler understand that `str`'s content's lifetime has been extended by adding it to `someVectorMember`? Or will it warn that we're extending `ptr` beyond the lifetime of `str`?

vlovich:
That part isn't clear to me. I think the way you've asked the question crystallizes for me why this didn't flag the integer -> string conversion that motivated this change. I think the C++ proposal that's implemented by this annotation isn't complete as copies/moves erase the compilers ability to track the lifetime. I'll ask on the clang mailing list though in case there's suggestions (annotating the constructors/assignment operators feels weird because I would think it ties the lifetime incorrectly to the thing you're copying instead of the thing it's lifetime bound to)

c++/src/kj/common.h
```diff
@@ -206,6 +206,32 @@ typedef unsigned char byte;
...
 #endif
 
+#ifdef __has_cpp_attribute
+#define KJ_HAS_CPP_ATTRIBUTE(x) __has_cpp_attribute(x)
+#else
+#define KJ_HAS_CPP_ATTRIBUTE(x) 0
+#endif
+
+#if KJ_HAS_CPP_ATTRIBUTE(clang::lifetimebound)
+#define KJ_LIFETIMEBOUND [[clang::lifetimebound]]
```
kentonv, line 216: I wonder if we'll need to give people a way to opt out of this? Since it did produce several false positives in our own code. Is it a warning by default, or an error?

vlovich: It's a warning. In our system I explicitly opted into having them be errors. I can change this to be an opt-in thing.

vlovich: Done. `KJ_ENABLE_LIFETIMEBOUND`

kentonv: It looks like you implemented an opt-in rather than an opt-out. To clarify I think this should be on by default, but I was concerned someone would whine to us that it breaks their code. OTOH they can presumably disable it using the appropriate `-w` flag so in retrospect I guess it probably doesn't need a flag at all. Sorry for not being cleaer.

PR #1248, opened 2021-05-21
Support registering interest in ancillary messages on AsyncIoStream

a-robinson:
This perhaps isn't the ideal API, but does at least offer a way to get
at underlying control messages if you really care about them.


a-robinson:
Hmm, looks like MacOS may not support SO_TIMESTAMP on TCP sockets, only on UDP. Is it a concern if I just disable the ancillary message tests on mac as well?

kentonv:
> Is it a concern if I just disable the ancillary message tests on mac as well?

Fine with me.

PR #1254, opened 2021-05-28
Add oldestQueuedMessage metric to TwoPartyVatNetwork

xortive:
This is another useful metric for backpressure, when protecting against queued message latency is important as well as protecting against arbitrary memory growth due to an unbounded amount queued message.

I'd have liked to use `kj::List`, but with only a forward declared `OutgoingMessageImpl` getting the pointer-to-member attribute wasn't possible. I'd appreciate advice if there's a way to get around this without polluting the header.

### DIFF
--- a/c++/src/capnp/rpc-twoparty.c++
+++ b/c++/src/capnp/rpc-twoparty.c++
@@ -30,14 +30,17 @@ TwoPartyVatNetwork::TwoPartyVatNetwork(
     kj::OneOf<MessageStream*, kj::Own<MessageStream>>&& stream,
     uint maxFdsPerMessage,
     rpc::twoparty::Side side,
-    ReaderOptions receiveOptions)
+    ReaderOptions receiveOptions,
+    const kj::MonotonicClock& clock)
 
     : stream(kj::mv(stream)),
       maxFdsPerMessage(maxFdsPerMessage),
       side(side),
       peerVatId(4),
       receiveOptions(receiveOptions),
-      previousWrite(kj::READY_NOW) {
+      previousWrite(kj::READY_NOW),
+      clock(clock),
+      currentOutgoingMessageSendTime(clock.now()) {
   peerVatId.initRoot<rpc::twoparty::VatId>().setSide(
       side == rpc::twoparty::Side::CLIENT ? rpc::twoparty::Side::SERVER
                                           : rpc::twoparty::Side::CLIENT);
@@ -48,25 +51,29 @@ TwoPartyVatNetwork::TwoPartyVatNetwork(
 }
 
 TwoPartyVatNetwork::TwoPartyVatNetwork(capnp::MessageStream& stream,
-                   rpc::twoparty::Side side, ReaderOptions receiveOptions)
-  : TwoPartyVatNetwork(stream, 0, side, receiveOptions) {}
+                   rpc::twoparty::Side side, ReaderOptions receiveOptions,
+                   const kj::MonotonicClock& clock)
+  : TwoPartyVatNetwork(stream, 0, side, receiveOptions, clock) {}
 
 TwoPartyVatNetwork::TwoPartyVatNetwork(
     capnp::MessageStream& stream,
     uint maxFdsPerMessage,
     rpc::twoparty::Side side,
-    ReaderOptions receiveOptions)
-    : TwoPartyVatNetwork(&stream, maxFdsPerMessage, side, receiveOptions) {}
+    ReaderOptions receiveOptions,
+    const kj::MonotonicClock& clock)
+    : TwoPartyVatNetwork(&stream, maxFdsPerMessage, side, receiveOptions, clock) {}
 
 TwoPartyVatNetwork::TwoPartyVatNetwork(kj::AsyncIoStream& stream, rpc::twoparty::Side side,
-                                       ReaderOptions receiveOptions)
+                                       ReaderOptions receiveOptions,
+                                       const kj::MonotonicClock& clock)
     : TwoPartyVatNetwork(kj::Own<MessageStream>(kj::heap<AsyncIoMessageStream>(stream)),
-                         0, side, receiveOptions) {}
+                         0, side, receiveOptions, clock) {}
 
 TwoPartyVatNetwork::TwoPartyVatNetwork(kj::AsyncCapabilityStream& stream, uint maxFdsPerMessage,
-                                       rpc::twoparty::Side side, ReaderOptions receiveOptions)
+                                       rpc::twoparty::Side side, ReaderOptions receiveOptions,
+                                       const kj::MonotonicClock& clock)
     : TwoPartyVatNetwork(kj::Own<MessageStream>(kj::heap<AsyncCapabilityMessageStream>(stream)),
-                         maxFdsPerMessage, side, receiveOptions) {}
+                         maxFdsPerMessage, side, receiveOptions, clock) {}
 
 MessageStream& TwoPartyVatNetwork::getStream() {
   KJ_SWITCH_ONEOF(stream) {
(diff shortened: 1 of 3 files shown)


kentonv:
Instead of maintaining a linked list, what if each message, when it begins writing itself to the socket, sets some field `currentOutgoingMessageCreationTime` to its creation time?

Do we care at all that this doesn't account for the messages that have been fully written to the socket buffer, but not yet read at the other end? I think for a unix socket that's typically 64kb worth of messages. In theory the sending end could use `ioctl(SIOCOUTQ)` to monitor how much data is in the buffer and remember what has been written in order to determine what message is at the front of the send queue, but I suppose that would be rather complicated... maybe not worth it.

a-robinson:
> Instead of maintaining a linked list, what if each message, when it begins writing itself to the socket, sets some field `currentOutgoingMessageCreationTime` to its creation time?

Yeah, something like this seems much simpler and similarly effective.

xortive:
> Instead of maintaining a linked list, what if each message, when it begins writing itself to the socket, sets some field `currentOutgoingMessageCreationTime` to its creation time?

Doh! I'm annoyed I missed this insight myself, I was not happy about having to add the linked list here :p I'll make the change!

> Do we care at all that this doesn't account for the messages that have been fully written to the socket buffer, but not yet read at the other end? I think for a unix socket that's typically 64kb worth of messages. In theory the sending end could use `ioctl(SIOCOUTQ)` to monitor how much data is in the buffer and remember what has been written in order to determine what message is at the front of the send queue, but I suppose that would be rather complicated... maybe not worth it.

I thought about it, but figuring out a point to actually check what's at the front of the send queue then find the equivalent message and when it was enqueued... seems pretty complicated.

c++/src/capnp/rpc-twoparty.c++
```diff
@@ -117,7 +124,8 @@ class TwoPartyVatNetwork::OutgoingMessageImpl final
 public:
   OutgoingMessageImpl(TwoPartyVatNetwork& network, uint firstSegmentWordSize)
       : network(network),
-        message(firstSegmentWordSize == 0 ? SUGGESTED_FIRST_SEGMENT_WORDS : firstSegmentWordSize) {}
+        message(firstSegmentWordSize == 0 ? SUGGESTED_FIRST_SEGMENT_WORDS : firstSegmentWordSize),
+        created(network.clock.now()) {}
```
a-robinson, line 128: It seems slightly more fair to set the timestamp based on when `send()` was called (since any waiting between the creation of the message and the calling of `send()` doesn't seem like the fault of the network), but I don't have any reason to believe it would make a particular difference in practice.

c++/src/capnp/rpc-twoparty.c++
```diff
@@ -172,12 +182,25 @@ public:
     return message.sizeInWords();
   }
 
+  kj::Duration getAge() {
```
a-robinson, line 185: Is this used by anything?

c++/src/capnp/rpc-twoparty.h
```diff
@@ -84,9 +91,14 @@ class TwoPartyVatNetwork: public TwoPartyVatNetworkBase,
   // to be sent on this connection. This may be useful for backpressure.
 
   size_t getCurrentQueueCount() { return currentQueueCount; }
-  // Get the count outgoing messages that are currently queued in memory waiting
+  // Get the count of outgoing messages that are currently queued in memory waiting
   // to be sent on this connection. This may be useful for backpressure.
 
+  kj::Duration getOldestQueuedMessageAge();
```
a-robinson, line 97: The naming here is a bit questionable, since it returns the age of the most recently *sent* message, not the oldest queued message. Or even if the name is ok, the comment could clarify what this value actually represents more accurately than by describing it as the "oldest outgoing message that is currently queued in memory waiting to be sent on this connection".

And doing the queue count check within here rather than always returning the age of the most recently sent message might theoretically be removing useful info -- maybe someone cares how recently a message was sent on the connection and this could be a good way to get that -- but at the same time there hasn't been any clear need for that functionality yet, so I suppose we don't need to worry about it.

a-robinson (approved):
LGTM other than a couple nits

c++/src/capnp/rpc-twoparty.c++
xortive: Probably doesn't matter in practice, but it's not hard to fix so I will

c++/src/capnp/rpc-twoparty.c++
xortive: nope, forgot to delete

c++/src/capnp/rpc-twoparty.h
kentonv: Yeah I'd argue the naming is reasonable, because if the next message in the queue _hasn't_ been sent yet, then we are in fact still in the process of sending the one before it, e.g. the previous message is only half-written to the socket. So in some sense it's still "queued"...

c++/src/capnp/rpc-twoparty.c++
kentonv: This could matter in practice. I have written code before that begins constructing a request, then waits for some further event before sending it.

kentonv:
Looks good once comments are addressed.

xortive:
removed `getAge()` and changed creation time to be set in `send()`

c++/src/capnp/rpc-twoparty.c++
```diff
@@ -150,8 +157,12 @@ public:
 
     network.previousWrite = KJ_ASSERT_NONNULL(network.previousWrite, "already shut down")
         .then([this]() {
-      return kj::evalNow([&]() { return network.getStream().writeMessage(fds, message); })
-          .catch_([this](kj::Exception&& e) {
+      return kj::evalNow([&]() {
+        created = network.clock.now();
```
a-robinson, line 161: Hm? Unless I'm misreading, isn't this too late to be setting `created`? We may have had to wait quite some time on `previousWrite` to resolve. I'd expect this variable to be set when `send()` is called, not after `previousWrite` is resolved.

We also don't really need `created` to be a member variable anymore, since it can just be declared as a local in `send()` and captured by the lambda that sets `network.currentOutgoingMessageCreationTime`. Although that doesn't really matter.

kentonv: Yeah we need to be setting `created` at the beginning of the `send()` method. (Maybe it should be called `sendTime`...)

xortive:
renamed `created` to `sendTime` and set it before waiting on `previousWrite`.

c++/src/capnp/rpc-twoparty.c++
```diff
@@ -148,10 +155,14 @@ public:
...
+    auto sendTime = network.clock.now();
     network.previousWrite = KJ_ASSERT_NONNULL(network.previousWrite, "already shut down")
-        .then([this]() {
-      return kj::evalNow([&]() { return network.getStream().writeMessage(fds, message); })
-          .catch_([this](kj::Exception&& e) {
+        .then([this, sendTime]() {
+      return kj::evalNow([&]() {
+        return network.getStream().writeMessage(fds, message);
+      }).then([this, sendTime]() {
+        network.currentOutgoingMessageSendTime = sendTime;
```
kentonv, line 164: Hmm, don't you want to set this _before_ starting the write?

c++/src/capnp/rpc-twoparty.h
```diff
@@ -84,9 +91,13 @@ class TwoPartyVatNetwork: public TwoPartyVatNetworkBase,
   // to be sent on this connection. This may be useful for backpressure.
 
   size_t getCurrentQueueCount() { return currentQueueCount; }
-  // Get the count outgoing messages that are currently queued in memory waiting
+  // Get the count of outgoing messages that are currently queued in memory waiting
   // to be sent on this connection. This may be useful for backpressure.
 
+  kj::Duration getOutgoingMessageWaitTime();
```
xortive, line 97: Updated the method name and comment here to reflect reality -- we aren't measuring the age of the oldest outgoing message, but how long it's been waiting to be written.

a-robinson (approved):
👍

Issue #1252, opened 2021-05-28
Segfault when running ASAN


vlovich:
Not sure the exact conditions, but running our unit tests under ASAN, I observe a segfault on Worker shutdown. I observe the root exception callback getting destroyed but then a new `ExceptionCallback` is constructed. The destruction of the RootCallback results in threadLocalPointer pointing to garbage memory. Similarly, clearing it out to null wouldn't do any good because `getExceptionCallback` will reference the wrong thing. I'm still investigating it but my hunch is that we should just maybe leak the default callback?

kentonv:
The root callback is a static local. If it's being destroyed, does that imply the process is already exiting, and running global destructors? This seems surprising since we tend to call `_exit()` to skip global destructors.

vlovich:
Yup. That’s exactly what happens (geez, without rr this would be more time consuming)

It’s an ordering problem. At exit we destroy the static local and then we destroy the TestCase which destroys a kj::ListLink which calls throwDestroyedWhileInList. The ordering should be reversed. This might be a problem in EW bin-test harness although the potential brittleness is concerning.

Need to look more into it but clearly appears to be some kind of static order initialization problem.

vlovich:
One way to solve this is to make the root callback global for the module instead of the function or using the constructor attribute. That carries a negative in that we always pay for initializing the root callback. Another alternative is to allocate it on the heap and leak the memory (and annotate it as OK so that LSAN is happy). Maybe just calling the getter that initializes it from the ListLink constructor might work too.

The other approach is to say it’s downstream’s problem to ensure they don’t have a global ListLink (or that if they do, that it’s initialized after the root callback).

My thought is that fixing ListLink to call that getter in the constructor or leaking the callback is a nice robust solution (specially leaking).

kentonv:
> One way to solve this is to make the root callback global for the module instead of the function

We have a strict no-global-constructors rule (i.e., no constructors that run at program startup) -- except for in tests, where it's allowed and used for test registrations.

> fixing ListLink to call that getter

This issue isn't in any way specific to ListLink. The general problem is that, at present, if a global destructor logs or throws (or calls anything that logs or throws), then it accesses the exception callback that has already been destroyed. Of course, a global destructor throwing is still a problem independent of this, but this obscures the underlying issue. It just happens to be that in this case, it's ListLink's destructor that is throwing.

I think the options are:
[...]

PR #1255, opened 2021-05-28
Sanitizer fixes

vlovich:
Fix the ASAN problem with RootExceptionCallback by intentionally leaking the pointer.
Fix the apparent leak in FiberPool::Impl (verified that kj/async-tests now pass).

Attempt a cleanup of the TODO in macros.h about RTTI/C++ exceptions. This one is
more "cause I had the macros & wanted to take a stab". Purely cosmetic & can easily
drop it from the stack if there's any concern.

### DIFF
--- a/c++/src/kj/async.c++
+++ b/c++/src/kj/async.c++
@@ -489,6 +489,13 @@ public:
       }
     }
 #endif
+
+    // Make sure we're not leaking anything from the global freelist either.
+    auto lock = freelist.lockExclusive();
+    auto dangling = kj::mv(*lock);
+    for (auto& stack: dangling) {
+      delete stack;
+    }
   }
 
   void setMaxFreelist(size_t count) {
--- a/c++/src/kj/exception.c++
+++ b/c++/src/kj/exception.c++
@@ -87,6 +87,24 @@
 #include <intrin.h>
 #endif
 
+#if KJ_HAS_COMPILER_FEATURE(address_sanitizer) || defined(__SANITIZE_ADDRESS__)
+#include <sanitizer/lsan_interface.h>
+#else
+static KJ_ALWAYS_INLINE(void __lsan_ignore_object(const void* p)) {}
+#endif
+// TODO(soon): Remove the LSAN stuff per https://github.com/capnproto/capnproto/pull/1255 feedback.
+
+namespace {
+template <typename T>
+inline T* lsanIgnoreObjectAndReturn(T* ptr) {
+  // Defensively lsan_ignore_object since the documentation doesn't explicitly specify what happens
+  // if you call this multiple times on the same object.
+  // TODO(soon): Remove this per https://github.com/capnproto/capnproto/pull/1255.
+  __lsan_ignore_object(ptr);
+  return ptr;
+}
+}
+
 namespace kj {
 
 StringPtr KJ_STRINGIFY(LogSeverity severity) {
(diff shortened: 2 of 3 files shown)


c++/src/kj/async.c++
```diff
@@ -489,6 +489,13 @@ public:
       }
     }
 #endif
+
+    // Make sure we're not leaking anything from the global freelist either.
```
kentonv, line 493: Doh, yes, this appears to be a legit leak.

(In practice `FiberPool`s are typically allocated globally, so the impact is minimal, but it's still definitely a bug.)

c++/src/kj/exception.c++
```diff
@@ -1070,9 +1076,30 @@ private:
...
+  // RootExceptionCallback's destructor (+ a separate pointer to share its value with this
+  // function). Since this would end up getting constructed during exit unwind, it would have the
+  // nice property of effectively being guaranteed to be evicted last.
+  //
+  // All this being said, I came back to leaking the object is the easiest tweak here:
+  //  * Can't go wrong
+  //  * Easy to maintain
+  //  * Throwing exceptions is bound to do be expensive and malloc-happy anyway, so the incremental
+  //    cost of 1 heap allocation is minimal.
+  __lsan_ignore_object(defaultCallback);
```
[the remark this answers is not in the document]
kentonv: Well, that seems more correct. What if LSAN is putting these pointers into a list somewhere, and each time you call __lsan_ignore_object() you're actually appending to that list? Then you've created a new kind of leak, since every time we call getExceptionCallback() -- which is quite often -- it's extending the list.

I don't know that it's implemented like that, but I also don't know that it isn't. Writing this code the right way seems easy, can we just do it? I'd like to have a habit of doing things right when it's easy to do so, rather than rely on assumptions that may or may not be accurate.

c++/src/kj/exception.c++
```diff
@@ -1070,9 +1076,30 @@ private:
...
-  static ExceptionCallback::RootExceptionCallback defaultCallback;
+  static auto defaultCallback = new ExceptionCallback::RootExceptionCallback();
+  // We allocate on the heap because some objects may throw in their destructors. If those objects
+  // had static storage, they might get fully constructed before the root callback. If they however
+  // then throw an exception during destruction, there would be a lifetime issue because their
+  // destructor would end up getting registered after the root callback's destructor. One solution
+  // is to just leak this pointer & allocate on first-use. The cost is that the initialization is
+  // mildly more expensive (+ we need to annotate sanitizers to ignore the problem). A great
+  // compiler annotation that would simply things would be one that allowed static variables to have
+  // their destruction omitted wholesale. That would allow us to avoid the heap but still have the
```
harrishancock, line 1087: > one that allowed static variables to have their destruction omitted wholesale.

I came up with this. Not sure about the smell, though -- I'm happy with the leak solution, too.

```cpp
ExceptionCallback& getExceptionCallback() {
  static union {
    ExceptionCallback::RootExceptionCallback defaultCallback;
  };

  // Use a second static to guarantee we call ctor() once.
  static auto initialized = [&defaultCallback]() { kj::ctor(defaultCallback); return true; }();

  ExceptionCallback* scoped = threadLocalCallback;
  return scoped != nullptr ? *scoped : defaultCallback;
}
```

vlovich: ooooo. I've never seen a trick like that. Filing that away. @kentonv do you like this better?

c++/src/kj/exception.c++
```diff
@@ -87,6 +87,22 @@
...
 
+#if KJ_HAS_COMPILER_FEATURE(address_sanitizer) || defined(__SANITIZE_ADDRESS__)
+#include <sanitizer/lsan_interface.h>
+#else
+static KJ_ALWAYS_INLINE(void __lsan_ignore_object(const void* p)) {}
+#endif
+
+namespace {
+template <typename T>
+KJ_ALWAYS_INLINE(T* lsan_ignore_object_and_return(T* ptr)) {
```
kentonv, line 98: Please use KJ naming style.

c++/src/kj/exception.c++
kentonv: I don't have a strong opinion.

The IIFE does seem like a nice way to accomplish one-time static initialization, compared to a separate global function.

vlovich: I thought about it & I know I tested the version as written but I don't have a convenient set up at the moment to retest the new code snippet. I'll leave a todo for now & implement this in a a follow-on PR

vlovich:
This fixed #1252

PR #1259, opened 2021-06-01
Accept TLS connections in their own loop

bcaimano:
This does the following:
- Adds a new general function ConnectionReceiver::run() to facilitate an async accept loop.
- Adds a queue and a TaskSet to the TlsConnectionReceiver to accept on its inner ConnectionReceiver without blocking on SSL_accept().
- Adds several new tests for the TlsConnectionReceiver.
- Converts several other TLS tests to use convenience helpers.

### DIFF
--- a/c++/src/kj/async-queue.h
+++ b/c++/src/kj/async-queue.h
@@ -1,0 +1,156 @@
+// Copyright (c) 2021 Cloudflare, Inc. and contributors
+// Licensed under the MIT License:
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+#pragma once
+
+#include "async.h"
+#include "common.h"
+#include "debug.h"
+#include "list.h"
+#include "memory.h"
+
+#include <list>
+
+KJ_BEGIN_HEADER
+
+namespace kj {
+
+template <typename T>
+class WaiterQueue {
+public:
+  // A WaiterQueue creates Nodes that blend newAdaptedPromise<T, Adaptor> and List<Node>.
+
+  WaiterQueue() = default;
+  KJ_DISALLOW_COPY(WaiterQueue);
+
+  Promise<T> wait() {
+    return newAdaptedPromise<T, Node>(queue);
+  }
+
+  void fulfill(T&& value) {
+    KJ_IREQUIRE(!empty());
+
+    auto& node = static_cast<Node&>(queue.front());
+    node.fulfiller.fulfill(kj::mv(value));
+    node.remove();
+  }
+
+  void reject(Exception&& exception) {
+    KJ_IREQUIRE(!empty());
+
+    auto& node = static_cast<Node&>(queue.front());
+    node.fulfiller.reject(kj::mv(exception));
+    node.remove();
+  }
+
+  bool empty() const {
+    return queue.empty();
+  }
+
+private:
+  struct BaseNode {
+    // This is a separate structure because List requires a predefined memory layout but
+    // newAdaptedPromise() only provides access to the Adaptor type in the ctor.
+
+    BaseNode(PromiseFulfiller<T>& fulfiller): fulfiller(fulfiller) {}
+
+    PromiseFulfiller<T>& fulfiller;
+    ListLink<BaseNode> link;
+  };
+
+  using Queue = List<BaseNode, &BaseNode::link>;
+
+  struct Node: public BaseNode {
+    Node(PromiseFulfiller<T>& fulfiller, Queue& queue): BaseNode(fulfiller), queue(queue) {
+      queue.add(*this);
+    }
+
+    ~Node() noexcept(false) {
+      // When the associated Promise is destructed, so is the Node thus we should leave the queue.
+      remove();
+    }
+
+    void remove() {
+      if(BaseNode::link.isLinked()){
+        queue.remove(*this);
+      }
+    }
+
+    Queue& queue;
+  };
+
+  Queue queue;
+};
+
+template <typename T>
+class ProducerConsumerQueue {
+  // ProducerConsumerQueue is an async FIFO queue.
+
+public:
+  void push(T v) {
+    // Push an existing value onto the queue.
+
+    if (!waiters.empty()) {
+      // We have at least one waiter, give the value to the oldest.
+      KJ_IASSERT(values.empty());
+
+      // Fulfill the first waiter and return without store our value.
+      waiters.fulfill(kj::mv(v));
+    } else {
+      // We don't have any waiters, store the value.
+      values.push_front(kj::mv(v));
+    }
+  }
+
+  void rejectAll(Exception e) {
+    // Reject all waiters with a given exception.
+
+    while (!waiters.empty()) {
+      auto newE = Exception(e);
+      waiters.reject(kj::mv(newE));
+    }
+  }
+
+  Promise<T> pop() {
+    // Eventually pop a value from the queue.
+    // Note that if your sinks lag your sources, the promise will always be ready.
+
+    if (!values.empty()) {
+      // We have at least one value, get the oldest.
+      KJ_IASSERT(waiters.empty());
+
+      auto value = kj::mv(values.back());
+      values.pop_back();
+      return kj::mv(value);
+    } else {
+      // We don't have any values, add ourselves to the waiting queue.
+      return waiters.wait();
+    }
+  }
+
+private:
+  std::list<T> values;
+  WaiterQueue<T> waiters;
+};
+
+}  // namespace kj
+
+KJ_END_HEADER
--- a/c++/src/kj/compat/tls-test.c++
+++ b/c++/src/kj/compat/tls-test.c++
@@ -428,6 +428,30 @@ struct TlsTest {
     options.trustedCertificates = kj::arrayPtr(&caCert, 1);
     return options;
   }
+
+  Promise<void> writeToServer(AsyncIoStream& client) {
+    return client.write("foo", 4);
+  }
+
+  Promise<void> readFromClient(AsyncIoStream& server) {
+    auto buf = heapArray<char>(4);
+
+    auto readPromise = server.read(buf.begin(), buf.size());
+
+    auto checkBuffer = [buf = kj::mv(buf)]() {
+      KJ_ASSERT(kj::StringPtr(buf.begin(), buf.end()-1) == kj::StringPtr("foo"));
+    };
+
+    return readPromise.then(kj::mv(checkBuffer));
+  }
+
+  void testConnection(AsyncIoStream& client, AsyncIoStream& server) {
+    auto writePromise = writeToServer(client);
+    auto readPromise = readFromClient(server);
+
+    writePromise.wait(io.waitScope);
+    readPromise.wait(io.waitScope);
+  };
 };
 
 KJ_TEST("TLS basics") {
@@ -442,12 +466,7 @@ KJ_TEST("TLS basics") {
   auto client = clientPromise.wait(test.io.waitScope);
   auto server = serverPromise.wait(test.io.waitScope);
 
-  auto writePromise = client->write("foo", 3);
-  char buf[4];
-  server->read(&buf, 3).wait(test.io.waitScope);
-  buf[3] = '\0';
-
-  KJ_ASSERT(kj::StringPtr(buf) == "foo");
+  test.testConnection(*client, *server);
 }
 
 KJ_TEST("TLS peer identity") {
@@ -487,12 +506,7 @@ KJ_TEST("TLS peer identity") {
     KJ_EXPECT(id->toString() == "(anonymous client)");
   }
 
-  auto writePromise = client.stream->write("foo", 3);
-  char buf[4];
-  server.stream->read(&buf, 3).wait(test.io.waitScope);
-  buf[3] = '\0';
-
-  KJ_ASSERT(kj::StringPtr(buf) == "foo");
+  test.testConnection(*client.stream, *server.stream);
 }
 
 KJ_TEST("TLS multiple messages") {
@@ -648,12 +662,7 @@ KJ_TEST("TLS SNI") {
   auto client = clientPromise.wait(test.io.waitScope);
   auto server = serverPromise.wait(test.io.waitScope);
 
-  auto writePromise = client->write("foo", 3);
-  char buf[4];
-  server->read(&buf, 3).wait(test.io.waitScope);
-  buf[3] = '\0';
-
-  KJ_ASSERT(kj::StringPtr(buf) == "foo");
+  test.testConnection(*client, *server);
 
   KJ_ASSERT(callback.callCount == 1);
 }
@@ -775,12 +784,7 @@ KJ_TEST("TLS client certificate verification") {
     KJ_ASSERT(id->hasCertificate());
     KJ_EXPECT(id->getCommonName() == "example.net");
 
-    auto writePromise = client->write("foo", 3);
-    char buf[4];
-    server.stream->read(&buf, 3).wait(test.io.waitScope);
-    buf[3] = '\0';
-
-    KJ_ASSERT(kj::StringPtr(buf) == "foo");
+    test.testConnection(*client, *server.stream);
   }
 
   // If verifyClients is off, client certificate is ignored, even if trusted.
@@ -917,35 +921,31 @@ public:
 KJ_TEST("TLS receiver basics") {
   TlsReceiverTest test;
 
-  auto wrapClient = [&](auto stream) {
+  auto clientPromise = test.baseReceiver->connect().then([&](auto stream) {
     return test.tlsClient.wrapClient(kj::mv(stream), "example.com");
-  };
+  });
+  auto serverPromise = test.receiver->accept();
 
-  auto writeToServer = [&](auto client) {
-    return test.writeToServer(kj::mv(client));
-  };
+  auto client = clientPromise.wait(test.io.waitScope);
+  auto server = serverPromise.wait(test.io.waitScope);
 
-  auto readFromClient = [&](auto server) {
-    return test.readFromClient(kj::mv(server));
-  };
-
-  auto clientPromise = test.baseReceiver->connect().then(wrapClient).then(writeToServer);
-  auto serverPromise = test.receiver->accept().then(readFromClient);
-
-  clientPromise.wait(test.io.waitScope);
-  serverPromise.wait(test.io.waitScope);
+  test.testConnection(*client, *server);
 }
 
 KJ_TEST("TLS receiver experiences pre-TLS error") {
   TlsReceiverTest test;
 
-  auto endPromise = test.receiver->onFatalError();
+  KJ_LOG(INFO, "Accepting before a bad connect");
+  auto promise = test.receiver->accept();
 
   KJ_LOG(INFO, "Disappointing our server");
-  KJ_EXPECT_LOG(ERROR, "TlsConnectionReceiver could not accept()");
   test.baseReceiver->badConnect();
 
-  KJ_EXPECT_THROW_RECOVERABLE_MESSAGE("Pipes are leaky", endPromise.wait(test.io.waitScope));
+  KJ_EXPECT_THROW_RECOVERABLE_MESSAGE("Pipes are leaky", promise.wait(test.io.waitScope));
+
+  KJ_LOG(INFO, "Trying to load a promise after failure");
+  KJ_EXPECT_THROW_RECOVERABLE_MESSAGE(
+      "Pipes are leaky", test.receiver->accept().wait(test.io.waitScope));
 }
 
 KJ_TEST("TLS receiver accepts multiple clients") {
@@ -956,11 +956,11 @@ KJ_TEST("TLS receiver accepts multiple clients") {
   };
 
   auto writeToServer = [&](auto client) {
-    return test.writeToServer(kj::mv(client));
+    return test.writeToServer(*client).attach(kj::mv(client));
   };
 
   auto readFromClient = [&](auto server) {
-    return test.readFromClient(kj::mv(server));
+    return test.readFromClient(*server).attach(kj::mv(server));
   };
 
   KJ_LOG(INFO, "Requesting a bunch of client connects");
@@ -993,11 +993,11 @@ KJ_TEST("TLS receiver does not stall on client that disconnects before ssl hands
   };
 
   auto writeToServer = [&](auto client) {
-    return test.writeToServer(kj::mv(client));
+    return test.writeToServer(*client).attach(kj::mv(client));
   };
 
   auto readFromClient = [&](auto server) {
-    return test.readFromClient(kj::mv(server));
+    return test.readFromClient(*server).attach(kj::mv(server));
   };
 
   constexpr auto kClientCount = 20;
@@ -1010,7 +1010,6 @@ KJ_TEST("TLS receiver does not stall on client that disconnects before ssl hands
   }
 
   KJ_LOG(INFO, "Requesting and resolving a client connect that hangs up before ssl connect");
-  KJ_EXPECT_LOG(ERROR, "Client disconnected during SSL_accept");
   KJ_ASSERT(test.baseReceiver->connect().wait(test.io.waitScope));
 
   KJ_LOG(INFO, "Requesting the second batch of client connects in parallel");
@@ -1040,11 +1039,11 @@ KJ_TEST("TLS receiver does not stall on hung client") {
   };
 
   auto writeToServer = [&](auto client) {
-    return test.writeToServer(kj::mv(client));
+    return test.writeToServer(*client).attach(kj::mv(client));
   };
 
   auto readFromClient = [&](auto server) {
-    return test.readFromClient(kj::mv(server));
+    return test.readFromClient(*server).attach(kj::mv(server));
   };
 
   constexpr auto kClientCount = 20;
@@ -1075,7 +1074,6 @@ KJ_TEST("TLS receiver does not stall on hung client") {
   joinPromises(clientPromises.releaseAsArray()).wait(test.io.waitScope);
 
   KJ_LOG(INFO, "Releasing the hung client");
-  KJ_EXPECT_LOG(ERROR, "Client disconnected during SSL_accept");
   hungClient = {};
 
   KJ_LOG(INFO, "Requesting one last server accept that we'll never resolve");
--- a/c++/src/kj/compat/tls-test.c++
+++ b/c++/src/kj/compat/tls-test.c++
@@ -26,17 +26,21 @@
 #endif
 
 #include "tls.h"
-#include <kj/test.h>
-#include <kj/async-io.h>
-#include <stdlib.h>
+
 #include <openssl/opensslv.h>
+
+#include <stdlib.h>
 
 #if _WIN32
 #include <winsock2.h>
+
 #include <kj/windows-sanity.h>
 #else
 #include <sys/socket.h>
 #endif
+
+#include <kj/async-io.h>
+#include <kj/test.h>
 
 namespace kj {
 namespace {
--- a/c++/src/kj/compat/tls-test.c++
+++ b/c++/src/kj/compat/tls-test.c++
@@ -818,6 +818,267 @@ KJ_TEST("TLS client certificate verification") {
   }
 }
 
+class MockConnectionReceiver final: public ConnectionReceiver {
+  // This connection receiver allows mocked async connection establishment without the network.
+
+  struct ClientRequest {
+    Maybe<Exception> maybeException;
+    Own<PromiseFulfiller<Own<AsyncIoStream>>> clientFulfiller;
+  };
+
+public:
+  MockConnectionReceiver(AsyncIoProvider& provider): provider(provider) {}
+
+  Promise<Own<AsyncIoStream>> accept() override {
+    return acceptImpl();
+  }
+
+  Promise<AuthenticatedStream> acceptAuthenticated() override {
+    return acceptImpl().then([](auto stream) -> AuthenticatedStream {
+      return { kj::mv(stream), LocalPeerIdentity::newInstance({}) };
+    });
+  }
+
+  uint getPort() override {
+    return 0;
+  }
+  void getsockopt(int, int, void*, uint*) override {}
+  void setsockopt(int, int, const void*, uint) override {}
+
+  Promise<Own<AsyncIoStream>> connect() {
+    // Mock a new successful connection to our receiver.
+    return connectImpl();
+  }
+
+  Promise<void> badConnect() {
+    // Mock a new failed connection to our receiver.
+    return connectImpl(KJ_EXCEPTION(DISCONNECTED, "Pipes are leaky")).ignoreResult();
+  }
+
+private:
+  Promise<Own<AsyncIoStream>> acceptImpl() {
+    if (clientRequests.empty()) {
+      KJ_ASSERT(!serverFulfiller);
+      auto paf = newPromiseAndFulfiller<void>();
+      serverFulfiller = kj::mv(paf.fulfiller);
+      return paf.promise.then([this] {
+        return acceptImpl();
+      });
+    }
+
+    // This is accepting in FILO order, it shouldn't matter in practice.
+    auto request = kj::mv(clientRequests.back());
+    clientRequests.removeLast();
+
+    KJ_IF_MAYBE(exception, kj::mv(request.maybeException)) {
+      request.clientFulfiller = nullptr;  // The other end had an issue, break the promise.
+      return kj::mv(*exception);
+    } else {
+      auto pipe = provider.newTwoWayPipe();
+      request.clientFulfiller->fulfill(kj::mv(pipe.ends[0]));
+      return kj::mv(pipe.ends[1]);
+    }
+  }
+
+  Promise<Own<AsyncIoStream>> connectImpl(Maybe<Exception> maybeException = nullptr) {
+    auto paf = newPromiseAndFulfiller<Own<AsyncIoStream>>();
+    clientRequests.add(ClientRequest{ kj::mv(maybeException), kj::mv(paf.fulfiller) });
+
+    if (auto fulfiller = kj::mv(serverFulfiller)) {
+      fulfiller->fulfill();
+    }
+
+    return kj::mv(paf.promise);
+  }
+
+  AsyncIoProvider& provider;
+
+  Own<PromiseFulfiller<void>> serverFulfiller;
+  Vector<ClientRequest> clientRequests;
+};
+
+class TlsReceiverTest final: public TlsTest {
+  // TlsReceiverTest augments TlsTest to test TlsConnectionReceiver.
+public:
+  TlsReceiverTest(): TlsTest() {
+    auto baseReceiverPtr = kj::heap<MockConnectionReceiver>(*io.provider);
+    baseReceiver = baseReceiverPtr.get();
+    receiver = tlsServer.wrapPort(kj::mv(baseReceiverPtr));
+  }
+
+  Own<ConnectionReceiver> receiver;
+  MockConnectionReceiver* baseReceiver;
+};
+
+KJ_TEST("TLS receiver basics") {
+  TlsReceiverTest test;
+
+  auto wrapClient = [&](auto stream) {
+    return test.tlsClient.wrapClient(kj::mv(stream), "example.com");
+  };
+
+  auto writeToServer = [&](auto client) {
+    return test.writeToServer(kj::mv(client));
+  };
+
+  auto readFromClient = [&](auto server) {
+    return test.readFromClient(kj::mv(server));
+  };
+
+  auto clientPromise = test.baseReceiver->connect().then(wrapClient).then(writeToServer);
+  auto serverPromise = test.receiver->accept().then(readFromClient);
+
+  clientPromise.wait(test.io.waitScope);
+  serverPromise.wait(test.io.waitScope);
+}
+
+KJ_TEST("TLS receiver experiences pre-TLS error") {
+  TlsReceiverTest test;
+
+  auto endPromise = test.receiver->onFatalError();
+
+  KJ_LOG(INFO, "Disappointing our server");
+  KJ_EXPECT_LOG(ERROR, "TlsConnectionReceiver could not accept()");
+  test.baseReceiver->badConnect();
+
+  KJ_EXPECT_THROW_RECOVERABLE_MESSAGE("Pipes are leaky", endPromise.wait(test.io.waitScope));
+}
+
+KJ_TEST("TLS receiver accepts multiple clients") {
+  TlsReceiverTest test;
+
+  auto wrapClient = [&](auto stream) {
+    return test.tlsClient.wrapClient(kj::mv(stream), "example.com");
+  };
+
+  auto writeToServer = [&](auto client) {
+    return test.writeToServer(kj::mv(client));
+  };
+
+  auto readFromClient = [&](auto server) {
+    return test.readFromClient(kj::mv(server));
+  };
+
+  KJ_LOG(INFO, "Requesting a bunch of client connects");
+  constexpr auto kClientCount = 20;
+  auto clientPromises = Vector<Promise<void>>();
+  for (auto i = 0; i < kClientCount; ++i) {
+    auto clientPromise = test.baseReceiver->connect().then(wrapClient).then(writeToServer);
+    clientPromises.add(kj::mv(clientPromise));
+  }
+
+  KJ_LOG(INFO, "Requesting and resolving a bunch of server accepts in sequence");
+  for (auto i = 0; i < kClientCount; ++i) {
+    // Resolve each receive in sequence like the Supervisor/Network.
+    test.receiver->accept().then(readFromClient).wait(test.io.waitScope);
+  }
+
+  KJ_LOG(INFO, "Resolving all of our client connects in parallel");
+  joinPromises(clientPromises.releaseAsArray()).wait(test.io.waitScope);
+
+  KJ_LOG(INFO, "Requesting one last server accept that we'll never resolve");
+  auto extraAcceptPromise = test.receiver->accept().then(readFromClient);
+  KJ_EXPECT(!extraAcceptPromise.poll(test.io.waitScope));
+}
+
+KJ_TEST("TLS receiver does not stall on client that disconnects before ssl handshake") {
+  TlsReceiverTest test;
+
+  auto wrapClient = [&](auto stream) {
+    return test.tlsClient.wrapClient(kj::mv(stream), "example.com");
+  };
+
+  auto writeToServer = [&](auto client) {
+    return test.writeToServer(kj::mv(client));
+  };
+
+  auto readFromClient = [&](auto server) {
+    return test.readFromClient(kj::mv(server));
+  };
+
+  constexpr auto kClientCount = 20;
+  auto clientPromises = Vector<Promise<void>>();
+
+  KJ_LOG(INFO, "Requesting the first batch of client connects in parallel");
+  for (auto i = 0; i < kClientCount / 2; ++i) {
+    auto clientPromise = test.baseReceiver->connect().then(wrapClient).then(writeToServer);
+    clientPromises.add(kj::mv(clientPromise));
+  }
+
+  KJ_LOG(INFO, "Requesting and resolving a client connect that hangs up before ssl connect");
+  KJ_EXPECT_LOG(ERROR, "Client disconnected during SSL_accept");
+  KJ_ASSERT(test.baseReceiver->connect().wait(test.io.waitScope));
+
+  KJ_LOG(INFO, "Requesting the second batch of client connects in parallel");
+  for (auto i = 0; i < kClientCount / 2; ++i) {
+    auto clientPromise = test.baseReceiver->connect().then(wrapClient).then(writeToServer);
+    clientPromises.add(kj::mv(clientPromise));
+  }
+
+  KJ_LOG(INFO, "Requesting and resolving a bunch of server accepts in sequence");
+  for (auto i = 0; i < kClientCount; ++i) {
+    test.receiver->accept().then(readFromClient).wait(test.io.waitScope);
+  }
+
+  KJ_LOG(INFO, "Resolving all of our client connects in parallel");
+  joinPromises(clientPromises.releaseAsArray()).wait(test.io.waitScope);
+
+  KJ_LOG(INFO, "Requesting one last server accept that we'll never resolve");
+  auto extraAcceptPromise = test.receiver->accept().then(readFromClient);
+  KJ_EXPECT(!extraAcceptPromise.poll(test.io.waitScope));
+}
+
+KJ_TEST("TLS receiver does not stall on hung client") {
+  TlsReceiverTest test;
+
+  auto wrapClient = [&](auto stream) {
+    return test.tlsClient.wrapClient(kj::mv(stream), "example.com");
+  };
+
+  auto writeToServer = [&](auto client) {
+    return test.writeToServer(kj::mv(client));
+  };
+
+  auto readFromClient = [&](auto server) {
+    return test.readFromClient(kj::mv(server));
+  };
+
+  constexpr auto kClientCount = 20;
+  auto clientPromises = Vector<Promise<void>>();
+
+  KJ_LOG(INFO, "Requesting the first batch of client connects in parallel");
+  for (auto i = 0; i < kClientCount / 2; ++i) {
+    auto clientPromise = test.baseReceiver->connect().then(wrapClient).then(writeToServer);
+    clientPromises.add(kj::mv(clientPromise));
+  }
+
+  KJ_LOG(INFO, "Requesting and resolving a client connect that never does ssl connect");
+  auto hungClient = test.baseReceiver->connect().wait(test.io.waitScope);
+  KJ_ASSERT(hungClient);
+
+  KJ_LOG(INFO, "Requesting the second batch of client connects in parallel");
+  for (auto i = 0; i < kClientCount / 2; ++i) {
+    auto clientPromise = test.baseReceiver->connect().then(wrapClient).then(writeToServer);
+    clientPromises.add(kj::mv(clientPromise));
+  }
+
+  KJ_LOG(INFO, "Requesting and resolving a bunch of server accepts in sequence");
+  for (auto i = 0; i < kClientCount; ++i) {
+    test.receiver->accept().then(readFromClient).wait(test.io.waitScope);
+  }
+
+  KJ_LOG(INFO, "Resolving all of our client connects in parallel");
+  joinPromises(clientPromises.releaseAsArray()).wait(test.io.waitScope);
+
+  KJ_LOG(INFO, "Releasing the hung client");
+  KJ_EXPECT_LOG(ERROR, "Client disconnected during SSL_accept");
+  hungClient = {};
+
+  KJ_LOG(INFO, "Requesting one last server accept that we'll never resolve");
+  auto extraAcceptPromise = test.receiver->accept().then(readFromClient);
+  KJ_EXPECT(!extraAcceptPromise.poll(test.io.waitScope));
+}
+
 #ifdef KJ_EXTERNAL_TESTS
 KJ_TEST("TLS to capnproto.org") {
   kj::AsyncIoContext io = setupAsyncIo();
--- a/c++/src/kj/compat/tls.c++
+++ b/c++/src/kj/compat/tls.c++
@@ -34,6 +34,7 @@
 #include <openssl/x509.h>
 #include <openssl/x509v3.h>
 
+#include <kj/async-queue.h>
 #include <kj/debug.h>
 #include <kj/vector.h>
 
@@ -162,11 +163,20 @@ public:
       }
     });
   }
+
   kj::Promise<void> accept() {
     // We are the server. Set SSL options to prefer server's cipher choice.
     SSL_set_options(ssl, SSL_OP_CIPHER_SERVER_PREFERENCE);
 
-    return sslCall([this]() { return SSL_accept(ssl); }).ignoreResult();
+    auto acceptPromise = sslCall([this]() {
+      return SSL_accept(ssl);
+    });
+    return acceptPromise.then([](size_t ret) {
+      if (ret == 0) {
+        kj::throwRecoverableException(
+            KJ_EXCEPTION(DISCONNECTED, "Client disconnected during SSL_accept()"));
+      }
+    });
   }
 
   kj::Own<TlsPeerIdentity> getIdentity(kj::Own<kj::PeerIdentity> inner) {
@@ -317,7 +327,7 @@ private:
             // According to documentation we shouldn't get here, because our BIO never returns an
             // "error". But in practice we do get here sometimes when the peer disconnects
             // prematurely.
-            KJ_FAIL_ASSERT("TLS protocol error");
+            return KJ_EXCEPTION(DISCONNECTED, "SSL unable to continue I/O");
           }
         default:
           KJ_FAIL_ASSERT("unexpected SSL error code", error);
@@ -407,29 +417,37 @@ private:
 // =======================================================================================
 // Implementations of ConnectionReceiver, NetworkAddress, and Network as wrappers adding TLS.
 
-class TlsConnectionReceiver final: public kj::ConnectionReceiver {
+class TlsConnectionReceiver final: public ConnectionReceiver, public TaskSet::ErrorHandler {
 public:
-  TlsConnectionReceiver(TlsContext& tls, kj::Own<kj::ConnectionReceiver> inner)
-      : tls(tls), inner(kj::mv(inner)) {}
+  TlsConnectionReceiver(TlsContext &tls, Own<ConnectionReceiver> inner)
+      : tls(tls), inner(kj::mv(inner)),
+        acceptLoopTask(acceptLoop().eagerlyEvaluate([this](Exception &&e) {
+          onAcceptFailure(kj::mv(e));
+        })),
+        tasks(*this) {}
+
+  void taskFailed(Exception&& e) override {
+    // TODO(soon) SSL connection failures may be a fact of normal operation but they may also be
+    // important diagnostic information. We should allow for an error handler to be passed in so
+    // that network issues that affect TLS can be more discoverable from the server side.
+    if (e.getType() != Exception::Type::DISCONNECTED) {
+      KJ_LOG(ERROR, "error accepting tls connection", kj::mv(e));
+    }
+  };
 
   Promise<Own<AsyncIoStream>> accept() override {
-    return inner->accept().then([this](kj::Own<AsyncIoStream> stream) {
-      return kj::evalNow([&] { return tls.wrapServer(kj::mv(stream)); })
-          .catch_([this](kj::Exception&& e) {
-        KJ_LOG(ERROR, "error accepting tls connection", e);
-        return accept();
-      });
+    return acceptAuthenticated().then([](AuthenticatedStream&& stream) {
+      return kj::mv(stream.stream);
     });
   }
 
-  Promise<kj::AuthenticatedStream> acceptAuthenticated() override {
-    return inner->acceptAuthenticated().then([this](kj::AuthenticatedStream stream) {
-      return kj::evalNow([&] { return tls.wrapServer(kj::mv(stream)); })
-          .catch_([this](kj::Exception&& e) {
-        KJ_LOG(ERROR, "error accepting tls connection", e);
-        return acceptAuthenticated();
-      });
-    });
+  Promise<AuthenticatedStream> acceptAuthenticated() override {
+    KJ_IF_MAYBE(e, maybeInnerException) {
+      // We've experienced an exception from the inner receiver, we consider this unrecoverable.
+      return Exception(*e);
+    }
+
+    return queue.pop();
   }
 
   uint getPort() override {
@@ -445,8 +463,50 @@ public:
   }
 
 private:
+  void onAcceptSuccess(AuthenticatedStream&& stream) {
+    // Queue this stream to go through SSL_accept.
+
+    auto acceptPromise = kj::evalNow([&] {
+      // Do the SSL acceptance procedure.
+      return tls.wrapServer(kj::mv(stream));
+    });
+
+    auto sslPromise = acceptPromise.then([this](auto&& stream) -> Promise<void> {
+      // This is only attached to the success path, thus the error handler will catch if our
+      // promise fails.
+      queue.push(kj::mv(stream));
+      return kj::READY_NOW;
+    });
+    tasks.add(kj::mv(sslPromise));
+  }
+
+  void onAcceptFailure(Exception&& e) {
+    // Store this exception to reject all future calls to accept() and reject any unfulfilled
+    // promises from the queue.
+    maybeInnerException = kj::mv(e);
+    queue.rejectAll(Exception(KJ_REQUIRE_NONNULL(maybeInnerException)));
+  }
+
+  Promise<void> acceptLoop() {
+    // Accept one connection and queue up the next accept on our TaskSet.
+
+    return inner->acceptAuthenticated().then(
+        [this](AuthenticatedStream&& stream) {
+      onAcceptSuccess(kj::mv(stream));
+
+      // Queue up the next accept loop immediately without waiting for SSL_accept()/wrapServer().
+      return acceptLoop();
+    });
+  }
+
   TlsContext& tls;
-  kj::Own<kj::ConnectionReceiver> inner;
+  Own<ConnectionReceiver> inner;
+
+  Promise<void> acceptLoopTask;
+  ProducerConsumerQueue<AuthenticatedStream> queue;
+  TaskSet tasks;
+
+  Maybe<Exception> maybeInnerException;
 };
 
 class TlsNetworkAddress final: public kj::NetworkAddress {
--- a/c++/src/kj/compat/tls.c++
+++ b/c++/src/kj/compat/tls.c++
@@ -22,19 +22,20 @@
 #if KJ_HAS_OPENSSL
 
 #include "tls.h"
+
 #include "readiness-io.h"
+
 #include <openssl/bio.h>
-#include <openssl/ssl.h>
-#include <openssl/err.h>
-#include <openssl/x509.h>
-#include <openssl/x509v3.h>
-#include <openssl/evp.h>
 #include <openssl/conf.h>
+#include <openssl/err.h>
+#include <openssl/evp.h>
 #include <openssl/ssl.h>
 #include <openssl/tls1.h>
+#include <openssl/x509.h>
+#include <openssl/x509v3.h>
+
 #include <kj/debug.h>
 #include <kj/vector.h>
-
 
 #if OPENSSL_VERSION_NUMBER < 0x10100000L
 #define BIO_set_init(x,v)          (x->init=v)
--- a/c++/src/kj/compat/tls.c++
+++ b/c++/src/kj/compat/tls.c++
@@ -35,6 +35,7 @@
 #include <kj/debug.h>
 #include <kj/vector.h>
 
+
 #if OPENSSL_VERSION_NUMBER < 0x10100000L
 #define BIO_set_init(x,v)          (x->init=v)
 #define BIO_get_data(x)            (x->ptr)
--- a/c++/src/kj/async-queue-test.c++
+++ b/c++/src/kj/async-queue-test.c++
@@ -1,0 +1,147 @@
+// Copyright (c) 2021 Cloudflare, Inc. and contributors
+// Licensed under the MIT License:
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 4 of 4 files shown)


c++/src/kj/compat/tls-test.c++
```diff
@@ -818,6 +822,241 @@ KJ_TEST("TLS client certificate verification") {
...
+};
+
+KJ_TEST("TLS receiver basics") {
+  TlsReceiverTest test;
+  ErrorNexus e;
+
+  test.receiver->run();
+
+  auto clientPromise = test.baseReceiver->connect()
+                           .then([&](auto stream) {
```
a-robinson, line 924: The indentation (and length of the next line) here is a bit extreme. As per [the style guide](https://github.com/capnproto/capnproto/blob/master/style-guide.md#spacing-and-bracing), we generally indent these kinds of `.then()` calls by either 0 or 4 spaces relative to the previous line. It does say "or line them up nicely with the previous line if it makes it easier to read", but I almost never see that done, especially not with this much indentation involved.

c++/src/kj/compat/tls-test.c++
```diff
@@ -818,6 +822,241 @@ KJ_TEST("TLS client certificate verification") {
...
+                             return test.tlsClient.wrapClient(kj::mv(stream), "example.com");
+                           })
+                           .then([&](auto client) {
+                             return test.writeToServer(kj::mv(client));
```
a-robinson, line 928: Not important, but I'm curious is there some reason why this doesn't use the `test.testConnection()` helper you added?

c++/src/kj/compat/tls-test.c++
```diff
@@ -818,6 +822,241 @@ KJ_TEST("TLS client certificate verification") {
...
+    receiver = tlsServer.wrapPort(kj::mv(baseReceiverPtr));
+  }
+
+  Own<ConnectionReceiver> receiver;
+  MockConnectionReceiver* baseReceiver;
+};
+
+KJ_TEST("TLS receiver basics") {
+  TlsReceiverTest test;
+  ErrorNexus e;
```
a-robinson, line 919: Super nit, but as far as I can tell none of the new tests use the `ErrorNexus` so it seems worth dropping these declarations for easier readability.

c++/src/kj/compat/tls-test.c++
```diff
@@ -818,6 +822,241 @@ KJ_TEST("TLS client certificate verification") {
...
+        .then([&](auto server) {
+          return test.readFromClient(kj::mv(server));
+        })
+        .wait(test.io.waitScope);
+  }
+
+  KJ_LOG(INFO, "Resolving all of our client connects in parallel");
+  joinPromises(clientPromises.releaseAsArray()).wait(test.io.waitScope);
+
+  KJ_LOG(INFO, "Requesting one last server accept that we'll never resolve");
```
a-robinson, line 999: If we want to at least minimally test that this doesn't resolve, you could add a `KJ_EXPECT(!extraAcceptPromise.poll(test.io.waitScope));` to the last line of the test.

c++/src/kj/compat/tls-test.c++
```diff
@@ -818,6 +822,241 @@ KJ_TEST("TLS client certificate verification") {
...
+                       })
+                       .then([&](auto client) {
+                         return test.writeToServer(kj::mv(client));
+                       });
+    clientPromises.add(kj::mv(promise));
+  }
+
+  KJ_LOG(INFO, "Requesting and resolving a client connect that never does ssl connect");
+  KJ_EXPECT_LOG(ERROR, "Client disconnected during SSL_accept");
+  KJ_ASSERT(test.baseReceiver->connect().wait(test.io.waitScope));
```
a-robinson, line 1028: Perhaps this is overly pedantic, but this doesn't cause a "hung client" as suggested by the test name, it causes a client that connects but then immediately disconnects before the SSL handshake.

This is a great test to include, but it seems worth also adding a case here where we don't just throw away the promise returned from `connect()`, but rather hold onto it to avoid the connection being disconnected.

a-robinson (commented):
Looks like you adapted to the code base quite quickly :) Apologies for the very picky comments, I just want to make sure you're aware of the general style guidelines early on.

c++/src/kj/compat/tls-test.c++
```diff
@@ -424,6 +427,27 @@ struct TlsTest {
...
   }
+
+  Promise<void> writeToServer(Own<AsyncIoStream> client) {
+    return client->write("foo\0", 4).attach(kj::mv(client));
+  }
+
+  Promise<void> readFromClient(Own<AsyncIoStream> server) {
+    auto buf = heapArray<char>(4);
+    return server->read(buf.begin(), buf.size())
+        .attach(kj::mv(server))  //
```
bretthoerner, line 438: Super minor, I was curious if you meant to comment about something here or if this `//` was just leftover.

c++/src/kj/compat/tls.c++
```diff
@@ -443,8 +531,31 @@ public:
   }
 
 private:
+  Promise<void> doOneAccept() {
+    // Accept one connection and queue up the next accept on our TaskSet
+
+    return inner->acceptAuthenticated().then(
+        [this](AuthenticatedStream&& stream) {
+          auto promise = kj::evalNow([&] {
```
bretthoerner, line 539: @a-robinson already mentioned the indentation style, but just to add one more that can be odd to newcomers for a while (or was to me): we de-dent the body of the continuation so that it's effectively 2 spaces in from main function body where it's being defined, so this `auto promise` would be de-dented 4 spaces, at which point you may decide to put the lambda capture and parameter list up right next to the `then` -- up to you.

```
return asyncFn().then(
    [cap1,cap2](auto p1, auto p2) {
  auto actualFnBodyHere = foo();
});
```

Or:
```
return asyncFn().then([cap1,cap2](auto p1, auto p2) {
  auto actualFnBodyHere = foo();
});
```

bretthoerner (commented):
I hate to only add nits, but @a-robinson covered everything else that I saw. Looks good. :)

c++/src/kj/compat/tls.c++
```diff
@@ -160,11 +161,21 @@ public:
...
   kj::Promise<void> accept() {
     // We are the server. Set SSL options to prefer server's cipher choice.
     SSL_set_options(ssl, SSL_OP_CIPHER_SERVER_PREFERENCE);
 
-    return sslCall([this]() { return SSL_accept(ssl); }).ignoreResult();
+    return sslCall([this]() {
+             return SSL_accept(ssl);
+           })
+        .then([](size_t ret) {
+          if (ret == 0) {
```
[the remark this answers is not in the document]
bcaimano: I did run into it, but only because I had been seeing what I could do with BIO_read. I think it's possible but I don't think it's likely we see it in practice.

c++/src/kj/compat/tls.c++
```diff
@@ -405,31 +416,108 @@ private:
...
+      return kj::fwd<Own<AsyncIoStream>>(stream.stream);
     });
   }
 
+  Promise<AuthenticatedStream> acceptAuthenticated() override {
+    return queue.pop();
+  }
+
+  Promise<void> run() override {
+    // TODO(pr) We could get away without this function by adding the task in the ctor.
```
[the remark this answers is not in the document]
bcaimano: I must admit that I lean towards defensive programming if it seems "free". I'm generally conflicted about this choice. On one end, it lays a foundation for a flexible accept loop in other ConnectionReceivers and I've see TODOs mentioning how that would be nice. It also allows for deferred init, which can be very nice for strictly ordered initialization. On the other hand, we could get away without changes in repos that use KJ if we start it in the ctor.

c++/src/kj/compat/tls.c++
```diff
@@ -443,8 +531,31 @@ public:
   }
 
 private:
+  Promise<void> doOneAccept() {
+    // Accept one connection and queue up the next accept on our TaskSet
+
+    return inner->acceptAuthenticated().then(
+        [this](AuthenticatedStream&& stream) {
+          auto promise = kj::evalNow([&] {
+            return tls.wrapServer(kj::fwd<kj::AuthenticatedStream>(stream));
```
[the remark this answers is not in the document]
bcaimano: Hmmmmm, I think if people are okay with mv I'd mildly prefer it. fwd has some advantages in code portability but actively decreases accessibility for junior devs. I'll change my usages.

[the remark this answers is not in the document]
bcaimano: If it took a moment, it's worth a comment :).

c++/src/kj/compat/tls.c++
```diff
@@ -405,31 +416,108 @@ private:
...
+  Promise<T> pop() {
+    // Eventually pop a value from the queue.
+    // Note that if your sink lags your sources, the promise will always be ready.
+
+    KJ_REQUIRE(!fulfiller, "InputQueue can only have one waiter");
+
+    if (head) {
+      // Fast path: There are already values in queue for us to take.
+      auto value = kj::mv(head->value);
+      ;
```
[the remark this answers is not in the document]
bcaimano: Yep, thanks for finding it! I've been testing out limited clang-format application and it has its ups and downs.

c++/src/kj/compat/tls-test.c++
bcaimano: Fair enough. I think I've convinced clang-format to play ball by penalizing whitespace more than assignment breaks. It adds an extra line...which probably couldn't hurt for comprehension anyway.

c++/src/kj/compat/tls-test.c++
bcaimano: It's mostly that testConnection takes a StreamPtr and not a Promise<StreamPtr>. I was hoping that distinguishing between async and sync signatures would make them easier to follow. That said, I don't want to pretend I spent that long thinking about it :).

c++/src/kj/compat/tls-test.c++
bcaimano: That sounds like an improvement to me!

c++/src/kj/compat/tls-test.c++
bcaimano: That's a good point. I might also be able to test our timeout mechanism also here.

bcaimano: Dang, the AsyncIoProvider is baked in a bit heavily into TlsTest. I could refactor it to be external to the TlsTest to get what I need to make a timer, but it seems a bit heavy for this PR.

c++/src/kj/compat/tls-test.c++
bcaimano: Well, no. It's actually a way to kludge clang-format into doing a nice function chaining. That said, this is an appropriate spot for a small comment explaining what we're doing.

c++/src/kj/compat/tls.c++
bcaimano: Oh man, that has a really negative on my ability to parse the async flow. I'll be doing a lot of trivial movement of lvalue lambdas.

c++/src/kj/compat/tls-test.c++
kentonv: Unfortunately clang-format doesn't really support our style (yet -- Vitali has been submitting some changes upstream). I don't recommend relying on it for now.

E.g. the `.then()` below here indents the lambda body too much. We indent lambda bodies one stop more than the surrounding function.

bcaimano:
Compared to what I'm used to, your nits are very kind! It should be ready for another look.

c++/src/kj/compat/tls.c++
```diff
@@ -405,31 +416,109 @@ private:
 // =======================================================================================
 // Implementations of ConnectionReceiver, NetworkAddress, and Network as wrappers adding TLS.
 
-class TlsConnectionReceiver final: public kj::ConnectionReceiver {
+template <typename T>
+class InputQueue {
+  // InputQueue is an async FIFO queue that has many sources but only one sink.
+  // TODO(now) Should I find a header home for this somewhere?
```
bretthoerner, line 422: @kentonv ?

c++/src/kj/compat/tls.c++
```diff
@@ -443,8 +532,39 @@ public:
...
+
+    auto loopAgainOnFailure = [this](Exception&& e) {
+      KJ_LOG(ERROR, "TlsConnectionReceiver could not accept()", kj::mv(e));
+
+      // Even though we had a bad accept(), we should still loop again.
+      tasks.add(doOneAccept());
+    };
+
+    return inner->acceptAuthenticated().then(
+        kj::mv(loopAgainOnSuccess), kj::mv(loopAgainOnFailure));
```
bretthoerner, line 560: Oh man, you're going to have a fun time reading the codebase. :)

c++/src/kj/compat/tls.c++
```diff
@@ -405,31 +416,108 @@ private:
...
+      localFulfiller->fulfill(kj::fwd<T>(v));
+      return;
+    }
+
+    // Slow path: We don't have a consumer, add it to the queue.
+    auto node = kj::heap<Node>(kj::fwd<T>(v));
+    if (head) {
+      Node* currentNode = head.get();
+      while (currentNode->next) {
+        // Run through the nodes until we don't have a next.
```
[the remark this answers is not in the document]
kentonv: FWIW it would be OK to use `std::deque` for this. Generally I think linked lists work better if the element type can have the `next` pointer embedded directly, avoiding the need to wrap it in a `Node` type that implies an extra allocation.

My one reservation with using `std::deque` is if we put this in a public header then that header has to include the std header, and we generally have a rule against including std headers in KJ headers for fear of people relying on the transitive includes, but I guess that's a pretty weak argument. (Also, using std container types always seems to lead to lots of frustration as they tend to get horribly confused by element types that are move-only or enforce transitive constness as KJ does. It's possible to work around, but annoying.)

Anyway up to you if you want to go that route or do the tail pointer thing.

c++/src/kj/compat/tls.c++
```diff
@@ -405,31 +416,108 @@ private:
...
+
+      return kj::mv(paf.promise);
+    }
+  }
+
+private:
+  struct Node {
+    // A simple intrusive linked list node which extends the lifetime of the next node.
+
+    Node(Promise<T>&& value): next{}, value{ kj::fwd<Promise<T>>(value) } {
```
kentonv, line 483: It looks like when you actually construct `Node`s it's always with a value of type `T`, not a `Promise`, so wrapping in a `Promise` is redundant and adds overhead.

c++/src/kj/compat/tls.c++
```diff
@@ -160,11 +161,21 @@ public:
...
+    return sslCall([this]() {
+             return SSL_accept(ssl);
+           })
+        .then([](size_t ret) {
+          if (ret == 0) {
+            kj::throwRecoverableException(
+                KJ_EXCEPTION(DISCONNECTED, "Client disconnected during SSL_accept()"));
+          }
+          KJ_ASSERT(ret > 0, "sslCall must return a non-negative value");
+        });
```
bretthoerner, line 178: This one is still out of style, maybe:
```c++
return sslCall([this]() {
  return SSL_accept(ssl);
}).then([](size_t ret) {
  if (ret == 0) {
    kj::throwRecoverableException(
        KJ_EXCEPTION(DISCONNECTED, "Client disconnected during SSL_accept()"));
  }
  KJ_ASSERT(ret > 0, "sslCall must return a non-negative value");
});
```

or

```c++
return sslCall([this]() { return SSL_accept(ssl); })
    .then([](size_t ret) {
  if (ret == 0) {
    kj::throwRecoverableException(
        KJ_EXCEPTION(DISCONNECTED, "Client disconnected during SSL_accept()"));
  }
  KJ_ASSERT(ret > 0, "sslCall must return a non-negative value");
});
```
You could even pull the first `.then` up, AFAIK.

c++/src/kj/compat/tls.c++
```diff
@@ -405,31 +416,108 @@ private:
 // =======================================================================================
 // Implementations of ConnectionReceiver, NetworkAddress, and Network as wrappers adding TLS.
 
-class TlsConnectionReceiver final: public kj::ConnectionReceiver {
+template <typename T>
+class InputQueue {
+  // InputQueue is an async FIFO queue that has many sources but only one sink.
+  // TODO(pr) Should I find a header home for this somewhere?
```
kentonv: One nice thing about putting it in a header would be that we can unit test it directly.

But the implementation does appear to be coupled to the specific use case in a few ways. If we put it in a header we might want to generalize to support multiple waiters, and maybe remove the `push(promise)` method (make `pushRaw()` be the normal `push()`). I dunno.

c++/src/kj/compat/tls.c++
```diff
@@ -405,31 +416,108 @@ private:
...
+      currentNode->next = kj::mv(node);
+    } else {
+      head = kj::mv(node);
+    }
+  }
+
+  Promise<void> push(Promise<T>&& p) {
+    // Eventually push a value onto the queue.
+
+    return kj::fwd<Promise<T>>(p).then([this](T&& v) {
```
kentonv, line 454: The `kj::fwd<Promise<T>>(p)` doesn't seem to be accomplishing anything here, vs. just `p`.

c++/src/kj/compat/tls.c++
```diff
@@ -443,8 +531,31 @@ public:
...
+            return tls.wrapServer(kj::fwd<kj::AuthenticatedStream>(stream));
+          });
+
+          tasks.add(queue.push(kj::mv(promise)));
+          tasks.add(doOneAccept());
+        },
+        [this](Exception&& e) {
+          KJ_LOG(ERROR, "TlsConnectionReceiver could not accept()", kj::fwd<Exception>(e));
+
+          // Even though we had a bad accept(), we should still loop again.
```
kentonv: Many errors thrown by `accept()` are permanent, meaning that if you call it again, it's just going to throw again, and you'll have a hot loop, which is pretty bad. Think of ENOTSOCK, for example.

Generally we have therefore taken the convention that an exception thrown by `accept()` must be presumed to be permanent. And as a result of that, `accept()` implementations need to be careful not to throw errors that they know to be one-offs. The latter rule is tricky and we've definitely screwed it up a few times, but it doesn't seem like there's any way around it.

c++/src/kj/compat/tls-test.c++
```diff
@@ -26,13 +26,16 @@
 #endif
 
 #include "tls.h"
-#include <kj/test.h>
```
kentonv, line 29: Seems like includes were reordered up here for no particular reason. Did your editor do that automatically? If so I suggest turning that feature off as it'll probably cause lots of noise diffs.

c++/src/kj/compat/tls-test.c++
a-robinson: Vitali did some work not too long ago on getting clang-format to work for our indentation preferences. I don't think it ever got finished, though, it looks like the internal PR for it is still open. Historically the extreme indentation of nested continuations is the main reason we haven't been able to use clang-format.

c++/src/kj/compat/tls.c++
bcaimano: Thanks for pointing this out. Sadly, I can't because of how sslCall works.

c++/src/kj/compat/tls.c++
bcaimano: Yeah, probably! I'll move towards improving readability where I can and put up with it where I can't.

c++/src/kj/compat/tls-test.c++
a-robinson: What do we need a timer for? I was just thinking something like:

```
  ...

  KJ_LOG(INFO, "Requesting and resolving a client connect that never does ssl connect");
  auto hungStream = test.baseReceiver->connect().wait(test.io.waitScope);

  ...
```

And you just leave `hungStream` around until the end of the test, verifying that its presence doesn't get in the way of any other connections getting accepted. The idea is that this would have failed previously (or at least slowed down the test until we gave up on waiting) because we wouldn't have accepted any other connections while waiting on `hungStream`'s handshake.

c++/src/kj/compat/tls.c++
kentonv: > I can't because of how sslCall works.

Can't what?

c++/src/kj/compat/tls-test.c++
```diff
@@ -424,6 +427,27 @@ struct TlsTest {
...
+  Promise<void> writeToServer(Own<AsyncIoStream> client) {
+    return client->write("foo\0", 4).attach(kj::mv(client));
+  }
+
+  Promise<void> readFromClient(Own<AsyncIoStream> server) {
+    auto buf = heapArray<char>(4);
+    return server->read(buf.begin(), buf.size())
+        .attach(kj::mv(server))  //
+        .then([buf = kj::mv(buf)]() {
+          KJ_ASSERT(kj::StringPtr(buf.begin()) == "foo");
```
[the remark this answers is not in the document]
bcaimano: Decent point on the potential for the corrupted buffer, I took it for granted that we wouldn't write more than the original size. I'll switch the way I construct the StringPtr.

c++/src/kj/compat/tls-test.c++
```diff
@@ -1055,6 +1054,60 @@ KJ_TEST("TLS receiver does not stall on hung client") {
   auto extraAcceptPromise = test.receiver->accept().then([&](auto server) {
```
a-robinson, line 1054: Looks like this could now use `.then(readFromClient)`

c++/src/kj/compat/tls.c++
```diff
@@ -160,11 +161,21 @@ public:
...
-    return sslCall([this]() { return SSL_accept(ssl); }).ignoreResult();
+    return sslCall([this]() {
+             return SSL_accept(ssl);
+           })
+        .then([](size_t ret) {
+          if (ret == 0) {
+            kj::throwRecoverableException(
+                KJ_EXCEPTION(DISCONNECTED, "Client disconnected during SSL_accept()"));
+          }
+          KJ_REQUIRE(ret > 0, "sslCall must return a non-negative value");
```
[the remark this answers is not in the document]
bcaimano: Yeah, I agree that it can't fail as the code is written now. I don't like an unhandled range though. I think this might make both of us feel better if I change the return type for sslCall.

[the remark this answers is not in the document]
bcaimano: Waiiiiit, it's size_t anyway. This isn't just redundant, it's impossible.

c++/src/kj/compat/tls-test.c++
a-robinson: Ah sorry, I hadn't realized you had already done this in your fixup and were just commenting about testing the timeout mechanism being too hard. This looks good 👍

c++/src/kj/compat/tls.c++
bcaimano: Yeah, you're right. Thanks for the catch.

c++/src/kj/compat/tls.c++
```diff
@@ -160,11 +161,21 @@ public:
...
     // We are the server. Set SSL options to prefer server's cipher choice.
     SSL_set_options(ssl, SSL_OP_CIPHER_SERVER_PREFERENCE);
 
-    return sslCall([this]() { return SSL_accept(ssl); }).ignoreResult();
+    return sslCall([this]() {
+             return SSL_accept(ssl);
+           })
+        .then([](size_t ret) {
+          if (ret == 0) {
+            kj::throwRecoverableException(
```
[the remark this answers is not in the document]
bcaimano: Interesting, thanks for the context.

c++/src/kj/compat/tls.c++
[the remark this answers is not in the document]
bcaimano: I was wondering if doing tail linking would be worthwhile. Pretty easy to implement that. 

I wouldn't mind writing OwnList if you think it's valuable, I'd probably TODO it here. (My trivial read is that I can generalize List to do both Own and pointer impls.)

c++/src/kj/compat/tls.c++
```diff
@@ -405,31 +416,108 @@ private:
...
-      });
+    if (auto localFulfiller = kj::mv(fulfiller)) {
+      // Fast path: We have a consumer waiting.
+      localFulfiller->fulfill(kj::fwd<T>(v));
+      return;
+    }
+
+    // Slow path: We don't have a consumer, add it to the queue.
+    auto node = kj::heap<Node>(kj::fwd<T>(v));
+    if (head) {
```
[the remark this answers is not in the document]
bcaimano: Hmmmmm, perhaps we should prevent boolean casts on Own then? Maybe<Own<T>> will serve my purposes but it's definitely not as easy to use as Own<T>. I'm doing a fair bit of `(*myExtractedMaybeValue)->`.

c++/src/kj/compat/tls.c++
bcaimano: I think the shift to doubly-linked list is enough of an increase in complexity that I want to write separate tests. Header it is!

c++/src/kj/compat/tls.c++
bcaimano: Mind if I leave it as `kj::mv`? I can see a world where we can optimize promises based on cv qualifiers.

c++/src/kj/compat/tls.c++
bcaimano: Hmmmm, I think my calculous on the hot loop looks like this: If we have to manually restart the process, then the hot loop with errors is actually better because we get _something_ done in that time frame. If we automatically restart the process, then the hot loop is obviously worse. I don't know what our situation is there.

bcaimano: On the tricky throwing conventions...I think we can probably make that less painful if we separate the fatal error path into a separate fulfiller. Errors that flow through the loop would be retriable. Errors that flow through the termination process are most definitely not.

c++/src/kj/compat/tls-test.c++
bcaimano: No, these were purposefully reordered. Alphabetized and categorized headers make it easier to track problems with non-idempotent headers. That said, I would like to ask what the ideal ordering of header categories is.

c++/src/kj/compat/tls-test.c++
bcaimano: Yep, thanks!

c++/src/kj/compat/tls.c++
bcaimano: The consolidation of the two `then()` lambdas can't happen because sslCall returns a promise, not a size_t.

c++/src/kj/compat/tls.c++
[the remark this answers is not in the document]
bcaimano: Hmmmmm, so I can see two potentially nice options then:
- We can make accept()/acceptAuthenticated() turn do one accept loop if we haven't invoked run().
- We can provide a different method instead of run(). `Promise<void> onFatalError()` or something less scary sounding.

c++/src/kj/compat/tls.c++
bretthoerner: Hmm... my snippets are only style/indentation changes from the code I see above in the diff view.

c++/src/kj/compat/tls.c++
bcaimano: I've worked in a codebase with lots of nested async lambdas before. The worst we ever had to do was separate our lambdas into semantic functions if we had too little horizontal space to work with.

For your given example, I'd write this:
```
if (
  really.long.expression.that(
    requires.wrapping.because.its.so.long()
  )
) {
  blockBody();
}
```

I think it boils down to preferring more clearly delineated blocks for me. The sudden left shift indicates a less nested block, which it is not.

c++/src/kj/compat/tls.c++
kentonv: > loop with errors is actually better because we get something done in that time frame.

Unfortunately, there a bunch of reasons that doesn't really work out:

* The KJ event loop likes to do depth-first scheduling because it's more efficient (better cache efficacy, etc.) and more predictable (fewer opportunities for a concurrent task to jump in and change things out from under you), but this means during a hot loop, no other events in the queue will make progress.
* For similar reasons, the KJ event loop also will not check for events from the OS until the in-memory queue is empty. So a hot loop will cause the thread to stop making any progress on I/O.
* Even if the KJ event loop tried to be a fair scheduler, a hot loop would tend to win the competition for CPU time over better-behaved tasks, leading better-behaved tasks to perform very badly. This is true both within the loop's thread itself, and across the system -- Linux is pretty bad at scheduling interactive tasks in the presence of CPU-saturating tasks.

So what would actually happen in this case is the process would saturate the CPU for 30-60 seconds, not make any progress on any I/O, and eventually get killed by the watchdog.

c++/src/kj/compat/tls.c++
kentonv: `.then()` (and most other methods of `Promise`) already moves the promise away implicitly. That's not something we could change now without breaking the world. So using `kj::mv` is redundant.

c++/src/kj/compat/tls.c++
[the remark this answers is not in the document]
kentonv: > Hmmmmm, perhaps we should prevent boolean casts on Own then?

Yeah, we probably should.

> Maybe<Own> will serve my purposes but it's definitely not as easy to use as Own. I'm doing a fair bit of `(*myExtractedMaybeValue)->`.

FWIW I usually do `myExtractedMaybeValue->get()->`, which is a lot easier to type.

I wish I could find a way to formulate the `KJ_IF_MAYBE(foo, bar)` macro such that `foo` ended up as a reference rather than a pointer.

c++/src/kj/compat/tls-test.c++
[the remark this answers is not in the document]
kentonv: It's not that we'd write more than the original size, but rather that when you construct a `StringPtr` from a `const char*` as you do here, it's going to scan memory until it finds a NUL byte.

c++/src/kj/compat/tls.c++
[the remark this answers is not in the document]
kentonv: BTW this kind of stuff is covered in the tour, if you haven't seen it already: https://github.com/capnproto/capnproto/blob/master/kjdoc/tour.md

c++/src/kj/compat/tls.c++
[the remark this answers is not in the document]
kentonv: > My trivial read is that I can generalize List to do both Own and pointer impls.

Eh I doubt that, it's pretty fundamentally different. With non-owned elements, the expectation is that the elements will remove themselves from the list when destroyed. It's commonly used to represent a list of waiters waiting for some event, where each waiter can potentially be canceled. (If you generalize this queue to allow multiple waiters, you should probably use a List<T> there.)

c++/src/kj/compat/tls-test.c++
kentonv: KJ style says that the first include should always be the corresponding header for the particular source file, which then proves that the header declares all its dependencies.

I'm not sure I understand how sorting the rest of the headers helps catch any problems... can you explain?

Regardless of whether we decide to institute header-sorting rules, we prefer to keep unrelated cleanups in their own separate commits from substantial code changes, to ease review. It's fine to include drive-by cleanup commits in a PR but they should be separate commits.

c++/src/kj/compat/tls.c++
```diff
@@ -405,31 +416,109 @@ private:
...
+    // Push an existing value onto the queue.
 
-  Promise<Own<AsyncIoStream>> accept() override {
-    return inner->accept().then([this](kj::Own<AsyncIoStream> stream) {
-      return kj::evalNow([&] { return tls.wrapServer(kj::mv(stream)); })
-          .catch_([this](kj::Exception&& e) {
-        KJ_LOG(ERROR, "error accepting tls connection", e);
-        return accept();
-      });
+    if (auto localFulfiller = kj::mv(fulfiller)) {
```
kentonv, line 428: `fulfiller` should be a `kj::Maybe`, rather than rely on nullable `Own`.

On another note, you may want to check `fulfiller->isWaiting()` to see whether the acceptor canceled the wait, in which case you should discard the fulfiller and proceed as if it wasn't there. It's not super-likely to be important in practice, but it's good to form a habit of writing cancellation-safe code. See also: https://github.com/capnproto/capnproto/blob/master/kjdoc/tour.md#cancellation

c++/src/kj/compat/tls.c++
kentonv: Yeah I only see one `.then()` here... are we looking at the same code?

c++/src/kj/compat/tls.c++
kentonv: Of course, people can reasonably disagree on what style is most readable. Consistency is important since people will always have an easier time reading patterns they are used to. It's definitely frustrating getting into a new codebase that uses different style, because everything seems less readable at first, since it's not what you're used to... but that doesn't necessarily mean the style is objectively less readable.

Personally I find it more difficult to follow code with the lambdas broken out like you have here, because it means the control flow bounces around more, and I have to cross-reference the names with the definitions. I'm used to being able to just read top-to-bottom and see everything in roughly the order it happens. Not saying that what I'm used to is better or worse than what you prefer, but it is what we do throughout the KJ and Workers codebases.

c++/src/kj/compat/tls.c++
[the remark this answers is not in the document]
bcaimano: Hmmmm, I probably would have needed your explanation here to understand https://github.com/capnproto/capnproto/blob/master/kjdoc/tour.md#throwing-and-catching-exceptions fully. I've looked through the tour but I doubt I have testable knowledge of its contents. :/

c++/src/kj/compat/tls.c++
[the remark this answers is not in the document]
bcaimano: If I tried to use a List for the waiters, I'd need to have each waiter come in with a reference to its ListLink, right? This seems like it would be a more classical condVar+storage style of queue.

c++/src/kj/compat/tls.c++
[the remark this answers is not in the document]
bcaimano: That is a lot easier to type/parse. Thanks!

I'd be interested in discussing different ways of phrasing KJ_IF_MAYBE equivalents. I doubt I've thought of every motivation you have for it, but I suspect we can satisfy at least some of them with an unnameable type that does ->/* forwarding.

c++/src/kj/compat/tls.c++
bcaimano: Hmmmm, that's good context to have. Out of curiosity, is there a maximum "depth" of the scheduling? I'd imagine at some point we'd be able to deduce that we're starving out other async sources.

c++/src/kj/compat/tls-test.c++
bcaimano: I think the canonical justification was from a textbook by John Lakos (https://stackoverflow.com/a/14243688). I'm sure you're familiar with the argument, no idea if it speaks to you. I think in practice, order within a category is what sometimes gives benefit. Imagine your stereotypical diamond problem with your includes:
`a.h` -> `b.h`
`a.h` -> `c.h`
`b.h` -> `d.h`
`c.h` -> `d.h`

If you alphabetize your headers, you have a guaranteed parse order [`d.h`, `b.h`, `c.h`, `a.h`]. Because of this parse order, you know that `b.h` cannot be subtly dependent on `c.h`. If you have an issue, it has to be the other way around. If you attempt to introduce a badly phrased dependency between those two headers, either you obey the existing ordering and it works or you violate it and it doesn't. Even in the case where someone has done something bad, it doesn't break the world.

This isn't an earth shattering benefit and it requires all include orders to be sorted. That said, I believe it's something worth working towards.

bcaimano: On the topic of separate commits, are they expected to go into master as separate commits? I'd guess yes but it does feel a bit odd if it's something like applying formatting before the actual work commit.

c++/src/kj/compat/tls.c++
bcaimano: That makes sense to me!

c++/src/kj/compat/tls-test.c++
bretthoerner: Separate commits yeah, but same PR is fine.

bcaimano:
Doing a rebase push fyi.

c++/src/kj/compat/tls.c++
```diff
@@ -35,6 +35,7 @@
 #include <kj/debug.h>
 #include <kj/vector.h>
 
+
```
bcaimano: Man, extra newlines in random patches haunt me.

c++/src/kj/compat/tls-test.c++
```diff
@@ -28,6 +28,7 @@
 #include "tls.h"
 
 #include <openssl/opensslv.h>
+#include <stdlib.h>
```
bcaimano, line 31: This is noise, I'll take it out.

c++/src/kj/compat/tls.c++
bcaimano: I think we are and I'm bad at explaining myself. We have two lambdas. I'd interpreted @bretthoerner's comment as saying that I could use the output of SSL_accept() in the same lambda to check for disconnect. I don't think we can do that because sslCall can be async. I might have misread the entire convo of course.

bretthoerner: Oh, no, my whole comment was style related. I meant "pull it up" as in remove a newline. But it looks like this code has changed a bunch since then anyway so this is moot.

c++/src/kj/compat/tls.c++
bcaimano: I understand that a linear vertical flow is helpful for understanding as well! Keeping in mind that this is async code, perhaps would be most readible to both of us as something like this:
```
auto promise = ...
auto onSuccess = ...
auto onFailure = ...
return promise.then(onSuccess, onFailure);
```

I rely a lot on horizontal indent to convey scoping. When the indentation shifts left for a block, that tells me that I've left a scope, which isn't true here. There's a lot in the codebase that's different than what I'm used to, but this is the one thing that feels like it decreases my ability to understand the code quickly even when I'm familiar with the style.

bcaimano: Thinking about this a bit more, I'd like to make it clear that I'm not trying to be trouble here. I have a vested interest in finding a style of coding that fits within our rules and is readable to us both. I'm curious if it's worth adding a line to the guide on preferring inline lambdas if you feel out of line actively thwarts comprehension.

bcaimano:
Btw, this should be good for another look!

c++/src/kj/compat/tls.c++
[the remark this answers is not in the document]
kentonv: Each call to `wait()` would use `newAdaptedPromise()` with an adapter defined privately in your queue class. The adapter would contain a ListLink.

c++/src/kj/compat/tls-test.c++
kentonv: Hah, that Lakos book is one of about three C++ books I've read, way back in the 90's. I don't think I've ever seen anyone reference it before TBH, but it definitely had a big influence on my style. That said, after following it religiously for a few years I eventually realized some of the advice was a bit overzealous...

In your link, the block quote seems to describe exactly the rule KJ follows (a source file should include its corresponding header first)... but then there is a bunch of text after that, beginning with "That is to say,", which seems to introduce a bunch of rules about the ordering of headers other than the first one. I don't see how those rules follow from the block quote... it seems to me that listing the source file's corresponding header first is sufficient to detect these problems on its own, and the ordering of the rest of the headers doesn't matter.

c++/src/kj/compat/tls.c++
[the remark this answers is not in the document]
bcaimano: I think I understand the shape of what you're suggesting now. The primary benefit I'm seeing is that we'd combine the allocation of the PromiseFulfiller with the Node. Is that accurate?

c++/src/kj/compat/tls-test.c++
bcaimano: I can't personally argue the linked comment from stack overflow, just seems like interesting background. I think I have laid out a reasonable justification for why the ordering of headers _does_ matter though in my own comments. If you want it struck from this PR, I'll accept that.

c++/src/kj/compat/tls.c++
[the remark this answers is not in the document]
kentonv: Right -- as opposed to using `OwnList<Own<PromiseFulfiller>>` or `std::list<Own<PromiseFulfiller>>`, that is.

In fact, I think `kj::newPromiseAndFulfiller` in itself does two or three allocations, whereas `newAdaptedPromise()` always does only one. So, even more savings.

And adapted promises can proactively remove themselves from the list on cancellation, rather than leaving behind orphaned fulfillers that have to be cleaned up later.

c++/src/kj/compat/tls.c++
```diff
@@ -162,11 +163,23 @@ public:
...
+    };
+
+    auto checkSslResult = [](size_t ret) {
+      if (ret == 0) {
+        kj::throwRecoverableException(
+            KJ_EXCEPTION(DISCONNECTED, "Client disconnected during SSL_accept()"));
+      }
+    };
+
+    return sslCall(kj::mv(sslAccept)).then(checkSslResult);
```
kentonv, line 182: Since you don't have `kj::mv()` on `checkSslResult`, I think this is capturing an lvalue reference, which is left dangling when `accept()` returns. KJ has a general convention that when a callback is expected, if you specify an lvalue, it is assumed you wanted to call back to that specific value by reference. I'll admit this convention is not my finest work -- in practice, almost no code ever actually intends to pass a callback by lvalue reference, and if it did, it would be clearer to wrap it in a lambda that explicitly captures the reference as a reference. Meanwhile, cases like this come up from time to time where someone accidentally passes a reference when they didn't intend to.

In this case, it doesn't result in any problem because the lambda doesn't actually capture anything, but if someone went and added a capture later, it would be corrupted.

I fear this problem will come up a lot more often if we're declaring callbacks on a separate line and then passing them into `.then()` as you've done here. Can we stick with the inline lambda style used throughout our code today?

c++/src/kj/compat/tls.c++
```diff
@@ -445,8 +459,58 @@ public:
...
+  }
+
   TlsContext& tls;
-  kj::Own<kj::ConnectionReceiver> inner;
+  Own<ConnectionReceiver> inner;
+
+  ProducerConsumerQueue<AuthenticatedStream> queue;
+  TaskSet tasks;
+
+  template<typename T>
```
kentonv, line 501: This template parameter doesn't seem to be used.

c++/src/kj/compat/tls.c++
```diff
@@ -445,8 +459,58 @@ public:
...
+      auto pushStream = [this](AuthenticatedStream&& stream) -> Promise<void> {
+        // This is only attached to the success path, thus the error handler will catch if our
+        // promise fails.
+        queue.push(kj::mv(stream));
+        return kj::READY_NOW;
+      };
+      tasks.add(promise.then(kj::mv(pushStream)));
+
+      // Queue up the next accept loop immediately without waiting for SSL_accept()/wrapServer().
+      tasks.add(doOneAccept());
```
kentonv, line 480: Why not `return doOneAccept();`? That's how we usually do loops like this.

(Probably `doOneAccept()` would be renamed to `acceptLoop()` in this case.)

c++/src/kj/compat/tls.c++
```diff
@@ -407,29 +420,30 @@ private:
...
-  Promise<kj::AuthenticatedStream> acceptAuthenticated() override {
-    return inner->acceptAuthenticated().then([this](kj::AuthenticatedStream stream) {
-      return kj::evalNow([&] { return tls.wrapServer(kj::mv(stream)); })
-          .catch_([this](kj::Exception&& e) {
-        KJ_LOG(ERROR, "error accepting tls connection", e);
-        return acceptAuthenticated();
-      });
-    });
+  Promise<AuthenticatedStream> acceptAuthenticated() override {
+    return queue.pop();
```
kentonv, line 442: If a fatal error occurred, then `accept()` should throw it, since that's what all existing uses of this class expect. You could do:

```
return queue.pop().exclusiveJoin(onFatalError());
```

c++/src/kj/compat/tls.c++
```diff
@@ -445,8 +459,58 @@ public:
...
+        return kj::READY_NOW;
+      };
+      tasks.add(promise.then(kj::mv(pushStream)));
+
+      // Queue up the next accept loop immediately without waiting for SSL_accept()/wrapServer().
+      tasks.add(doOneAccept());
+    };
+
+    auto loopAgainOnFailure = [this](Exception&& e) {
+      KJ_LOG(ERROR, "TlsConnectionReceiver could not accept()", e);
```
kentonv, line 484: I'd argue that if someone is actually listening using `onFatalError()`, we shouldn't log, since it should be up to the caller to decide if and how they want to log the error. But see my comments on whether `onFatalError()` is the right interface in the first place...

c++/src/kj/list.h
```diff
@@ -202,6 +203,231 @@ class ListIterator {
   friend class List<T, link>;
 };
 
-} // namespace kj
+template <typename T>
+class OwnList {
+  // A linked list that does its own memory allocation.
+  //
+  // Keep in mind that each new element requires the allocation of two pointers and a T. This is
+  // very little if T is actually an Own<U> but can be quite expensive if T is a large type.
```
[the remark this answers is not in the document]
bcaimano: Well, I don't think an obvious comment is a bad thing personally. Good point about misstating the layout of Own. I'll explicitly call out its use.

c++/src/kj/list.h
```diff
@@ -202,6 +203,231 @@ class ListIterator {
...
+
+  struct Node {
+    // A simple linked list node which extends the lifetime of the next node.
+
+    Node(T&& value): value{ kj::mv(value) } {}
+
+    Maybe<Own<Node>> maybeNext;
+    Node* prev = nullptr;
```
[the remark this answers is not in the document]
bcaimano: I'm familiar with this variant of linked lists. I appreciate that it decreases the need for branching logic. In some situations, it may lead what was previously a single deref into a double deref, which is less of an issue since we use tcmalloc. I do worry that it makes code more difficult to follow for a casual reader. I've rewritten to a variant of this with a paranoid amount of comments.

c++/src/kj/list.h
```diff
@@ -202,6 +203,231 @@ class ListIterator {
...
+
+  using MutableIterator = Iterator<T>;
+  using ConstIterator = Iterator<const T>;
+
+  OwnList() = default;
+  KJ_DISALLOW_COPY(OwnList);
+
+  void pushBack(T v) {
+    // Push a value onto the end of the list.
+    // TODO(now) Should this be add[Back]() instead?
```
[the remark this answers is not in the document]
bcaimano: Hmmmm, I think `add()` + `add(Iterator)` can possible be more within the culture then. `pushFront()` is just `add(begin())`. I think I lean in favor of returning the next iterator from whatever does the erasing. I  think this can be equivalent to `popBack()` and `popFront()` if I implement `remove(Iterator)` and `Iterator::operator--()`. (No idea how strongly you feel about it being distinct from `List::remove()`.)

c++/src/kj/common.h
```diff
@@ -1255,6 +1255,8 @@ inline T* readMaybe(T* ptr) { return ptr; }
 }  // namespace _ (private)
 
 #define KJ_IF_MAYBE(name, exp) if (auto name = ::kj::_::readMaybe(exp))
+#define KJ_WHILE_MAYBE(name, exp) while (auto name = ::kj::_::readMaybe(exp))
+//TODO(now) Do I need to write tests for KJ_WHILE_MAYBE?
```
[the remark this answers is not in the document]
bcaimano: Ugh, I keep running into situations where it's useful and then drifting away from needing it. I might just toss it at you in a separate PR if it comes up another time. :/

c++/src/kj/queue.h
```diff
@@ -0,0 +1,83 @@
...
+
+    // We don't have any waiters, store the value for the first one.
+    values.pushBack(kj::mv(v));
+  }
+
+  Promise<T> pop() {
+    // Eventually pop a value from the queue.
+    // Note that if your sinks lag your sources, the promise will always be ready.
+
+    KJ_IF_MAYBE (value, values.popFront()) {
```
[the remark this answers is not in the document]
bcaimano: Yeah, you really have my sympathy with this one :(.

c++/src/kj/compat/tls.c++
bcaimano: Hmmmm, yeah, not sure if I had a good motivation here.

c++/src/kj/compat/tls.c++
bcaimano: Hmmmm, I don't think that works because the Promises have separate types. I can think of a few different solutions but most of them aren't so great. :/ Dropping a comment in place of a better idea for now.

c++/src/kj/async-io.h
```diff
@@ -419,6 +419,10 @@ class ConnectionReceiver {
   // For backwards-compatibility, the default implementation of this method calls `accept()` and
   // then adds `UnknownPeerIdentity`.
 
+  virtual Promise<void> onFatalError();
```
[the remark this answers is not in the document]
bcaimano: I like the `listen()` interface. I might suggest `kj::Promise<void> listen(ConnectionReceiver::Handler&)` instead, it gives us slightly more design space because we can expand the interface for `Handler` at will.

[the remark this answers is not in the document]
kentonv: Hmm, I somewhat agree, and note that it would save an allocation as well (kj::Function requires an allocation, but `Handler&` could point at the parent object).

However, the `kj::Function` approach is significantly less verbose for callers, since they don't have to define a class.

`kj::TaskSet`, for example, requires an `ErrorHandler&`, and my experience has been that it's actually pretty painful every time I want to create a `kj::TaskSet` and have to figure out what's going to satisfy the `ErrorHandler&`... in retrospect I wish I made it accept a `kj::Function`.

So I'm a bit torn here but I think I lean more towards the `kj::Function` approach.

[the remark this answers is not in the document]
kentonv: I guess maybe the `kj::Function`-based `listen()` could be provided as a non-virtual method that is implemented in terms of the `Handler&`-based `listen()`...

c++/src/kj/compat/tls.c++
kentonv: Since `onFatalError()` never resolves, only throws, you can do:

```
return queue.pop().exclusiveJoin(
    onFatalError().then([]() -> AuthenticatedStream { KJ_UNREACHABLE; }));
```

c++/src/kj/list.h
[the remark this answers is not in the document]
kentonv: > I think I lean in favor of returning the next iterator from whatever does the erasing.

I definitely understand your preference here -- but I think it's confusing if List and OwnList work differently.

I guess they already work differently due to intrusive vs. non-intrusive.

But... Hmm, maybe we've lost track of the bigger picture here. When is OwnList better than std::deque? It seems like there are two possible advantages:
1. Supporting efficient insertion and deletion mid-list -- but our present use case doesn't need this.
2. Avoiding allocations on insertion and deletion, by relying on an embedded list link (intrusive linked list). But this is not the design being proposed, and our present use case wouldn't benefit from it.
3. Avoiding std in favor of KJ-native data structures. But then, why a new list type, vs. introducing a `kj::Deque`, which would be more efficient?

Maybe we should just use `std::deque` for now and leave it as a TODO to create a `kj::Deque`?

I am worried that if we introduce a non-intrusive `OwnList` now, then it's harder to introduce an intrusive version later.

Sorry to give this feedback so late in the process. :/

c++/src/kj/compat/tls.c++
bcaimano: It is capturing an lvalue reference. I wouldn't say it's unintentional. I did encounter some segfaults from use-after-move in other places and chose to leave it unmoved when I could because it would be an easier skim. I agree about the capture situation. I'm uncomfortable with two points here:
- We're relying on an undocumented convention to prevent programmer error.
- We generally have made stateful functor usage unsafe.

I'll formulate it differently here to avoid that. Going forward, I'll try to see if I can either make then() have stronger guarantees or stand up a clang-tidy with bugprone-use-after-move.

c++/src/kj/async-io.h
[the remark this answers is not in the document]
bcaimano: Unfortunately, when I gave it a shot it really started looking like the proper behavior for mixed accept() and listen() would be painful to guarantee. On the upside, I think I've figured out a way to phrase your concerns on accept() exceptions into something that avoids a separate promise.

c++/src/kj/list.h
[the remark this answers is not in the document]
bcaimano: It is a mite frustrating to hear now but it's a good conversation to be having. I have two points that may or may not be valuable to you:
- There is one more advantage: Traditional linked lists are generally more efficient for small size sequential data storage. This is actually what made me shy away from using a deque here, my use case maintains a list of one in the ideal case.
- If I was writing a `kj::OwnDeque<T>`, I'd write a `kj::OwnList<T>` as well because I would consider it syntactic sugar on top of `kj::OwnList<DequeBlock<T>>`.

c++/src/kj/async-io.h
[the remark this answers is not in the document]
kentonv: FWIW I wouldn't expect to support mixed `accept()` and `listen()`. It's already the case that you can't have multiple overlapping calls to `accept()`, so also saying you can't have overlapping `listen()` seems reasonable.

c++/src/kj/list.h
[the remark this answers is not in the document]
kentonv: I think `kj::Deque<T>` would "own" the elements by default, just like `kj::Vector<T>`, `kj::Array<T>`, etc. So there's no need for `kj::OwnDeque<T>`.

Are deques normally implemented using a list of blocks? I was thinking it could be implemented as something like a cross between a vector an a ring buffer...

> Traditional linked lists are generally more efficient for small size sequential data storage.

Hmm, is that because `std::deque` allocates a large initial buffer? `kj::Vector` starts with a 4-element array FWIW (unless you pass something different to the constructor).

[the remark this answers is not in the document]
bcaimano: > Are deques normally implemented using a list of blocks? I was thinking it could be implemented as something like a cross between a vector an a ring buffer...

I must admit I had to go do a bit of research to confirm my understanding. std::deque is a "vector of vectors", not a "list of vectors" like I remembered. The big advantage to that is that you get bounded cost for insertion and removal in the middle of the queue. This flexibility is considered attractive for a general library type.

Ring buffers are a perfectly valid way to implement the core promises of a deque and people do use those sort of deques. The major pain point is the same as most vectors, you can end up with "stop the world" moments when it needs to grow. In my past life in fintech, we used a lot of fixed size ring buffers for queues. Honestly, they work pretty well as long as you think hard about what happens when you hit the limit. I'd be happy to write a ring buffer type instead if you'd prefer that.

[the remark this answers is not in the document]
bcaimano: > Hmm, is that because std::deque allocates a large initial buffer? kj::Vector starts with a 4-element array FWIW (unless you pass something different to the constructor).

IIRC, both `std::deque` and `std::vector` do no allocations until first use. But yeah, deque has a 512B (1/8th a page) block size. That's a sizable reservation that we'll only use if something is wrong.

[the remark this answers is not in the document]
kentonv: `kj::Vector` also doesn't allocate until first use, but then it allocates an array of length 4.

STL's tendency to disregard small containers in optimization decisions is pretty annoying. (Though I suppose this is an implementation-specific decision. Note that we use libc++, not libstdc++. I wonder if they make the same decisions.)

Is there any way to instruct `std::deque` to reserve a smaller amount of space to start? If so maybe we could just live with that.

IMO vector-of-vectors is overcomplicated. If we really want to implement our own deque, I'd make it a ring buffer on top of a single vector. However, this may still be somewhat complicated to implement as kj::Vector at present is just a thin wrapper around kj::ArrayBuilder which is not designed for the ring buffer scenario. Presumably we want to control constructor and destructor calls so that the unused slots aren't constructed at all, rather than default-initializing them. kj::ArrayBuilder is designed to deal with that. But we probably can't just extend it to support a ring buffer as that'll increase the size and complexity of all ArrayBuilders (and Vectors) to support an obscure use case. So I suppose our Deque type will have to do all its own allocation and constructor management, duplicating a lot of ArrayBuilder functionality.

This would all be useful to have around long-term, but it's quite a yak shave for this ticket. I don't think it's the best thing to spend time on at the moment.

So I think I'd say you should either:
a. Just use whatever STD type makes the most sense.
b. Implement a simple owning linked list, but don't try to make it public for now (so that we can defer the decision on whether it should be intrusive until we have more use cases).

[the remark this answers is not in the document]
bcaimano: Design concerns aside, I think I'll just use a `std::list`. If we see issues, we can switch to `std::deque` or make our own type.

bcaimano:
This should be ready for another look! (With rebases, sadly)

a-robinson:
@kentonv if you're able to prioritize this while catching up on your backlog, we think it'd fix a recurring problem that's kept popping up lately so it'd be nice to get in this week.

kentonv:
@a-robinson sorry, said backlog meant I didn't see your message until now. A chat ping would have reached me faster. Anyway, I'm reviewing now.

c++/src/kj/compat/tls.c++
```diff
@@ -445,8 +457,53 @@ public:
...
+    // Accept one connection and queue up the next accept on our TaskSet.
+
+    return inner->acceptAuthenticated().then(
+        [this](AuthenticatedStream&& stream) {
+      return onAcceptSuccess(kj::mv(stream));
+        },
+        [this](Exception&& e) {
+      onAcceptFailure(kj::mv(e));
+      return kj::READY_NOW;
+    });
```
kentonv, line 497: We would normally format this like:

```
    return inner->acceptAuthenticated().then(
        [this](AuthenticatedStream&& stream) {
      return onAcceptSuccess(kj::mv(stream));
    }, [this](Exception&& e) {
      onAcceptFailure(kj::mv(e));
      return kj::READY_NOW;
    });
```

c++/src/kj/compat/tls.c++
```diff
@@ -445,8 +457,53 @@ public:
...
+    // Store this exception to reject all future calls to accept() and reject any unfulfilled
+    // promises from the queue.
+    maybeInnerException = kj::mv(e);
+    queue.rejectAll(Exception(KJ_REQUIRE_NONNULL(maybeInnerException)));
+  }
+
+  Promise<void> acceptLoop() {
+    // Accept one connection and queue up the next accept on our TaskSet.
+
+    return inner->acceptAuthenticated().then(
```
kentonv, line 490: If `inner->acceptAuthenticated()` throws synchronously, it looks like the exception won't be caught and will bubble up to the TaskSet, causing `taskFailed()` to be invoked, which merely logs the exception. The loop will be broken but the application isn't notified so it'll hang forever.

I think to solve this, the `acceptLoop()` promise should _not_ go in the `TaskSet`, but should instead be stored as a separate member `Promise<void> acceptLoopTask`, and should have an explicit error handler like:

```
acceptLoopTask = acceptLoop().eagerlyEvaluate([this](kj::Exception&& e) {
  onAcceptFailure(kj::mv(e));
});
```

Then you can also remove the async error handler here, since you can rely on the top-level handler to take care of both sync and async exceptions.

c++/src/kj/compat/tls.c++
```diff
@@ -445,8 +457,53 @@ public:
...
+      return kj::READY_NOW;
+    });
+    tasks.add(kj::mv(sslPromise));
+
+    // Queue up the next accept loop immediately without waiting for SSL_accept()/wrapServer().
+    return acceptLoop();
+  }
+
+  void onAcceptFailure(Exception&& e) {
+    KJ_LOG(ERROR, "TlsConnectionReceiver could not accept()", e);
```
kentonv, line 479: Since we're propagating the exception to the app, I would leave it up to the app to decide whether to log.

c++/src/kj/compat/tls.c++
```diff
@@ -445,8 +457,53 @@ public:
...
+    maybeInnerException = kj::mv(e);
+    queue.rejectAll(Exception(KJ_REQUIRE_NONNULL(maybeInnerException)));
+  }
+
+  Promise<void> acceptLoop() {
+    // Accept one connection and queue up the next accept on our TaskSet.
+
+    return inner->acceptAuthenticated().then(
+        [this](AuthenticatedStream&& stream) {
+      return onAcceptSuccess(kj::mv(stream));
```
kentonv, line 492: I find it a little weird that `acceptLoop()` and `onAcceptSuccess()` do mutual recursion here. It might be a little easier to see that this is a loop if `onAcceptSuccess()` returned void and here we did:

```
onAcceptSuccess(kj::mv(stream));  // will queue a handshake task to `tasks`
return acceptLoop();
```

c++/src/kj/compat/tls.c++
```diff
@@ -407,29 +417,31 @@ private:
...
 public:
-  TlsConnectionReceiver(TlsContext& tls, kj::Own<kj::ConnectionReceiver> inner)
-      : tls(tls), inner(kj::mv(inner)) {}
+  TlsConnectionReceiver(TlsContext& tls, Own<ConnectionReceiver> inner)
+      : tls(tls), inner(kj::mv(inner)), tasks(*this) {
+    // Start up the connection loop.
+    tasks.add(acceptLoop());
+  }
+
+  void taskFailed(Exception&& e) override {
```
kentonv, line 428: FWIW I usually inherit privately from `TaskSet::ErrorHandler`, and declare `taskFailed()` as a private method, to avoid making it part of the public interface. But this whole class is private to this translation unit anyway so it doesn't really matter here. (No need to action this comment.)

c++/src/kj/compat/tls.c++
```diff
@@ -162,11 +163,20 @@ public:
...
     SSL_set_options(ssl, SSL_OP_CIPHER_SERVER_PREFERENCE);
 
-    return sslCall([this]() { return SSL_accept(ssl); }).ignoreResult();
+    auto acceptPromise = sslCall([this]() {
+      return SSL_accept(ssl);
+    });
+    return acceptPromise.then([](size_t ret) {
+      if (ret == 0) {
+        kj::throwRecoverableException(
+            KJ_EXCEPTION(DISCONNECTED, "Client disconnected during SSL_accept()"));
```
kentonv, line 177: This will eventually get `KJ_LOG()`ed in `TlsConnectionReceiver::taskFailed()`, but this doesn't seem like an error that we want generating Sentry reports. Maybe `taskFailed()` should avoid logging `DISCONNECTED` exceptions?

c++/src/kj/async-queue.h
```diff
@@ -0,0 +1,156 @@
...
+    node.remove();
+  }
+
+  bool empty() const {
+    return queue.empty();
+  }
+
+private:
+  struct BaseNode {
+    // This is a separate structure because List requires a predefined memory layout but
```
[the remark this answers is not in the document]
bcaimano: It only works after Node defines its memory layout, which means we can't name the specialized List within BaseNode. This is because List requires a pointer to member template parameter. We would be able to avoid this if it took in a stateless functor such that `getLinkFunctor(Item&) -> Link&`.

c++/src/kj/compat/tls.c++
bcaimano: I think I like this way better overall. Thanks!

I should ask in this context: What is the criteria for a Promise to be "waiting"? I could see several different criteria, and I've been nervous that keeping a `Promise` member would not count.

c++/src/kj/compat/tls.c++
bcaimano: I'm concerned that we may lack any other avenue to detect a burst of aborted TLS connection attempts. Would you be okay if I keep as is but drop a TODO(someday) explaining that we may want to do this in the future?

c++/src/kj/async-queue.h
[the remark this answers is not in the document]
kentonv: Oh I see, this is because you have `Queue& queue` as a member of `Node`, and `Queue` is a typedef of `List<BaseNode>`.

But instead of doing that, you could have `WaiterQueue& queue` as the member of `Node`, then access the list as `queue.list`. That's what I usually do.

c++/src/kj/compat/tls.c++
kentonv: A `Promise` that is just sitting as a member will execute, with one caveat: trailing `.then()` continuations that only transform the result (they do not return another Promise) will be executed lazily when the result is actually needed. However, using `.eagerlyEvaluate()` prevents this. So the general rule of thumb is: if you're going to keep a `Promise` around, make sure to call `.eagerlyEvaluate()` on it. This also forces you to provide an exception handler, so that errors aren't silently ignored.

c++/src/kj/compat/tls.c++
kentonv: In general I would say that aborted connection attempts are really up to the client to report -- the server does not know if it's an error or not, so shouldn't report it as one. We could arguably add a metric to count these, so that we could alert on sudden spikes. But reporting a Sentry error is definitely not what we want here. We know for a fact that we have security scanners and such which are making these connections and dropping them today, and that's not a problem, but our goal with Sentry is to avoid reporting anything that isn't actionable (which, obviously, we're quite a ways from today, but that's the goal). So we really can't report these to Sentry.

bcaimano: Hmmmm, in my experience, I have seen logs on both sides prove to be valuable. I feel strongly that we shouldn't ignore these on the server side.

How about I add either a `kj::Maybe<kj::Own<TaskSet::ErrorHandler>>` or a `kj::Maybe<kj::Function<void(Exception&&)>>` to TLS context? Then the user can handle it how they want downstream.

In our case, I'd probably log in a why that does not report to Sentry.

kentonv: I'm OK with adding a new callback to handle connections that drop during the handshake, but can we plan on doing that in a separate PR? For now I think the best default behavior is to log only non-disconnected exceptions.

bcaimano:
Pushing up a rebase on master fyi

bcaimano:
Okay, added in the DISCONNECTED exemption with a todo. I also switched the most likely failure point to use DISCONNECTED instead of FAILED.

c++/src/kj/compat/tls.c++
```diff
@@ -315,7 +327,7 @@ private:
             // According to documentation we shouldn't get here, because our BIO never returns an
             // "error". But in practice we do get here sometimes when the peer disconnects
             // prematurely.
-            KJ_FAIL_ASSERT("TLS protocol error");
+            return KJ_EXCEPTION(DISCONNECTED, "SSL unable to continue I/O");
```
kentonv, line 330: Out of curiosity, did you actually see this error manifest in some testing? Or is this change based on the comment above (which, to be fair, does seem to suggest that DISCONNECTED is correct here).

bcaimano: This one I actually did see in testing when I let my client side disappear---I was confused when I seemed to still need the KJ_EXPECT_LOG despite adding the exemption. IIRC, most disconnections will surface as ECONNRESET/ECONNREFUSED during socket operations. I believe SSL will only politely close the socket with a closure alert if there was a protocol violation.

kentonv:
Bad news... somehow the TLS tests are broken on Windows. That's going to be a pain to debug. :(

```
2021-07-07T23:49:10.6436579Z 2: [ TEST ] compat\tls-test.c++:457: TLS basics
2021-07-07T23:49:11.0192994Z 2: kj\async-io-win32.c++:385: failed: WSARecv(): #64 The specified network name is no longer available.
2021-07-07T23:49:11.0194080Z 2: stack: 64589c 704c90 870e80 668718 668670 5b8f64
2021-07-07T23:49:11.0195532Z 2: [ FAIL ] compat\tls-test.c++:457: TLS basics (375582 μs)
2021-07-07T23:49:11.0196575Z 2: [ TEST ] compat\tls-test.c++:472: TLS peer identity
2021-07-07T23:49:11.0260346Z 2: kj\async-io-win32.c++:385: failed: WSARecv(): #64 The specified network name is no longer available.
2021-07-07T23:49:11.0261293Z 2: stack: 64589c 704c90 870e80 668718 668670 5b8f64
2021-07-07T23:49:11.0261979Z 2: [ FAIL ] compat\tls-test.c++:472: TLS peer identity (6804 μs)
2021-07-07T23:49:11.0263400Z 2: [ TEST ] compat\tls-test.c++:512: TLS multiple messages
2021-07-07T23:49:11.0328371Z 2: [ PASS ] compat\tls-test.c++:512: TLS multiple messages (6827 μs)
2021-07-07T23:49:11.0329242Z 2: [ TEST ] compat\tls-test.c++:552: TLS zero-sized write
2021-07-07T23:49:11.0396684Z 2: [ PASS ] compat\tls-test.c++:552: TLS zero-sized write (6790 μs)
2021-07-07T23:49:11.0397510Z 2: [ TEST ] compat\tls-test.c++:598: TLS full duplex
2021-07-07T23:49:11.2841713Z 2: [ PASS ] compat\tls-test.c++:598: TLS full duplex (244380 μs)
2021-07-07T23:49:11.2842703Z 2: [ TEST ] compat\tls-test.c++:649: TLS SNI
2021-07-07T23:49:11.2924819Z 2: [ PASS ] compat\tls-test.c++:649: TLS SNI (8371 μs)
2021-07-07T23:49:11.2925795Z 2: [ TEST ] compat\tls-test.c++:685: TLS certificate validation
2021-07-07T23:49:11.3239831Z 2: [ PASS ] compat\tls-test.c++:685: TLS certificate validation (31435 μs)
2021-07-07T23:49:11.3241297Z 2: [ TEST ] compat\tls-test.c++:703: TLS client certificate verification
2021-07-07T23:49:11.3567202Z 2: kj\async-io-win32.c++:385: failed: WSARecv(): #64 The specified network name is no longer available.
2021-07-07T23:49:11.3568192Z 2: stack: 64589c 704c90 870e80 668718 668670 5b8f64
2021-07-07T23:49:11.3569356Z 2: [ FAIL ] compat\tls-test.c++:703: TLS client certificate verification (32688 μs)
2021-07-07T23:49:11.3570654Z 2: [ TEST ] compat\tls-test.c++:921: TLS receiver basics
2021-07-07T23:49:11.3636453Z 2: kj\async-io-win32.c++:372: disconnected: WSARecv(): #10053 An established connection was aborted by the software in your host machine.
2021-07-07T23:49:11.3637652Z 2: stack: 704c90 870e80 668718 668670 5b8f64
2021-07-07T23:49:11.3638778Z 2: [ FAIL ] compat\tls-test.c++:921: TLS receiver basics (6957 μs)
2021-07-07T23:49:11.3639601Z 2: [ TEST ] compat\tls-test.c++:943: TLS receiver experiences pre-TLS error
2021-07-07T23:49:11.3640575Z 2: [ PASS ] compat\tls-test.c++:943: TLS receiver experiences pre-TLS error (269 μs)
2021-07-07T23:49:11.3643052Z 2: [ TEST ] compat\tls-test.c++:959: TLS receiver accepts multiple clients
2021-07-07T23:49:11.4907346Z 2: [ PASS ] compat\tls-test.c++:959: TLS receiver accepts multiple clients (126759 μs)
2021-07-07T23:49:11.4908591Z 2: [ TEST ] compat\tls-test.c++:996: TLS receiver does not stall on client that disconnects before ssl handshake
2021-07-07T23:49:11.6201385Z 2: [ PASS ] compat\tls-test.c++:996: TLS receiver does not stall on client that disconnects before ssl handshake (129319 μs)
2021-07-07T23:49:11.6203073Z 2: [ TEST ] compat\tls-test.c++:1042: TLS receiver does not stall on hung client
2021-07-07T23:49:11.7465205Z 2: [ PASS ] compat\tls-test.c++:1042: TLS receiver does not stall on hung client (126339 μs)
```

kentonv:
I think the problems here most likely have to do with the fact that you're destroying the client stream immediately after `write()` completes in `writeToServer()`, before the other end has actually had a chance to read. It seems Windows is reacting to this by reporting an error on the read end, rather than simply reporting EOF. This may or may not be a deeper bug in the Windows async implementation, but I don't think you should try to fix that in your PR.

I think destroying the client so early is probably not what you want to do anyway. I'd adjust the code so that `writeToServer` and `readFromClient` take `AsyncIoStream&` as a parameter, rather than taking an `Own`. Then make sure that the objects remain in-scope until both operations are complete.

bcaimano:
>I think destroying the client so early is probably not what you want to do anyway. I'd adjust the code so that writeToServer and readFromClient take AsyncIoStream& as a parameter, rather than taking an Own. Then make sure that the objects remain in-scope until both operations are complete.

Seems reasonable to me, I'll try that now. Thanks for investigating this for me! As always, I'm happy to do the leg work for a future improvement if desired on Windows.

kentonv:
@bcaimano Well, all I did was look at the CI log, looked at the code, and made my best guess. :)

bcaimano:
Okay, looking slightly deeper, there might be something we want to improve eventually in the Windows implementation. Most of the failing tests used `testConnection()`. The one that didn't was a new test that probably should have done so. There is one fundamental way in which testConnection is different from how most of my new tests work: It waits for `writePromise` and then the `readPromise` instead of using `joinPromises(...)`. This means that when we disconnect the client side before we wait on the read to finish. This implies that the act of reading relies on an open remote end and not just a buffer of available data.

kentonv:
GitHub was having serious problems displaying this thread, so I just went through and marked all the conversations resolved in hopes that helps...

bcaimano:
> GitHub was having serious problems displaying this thread, so I just went through and marked all the conversations resolved in hopes that helps...

We may have talked it to death!

PR #1261, opened 2021-06-07
Allow parsing URLs without contextual validation

bcaimano:
This does the following:
- Separates fragment behavior into a Url::Option member.
- Separates URL parsing from context validation.
- Adds a new Url::Context entry, ARBITRARY, that performs no validation.

### DIFF
--- a/c++/src/kj/compat/url.c++
+++ b/c++/src/kj/compat/url.c++
@@ -45,9 +45,9 @@ const parse::CharGroup_& getEndPathPart(Url::Options::MergeFragment mergeFragmen
   static constexpr auto END_PATH_PART_REQUEST = parse::anyOfChars("/?");
 
   switch (mergeFragment) {
-    case Url::Options::MergeFragment::kAuto:  KJ_UNREACHABLE;
-    case Url::Options::MergeFragment::kNo:    return END_PATH_PART_HREF;
-    case Url::Options::MergeFragment::kYes:   return END_PATH_PART_REQUEST;
+    case Url::Options::MergeFragment::AUTO:  KJ_UNREACHABLE;
+    case Url::Options::MergeFragment::NO:    return END_PATH_PART_HREF;
+    case Url::Options::MergeFragment::YES:   return END_PATH_PART_REQUEST;
   }
 
   KJ_UNREACHABLE;
@@ -58,9 +58,9 @@ const parse::CharGroup_& getEndQueryPart(Url::Options::MergeFragment mergeFragme
   static constexpr auto END_QUERY_PART_REQUEST = parse::anyOfChars("&");
 
   switch (mergeFragment) {
-    case Url::Options::MergeFragment::kAuto:  KJ_UNREACHABLE;
-    case Url::Options::MergeFragment::kNo:    return END_QUERY_PART_HREF;
-    case Url::Options::MergeFragment::kYes:   return END_QUERY_PART_REQUEST;
+    case Url::Options::MergeFragment::AUTO:  KJ_UNREACHABLE;
+    case Url::Options::MergeFragment::NO:    return END_QUERY_PART_HREF;
+    case Url::Options::MergeFragment::YES:   return END_QUERY_PART_REQUEST;
   }
 
   KJ_UNREACHABLE;
@@ -250,7 +250,7 @@ Maybe<Url> Url::tryParseWithoutContext(StringPtr text, Options options) {
   }
 
   if (text.startsWith("#")) {
-    if (options.mergeFragment == Options::MergeFragment::kNo) {
+    if (options.mergeFragment == Options::MergeFragment::NO) {
       result.fragment = percentDecode(text.slice(1), err, options);
       text = {};
     } else {
@@ -271,23 +271,23 @@ Maybe<Url> Url::tryParseWithoutContext(StringPtr text, Options options) {
 Maybe<Url> Url::tryParse(StringPtr text, Context context, Options options) {
   switch (context) {
     case Context::REMOTE_HREF: {
-      if (options.mergeFragment == Options::MergeFragment::kAuto) {
-        options.mergeFragment = Options::MergeFragment::kNo;
+      if (options.mergeFragment == Options::MergeFragment::AUTO) {
+        options.mergeFragment = Options::MergeFragment::NO;
       }
     } break;
     case Context::HTTP_PROXY_REQUEST: {
-      if (options.mergeFragment == Options::MergeFragment::kAuto) {
-        options.mergeFragment = Options::MergeFragment::kYes;
+      if (options.mergeFragment == Options::MergeFragment::AUTO) {
+        options.mergeFragment = Options::MergeFragment::YES;
       }
     } break;
     case Context::HTTP_REQUEST: {
-      if (options.mergeFragment == Options::MergeFragment::kAuto) {
-        options.mergeFragment = Options::MergeFragment::kYes;
+      if (options.mergeFragment == Options::MergeFragment::AUTO) {
+        options.mergeFragment = Options::MergeFragment::YES;
       }
     } break;
     case Context::ARBITRARY: {
-      if (options.mergeFragment == Options::MergeFragment::kAuto) {
-        options.mergeFragment = Options::MergeFragment::kNo;
+      if (options.mergeFragment == Options::MergeFragment::AUTO) {
+        options.mergeFragment = Options::MergeFragment::NO;
       }
     } break;
   }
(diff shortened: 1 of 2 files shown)


c++/src/kj/compat/url.h
```diff
@@ -43,6 +43,14 @@ struct UrlOptions {
   // silently removed. In other words, setting this false causes consecutive slashes in the path or
   // consecutive ampersands in the query to be collapsed into one, whereas if true then they
   // produce empty components.
+
+  enum class MergeFragment {
+    kAuto, // Depend on the context if we merge the fragment into the previous component.
```
kentonv, line 48: We name enumerants (and constants in general) using UPPER_SNAKE_CASE. I know there are a lot of problems with this naming style, but it is what we use and we should stay consistent.

bcaimano: That's fine by me! I think the primary argument I've seen is macro collision. I'd rather people just use less macros when they can.

c++/src/kj/compat/url.c++
```diff
@@ -40,27 +40,27 @@ constexpr auto END_AUTHORITY = parse::anyOfChars("/?#");
...
+const parse::CharGroup_& getEndPathPart(Url::Options::MergeFragment mergeFragment) {
   static constexpr auto END_PATH_PART_HREF = parse::anyOfChars("/?#");
   static constexpr auto END_PATH_PART_REQUEST = parse::anyOfChars("/?");
 
-  switch (context) {
-    case Url::REMOTE_HREF:        return END_PATH_PART_HREF;
-    case Url::HTTP_PROXY_REQUEST: return END_PATH_PART_REQUEST;
-    case Url::HTTP_REQUEST:       return END_PATH_PART_REQUEST;
+  switch (mergeFragment) {
+    case Url::Options::MergeFragment::AUTO:  KJ_UNREACHABLE;
```
vlovich, line 48: It's not immediately clear to me this is actually unreachable, especially since the default value for the variable being supplied to this function is `AUTO`. Should this perhaps be a `KJ_REQUIRE` so that we generate an exception instead of undefined behavior?

Perhaps you're already aware (but in case you're not), `KJ_UNREACHABLE` expands to `__builtin_unreachable` which is a signal to the compiler that hitting this code path is impossible & the compiler can treat it as undefined behavior (aka UB) for optimization purposes (i.e. colloquially one should think of hitting a UB statement as potentially running `rm -rf /`).

For example, this is generally fine when outside of a an exhaustive `switch` statement (as seen below) over a closed enum (i.e. generally what enums internal to a program represent). The reason for this is C/C++ define enums as open & thus that knowledge isn't available to the compiler to understand your intention (hence the builtin). Arguably, this could have been fixed by making `enum class` closed by default or something but alas we're stuck with `__builtin_unreachable`.

Alternatively, you could make the `mergeFragment` variable a `Maybe` & get rid of the `AUTO` state. That would remove the `AUTO` state from switch statements (& thus ensure you did `KJ_REQUIRE_NONNULL` call when unwrapping).

c++/src/kj/compat/url.c++
```diff
@@ -260,21 +250,146 @@ Maybe<Url> Url::tryParse(StringPtr text, Context context, Options options) {
...
+    } break;
+  }
+
+  KJ_IF_MAYBE (parsedUrl, tryParseWithoutContext(text, options)) {
+    if (parsedUrl->isValidWithContext(context)) {
+      return kj::mv(*parsedUrl);
+    } else {
+      // TODO(now) Should I leave these logs in? They kinda feel like a poor man's equivalent of
+      // emitting exceptions.
+      KJ_LOG(INFO, "URL failed validation", text, *parsedUrl);
```
vlovich, line 301: They weren't here previously AFAICT so I'd vote for removing them.

c++/src/kj/compat/url.c++
```diff
@@ -260,21 +250,146 @@ Maybe<Url> Url::tryParse(StringPtr text, Context context, Options options) {
...
+      return nullptr;
+    }
+  } else {
+    KJ_LOG(INFO, "URL failed construction", text);
+    return nullptr;
+  }
```
vlovich, line 307: Can just do a simple `return nullptr` at the end without any `else` since all the `else` cases fall through the same way (once log statements are removed).

c++/src/kj/compat/url.h
```diff
@@ -139,9 +151,15 @@ struct Url {
   static Maybe<Url> tryParse(StringPtr text, Context context = REMOTE_HREF, Options options = {});
   // Parse an absolute URL.
 
+  static Maybe<Url> tryParseWithoutContext(StringPtr text, Options options = {});
```
vlovich, line 154: Totally optional nit: what about just changing `tryParse` to take a `kj::Maybe<Context>`? Then to not pass in context you just pass in `nullptr`? I could see an argument either way but figured I'd mention it just as a discussion point that can be ignored.

vlovich (commented):
I'd love to see some more test cases, particularly with regards to the new `ARBITRARY` enum as well as the new `isValidWithContext` code path.

c++/src/kj/compat/url.c++
bcaimano: I think I might have been a bit too coy with my TODO. I'm mildly concerned that we provide no channel for explaining why we rejected the URI. I'm guessing this hasn't been an issue in practice but I figured I should check to see if an exception-based form of tryParse would be valuable.

c++/src/kj/compat/url.c++
bcaimano: Agreed. What motivates you to ask for this? I was looking for the cleanliness of a double branch where each of the three paths returns without a jump.

c++/src/kj/compat/url.h
bcaimano: Hmmmmm, if you think that fits with the codebase style, that's probably a better choice for maintaining a smaller public API that we have to maintain. IIRC `Maybe` doesn't have an `explicit`-decorated ctor, so we should be able to do that without breaking people.

c++/src/kj/compat/url.c++
bcaimano: This is interesting. In the form I have now, you are correct, it is reachable if someone invokes `tryParseWithoutContext()`. That said, if I do your suggestion with `Maybe<Context>` in the public signature, this can become truly unreachable. I'd rather `KJ_FAIL_REQUIRE` here I think, that way I can explain why this should never happen.

I was a little surprised to read that it uses a `__builtin` since those usually aren't cross-platform. Traditionally, I've seen either an uncatchable exception thrown or an abort instead.

bcaimano:
> I'd love to see some more test cases, particularly with regards to the new `ARBITRARY` enum as well as the new `isValidWithContext` code path.

Sure, it's pretty easy to phrase these tests in an iterative way.

c++/src/kj/compat/url.h
```diff
@@ -124,10 +133,13 @@ struct Url {
     // path, and query, but omits userInfo (which should be used to construct the Authorization
     // header) and fragment (which should not be transmitted).
 
-    HTTP_REQUEST
+    HTTP_REQUEST,
     // The path to place in the first line of a regular HTTP request. This includes only the path
     // and query. Scheme, user, host, and fragment are omitted.
 
+    ARBITRARY,
+    // An arbitrary URL that may omit any component.
```
kentonv, line 141: So, what does this mean exactly? Is it meant to parse a relative URL? Or is it really an _absolute_ URL that can omit any component?

If it's an absolute URL missing components, can you name any existing applications where such a thing today? I feel like this isn't a well-defined concept and I'm not really sure if we should be inventing it from scratch.

If it's meant to be a relative URL, then I'm not sure if this is the right approach. Generally, the right way to parse a relative URL is by first obtaining the base URL, and then calling `parseRelative()` on that. I don't think the URL parser is set up to parse a relative URL without a base to evaluate it against. Two specific problems that I think come up here:

* How is a leading `../` handled? When parsing a URL relative to an existing URL, we evaluate these, so the final result doesn't contain `..`. But when parsing a truly relative URL I think you'd want to preserve them for later evaluation?
* Let's say we want to extend this class in the future to support non-HTTP URL schemes. Different schemes might have completely different parsing rules, right? So how can we parse a relative URL without as scheme, unless we have a base URL to evaluate against which defines the scheme?

Basically, I'm uncomfortable as I don't think this class was ever intended to represent a relative URL and it's not obvious to me that extending it that way won't lead to problems.

Sorry to bring this up late in the process. In the future we should probably try to discuss major design changes like this before code is written.

PR #1263, opened 2021-06-10
Make cmake install more portable on windows

VladimirVR:
Current implementation might produce the following CMake error:

```
CMake Error: failed to create symbolic link ...: operation not permitted
```


kentonv:
Does this mean that #1240 is no longer relevant? Should the changes from that PR be reverted?

VladimirVR:
> Does this mean that #1240 is no longer relevant? Should the changes from that PR be reverted?

I would say yes, because this change actually makes it work instead of just reporting a warning to user.
I can revert original line within my pull request.

prince-chrismc (approved):

harrishancock (approved):
Works for me, thanks!

PR #1264, opened 2021-06-18
Improve emacs packaging

syohex:
- add package footer
- add auto-mode-alist setting


kentonv:
Hmm I don't know anything about emacs. It looks like @netguy204 is the original author of these files. @netguy204, do you have thoughts on this? I don't know anything about emacs...

@syohex, note that #1265 also edits these files and was posted a couple days after your PR... weird coincidence?

syohex:
It looks #1265 is better. I've closed this PR.

PR #1266, opened 2021-06-24
Implement a lazy Maybe::orDefault

vlovich:
It's often times more terse/readable to use a lazy `orDefault` version
instead of writing out the KJ_IF_MAYBE logic by hand.

### DIFF
--- a/c++/src/kj/common-test.c++
+++ b/c++/src/kj/common-test.c++
@@ -83,7 +83,9 @@ TEST(Common, Maybe) {
     EXPECT_FALSE(ranLazy);
 
     KJ_IF_MAYBE(v, m) {
-      const int& ref = m.orDefault([notUsed = 5]() -> const int& { return notUsed; });
+      const int& ref = m.orDefault([notUsed = 5]() -> const int& {
+        return std::cref(notUsed);
+      });
 
       EXPECT_EQ(ref, *v);
       EXPECT_EQ(&ref, v);
--- a/c++/src/kj/common-test.c++
+++ b/c++/src/kj/common-test.c++
@@ -83,9 +83,7 @@ TEST(Common, Maybe) {
     EXPECT_FALSE(ranLazy);
 
     KJ_IF_MAYBE(v, m) {
-      const int& ref = m.orDefault([notUsed = 5]() -> const int& {
-        return std::cref(notUsed);
-      });
+      const int& ref = m.orDefault([notUsed = 5]() -> const int& { return notUsed; });
 
       EXPECT_EQ(ref, *v);
       EXPECT_EQ(&ref, v);
--- a/c++/src/kj/common-test.c++
+++ b/c++/src/kj/common-test.c++
@@ -83,7 +83,7 @@ TEST(Common, Maybe) {
     EXPECT_FALSE(ranLazy);
 
     KJ_IF_MAYBE(v, m) {
-      const int& ref = m.orDefault([notUsed = 5]() -> const int& { return notUsed; });
+      const int& ref = m.orDefault([notUsed = 5]() -> int& { return notUsed; });
 
       EXPECT_EQ(ref, *v);
       EXPECT_EQ(&ref, v);
--- a/c++/src/kj/common-test.c++
+++ b/c++/src/kj/common-test.c++
@@ -75,6 +75,37 @@ TEST(Common, Maybe) {
       ADD_FAILURE();
     }
     EXPECT_EQ(123, m.orDefault(456));
+    bool ranLazy = false;
+    EXPECT_EQ(123, m.orDefault([&] {
+      ranLazy = true;
+      return 456;
+    }));
+    EXPECT_FALSE(ranLazy);
+
+    KJ_IF_MAYBE(v, m) {
+      const int& ref = m.orDefault([notUsed = 5]() -> const int& { return notUsed; });
+
+      EXPECT_EQ(ref, *v);
+      EXPECT_EQ(&ref, v);
+
+      const int& ref2 = m.orDefault([notUsed = 5]() -> int { return notUsed; });
+      EXPECT_NE(&ref, &ref2);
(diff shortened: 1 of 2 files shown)


c++/src/kj/common.h
```diff
@@ -1379,6 +1379,25 @@ class Maybe {
     }
   }
 
+  template <typename F, typename RetOk = EnableIf<isSameType<T, decltype(instance<F>()())>()>>
```
kentonv, line 1382: Probably should use canConvert() instead of isSameType()?

vlovich: Is that testing implicit or explicit conversion? If the latter then there needs to be an additional T constructor to wrap the result, right?

kentonv: `canConvert()` only tests for implicit conversion, which is what we want here I think.

c++/src/kj/common.h
```diff
@@ -1377,6 +1377,25 @@ class Maybe {
     }
   }
 
+  template <typename F, typename RetOk = EnableIf<canConvert<T, decltype(instance<F>()())>()>>
+  T orDefault(F&& lazyDefaultValue) const & {
```
kentonv, line 1381: Hmm, looking closer at this, I guess it's sort of awkward that the `const &` version of this returns a `T` and not `const T&` (which is what the non-lazy version returns). This means this method will only work when `T` is a type that has a const copy constructor.

But if you change this to return `const T&`, then that requires that the callback function returns exactly `const T&`, which may be inconvenient in cases where `T` is in fact const-copyable (e.g. `Maybe<int>`).

I wonder if what we really need here is to let the compiler deduce an appropriate type, the same way it would for a ternary operator. That is:

```
template <typename F, typename Result = decltype(instance<bool>() ? instance<const T&>() : instance<F>()())>
Result orDefault(F&& lazyDefaultValue) const & {
```

I think in this case, if the lazy callback does return a `const T&`, then the compiler will deduce the result as `const T&`, but otherwise it'll figure out an appropriate middleground that both expressions can be cast to?

On another note, you are missing an overload for non-const &, which is important to cover due to the enforcement of transitive constness in KJ. You should probably also cover `const &&` for completeness, though admittedly that case rarely comes up in practice.

vlovich: The reason I omitted the `T&` was precisely for the reason of returning `T&`/`const T&`. I'll play around with this & see if what your recommending works out without any dangerous edge cases.

vlovich: I did the change & it seems to work. I'm a little paranoid that I may have missed a test case of some kind but I can't think what it might be. I spot checked some usages that should be illegal & they resulted in compilation errors but hard to know if I was thorough & those are always annoying to write unit tests for.

c++/src/kj/common-test.c++
```diff
@@ -75,6 +75,37 @@ TEST(Common, Maybe) {
...
     EXPECT_EQ(123, m.orDefault(456));
+    bool ranLazy = false;
+    EXPECT_EQ(123, m.orDefault([&] {
+      ranLazy = true;
+      return 456;
+    }));
+    EXPECT_FALSE(ranLazy);
+
+    KJ_IF_MAYBE(v, m) {
+      const int& ref = m.orDefault([notUsed = 5]() -> const int& { return notUsed; });
```
vlovich, line 86: Hmm... this fails to build on MSVC because for some reason it selects the `&` override instead of `const &` & fails with a hard error instead of continuing with other overloads (but how did the `Result` template parameter even get evaluated?! Should have been a SFINAE failure. It's also a mystery why it says the evaluation of `lazyDefaultValue()` is returning a `const int` when it explicitly returns a `const int&` (i.e. I'd expect the error message to say it can't convert a `const int &` to `int&`

c++/src/kj/common.h
```diff
@@ -1377,6 +1377,46 @@ class Maybe {
...
+    if (ptr == nullptr) {
+      return lazyDefaultValue();
+    } else {
+      return *ptr;
+    }
+  }
+
+  template <typename F,
+      typename Result = decltype(instance<bool>() ? instance<const T&>() : instance<F>()())>
+  const Result orDefault(F&& lazyDefaultValue) const & {
```
kentonv, line 1392: I think `Result` will already be derived to include `const` where needed, and declaring it `const` again here doesn't mean anything.

c++/src/kj/common-test.c++
kentonv: I think that choosing the `&` overload is correct. In this case, `decltype(instance<bool>() ? instance<T&>() : instance<F>()())` should find that the branches of the `?:` are `int&` and `const int&`, which it should reconcile by choosing `const int&` as the result type, since that's the common type both can be converted to. That is to say, the following is valid according to GCC and Clang:

```
const int& foo(bool b, int& r1, const int& r2) {
  return b ? r1 : r2;
}
```

But it appears that MSVC is incorrectly choosing `int&` as the result type. That seems like an MSVC bug.

I think we can adjust the test to avoid this case, by either returning a non-const reference here, or returning an rvalue. I don't think people will commonly return references from these callbacks anyway.

c++/src/kj/common-test.c++
```diff
@@ -75,6 +75,37 @@ TEST(Common, Maybe) {
...
     EXPECT_EQ(123, m.orDefault(456));
+    bool ranLazy = false;
+    EXPECT_EQ(123, m.orDefault([&] {
+      ranLazy = true;
+      return 456;
+    }));
+    EXPECT_FALSE(ranLazy);
+
+    KJ_IF_MAYBE(v, m) {
+      const int& ref = m.orDefault([notUsed = 5]() -> int& { return notUsed; });
```
kentonv, line 86: Compiler error because the lambda isn't mutable, but it's also sorta weird that you're returning a reference to a capture which will go away immediately when the lambda is destroyed. Maybe return a reference to a local variable on the stack instead?

PR #1270, opened 2021-06-28
Fix release build for GCC 11

### LINKED ISSUE #1268: __lsan_ignore_object attribute fails with GCC 11
#1255 causes a build failure with GCC 11.1.0, since it's applying a GCC attribute to a function definition rather than a declaration:

```
In file included from src/kj/memory.h:24,
                 from src/kj/exception.h:24,
                 from src/kj/exception.c++:37:
src/kj/common.h:197:50: error: attributes are not allowed on a function-definition                             
  197 | #define KJ_ALWAYS_INLINE(...) inline __VA_ARGS__ __attribute__((always_inline))                        
      |                                                  ^~~~~~~~~~~~~
src/kj/exception.c++:93:8: note: in expansion of macro ‘KJ_ALWAYS_INLINE’
   93 | static KJ_ALWAYS_INLINE(void __lsan_ignore_object(const void* p)) {}
      |        ^~~~~~~~~~~~~~~~
```

Adding a separate declaration before the dummy definition of `__lsan_ignore_object` is one possible fix:

```
static KJ_ALWAYS_INLINE(void __lsan_ignore_object(const void* p));
static void __lsan_ignore_object(const void* p) {}
```

vlovich:
Fixes #1268

### DIFF
--- a/c++/src/kj/exception.c++
+++ b/c++/src/kj/exception.c++
@@ -90,7 +90,7 @@
 #if KJ_HAS_COMPILER_FEATURE(address_sanitizer) || defined(__SANITIZE_ADDRESS__)
 #include <sanitizer/lsan_interface.h>
 #else
-static KJ_ALWAYS_INLINE(void __lsan_ignore_object(const void* p)) {}
+static void __lsan_ignore_object(const void* p) {}
 #endif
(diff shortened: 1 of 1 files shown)


kentonv:
(This is actually a fix for release builds under GCC in general, not just GCC 11.)

PR #1273, opened 2021-07-05
Annotate all fallthroughs.

kentonv:
This allows the code to compile with -Wimplicit-fallthrough on both GCC and Clang. Note that Clang's version of the warning doesn't seem to accept comments the way GCC's does (or I couldn't figure out how to make a comment that it accepted), so I am using the explicit annotation everywhere instead. Technically this annotation wasn't recognized until C++17, but GCC seems to accept it with C++14. I'm hoping the others do to (we'll see what CI says), otherwise I'll need a KJ_FALLTHROUGH macro I guess.

### DIFF
--- a/c++/src/capnp/compiler/capnp.c++
+++ b/c++/src/capnp/compiler/capnp.c++
@@ -904,7 +904,7 @@ private:
               state = COMMENT;
               break;
             }
-            // fallthrough
+            KJ_FALLTHROUGH;
           case NORMAL:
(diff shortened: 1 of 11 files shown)


kentonv:
Dang. MSVC 2015 doesn't support the annotation by default. I guess I'll make a macro.

vlovich:
LGTM.

a-robinson (approved):

PR #1274, opened 2021-07-07
Clarify that kj's async Timer only updates when the event loop waits

jyn514:
Th original comment was misleading to me. I understood it to mean "Timer counts
on its own, but may get slightly out of sync until the start of the event loop,"
especially because of the "guaranteed to be synchronized" bit. The actual
meaning is "At any given point, Timer is frozen in time, and is only updated at
the start of the event loop."

cc @a-robinson

### DIFF
--- a/c++/src/kj/timer.h
+++ b/c++/src/kj/timer.h
@@ -41,6 +41,13 @@ class Timer: public MonotonicClock {
   // guaranteed to be synchronized with the `MonotonicClock` returned by
   // `systemPreciseMonotonicClock()` (or, more precisely, is updated to match that clock whenever
   // the loop waits).
+  //
+  // Note that the value returned by `Timer::now()` only changes each time the
+  // event loop waits for I/O from the system. While the event loop is actively
+  // running, the time stays constant. This is intended to make behavior more
+  // deterministic and reproducible. However, if you need up-to-the-cycle
+  // accurate time, then `Timer::now()` is not appropriate. Instead, use
+  // `systemPreciseMonotonicClock()` directly in this case.
 
 public:
   virtual TimePoint now() const = 0;


a-robinson (approved):
Looks reasonable to me, but I'll leave for Kenton to decide on.

c++/src/kj/timer.h
```diff
@@ -38,9 +38,8 @@ class Timer: public MonotonicClock {
   // date as tracked by the system is manually modified.
   //
   // That said, the `Timer` returned by `kj::setupAsyncIo().provider->getTimer()` in particular is
-  // guaranteed to be synchronized with the `MonotonicClock` returned by
-  // `systemPreciseMonotonicClock()` (or, more precisely, is updated to match that clock whenever
-  // the loop waits).
+  // based on the `MonotonicClock` returned by `systemPreciseMonotonicClock()`, but only updates
+  // when the event loop waits.
```
kentonv, line 42: Hmm, to me the change doesn't really make the text any more or less understandable.

What about adding a whole new paragraph emphasizing the update-on-wait behavior? Something like:

> Note that the value returned by `Timer::now()` only changes each time the event loop waits for I/O from the system. While the event loop is actively running, the time stays constant. This is intended to make behavior more deterministic and reproducible. However, if you need up-to-the-cycle accurate time, then `Timer::now()` is not appropriate. Instead, use `systemPreciseMonotonicClock()` directly in this case.

FWIW, `now()`'s documentation also mentions this behavior. This may just be a case where no amount of documentation is going to help, because few people are going to carefully read the docs...

jyn514: > What about adding a whole new paragraph emphasizing the update-on-wait behavior? Something like:

Sure, done.

PR #1276, opened 2021-07-09
Make TSAN happy about variable we double-check on destruction

vlovich:
Not actually a bug but TSAN is unhappy that this value is read
non-atomically (even though the non-atomic read here is safe since we're
destroying the object anyway).

### DIFF
--- a/c++/src/kj/async-xthread-test.c++
+++ b/c++/src/kj/async-xthread-test.c++
@@ -29,6 +29,8 @@
 #include "mutex.h"
 #include <kj/test.h>
 
+#include <atomic>
+
 #if _WIN32
 #include <windows.h>
 #include "windows-sanity.h"
@@ -298,13 +300,16 @@ KJ_TEST("cancel cross-thread event before it runs") {
       exec = &KJ_ASSERT_NONNULL(*lock);
     }
 
-    volatile bool called = false;
+    std::atomic<bool> called(false);
     {
-      Promise<uint> promise = exec->executeAsync([&]() { called = true; return 123u; });
+      Promise<uint> promise = exec->executeAsync([&]() {
+        called.store(true, std::memory_order_relaxed);
+        return 123u;
+      });
       delay();
       KJ_EXPECT(!promise.poll(waitScope));
     }
-    KJ_EXPECT(!called);
+    KJ_EXPECT(!called.load(std::memory_order_relaxed));
 
     *executor.lockExclusive() = nullptr;
   })();
@@ -342,12 +347,12 @@ KJ_TEST("cancel cross-thread event while it runs") {
     }
 
     {
-      volatile bool called = false;
+      std::atomic<bool> called(false);
       Promise<uint> promise = exec->executeAsync([&]() -> kj::Promise<uint> {
-        called = true;
+        called.store(true, std::memory_order_relaxed);
         return kj::NEVER_DONE;
       });
-      while (!called) {
+      while (!called.load(std::memory_order_relaxed)) {
         delay();
       }
       KJ_EXPECT(!promise.poll(waitScope));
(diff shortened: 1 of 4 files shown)


vlovich:
I'm still seeing a bunch more TSAN failures in the following three tests when run locally:

[kj/async-xthread-test](https://pastebin.com/zifhLDeZ)
[kj/async-unix-xthread-test](https://pastebin.com/p0pAEZyP)
[capnp/rpc-twoparty-test](https://pastebin.com/CN8kh9F1)

These all seem to contain much more subtle/difficult race conditions that will take much more time to resolve.

c++/src/kj/mutex.c++
```diff
@@ -210,7 +210,15 @@ Mutex::AcquiredMetadata Mutex::lockedInfo() const {
 Mutex::Mutex(): futex(0) {}
 Mutex::~Mutex() {
   // This will crash anyway, might as well crash with a nice error message.
+
+#if defined(__clang__) || !defined(_MSC_VER)
```
kentonv, line 214: The futex code is Linux-only so I don't think you need this conditional.

c++/src/kj/async-xthread-test.c++
```diff
@@ -49,6 +49,30 @@ inline void delay() { usleep(10000); }
 namespace kj {
 namespace {
 
+// Some trampolines to make TSAN happy about having to read/write volatile values.
+#if defined(_MSC_VER) && !__clang__
+template <typename T>
+inline T xthread_read(volatile T& var) {
```
kentonv, line 55: Names should use KJ style.

But I'd be totally fine with just using std::atomic in these unit tests, which I think would be a lot clearer. (The only place where std::atomic bothers me is in public headers... but even there I think I'm coming around to it.)

c++/src/kj/async-xthread-test.c++
```diff
@@ -298,13 +300,16 @@ KJ_TEST("cancel cross-thread event before it runs") {
...
-      Promise<uint> promise = exec->executeAsync([&]() { called = true; return 123u; });
+      Promise<uint> promise = exec->executeAsync([&]() {
+        called.store(true, std::memory_order::relaxed);
+        return 123u;
+      });
       delay();
       KJ_EXPECT(!promise.poll(waitScope));
     }
-    KJ_EXPECT(!called);
+    KJ_EXPECT(!called.load(std::memory_order::relaxed));
```
kentonv, line 312: Should be `std::memory_order_relaxed`

c++/src/kj/mutex.c++
```diff
@@ -210,7 +210,10 @@ Mutex::AcquiredMetadata Mutex::lockedInfo() const {
 Mutex::Mutex(): futex(0) {}
 Mutex::~Mutex() {
   // This will crash anyway, might as well crash with a nice error message.
-  KJ_ASSERT(futex == 0, "Mutex destroyed while locked.") { break; }
+  // Atomic load to make TSAN happy.
```
kentonv, line 213: Wait, hold on... if TSAN is alerting on this, doesn't that suggest an actual problem? It suggests that a thread is trying to destroy a Mutex without having actually synchronized with the last thread that unlocked the Mutex. That suggests it's possible that, under the right circumstances, the Mutex could end up being destroyed while still locked.

c++/src/kj/refcount.c++
```diff
@@ -48,7 +48,12 @@ void Refcounted::disposeImpl(void* pointer) const {
 // Atomic (thread-safe) refcounting
 
 AtomicRefcounted::~AtomicRefcounted() noexcept(false) {
+#if defined(__clang__) || !defined(_MSC_VER)
+  KJ_ASSERT(__atomic_load_n(&refcount, __ATOMIC_RELAXED) == 0,
```
kentonv, line 52: Similarly here... in correct operation, this thread should have been the one which decremented the refcount to zero. If this thread was the last one to access the value, and no other thread will access it in the future, then it should not be a problem for it to read the value non-atomically. Why is TSAN alerting on this?

vlovich:
You were totally write about the missing annotations being the likely culprit behind all the false positives. I'll take a stab at adding them to KJ.

PR #1279, opened 2021-07-13
Add new macro[s] to explicitly disallow copy and move ctors

bcaimano:


vlovich:
My understanding is that this is to try to solve the problem that someone uses a `KJ_DISALLOW_COPY` type & only tests release builds & thus `auto x = NonCopyable()` passes in release and fails in debug. Is that the gist of the problem?

Personally, the distinction between `KJ_DISALLOW_COPY` and `KJ_DISALLOW_COPY_AND_MOVE` is subtle enough in practice, that I wonder if it's not better solved through better CI coverage. At a minimum, I think this could use with some documentation that describes when to use one vs the other (e.g. "Rule of thumb is `KJ_DISALLOW_COPY_AND_MOVE` unless you need to be able to move the type).

kentonv:
@vlovich No, that's not it. Compilers are only allowed to elide a copy/move in the case that it would have been legal. Optimization level doesn't change what errors are reported here.

My understanding is that @bcaimano feels that the compiler produces better error messages if the move constructors are explicitly deleted, compared to the case where they are not defined at all but the copy constructor is deleted.

@bcaimano, can you give a specific example of where this produces better error messages?

bcaimano:
In short, compilers default to complaints about copy even in contexts where you'd expect them to use the move ctor: https://godbolt.org/z/EhKq84qMP. The underlying reason, to my understanding, is that defining the copy ctor but not the move ctor prevents the implicit generation of the move ctor. This is not the same thing as implicitly generating a deleted move ctor.

There's also the milder gripe that KJ_DISALLOW_COPY is ambiguous for resource types. Is there a move ctor defined somewhere else in the class? Specifying the full rule of five (or a ctor/macro that is equivalent to it) means you know at a glance what copy/move semantics for the class are.

Regarding the elision, the rules are more permissive than you may think and vary by compiler (https://en.cppreference.com/w/cpp/language/copy_elision). Sadly, the phrase "permitted but not required" comes up here. Note that in the godbolt above, gcc permits elision but clang does not.

bcaimano:
Closing since https://github.com/capnproto/capnproto/pull/1594 landed. (Didn't even remember that this was open.)

PR #1280, opened 2021-07-17
Fix release build

kentonv:


kentonv:
@harrishancock this is actually ready for review now.

kentonv:
It appears that GitHub Actions' ubuntu-latest is busted, can't install cmake anymore. Ironically, this makes all the Quick Tests fail, but the Release Tests succeed because they are inexplicably using ubuntu-18.04 instead. Maybe I'll switch it.

kentonv:
Oh duh I misread, it's quick-tests that were weirdly using ubuntu-18.04, and ubuntu-latest that works fine.

kentonv:
Oh, right, we were using ubuntu-18.04 because it has older compilers available. Sigh.

PR #1281, opened 2021-07-20
Cleanup mutex tests & SourceLocation equality

vlovich:
Rather than using magic line numbers, just supply the location used into
the functions so that we can just use plain equality in mutex tests.

Similarly, define an equality operator for SourceLocation so that they can be
used with `KJ_REQUIRE` macros.

### DIFF
--- a/c++/src/kj/source-location.h
+++ b/c++/src/kj/source-location.h
@@ -63,12 +63,21 @@ public:
   {}
 #endif
 
+#if KJ_COMPILER_SUPPORTS_SOURCE_LOCATION
+  // This can only be exposed if we actually generate valid SourceLocation objects as otherwise all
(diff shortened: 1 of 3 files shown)


c++/src/kj/source-location.h
```diff
@@ -63,12 +63,17 @@ class SourceLocation {
   {}
 #endif
 
+  constexpr bool operator==(const SourceLocation& o) const {
```
vlovich, line 66: Hmm... I think this can only be exposed under `KJ_COMPILER_SUPPORTS_SOURCE_LOCATION` as otherwise all `SourceLocation` instances would be equal & that's a nonsensical result.

vlovich: Fixed in latest revision.

PR #1284, opened 2021-07-22
Print reader contention location

vlovich:
This should provide an additional breadcrumb to debug the contended mutex case.

In the future we might want to hide this behind a dedicated separate macro to avoid
the bloat of the `kj::Mutex`.

### DIFF
--- a/c++/src/kj/mutex.c++
+++ b/c++/src/kj/mutex.c++
@@ -258,6 +258,9 @@ bool Mutex::lock(Exclusivity exclusivity, Maybe<Duration> timeout, LockSourceLoc
         }
       }
       acquiredExclusive(TRACK_ACQUIRED_TID(), location);
+#if KJ_CONTENTION_WARNING_THRESHOLD
+      printContendedReader = false;
+#endif
       break;
     case SHARED: {
       uint state = __atomic_add_fetch(&futex, 1, __ATOMIC_ACQUIRE);
@@ -294,6 +297,15 @@ bool Mutex::lock(Exclusivity exclusivity, Maybe<Duration> timeout, LockSourceLoc
         }
         state = __atomic_load_n(&futex, __ATOMIC_ACQUIRE);
       }
+
+#ifdef KJ_CONTENTION_WARNING_THRESHOLD
+      if (__atomic_load_n(&printContendedReader, __ATOMIC_RELAXED)) {
+        // Double-checked lock avoids the CPU needing to acquire the lock in most cases.
+        if (__atomic_exchange_n(&printContendedReader, false, __ATOMIC_RELAXED)) {
+          KJ_LOG(WARNING, "Acquired contended lock", location, kj::getStackTrace());
+        }
+      }
+#endif
 
       // We just want to record the lock being acquired somewhere but the specific location doesn't
       // matter. This does mean that race conditions could occur where a thread might read this
@@ -364,6 +376,18 @@ void Mutex::unlock(Exclusivity exclusivity, Waiter* waiterToSkip) {
         }
       }
 
+#ifdef KJ_CONTENTION_WARNING_THRESHOLD
+      uint readerCount;
+      {
+        uint oldState = __atomic_load_n(&futex, __ATOMIC_RELAXED);
+        readerCount = oldState & SHARED_COUNT_MASK;
+        if (readerCount >= KJ_CONTENTION_WARNING_THRESHOLD) {
+          // Atomic not needed because we're still holding the exclusive lock.
+          printContendedReader = true;
+        }
+      }
+#endif
+
       // Didn't wake any waiters, so wake normally.
       uint oldState = __atomic_fetch_and(
           &futex, ~(EXCLUSIVE_HELD | EXCLUSIVE_REQUESTED), __ATOMIC_RELEASE);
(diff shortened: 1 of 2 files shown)


vlovich:
Hmm... @kentonv Linux CI seems busted. It's trying to fetch http://azure.archive.ubuntu.com/ubuntu/pool/main/c/cmake/cmake_3.10.2-1ubuntu2.18.04.1_amd64.deb but that URL is only hosting 18.04.2_amd64.deb. https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu1804-README.md seems to indicate it should be on 18.04.5 LTS anyway but I could find no knob in the workflow actions to specify the actual revision to be used.

c++/src/kj/mutex.c++
```diff
@@ -364,6 +373,19 @@ void Mutex::unlock(Exclusivity exclusivity, Waiter* waiterToSkip) {
...
       }
 
+#ifdef KJ_CONTENTION_WARNING_THRESHOLD
+      {
+        uint oldState = __atomic_load_n(&futex, __ATOMIC_RELAXED);
+        uint readerCount = oldState & SHARED_COUNT_MASK;
+        if (readerCount >= KJ_CONTENTION_WARNING_THRESHOLD) {
+          // Atomic not needed because we're still holding the exclusive lock.
+          printContendedReader = true;
+          KJ_LOG(WARNING, "excessively many readers were waiting on this lock", readerCount,
```
kentonv, line 383: Maybe we should still delay this `KJ_LOG` until after releasing the lock? So we're not making the readers wait through yet another syscall...

c++/src/kj/mutex.c++
```diff
@@ -295,6 +298,12 @@ bool Mutex::lock(Exclusivity exclusivity, Maybe<Duration> timeout, LockSourceLoc
         state = __atomic_load_n(&futex, __ATOMIC_ACQUIRE);
       }
 
+#ifdef KJ_CONTENTION_WARNING_THRESHOLD
+      if (__atomic_exchange_n(&printContendedReader, false, __ATOMIC_RELAXED)) {
```
kentonv, line 302: This exchange is an expensive atomic requiring hardware-level locking. Can we do a double-check, i.e. check if the value is `true` first using `__atomic_load_n` (which is cheap), and only do the exchange if so?

c++/src/kj/mutex.c++
vlovich: Hmm.... I would imagine the backtrace, string formatting, & memory allocations dominate the cost of the actual write syscall. Even still I think it's OK to have this here since this is a rare code path & when I was debugging this issue I would always end up moving this log statement here (makes it easier to set a breakpoint & look at the stack traces of other threads).

The other reason I want this here is that in the future I want to kick off a contention walk of 1 or more shared mutexes to get the lock location or even the whole stack trace of where the readers are at. Usually that information is sufficient to analyze the problem & fix it (barring some sources of real contention I suspect we do have). I don't have immediate plans to do that extension in the near term as I worried about scope creep for now. I want to land the contention walker in EW & make sure it's working to help disambiguate the SIGINT stuff well before we expand its use-case slightly to handle this issue.

c++/src/kj/mutex.c++
vlovich: Would that essentially be the same thing as writing it via `atomic_compare_exchange`? I'm a little surprised that exchange is more expensive than doing a double-check although if I think about it I could see why that's the case. Is there a good tutorial you've read on the different atomic operations & the performance mental model I should have about them?

vlovich: Oh interesting. CAS is equivalent to atomic_exchange here but the double-checked lock is much faster. Not sure I understand why. https://quick-bench.com/q/2nDIxLb7d9bptd-stO03ePUh8x8

c++/src/kj/mutex.c++
kentonv: Oh yeah, the backtrace is also a pretty expensive thing to do while still holding a lock that we know is highly contended.

All I'm suggesting is delaying the actual logging to happen after the futex has been signaled. Yes, I understand it's not common but it's also not hard to defer the logging...

c++/src/kj/mutex.c++
kentonv: A relaxed atomic load is "free" -- does not require any hardware-level locking. (On x86, an acquire-load is also "free", since memory ordering is guaranteed consistent.)

But any operation that has to do a read and a write together, atomically, requires hardware-level locking to make sure the read+write appears atomic to other cores.

I suppose arguably a compare-and-swap _could_ have the double-check optimization built into hardware, i.e. the CPU could do the comparison without a lock first, and skip the lock if it fails. But, that "optimization" might not be a win in most use cases, and it's easy to let software do the optimization if it wants it, so I wouldn't expect CPUs attempt such optimization.

So basically, with the double-checked approach here you avoid a hardware lock in the common case.

c++/src/kj/mutex.c++
```diff
@@ -364,6 +376,18 @@ void Mutex::unlock(Exclusivity exclusivity, Waiter* waiterToSkip) {
         }
       }
 
+#ifdef KJ_CONTENTION_WARNING_THRESHOLD
+      uint readerCount;
+      {
+        uint oldState = __atomic_load_n(&futex, __ATOMIC_RELAXED);
+        uint readerCount = oldState & SHARED_COUNT_MASK;
```
kentonv, line 383: Need to remove `uint` here.

vlovich: Oof. That would have been bad. Let me add a unit test just to make sure I don't accidentally break it again.

PR #1285, opened 2021-07-23
Add constexpr annotation to misc StringPtr methods

vlovich:
Make begin/end constexpr as they are for ArrayPtr (+ make comparison
against nullptr similarly constexpr).


kentonv:
LGTM but can you rebase on master to get the CI fix?

PR #1286, opened 2021-07-26
Lock contention tests

vlovich:
#1284 might have accidentally broken the lock contention code if not caught in code review. Add a unit test to protect accidental breakage in the future.

### DIFF
--- a/c++/src/kj/mutex-test.c++
+++ b/c++/src/kj/mutex-test.c++
@@ -42,6 +42,10 @@
 #include <unistd.h>
 #endif
 
+#ifdef KJ_CONTENTION_WARNING_THRESHOLD
+#include <vector>
+#endif
+
 #if KJ_TRACK_LOCK_BLOCKING
 #include <syscall.h>
 #include <signal.h>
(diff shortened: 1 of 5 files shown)


c++/src/kj/mutex-test.c++
```diff
@@ -855,5 +859,40 @@ KJ_TEST("get location of shared mutex") {
...
+  LockSourceLocation exclusiveLockLocation;
+  mutex.lock(_::Mutex::EXCLUSIVE, nullptr, exclusiveLockLocation);
+
+  bool seenContendedLockLog = false;
+
+  auto threads = kj::heapArrayBuilder<kj::Own<kj::Thread>>(KJ_CONTENTION_WARNING_THRESHOLD);
+  for (auto i: kj::zeroTo(KJ_CONTENTION_WARNING_THRESHOLD)) {
+    (void)i;
+    threads.add(kj::heap<kj::Thread>([&mutex, &seenContendedLockLog]() {
+      _::LogExpectation expectation(LogSeverity::WARNING, "Acquired contended lock");
```
kentonv, line 874: `_::LogExpectation` is meant to implement `KJ_EXPECT_LOG`. It's kind of weird to go extend `KJ_EXPECT_LOG`'s privaet implementation to serve another purpose and then directly call into it from somewhere else... Maybe better to implement a custom `ExceptionCallback` for this test?

c++/src/kj/test-helpers.c++
```diff
@@ -32,13 +32,8 @@ namespace kj {
...
 
 bool hasSubstring(StringPtr haystack, StringPtr needle) {
-  // TODO(perf): This is not the best algorithm for substring matching.
   if (needle.size() <= haystack.size()) {
-    for (size_t i = 0; i <= haystack.size() - needle.size(); i++) {
-      if (haystack.slice(i).startsWith(needle)) {
-        return true;
-      }
-    }
+    return strstr(haystack.cStr(), needle.cStr()) != nullptr;
```
kentonv, line 36: Technically this breaks support for strings with NUL bytes in them... did you verify it's never used with NUL bytes?

vlovich: No I didn't try NUL bytes as I clearly didn't think of it (although apparently none of the in-tree tests require this). I'll switch to memmem on Linux then.

c++/src/kj/mutex-test.c++
vlovich: Yeah, I was just being lazy.

c++/src/kj/test-helpers.c++
kentonv: Admittedly NUL bytes are probably never used in test strings, but OTOH the performance of this helper probably doesn't really matter since it's just for tests.

c++/src/kj/test-helpers.c++
```diff
@@ -28,17 +32,35 @@
...
 
+#include <functional>
+
 namespace kj {
 namespace _ {  // private
 
 bool hasSubstring(StringPtr haystack, StringPtr needle) {
-  // TODO(perf): This is not the best algorithm for substring matching.
   if (needle.size() <= haystack.size()) {
+    // Boyer Moore Horspool wins https://quick-bench.com/q/RiKdKduhdLb6x_DfS1fHaksqwdQ
```
kentonv, line 42: Well, you've done the work now, but I feel like this was a lot of effort spent optimizing a test helper that only ever runs on small strings anyway...

kentonv (approved):
LGTM whenever the CI gets around to running.

c++/src/kj/test-helpers.c++
vlovich: Yeah... had some time to kill waiting for CI :).

kentonv:
Canceled and re-queued the CI, seems to be actually running now.

PR #1287, opened 2021-07-27
Fix UBSAN errors.

kentonv:
Motivated by: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=33710

Clusterfuzz considers any UBSAN violation to be a possible security flaw. In practice, none of the actual errors are likely to be exploitable.

This commit fixes all errors reported by `capnp-test` with UBSAN enabled.

### DIFF
--- a/c++/src/capnp/dynamic.c++
+++ b/c++/src/capnp/dynamic.c++
@@ -1733,15 +1733,26 @@ int64_t unsignedToSigned<int64_t>(unsigned long long value) {
 
 template <typename T, typename U>
 T checkRoundTrip(U value) {
-#if __aarch64__
-  // Work around an apparently broken compiler optimization on Clang / arm64. It appears that
-  // for T = int8_t, U = double, and value = 128, the compiler incorrectly believes that the
-  // round-trip does not change the value, where in fact it should change to -128. Similar problems
-  // exist for various other types and inputs -- json-test seems to exercise several problem cases.
-  // The problem only exists when compiling with optimization. In any case, declaring the variable
-  // `volatile` kills the optimization.
-  volatile
-#endif
+  T result = value;
+  KJ_REQUIRE(U(result) == value, "Value out-of-range for requested type.", value) {
+    // Use it anyway.
+    break;
+  }
+  return result;
+}
+
+template <typename T, typename U>
+T checkRoundTripFromFloat(U value) {
+  // When `U` is `float` or `double`, we have to use a different approach, because casting an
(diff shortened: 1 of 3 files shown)


kentonv:
We should probably add a UBSAN pass to CI. I would have done it here but I wasn't able to immediately figure out how to make it abort on failures (as opposed to just printing a message).

cc @vlovich

c++/src/capnp/dynamic.c++
```diff
@@ -1733,15 +1733,26 @@ int64_t unsignedToSigned<int64_t>(unsigned long long value) {
...
+  // When `U` is `float` or `double`, we have to use a different approach, because casting an
+  // out-of-range float to an integer is, surprisingly, UB.
+  constexpr T MIN = kj::minValue;
+  constexpr T MAX = kj::maxValue;
+  KJ_REQUIRE(value >= U(MIN), "Value out-of-range for requested type.", value) {
+    return MIN;
+  }
+  KJ_REQUIRE(value <= U(MAX), "Value out-of-range for requested type.", value) {
+    return MAX;
+  }
```
jclee, line 1755: I guess if value is exactly min or max, it's in range for the conversion but we'll still warn for it?  Or is this maybe to handle the case where high-magnitude 64-bit ints don't have exact representation as a double?

vlovich:
I think `-fno-sanitize-recover=all` is the magic to make UBSAN errors hard (GCC docs appear to be a bit better here: https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html, but I don't know if clang implements `all`... generally my impression is they share the same underlying implementation & it just varies which compiler version picked up which snapshot of the sanitizer codebase).

jclee (approved):

c++/src/capnp/dynamic.c++
kentonv: We're converting from a floating-point (`U`, which is either `float` or `double`) to an integer (`T`). The goal is to verify that this conversion does not lose information. It turns out that this conversion is undefined behavior when the floating point would overflow or underflow the integer type, so we have to check for that case first. We also want to reject the float if it's not an integer value itself, which we do by checking if it round-trips.

A value of exactly MIN or MAX should pass the asserts here. Not sure what you mean about warning for it -- there are no warnings here, only exceptions.

jclee: Got it... I was misinterpreting the checks, and also needed to refresh my understanding of how the recovery blocks worked.

Issue #1208, opened 2021-04-09
Building OpenBSD 6.9 error: use of undeclared identifier 'getcontext'


murilobsd:
```
# os
OpenBSD 6.9 amd64
# clang version
OpenBSD clang version 10.0.1
Target: amd64-unknown-openbsd6.9
Thread model: posix
InstalledDir: /usr/bin
```

Process:

```
git clone https://github.com/sandstorm-io/capnproto.git
cd capnproto/c++
export AUTOCONF_VERSION=2.71
export AUTOMAKE_VERSION=1.16
export LIBTOOL=/usr/local/lib/libtool
autoreconf -i
./configure CC=clang CXX=clang++
make -j6 check
```
[configure.log](https://envs.sh/cp.log)

Out:

```
libtool: compile:  clang++ -DHAVE_CONFIG_H -I. -I./src -I./src -DKJ_HEADER_WARNINGS -DCAPNP_HEADER_WARNINGS -DCAPNP_INCLUDE_DIR=\"/usr/local/include\" -pthread -O2 -DNDEBUG -pthread -DKJ_HAS_ZLIB -DKJ_HAS_OPENSSL -MT src/capnp/rpc.capnp.lo -MD -MP -MF src/capnp/.deps/rpc.capnp.Tpo -c src/capnp/rpc.capnp.c++  -fPIC -DPIC -o src/capnp/.libs/rpc.capnp.o
src/kj/async.c++:56:10: fatal error: 'ucontext.h' file not found
#include <ucontext.h>
         ^~~~~~~~~~~~
1 error generated.
*** Error 1 in target 'src/kj/async.lo'
src/kj/async-io-unix.c++:1446:20: error: variable has incomplete type 'struct ucred'
      struct ucred creds;
                   ^
src/kj/async-io-unix.c++:1446:14: note: forward declaration of 'ucred'
      struct ucred creds;
             ^
```

OpenBSD doesn't have <ucontext.h>. ucontext_t lives in <signal.h>, so i change:

[aex/kj/async.c++](https://github.com/capnproto/capnproto/blob/6b5bcc2c6e954bc6e167ac581eb628e5a462a469/c%2B%2B/src/kj/async.c%2B%2B#L56)
```c
#if __OpenBSD__
#include <signal.h>
#else
#include <ucontext.h>
#endif
```
and now the problem is this! Can someone guide me about this error?

```
src/kj/async.c++:1354:16: error: use of undeclared identifier 'getcontext'; did you mean
[...]

PR #1289, opened 2021-07-31
Fix stack overflow potential destroying TaskSet

vlovich:
If the size of the linked list is really large (e.g. 20000000), it's easy to generate a stack overflow
because the compiler will unwind the object graph through the stack. Switch this to unwind with
a fixed size loop to avoid that problem. Obviously a large list like this can take time to unwind
anyway, so it may be worthwhile to consider adding limits to how large a task set that is directly
under untrusted control is allowed to get.

I don't have a Windows machine so I haven't verified that the stack test without the fix fails on its
own there but presumably this simple test uses roughly the same amount of stack on all platforms.

### DIFF
--- a/c++/src/kj/async-test.c++
+++ b/c++/src/kj/async-test.c++
@@ -25,6 +25,19 @@
 #include "mutex.h"
 #include "thread.h"
 
+// TODO(later): Refactor this more cleanly so that the KJ library itself defines this.
+#if !__BIONIC__
+#define KJ_FIBERS_AVAILABLE 1
+#else
+#define KJ_FIBERS_AVAILABLE 0
+#endif
+
+#if !KJ_FIBERS_AVAILABLE
+// For bionic & OpenBSD, enables the regression test for kj::TaskSet::~TaskSet potentially causing a
+// stack overflow.
+#include <pthread.h>
+#endif
+
 namespace kj {
 namespace {
 
@@ -733,6 +746,45 @@ TEST(Async, TaskSet) {
   EXPECT_EQ(1u, errorHandler.exceptionCount);
 }
 
+TEST(Async, LargeTaskSetDestruction) {
+  static constexpr size_t stackSize = 200 * 1024;
+
+  static auto testBody = [] {
+
+    ErrorHandlerImpl errorHandler;
+    TaskSet tasks(errorHandler);
+
+    for (int i = 0; i < stackSize / sizeof(void*); i++) {
+      tasks.add(kj::NEVER_DONE);
+    }
+  };
+
+#if KJ_FIBERS_AVAILABLE
+  EventLoop loop;
+  WaitScope waitScope(loop);
+
+  startFiber(stackSize,
+      [](WaitScope&) mutable {
+    testBody();
+  }).wait(waitScope);
+
+#else
+  pthread_attr_t attr;
+  KJ_REQUIRE(0 == pthread_attr_init(&attr));
+  KJ_DEFER(KJ_REQUIRE(0 == pthread_attr_destroy(&attr)));
+
+  KJ_REQUIRE(0 == pthread_attr_setstacksize(&attr, stackSize));
+  pthread_t thread;
+  KJ_REQUIRE(0 == pthread_create(&thread, &attr, [](void*) -> void* {
+    EventLoop loop;
+    WaitScope waitScope(loop);
+    testBody();
+    return nullptr;
+  }, nullptr));
+  KJ_REQUIRE(0 == pthread_join(thread, nullptr));
+#endif
+}
+
 TEST(Async, TaskSet) {
   EventLoop loop;
   WaitScope waitScope(loop);
@@ -946,7 +998,7 @@ KJ_TEST("exclusiveJoin both events complete simultaneously") {
   KJ_EXPECT(!joined.poll(waitScope));
 }
 
-#if !__BIONIC__ && !KJ_NO_EXCEPTIONS
+#if KJ_FIBERS_AVAILABLE && !KJ_NO_EXCEPTIONS
 KJ_TEST("start a fiber") {
   EventLoop loop;
   WaitScope waitScope(loop);
--- a/c++/src/kj/async.c++
+++ b/c++/src/kj/async.c++
@@ -278,6 +278,16 @@ public:
     node->onReady(this);
   }
 
+  Own<Task> pop() {
+    KJ_IF_MAYBE(n, next) { n->get()->prev = prev; }
+    Own<Task> self = kj::mv(KJ_ASSERT_NONNULL(*prev));
+    KJ_ASSERT(self.get() == this);
+    *prev = kj::mv(next);
+    next = nullptr;
+    prev = nullptr;
+    return self;
+  }
+
   Maybe<Own<Task>> next;
   Maybe<Own<Task>>* prev = nullptr;
 
@@ -307,14 +317,7 @@ protected:
     }
 
     // Remove from the task list.
-    KJ_IF_MAYBE(n, next) {
-      n->get()->prev = prev;
-    }
-    Own<Event> self = kj::mv(KJ_ASSERT_NONNULL(*prev));
-    KJ_ASSERT(self.get() == this);
-    *prev = kj::mv(next);
-    next = nullptr;
-    prev = nullptr;
+    auto self = pop();
 
     KJ_IF_MAYBE(f, taskSet.emptyFulfiller) {
       if (taskSet.tasks == nullptr) {
--- a/c++/src/kj/async.c++
+++ b/c++/src/kj/async.c++
@@ -259,17 +259,6 @@ void Canceler::AdapterImpl<void>::cancel(kj::Exception&& e) {
 
 TaskSet::TaskSet(TaskSet::ErrorHandler& errorHandler)
   : errorHandler(errorHandler) {}
-
-TaskSet::~TaskSet() noexcept(false) {
-  // You could argue it is dubious, but some applications would like for the destructor of a
-  // task to be able to schedule new tasks. So when we cancel our tasks... we might find new
-  // tasks added! We'll have to repeatedly cancel.
-  while (tasks != nullptr) {
-    auto toDestroy = kj::mv(tasks);
-    tasks = nullptr;
-  }
-}
-
 class TaskSet::Task final: public _::Event {
 public:
   Task(TaskSet& taskSet, Own<_::PromiseNode>&& nodeParam)
@@ -339,6 +328,16 @@ private:
   TaskSet& taskSet;
   Own<_::PromiseNode> node;
 };
+
+TaskSet::~TaskSet() noexcept(false) {
+  // You could argue it is dubious, but some applications would like for the destructor of a
+  // task to be able to schedule new tasks. So when we cancel our tasks... we might find new
+  // tasks added! We'll have to repeatedly cancel. Additionally, we need to make sure that we destroy
+  // the items in a loop to prevent any issues with stack overflow.
+  while (tasks != nullptr) {
+    auto removed = KJ_REQUIRE_NONNULL(tasks)->pop();
+  }
+}
 
 void TaskSet::add(Promise<void>&& promise) {
   auto task = heap<Task>(*this, _::PromiseNode::from(kj::mv(promise)));


c++/src/kj/common.h
```diff
@@ -347,6 +347,14 @@ KJ_NORETURN(void unreachable());
 #define KJ_FALLTHROUGH
 #endif
 
+#if KJ_HAS_CPP_ATTRIBUTE(nodiscard)
+#define KJ_MUST_USE_RESULT [[nodiscard]]
```
kentonv, line 351: Is this different from `KJ_WARN_UNUSED_RESULT`?

c++/src/kj/async.c++
```diff
@@ -278,6 +267,16 @@ public:
     node->onReady(this);
   }
 
+  KJ_MUST_USE_RESULT Own<Task> pop() {
```
kentonv, line 270: Why `KJ_MUST_USE_RESULT`? Technically the call site in `~TaskSet()` would work just fine even if it _didn't_ store the result to a local, wouldn't it?

c++/src/kj/async-test.c++
```diff
@@ -733,6 +739,41 @@ TEST(Async, TaskSet) {
...
+    ErrorHandlerImpl errorHandler;
+    TaskSet tasks(errorHandler);
+
+    for (int i = 0; i < stackSize / sizeof(void*); i++) {
+      tasks.add(kj::NEVER_DONE);
+    }
+  };
+
+#if _WIN32
+  HANDLE thread = CreateThread(nullptr, stackSize, [](void*) -> DWORD {
```
kentonv, line 757: This is fine (assuming you can get CI to pass), but FWIW, another option might have been to use fibers. The fibers API lets you specify the stack size. (However you'd have to skip the test on platforms that don't support fibers.)

kentonv:
FWIW I've had success testing Windows changes using MinGW + Wine on Linux.

On Debian I install mingw-w64, wine64, and wine-binfmt (so that executing a Windows exe on the command line automatically invokes wine). Then I do:

```
./configure --host=x86_64-w64-mingw32 --disable-shared CXXFLAGS='-static-libgcc -static-libstdc++ -g'
```

c++/src/kj/async.c++
vlovich: Yeah. I added this when I needed the result but I found a way to simplify. I'll remove it.

c++/src/kj/common.h
vlovich: No, it's just the C++17 way of doing things.

c++/src/kj/async-test.c++
vlovich: Is there a macro for that? Skimming the code I don't see which platforms fibers are unavailable on.

c++/src/kj/async.c++
```diff
@@ -337,6 +329,16 @@ private:
   Own<_::PromiseNode> node;
 };
 
+TaskSet::~TaskSet() noexcept(false) {
+  // You could argue it is dubious, but some applications would like for the destructor of a
+  // task to be able to schedule new tasks. So when we cancel our tasks... we might find new
+  // tasks added! We'll have to repeatedly cancel. Additionally, we need to make sure that we destroy
+  // the items in a loop to prevent any issues with stack overflow.
+  while (tasks != nullptr) {
+    auto removed = KJ_REQUIRE_NONNULL(tasks)->pop();
```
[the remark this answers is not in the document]
kentonv: I think the ordering has always been FIFO. In `TaskSet::Task`, the field `node` is the one that holds the promise being executed. It appears last in the class, so it would be the first thing to be destroyed in the destructor. Meanwhile, `next` is the link to the next member of the list. It's the first thing in the class, so it would be destroyed last. Hence, when the head of the list is destroyed, it'll cancel its own promise _before_ destroying the next list element, and hence the cancellations occur in list order, not reverse.

Anyway if it were actually LIFO and being switch to FIFO, that's fine too.

c++/src/kj/common.h
kentonv: Looks like you don't need it anyway.

c++/src/kj/async-test.c++
kentonv: At present they are only missing on `__BIONIC__`, though #1208 suggests OpenBSD needs to be added, and I proposed there that we create a `KJ_USE_FIBERS` macro to encapsulate this and let people override their platform defaults... but I wouldn't do that in this PR obviously.

c++/src/kj/async.c++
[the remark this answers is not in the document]
vlovich: Good point. It does still change the ordering slightly because of the batched destruction logic, so now earlier cancellations can preempt later ones enqueuing additional tasks. In other words, if we had 999 tasks and cancelling the first one caused the 1000th to be enqueued, in the old model we’d cancel all 999 first before cancelling the 1000th whereas now we’ll cancel 1, 1000, 2, … 999.


Anyway, I didn’t think that would be a problem so good to double check. And maybe it’s even better that if we’re shutting down the task set that we just reject the promise during the attempt to add it rather than allocating some space for it on the list just so that we can do some pointer arithmetic.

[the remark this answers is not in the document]
kentonv: Yeah I think that's fine.

c++/src/kj/async.c++
```diff
@@ -278,6 +267,16 @@ public:
     node->onReady(this);
   }
 
+  Own<Task> pop() {
+    KJ_IF_MAYBE(n, next) { n->get()->prev = prev; }
+    Own<Task> self = kj::mv(KJ_ASSERT_NONNULL(*prev));
+    KJ_ASSERT(self.get() == this);
+    *prev = kj::mv(next);
+    next = nullptr;
```
vlovich, line 275: Oh yeah, I wanted to ask about this line. It's superfluous right, because we just moved `next` and `Maybe` nullifies after a move I thought. Maybe just a `KJ_DASSERT(next == nullptr)`?

c++/src/kj/async.c++
```diff
@@ -370,7 +379,7 @@ Promise<void> TaskSet::onEmpty() {
     }
   }
 
-  if (tasks == nullptr) {
+  if (shuttingDown) {
```
vlovich, line 382: I noticed this oversight via manual inspection. Added a test to protect against regressing that in the future.

c++/src/kj/async-test.c++
vlovich: I added `KJ_USE_FIBERS` to the top of the file so that it's there whenever `KJ_USE_FIBERS` is implemented in the codebase properly. It was a little unclear if fibers required exceptions but I assumed that that protection is in place in the test for a different reason.

c++/src/kj/async.c++
```diff
@@ -337,7 +329,24 @@ private:
...
+  // complicates pop() not working since tasks is required to point to the head of the list for that
+  // method to work.
+  shuttingDown = true;
+  while (tasks != nullptr) {
+    auto removed = KJ_REQUIRE_NONNULL(tasks)->pop();
+  }
+}
+
 void TaskSet::add(Promise<void>&& promise) {
+  if (shuttingDown) {
```
vlovich, line 344: Minor behavioural change since last review but since I was adding `shuttingDown` anyway so that `onEmpty` behaviour was correct.

c++/src/kj/async.c++
kentonv: What oversight?

This change looks wrong. If you call `onEmpty()` and the task set is already empty, then the promise should complete immediately. That's what the previous code implemented. The new code only returns READY_NOW if the TaskSet is currently being destroyed. If it's empty, but not being destroyed, the new code will incorrectly cause the caller to wait.

c++/src/kj/async.c++
kentonv: This change isn't clearly correct, because it means the promise will be destroyed synchronously, whereas previously it was destroyed asynchronously. What if the destruction takes a lock, but the caller already holds that lock? I would be more comfortable leaving it the way it was.

c++/src/kj/async.c++
kentonv: I think the way it was previously was fine.

1. I doubt anyone calls `onEmpty()` on a `TaskSet` from within its own destructor.
2. If they _were_ doing so before, the behavior already wasn't guaranteed -- `onEmpty()` would only return READY_NOW if nothing else happened to add a task to the TaskSet just prior, during other promise destructors.
3. It we wanted to get pedantic, if `onEmpty()` were to be called as a result of cancelling promises in `~TaskSet()`, it really ought to _never_ return READY_NOW, because the TaskSet isn't actually empty until all the promises have been canceled. So if anything, the test should be `tasks == nullptr && !shuttingDown`. However, I don't think it actually matters.

c++/src/kj/async.c++
kentonv: Historically, a moved-from `Maybe` was not guaranteed to be null. We recently decided to make it so, after seeing one or two bugs go by that would have been prevented by this. But I'm not sure how I feel about people relying on it.

I'd just leave the code as it is. The compiler should be able to optimize out the redundant nullification anyway (not that it would matter if it didn't).

c++/src/kj/async-test.c++
```diff
@@ -733,6 +746,45 @@ TEST(Async, TaskSet) {
   EXPECT_EQ(1u, errorHandler.exceptionCount);
 }
 
+TEST(Async, LargeTaskSetDestruction) {
```
kentonv, line 772: Minor point, but probably commits should be ordered so that the test for a bug is added _after_ the bug is fixed, otherwise there's a point in the commit history where the tests are broken, which may interfere with things like `git bisect`.

kentonv (changes requested):
The first three commits look good. The last two commits should be removed.

c++/src/kj/async-test.c++
vlovich: Ok, I'll re-order. I usually prefer having the broken commit first so that you know that you really fixed the issue, but I can see the annoying impact it might have on git bisect.

c++/src/kj/async.c++
vlovich: Hmm.... I was just going from the logic from before

```
  while (tasks != nullptr) {
    auto toDestroy = kj::mv(tasks);
    tasks = nullptr;
  }
```

In this case, all the tasks that get destroyed see `onEmpty` return `READY_NOW`. While I agree with you on point #2, I'm worried about Hyrum's law in this case even for Cloudflare's use cases. Specifically, if two tasks enqueued were accidentally using `onEmpty` in their destruction, they would have been fine previously but now they'd be throwing an exception because the first would create a promise fulfiller pair & the second would throw an exception.

Beyond just the exception, it also changes the logic of cancellation. Before, during destruction, the `onEmpty` would get satisfied & any logic triggered as a result would run. If we use `tasks == nullptr && !shuttingDown` then we'll create the `emptyFulfiller` but it'll never get triggered.

Admittedly it's probably a corner case, but I'd rather separate bug fixes & behavioral changes when possible.

c++/src/kj/async.c++
vlovich: Added a comment so that a future reader might think twice about this optimization as well.

c++/src/kj/async.c++
kentonv: Again, the old behavior was actually incorrect, and anyone relying on it would presumably be fundamentally broken. Changing them to throw an exception would probably be a good thing, to uncover these bugs.

But also, no one is relying on it. ​It doesn't make sense to call `onEmpty()` on a task set that you don't own, and if you do own the task set, then you wouldn't be calling this after the task set is already being destroyed.

To make extra sure, I went ahead and looked at all call sites of this method. As expected, no one is calling it in a place where the TaskSet could possibly be in the process of being destroyed.

So this commit is trying to maintain buggy behavior that no one actually relies on. Let's just remove it.

c++/src/kj/async-test.c++
```diff
@@ -25,6 +25,19 @@
 #include "mutex.h"
 #include "thread.h"
 
+// TODO(later): Refactor this more cleanly so that the KJ library itself defines this.
+#if !__BIONIC__ && !__OpenBSD__
```
kentonv, line 29: We haven't actually fixed `async.c++` yet to recognize that OpenBSD doesn't have fibers, so this is inconsistent with the check being performed there...

kentonv (approved):

vlovich:
Hmm.... flaky test apparently:

```
[ TEST ] kj/mutex-test.c++:391: legacy test: Mutex/WhenWithTimeoutPreciseTiming
Failed to start process notifications for pid 15415 (19)
kj/mutex-test.c++:419: failed: time not within expected bounds even after retries
[ FAIL ] kj/mutex-test.c++:391: legacy test: Mutex/WhenWithTimeoutPreciseTiming (3926733 μs)
stack: 107a0fe1d 108a7badc
    kj::(anonymous namespace)::TestCase391::run() (in capnp-test) + 445: returning here
```

Flaked on [MacOS](https://github.com/capnproto/capnproto/pull/1289/checks?check_run_id=3222844743#step:4:1142)

vlovich:
On Arch I had to also install `mingw-w64-winpthreads` and run with `WINEPATH="/usr/lib/gcc/i686-w64-mingw32/11.1.0/;/usr/lib/gcc/x86_64-w64-mingw32/11.1.0;/usr/x86_64-w64-mingw32/bin;/usr/i686-w64-mingw32/bin"`

PR #1290, opened 2021-08-03
Fix promise chain stack overflow

vlovich:
This closes the ability for tail call optimization failures (on purpose or accidental) to generate
a stack overflow on destruction. KJ can now be built with a `KJ_CHAIN_PROMISE_DEPTH_WARNING_THRESHOLD`
to generate a warning whenever the promise chain is too long (similar to `KJ_CONTENTION_WARNING_THRESHOLD`).

@kentonv this wasn't actually that much work & I think this makes things a bit easier to debug. While Sentry does have contextual
information, it requires quite a bit of knowledge about the application using KJ (& sometimes both the application & KJ) to
convert it to a reasonable guess. I know your analysis was straightforward for you, but I have trouble following it & I suspect
other new people will struggle to perform the same analysis you did. However, now, if the warning threshold is set, we'll get a
precise stack trace of the issue & can even set this fairly aggressively within Workers since I suspect anything above a
relatively small number like is going to be an issue we want to examine on our end.

### DIFF
--- a/c++/src/kj/async.c++
+++ b/c++/src/kj/async.c++
@@ -145,6 +145,7 @@ public:
   void tracePromise(_::TraceBuilder& builder, bool stopAtNextEvent) override {
     builder.add(reinterpret_cast<void*>(&kj::evalLater<DummyFunctor>));
   }
+  void cancelDependenciesInto(Vector<Own<PromiseNode>>&) override final {}
 };
 
 class YieldHarderPromiseNode final: public _::PromiseNode {
@@ -158,6 +159,7 @@ public:
   void tracePromise(_::TraceBuilder& builder, bool stopAtNextEvent) override {
     builder.add(reinterpret_cast<void*>(&kj::evalLast<DummyFunctor>));
   }
+  void cancelDependenciesInto(Vector<Own<PromiseNode>>&) override final {}
 };
 
 class NeverDonePromiseNode final: public _::PromiseNode {
@@ -171,6 +173,7 @@ public:
   void tracePromise(_::TraceBuilder& builder, bool stopAtNextEvent) override {
     builder.add(_::getMethodStartAddress(kj::NEVER_DONE, &_::NeverDone::wait));
   }
+  void cancelDependenciesInto(Vector<Own<PromiseNode>>&) override final {}
 };
 
 }  // namespace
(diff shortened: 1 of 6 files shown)


vlovich:
Setting up a Windows VM with VS to debug the problem. Cross MingW fails in a different way.

c++/src/kj/async.c++
```diff
@@ -2346,7 +2410,15 @@ ChainPromiseNode::ChainPromiseNode(Own<PromiseNode> innerParam)
   inner->onReady(this);
 }
 
-ChainPromiseNode::~ChainPromiseNode() noexcept(false) {}
+ChainPromiseNode::~ChainPromiseNode() noexcept(false) {
+  cancelDependenciesWithoutStackOverflow();
+}
+
+void ChainPromiseNode::cancelDependenciesInto(Vector<Own<PromiseNode>> &toCancel) {
```
vlovich, line 2417: This logic is broken but for some reason only Windows has coverage. Investigating.

vlovich: Nope. Found it. It's in the `cancelDependenciesWithoutStackOverflow` implementation.

kentonv:
This adds a fair amount of bookkeeping -- including allocations (for the Vector) -- inside the destructor of most PromiseNode implementations, which KJ async code creates lots and lots of. Have you measured the performance impact?

vlovich:
I have not. Should I profile it via measuring RPS of a load test or is there a different benchmark I should run?

vlovich:
The other approach that I can take is a hybrid one. Either:
A) Modify Vector to be constructable from an ArrayBuilder that's initially wrapping some space on the stack.
B) Unwind the promises via the stack up to some limit & then switch to this heap approach once the limit is exceeded.

Option A feels like the simplest. Aside from the allocations (which only happen if you have a dependency chain), is there any other book keeping you're seeing as expensive? Most of this seems like not much more than what the stack is going to do when it unwinds...

vlovich:
FWIW, I bumped the `LargeInclusiveJoinPromisesChain` test to an iteration count of 102400 & made the stack size infinite.

Debug without stack overflow protection the test takes ~360.635ms
Debug with stack overflow protection takes ~265.406ms

Release build without stack overflow protection takes: ~104.905 ms
Release build with stack overflow protection takes: ~70.617 ms

This is obviously skewed to just one test so a more complex/real example might show different behavior, but there's at least some indication that this is counterintuitively a performance improvement. I think the reason is that there's actually less book keeping happening than when the stack is used because here you're manipulating roughly a pointer per promise, maybe a couple of pointers. In the stack unwinding you have to push the entire frame resulting in quite a bit of pointer-sized registers being spilled (and backtracking).

Interestingly, if I give the Vector an initial stack buffer to use it's actually slower, not sure why.

vlovich:
Using Worker's load test on my laptop, I get a consistent RPS of ~3182 with this change (optimized build, cycle random 50, repeat 10s). Without the change it's ~-3290~3423 so maybe a regression of 3%-7%? Obviously workload dependent.

kentonv:
3%-7% is a pretty severe regression. I don't think this is worth it.

vlovich:
Yeah. I plan to rewrite this to switch to the slow path once the promise stack depth exceeds a certain length. That should remove the penalty for the common path while ensuring it's not possible to construct a stack overflow.

vlovich:
I've run out of ideas on how this could be improved in KJ.

vlovich:
I did try out making a slow vs fast path. It still showed a performance regression for some reason & I'm not sure why. I wouldn't think calling a virtual method to do the unwind manually is slower than what the compiler generates for the destructors it walks during unwind...

vlovich:
Looks like pushing to a branch after the PR is closed doesn't update the diff. My last attempt at using regular unwind unless there's > 500 promises in a chain is https://github.com/vlovich/capnproto/tree/fix-promise-chain-stack-overflow.

PR #1293, opened 2021-08-06
Log the time spent contending the lock

vlovich:
Maybe we should also include the time spent holding the exclusive lock
(or at least in KJ_DEBUG?), but I suspect this is sufficient.

### DIFF
--- a/c++/src/kj/mutex.c++
+++ b/c++/src/kj/mutex.c++
@@ -263,6 +263,10 @@ bool Mutex::lock(Exclusivity exclusivity, Maybe<Duration> timeout, LockSourceLoc
 #endif
       break;
     case SHARED: {
+#if KJ_CONTENTION_WARNING_THRESHOLD
+      kj::Maybe<kj::TimePoint> contentionWaitStart;
+#endif
+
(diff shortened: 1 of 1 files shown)


c++/src/kj/mutex.c++
```diff
@@ -263,6 +263,13 @@ bool Mutex::lock(Exclusivity exclusivity, Maybe<Duration> timeout, LockSourceLoc
 #endif
       break;
     case SHARED: {
+#if KJ_CONTENTION_WARNING_THRESHOLD
+      // We could have the exclusive mutex tell us how long it was holding the lock. That would be
+      // the nicest. However, I'm hesitant to bloat the structure. I suspect having a reader tell
+      // us how long it was waiting for is probably a good proxy.
+      auto contentionWaitStart = kj::systemPreciseMonotonicClock().now();
```
kentonv, line 270: Can you move this down so that in the common, uncontended case, we don't read the clock at all?

vlovich: It complicates the logic slightly but seems reasonable. Let me know what you think.

PR #1294, opened 2021-08-07
Fix large taskset test when building with KJ_NO_EXCEPTIONS

jclee:
Looks like `-fno-exceptions` builds have been failing in `Async/LargeTaskSetDestruction` since b6e26ae was merged, because FiberStack's constructor fails when exceptions are turned off.


kentonv:
Thanks!

PR #1295, opened 2021-08-07
Fix most 0.9 milestone issues

kentonv:


c++/src/kj/string.c++
```diff
@@ -489,6 +461,76 @@ char* FloatToBuffer(float value, char* buffer) {
...
+
+double NoLocaleStrtod(const char* text, char** original_endptr) {
+  // We cannot simply set the locale to "C" temporarily with setlocale()
+  // as this is not thread-safe.  Instead, we try to parse in the current
+  // locale first.  If parsing stops at a '.' character, then this is a
+  // pretty good hint that we're actually in some other locale in which
+  // '.' is not the radix character.
+
+  char* temp_endptr;
+  double result = strtod(text, &temp_endptr);
```
vlovich, line 503: Hmm.... what if we leverage https://github.com/lemire/fast_double_parser? Then this parsing would be ~16x faster (8x faster than strtod & you call strtod twice, + there's a sprintf in there) & simpler to reason about, although I'm sure it's correct since you're just copying your old protobuf implementation. It's also only a single header file so there's not even any build files that would change...

vlovich (approved):

c++/src/kj/string.c++
vlovich: Also, even if we don't take it, reading the source there, I like it's locale solution much better than what we have here using strtod_l for most systems except cygwin & solaris where it has a different solution. Unclear which is faster in that path - I suspect they still win because a small malloc seems faster than doing 2*strtod + sprintf.

vlovich: Oh, and reading the code, it looks like they technically do have a memory leak on exit that valgrind/asan might complain about (would need to change the code so that the locale is freed on shutdown)

c++/src/kj/filesystem-disk-unix.c++
```diff
@@ -1102,7 +1102,10 @@ public:
 
       KJ_SYSCALL_HANDLE_ERRORS(syscall(SYS_renameat2,
           fromDirFd, fromPath.cStr(), fd.get(), toPath.cStr(), RENAME_EXCHANGE)) {
-        case ENOSYS:
+        case ENOSYS:  // Syscall not supported by kernel.
+        case EINVAL:  // Maybe we screwed up, or maybe the syscall is not supported by the
+                      // filesystem. Unfortunately, there's no way to tell, so assume the latter.
+                      // ZFS in particular apparently produces EINVAL.
```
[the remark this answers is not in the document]
kentonv: Agreed 100%.

c++/src/kj/string.c++
kentonv: The code I copied from protobuf is code I wrote. :)

But I wrote it ~15 years ago. Absurdly enough, there have been advances in both floating-point parsing and stringification since then. I think it would make a lot of sense to investigate the state of the art and possibly bring in some better implementations here. But, my goal for this PR was to fix the bugs as quickly as possible in order to do a release... we can investigate these options another time.

c++/src/capnp/compiler/capnp.c++
```diff
@@ -1287,8 +1287,13 @@ private:
       return IMPOSSIBLE;
     }
     if ((prefix[3] & 0x80) != 0) {
-      // Offset is negative (invalid).
-      return IMPOSSIBLE;
+      if (prefix[0] == 0xff && prefix[1] == 0xff && prefix[2] == 0xff && prefix[3] == 0xff &&
```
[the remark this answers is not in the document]
kentonv: "Should", yes. But TBH I don't want to spend time on it.

PR #1296, opened 2021-08-08
Fixes #1288.

### LINKED ISSUE #1288: Windows from-zip installation instructions omit stream.capnp
The [Windows installation instructions](https://capnproto.org/install.html#installation-windows) say:

> From Release Zip
>    1. Download Cap’n Proto Win32 build: https://capnproto.org/capnproto-c++-win32-0.8.0.zip
>    2. Find capnp.exe, capnpc-c++.exe, and capnpc-capnp.exe under capnproto-tools-win32-0.8.0 in the zip and copy them somewhere.
> 
> If you don’t care about C++ support, you can stop here. The compiler exe can be used with plugins provided by projects implementing Cap’n Proto in other languages.

These instructions fail to mention that if you want support for the `stream` keyword (and even if you don't care about C++), then you will also need `stream.capnp` (and maybe `c++.capnp`, which is referenced by it).

See https://github.com/capnproto/capnproto-rust/issues/230 .

kentonv:


kentonv:
Doc-only change. Merging without review so that I can proceed with a release.

PR #1302, opened 2021-08-11
Add address to connect() error message

jasnell:
To make the "connect()" error message a bit more useful, add the address that was being attempted.

/cc @kentonv @vlovich 

Signed-off-by: James M Snell <jasnell@gmail.com>

### DIFF
--- a/c++/src/kj/async-io-unix.c++
+++ b/c++/src/kj/async-io-unix.c++
@@ -1286,7 +1286,8 @@ public:
           // Fine.
           break;
         } else if (error != EINTR) {
-          KJ_FAIL_SYSCALL("connect()", error) { break; }
+          auto address = SocketAddress(addr, addrlen).toString();
+          KJ_FAIL_SYSCALL("connect()", error, address) { break; }
           return Own<AsyncIoStream>();
         }
       } else {


c++/src/kj/async-io-unix.c++
```diff
@@ -1286,7 +1293,8 @@ public:
           // Fine.
           break;
         } else if (error != EINTR) {
-          KJ_FAIL_SYSCALL("connect()", error) { break; }
+          auto address = SocketAddress::toString(addr, addrlen);
```
kentonv, line 1296: Hmm why the big refactoring of `toString()` into a static method instead of doing:

```
auto address = SocketAddress(addr, addrlen).toString();
```

jasnell: Just a style choice really, avoiding the additional stack allocation. But if the `SocketAddress(addr, addrlen)...` approach is what you prefer then I'm good with that.

kentonv: Well, since it's only called when throwing an exception anyway, I think the copy isn't going to make much difference. I do prefer to minimize code churn when not necessary, so yeah, I think I'd prefer the smaller diff here.

jasnell: Sounds good!

Issue #1167, opened 2021-01-23
Failing to build on musl libc due to missing getcontext/setcontext


roshii:
I'm trying to build on Alpine Linux v3.13 from release tarball v0.8.0 but  failing due to missing commands listed here after.
From what i [understand](https://github.com/9fans/plan9port/issues/36) [`libucontext`](https://github.com/kaniini/libucontext) seem to implement these missing commands but i fail to make anything out of it.

```
/usr/lib/gcc/x86_64-alpine-linux-musl/9.3.0/../../../../x86_64-alpine-linux-musl/bin/ld: ./.libs/libkj-async.so: undefined reference to `swapcontext'
/usr/lib/gcc/x86_64-alpine-linux-musl/9.3.0/../../../../x86_64-alpine-linux-musl/bin/ld: ./.libs/libkj-async.so: undefined reference to `getcontext'
/usr/lib/gcc/x86_64-alpine-linux-musl/9.3.0/../../../../x86_64-alpine-linux-musl/bin/ld: ./.libs/libkj-async.so: undefined reference to `makecontext'
collect2: error: ld returned 1 exit status
make[2]: *** [Makefile:1838: capnp-afl-testcase] Error 1
```

kentonv:
Interesting, I guess musl doesn't implement these, maybe because POSIX considers them deprecated. But POSIX offers no replacement and this is the only portable way to implement fibers.

Some ways we could address this:
* Stopgap: You could maybe install `libucontext` and then do `./configure LIBS='-lucontext'` to force Cap'n Proto to link against it.
* We could extend the source code in `kj/async.c++` to detect musl via `#ifdef`, similar to how it detects Bionic, and disable fibers in this case.
* We could add a new build flag to disable fibers, so that the next person who discovers their libc implementation doesn't support these functions, they can disable fibers and move on.
* We could actually adopt libucontext into KJ so that we don't depend on it being provided by the system at all. This would give us more portability and protect us against future libc versions dropping support for these. We might even be able to optimize them a bit for our specific use case. But this option is probably a fair amount of work.

Happy to accept PRs.

roshii:
Running `./configure LIBS='-lucontext'` turned out to be successful, many thanks :)

imho, a build flag would be the most appropriate as it would give musl users a choice to user fibers or not. i have little experience in c++ but i like challenges and i may try to implement this flag within the year.

p.s. i'm not closing the issue as i'm not sure how you want to keep track of things. feel free to do so however.

[...]

PR #1313, opened 2021-08-20
Add support for OpenBSD

ZhanYF:
It's been on my to-do list for quite some time now, finally got a chance to do it. This PR aims to:

  - fix test for systems with bsd cmp
    - a single dash is required for bsd style cmp to accept input from stdin
  - Allow disable fibers with KJ_USE_FIBERS
    - as discussed in https://github.com/capnproto/capnproto/issues/1208 and https://github.com/capnproto/capnproto/issues/1167
  - Use appropriate ldconfig command for OpenBSD 

Tested on OpenBSD 7.0-beta/AMD64.

output of `make check`: https://github.com/ZhanYF/static/blob/main/checkLog
  - 1 of 3 tests failed, but I guess that's due to not building fibers?

any comment?  I have to admit I'm not familiar with capnproto/fibers/c++ in general, so please feel free to tweak it :)

### DIFF
--- a/c++/Makefile.am
+++ b/c++/Makefile.am
@@ -421,10 +421,15 @@ capnpc_c___SOURCES = src/capnp/compiler/capnpc-c++.c++
 # Also attempt to run ldconfig, because otherwise users get confused.  If
 # it fails (e.g. because the platform doesn't have it, or because the
 # user doesn't have root privileges), don't worry about it.
+#
+# We need to specify the path for OpenBSD.
 install-exec-hook:
 	ln -sf capnp $(DESTDIR)$(bindir)/capnpc
-	ldconfig < /dev/null > /dev/null 2>&1 || true
-
+	if [ `uname` == 'OpenBSD' ]; then \
+	        (ldconfig /usr/local/lib /usr/lib /usr/X11R6/lib > /dev/null 2>&1 || true); \
+	else \
+		ldconfig < /dev/null > /dev/null 2>&1 || true; \
+	fi
 uninstall-hook:
 	rm -f $(DESTDIR)$(bindir)/capnpc
 
--- a/c++/src/capnp/compiler/capnp-test.sh
+++ b/c++/src/capnp/compiler/capnp-test.sh
@@ -77,7 +77,7 @@ $CAPNP convert json:binary $SCHEMA TestAllTypes < $TESTDATA/pretty.json | cmp $T
 $CAPNP convert json:binary $SCHEMA TestAllTypes < $TESTDATA/short.json | cmp $TESTDATA/binary - || fail short json to binary
 
 $CAPNP convert json:binary $JSON_SCHEMA TestJsonAnnotations -I"$SRCDIR" < $TESTDATA/annotated.json | cmp $TESTDATA/annotated-json.binary - || fail annotated json to binary
-$CAPNP convert binary:json $JSON_SCHEMA TestJsonAnnotations -I"$SRCDIR" < $TESTDATA/annotated-json.binary | cmp $TESTDATA/annotated.json - || fail annotated json to binary
+$CAPNP convert binary:json $JSON_SCHEMA TestJsonAnnotations -I"$SRCDIR" < $TESTDATA/annotated-json.binary | cmp $TESTDATA/annotated.json - || fail annotated binary to json
 
 [ "$(echo '(foo = (text = "abc"))' | $CAPNP convert text:text "$SRCDIR/capnp/test.capnp" BrandedAlias)" = '(foo = (text = "abc"), uv = void)' ]  || fail branded alias
 [ "$(echo '(foo = (text = "abc"))' | $CAPNP convert text:text "$SRCDIR/capnp/test.capnp" BrandedAlias.Inner)" = '(foo = (text = "abc"))' ]  || fail branded alias
--- a/c++/src/kj/async-io-unix.c++
+++ b/c++/src/kj/async-io-unix.c++
@@ -1480,7 +1480,9 @@ kj::Own<PeerIdentity> SocketAddress::getIdentity(kj::LowLevelAsyncIoProvider& ll
       // seen vague references on the internet saying that a PID of 0 and a UID of uid_t(-1) are used
       // as invalid values.
 
-#if defined(SO_PEERCRED)
+// OpenBSD defines SO_PEERCRED but uses a different interface for it
+// hence we're falling back to LOCAL_PEERCRED
+#if defined(SO_PEERCRED) && !__OpenBSD__
       struct ucred creds;
       uint length = sizeof(creds);
       stream.getsockopt(SOL_SOCKET, SO_PEERCRED, &creds, &length);
@@ -1492,7 +1494,7 @@ kj::Own<PeerIdentity> SocketAddress::getIdentity(kj::LowLevelAsyncIoProvider& ll
       }
 
 #elif defined(LOCAL_PEERCRED)
-      // MacOS / FreeBSD
+      // MacOS / FreeBSD / OpenBSD
       struct xucred creds;
       uint length = sizeof(creds);
       stream.getsockopt(SOL_LOCAL, LOCAL_PEERCRED, &creds, &length);
--- a/c++/src/kj/async.c++
+++ b/c++/src/kj/async.c++
@@ -53,8 +53,12 @@
 #include <windows.h>  // for Sleep(0) and fibers
 #include "windows-sanity.h"
 #else
+
+#if KJ_USE_FIBERS
 #include <ucontext.h>
 #include <setjmp.h>    // for fibers
+#endif
+
 #include <sys/mman.h>  // mmap(), for allocating new stacks
 #include <unistd.h>    // sysconf()
 #include <errno.h>


jclee:
FYI, I expect you'd want to amend this line in async-test as well, which might help with the failing test:
https://github.com/capnproto/capnproto/blob/4079a9c/c%2B%2B/src/kj/async-test.c%2B%2B#L29

c++/Makefile.am
```diff
@@ -421,10 +421,14 @@ capnpc_c___SOURCES = src/capnp/compiler/capnpc-c++.c++
 # Also attempt to run ldconfig, because otherwise users get confused.  If
 # it fails (e.g. because the platform doesn't have it, or because the
 # user doesn't have root privileges), don't worry about it.
+#
+# We need to specify the path for OpenBSD.
 install-exec-hook:
 	ln -sf capnp $(DESTDIR)$(bindir)/capnpc
 	ldconfig < /dev/null > /dev/null 2>&1 || true
```
kentonv, line 428: Do you intentionally want to run `ldconfig` bare once, and then a second time with paths, on OpenBSD? Or should this be in an "else" block below?

c++/src/kj/async-io-unix.c++
```diff
@@ -85,7 +85,7 @@ void setCloseOnExec(int fd) {
 }
 
 static constexpr uint NEW_FD_FLAGS =
-#if __linux__ && !__BIONIC__
+#if __linux__ && KJ_USE_FIBERS
```
kentonv, line 88: It looks like you assumed every instance of `!__BIONIC__` is guarding fiber code, but there are actually many reasons why we check for Bionic in different parts of the code. Most of these ifdefs are unrelated to fibers. This one here actually only matches Linux, so obviously it doesn't need to be updated to support OpenBSD. Can you remove the changes that aren't fiber-related?

In fact, nothing in this file is fiber-related.

c++/src/kj/async-io-unix.c++
```diff
@@ -1483,7 +1483,7 @@ kj::Own<PeerIdentity> SocketAddress::getIdentity(kj::LowLevelAsyncIoProvider& ll
         result.uid = creds.uid;
       }
 
-#elif defined(LOCAL_PEERCRED)
+#elif defined(LOCAL_PEERCRED) && __OpenBSD__
```
kentonv, line 1486: It looks like your change on this line will actually break platforms other than OpenBSD that use this block of code.

kentonv (changes requested):
Hi @ZhanYF,

Thanks for working on this!

c++/Makefile.am
ZhanYF: Technically an 'else' block is not required since the second `ldconfig` will overwrite the effect of the first `ldconfig`, but I see how an 'else' block can make things better. Updated.

c++/src/kj/async-io-unix.c++
ZhanYF: Thanks for letting me know that, I restricted the scope the KJ_USE_FIBERS and reverted the change done on this file. Please take a look :)

c++/src/kj/async-io-unix.c++
ZhanYF: This is a mistake and is now reverted, thanks!

ZhanYF:
> FYI, I expect you'd want to amend this line in async-test as well, which might help with the failing test:
> https://github.com/capnproto/capnproto/blob/4079a9c/c%2B%2B/src/kj/async-test.c%2B%2B#L29

This indeed helped, but there are still some failed tests which I have no idea how to fix, e.g.:
```
kj/mutex.c++:963: error: pthread_mutex_destroy(&waiter.stupidMutex); strerror(pthreadError) = Invalid argument
kj/mutex.c++:963: pthread_mutex_destroy(&waiter.stupidMutex); strerror(pthreadError) = Invalid argument
stack: 
```

log:
https://raw.githubusercontent.com/ZhanYF/static/main/checkLog.2

Any idea on what might be causing that?

vlovich (commented):
It might be a good idea to set up CI for OpenBSD (https://github.com/cross-platform-actions/openbsd-builder ?). Otherwise, the bitrot risk seems high.

Regarding the failed tests, is that for the version of code that's pushed or some version you're testing locally? I'm assuming the tests are being run on OpenBSD? I don't know why the tests would be failing on OpenBSD. It looks like it's using the pthread-based implementation & I'm not sure how well that branch is covered by automation. I think you'll have to dig in & debug what's going on if you want to figure that part out. If you're just using plain cap'n'proto messages, then mutexes don't really come up. They're needed if you're using KJ's event loop (& cap'n'proto RPC since it's built on top of that).

That being said, AFAICT, OpenBSD has futex as of 6.2. Not sure if that's the minimum version that can be targeted for your use-case, but, if so, it might be to adjust the macro at the top of `mutex.h` to define `KJ_USE_FUTEX` to 1 by default on OpenBSD & see if that bypasses the problem.

c++/Makefile.am
```diff
@@ -421,10 +421,15 @@ capnpc_c___SOURCES = src/capnp/compiler/capnpc-c++.c++
 # Also attempt to run ldconfig, because otherwise users get confused.  If
 # it fails (e.g. because the platform doesn't have it, or because the
 # user doesn't have root privileges), don't worry about it.
+#
+# We need to specify the path for OpenBSD.
 install-exec-hook:
 	ln -sf capnp $(DESTDIR)$(bindir)/capnpc
-	ldconfig < /dev/null > /dev/null 2>&1 || true
-
+	if [ `uname` == 'OpenBSD' ]; then \
```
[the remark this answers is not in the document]
kentonv: This is an automake file, so we need to assume bourne shell.

c++/src/kj/async-io-unix.c++
```diff
@@ -1472,7 +1472,7 @@ kj::Own<PeerIdentity> SocketAddress::getIdentity(kj::LowLevelAsyncIoProvider& ll
       // seen vague references on the internet saying that a PID of 0 and a UID of uid_t(-1) are used
       // as invalid values.
 
-#if defined(SO_PEERCRED)
+#if defined(SO_PEERCRED) && !__OpenBSD__
```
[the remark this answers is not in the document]
ZhanYF: I don't think so, this is intentional as OpenBSD requires `struct xucred creds` here. Reverting it will cause the following error:
```
src/kj/async-io-unix.c++:1476:20: error: variable has incomplete type 'struct ucred'
      struct ucred creds;
                   ^
src/kj/async-io-unix.c++:1476:14: note: forward declaration of 'ucred'
      struct ucred creds;
```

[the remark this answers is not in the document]
ZhanYF: It seems like I forget to push the fully reverted version last time, this file should be good now. Thanks for pointing that out.

c++/src/kj/async.c++
```diff
@@ -1443,7 +1452,7 @@ FiberStack::FiberStack(size_t stackSizeParam)
   // We can create fibers before we convert the main thread into a fiber in FiberBase
   KJ_WIN32(osFiber = CreateFiber(stackSize, &StartRoutine::run, this));
 
-#elif !__BIONIC__
+#elif !__BIONIC__ && KJ_USE_FIBERS
```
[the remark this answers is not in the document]
ZhanYF: It is indeed redundant here, forgot to change it last time, should be OK now.

ZhanYF:
> It might be a good idea to set up CI for OpenBSD (https://github.com/cross-platform-actions/openbsd-builder ?). Otherwise, the bitrot risk seems high.

I agree, it's just that the current approach used in cross-platform-actions is relying on macOS for nested virtualization, and since I don't have a mac, the debugging process has not been very smooth. My limited test results show that `super-test.sh` (and probably other scripts) will need to be tweaked to be compatible with bsd style tools. Perhaps I should open another PR/issue for that?

> That being said, AFAICT, OpenBSD has futex as of 6.2. Not sure if that's the minimum version that can be targeted for your use-case, but, if so, it might be to adjust the macro at the top of `mutex.h` to define `KJ_USE_FUTEX` to 1 by default on OpenBSD & see if that bypasses the problem.

The current futex syscall on OpenBSD only supports three operations `(FUTEX_WAIT/FUTEX_WAKE/FUTEX_REQUEUE)` and mutex.c++ needs `FUTEX_WAIT_BITSET_PRIVATE`, so I don't think I can get it to work now. However since I don't need it for my use case it's not really a road blocker for me.

c++/src/kj/async-io-unix.c++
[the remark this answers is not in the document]
kentonv: Can you add a comment here that OpenBSD defines `SO_PEERCRED` but uses a different interface for it, hence we're falling back to `LOCAL_PEERCRED`?

kentonv (changes requested):
Why did you remove a bunch of CI tests?

sobomax:
Just a quick note that I've integrated some __OpenBSD__ conditionals into KJ_USE_FIBERS in PR #1321

c++/src/kj/async-io-unix.c++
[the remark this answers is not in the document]
ZhanYF: Agree, done.

ZhanYF:
Sorry, the CI removal part was a mistake, it was meant to be for my WIP CI test branch. Reverted.

kentonv:
#1321 has been merged covering a lot of similar ground to your PR. Can you rebase and reduce this PR to just the OpenBSD specifics?

c++/src/capnp/compiler/capnp-test.sh
```diff
@@ -77,7 +77,7 @@ $CAPNP convert json:binary $SCHEMA TestAllTypes < $TESTDATA/pretty.json | cmp $T
 $CAPNP convert json:binary $SCHEMA TestAllTypes < $TESTDATA/short.json | cmp $TESTDATA/binary - || fail short json to binary
 
 $CAPNP convert json:binary $JSON_SCHEMA TestJsonAnnotations -I"$SRCDIR" < $TESTDATA/annotated.json | cmp $TESTDATA/annotated-json.binary - || fail annotated json to binary
-$CAPNP convert binary:json $JSON_SCHEMA TestJsonAnnotations -I"$SRCDIR" < $TESTDATA/annotated-json.binary | cmp $TESTDATA/annotated.json - || fail annotated json to binary
+$CAPNP convert binary:json $JSON_SCHEMA TestJsonAnnotations -I"$SRCDIR" < $TESTDATA/annotated-json.binary | cmp $TESTDATA/annotated.json - || fail annotated binary to json
```
ZhanYF, line 80: I understand this is not OpenBSD specifics but since we are here perhaps we can also fix this line

kentonv (approved):

kentonv:
Thanks!

kentonv:
Hi @ZhanYF,

I've put together a point release including your OpenBSD changes:

https://capnproto.org/capnproto-c++-0.9.1-rc1.tar.gz

Can you verify it works?

ZhanYF:
Thanks for that, I can confirm this works on the latest OpenBSD snapshot w/ amd64.

P.S.: I've got the GitHub CI pipeline working now, but tests for known-broken features generate a large amount of noise, should I patch those tests out in my pipeline?

kentonv:
@ZhanYF Hmm, what features are "known-broken"? We could `#ifdef`-out those tests on OpenBSD if needed, though obviously it would be better to fix the bugs if at all possible.

ZhanYF:
@kentonv 

Sorry for the delay, I'm busy at the moment so can't investigate all the errors.
Here is a list of failed tests:

* Everything has to do with mutex, because at the moment futex syscall on OpenBSD only supports three operations (FUTEX_WAIT/FUTEX_WAKE/FUTEX_REQUEUE) and mutex.c++ needs at least FUTEX_WAIT_BITSET_PRIVATE
  * This is unlikely to change in near future and generates the most noise, so we should probably `ifdef` them out

* Filesystem related things:
```
kj/filesystem-disk-test.c++:339: failed: expected kj::str(mapping.slice(0, 6).asChars()) == "quxbaz" [fDobaz == quxbaz]
kj/filesystem-disk-test.c++:340: failed: expected kj::str(writableMapping->get().slice(0, 6).asChars()) == "quxbaz" [fDobaz == quxbaz]
kj/filesystem-disk-test.c++:344: failed: expected kj::str(mapping.slice(12, 17).asChars()) == "corge" [^@^@^@^@^@ == corge]
kj/filesystem-disk-unix.c++:919: failed: create(path): Operation not permitted; path = .kj-tmp.81341.19.foo.partial
```

They seem to be related to kj, but I don't know how to debug that.

* TLS errors 
```
kj/compat/tls-test.c++:682: failed: expected ::kj::_::hasSubstring(e->getDescription(), message); exception description didn't contain expected substring; *e = kj/compat/tls.c++:162: failed: TLS peer's certificate is not trusted; reason = error number 62
kj/compat/tls-test.c++:732: failed: expected ::kj::_::hasSubstring(e->getDescription(), "alert"); exception description didn't contain expected substring; *e = kj/compat/tls.c++:66: failed: OpenSSL error; message = error:1404C45C:SSL routines:ST_OK:reason(1116)
```

This might be due to the fact that OpenBSD uses LibreSSL instead of OpenSSL, but I'm not sure.
[log.test.txt](https://github.com/capnproto/capnproto/files/7238085/log.test.txt)
[log.test.txt](https://github.com/capnproto/capnproto/files/7238090/log.test.txt)


* Xthread error 

```
[ FAIL ] kj/async-xthread-test.c++:473: cross-thread cancellation cycle (889 μs)
[ FAIL ] kj/async-xthread-test.c++:547: call own thread's executor (625 μs)
```
* Some legacy tests for async io 
```
[ TEST ] kj/async-io-test.c++:701: legacy test: AsyncIo/ScmRightsTruncatedEven
[ FAIL ] kj/async-io-test.c++:701: legacy test: AsyncIo/ScmRightsTruncatedEven (754 μs)
```

kentonv:
> Everything has to do with mutex, because at the moment futex syscall on OpenBSD only supports three operations

KJ only uses futex if `__linux__` is defined. OpenBSD should be falling back to the pthread-based mutex implementation. So I'm confused how you'd be seeing futex issues...

> Filesystem related things:

Yikes, looks like some sort of corruption when using mmap?

> TLS errors

Oh great, LibreSSL produces different error strings than either upstream OpenSSL or BoringSSL? I guess we'll need to adjust the test to accept any of them.

> Xthread error

Were there any error messages associated with these two failures?

> Some legacy tests for async io

To be clear the word "legacy" here refers to how the test was declared (using the googletest syntax vs. KJ syntax), but the test is still very much valid. It looks like OpenBSD may need to be added to the `#if __APPLE__ || __FreeBSD__` check in `async-io-unix.c++` that works around a bug in `SCM_RIGHTS` handling present on those systems.

ZhanYF:
Hi, sorry for the looong delay, I got way too busy recently...

-----

> KJ only uses futex if `__linux__` is defined. OpenBSD should be falling back to the pthread-based mutex implementation. So I'm confused how you'd be seeing futex issues...

You're right, sorry for the confusion.  This is not the default and I got those errors because I was testing with `KJ_USE_FUTEX` defined, due to the pthread-based implementation being broken. Unfortunately both the FUTEX implementation and the pthread-based one are broken. Here is the errors I got when using the pthread-based implementation:

kj/mutex.c++:963: pthread_mutex_destroy(&waiter.stupidMutex); strerror(pthreadError) = Invalid argument
stack: 
kj/mutex.c++:963: error: pthread_mutex_destroy(&waiter.stupidMutex); strerror(pthreadError) = Invalid argument
kj/mutex.c++:963: pthread_mutex_destroy(&waiter.stupidMutex); strerror(pthreadError) = Invalid argument
stack: 
kj/mutex.c++:963: pthread_mutex_destroy(&waiter.stupidMutex); strerror(pthreadError) = Invalid argument
stack: 
kj/mutex.c++:963: pthread_mutex_destroy(&waiter.stupidMutex); strerror(pthreadError) = Invalid argument
stack: 
[ FAIL ] kj/mutex-test.c++:262: legacy test: Mutex/WhenWithTimeout (279188 <CE><BC>s)

looks like the `pthread_mutex_destroy(&waiter.stupidMutex);` call is the source of our problem here, errors disappeared after commenting it out, I tried to debug it but with no success, will come back to that later.

-----

> Yikes, looks like some sort of corruption when using mmap?

Yes, part of the string is corrupted. I tried to debug it but that rabbit hole is too deep...

----

> Oh great, LibreSSL produces different error strings than either upstream OpenSSL or BoringSSL? I guess we'll need to adjust the test to accept any of them.

agree, after adding those strings to the test the errors went away.

----

>>    Xthread error

> Were there any error messages associated with these two failures?

The Xthread related failure is also caused by `pthread_mutex_destroy` not working properly and the messages are the same.

----

> It looks like OpenBSD may need to be added to the `#if __APPLE__ || __FreeBSD__` check in async-io-unix.c++ that works around a bug in SCM_RIGHTS handling present on those systems.

Thanks for the pointer! It does seem to be the case and the this fixes the following failure:
`kj/async-io-test.c++:701: legacy test: AsyncIo/ScmRightsTruncatedEven`

...  but there is still another failure in this category:

[ TEST ] kj/async-io-test.c++:174: legacy test: AsyncIo/UnixSocket
kj/async-io-test.c++:206: LocalPeerIdentity for unix socket had null UID
stack: 
kj/async-io-test.c++:234: LocalPeerIdentity for unix socket had null UID
stack: 
[ FAIL ] kj/async-io-test.c++:174: legacy test: AsyncIo/UnixSocket (612 _s))

-----

humm... maybe I should send a PR to fix the libressl/ScmRights failures?

kentonv:
> pthread_mutex_destroy(&waiter.stupidMutex); strerror(pthreadError) = Invalid argument

I think this is because we're using `PTHREAD_MUTEX_INITIALIZER` to initialize the mutex instead of `pthread_mutex_init()`... technically `PTHREAD_MUTEX_INITIALIZER` should only be used for static-global mutexes which would never be destroyed. I think switching to `pthread_mutex_init()` at the start of `Mutex::wait()` should solve it. Probably should do the same with `PTHREAD_COND_INITIALIZER` used there as well.

> LocalPeerIdentity for unix socket had null UID

I guess this suggests that `LOCAL_PEERCRED` is not working? You'll have to debug this. This test is not very important as Cap'n Proto doesn't really use this functionality, it's just something KJ supports for applications to use.

Issue #1311, opened 2021-08-19
G++ 11 warnings


vlovich:
https://github.com/capnproto/capnproto/blob/master/c%2B%2B/src/capnp/compiler/node-translator.c%2B%2B#L2398
```
    /ekam-provider/canonical/capnp/compiler/node-translator.c++: In member function ‘capnp::Orphan<capnp::List<capnp::schema::Annotation, capnp::Kind::STRUCT> > capnp::compiler::NodeTranslator::compileAnnotationApplications(capnp::List<capnp::
      compiler::Declaration::AnnotationApplication, capnp::Kind::STRUCT>::Reader, kj::StringPtr)’:
    /ekam-provider/canonical/capnp/compiler/node-translator.c++:2398:14: warning: ‘*(short unsigned int*)((char*)&kind + offsetof(kj::_::NullableValue<capnp::compiler::Declaration::Which>,kj::_::NullableValue<capnp::compiler::Declaration::Which>:
      :<unnamed>))’ may be used uninitialized in this function [-Wmaybe-uninitialized]
     2398 |       } else if (*kind != Declaration::ANNOTATION) {
          |              ^~
```
Looks like a legit bug that'll be a crash as the only way to hit this branch is if `kind` is nullptr. Not sure what the uninitialized is about as it should be warning about a potential null derference? Not sure what the fix is.

```
    In member function ‘capnp::Type capnp::Schema::BrandArgumentList::operator[](capnp::uint) const’,
        inlined from ‘capnp::Type capnp::Schema::BrandArgumentList::operator[](capnp::uint) const’ at /ekam-provider/canonical/capnp/schema.c++:404:6:
    /ekam-provider/canonical/capnp/schema.c++:431:12: warning: ‘*(short unsigned int*)((char*)&result + offsetof(capnp::Type, capnp::Type::<unnamed>))’ may be used uninitialized [-Wmaybe-uninitialized]
      431 |     result = Type(static_cast<schema::Type::Which>(binding.which), binding.schema);
          |     ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    /ekam-provider/canonical/capnp/schema.c++: In member function ‘capnp::Type capnp::Schema::BrandArgumentList::operator[](capnp::uint) const’:
    /ekam-provider/canonical/capnp/schema.c++:417:8: note: ‘result’ declared here
[...]

PR #1314, opened 2021-08-21
Fix copy-pasto resulting in uninitialized use of `kind`.

kentonv:
In practice this would only result in incorrect behavior if the uninitialized value turned out to be equal to a certain non-zero constant. Hence, in practice, incorrect behavior was unlikely and maybe impossible. I wasn't able to trigger it in tests.

This is one part of #1311.

### DIFF
--- a/c++/src/capnp/testdata/errors.txt
+++ b/c++/src/capnp/testdata/errors.txt
@@ -51,13 +51,14 @@ file:136:3-10: error: 'dupName' previously defined here.
 file:139:15-16: error: Duplicate ordinal number.
 file:138:15-16: error: Ordinal @2 originally used here.
 file:142:7-16: error: Declaration recursively depends on itself.
-file:148:14-27: error: Not enough generic parameters.
-file:149:15-47: error: Too many generic parameters.
-file:150:18-49: error: Double-application of generic parameters.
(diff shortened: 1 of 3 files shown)


c++/src/capnp/testdata/errors.txt
```diff
@@ -51,13 +51,14 @@ file:136:3-10: error: 'dupName' previously defined here.
...
-file:149:15-47: error: Too many generic parameters.
-file:150:18-49: error: Double-application of generic parameters.
-file:151:38-43: error: Sorry, only pointer types can be used as generic parameters.
-file:154:30-44: error: Embeds can only be used when Text, Data, or a struct is expected.
-file:155:37-51: error: Couldn't read file for embed: no-such-file
-file:161:23-27: error: Only pointer parameters can declare their default as 'null'.
-file:162:10-16: error: 'stream' can only appear after '->', not before.
-file:162:10-16: error: A method declaration uses streaming, but '/capnp/stream.capnp' is not found in the import path. This is a standard file that should always be installed with the Cap'n Proto compiler.
-file:157:20-45: error: Import failed: nosuchfile-unused.capnp
+file:145:19-20: error: 'T' is not an annotation.
```
kentonv, line 54: This is the only new error. The rest of the file is the same errors as before but with all the line numbers incremented.

PR #1315, opened 2021-08-21
Fix build regression introduced in d3b0dc209 - the SOL_LOCAL is not defined on FreeBSD 11.x and below.

sobomax:
On FreeBSD versions <= 11 the magic value of "0" would be used instead.

### DIFF
--- a/c++/src/kj/async-io-unix.c++
+++ b/c++/src/kj/async-io-unix.c++
@@ -55,6 +55,11 @@
 #include <sys/ucred.h>
 #endif
 
+#if !defined(SOL_LOCAL) && (__FreeBSD__ || __DragonflyBSD__)
+// On DragonFly or FreeBSD < 12.2 you're supposed to use 0 for SOL_LOCAL.
+#define SOL_LOCAL 0
+#endif
+
 namespace kj {
 
 namespace {


c++/src/kj/async-io-unix.c++
```diff
@@ -1487,7 +1487,11 @@ kj::Own<PeerIdentity> SocketAddress::getIdentity(kj::LowLevelAsyncIoProvider& ll
       // MacOS / FreeBSD
       struct xucred creds;
       uint length = sizeof(creds);
+#if defined SOL_LOCAL
```
vlovich, line 1490: Perhaps cleaner to have this at the top of the file instead?

```
#ifndef SOL_LOCAL
// DragonFly or FreeBSD < 12.2
#define SOL_LOCAL 0
#endif
```

kentonv: I agree, and also I would want to restrict this to the operating systems where it is known to be correct, so...

#if !defined(SOL_LOCAL) && (__FreeBSD__ || __DragonflyBSD__)
// On DragonFly or FreeBSD < 12.2 you're supposed to use 0 for SOL_LOCAL.
#define SOL_LOCAL 0
#endif

sobomax:
@vlovich @kentonv thanks, the patch has been updated as suggested.

kentonv:
Thanks!

sobomax:
> Thanks!

Cygwin seems to be broken but I don't think that's my fault.

kentonv:
@sobomax Yeah I saw that but it looks unrelated. Probably a cygwin update broke us, bah.

PR #1317, opened 2021-08-23
Check that file descriptor is in binary mode

### LINKED ISSUE #1316: Set file descriptor to binary mode on Windows?
I wasted quite some time figuring out why we got corrupted Cap'n Proto messages from stdout on Windows. Turns out the Visual C++ runtime was converting LF to CRLF. The nasty part about this issue is that it only bites you if you happen to have an 0x0A in your serialized message. I wonder if it would make sense to make `writeMessageToFd` call [`_setmode(fd, _O_BINARY);`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/setmode?view=msvc-160) on Windows?

btj:
Prevents the Windows C runtime from converting LF to CRLF.
Fixes #1316


c++/src/capnp/serialize.c++
```diff
@@ -271,6 +271,11 @@ void readMessageCopy(kj::InputStream& input, MessageBuilder& target,
 
 void writeMessage(kj::OutputStream& output, kj::ArrayPtr<const kj::ArrayPtr<const word>> segments) {
   KJ_REQUIRE(segments.size() > 0, "Tried to serialize uninitialized message.");
+#ifdef _WIN32
+#ifndef NDEBUG
```
vlovich, line 275: Why does debug mode matter here?

vlovich (commented):
I think it’s best if @kentonv  weighs in.

c++/src/capnp/serialize.c++
btj: Just a tiny optimisation. In release mode, the `requireIsInBinaryMode()` call is a no-op anyway, but this removes even the call itself.

c++/src/kj/io.h
```diff
@@ -85,6 +85,14 @@ class OutputStream {
   // Equivalent to write()ing each byte array in sequence, which is what the default implementation
   // does.  Override if you can do something better, e.g. use writev() to do the write in a single
   // syscall.
+
+#ifdef _WIN32
+  virtual void requireIsInBinaryMode();
```
kentonv, line 90: Hmm, I'm not so excited about polluting the streams APIs with this.

But I realize after some thought that we probably can't enforce binary mode on _all_ streams since we use `FdOutputStream` to write text to the console in some places.

Ideas:
* Only check in `writeMessageToFd()`. If people manually set up an `FdOutputStream` and use `writeMessage()`, then it's on them to ensure binary mode.
* The `write()` method could, in debug mode, check if the buffer being written contains NUL ('\0') bytes. If so, that's a strong indication that it's binary. Then we could warn/throw if the fd is in text mode. Almost all serialized capnp will contain some zeros so this should work well for detecting the problem in the Cap'n Proto use case. You could even limit it to checking only the first 8 bytes and you'd probably catch all capnp messages.
* We could always do the checking in `FdOutputStream`'s constructor, except when the output is stdout/stderr, which we'd assume is a legitimate place to write text. I'd be surprised if anyone's using `FdOutputStream` to write text to somewhere other than stdout/stderr... nevertheless this could still possibly break people.

btj: OK, I'll add the check in `writeMessageToFd()`, and additionally, in FdOutputStream::write, if the fourth byte is NUL, I check that the stream is in binary mode. Since it's unlikely that the message has more than 2^24 segments, this is likely to catch all non-packed Cap'n Proto messages being written. This is unlikely to catch packed messages, though. But this seems like a good trade-off.

btj:
I ended up only doing the check in `writeMessageToFd` for now. The NUL check is a bit hacky: there are programs (such as `find -print0`) that output text with NULs.

The only complete solution that I see right now would be to extend the streams API on Windows. On Windows, text mode vs binary mode is a reality that we might as well take into account wherever relevant, including in the streams API. Short of that, doing the check just in `writeMessageToFd` seems like a good compromise.

c++/src/capnp/serialize.c++
vlovich: Oh my bad. I misread this as saying this optimization is live only in release builds which confused me.

vlovich:
I think it would be appropriate to have `FdOutputStream` do this check but maybe there's a caveat for someone using `TextCodec`/`JsonCodec` + `FdOutputStream`? Really it's not just that it's a bad idea to do line ending translation in low level i/o routines. Window's line ending handling hasn't held up in general the test of time when every other platform uses `\n` everywhere :\.

vlovich:
Hmm... @kentonv the cygwin build seems to be failing because `uncaught_exception` is deprecated: 
```
src/kj/exception.c++: In member function ‘virtual void kj::ExceptionCallback::RootExceptionCallback::onRecoverableException(kj::Exception&&)’:
src/kj/exception.c++:1020:14: error: ‘bool std::uncaught_exception()’ is deprecated [-Werror=deprecated-declarations]
```
I thought that was a suppressed warning...

c++/src/capnp/serialize.c++
```diff
@@ -301,6 +305,9 @@ void writeMessage(kj::OutputStream& output, kj::ArrayPtr<const kj::ArrayPtr<cons
 StreamFdMessageReader::~StreamFdMessageReader() noexcept(false) {}
 
 void writeMessageToFd(int fd, kj::ArrayPtr<const kj::ArrayPtr<const word>> segments) {
+#ifdef _WIN32
+  KJ_REQUIRE(_setmode(fd, _O_BINARY) == _O_BINARY, "Tried to write a message to a file descriptor that is in text mode. Set the file descriptor to binary mode by calling the _setmode Windows CRT function.");
```
kentonv, line 309: I think we should probably set the FD back to text mode before throwing.

```suggestion
    auto oldMode = _setmode(fd, _O_BINARY);
    if (oldMode != _O_BINARY) {
      _setmode(fd, oldMode);
      KJ_FAIL_REQUIRE("Tried to write a message to a file descriptor that is in text mode. Set the "
          "file descriptor to binary mode by calling the _setmode Windows CRT function, or passing "
          "_O_BINARY to _open().");
    }
```

(Also added a mention of _open() to the error message, and applied line wrapping to comply with KJ style rules.)

PR #1318, opened 2021-08-23
Minor doc improvements

garrison:
See each commit for explanation.


kentonv:
Hey I didn't review this.

The changes to add `../` on the style-guide references might help GitHub since it doesn't seem to be able to follow symlinks, but it breaks the way I was working on generating a web site out of this...

PR #1321, opened 2021-08-26
Add Cirrus CI configuration to test capnproto on a few actively maintained FreeBSD versions and fix few issues

sobomax:
This should run automatically on PRs and commits and will ensure no regressions are made over time (might need project maintainers to make few clicks to enable).

Few issues preventing unit test suite from completing have been identified and fixed.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,0 +1,34 @@
+env:
+  CIRRUS_CLONE_DEPTH: 1
+
+freebsd_task:
+  matrix:
+    - name: FreeBSD 13.0 (GCC 10 from packages)
+      # RunCatchingExceptionsOtherException fails on 13.0 with system Clang
+      # 11.0 and ports Clang 10/11 as well, so GCC 10 is used instead.
+      freebsd_instance:
+        image_family: freebsd-13-0
+      preinstall_script:
+        # Stock clang11 fails some exception unit tests
+        pkg install -y gcc10
+      env:
+        CC: gcc10
+        CXX: g++10
+    - name: FreeBSD 12.2 (System Clang 10)
+      freebsd_instance:
+        image_family: freebsd-12-2
+    - name: FreeBSD 11.4 (System Clang 10)
+      freebsd_instance:
+        image_family: freebsd-11-4
+  install_script:
+    pkg install -y automake autoconf libtool
+  compiler_version_script:
+    ${CXX:-"c++"} --version
+  autoreconf_script:
+    - cd c++ && autoreconf -i
+  configure_script:
+    - cd c++ && ./configure
+  build_script:
+    - make -C c++
+  test_script:
+    - make -C c++ check
--- a/c++/src/kj/async-io-test.c++
+++ b/c++/src/kj/async-io-test.c++
@@ -299,11 +299,11 @@ TEST(AsyncIo, AncillaryMessageHandlerNoMsg) {
 }
 #endif
 
-// We only support ancillary messages on Unix.
-// MacOS only supports SO_TIMESTAMP on datagram sockets, so this test won't work.
-// Android fails this test for unknown reasons and I don't care because it's an extremely obscure
-// feature anyway.
-#if !_WIN32 && !__CYGWIN__  && !__APPLE__ && !__ANDROID__
+// This test uses SO_TIMESTAMP on a SOCK_STREAM, which is only supported by Linux. Ideally we'd
+// rewrite the test to use some other message type that is widely supported on streams. But for
+// now we just limit the test to Linux. Also, it doesn't work on Android for some reason, and it
+// isn't worth investigating, so we skip it there.
+#if __linux__ && !__ANDROID__
 TEST(AsyncIo, AncillaryMessageHandler) {
   auto ioContext = setupAsyncIo();
   auto& network = ioContext.provider->getNetwork();
--- a/c++/src/kj/async-test.c++
+++ b/c++/src/kj/async-test.c++
@@ -25,16 +25,7 @@
 #include "mutex.h"
 #include "thread.h"
 
-// TODO(later): Refactor this more cleanly so that the KJ library itself defines this.
-#if !__BIONIC__ && !KJ_NO_EXCEPTIONS
-#define KJ_FIBERS_AVAILABLE 1
-#else
-#define KJ_FIBERS_AVAILABLE 0
-#endif
-
-#if !KJ_FIBERS_AVAILABLE
-// For bionic, OpenBSD, and no-exception builds, enables the regression test for
-// kj::TaskSet::~TaskSet potentially causing a stack overflow.
+#if !KJ_USE_FIBERS
 #include <pthread.h>
 #endif
 
@@ -759,7 +750,7 @@ TEST(Async, LargeTaskSetDestruction) {
     }
   };
 
-#if KJ_FIBERS_AVAILABLE
+#if KJ_USE_FIBERS
   EventLoop loop;
   WaitScope waitScope(loop);
 
@@ -998,7 +989,7 @@ KJ_TEST("exclusiveJoin both events complete simultaneously") {
   KJ_EXPECT(!joined.poll(waitScope));
 }
 
-#if KJ_FIBERS_AVAILABLE && !KJ_NO_EXCEPTIONS
+#if KJ_USE_FIBERS
 KJ_TEST("start a fiber") {
   EventLoop loop;
   WaitScope waitScope(loop);
--- a/c++/src/kj/async.c++
+++ b/c++/src/kj/async.c++
@@ -449,11 +449,13 @@ private:
 
   struct StartRoutine;
 
+#if KJ_USE_FIBERS
 #if _WIN32 || __CYGWIN__
   void* osFiber;
-#elif !__BIONIC__
+#else
   struct Impl;
   Impl* impl;
+#endif
 #endif
 
   [[noreturn]] void run();
@@ -1311,7 +1313,8 @@ const Executor& getCurrentThreadExecutor() {
 
 namespace _ {  // private
 
-#if !(_WIN32 || __CYGWIN__ || __BIONIC__)
+#if KJ_USE_FIBERS
+#if !(_WIN32 || __CYGWIN__)
 struct FiberStack::Impl {
   // This struct serves two purposes:
   // - It contains OS-specific state that we don't want to declare in the header.
@@ -1383,6 +1386,7 @@ struct FiberStack::Impl {
   }
 };
 #endif
+#endif
 
 struct FiberStack::StartRoutine {
 #if _WIN32 || __CYGWIN__
@@ -1436,14 +1440,13 @@ FiberStack::FiberStack(size_t stackSizeParam)
     // Force stackSize to a reasonable minimum.
     : stackSize(kj::max(stackSizeParam, 65536))
 {
-#if KJ_NO_EXCEPTIONS
-  KJ_UNIMPLEMENTED("Fibers are not implemented because exceptions are disabled");
 
-#elif _WIN32 || __CYGWIN__
+#if KJ_USE_FIBERS
+#if _WIN32 || __CYGWIN__
   // We can create fibers before we convert the main thread into a fiber in FiberBase
   KJ_WIN32(osFiber = CreateFiber(stackSize, &StartRoutine::run, this));
 
-#elif !__BIONIC__
+#else
   // Note: Nothing below here can throw. If that changes then we need to call Impl::free(impl)
   //   on exceptions...
   ucontext_t context;
@@ -1460,6 +1463,10 @@ FiberStack::FiberStack(size_t stackSizeParam)
   if (_setjmp(impl->originalJmpBuf) == 0) {
     setcontext(&context);
   }
+#endif
+#else
+#if KJ_NO_EXCEPTIONS
+  KJ_UNIMPLEMENTED("Fibers are not implemented because exceptions are disabled");
 #else
   KJ_UNIMPLEMENTED(
       "Fibers are not implemented on this platform because its C library lacks setcontext() "
@@ -1467,13 +1474,16 @@ FiberStack::FiberStack(size_t stackSizeParam)
       "We can likely make it happen using assembly, but didn't want to try unless it was "
       "actually needed.");
 #endif
+#endif
 }
 
 FiberStack::~FiberStack() noexcept(false) {
+#if KJ_USE_FIBERS
 #if _WIN32 || __CYGWIN__
   DeleteFiber(osFiber);
-#elif !__BIONIC__
+#else
   Impl::free(impl, stackSize);
+#endif
 #endif
 }
 
@@ -1545,23 +1555,27 @@ Maybe<Own<Event>> FiberBase::fire() {
 void FiberStack::switchToFiber() {
   // Switch from the main stack to the fiber. Returns once the fiber either calls switchToMain()
   // or returns from its main function.
+#if KJ_USE_FIBERS
 #if _WIN32 || __CYGWIN__
   SwitchToFiber(osFiber);
-#elif !__BIONIC__
+#else
   if (_setjmp(impl->originalJmpBuf) == 0) {
     _longjmp(impl->fiberJmpBuf, 1);
   }
+#endif
 #endif
 }
 void FiberStack::switchToMain() {
   // Switch from the fiber to the main stack. Returns the next time the main stack calls
   // switchToFiber().
+#if KJ_USE_FIBERS
 #if _WIN32 || __CYGWIN__
   SwitchToFiber(getMainWin32Fiber());
-#elif !__BIONIC__
+#else
   if (_setjmp(impl->fiberJmpBuf) == 0) {
     _longjmp(impl->originalJmpBuf, 1);
   }
+#endif
 #endif
 }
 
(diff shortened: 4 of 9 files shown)


c++/src/kj/string-test.c++
```diff
@@ -282,7 +282,7 @@ KJ_TEST("float stringification and parsing is not locale-dependent") {
   KJ_EXPECT("1.5"_kj.parseAs<double>() == 1.5);
 
   if (setlocale(LC_NUMERIC, "es_ES") == nullptr &&
-      setlocale(LC_NUMERIC, "es_ES.utf8") == nullptr) {
+      setlocale(LC_NUMERIC, "es_ES.UTF-8") == nullptr) {
```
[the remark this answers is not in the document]
sobomax: OK, I've added es_ES.utf8 back. Thanks!

c++/src/kj/async.c++
```diff
@@ -1311,7 +1313,7 @@ const Executor& getCurrentThreadExecutor() {
 
 namespace _ {  // private
 
-#if !(_WIN32 || __CYGWIN__ || __BIONIC__)
+#if KJ_USE_FIBERS
```
kentonv, line 1316: Looks like the meaning of the check changed here? `_WIN32` and `__CYGWIN__` should avoid this block.

c++/src/kj/async.c++
```diff
@@ -1460,21 +1463,25 @@ FiberStack::FiberStack(size_t stackSizeParam)
...
   }
+#endif
 #else
   KJ_UNIMPLEMENTED(
       "Fibers are not implemented on this platform because its C library lacks setcontext() "
       "and friends. If you'd like to see fiber support added, file a bug to let us know. "
       "We can likely make it happen using assembly, but didn't want to try unless it was "
       "actually needed.");
 #endif
+#endif
```
kentonv, line 1474: I can't figure out what `#if` this is matching. Maybe time to add comments to the `#endif`s to make clear? Like:

```
#endif  // KJ_USE_FIBERS
```

kentonv: Oh I see it's for KJ_NO_EXCEPTIONS. Perhaps we should roll that into `KJ_USE_FIBERS`? I.e. make sure to define `KJ_USE_FIBERS 0` when `KJ_NO_EXCEPTIONS` is enabled.

c++/src/kj/async-test.c++
```diff
@@ -26,7 +26,7 @@
 #include "thread.h"
 
 // TODO(later): Refactor this more cleanly so that the KJ library itself defines this.
-#if !__BIONIC__ && !KJ_NO_EXCEPTIONS
+#if KJ_USE_FIBERS && !KJ_NO_EXCEPTIONS
 #define KJ_FIBERS_AVAILABLE 1
```
kentonv, line 30: Can we just remove this define entirely and replace `KJ_FIBERS_AVAILABLE` with `KJ_USE_FIBERS`? (See the TODO, above.)

c++/src/kj/async-io-test.c++
```diff
@@ -300,10 +300,10 @@ TEST(AsyncIo, AncillaryMessageHandlerNoMsg) {
 #endif
 
 // We only support ancillary messages on Unix.
-// MacOS only supports SO_TIMESTAMP on datagram sockets, so this test won't work.
+// MacOS & FreeBSD only support SO_TIMESTAMP on datagram sockets, so this test won't work.
 // Android fails this test for unknown reasons and I don't care because it's an extremely obscure
 // feature anyway.
-#if !_WIN32 && !__CYGWIN__  && !__APPLE__ && !__ANDROID__
+#if !(_WIN32 || __CYGWIN__  || __APPLE__ || __ANDROID__ || __FreeBSD__)
```
kentonv, line 306: Nitpick, but I feel that the negation-of-ors is harder to follow than the previous ands-of-negations, can we keep it the old way and just add freeBSD to the end?

c++/src/kj/async.c++
sobomax: Fixed.

c++/src/kj/async.c++
sobomax: Done. It's bit hairy now that you start considering all permutations of what might be defined and what not.... But let's see if you like it and go from there. )

c++/src/kj/async-test.c++
sobomax: Done.

c++/src/kj/async.c++
```diff
@@ -451,7 +451,7 @@ private:
 
 #if _WIN32 || __CYGWIN__
   void* osFiber;
-#elif !__BIONIC__
+#elif !(__BIONIC__ || __FreeBSD__)
```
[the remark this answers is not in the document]
sobomax: Done. See my comment on that section becoming quite hairy. :)

c++/src/kj/async-io-test.c++
sobomax: Done. Your version is 3 character longer. :-P

```
-#if !(_WIN32 || __CYGWIN__  || __APPLE__ || __ANDROID__ || __FreeBSD__)
+#if !_WIN32 && !__CYGWIN__  && !__APPLE__ && !__ANDROID__ && !__FreeBSD__
```

sobomax:
> This looks very promising, especially the CI. For the Fibers, could you move the fibers to be using a `KJ_USE_FIBERS` macro so that we don't have this logic duplicated all over as discussed in #1313, #1208, & #1167?

Done, thanks!

c++/src/kj/async-test.c++
```diff
@@ -25,14 +25,7 @@
...
-// TODO(later): Refactor this more cleanly so that the KJ library itself defines this.
-#if !__BIONIC__ && !KJ_NO_EXCEPTIONS
-#define KJ_FIBERS_AVAILABLE 1
-#else
-#define KJ_FIBERS_AVAILABLE 0
-#endif
-
-#if !KJ_FIBERS_AVAILABLE
+#if !KJ_USE_FIBERS
 // For bionic, OpenBSD, and no-exception builds, enables the regression test for
```
sobomax, line 29: @vlovich This comment seems to be somewhat stale now. Not sure if I just need to remove it?

kentonv: Yeah, I'd remove this.

c++/src/kj/async.c++
```diff
@@ -1460,6 +1463,7 @@ FiberStack::FiberStack(size_t stackSizeParam)
   if (_setjmp(impl->originalJmpBuf) == 0) {
     setcontext(&context);
   }
+#endif
 #else
   KJ_UNIMPLEMENTED(
```
kentonv, line 1468: Can we bring back the custom error message for KJ_NO_EXCEPTIONS:

```
#if KJ_NO_EXCEPTIONS
    KJ_UNIMPLEMENTED("Fibers are not implemented because exceptions are disabled");
#else
    KJ_UNIMPLEMENTED(
      "Fibers are not implemented on this platform because its C library lacks setcontext() "
      "and friends. If you'd like to see fiber support added, file a bug to let us know. "
      "We can likely make it happen using assembly, but didn't want to try unless it was "
      "actually needed.");
#endif
```

c++/src/kj/async.h
```diff
@@ -27,6 +27,23 @@
...
+#define KJ_USE_FIBERS 1
+#else
+#define KJ_USE_FIBERS (!KJ_NO_EXCEPTIONS)
+#endif /* KJ_NO_EXCEPTIONS */
+#endif /* unsupported platforms */
+#else
+#if defined(KJ_NO_EXCEPTIONS) && KJ_NO_EXCEPTIONS && KJ_USE_FIBERS
+#error Fibers are not implemented because exceptions are disabled
+#endif /* KJ_NO_EXCEPTIONS */
+#endif /* KJ_USE_FIBERS */
```
kentonv, line 45: Let's use indentation here, and simplify things a bit. There's no need to check if KJ_NO_EXCEPTIONS is defined; if it's not defined, the preprocessor treats it as falsy. We generally use `#if KJ_NO_EXCEPTIONS` without first checking if it's defined elsewhere in KJ code.

```suggestion
#ifndef KJ_USE_FIBERS
  #if __BIONIC__ || __FreeBSD__ || __OpenBSD__ || KJ_NO_EXCEPTIONS
    // These platforms don't support fibers.
    #define KJ_USE_FIBERS 0
  #else
    #define KJ_USE_FIBERS 1
  #endif
#else
  #if KJ_NO_EXCEPTIONS && KJ_USE_FIBERS
    #error "Fibers cannot be enabled when exceptions are disabled."
  #endif
#endif
```

c++/src/kj/async-io-test.c++
```diff
@@ -300,10 +300,10 @@ TEST(AsyncIo, AncillaryMessageHandlerNoMsg) {
 #endif
 
 // We only support ancillary messages on Unix.
-// MacOS only supports SO_TIMESTAMP on datagram sockets, so this test won't work.
+// MacOS & FreeBSD only support SO_TIMESTAMP on datagram sockets, so this test won't work.
 // Android fails this test for unknown reasons and I don't care because it's an extremely obscure
 // feature anyway.
-#if !_WIN32 && !__CYGWIN__  && !__APPLE__ && !__ANDROID__
+#if !_WIN32 && !__CYGWIN__  && !__APPLE__ && !__ANDROID__ && !__FreeBSD__
```
kentonv, line 306: Actually looking at this again, if we're literally filtering every platform except linux, maybe we should just make it `#if __linux__ && !__ANDROID__`?

```suggestion
// SO_TIMESTAMP specifically seems to have lots of portability issues, so we only even attempt this
// test on Linux. Even Android fails this test for some unknown reason, so we avoid it there. This
// is an extremely obscure feature and it's just not worth the time to debug the test on every
// platform.
#if __linux__ && !__ANDROID__
```

.cirrus.yml
```diff
@@ -0,0 +1,34 @@
+env:
```
kentonv, line 1: Is there really no way to run FreeBSD on GitHub Actions? I was pretty happy that we managed to consolidate our CI a while back...

sobomax: Not without major hassle, unfortunately. Like via quemu or something of that sort. But in my experience using Cirrus with other projects it has quite low overhead and is otherwise "set and forget". Examples of some other projects that use both Actions and Cirrus are git itself: https://github.com/git/git/runs/3416539724

c++/src/kj/async-io-test.c++
sobomax: That comment is somewhat misleading in my view, I must point out. There is nothing wrong with SO_TIMESTAMP support (it's been around way before Linux came along), the only problem is that the test case specifically written around using that feature with SOCK_STREAM, a (mis-)feature that has been added into Linux kernel around 2009. I even dug a commit message at some point, but is too lazy to find it out. The whole idea of timestamping a data on a stream connection has no much sense, since you might be reading data from multiple TCP packets in the same recvmsg() call.

sobomax: P.S. The correct way to fix this would be to re-write test case to use datagram sockets, IMHO. (not volunteering) :)

c++/src/kj/async.c++
sobomax: Done. I am wondering if it's d3adc0d3() now with an #error statement in the async.h?

c++/src/kj/async-test.c++
sobomax: Removed.

c++/src/kj/async-io-test.c++
kentonv: Sure, I didn't write this test and wasn't aware of the specific details. I believe the test is specifically trying to test ancillary messages on streams, though, so changing it to test datagram sockets would defeat the purpose. How about:

```
// This test uses SO_TIMESTAMP on a SOCK_STREAM, which is only supported by Linux. Ideally we'd
// rewrite the test to use some other message type that is widely supported on streams. But for
// now we just limit the test to Linux. Also, it doesn't work on Android for some reason, and it
// isn't worth investigating, so we skip it there.
```

c++/src/kj/async.c++
kentonv: No... that error will only happen if `KJ_USE_FIBERS=1`, but this code here is for `KJ_USE_FIBERS=0`.

kentonv (approved):
This looks good but I can't merge as it's still marked "draft".

sobomax:
> This looks good but I can't merge as it's still marked "draft".

Give me few minutes, I am polishing few things. The last forced push is still pending, I'll move it into ready state once it's in.

kentonv:
Unfortunately GitHub doesn't provide diffs across force-pushes so I'd actually prefer that you use incremental commits and I can squash-merge later. Otherwise I have to review the whole thing every time...

sobomax:
> Unfortunately GitHub doesn't provide diffs across force-pushes so I'd actually prefer that you use incremental commits and I can squash-merge later. Otherwise I have to review the whole thing every time...

Oh, ok, will keep it in mind. You never know. I remember some of the other maintainers have been upset by me having 10-20 commits in the PR and demanding me to do squashing. So it's kinda habitual now...

sobomax:
@kentonv I can actually see the diff between forced requests (b600c9a was immediately before approval, 7a82d79 is the very last one). Not sure if it would work for code pulled from the github:

```
$ git diff   b600c9a..7a82d79
diff --git a/c++/src/kj/async-io-test.c++ b/c++/src/kj/async-io-test.c++
index bc833716..e2d12401 100644
--- a/c++/src/kj/async-io-test.c++
+++ b/c++/src/kj/async-io-test.c++
@@ -299,11 +299,11 @@ TEST(AsyncIo, AncillaryMessageHandlerNoMsg) {
 }
 #endif

-// We only support ancillary messages on Unix.
-// MacOS & FreeBSD only support SO_TIMESTAMP on datagram sockets, so this test won't work.
-// Android fails this test for unknown reasons and I don't care because it's an extremely obscure
-// feature anyway.
-#if !_WIN32 && !__CYGWIN__  && !__APPLE__ && !__ANDROID__ && !__FreeBSD__
+// This test uses SO_TIMESTAMP on a SOCK_STREAM, which is only supported by Linux. Ideally we'd
+// rewrite the test to use some other message type that is widely supported on streams. But for
+// now we just limit the test to Linux. Also, it doesn't work on Android for some reason, and it
+// isn't worth investigating, so we skip it there.
+#if __linux__ && !__ANDROID__
 TEST(AsyncIo, AncillaryMessageHandler) {
   auto ioContext = setupAsyncIo();
   auto& network = ioContext.provider->getNetwork();
$

File: .cirrus.yml [hunk 12]
```diff
@@ -0,0 +1,34 @@
...
+      freebsd_instance:
+        image_family: freebsd-11-4
+  install_script:
+    pkg install -y automake autoconf libtool
+  compiler_version_script:
+    ${CXX:-"c++"} --version
+  autoreconf_script:
+    - cd c++ && autoreconf -i
+  configure_script:
+    - cd c++ && ./configure
```
kentonv, line 30: Hey sorry, I should have looked more closely at this earlier. Can you use `super-test.sh` like all the other Unix CI targets, rather than building manually? It's specifically intended to be our CI script.

kentonv: Though if it's easier to merge this now and get CI configured before iterating on that, I'm OK with that... but we should have this use `super-test.sh` long-term.

sobomax: @kentonv I gave it a brief try and it looks like super-test.sh has some bash-only features and also fails to detect system Clang for some reason on FreeBSD. Therefore, I suggest perhaps flushing this PR since it affects other PRs in pipeline and I'll open a new one to switch Cirus to use super-test.sh once I have a chance to massage it to complete on FreeBSD. Thanks!

sobomax:
Hi @kentonv any chance to get this merged today, so that I can start massaging super-test.sh to make it usable on FreeBSD/Cyrus? I've got some interesting new failures coming out of it, having a working CI in master would be a good help to test against baseline.

```
ld: error: undefined symbol: kj::InFlightExceptionIterator::InFlightExceptionIterator()
>>> referenced by exception-test.c++
>>>               src/kj/capnp_test-exception-test.o:(kj::_::(anonymous namespace)::TestCase212::run())

ld: error: undefined symbol: kj::InFlightExceptionIterator::next()
>>> referenced by exception-test.c++
>>>               src/kj/capnp_test-exception-test.o:(kj::_::(anonymous namespace)::TestCase212::run())
>>> referenced by exception-test.c++
>>>               src/kj/capnp_test-exception-test.o:(kj::_::(anonymous namespace)::TestCase212::run())
>>> referenced by exception-test.c++
>>>               src/kj/capnp_test-exception-test.o:(kj::_::(anonymous namespace)::TestCase212::run())
c++: error: linker command failed with exit code 1 (use -v to see invocation)
*** [capnp-test] Error code 1

kentonv:
Thanks for working on this!

Note that I only do code reviews M/W/F, and I have dozens of code reviews every week, so I may be a bit slow at times. Sorry.

sobomax:
No worries, @kentonv thanks for taking care of this. Could you also look into enabling Cirrus CI for the capnproto project, so that it would integrate FreeBSD runs into PRs and commits? You can find instructions here: https://cirrus-ci.org/guide/quick-start/

I am on to get super-test running there.

kentonv:
Hi @sobomax,

I've put together a point release including the FreeBSD changes:

https://capnproto.org/capnproto-c++-0.9.1-rc1.tar.gz

Can you verify it works?

Sorry I haven't gotten around to setting up the CI yet.

PR #1331, opened 2021-09-02
Add new Exception::Type::CANCELED type

jasnell:
For use when indicating that an operation was explicitly canceled.

### DIFF
--- a/c++/src/capnp/rpc.capnp
+++ b/c++/src/capnp/rpc.capnp
@@ -1194,6 +1194,9 @@ struct Exception {
     # The server doesn't implement the requested method. If there is some other method that the
     # client could call (perhaps an older and/or slower interface), it should try that instead.
     # Otherwise, this should be treated like `failed`.
+
+    canceled @4;
+    # The operation was canceled.
   }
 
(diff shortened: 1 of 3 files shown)


kentonv:
Sorry, we can't add a new exception type so easily. The four exception types have existed since the very beginning of KJ and lots of code has been written around them, including the Cap'n Proto RPC protocol. The philosophy behind them is here:

https://github.com/capnproto/capnproto/blob/master/style-guide.md#exceptions

The right type for cancellation is not completely clear, but DISCONNECTED is usually the best bet. OVERLOADED might make sense if the reason for cancellation is specifically a timeout.

kentonv:
Actually, the really right answer is that whoever is invoking the cancellation should specify what exception to throw. But... that's obviously not an option for AbortController. I think DISCONNECTED is the right answer here.

jasnell:
Yep, that's fine. Personally I'd prefer a bit more specificity between `DISCONNECTED` for whatever reason and `CANCELED` by user action, but `DISCONNECTED` certainly works, especially if adding a new error type is difficult.

PR #1332, opened 2021-09-03
Disable coroutine trace test on arm64

harrishancock:
This failed in our downstream project's CI, but I can't reproduce the failure locally. I don't have time to dig too deep, so let's just disable this test as flakey.


vlovich:
I bet you this is also the same reason why the Sentry tests are missing stack traces on ARM64.

harrishancock:
Gonna merge since this is trivial.

PR #1335, opened 2021-09-09
Various TreeIndex debugging improvements

kentonv:
There's one bug fix, but I don't think it is possible to trigger in practice.

### DIFF
--- a/c++/src/kj/table-test.c++
+++ b/c++/src/kj/table-test.c++
@@ -24,6 +24,8 @@
 #include <set>
 #include <unordered_set>
 #include "hash.h"
+#include "time.h"
+#include <stdlib.h>
 
 namespace kj {
 namespace _ {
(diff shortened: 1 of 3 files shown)


c++/src/kj/table-test.c++
```diff
@@ -907,6 +908,69 @@ KJ_TEST("large tree table") {
   }
 }
 
+KJ_TEST("TreeIndex fuzz test") {
+  // A test which randomly modifies a TreeIndex to try to discover buggy state changes.
```
vlovich, line 912: Can you please expand on the set of buggy state changes this test is trying to fuzz?

c++/src/kj/table.c++
```diff
@@ -651,7 +651,7 @@ void BTreeImpl::renumber(uint oldRow, uint newRow, const SearchKey& searchKey) {
     auto& node = tree[pos].parent;
     uint indexInParent = searchKey.search(node);
     pos = node.children[indexInParent];
-    if (node.keys[indexInParent] == oldRow) {
+    if (indexInParent < kj::size(node.keys) && node.keys[indexInParent] == oldRow) {
```
vlovich, line 669: Can this be triggered via a synthetic test? Might be useful to at least right an explicit test to avoid a regression in the future. If there's no way to trigger it, then is it actually a bug or just something that needs your commit message as a comment?

c++/src/kj/table-test.c++
```diff
@@ -907,6 +908,69 @@ KJ_TEST("large tree table") {
...
+
+  uint seed = (kj::systemPreciseCalendarClock().now() - kj::UNIX_EPOCH) / kj::NANOSECONDS;
+  KJ_CONTEXT(seed);  // print the seed if the test fails
+  srand(seed);
+
+  Table<uint, TreeIndex<UintCompare>> table;
+
+  auto randomInsert = [&]() {
+    table.upsert(rand(), [](auto&&, auto&&) {});
```
vlovich, line 921: Nit: C rand() is pretty bad. STL's [talk](https://channel9.msdn.com/Events/GoingNative/2013/rand-Considered-Harmful) on this is pretty good although I'm sure you've seen it. The main issues are that it's an extremely bad PRNG as it only generates numbers in [0, 65535] & even then it's pretty biased in its output. Probably is fine for this test, but maybe use the C++11 rand routines instead?

```
std::random_device rnd;
std::mt19937_64 eng(rnd())
std::uniform_int_distribution<uint> dist;
auto rng = [&rnd, &dist] { return dist(eng); };

uint randomNumber = rng();
```

The mersenne twister isn't the best PRNG engine (I like the PCG family more) & `random_device` probably uses `/dev/random` instead of `getrandom`, but all of this should be fine for tests.

vlovich (approved):
Should the build configuration change in some way so that the CI for KJ is testing with `KJ_DEBUG_TABLE_IMPL`?

a-robinson (approved):

c++/src/kj/table-test.c++
kentonv: It hasn't found any.

c++/src/kj/table.c++
kentonv: > Can this be triggered via a synthetic test?

As explained in the commit message, it's not actually possible to trigger bad behavior from this.

I don't think it's necessary to write a comment explaining why we need to avoid accessing an array out-of-bounds.

c++/src/kj/table-test.c++
kentonv: The C++11 interface is atrocious. Why should I have to construct three different classes with obscure names just to make a damned random number? This is an interface designed by mathematicians.

`rand()` is fine for the present use case.

PR #1340, opened 2021-09-15
Add examples of using `filesystem.h` to the KJ tour and docs.

jyn514:
I had trouble with this while working on improvements to Ekam; it seems useful to document.

cc @vlovich - is this the sort of documentation you were thinking of? I certainly would have found it useful.
I also noticed that `capnp/schema-parser.h` has lots of useful examples around `parseFromDirectory` - maybe it makes sense to link to those from the kj docs? Or move parts of them to kj? I only found them by searching for `newDiskFilesystem`.

### DIFF
--- a/c++/src/capnp/serialize.h
+++ b/c++/src/capnp/serialize.h
@@ -57,10 +57,11 @@ public:
   //
   //     auto fs = kj::newDiskFilesystem();
   //     auto file = fs->getCurrent().openFile(kj::Path::parse("mydata"));
-  //     kj::Array<const kj::byte> fileBuffer = file->mmap(0, file->stat().size).releaseAsChars()
+  //     kj::Array<const kj::byte> fileBuffer = file->mmap(0, file->stat().size);
   //     kj::ArrayPtr<const capnp::word> messageBuffer(
-  //       reinterpret_cast<const capnp::word *>(fileBuffer.begin()), fileBuffer.size() / sizeof(capnp::word));
-  //     capnp::FlatArrayMessageReader capnpReader { messageBuffer };
+  //       reinterpret_cast<const capnp::word *>(fileBuffer.begin()),
+  //       fileBuffer.size() / sizeof(capnp::word));
+  //     capnp::FlatArrayMessageReader capnpReader(messageBuffer);
 
   kj::ArrayPtr<const word> getSegment(uint id) override;
 
--- a/c++/src/capnp/serialize.h
+++ b/c++/src/capnp/serialize.h
@@ -53,7 +53,14 @@ class FlatArrayMessageReader: public MessageReader {
 
 public:
   FlatArrayMessageReader(kj::ArrayPtr<const word> array, ReaderOptions options = ReaderOptions());
-  // The array must remain valid until the MessageReader is destroyed.
+  // The array must remain valid until the MessageReader is destroyed. Example:
+  //
+  //     auto fs = kj::newDiskFilesystem();
+  //     auto file = fs->getCurrent().openFile(kj::Path::parse("mydata"));
+  //     kj::Array<const kj::byte> fileBuffer = file->mmap(0, file->stat().size).releaseAsChars()
+  //     kj::ArrayPtr<const capnp::word> messageBuffer(
+  //       reinterpret_cast<const capnp::word *>(fileBuffer.begin()), fileBuffer.size() / sizeof(capnp::word));
+  //     capnp::FlatArrayMessageReader capnpReader { messageBuffer };
 
   kj::ArrayPtr<const word> getSegment(uint id) override;
 
@@ -169,7 +176,11 @@ void writeMessage(kj::OutputStream& output, kj::ArrayPtr<const kj::ArrayPtr<cons
 // Specializations for reading from / writing to file descriptors.
 
 class StreamFdMessageReader: private kj::FdInputStream, public InputStreamMessageReader {
-  // A MessageReader that reads from a stream-based file descriptor.
+  // A MessageReader that reads from a stream-based file descriptor. Example:
+  //
+  //     auto fs = kj::newDiskFilesystem();
+  //     auto file = fs->getCurrent().openFile(kj::Path::parse("mydata"));
+  //     capnp::StreamFdMessageReader capnpReader { KJ_ASSERT_NONNULL(file->getFd()) };
 
 public:
   StreamFdMessageReader(int fd, ReaderOptions options = ReaderOptions(),
--- a/kjdoc/tour.md
+++ b/kjdoc/tour.md
@@ -1022,19 +1022,20 @@ KJ provides an advanced, cross-platform filesystem API in `kj/filesystem.h`. Fea
 * Sparse files ("hole punching"), copy-on-write file cloning (`FICLONE`, `FICLONERANGE`), `sendfile()`-based copying, `renameat2()` atomic replacements, and more will automatically be used when available.
 
 ```c++
-// NOTE: in most cases, you should construct the `Filesystem` at the entry point of your program,
-// then pass it in as a parameter wherever it's used. This allows test implementations or user-space
-// filesystems.
+// NOTE: in most cases, you should construct the `Filesystem` at the entry
+// point of your program, then pass it in as a parameter wherever it's used.
+// This allows dependency injection of alternative filesystem implementations
+// for testing or other interesting use cases.
 kj::Own<kj::Filesystem> fs = kj::newDiskFilesystem();
-kj::Own<const kj::ReadableFile>> file = fs->getCurrent().openFile(kj::Path::parse("README.md"));
+kj::Own<const kj::ReadableFile>> file =
+    fs->getCurrent().openFile(kj::Path::parse("README.md"));
 kj::String text = file->readAllText();
 ```
 
-For large files, consider using mmap when you want to read the full contents of a file, which uses
-on demand paging as contents are accessed rather than buffering the contents into memory first.
+For large files, consider using mmap when you want to read the full contents of a file, which uses on demand paging as contents are accessed rather than buffering the contents into memory first.
 
 ```c++
-kj::Array<const char> fileBuffer = file->mmap(0, file->stat().size).releaseAsChars();
+kj::Array<const kj::byte> fileBuffer = file->mmap(0, file->stat().size);
 ```
 
 See the API reference (header file) for details.
--- a/kjdoc/tour.md
+++ b/kjdoc/tour.md
@@ -1021,7 +1021,24 @@ KJ provides an advanced, cross-platform filesystem API in `kj/filesystem.h`. Fea
 * Symlinks, hard links, listing directories, recursive delete, recursive create parents, recursive copy directory, memory mapping, and unnamed temporary files are all exposed and easy to use.
 * Sparse files ("hole punching"), copy-on-write file cloning (`FICLONE`, `FICLONERANGE`), `sendfile()`-based copying, `renameat2()` atomic replacements, and more will automatically be used when available.
 
+```c++
+// NOTE: in most cases, you should construct the `Filesystem` at the entry point of your program,
+// then pass it in as a parameter wherever it's used. This allows test implementations or user-space
+// filesystems.
+kj::Own<kj::Filesystem> fs = kj::newDiskFilesystem();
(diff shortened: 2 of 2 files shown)


kjdoc/tour.md
```diff
@@ -1021,6 +1021,19 @@ KJ provides an advanced, cross-platform filesystem API in `kj/filesystem.h`. Fea
 * Symlinks, hard links, listing directories, recursive delete, recursive create parents, recursive copy directory, memory mapping, and unnamed temporary files are all exposed and easy to use.
 * Sparse files ("hole punching"), copy-on-write file cloning (`FICLONE`, `FICLONERANGE`), `sendfile()`-based copying, `renameat2()` atomic replacements, and more will automatically be used when available.
 
+```c++
```
vlovich, line 1024: This needs some kind of a preface I think before just dumping the code. Something like "Here are API examples for accessing files."

kjdoc/tour.md
```diff
@@ -1021,6 +1021,19 @@ KJ provides an advanced, cross-platform filesystem API in `kj/filesystem.h`. Fea
...
+auto fs = kj::newDiskFilesystem();
```
vlovich, line 1025: This needs a comment clarifying that you are recommended to construct this at the outermost part of your program (i.e. main) & dependency inject the implementation where needed (that way, for example, you could have a test implementation that roots you somewhere or be given a user space network filesystem implementation)

vlovich (commented):
Maybe a similar example snippet for `FlatArrayMessageReader` & how to use it with `mmap`?

The `parseFromDirectory` documentation pointer seems like a good idea.

I'll defer to @kentonv for approval.

kjdoc/tour.md
jyn514: Hmm, the rest of the tour uses a pretty similar format: https://github.com/capnproto/capnproto/blob/363cefdf46d37f399ba705ba9d9c93e7fa74bcbe/kjdoc/tour.md#unit-testing-tips
I can change it if you like, though.

kjdoc/tour.md
jyn514: Sure, done.

jyn514:
> Maybe a similar example snippet for FlatArrayMessageReader & how to use it with mmap?

How *would* I use it with mmap? I'm getting errors that `ArrayPtr<byte>` can't be converted to `ArrayPtr<word>`.
```c++
  auto fileBuffer = file->mmap(0, file->stat().size);
  capnp::FlatArrayMessageReader capnpReader { fileBuffer.asBytes() }
```
```
✘ compile: kj_fstest.c++
    /ekam-provider/canonical/kj_fstest.c++:12:33: error: no matching constructor for initialization of '
      capnp::FlatArrayMessageReader'
    /ekam-provider/c++header/capnp/serialize.h:50:7: note: candidate constructor (the implicit copy constructor) not
      viable: no known conversion from 'ArrayPtr<PropagateConst<const unsigned char, kj::byte>>' (aka 'ArrayPtr<
      const unsigned char>') to 'const capnp::FlatArrayMessageReader' for 1st argument
    /ekam-provider/c++header/capnp/serialize.h:50:7: note: candidate constructor (the implicit move constructor) not
      viable: no known conversion from 'ArrayPtr<PropagateConst<const unsigned char, kj::byte>>' (aka 'ArrayPtr<
      const unsigned char>') to 'capnp::FlatArrayMessageReader' for 1st argument
    /ekam-provider/c++header/capnp/serialize.h:55:3: note: candidate constructor not viable: no known conversion 
      from 'ArrayPtr<PropagateConst<const unsigned char, kj::byte>>' to 'ArrayPtr<const capnp::word>' for 1st 
      argument
```

vlovich:
I would imagine something like this:
```
auto fileBuffer = file->mmap(0, file->stat().size);
KJ_ASSERT(fileBuffer.size() % sizeof(capnp::word) == 0, "capnp messages must be multiples of 8 bytes");
kj::ArrayPtr<capnp::word> messageBuffer(reinterpret_cast<const capnp::word>(fileBuffer.begin()), fileBuffer.size() / sizeof(capnp::word));
capnp::FlatArrayMessageReader capnpReader { messageBuffer };
```

c++/src/capnp/serialize.h
```diff
@@ -53,7 +53,14 @@ class FlatArrayMessageReader: public MessageReader {
 
 public:
   FlatArrayMessageReader(kj::ArrayPtr<const word> array, ReaderOptions options = ReaderOptions());
-  // The array must remain valid until the MessageReader is destroyed.
+  // The array must remain valid until the MessageReader is destroyed. Example:
+  //
+  //     auto fs = kj::newDiskFilesystem();
+  //     auto file = fs->getCurrent().openFile(kj::Path::parse("mydata"));
+  //     kj::Array<const kj::byte> fileBuffer = file->mmap(0, file->stat().size).releaseAsChars()
```
kentonv, line 60:
```suggestion
  //     kj::Array<const kj::byte> fileBuffer = file->mmap(0, file->stat().size);
```

c++/src/capnp/serialize.h
```diff
@@ -53,7 +53,14 @@ class FlatArrayMessageReader: public MessageReader {
...
+  //     kj::ArrayPtr<const capnp::word> messageBuffer(
+  //       reinterpret_cast<const capnp::word *>(fileBuffer.begin()), fileBuffer.size() / sizeof(capnp::word));
```
kentonv, line 62: 100-char limit

```suggestion
  //       reinterpret_cast<const capnp::word *>(fileBuffer.begin()),
  //       fileBuffer.size() / sizeof(capnp::word));
```

c++/src/capnp/serialize.h
```diff
@@ -53,7 +53,14 @@ class FlatArrayMessageReader: public MessageReader {
...
+  //     capnp::FlatArrayMessageReader capnpReader { messageBuffer };
```
kentonv, line 63: Braced initializers for non-structs weird me out, and you used a parenthesized initializer on the previous line anyway, so let's be consistent.

```suggestion
  //     capnp::FlatArrayMessageReader capnpReader(messageBuffer);
```

c++/src/capnp/serialize.h
```diff
@@ -169,7 +176,11 @@ void writeMessage(kj::OutputStream& output, kj::ArrayPtr<const kj::ArrayPtr<cons
 // Specializations for reading from / writing to file descriptors.
 
 class StreamFdMessageReader: private kj::FdInputStream, public InputStreamMessageReader {
-  // A MessageReader that reads from a stream-based file descriptor.
+  // A MessageReader that reads from a stream-based file descriptor. Example:
+  //
+  //     auto fs = kj::newDiskFilesystem();
+  //     auto file = fs->getCurrent().openFile(kj::Path::parse("mydata"));
+  //     capnp::StreamFdMessageReader capnpReader { KJ_ASSERT_NONNULL(file->getFd()) };
```
kentonv, line 183: This isn't a great demonstration of the KJ filesystem APIs since you're just pulling out the underlying FD. This won't actually even work on Windows, which doesn't use file descriptors for the filesystem.

An example using `open()` might be better:

```suggestion
  //     int fd = open("mydata", O_RDONLY);
  //     assert(fd >= 0);
  //     capnp::StreamFdMessageReader capnpReader(fd);
```

That said I'm not sure this is particularly exciting either, and it still has a bunch of weird issues on Windows: `open` and `O_RDONLY` both need to be prefixed with `_` on Windows, and `_O_BINARY` is really needed too, but other systems don't have that flag.

Plus, people should not be using `StreamFdMessageReader` for files anyway, they should be using mmap, like in your other example. The mmap approach is usually far superior. `StreamFdMessageReader` is really intended for sockets, pipes, etc.

So... I dunno, maybe there isn't a good example to show here.

kjdoc/tour.md
```diff
@@ -1021,7 +1021,24 @@ KJ provides an advanced, cross-platform filesystem API in `kj/filesystem.h`. Fea
 * Symlinks, hard links, listing directories, recursive delete, recursive create parents, recursive copy directory, memory mapping, and unnamed temporary files are all exposed and easy to use.
 * Sparse files ("hole punching"), copy-on-write file cloning (`FICLONE`, `FICLONERANGE`), `sendfile()`-based copying, `renameat2()` atomic replacements, and more will automatically be used when available.
 
+```c++
+// NOTE: in most cases, you should construct the `Filesystem` at the entry point of your program,
+// then pass it in as a parameter wherever it's used. This allows test implementations or user-space
+// filesystems.
```
kentonv, line 1027:
```suggestion
// NOTE: in most cases, you should construct the `Filesystem` at the entry
// point of your program, then pass it in as a parameter wherever it's used.
// This allows dependency injection of alternative filesystem implementations
// for testing or other interesting use cases.
```

"Dependency injection" is a common term describing the point here.

I wrapped this at 80 chars since this code will go into a document / web page where width may be more constrained.

kjdoc/tour.md
```diff
@@ -1021,7 +1021,24 @@ KJ provides an advanced, cross-platform filesystem API in `kj/filesystem.h`. Fea
...
+kj::Own<kj::Filesystem> fs = kj::newDiskFilesystem();
+kj::Own<const kj::ReadableFile>> file = fs->getCurrent().openFile(kj::Path::parse("README.md"));
```
kentonv, line 1029:
```suggestion
kj::Own<const kj::ReadableFile>> file =
    fs->getCurrent().openFile(kj::Path::parse("README.md"));
```

kjdoc/tour.md
```diff
@@ -1021,7 +1021,24 @@ KJ provides an advanced, cross-platform filesystem API in `kj/filesystem.h`. Fea
...
+kj::String text = file->readAllText();
+```
+
+For large files, consider using mmap when you want to read the full contents of a file, which uses
+on demand paging as contents are accessed rather than buffering the contents into memory first.
```
kentonv, line 1034:
```suggestion
For large files, consider using mmap when you want to read the full contents of a file, which uses on demand paging as contents are accessed rather than buffering the contents into memory first.
```

We actually don't wrap prose in this document.

kjdoc/tour.md
```diff
@@ -1021,7 +1021,24 @@ KJ provides an advanced, cross-platform filesystem API in `kj/filesystem.h`. Fea
...
+
+```c++
+kj::Array<const char> fileBuffer = file->mmap(0, file->stat().size).releaseAsChars();
```
kentonv, line 1037:
```suggestion
kj::Array<const kj::byte> fileBuffer = file->mmap(0, file->stat().size);
```

I think it's useful to make clear that `mmap()` returns `Array<const byte>`. Converting to chars is a separate concern and I suspect most mmap users may not actually want to do that.

kjdoc/tour.md
```diff
@@ -1021,7 +1021,24 @@ KJ provides an advanced, cross-platform filesystem API in `kj/filesystem.h`. Fea
...
+```
+
 See the API reference (header file) for details.
+See `parseFromDirectory` in `capnp/schema-parser.h` for more examples.
```
kentonv, line 1041: This document is about KJ, which is intended to be independent of Cap'n Proto and may actually split into its own project someday, so I'm not sure I like referring to capnp source files like this. Also the examples there are intertwined with SchemaParser. If we need better examples for KJ, we should probably write them directly in `filesystem.h`.

jyn514: I'll add some more examples in filesystem.h.

c++/src/capnp/serialize.h
kentonv: > That's a case where you can't use mmap.

Why not? Temp files are files.

(Also a newly-created temp file would be empty so I guess there's nothing to read.)

Maybe the example could open a named pipe? That's a case where you actually can't use mmap...

kjdoc/tour.md
kentonv: Sadly, userspace virtual filesystems are not a standard feature of operating systems. Technically on Linux there is FUSE but setting that up is complicated and requires privileges.

So yeah, the in-memory FS stuff is really meant for passing to KJ-based libraries. The most common use case would be unit tests, so that you can test code that does file manipulation without creating real files on disk.

jyn514:
I am not planning to spend much more time working on this. Feel free to close it if you don't think it's an improvement as-is.

PR #1342, opened 2021-09-19
Fix build on MSVC with conformant preprocessor enabled

### LINKED ISSUE #1307: KJ_IREQUIRE fails to compile on MSVC with /Zc:preprocessor enabled
The `KJ_IREQUIRE` macro currently branches based on MSVC vs. other compilers in its attempt to make the leading comma in a `__VA_ARGS__` block conditional. It uses `"" #__VA_ARGS__, __VA_ARGS__` on MSVC, and `#__VA_ARGS__, ##__VA_ARGS__` elsewhere. (I'm not sure what either does, are you stringifying `__VA_ARGS__`? In any case, `__VA_ARGS__` commas are a hard problem solved differently across compilers and versions, and I understand the difficulties and edge cases.)

https://github.com/capnproto/capnproto/blob/master/c++/src/kj/common.h#L297-L313

However, on recent versions of MSVC 2019, the [`/Zc:preprocessor` flag](https://devblogs.microsoft.com/cppblog/announcing-full-support-for-a-c-c-conformant-preprocessor-in-msvc/) makes its preprocessor act more similarly to GCC, not omitting the leading comma and instead depending on `##__VA_ARGS__` or `__VA_OPT__` to drop the leading comma. I'm not sure exactly what the difference is though. In any case, with `/Zc:preprocessor` passed, the MSVC case fails to compile and only the non-MSVC case works.

The blog says:

> The conformant preprocessor can be tested by checking if the macro _MSVC_TRADITIONAL is defined and set to 0.

And subsequently presents the incorrect code for testing if the conformant preprocessor is turned off:

```cpp
#if defined(_MSVC_TRADITIONAL) && _MSVC_TRADITIONAL
// old behavior
#else
// new behavior
#endif
```

This triggers `new behavior` both on old MSVC versions before `_MSVC_TRADITIONAL` existed, *and* on newer versions with `_MSVC_TRADITIONAL=0`. And this is *after* a commenter "corrected" the original blog post. Boolean logic is hard, De Morgan's laws are hard... I think Microsoft should've added a `__MSVC_NEW_PREPROCESSOR` flag instead, to simplify testing for the new preprocessor (1) and old preprocessor (unset or 0).

In any case, to to test for new behavior, you need `#if defined(_MSVC_TRADITIONAL) && !_MSVC_TRADITIONAL`, and to test for old behavior, you need `#if !defined(_MSVC_TRADITIONAL) || _MSVC_TRADITIONAL`. I think I got it right? Double negatives are hard.

In summary, the most concise fix is to switch to `#if _MSC_VER && !defined(__clang__) && (!defined(_MSVC_TRADITIONAL) || _MSVC_TRADITIONAL)`, but this turns out to be a rather long line of text.

There may be other ways to write the command, maybe just `_MSVC_TRADITIONAL` but this falsely detects MSVC before the macro [was introduced](https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-160#:~:text=_msvc_traditional%3A) (pre-VS2017 and older versions of VS2017 and VS2019) as "new preprocessor".

One alternative approach that doesn't introduce a new layer of nesting is:

```cpp
#ifndef _MSVC_TRADITIONAL
#define _MSVC_TRADITIONAL 1
#endif
#if _MSC_VER && !defined(__clang__) && _MSVC_TRADITIONAL
```

nyanpasu64:
MSVC's conformant preprocessor (`/Zc:preprocessor`) adds support for the `##__VA_ARGS__` extension and the `__VA_OPT__(,)` C++20 feature.

(OUTDATED) Technically `_MSVC_TRADITIONAL` is a [reserved identifier](https://stackoverflow.com/q/228783) and we shouldn't be defining it, though I don't think it's a problem in practice. If you don't like this, whether because it defines a reserved identifier, or because defining the identifier in a library header will affect users which check for the macro, there are alternative (but less clear IMO) approaches I listed in #1307.

Fixes #1307.

### DIFF
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -294,7 +294,10 @@ KJ_NORETURN(void unreachable());
 }  // namespace _ (private)
 
 #ifdef KJ_DEBUG
-#if _MSC_VER && !defined(__clang__)
+#if _MSC_VER && !defined(__clang__) && (!defined(_MSVC_TRADITIONAL) || _MSVC_TRADITIONAL)
+#define KJ_MSVC_TRADITIONAL_CPP 1
+#endif
+#if KJ_MSVC_TRADITIONAL_CPP
 #define KJ_IREQUIRE(condition, ...) \
     if (KJ_LIKELY(condition)); else ::kj::_::inlineRequireFailure( \
         __FILE__, __LINE__, #condition, "" #__VA_ARGS__, __VA_ARGS__)


c++/src/kj/common.h
```diff
@@ -294,7 +294,10 @@ KJ_NORETURN(void unreachable());
 }  // namespace _ (private)
 
 #ifdef KJ_DEBUG
-#if _MSC_VER && !defined(__clang__)
+#ifndef _MSVC_TRADITIONAL
```
vlovich, line 297: Could use a comment about that this is comparability support for Visual Studio < 2017 version 15.8.

I think a more elegant way that doesn’t run afoul of reserved identifiers is to do something like:
```
#define KJ_MSVC_TRADITIONAL_CPP (!defined(_MSVC_TRADITIONAL) || _MSVC_TRADITIONAL)
```

Then you can just use `KJ_MSVC_TRADITIONAL_CPP`.

nyanpasu64: I tried this:

```
#define KJ_MSVC_TRADITIONAL_CPP (!defined(_MSVC_TRADITIONAL) || _MSVC_TRADITIONAL)
#if _MSC_VER && !defined(__clang__) && KJ_MSVC_TRADITIONAL_CPP
```

It turns out clean in MSVC and Clang compilations, but the Qt Creator IDE reports the following warning:

```
-Wexpansion-to-defined:
common.h:298:40: warning: macro expansion producing 'defined' has undefined behavior
common.h:297:35: note: expanded from macro 'KJ_MSVC_TRADITIONAL_CPP'
```

kentonv: Sounds like you probably need to do:

```
#if _MSC_VER && (!defined(_MSVC_TRADITIONAL) || _MSVC_TRADITIONAL)
#define KJ_MSVC_TRADITIONAL_CPP 1
#endif
```

nyanpasu64: Why does your `#if` condition include `_MSC_VER` but not `!defined(__clang__)`? IMO I'd either remove the `_MSC_VER` check (this approach sets `KJ_MSVC_TRADITIONAL_CPP=1` even on non-MSVC compilers, but that should be fine if nobody checks it):

```cpp
#if !defined(_MSVC_TRADITIONAL) || _MSVC_TRADITIONAL
#define KJ_MSVC_TRADITIONAL_CPP 1
#endif
#if _MSC_VER && !defined(__clang__) && KJ_MSVC_TRADITIONAL_CPP
```

Or add the `!defined(__clang__)` check:

```cpp
#if _MSC_VER && !defined(__clang__) && (!defined(_MSVC_TRADITIONAL) || _MSVC_TRADITIONAL)
#define KJ_MSVC_TRADITIONAL_CPP 1
#endif
#if KJ_MSVC_TRADITIONAL_CPP
```

At which point the one-liner is simpler, and better since it doesn't leak a #define (though worse since it doesn't give the expression a name):

```cpp
#if _MSC_VER && !defined(__clang__) && (!defined(_MSVC_TRADITIONAL) || _MSVC_TRADITIONAL)
```

kentonv: I forgot about clang. Yes that should probably be in there too.

It seemed easier to define `KJ_MSVC_TRADITIONAL_CPP` once and then use that everywhere rather than have a multi-part check at every use site.

nyanpasu64: In the part of capnproto I build in my vendored project, this header is the only place I had to fix to make it build on conformant MSVC. Honestly I feel it's not worth the trouble to make capnproto's source files build properly on conformant MSVC.

## My attempt to make it build

When building the entire capnproto project on MSVC with `add_compile_options(/Zc:preprocessor)`, I found that debug.h and test.h also depended on the preprocessor. Both files already see `KJ_MSVC_TRADITIONAL_CPP` when needed, but neither directly includes common.h; should I add an explicit `#include "common.h"`?

Worse yet, the GCC branch of `KJ_ASSERT_NONNULL`/`KJ_REQUIRE_NONNULL` fails to build on MSVC due to using `({...})` statement expressions, and the MSVC branch fails to build on MSVC due to the preprocessor differences. How should this be handled?

Assuming you don't want 3 code paths, one possibility is to stop using statement expressions in the GCC path, and switch conformant MSVC to the GCC path. I implemented this change locally, and got the entire capnproto to compile, except for a single error where since debug.h takes the GCC branch, it triggers the "MSVC mis-deduces `ContextImpl<decltype(func)>` as `ContextImpl<int>`" bug again... 😢

Another is to not support building `KJ_REQUIRE_NONNULL` and `KJ_ASSERT_NONNULL` in conformant MSVC (debug.h is exported to clients through "capnp/message.h", but these macros are never expanded in public headers, only private source files). Then when all supported MSVC versions support the conformant preprocessor, switch the MSVC codepath to the conformant preprocessor but keep MSVC hacks as necessary. However this may have issues if you ever want to support both MSVC versions that lack the new preprocessor, and MSVC versions that lack the old one.

## Compiler tiers?

Is capnproto-c++'s main testing environment Linux GCC (and perhaps Clang), with Windows MSVC and Clang as less-supported, less-tested targets that frequently run into build issues?

kentonv: Every file includes `kj/common.h`, though sometimes it may be transitively through other headers.

I think we should be able to change `({...})` to IIFE lambdas if necessary.

`debug.h` is intended to be useful for any project that uses KJ, so we will need to fix it at some point. But I won't force you to do it if you don't want to. I would prefer, though, that `KJ_MSVC_TRADITIONAL_CPP` is only true when we're actually using MSVC, otherwise I think the macro is confusing. This should be a minor change to your PR.

nyanpasu64: Force pushed. Hopefully it's good enough for now.

PR #1344, opened 2021-09-19
Fix building test-helpers.c++ hasSubstring() on Windows C++17

nyanpasu64:
When I vendored capnproto into a C++20 project, test-helpers.c++ failed to build on Windows Clang for two reasons:

- std::boyer_moore_horspool_searcher() takes begin and end, not begin and size.
- std::search() requires #include <algorithm>.

This makes me think that the boyer_moore_horspool_searcher codepath was never tested to work. I did not test to ensure that `hasSubstring()` functions correctly, only that it compiles. There may be undiscovered logical bugs even after this PR is merged, so you should test it further yourself.

----

The build doesn't fail on Linux because it takes the memmem codepath. Non-vendored standalone capnproto builds don't fail on Windows, because standalone capnproto configures both Clang and MSVC in C++14 mode, causing them to take the TODO(perf) path.

----

Also why is ["test-helpers.c++" included in `kj_sources_lite`](https://github.com/capnproto/capnproto/blob/73b35402140c5afb54a0fe89bd6658d0dc22aea9/c%2B%2B/src/kj/CMakeLists.txt#L20)? I don't need it, since it's never used in library-only builds of kj and capnproto. Could it be moved to `kj_sources_heavy`, or only included if `BUILD_TESTING=ON`?

### DIFF
--- a/c++/src/kj/test-helpers.c++
+++ b/c++/src/kj/test-helpers.c++
@@ -34,8 +34,6 @@
 #include <process.h>
 #endif
 
-#include <functional>
-
 namespace kj {
 namespace _ {  // private
 
@@ -49,9 +47,6 @@ bool hasSubstring(StringPtr haystack, StringPtr needle) {
 
 #if !defined(_WIN32)
     return memmem(haystack.begin(), haystack.size(), needle.begin(), needle.size()) != nullptr;
-#elif defined(__cpp_lib_boyer_moore_searcher)
-    std::boyer_moore_horspool_searcher searcher{needle.begin(), needle.size()};
-    return std::search(haystack.begin(), haystack.end(), searcher) != haystack.end();
 #else
     // TODO(perf): This is not the best algorithm for substring matching. strstr can't be used
     //   because this is supposed to be safe to call on strings with embedded nulls.


vlovich:
Hmmm… maybe I somehow forgot to upload my fix for the Boyer Moore path before it got merged? I do recall this change for some reason.

lgtm.

> Also why is "test-helpers.c++" included in kj_sources_lite? I don't need it, since it's never used in library-only builds of kj and capnproto. Could it be moved to kj_sources_heavy, or only included if BUILD_TESTING=ON?

You’re right. Please feel free to move it to kj-test_sources

kentonv:
> You’re right. Please feel free to move it to kj-test_sources

No, this code was intentionally placed where it is. See the description of ca7304b52ed3892b858d1dcf4ded3dd94a13a1ad.

nyanpasu64:
In that case, should this file be kept in `kj_sources_lite` or moved to `kj_sources_heavy`?

kentonv:
Let's just delete the `defined(__cpp_lib_boyer_moore_searcher)` block. As you point out, it doesn't appear to have been tested, it isn't covered by our CI, and it isn't actually needed in any case since performance is not relevant here. As this PR demonstrates, having it probably creates more maintenance burden than it solves.

> In that case, should this file be kept in kj_sources_lite or moved to kj_sources_heavy?

Not sure, do we build tests in lite mode?

kentonv:
Note that I don't really recommend using lite mode. It was only ever there to support old versions of MSVC that couldn't compile the whole library, but we don't support those versions at all anymore. The right way to minimize code size is using dead code elimination when linking.

nyanpasu64:
> The right way to minimize code size is using dead code elimination when linking.

Shrug. The way I'm using capnproto is to vendor a fixed version of the library (containing patches which I've been upstreaming in my recent PRs) in my source code, and have users build it along with the rest of the program. I enable lite mode to compile a smaller subset of the library. This speeds up clean builds, which is not a big deal on a beefy machine, but is important on cheaper/older/laptop CPUs and CI servers.

This has some disadvantages though. For example, every time I edit the schema, I'm forced to run a compatible version of `capnpc` myself and commit the generated C++ files. This is because users may not have capnpc to regenerate the C++ files at build time, since I don't make users install capnp/capnpc from their distro, and I don't compile capnpc from my vendored capnp in lite mode. However, this has the advantage of simplifying the build process for users who don't edit the schema.

As an example of the difficulties of not committing autogenerated code, one time I tried to trace Anbox's OpenGL proxying to diagnose an app crash, but struggled since it depended on autogenerated C/C++ files not found in the repo and only created when you build the project (which I didn't do).

In any case, I don't plan to stop using lite mode for the time being.

nyanpasu64:
> Let's just delete the `defined(__cpp_lib_boyer_moore_searcher)` block. As you point out, it doesn't appear to have been tested, it isn't covered by our CI, and it isn't actually needed in any case since performance is not relevant here. As this PR demonstrates, having it probably creates more maintenance burden than it solves.

My force-push partly reverts bd88a9157e81000168e03e281172da4fb04e5e45. `#include <functional>` was added in that commit, so I assume it's unused in the rest of this file and can safely be removed. I glanced through the file and didn't see any `std::` usages, so I hope it's OK.

PR #1346, opened 2021-09-24
Make hashCode behave more consistently

vlovich:
Currently `hashCode(<string literal>)` takes the pointer-path meaning it
hashes just the pointer. This creates a weirdness in the API because
`hashCode(strPtr) != hashCode(strPtr.cStr())`. Disable hashCode for
naked C strings to avoid the ambiguity.
    
Similarly block all C arrays for a similar weirdness between T[n] and
ArrayPtr.

### DIFF
--- a/c++/src/kj/hash.h
+++ b/c++/src/kj/hash.h
@@ -90,6 +90,8 @@ struct HashCoder {
 
   template <typename T>
   uint operator*(T* ptr) const {
+    static_assert(!isSameType<Decay<T>, char>(), "Wrap in StringPtr if you want to hash string "
+        "contents. If you want to hash the pointer, cast to void*");
     if (sizeof(ptr) == sizeof(uint)) {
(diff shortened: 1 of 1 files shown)


c++/src/kj/hash.h
```diff
@@ -128,6 +130,13 @@ static KJ_CONSTEXPR(const) HashCoder HASHCODER = HashCoder();
 inline uint hashCode(uint value) { return value; }
 template <typename T>
 inline uint hashCode(T&& value) { return hashCode(_::HASHCODER * kj::fwd<T>(value)); }
+template <typename T, size_t N>
+inline uint hashCode(T (&arr)[N]) {
```
vlovich, line 134: This had to be here instead of as a member function on `HashCoder` because otherwise the `T[N]` was getting decayed by kj::fwd to `T*`.

PR #1347, opened 2021-09-24
Add KJ_UNWRAP_OR_RETURN macro.

kentonv:
This provides convenient syntax to return early from a function if a maybe is null, while avoiding the need to KJ_ASSERT_NONNULL() immediately after the check.

Unfortunately, the only way to implement this was with GCC statement expressions, a non-standard feature that MSVC does not support. So we can't actually use this in KJ proper, but it can be used by dependent projects that don't care about Windows.

cc @kkoenig

### DIFF
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -1260,6 +1260,60 @@ inline T* readMaybe(T* ptr) { return ptr; }
 
 #define KJ_IF_MAYBE(name, exp) if (auto name = ::kj::_::readMaybe(exp))
 
+#if __GNUC__
+// These two macros provide a friendly syntax to extract the value of a Maybe or return early.
+//
+// Use KJ_UNWRAP_OR_RETURN if you just want to return a simple value when the Maybe is null:
+//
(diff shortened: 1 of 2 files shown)


c++/src/kj/common.h
```diff
@@ -1260,6 +1260,60 @@ inline T* readMaybe(T* ptr) { return ptr; }
...
+    } \
+    kj::mv(_kj_result); \
+  }))
+
+#define KJ_UNWRAP_OR(value, block) \
+  (*({ \
+    auto _kj_result = ::kj::_::readMaybe(value); \
+    if (!_kj_result) { \
+      block; \
+      asm("KJ_UNWRAP_OR_block_is_missing_return_statement\n"); \
```
a-robinson, line 1311: Heh, that's an interesting way to generate a compiler error.

a-robinson (approved):
Nice! I'm excited to use this.

c++/src/kj/common.h
kentonv: I found it on Stack Overflow and I love it.

PR #1348, opened 2021-09-27
Improve TLS accept error handling

bcaimano:
Timeout errors are OVERLOADED, so our existing "interestingness" suppression did not cover them. Additionally, we want to record these errors as a metric, so we'll need a custom handler.


c++/src/kj/exception.h
```diff
@@ -440,6 +440,13 @@ kj::ArrayPtr<void* const> computeRelativeTrace(
 //
 // This is useful for debugging, when reporting several related traces at once.
 
+class ErrorHandler {
+  // A sink for Exceptions. Implement this interface when you need to provide custom error handling
+  // for generic code.
+public:
+  virtual void taskFailed(kj::Exception&& exception) = 0;
```
kentonv, line 447: The name `taskFailed()` is kind of specific to `TaskSet`. I'm not sure this is the right interface to make more general in this way.

In fact, I regret defining this interface for `TaskSet`. I think I should have had it take a `kj::Function<void(kj::Exception&&)>` to the constructor instead.

Can we do that instead, for the TLS error handler callback?

c++/src/kj/compat/tls.c++
```diff
@@ -419,20 +419,26 @@ private:
...
-    // that network issues that affect TLS can be more discoverable from the server side.
-    if (e.getType() != Exception::Type::DISCONNECTED) {
-      KJ_LOG(ERROR, "error accepting tls connection", kj::mv(e));
-    }
+    // Note that if an ErrorHandler was provided to the ctor, this will not be used.
+    switch (e.getType()) {
+      case Exception::Type::DISCONNECTED:
+      case Exception::Type::OVERLOADED: {
+        return;
+      } break;
```
kentonv, line 436: Nitpick: This bracing style for switch cases isn't used anywhere else in KJ.

* We normally wouldn't add braces if there's no variables defined in the scope.
* When we do have braces, the `break` goes inside the braces.
* We don't normally write `break` if the case already ended in a `return`.

I understand that you may have different preferences here but consistency is important. Using an inconsistent style may lead other readers to wonder why the code here looks different and whether they are missing something important.

Note that we compile with `-Werror=implicit-fallthrough` which will catch any case fallthrough that isn't explicitly annotated.

c++/src/kj/compat/tls.c++
```diff
@@ -419,20 +419,26 @@ private:
...
-    // TODO(someday): SSL connection failures may be a fact of normal operation but they may also
-    // be important diagnostic information. We should allow for an error handler to be passed in so
-    // that network issues that affect TLS can be more discoverable from the server side.
-    if (e.getType() != Exception::Type::DISCONNECTED) {
-      KJ_LOG(ERROR, "error accepting tls connection", kj::mv(e));
-    }
+    // Note that if an ErrorHandler was provided to the ctor, this will not be used.
+    switch (e.getType()) {
+      case Exception::Type::DISCONNECTED:
+      case Exception::Type::OVERLOADED: {
```
kentonv, line 434: Ideally it would be better if the timeout code wouldn't produce an OVERLOADED exception in the first place, rather than have this code trying to detect timeout errors by catching this error type.

The OVERLOADED error is a side-effect of using `timer.timeoutAfter()`, but it's not actually even the right error here, as this timeout doesn't actually suggest anything is overloaded. The reason `timeoutAfter()` chooses OVERLOADED as its exception type is because it's meant to be used on the client side when talking to a remote server -- if the server doesn't respond, we guess it's overloaded. But here we're on the server side and waiting for a client, and the assumption doesn't really hold.

The right way to get more control over how timeouts manifest is to use `promise.exclusiveJoin()` with `timer.afterDelay(timeout)`. This is e.g. how KJ's HTTP server implements timeouts. You could, for example, treat timeouts as DISCONNECTED with something like:

```
promise = promise.exclusiveJoin(
    timer.afterDelay(timeout).then([]() -> kj::Promise<void> {
  return KJ_EXCEPTION(DISCONNECTED, "client timed out during TLS handshake");
}))
```

Or, if we didn't want to think of timeouts as errors at all (my preference), but we still wanted to allow a custom error handler to gather metrics about them, we'd do something like:

```
promise = promise.exclusiveJoin(
    timer.afterDelay(timeout).then([this]() {
  errorHandler.notifyTimeout();
}))
```

This requires a dedicated error handler interface, though.

In retrospect I sort of regret adding `timer.timeoutAfter()`, or at least I wish I'd made it require the caller to specify the exception to use on error. I think the default of OVERLOADED has led to a lot of confusion about exception types...

c++/src/kj/exception.h
bcaimano: Yeah, happy to do so!

c++/src/kj/compat/tls.c++
bcaimano: I was wondering if you'd prefer the afterDelay form. I think I do as well. I'll switch it over.

If you're okay with it, I'll skip the custom error handler.

bcaimano:
Okay, this should be ready for another look. I had to take a `kj::Function<>&` so that `TlsContext::Options` could still have the same copy/move semantics. Hopefully, it's isomorphic to `kj::TaskSet::ErrorHandler&`.

c++/src/kj/compat/tls.h
```diff
@@ -97,6 +101,9 @@ class TlsContext {
 
     kj::Maybe<kj::Duration> acceptTimeout;
     // Timeout applied to accepting a new TLS connection. `timer` is required if this is set.
+
+    kj::Maybe<TlsErrorHandler&> acceptErrorHandler;
```
kentonv, line 105: Hmm, it's unusual to take a `kj::Function` by reference like this, and I worry it's going to lead to some awkward code and maybe UAF potential as people don't realize they have to keep the function object itself alive.

Did you find that a lot of stuff breaks if `Options` is converted to move-only semantics?

kentonv (commented):
Sorry for the slow response, this got buried in my inbox while I was moving last week.

c++/src/kj/compat/tls.h
bcaimano: > Hmm, it's unusual to take a kj::Function by reference like this, and I worry it's going to lead to some awkward code and maybe UAF potential as people don't realize they have to keep the function object itself alive.

That's a fair concern. I'm guessing you'd consider `kj::Function` a value type? And pass by reference is only appropriate for resource types?

> Did you find that a lot of stuff breaks if Options is converted to move-only semantics?

I think our production code itself builds with move-only semantics but it broke a test or two. If you're okay with trivially breaking some downstream builds here, I'm fine making the change. I can also make a nested class refcounted wrapper if you'd prefer that.

kentonv (approved):
LGTM, please squash fixup and then I'll merge.

PR #1358, opened 2021-10-14
Move BUILD_TESTING to root

### LINKED ISSUE #1323: 0.9.0: cmake do not provide test suite with BUILD_TESTING=ON
```console
[tkloczko@barrel x86_64-redhat-linux-gnu]$ ctest -vv
Test project /home/tkloczko/rpmbuild/BUILD/capnproto-c++-0.9.0/x86_64-redhat-linux-gnu
No tests were found!!!
```

vlovich:
This fixes #1323 since enable_testing has to be in the top-level
CMakeLists.txt file before any add_subdirectory call.

https://stackoverflow.com/questions/30250494/ctest-not-detecting-tests

### DIFF
--- a/c++/CMakeLists.txt
+++ b/c++/CMakeLists.txt
@@ -4,6 +4,7 @@ set(VERSION 0.10-dev)
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 
+include(CTest)
 include(CheckIncludeFileCXX)
 include(GNUInstallDirs)
 if(MSVC)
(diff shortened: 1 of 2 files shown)


Sarcasm:
FYI: if you use `include(CTest)` versus calling `enable_testing` you get the `BUILD_TESTING` option declared automatically, see:

- https://cmake.org/cmake/help/latest/module/CTest.html

So this, should be equivalent:

```patch
-option(BUILD_TESTING "Build unit tests and enable CTest 'check' target." ON)
-if (BUILD_TESTING)
-  enable_testing()
-endif()
+include(CTest)
```

c++/CMakeLists.txt
```diff
@@ -25,7 +25,6 @@ set(INSTALL_TARGETS_DEFAULT_ARGS
 
 # Options ======================================================================
 
-option(BUILD_TESTING "Build unit tests and enable CTest 'check' target." ON)
```
harrishancock, line 28: Hmm, does this break folks who use the `capnproto/c++/CMakeLists.txt` as the top-level CMakeLists.txt? As I recall, that file predates the `capnproto/CMakeLists.txt`, which, I think, was only introduced because Conan doesn't support non-root top-level CMakeLists.txt files.

Or was the `c++/CMakeLists.txt` always broken because we don't call `enable_testing()`?

PR #1371, opened 2021-10-30
Track source locations where Events were created in order to provide extra debugging info.

kentonv:
In the Workers Runtime, the performance cost of this change seems to be well within the noise.


kentonv:
The Cygwin failure is consistent but doesn't seem related to this change. Seems like it is having trouble looking up `localhost`. Not going to debug it now.

vlovich (approved):
LGTM

PR #1376, opened 2021-11-10
RFC/WIP: Attempt to define a CAPNP_EMBEDDED compilation option.

litghost:
This PR is a request for comment on the `CAPNP_EMBEDDED` option and how it is wound into the library.  The idea is to cut out portions of the kj and capnp library that won't exist in a bare metal embedded system.  This is suitable for:
 - Targets that support C++14/17 but lacks exceptions, RTTI, etc.
 - Cannot assume existence of POSIX support.
    - No file descriptors.
    - No file system.
    - No mmap.
 - Threading is not in use.
   - Library will **not** be safe to use in interrupt context as well as main context.

### DIFF
--- a/c++/src/capnp/arena.h
+++ b/c++/src/capnp/arena.h
@@ -102,7 +102,7 @@ private:
   KJ_DISALLOW_COPY(ReadLimiter);
 
   KJ_ALWAYS_INLINE(void setLimit(uint64_t newLimit)) {
-#if defined(__GNUC__) || defined(__clang__)
+#if (defined(__GNUC__) || defined(__clang__)) && !defined(KJ_NO_THREADS)
     __atomic_store_n(&limit, newLimit, __ATOMIC_RELAXED);
 #else
     limit = newLimit;
@@ -110,7 +110,7 @@ private:
   }
 
   KJ_ALWAYS_INLINE(uint64_t readLimit() const) {
-#if defined(__GNUC__) || defined(__clang__)
+#if (defined(__GNUC__) || defined(__clang__)) && !defined(KJ_NO_THREADS)
     return __atomic_load_n(&limit, __ATOMIC_RELAXED);
 #else
     return limit;
--- a/c++/src/capnp/serialize-packed.c++
+++ b/c++/src/capnp/serialize-packed.c++
@@ -439,6 +439,8 @@ PackedMessageReader::PackedMessageReader(
 
 PackedMessageReader::~PackedMessageReader() noexcept(false) {}
 
+#ifndef KJ_NO_FD
+
 PackedFdMessageReader::PackedFdMessageReader(
     int fd, ReaderOptions options, kj::ArrayPtr<word> scratchSpace)
     : FdInputStream(fd),
@@ -454,6 +456,8 @@ PackedFdMessageReader::PackedFdMessageReader(
                           options, scratchSpace) {}
 
 PackedFdMessageReader::~PackedFdMessageReader() noexcept(false) {}
+
+#endif
 
 void writePackedMessage(kj::BufferedOutputStream& output,
                         kj::ArrayPtr<const kj::ArrayPtr<const word>> segments) {
@@ -472,10 +476,14 @@ void writePackedMessage(kj::OutputStream& output,
   }
 }
 
+#ifndef KJ_NO_FD
+
 void writePackedMessageToFd(int fd, kj::ArrayPtr<const kj::ArrayPtr<const word>> segments) {
   kj::FdOutputStream output(fd);
   writePackedMessage(output, segments);
 }
+
+#endif
 
 size_t computeUnpackedSizeInWords(kj::ArrayPtr<const byte> packedBytes) {
   const byte* ptr = packedBytes.begin();
--- a/c++/src/kj/exception.c++
+++ b/c++/src/kj/exception.c++
@@ -38,7 +38,9 @@
 #include "string.h"
 #include "debug.h"
 #include "threadlocal.h"
+#ifndef KJ_NO_POSIX
 #include "miniposix.h"
+#endif
 #include "function.h"
 #include "main.h"
 #include <stdlib.h>
@@ -46,7 +48,7 @@
 #include <new>
 #include <signal.h>
 #include <stdint.h>
-#ifndef _WIN32
+#if !defined(_WIN32) && !defined(KJ_NO_POSIX)
 #include <sys/mman.h>
 #endif
 #include "io.h"
@@ -609,13 +611,16 @@ namespace {
                          "\nstack: ", stringifyStackTraceAddresses(trace),
                          stringifyStackTrace(trace), '\n');
 
+#ifndef KJ_NO_POSIX
   FdOutputStream(STDERR_FILENO).write(message.begin(), message.size());
   _exit(1);
+#endif
 }
 
 }  // namespace
 
 void printStackTraceOnCrash() {
+#ifndef KJ_NO_POSIX
   // Set up alternate signal stack so that stack overflows can be handled.
   stack_t stack;
   memset(&stack, 0, sizeof(stack));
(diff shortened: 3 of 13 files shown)


c++/src/capnp/arena.h
```diff
@@ -102,15 +102,15 @@ class ReadLimiter {
   KJ_DISALLOW_COPY(ReadLimiter);
 
   KJ_ALWAYS_INLINE(void setLimit(uint64_t newLimit)) {
-#if defined(__GNUC__) || defined(__clang__)
+#if (defined(__GNUC__) || defined(__clang__)) && !defined(KJ_NO_THREADS)
```
kentonv, line 105: Are the `__atomic_*` builtins not supported on embedded systems? These are compiler intrinsics which usually don't require any special instructions, they just prevent the compiler from re-ordering operations. I thought they existed on all targets...

c++/src/capnp/serialize-packed.c++
```diff
@@ -439,6 +439,8 @@ PackedMessageReader::PackedMessageReader(
 
 PackedMessageReader::~PackedMessageReader() noexcept(false) {}
 
+#ifndef KJ_NO_FD
```
kentonv, line 442: Our preferred style throughout KJ code is `#if !KJ_NO_FD`, relying on the fact that undefined macros are implicitly falsy, rather than checking whether the macro is defined. (Note that the compiler flag `-D` defaults to using a value of `1`, so it works with this style.)

c++/src/kj/exception.c++
```diff
@@ -1263,7 +1282,11 @@ size_t sharedSuffixLength(kj::ArrayPtr<void* const> a, kj::ArrayPtr<void* const>
 
 kj::ArrayPtr<void* const> computeRelativeTrace(
     kj::ArrayPtr<void* const> trace, kj::ArrayPtr<void* const> relativeTo) {
+#ifndef KJ_NO_POSIX
```
kentonv, line 1285: When an `#else` is present, we prefer the positive case comes first, otherwise the else clause is a double-negative which can be hard to think about. (Actually in this case it's worse, since the flag itself is a negative, so the `#ifndef` is a double-negative and the `#else` is a triple-negative.)

Can you please reverse all the `#ifndef`s that are followed by `#else`?

c++/src/kj/mutex.c++
```diff
@@ -850,6 +850,15 @@ void Once::reset() {
   InitOnceInitialize(&coercedInitOnce);
 }
 
+#elif defined(KJ_NO_THREADS)
+
+Mutex::Mutex() {}
+Mutex::~Mutex() {}
+bool Mutex::lock(Exclusivity exclusivity, Maybe<Duration> timeout, NoopSourceLocation) {
+  return true;
```
kentonv, line 858: Perhaps in no-threads mode, `kj::Mutex` should still use a boolean to track whether it is locked, and abort if a recursive lock is attempted? (Such a recursive lock would deadlock when threads are enabled.)

c++/src/kj/mutex.h
```diff
@@ -197,6 +197,8 @@ class Mutex {
 #elif _WIN32 || __CYGWIN__
   uintptr_t srwLock;  // Actually an SRWLOCK, but don't want to #include <windows.h> in header.
 
+#elif defined(KJ_NO_THREADS)
+  uint32_t dummy_mutex;
```
kentonv, line 201: Please keep the naming style consistent -- this should be `dummyMutex`.

What does this variable do, though? It looks like it is unused. Can you just delete it?

Similarly for the other "dummy" fields below.

c++/src/kj/mutex.h
```diff
@@ -285,6 +289,8 @@ class Once {
     // Fast path check to see if runOnce() would simply return immediately.
 #if KJ_USE_FUTEX
     return __atomic_load_n(&futex, __ATOMIC_ACQUIRE) == INITIALIZED;
+#elif defined(KJ_NO_THREADS)
+    return true;
```
kentonv, line 293: This is not correct. When threads are disabled, we still need to run the initializer, and make sure it is run just once. This line appears to result in the initializer never running at all?

c++/src/kj/mutex.h
```diff
@@ -236,6 +238,8 @@ class Mutex {
 #elif _WIN32 || __CYGWIN__
     uintptr_t condvar;
     // Actually CONDITION_VARIABLE, but don't want to #include <windows.h> in header.
+#elif defined(KJ_NO_THREADS)
+    uint32_t dummy_condvar;
```
kentonv, line 242: Condvars aren't actually implemented (and can't be implemented) in the absence of threads, so should the whole condvar interface (`wait()` / `when()`) be `#if`ed out in no-threads mode?

c++/src/kj/exception.c++
```diff
@@ -1263,7 +1282,11 @@ size_t sharedSuffixLength(kj::ArrayPtr<void* const> a, kj::ArrayPtr<void* const>
...
   using miniposix::ssize_t;
+#else
+  using ssize_t = int;
```
kentonv, line 1288: Maybe should be `intptr_t`?

c++/src/kj/exception.c++
```diff
@@ -609,13 +611,16 @@ namespace {
                          "\nstack: ", stringifyStackTraceAddresses(trace),
                          stringifyStackTrace(trace), '\n');
 
+#ifndef KJ_NO_POSIX
```
kentonv, line 614: Removing the last two lines of the crash handler seems to make it useless. Should the entire `crashHandler()` function be `#if`'d out instead? It looks like it won't be used anyway.

c++/src/kj/exception.c++
```diff
@@ -609,13 +611,16 @@ namespace {
...
   FdOutputStream(STDERR_FILENO).write(message.begin(), message.size());
   _exit(1);
+#endif
 }
 
 }  // namespace
 
 void printStackTraceOnCrash() {
+#ifndef KJ_NO_POSIX
```
kentonv, line 623: Instead of removing the body of `printStackTraceOnCrash()`, maybe the whole function should not be defined? It's weird to let people call it but have it do nothing.

c++/src/kj/exception.c++
```diff
@@ -1047,6 +1054,7 @@ public:
     text = str(kj::repeat('_', contextDepth), file, ":", line, ": ", severity, ": ",
                mv(text), '\n');
 
+#ifndef KJ_NO_POSIX
```
kentonv, line 1057: I guess logs just won't go anywhere unless the application defines an ExceptionCallback? This seems kind of bad as those logs are usually important for debugging problems... I worry about people not even realizing that they're missing the logs.

kentonv (commented):
Sorry for the long delay in replying.

In general this seems like a reasonable idea.

It would be nice to have some sort of a CI test to verify that this mode actually does compile, otherwise we'll surely break it. It would be even nicer to have unit tests... but I suppose it's pretty hard to define unit tests when you don't even have a standard output for writing a test log. What is the standard practice here?

c++/src/kj/mutex.c++
litghost: I like that!

c++/src/kj/mutex.h
litghost: There is a `static_assert` that explodes if the Mutex has a too small size.  I'll paste it in.

c++/src/kj/exception.c++
litghost: > It would be nice to have some sort of a CI test to verify that this mode actually does compile, otherwise we'll surely break it. 

Agreed, will do.

> It would be even nicer to have unit tests... but I suppose it's pretty hard to define unit tests when you don't even have a standard output for writing a test log. What is the standard practice here?

We can compile and test the unit tests on the host system.  I'll work on something.

litghost:
> It would be nice to have some sort of a CI test to verify that this mode actually does compile, otherwise we'll surely break it. 

Agreed, will do.

> It would be even nicer to have unit tests... but I suppose it's pretty hard to define unit tests when you don't even have a standard output for writing a test log. What is the standard practice here?

We can compile and test the unit tests on the host system.  I'll work on something.

c++/src/kj/mutex.h
litghost: Valid, I'll fix

PR #1377, opened 2021-11-12
Work around old OpenSSL versions not being able to verify Let's Encrypt certificate chains.

kentonv:
As of September 30, 2021, old versions of OpenSSL (1.0.x) could no longer verify Let's Encrypt certificate chains by default, due to the problems described here:

https://letsencrypt.org/docs/dst-root-ca-x3-expiration-september-2021/
https://community.letsencrypt.org/t/openssl-client-compatibility-changes-for-let-s-encrypt-certificates/143816

It turns out the problem can be solved by setting a flag. That flag actually became the default in 1.1.0.

Of course, no one should be using OpenSSL 1.0.x anymore, but the various OpenSSL forks fixed this problem at different times, and the FIPS-certified version of BoringSSL is known to have this problem as well, thus requiring the work-around.

### DIFF
--- a/c++/src/kj/compat/tls.c++
+++ b/c++/src/kj/compat/tls.c++
@@ -151,6 +151,14 @@ public:
       throwOpensslError();
     }
 
+    // As of OpenSSL 1.1.0, X509_V_FLAG_TRUSTED_FIRST is on by default. Turning it on for older
+    // versions -- as well as certain OpenSSL-compatible libraries -- fixes the problem described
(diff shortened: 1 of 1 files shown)


vlovich:
Nice find! Lgtm

vlovich (approved):

kentonv:
BTW I found the work-around here: https://www.openssl.org/blog/blog/2021/09/13/LetsEncryptRootCertExpire/

PR #1379, opened 2021-11-13
Tweak BTreeImpl::verifyNode().

kentonv:
Previously we would compare slot `i` against slot `i+1` to check that ordering is correct. However, the basic validity of slot `i+1` hadn't been verified yet. It it was wildly invalid, this could segfault. If we instead compare `i-1` against `i`, then we know both slots have been basically verified already. If a slot is totally broken, this should now throw instead of segfaulting.

### DIFF
--- a/c++/src/kj/table.c++
+++ b/c++/src/kj/table.c++
@@ -264,9 +264,9 @@ size_t BTreeImpl::verifyNode(size_t size, FunctionParam<bool(uint, uint)>& f,
     for (auto i: kj::zeroTo(n)) {
       KJ_ASSERT(*parent.keys[i] < size, n, i);
       total += verifyNode(size, f, parent.children[i], height - 1, parent.keys[i]);
-      if (i + 1 < n) {
-        KJ_ASSERT(f(*parent.keys[i], *parent.keys[i + 1]),
-            n, i, parent.keys[i], parent.keys[i + 1]);
+      if (i > 0) {
+        KJ_ASSERT(f(*parent.keys[i - 1], *parent.keys[i]),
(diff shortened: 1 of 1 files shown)


c++/src/kj/table.c++
```diff
@@ -264,9 +264,9 @@ size_t BTreeImpl::verifyNode(size_t size, FunctionParam<bool(uint, uint)>& f,
     for (auto i: kj::zeroTo(n)) {
       KJ_ASSERT(*parent.keys[i] < size, n, i);
```
kentonv, line 265: Specifically, this line verifies that `keys[i]` is a valid index into the table. We should verify this before trying to access the row it targets.

c++/src/kj/table.c++
```diff
@@ -279,13 +279,14 @@ size_t BTreeImpl::verifyNode(size_t size, FunctionParam<bool(uint, uint)>& f,
...
-            n, i, leaf.rows[i], leaf.rows[i + 1]);
-      } else if (maxRow != nullptr) {
-        KJ_ASSERT(leaf.rows[n-1] == maxRow);
+      if (i > 0) {
+        KJ_ASSERT(f(*leaf.rows[i - 1], *leaf.rows[i]),
+            n, i, leaf.rows[i - 1], leaf.rows[i]);
       }
     }
+    if (maxRow != nullptr) {
+      KJ_ASSERT(leaf.rows[n-1] == maxRow, n);
```
harrishancock, line 288: `n` is guaranteed to be > 0 here? (This would be a preexisting problem if it were one, of course.)

harrishancock (approved):

PR #1380, opened 2021-11-13
Fix/refactor AsyncTee

kentonv:
This eliminates the vector of branch state structs and replaces them with a `List` of branch objects.


harrishancock (approved):
Looks like a nice improvement. My only comment is if it's easy to add a regression test for the UAF, that might be worthwhile.

PR #1385, opened 2021-11-18
Log exceptions when serving HTTP errors by default

jyn514:
Previously, KJ would return a 5XX error to the client, but wouldn't actually log the error anywhere,
making it difficult to tell what went wrong.

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -5528,7 +5528,7 @@ kj::Promise<void> HttpServerErrorHandler::handleApplicationError(
   }
 
   KJ_IF_MAYBE(r, response) {
-    KJ_LOG(INFO, "Internal error serving HTTP response", exception);
+    KJ_LOG(INFO, "threw exception while serving HTTP response", exception);
 
     HttpHeaderTable headerTable {};
     HttpHeaders headers(headerTable);
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -5528,6 +5528,8 @@ kj::Promise<void> HttpServerErrorHandler::handleApplicationError(
   }
 
   KJ_IF_MAYBE(r, response) {
+    KJ_LOG(INFO, "Internal error serving HTTP response", exception);
+
     HttpHeaderTable headerTable {};
     HttpHeaders headers(headerTable);
     headers.set(HttpHeaderId::CONTENT_TYPE, "text/plain");


kentonv:
Different applications may have different opinions on whether they want this logging, so changing it for everyone seems rash.

We already override this callback in the Workers Runtime, so this won't even affect Workers. What are you using it for?

jyn514:
> What are you using it for?

Parts of the test suite have HTTP servers and were hitting 500 errors. @xortive suggested that I add the logging in KJ rather than adding it to each server one by one.

xortive:
I'd expect the default behavior for a HttpService to log exceptions -- it's pretty confusing during the development process that they return 500s and don't log anywhere. Production services probably should be implementing their own callback anyways, but it's nice for development ones to not have to just to be able to debug.

kentonv:
For better or worse, HttpServer was designed to, by default, return exception stack traces in the response body, rather than log them. That may or may not have been the right design. But Workers isn't the only thing built on KJ HTTP. How do you propose we determine whether changing the default is the right thing for all users?

jyn514:
When would it be the *wrong* decision? At worse it makes the logs slightly more noisy and you still have the option to override the method.

kentonv:
1. This change broke our tests. If it broke our tests it could break other people's tests too.
2. "makes the logs slightly more noisy" is something you could say about Workers specifically, which happily spams logs for all kinds of stuff. But not everyone uses logs the same way and some may not be so accepting of sudden new log spam.

xortive:
I see your point with your objections, but before we abandon improving the situation , can we do something configurable? either:

* something like `ExceptionCallback` where the current behavior is the default, but can be overridden by creating a `HttpServerErrorHandler` on the stack that's used for all new `HttpService`s in that thread by default, unless a different one is specified at creation time.

* build-time #define, but this feels less elegant and is a pattern I haven't seen much in KJ.

The "right answer" for behavior of error logging here I think varies more by team/codebase, rather than where a given `HttpService` is used, and it'd be nice to be able to say all `HttpService`s made in the test harness, log to `stderr`.

kentonv:
@xortive I don't really get the motivation here. We don't have that many places in our codebase that create HttpServer, how hard can it be to have each one use some alternative HttpServerErrorHandler?

`ExceptionCallback` works the way it does because it's supposed to to be a work-around for people who don't want to enable exception handling. In general I think relying on ambient thread-locals in this way is not good design.

Concretely, it would be very weird if some third-party library that happens to use `HttpServer` behaved totally differently when called in one thread vs. another.

xortive:
Smaller one-off ones are getting more use within C++ tests, and I felt that it was rather confusing for this one part of KJ/capnp to swallow exceptions when the other are relatively noisy about logging them like the RpcSystem logging, which seems analogous to what we'd like to achieve here and is rather useful during debugging: https://github.com/capnproto/capnproto/blob/e5af75ff0513e6d971fa7b48e8108427766e51f9/c%2B%2B/src/capnp/rpc.c%2B%2B#L160

I feel we should at least be consistent between various kinds of server implementations, coming from writing `capnp::Capability::Server`s first it's not obvious when writing an HttpServer that you need to opt-in to getting logging -- you have to read a good bit of `http.h` to discover the options for it.

kentonv:
Cap'n Proto RPC only logs errors at `INFO` level, which is silent by default. I'm totally happy to add an `INFO` log here, but the current PR introduces an `ERROR` log.

jyn514:
Oh! Well in that case I'll just change it to `INFO`, that's on by default for tests.

kentonv:
SGTM, sorry I should have thought of that earlier.

c++/src/kj/compat/http.c++
```diff
@@ -5528,6 +5528,8 @@ kj::Promise<void> HttpServerErrorHandler::handleApplicationError(
   }
 
   KJ_IF_MAYBE(r, response) {
+    KJ_LOG(INFO, "Internal error serving HTTP response", exception);
```
kentonv, line 5531:
```suggestion
    KJ_LOG(INFO, "threw exception while serving HTTP response", exception);
```

PR #1387, opened 2021-11-20
Lie about the true size of the fiber stack on M1 Macs

### LINKED ISSUE #1386: EXC_BAD_ACCESS in Async/LargeTaskSetDestruction on M1 Arm Mac
Environment: M1 Macbook
 (`uname -a`: `Darwin zacharys-mbp.lan 21.1.0 Darwin Kernel Version 21.1.0: Wed Oct 13 17:33:01 PDT 2021; root:xnu-8019.41.5~1/RELEASE_ARM64_T6000 arm64`)
Build Configuration: `./configure --disable-shared CFLAGS='-g' CXXFLAGS='-g'`

Backtrace:
```
[ TEST ] kj/async-test.c++:740: legacy test: Async/LargeTaskSetDestruction
Process 61930 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=2, address=0x103ad4000)
    frame #0: 0x000000019184be00 libsystem_platform.dylib`__bzero + 32
libsystem_platform.dylib`__bzero:
->  0x19184be00 <+32>: stnp   q0, q0, [x0]
    0x19184be04 <+36>: stnp   q0, q0, [x0, #0x20]
    0x19184be08 <+40>: add    x3, x0, #0x40             ; =0x40
    0x19184be0c <+44>: and    x3, x3, #0xffffffffffffffc0
Target 0: (capnp-test) stopped.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=2, address=0x103ad4000)
  * frame #0: 0x000000019184be00 libsystem_platform.dylib`__bzero + 32
    frame #1: 0x000000019184ecd0 libsystem_platform.dylib`makecontext + 72
    frame #2: 0x0000000100a7bdd0 capnp-test`kj::_::FiberStack::FiberStack(this=0x000060000020c000, stackSizeParam=204800) at async.c++:1467:3
    frame #3: 0x0000000100a7c1c4 capnp-test`kj::_::FiberStack::FiberStack(this=0x000060000020c000, stackSizeParam=204800) at async.c++:1448:1
    frame #4: 0x0000000100a7c7e4 capnp-test`kj::Own<kj::_::FiberStack> kj::heap<kj::_::FiberStack, unsigned long&>(params=0x000000016fdfc880) at memory.h:459:21
    frame #5: 0x0000000100a7c6b8 capnp-test`kj::_::FiberBase::FiberBase(this=0x0000000104804110, stackSize=204800, result=0x0000000104804190, location=(fileName = "src/kj/async-test.c++", function = "run", lineNumber = 757, columnNumber = 3)) at async.c++:1507:47
    frame #6: 0x000000010032fa74 capnp-test`kj::_::Fiber<kj::(anonymous namespace)::TestCase740::run()::$_58>::Fiber(this=0x0000000104804110, stackSize=204800, func=0x000000016fdfcaff, location=<unavailable>)::TestCase740::run()::$_58&&, kj::SourceLocation) at async-inl.h:1045:9
    frame #7: 0x000000010032f9c0 capnp-test`kj::_::Fiber<kj::(anonymous namespace)::TestCase740::run()::$_58>::Fiber(this=0x0000000104804110, stackSize=204800, func=0x000000016fdfcaff, location=<unavailable>)::TestCase740::run()::$_58&&, kj::SourceLocation) at async-inl.h:1045:75
    frame #8: 0x000000010032f7e4 capnp-test`kj::Own<kj::_::Fiber<kj::(anonymous namespace)::TestCase740::run()::$_58> > kj::heap<kj::_::Fiber<kj::(anonymous namespace)::TestCase740::run(params=0x000000016fdfca90, params=0x000000016fdfcaff, params=0x000000016fdfcae0)::$_58>, unsigned long&, kj::(anonymous namespace)::TestCase740::run()::$_58, kj::SourceLocation&>(unsigned long&, kj::(anonymous namespace)::TestCase740::run()::$_58&&, kj::SourceLocation&) at memory.h:459:21
    frame #9: 0x000000010032f66c capnp-test`decltype(stackSize=204800, func=0x000000016fdfcaff, location=(fileName = "src/kj/async-test.c++", function = "run", lineNumber = 757, columnNumber = 3))::TestCase740::run()::$_58, kj::WaitScope&>::Type*)(nullptr), false)) kj::startFiber<kj::(anonymous namespace)::TestCase740::run()::$_58>(unsigned long, kj::(anonymous namespace)::TestCase740::run()::$_58&&, kj::SourceLocation) at async-inl.h:1270:36
    frame #10: 0x000000010032f57c capnp-test`kj::(anonymous namespace)::TestCase740::run(this=0x0000000100f764e0) at async-test.c++:757:3
    frame #11: 0x0000000100b3a498 capnp-test`kj::TestRunner::run(this=0x000000016fdfd1a8)::'lambda'()::operator()() const at test.c++:300:11
    frame #12: 0x0000000100b3a44c capnp-test`kj::_::RunnableImpl<kj::TestRunner::run()::'lambda'()>::run(this=0x000000016fdfd1a0) at exception.h:319:5
    frame #13: 0x0000000100b5aef4 capnp-test`kj::_::runCatchingExceptions(runnable=0x000000016fdfd1a0) at exception.c++:1327:14
    frame #14: 0x0000000100b39778 capnp-test`kj::Maybe<kj::Exception> kj::runCatchingExceptions<kj::TestRunner::run(func=0x000000016fdfd378)::'lambda'()>(kj::TestRunner::run()::'lambda'()&&) at exception.h:332:10
    frame #15: 0x0000000100b390c0 capnp-test`kj::TestRunner::run(this=0x000000016fdff570) at test.c++:300:11
    frame #16: 0x0000000100b38dc4 capnp-test`auto kj::TestRunner::getMain(this=0x00006000002012f0, s=0x000000016fdff570)::'lambda3'(auto&, auto&&...)::operator()<kj::TestRunner>(auto&, auto&&...) at test.c++:208:27
    frame #17: 0x0000000100b38d98 capnp-test`auto kj::_::BoundMethod<kj::TestRunner&, kj::TestRunner::getMain()::'lambda3'(auto&, auto&&...), kj::TestRunner::getMain()::'lambda4'(auto&, auto&&...)>::operator(this=0x00006000002012e8)<>() at function.h:263:12
    frame #18: 0x0000000100b38d68 capnp-test`kj::Function<kj::MainBuilder::Validity ()>::Impl<kj::_::BoundMethod<kj::TestRunner&, kj::TestRunner::getMain(this=0x00006000002012e0)::'lambda3'(auto&, auto&&...), kj::TestRunner::getMain()::'lambda4'(auto&, auto&&...)> >::operator()() at function.h:142:14
    frame #19: 0x0000000100ba9e84 capnp-test`kj::Function<kj::MainBuilder::Validity ()>::operator(this=0x0000600003b000c8)() at function.h:119:12
    frame #20: 0x0000000100ba8d70 capnp-test`kj::MainBuilder::MainImpl::operator(this=0x0000600000201308, programName=StringPtr @ 0x000000016fdfe8e0, params=ArrayPtr<const kj::StringPtr> @ 0x000000016fdfe8d0)(kj::StringPtr, kj::ArrayPtr<kj::StringPtr const>) at main.c++:623:5
    frame #21: 0x0000000100bb3634 capnp-test`kj::Function<void (kj::StringPtr, kj::ArrayPtr<kj::StringPtr const>)>::Impl<kj::MainBuilder::MainImpl>::operator(this=0x0000600000201300, params=StringPtr @ 0x000000016fdfe960, params=ArrayPtr<const kj::StringPtr> @ 0x000000016fdfe950)(kj::StringPtr, kj::ArrayPtr<kj::StringPtr const>) at function.h:142:14
    frame #22: 0x0000000100baa584 capnp-test`kj::Function<void (kj::StringPtr, kj::ArrayPtr<kj::StringPtr const>)>::operator(this=0x000000016fdff560, params=StringPtr @ 0x000000016fdfe9e0, params=ArrayPtr<const kj::StringPtr> @ 0x000000016fdfe9d0)(kj::StringPtr, kj::ArrayPtr<kj::StringPtr const>) at function.h:119:12
    frame #23: 0x0000000100baefdc capnp-test`kj::runMainAndExit(this=0x000000016fdfefd8)>&&, int, char**)::$_0::operator()() const at main.c++:228:5
    frame #24: 0x0000000100baef6c capnp-test`kj::_::RunnableImpl<kj::runMainAndExit(kj::ProcessContext&, kj::Function<void (kj::StringPtr, kj::ArrayPtr<kj::StringPtr const>)>&&, int, char**)::$_0>::run(this=0x000000016fdfefd0) at exception.h:319:5
    frame #25: 0x0000000100b5aef4 capnp-test`kj::_::runCatchingExceptions(runnable=0x000000016fdfefd0) at exception.c++:1327:14
    frame #26: 0x0000000100ba53f4 capnp-test`kj::Maybe<kj::Exception> kj::runCatchingExceptions<kj::runMainAndExit(func=0x000000016fdff0c8)>&&, int, char**)::$_0>(kj::runMainAndExit(kj::ProcessContext&, kj::Function<void (kj::StringPtr, kj::ArrayPtr<kj::StringPtr const>)>&&, int, char**)::$_0&&) at exception.h:332:10
    frame #27: 0x0000000100ba51e4 capnp-test`kj::runMainAndExit(context=0x000000016fdff590, func=0x000000016fdff560, argc=1, argv=0x000000016fdff738)>&&, int, char**) at main.c++:228:5
    frame #28: 0x0000000100b3740c capnp-test`main(argc=1, argv=0x000000016fdff738) at test.c++:363:1
    frame #29: 0x00000001038c90f4 dyld`start + 520
```

Thank you for all your work on capnp!

Dr-Emann:
`makecontext` on the new M1 macs calls `bzero` on the entire stack configured in the `ucontext_t`. Because we set up a guard page, this was faulting. Instead, only pass the usable stack, leaving the guard page below the configured stack start.

Fixes #1386

### DIFF
--- a/c++/src/kj/async.c++
+++ b/c++/src/kj/async.c++
@@ -1337,9 +1337,8 @@ struct FiberStack::Impl {
 #ifndef MAP_STACK
 #define MAP_STACK 0
 #endif
-
     size_t pageSize = getPageSize();
-    size_t allocSize = stackSize + pageSize;  // size plus guard page
+    size_t allocSize = stackSize + pageSize;  // size plus guard page and impl
 
     // Allocate virtual address space for the stack but make it inaccessible initially.
     // TODO(someday): Does it make sense to use MAP_GROWSDOWN on Linux? It's a kind of bizarre flag
@@ -1365,8 +1364,18 @@ struct FiberStack::Impl {
     // Note: mmap() allocates zero'd pages so we don't have to memset() anything here.
 
     KJ_SYSCALL(getcontext(context));
-    context->uc_stack.ss_size = stackSize - sizeof(Impl);
+#if __APPLE__ && __aarch64__
+    // Per issue #1386, apple on arm64 zeros the entire configured stack.
+    // For performance reasons, we rely on being able to allocate larger stacks
+    // than needed, and allow them to be lazily paged in. Instead, we lie:
+    // we allocate the full size, but tell the ucontext the stack is the last
+    // page only.
+    context->uc_stack.ss_size = pageSize - sizeof(Impl);
+    context->uc_stack.ss_sp = reinterpret_cast<byte*>(stack) + stackSize - min(pageSize, stackSize);
+#else
+    context->uc_stack.ss_size = stackSize;
     context->uc_stack.ss_sp = stack;
+#endif
     context->uc_stack.ss_flags = 0;
     // We don't use uc_link since our fiber start routine runs forever in a loop to allow for
     // reuse. When we're done with the fiber, we just destroy it, without switching to it's
@@ -1387,7 +1396,7 @@ struct FiberStack::Impl {
 #ifndef _SC_PAGESIZE
 #define _SC_PAGESIZE _SC_PAGE_SIZE
 #endif
-    static size_t result = sysconf(_SC_PAGE_SIZE);
+    static size_t result = sysconf(_SC_PAGESIZE);
     return result;
   }
 };
--- a/c++/src/kj/async.c++
+++ b/c++/src/kj/async.c++
@@ -1345,28 +1345,28 @@ struct FiberStack::Impl {
     // TODO(someday): Does it make sense to use MAP_GROWSDOWN on Linux? It's a kind of bizarre flag
     //   that causes the mapping to automatically allocate extra pages (beyond the range specified)
     //   until it hits something...
-    void* stack = mmap(nullptr, allocSize, PROT_NONE,
+    void* stackMapping = mmap(nullptr, allocSize, PROT_NONE,
         MAP_ANONYMOUS | MAP_PRIVATE | MAP_STACK, -1, 0);
-    if (stack == MAP_FAILED) {
+    if (stackMapping == MAP_FAILED) {
       KJ_FAIL_SYSCALL("mmap(new stack)", errno);
     }
     KJ_ON_SCOPE_FAILURE({
-      KJ_SYSCALL(munmap(stack, allocSize)) { break; }
+      KJ_SYSCALL(munmap(stackMapping, allocSize)) { break; }
     });
 
+    void* stack = reinterpret_cast<byte*>(stackMapping) + pageSize;
     // Now mark everything except the guard page as read-write. We assume the stack grows down, so
     // the guard page is at the beginning. No modern architecture uses stacks that grow up.
-    KJ_SYSCALL(mprotect(reinterpret_cast<byte*>(stack) + pageSize,
-                        stackSize, PROT_READ | PROT_WRITE));
+    KJ_SYSCALL(mprotect(stack, stackSize, PROT_READ | PROT_WRITE));
 
     // Stick `Impl` at the top of the stack.
-    Impl* impl = (reinterpret_cast<Impl*>(reinterpret_cast<byte*>(stack) + allocSize) - 1);
+    Impl* impl = (reinterpret_cast<Impl*>(reinterpret_cast<byte*>(stack) + stackSize) - 1);
 
     // Note: mmap() allocates zero'd pages so we don't have to memset() anything here.
 
     KJ_SYSCALL(getcontext(context));
-    context->uc_stack.ss_size = allocSize - sizeof(Impl);
-    context->uc_stack.ss_sp = reinterpret_cast<char*>(stack);
+    context->uc_stack.ss_size = stackSize - sizeof(Impl);
+    context->uc_stack.ss_sp = stack;
     context->uc_stack.ss_flags = 0;
     // We don't use uc_link since our fiber start routine runs forever in a loop to allow for
     // reuse. When we're done with the fiber, we just destroy it, without switching to it's
--- a/c++/src/kj/async.c++
+++ b/c++/src/kj/async.c++
@@ -1374,10 +1374,10 @@ struct FiberStack::Impl {
     // page only. This appears to work as no particular bounds checks or
     // anything are set up based on what we say here.
     context->uc_stack.ss_size = pageSize - sizeof(Impl);
-    context->uc_stack.ss_sp = reinterpret_cast<byte*>(stack) + stackSize - min(pageSize, stackSize);
+    context->uc_stack.ss_sp = reinterpret_cast<char*>(stack) + stackSize - min(pageSize, stackSize);
 #else
-    context->uc_stack.ss_size = stackSize;
-    context->uc_stack.ss_sp = stack;
+    context->uc_stack.ss_size = stackSize - sizeof(Impl);
+    context->uc_stack.ss_sp = reinterpret_cast<char*>(stack);
 #endif
     context->uc_stack.ss_flags = 0;
     // We don't use uc_link since our fiber start routine runs forever in a loop to allow for
--- a/c++/src/kj/async.c++
+++ b/c++/src/kj/async.c++
@@ -1366,10 +1366,13 @@ struct FiberStack::Impl {
     KJ_SYSCALL(getcontext(context));
 #if __APPLE__ && __aarch64__
     // Per issue #1386, apple on arm64 zeros the entire configured stack.
-    // For performance reasons, we rely on being able to allocate larger stacks
-    // than needed, and allow them to be lazily paged in. Instead, we lie:
+    // But this is redundant, since we just allocated the stack with mmap() which
+    // returns zero'd pages. Re-zeroing is both slow and results in prematurely
+    // allocating pages we may not need -- it's normal for stacks to rely heavily
+    // on lazy page allocation to avoid wasting memory. Instead, we lie:
     // we allocate the full size, but tell the ucontext the stack is the last
-    // page only.
+    // page only. This appears to work as no particular bounds checks or
+    // anything are set up based on what we say here.
     context->uc_stack.ss_size = pageSize - sizeof(Impl);
     context->uc_stack.ss_sp = reinterpret_cast<byte*>(stack) + stackSize - min(pageSize, stackSize);
 #else
--- a/c++/src/kj/async.c++
+++ b/c++/src/kj/async.c++
@@ -1373,7 +1373,7 @@ struct FiberStack::Impl {
     // we allocate the full size, but tell the ucontext the stack is the last
     // page only. This appears to work as no particular bounds checks or
     // anything are set up based on what we say here.
-    context->uc_stack.ss_size = pageSize - sizeof(Impl);
+    context->uc_stack.ss_size = min(pageSize, stackSize) - sizeof(Impl);
     context->uc_stack.ss_sp = reinterpret_cast<char*>(stack) + stackSize - min(pageSize, stackSize);
 #else
     context->uc_stack.ss_size = stackSize - sizeof(Impl);


Dr-Emann:
I disassembled the `makecontext` impl: Even on the m1 mac, the x64 version of the function does not perform the `bzero`, so it's just on arm mode (maybe also IOS?).

`make check` passes with this patch on an M1 mac, and it feels maybe more correct anyway.

kentonv:
Yikes, that bzero() is very bad even if it doesn't fault. The memory is freshly-allocated using mmap(), so it's already zero'd. Moreover, normally these pages would be allocated lazily on first access, so that the stack only uses as much physical memory as is actually needed for its high point. If it gets bzero()ed up front, that will force immediate physical allocation of the entire stack space. We usually rely on stack space being lazily allocated so that we can allocate huge stacks and not worry that we're wasting memory (only wasting virtual space, which on 64-bit is plentiful).

Is there any way we can prevent the system from calling bzero() here?

I'm very surprised it does this as it seems disastrous for performance.

Dr-Emann:
No, one of the first things `makecontext` does is to unconditionally bzero the stack (after ensuring: the passed context is non-null, there are less than 0x100001 arguments, and the stack pointer is non-null.)

The `ucontext.h` file does have a comment that "These routines are DEPRECATED and should not be used." and are annotated with `__API_DEPRECATED("No longer supported", macos(10.5, 10.6))`.

I know macos does make use of memory compression, so the actual memory use may not be a terrible problem, but the performance of creating will probably be worse.

Dr-Emann:
It would be horrific, and terrible, and almost certainly a bad idea, but it _looks_ from my inexperienced reading the disassembled function that we _could_ do something crazy like point `makecontext` at the last page of our stack, and modify it to be larger after calling `makecontext`. I think it would be better to just say fibers may be slow to create on arm macos, and/or deprecated because macos won't support the needed calls, and might someday break even worse and there's nothing we can do.

kentonv:
Yeah I was wondering, is there actually anything preventing you from lying about the stack size? Does it actually do bounds checks during execution?

I think there are two reasonable options:
1. Lie about the stack size to avoid the bzero().
2. Declare that fibers aren't supported on M1 MacOS. They're already missing on several other platforms anyway.

Dr-Emann:
I __think__ based on my reading of the assembly of `setcontext` and `makecontext`, it seems safe, and experimentally, the following program works, and seems to jump correctly between the new stack and the original stack (and is much faster with `LIE_ABOUT_SIZE=1`):

```c
// Compiled with -O0 so the writes in `run` aren't optimized away
#include <stdio.h>
#include <sys/mman.h>
#define _XOPEN_SOURCE
#include <stdlib.h>
#include <string.h>
#include <ucontext.h>
#include <unistd.h>

// 100 GiB should take some time to bzero
#define STACK_SIZE (100UL * 1024 * 1024 * 1024)

#define LIE_ABOUT_SIZE 1
#define FIXUP_LIE 0

size_t page_size;

int did_run = 0;

void run(int x, int y)
{
    char z[STACK_SIZE - 2048];
    // Use the first and last byte
    z[0] = 99;
    z[sizeof(z) - 1] = 99;
    printf("x: %d, y: %d, z: %d\n", x, y, z[0]);
    did_run = 1;
}

int main()
{
    page_size = sysconf(_SC_PAGE_SIZE);
    ucontext_t ctx;
    getcontext(&ctx);
    void *stack_mapping = mmap(NULL, page_size + STACK_SIZE, PROT_NONE,
                       MAP_ANONYMOUS | MAP_PRIVATE, -1, 0);
    void *stack = (char *) stack_mapping + page_size;
    mprotect(stack, STACK_SIZE, PROT_READ | PROT_WRITE);

    ctx.uc_stack.ss_sp = stack;
    ctx.uc_stack.ss_size = STACK_SIZE;
#if LIE_ABOUT_SIZE
    ctx.uc_stack.ss_sp = stack + STACK_SIZE - page_size;
    ctx.uc_stack.ss_size = page_size;
#endif
    ctx.uc_stack.ss_flags = 0;
    ucontext_t next_ctx;
    getcontext(&next_ctx);
    if (did_run) {
        exit(22);
    }

    ctx.uc_link = &next_ctx;

    makecontext(&ctx, run, 2, 10, 20);

#if LIE_ABOUT_SIZE && FIXUP_LIE
    ctx.uc_stack.ss_sp = stack;
    ctx.uc_stack.ss_size = STACK_SIZE;
#endif

    setcontext(&ctx);

    return 0;
}
```

kentonv:
I am in favor of lying then.

c++/src/kj/async.c++
```diff
@@ -1337,36 +1337,45 @@ struct FiberStack::Impl {
...
 
     KJ_SYSCALL(getcontext(context));
-    context->uc_stack.ss_size = allocSize - sizeof(Impl);
-    context->uc_stack.ss_sp = reinterpret_cast<char*>(stack);
+#if __APPLE__ && __aarch64__
+    // Per issue #1386, apple on arm64 zeros the entire configured stack.
+    // For performance reasons, we rely on being able to allocate larger stacks
+    // than needed, and allow them to be lazily paged in. Instead, we lie:
+    // we allocate the full size, but tell the ucontext the stack is the last
+    // page only.
```
kentonv, line 1372:
```suggestion
    // Per issue #1386, apple on arm64 zeros the entire configured stack.
    // But this is redundant, since we just allocated the stack with mmap() which
    // returns zero'd pages. Re-zeroing is both slow and results in prematurely
    // allocating pages we may not need -- it's normal for stacks to rely heavily
    // on lazy page allocation to avoid wasting memory. Instead, we lie:
    // we allocate the full size, but tell the ucontext the stack is the last
    // page only. This appears to work as no particular bounds checks or
    // anything are set up based on what we say here.
```

c++/src/kj/async.c++
```diff
@@ -1337,36 +1337,45 @@ struct FiberStack::Impl {
...
+    context->uc_stack.ss_size = pageSize - sizeof(Impl);
+    context->uc_stack.ss_sp = reinterpret_cast<byte*>(stack) + stackSize - min(pageSize, stackSize);
+#else
+    context->uc_stack.ss_size = stackSize;
```
kentonv, line 1376: Did we lose the `- sizeof(Impl)` here?

c++/src/kj/async.c++
```diff
@@ -1337,36 +1337,45 @@ struct FiberStack::Impl {
...
+    context->uc_stack.ss_sp = stack;
```
kentonv, line 1377: Looks like you removed the `reinterpret_cast<char*>` but it is actually needed on some platforms.

c++/src/kj/async.c++
Dr-Emann: I kept going back and forth on adding `sizeof(Impl)` to the size of the allocation, rather than decreasing the usable stack size by `sizeof(Impl)`. Looks like I missed a case of going back.

c++/src/kj/async.c++
```diff
@@ -1337,36 +1337,48 @@ struct FiberStack::Impl {
...
+#if __APPLE__ && __aarch64__
+    // Per issue #1386, apple on arm64 zeros the entire configured stack.
+    // But this is redundant, since we just allocated the stack with mmap() which
+    // returns zero'd pages. Re-zeroing is both slow and results in prematurely
+    // allocating pages we may not need -- it's normal for stacks to rely heavily
+    // on lazy page allocation to avoid wasting memory. Instead, we lie:
+    // we allocate the full size, but tell the ucontext the stack is the last
+    // page only. This appears to work as no particular bounds checks or
+    // anything are set up based on what we say here.
+    context->uc_stack.ss_size = pageSize - sizeof(Impl);
```
kentonv, line 1376: I think this needs to be consistent with the next line?

```suggestion
    context->uc_stack.ss_size = min(pageSize, stackSize) - sizeof(Impl);
```

kentonv:
Thanks! Sorry this got stuck under a pile in my inbox for a while. :(

PR #1390, opened 2021-11-25
Introduce ConnectionReceiver implementation that aggregates from multiple underlying receivers.

kentonv:
Use this to implement binding to a hostname that resolves to multiple addresses. This is common for `localhost`, which may return both ipv4 and ipv6 addresses.

(I didn't actually create this class for this purpose, but it was super-easy to use it to resolve this very old TODO...)

### DIFF
--- a/c++/src/kj/async-io-test.c++
+++ b/c++/src/kj/async-io-test.c++
@@ -2918,5 +2918,78 @@ KJ_TEST("import socket FD that's already broken") {
 #endif  // !__CYGWIN__
 #endif  // !_WIN32
 
+KJ_TEST("AggregateConnectionReceiver") {
+  EventLoop loop;
+  WaitScope ws(loop);
+
+  auto pipe1 = newCapabilityPipe();
+  auto pipe2 = newCapabilityPipe();
+
+  auto receiversBuilder = kj::heapArrayBuilder<Own<ConnectionReceiver>>(2);
+  receiversBuilder.add(kj::heap<CapabilityStreamConnectionReceiver>(*pipe1.ends[0]));
+  receiversBuilder.add(kj::heap<CapabilityStreamConnectionReceiver>(*pipe2.ends[0]));
+
+  auto aggregate = newAggregateConnectionReceiver(receiversBuilder.finish());
+
+  CapabilityStreamNetworkAddress connector1(nullptr, *pipe1.ends[1]);
+  CapabilityStreamNetworkAddress connector2(nullptr, *pipe2.ends[1]);
+
+  auto connectAndWrite = [&](NetworkAddress& addr, kj::StringPtr text) {
+    return addr.connect()
+        .then([text](Own<AsyncIoStream> stream) {
+      auto promise = stream->write(text.begin(), text.size());
+      return promise.attach(kj::mv(stream));
+    }).eagerlyEvaluate([](kj::Exception&& e) {
+      KJ_LOG(ERROR, e);
+    });
+  };
+
+  auto acceptAndRead = [&](ConnectionReceiver& socket, kj::StringPtr expected) {
+    return socket
+        .accept().then([](Own<AsyncIoStream> stream) {
+      auto promise = stream->readAllText();
+      return promise.attach(kj::mv(stream));
+    }).then([expected](kj::String actual) {
+      KJ_EXPECT(actual == expected);
+    }).eagerlyEvaluate([](kj::Exception&& e) {
+      KJ_LOG(ERROR, e);
+    });
+  };
+
+  auto connectPromise1 = connectAndWrite(connector1, "foo");
+  KJ_EXPECT(!connectPromise1.poll(ws));
+  auto connectPromise2 = connectAndWrite(connector2, "bar");
+  KJ_EXPECT(!connectPromise2.poll(ws));
+
+  acceptAndRead(*aggregate, "foo").wait(ws);
+
+  auto connectPromise3 = connectAndWrite(connector1, "baz");
+  KJ_EXPECT(!connectPromise3.poll(ws));
+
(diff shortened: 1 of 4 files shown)


vlovich (approved):
Is it worthwhile to do a bit of a stress test to make sure that concurrent inbound requests are handled correctly & never lost? The current test is a bit too "happy path" for my taste although the code seems relatively straightforward enough (at least after reading it a few times to grok what was happening) that it may not be worth the extra test code. Up to you.

c++/src/kj/async-io.c++
```diff
@@ -2764,6 +2764,121 @@ String CapabilityStreamNetworkAddress::toString() {
...
+      }
+    }
+
+    PromiseFulfiller<AuthenticatedStream>& fulfiller;
+    AggregateConnectionReceiver& parent;
+    ListLink<Waiter> link;
+  };
+
+  List<Waiter, &Waiter::link> waiters;
+  std::deque<Promise<AuthenticatedStream>> backlog;
```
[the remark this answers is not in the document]
kentonv: Eh... O(log n) is basically O(1).

c++/src/kj/async-io.c++
```diff
@@ -2764,6 +2764,121 @@ String CapabilityStreamNetworkAddress::toString() {
...
+      auto result = kj::mv(backlog.front());
+      backlog.pop_front();
+      return result;
+    }
+  }
+
+  uint getPort() override {
+    return receivers[0]->getPort();
+  }
+  void getsockopt(int level, int option, void* value, uint* length) override {
```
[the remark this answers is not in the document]
kentonv: It may be used. If someone listens on `localhost:1234`, `localhost` will resolve to both `127.0.0.1` and `::1`. In the past, we'd only actually listen on the first address and ignore the second, but this PR makes us listen on both, using `AggregateConnectionReceiver`. If someone then calls `getsockopt()` on such a socket, what should happen? The implementation here maintains the old behavior.

c++/src/kj/async-io-test.c++
```diff
@@ -2918,5 +2918,78 @@ KJ_TEST("import socket FD that's already broken") {
...
+
+  KJ_EXPECT(!acceptPromise1.poll(ws));
+  KJ_EXPECT(!acceptPromise2.poll(ws));
+  KJ_EXPECT(!acceptPromise3.poll(ws));
+
+  // Cancel one of the acceptors...
+  { auto drop = kj::mv(acceptPromise2); }
+
+  connectAndWrite(connector2, "qux").wait(ws);
+  connectAndWrite(connector1, "grault").wait(ws);
```
jclee, line 2988: I guess one thing that the test doesn't explicitly test the case where connections come in in the reverse order of the accept calls, although I assume the implementation should handle this fine.  But reversing these two lines would help test that?

c++/src/kj/async-io.c++
```diff
@@ -2764,6 +2764,121 @@ String CapabilityStreamNetworkAddress::toString() {
 
 // =======================================================================================
 
+namespace {
+
+class AggregateConnectionReceiver final: public ConnectionReceiver {
+public:
+  AggregateConnectionReceiver(Array<Own<ConnectionReceiver>> receiversParam)
```
jclee, line 2771: I guess if someone for some reason passed an empty array of receivers, the `accept` methods would still work correctly, in the sense that they'd return promises that never resolve.  But it looks like the couple uses of `receivers[0]` are only guarded by an `KJ_IREQUIRE`, so maybe we should require a non-empty array, either in those methods or in the constructor?

c++/src/kj/async-io-test.c++
```diff
@@ -2918,5 +2918,78 @@ KJ_TEST("import socket FD that's already broken") {
...
+
+  auto acceptAndRead = [&](ConnectionReceiver& socket, kj::StringPtr expected) {
+    return socket
+        .accept().then([](Own<AsyncIoStream> stream) {
+      auto promise = stream->readAllText();
+      return promise.attach(kj::mv(stream));
+    }).then([expected](kj::String actual) {
+      KJ_EXPECT(actual == expected);
+    }).eagerlyEvaluate([](kj::Exception&& e) {
+      KJ_LOG(ERROR, e);
```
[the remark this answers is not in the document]
kentonv: Logging at level `ERROR` during a test will cause the test to fail.

c++/src/kj/async-io-test.c++
```diff
@@ -2918,5 +2918,78 @@ KJ_TEST("import socket FD that's already broken") {
...
+    });
+  };
+
+  auto acceptAndRead = [&](ConnectionReceiver& socket, kj::StringPtr expected) {
+    return socket
+        .accept().then([](Own<AsyncIoStream> stream) {
+      auto promise = stream->readAllText();
+      return promise.attach(kj::mv(stream));
+    }).then([expected](kj::String actual) {
+      KJ_EXPECT(actual == expected);
```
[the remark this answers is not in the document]
kentonv: Yep.

c++/src/kj/async-io-test.c++
kentonv: Hmm, not sure what you mean. Connections are accepted in the order in which they are made, so if we reversed these two lines, the test should fail.

c++/src/kj/async-io.c++
kentonv: Yeah, we should probably KJ_REQUIRE that the array is non-empty...

c++/src/kj/async-io-test.c++
jclee: Oh, right...  I think I may have been thinking of reversing the connector order, which you're already doing here.

PR #1391, opened 2021-11-27
Update old repo url with new

Bambooin:


kentonv:
Thanks for doing this! Sorry for the delay.

I'm not sure I agree with the second commit. That's a change to the documentation. A human reader familiar with `-j` will be able to figure out that they can change the number. For a reader not familiar with `-j`, introducing `$(nproc)` adds additional cognitive load -- especially if their system does not have an `$(nproc)` command, in which case the command will fail and the user may not understand what to do.

Do you mind removing the second commit so we can merge the first one?

Bambooin:
I have removed the second commit, you can merge the first one

For the `$(nproc)`, I check it works on Unix and Linux before commit

https://github.com/osfans/trime/pull/594

We use capnproto in our project, and change to `$(nproc)` can speed build for us

jclee:
@Bambooin FYI, `nproc` was not a supported command on the Mac that I tried it on.  (I figured you'd want to be aware of this, since your project's readme says it supports building on Mac.)

Bambooin:
Checked that `nproc` doesn't work in macOS, it's my bad

We build Trime in macOS with CMake, and `script/dependency.sh` file is used only for ci

kentonv:
Thanks!

PR #1405, opened 2022-01-11
HTTP concurrency limiter: Avoid unnecessary recursion

jclee:
...when many pending connections are canceled at once

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -4266,12 +4266,15 @@ private:
   };
 
   void serviceQueue() {
-    if (concurrentRequests >= maxConcurrentRequests) { return; }
-    if (pendingRequests.empty()) { return; }
-
-    auto fulfiller = kj::mv(pendingRequests.front());
-    pendingRequests.pop();
-    fulfiller->fulfill(ConnectionCounter(*this));
+    while (concurrentRequests < maxConcurrentRequests && !pendingRequests.empty()) {
+      auto fulfiller = kj::mv(pendingRequests.front());
+      pendingRequests.pop();
+      // ConnectionCounter's destructor calls this function, so we can avoid unnecessary recursion
+      // if we only create a ConnectionCounter when we find a waiting fulfiller.
+      if (fulfiller->isWaiting()) {
+        fulfiller->fulfill(ConnectionCounter(*this));
+      }
+    }
   }
 
   void fireCountChanged() {


a-robinson (approved):
LGTM, although I have no knowledge of how often promises in `pendingRequests` are no longer waiting.

c++/src/kj/compat/http.c++
```diff
@@ -4266,12 +4266,15 @@ private:
...
-    if (pendingRequests.empty()) { return; }
-
-    auto fulfiller = kj::mv(pendingRequests.front());
-    pendingRequests.pop();
-    fulfiller->fulfill(ConnectionCounter(*this));
+    while (concurrentRequests < maxConcurrentRequests && !pendingRequests.empty()) {
+      auto fulfiller = kj::mv(pendingRequests.front());
+      pendingRequests.pop();
+      // ConnectionCounter's destructor calls this function, so we can avoid unnecessary recursion
+      // if we only create a ConnectionCounter when we find a waiting fulfiller.
```
kentonv, line 4273: Is this a change in behavior in the case that there are multiple pending requests that are waiting? Previously only one waiting request would be fulfilled, but now they all are?

jclee: I think it actually ends up being the same behavior... I think the code maintains an invariant that if `!pendingRequests.empty()` then `concurrentRequests` is either `maxConcurrentRequests` or `maxConcurrentRequests - 1`.  Due to the while loop condition, `concurrentRequests` should be `maxConcurrentRequests - 1` here, so calling `fulfiller->fulfill(ConnectionCounter(*this));` on a waiting fulfiller should increment `concurrentRequests` and exit the while loop -- assuming that `fulfill()` never runs synchronously but instead is always asynchronously queued for later execution in the event loop, and that calling `fulfill()` on a waiting fulfiller will never synchronously destruct its payload.

Admittedly, this is non-obvious...  I did have a `return` after the `fulfill()` in an earlier version of the code, but removed it because it had no effect.  I'll try adding some `KJ_ASSERT`s to demonstrate the invariants.  I think it might also be a little more clear if I inline `serviceQueue()` in the `ConnectionCounter` destructor, since that's the only place it is called.

kentonv: Ah I get it now. Actually I think I like the code as-is, no need for extra asserts... if for some reason `concurrentRequests` had been decremented by two, then you actually would want to fulfill two of the fulfillers.

PR #1408, opened 2022-01-13
Build fails with -O0 on Ubuntu 20.04

JonasVautherin:
I don't really understand why it passed the current CI, but with the latest PR in capnproto, both the tests and samples fail to build for me. It works from commit 8009588ff84cbdf233f6d23d1d507462b050b427.

The error I get is:

```
[ 44%] Linking CXX executable kj-heavy-tests
/usr/bin/ld: libkj-async.a(async-io-unix.c++.o): in function `kj::(anonymous namespace)::AsyncStreamFd::splicePumpLoop(kj::(anonymous namespace)::AsyncStreamFd&, int, int, unsigned long, unsigned long, unsigned long)::{lambda()#3}::operator()() const':
async-io-unix.c++:(.text+0x1f1d): undefined reference to `kj::(anonymous namespace)::AsyncStreamFd::MAX_SPLICE_LEN'
```

I don't think that it is necessary to add this github workflow, because capnproto already should cover that in the CI. I just thought it would be easier to report like this (and I don't understand why the current CI did not catch this). As can be seen, all I do is configuring and building capnproto and the samples.

EDIT: actually I realize that the CI is using the Makefile, so maybe it would make sense to add this workflow to check that it builds with CMake, right?

### DIFF
--- a/c++/src/kj/async-io-unix.c++
+++ b/c++/src/kj/async-io-unix.c++
@@ -804,6 +804,10 @@ private:
   }
 };
 
+#if __linux__
+constexpr size_t AsyncStreamFd::MAX_SPLICE_LEN;
+#endif  // __linux__
+
 // =======================================================================================
 
 class SocketAddress {


jclee:
Wonder if this might be due to different versions of C++ having different treatment for whether a "static constexpr" member needs both a declaration and definition? https://stackoverflow.com/questions/8016780/undefined-reference-to-static-constexpr-char

Might also be influenced by the compiler version or its optimization level...  looks like the ubuntu environment in your build is using gcc 9.3.0, not sure what the optimization level or the C++ language level is.

kentonv:
Hmm, the existing CI script already does build samples, including both with autotools and cmake:

https://github.com/capnproto/capnproto/blob/ab12d40915889893cba9c0ee8e7ba43bf87361dc/super-test.sh#L456-L468

The CI runs those tests for old and new versions of GCC and clang.

So I don't understand why the new test fails. Could it be that different optimization levels are being applied?

jclee:
I suppose the cmake invocation in `super-test.sh` would be influenced by the `CXXFLAGS` value it sets earlier, which does "-O2": https://github.com/capnproto/capnproto/blob/ab12d40915889893cba9c0ee8e7ba43bf87361dc/super-test.sh#L353

Not sure what the cmake invocation in the new CI stage would use...  gcc docs say the default is `-o0`, which certainly sounds like it might aggressively avoid inlining the member access, making the link error more likely.

JonasVautherin:
I tried different optimization levels on an Ubuntu 20.04 to make sure, and it seems like you are right:

* Building with `-O3` works
* Building with `-O2` works
* Building with `-O1` works
* Building with `-O0` __fails__

kentonv:
Yeah I believe `MAX_SPLICE_LEN` needs an out-of-line definition. Happy to accept a PR or otherwise I'll get to it... sometime.

JonasVautherin:
I tried inlining the variable, keeping my CI workflow just to check if that works (it works locally).

c++/src/kj/async-io-unix.c++
```diff
@@ -804,6 +804,8 @@ private:
   }
 };
 
+constexpr size_t kj::AsyncStreamFd::MAX_SPLICE_LEN;
```
kentonv, line 807: You need `#if __linux__` around this to match the declaration.

JonasVautherin: ooooh that explains things :facepalm:

kentonv:
Thanks!

PR #1409, opened 2022-01-15
Add support for musl

### LINKED ISSUE #1167: Failing to build on musl libc due to missing getcontext/setcontext

JonasVautherin:
Add support for musl systems both for cmake and makefile (related to #1167).

* For makefile, it searches for the `getcontext` symbol in `libucontext`, and adds `-lucontext` if it is found.
* For cmake, it searches for `libucontext` using `pkg-config`, and links it if found.

Note: this PR makes the assumption that a system that has `libucontext` needs it for the `getcontext`, `makecontext` and `setcontext` symbols.

The CI fails for two tests, I believe because docker on the github workflow does not support ipv6 (see https://github.com/actions/virtual-environments/issues/668):

```
2022-01-15T01:25:57.4242371Z kj/async-io-unix.c++:865: failed: ::bind(sockfd, &addr.generic, addrlen): Address not available; toString() = [::1]:0
2022-01-15T01:25:57.4242816Z [ FAIL ] capnp/ez-rpc-test.c++:32: legacy test: EzRpc/Basic (354 μs)
2022-01-15T01:25:57.4249502Z kj/async-io-unix.c++:865: failed: ::bind(sockfd, &addr.generic, addrlen): Address not available; toString() = [::1]:0
2022-01-15T01:25:57.4250008Z [ FAIL ] capnp/ez-rpc-test.c++:49: legacy test: EzRpc/DeprecatedNames (650 μs)
```

One alternative would be to not add the alpine build to the CI (which is a bit sad), another would be to add some kind of `--no-ipv6` flag to `super-test.sh`, but I'm not sure how much work that requires.

Resolves #1167.

### DIFF
--- a/c++/CMakeLists.txt
+++ b/c++/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.4)
+cmake_minimum_required(VERSION 3.6)
 project("Cap'n Proto" CXX)
 set(VERSION 0.10-dev)
 
@@ -62,6 +62,50 @@ elseif (WITH_OPENSSL STREQUAL "AUTO")
   endif()
 elseif (WITH_OPENSSL)
   find_package(OpenSSL REQUIRED COMPONENTS Crypto SSL)
+endif()
+
+set(WITH_FIBERS "AUTO" CACHE STRING
+  "Whether or not to build libkj-async with fibers")
+# define list of values GUI will offer for the variable
+set_property(CACHE WITH_FIBERS PROPERTY STRINGS AUTO ON OFF)
+
+# CapnProtoConfig.cmake.in needs this variable.
+set(_WITH_LIBUCONTEXT OFF)
+
+if (WITH_FIBERS OR WITH_FIBERS STREQUAL "AUTO")
+  set(_capnp_fibers_found OFF)
+  if (WIN32 OR CYGWIN)
+    set(_capnp_fibers_found ON)
+  else()
+    # Fibers need makecontext, setcontext, getcontext, swapcontext that may be in libc,
+    # or in libucontext (e.g. for musl).
+    # We assume that makecontext implies that the others are present.
+    include(CheckLibraryExists)
+    check_library_exists(c makecontext "" HAVE_UCONTEXT_LIBC)
+    if (HAVE_UCONTEXT_LIBC)
+      set(_capnp_fibers_found ON)
+    else()
+      # Try with libucontext
+      find_package(PkgConfig)
+      if (PKG_CONFIG_FOUND)
+        pkg_check_modules(libucontext IMPORTED_TARGET libucontext)
+        if (libucontext_FOUND)
+          set(_WITH_LIBUCONTEXT ON)
+          set(_capnp_fibers_found ON)
+        endif()
+      else()
+        set(_capnp_fibers_found OFF)
+      endif()
+    endif()
+  endif()
+
+  if (_capnp_fibers_found)
+    set(WITH_FIBERS ON)
+  elseif(WITH_FIBERS STREQUAL "AUTO")
+    set(WITH_FIBERS OFF)
+  else()
+    message(FATAL_ERROR "Missing 'makecontext', 'getcontext', 'setcontext' or 'swapcontext' symbol in libc and no libucontext found: KJ won't be able to build with fibers. Disable fibers (-DWITH_FIBERS=OFF).")
+  endif()
 endif()
 
 if(MSVC)
--- a/c++/configure.ac
+++ b/c++/configure.ac
@@ -32,6 +32,11 @@ AC_ARG_WITH([openssl],
     [build libkj-tls by linking against openssl @<:@default=check@:>@])],
   [],[with_openssl=check])
 
+AC_ARG_WITH([fibers],
+  [AS_HELP_STRING([--with-fibers],
+    [build libkj-async with fibers @<:@default=check@:>@])],
+  [],[with_fibers=check])
+
 AC_ARG_ENABLE([reflection], [
   AS_HELP_STRING([--disable-reflection], [
     compile Cap'n Proto in "lite mode", in which all reflection APIs (schema.h, dynamic.h, etc.)
@@ -195,8 +200,46 @@ AS_IF([test "$with_openssl" != no], [
 ])
 AM_CONDITIONAL([BUILD_KJ_TLS], [test "$with_openssl" != no])
 
-# CapnProtoConfig.cmake.in needs this variable.
-AC_SUBST(WITH_OPENSSL, $with_openssl)
+# Fibers need the symbols getcontext, setcontext, swapcontext and makecontext.
+# We assume that makecontext implies the rest.
+AS_IF([test "$with_fibers" != no], [
+  libc_supports_fibers=yes
+  AC_SEARCH_LIBS([makecontext], [], [], [
+    libc_supports_fibers=no
+  ])
+
+  AS_IF([test "$libc_supports_fibers" = yes], [
+      with_fibers=yes
+    ], [
+    # If getcontext does not exist in libc, try with libucontext
+    ucontext_supports_fibers=yes
+    AC_CHECK_LIB(ucontext, [makecontext], [], [
+      ucontext_supports_fibers=no
+    ])
+    AS_IF([test "$ucontext_supports_fibers" = yes], [
+      ASYNC_LIBS="$ASYNC_LIBS -lucontext"
+      with_fibers=yes
+    ], [
+      AS_IF([test "$with_fibers" = yes], [
+        AC_MSG_ERROR([Missing symbols required for fibers (makecontext, setcontext, ...). Disable fibers (--without-fibers) or install libucontext])
+      ], [
+        AC_MSG_WARN([could not find required symbols (makecontext, setcontext, ...) -- won't build with fibers])
+        with_fibers=no
+      ])
+    ])
+  ])
+])
+AS_IF([test "$with_fibers" = yes], [
+  CXXFLAGS="$CXXFLAGS -DKJ_USE_FIBERS"
+], [
+  CXXFLAGS="$CXXFLAGS -DKJ_USE_FIBERS=0"
+])
+
+# CapnProtoConfig.cmake.in needs these variables,
+# we force them to NO because we don't need the CMake dependency for them,
+# the dependencies are provided by the .pc files.
+AC_SUBST(WITH_OPENSSL, NO)
+AC_SUBST(_WITH_LIBUCONTEXT, NO)
 
 AM_CONDITIONAL([HAS_FUZZING_ENGINE], [test "x$LIB_FUZZING_ENGINE" != "x"])
 
--- a/c++/src/kj/CMakeLists.txt
+++ b/c++/src/kj/CMakeLists.txt
@@ -137,6 +137,15 @@ if(NOT CAPNP_LITE)
   add_library(kj-async ${kj-async_sources})
   add_library(CapnProto::kj-async ALIAS kj-async)
   target_link_libraries(kj-async PUBLIC kj)
+  if(WITH_FIBERS)
+    target_compile_definitions(kj-async PUBLIC KJ_USE_FIBERS)
+    if(_WITH_LIBUCONTEXT)
+      target_link_libraries(kj-async PUBLIC PkgConfig::libucontext)
+    endif()
+  else()
+    target_compile_definitions(kj-async PUBLIC KJ_USE_FIBERS=0)
+  endif()
+
   if(UNIX)
     # external clients of this library need to link to pthreads
     target_compile_options(kj-async INTERFACE "-pthread")
@@ -182,7 +191,7 @@ if(NOT CAPNP_LITE)
   add_library(kj-tls ${kj-tls_sources})
   add_library(CapnProto::kj-tls ALIAS kj-tls)
   target_link_libraries(kj-tls PUBLIC kj-async)
-  if (WITH_OPENSSL)
+  if(WITH_OPENSSL)
     target_compile_definitions(kj-tls PRIVATE KJ_HAS_OPENSSL)
     target_link_libraries(kj-tls PRIVATE OpenSSL::SSL OpenSSL::Crypto)
   endif()
@@ -270,7 +279,7 @@ if(BUILD_TESTING)
       compat/tls-test.c++
     )
     target_link_libraries(kj-heavy-tests kj-http kj-gzip kj-tls kj-async kj-test kj)
-    if (WITH_OPENSSL)
+    if(WITH_OPENSSL)
       set_property(
         SOURCE compat/tls-test.c++
         APPEND PROPERTY COMPILE_DEFINITIONS KJ_HAS_OPENSSL
--- a/.github/workflows/quick-test.yml
+++ b/.github/workflows/quick-test.yml
@@ -10,6 +10,15 @@ on:
     - 'release-*'
 
 jobs:
+  Linux-musl:
+    runs-on: ubuntu-latest
+    container: alpine:latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: install dependencies
+        run: apk add autoconf automake build-base cmake libtool libucontext-dev linux-headers openssl-dev
+      - name: super-test
+        run: ./super-test.sh quick
   Linux:
     runs-on: ubuntu-latest
     strategy:
--- a/c++/cmake/CapnProtoConfig.cmake.in
+++ b/c++/cmake/CapnProtoConfig.cmake.in
@@ -62,6 +62,38 @@ if (@WITH_OPENSSL@)  # WITH_OPENSSL
   endif()
 endif()
 
+if (@_WITH_LIBUCONTEXT@) # _WITH_LIBUCONTEXT
+  set(forwarded_config_flags)
+  if(CapnProto_FIND_QUIETLY)
+    list(APPEND forwarded_config_flags QUIET)
+  endif()
+  if(CapnProto_FIND_REQUIRED)
+    list(APPEND forwarded_config_flags REQUIRED)
+  endif()
+  # If the consuming project called find_package(CapnProto) with the QUIET or REQUIRED flags, forward
+  # them to calls to find_package(PkgConfig) and pkg_check_modules(). Note that find_dependency()
+  # would do this for us in the former case, but there is no such forwarding wrapper for
+  # pkg_check_modules().
+
+  find_package(PkgConfig ${forwarded_config_flags})
+  if(NOT ${PkgConfig_FOUND})
+    # If we're here, the REQUIRED flag must not have been passed, else we would have had a fatal
+    # error. Nevertheless, a diagnostic for this case is probably nice.
+    if(NOT CapnProto_FIND_QUIETLY)
+      message(WARNING "pkg-config cannot be found")
+    endif()
+    set(CapnProto_FOUND OFF)
+    return()
+  endif()
+
+  if (CMAKE_VERSION VERSION_LESS 3.6)
+    # CMake >= 3.6 required due to the use of IMPORTED_TARGET
+    message(SEND_ERROR "libucontext support requires CMake >= 3.6.")
+  endif()
+
+  pkg_check_modules(libucontext IMPORTED_TARGET ${forwarded_config_flags} libucontext)
+endif()
+
 include("${CMAKE_CURRENT_LIST_DIR}/CapnProtoTargets.cmake")
 include("${CMAKE_CURRENT_LIST_DIR}/CapnProtoMacros.cmake")
 
--- a/c++/src/kj/async-io-test.c++
+++ b/c++/src/kj/async-io-test.c++
@@ -373,9 +373,17 @@ bool systemSupportsAddress(StringPtr addr, StringPtr service = nullptr) {
   // Can getaddrinfo() parse this addresses? This is only true if the address family (e.g., ipv6)
   // is configured on at least one interface. (The loopback interface usually has both ipv4 and
   // ipv6 configured, but not always.)
+  struct addrinfo hints;
+  hints.ai_family = AF_UNSPEC;
+  hints.ai_socktype = 0;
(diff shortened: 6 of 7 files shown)


kentonv:
> this PR makes the assumption that a system that has libucontext needs it for the getcontext, makecontext and setcontext symbols.

Can we specifically detect if getcontext/makecontext is _missing_ from the base libc before looking for it in libucontext? I don't like the idea that the presence of libucontext on a glibc system would cause KJ to behave differently.

> The CI fails for two tests, I believe because docker on the github workflow does not support ipv6

It looks to me like the system is misconfigured: `getaddrinfo("localhost")` is returning the ipv6 address `::1`, but `getaddrinfo()` is supposed to avoid returning ipv6 addresses if ipv6 is not available.

c++/configure.ac
```diff
@@ -195,6 +200,19 @@ AS_IF([test "$with_openssl" != no], [
...
 AM_CONDITIONAL([BUILD_KJ_TLS], [test "$with_openssl" != no])
 
+AS_IF([test "$with_fibers" != no], [
+  CXXFLAGS="$CXXFLAGS -DKJ_USE_FIBERS"
+
+  AC_SEARCH_LIBS([getcontext], [], [], [
+    # If getcontext does not exist in libc, try to link libucontext
+    AC_CHECK_LIB(ucontext, getcontext, [
+      ASYNC_LIBS="$ASYNC_LIBS -lucontext"
+    ])
```
kentonv, line 210: I think we need a failure clause on this AC_CHECK_LIB to disable fibers? (Or error out if `--with-fibers` was specified explicitly.)

c++/src/kj/CMakeLists.txt
```diff
@@ -137,6 +137,26 @@ if(NOT CAPNP_LITE)
   add_library(kj-async ${kj-async_sources})
   add_library(CapnProto::kj-async ALIAS kj-async)
   target_link_libraries(kj-async PUBLIC kj)
+  if (KJ_USE_FIBERS)
```
kentonv, line 140: I'll need @harrishancock's help to review cmake stuff...

c++/configure.ac
JonasVautherin: Yep good point! However, I am not a big fan of changing the default based on what's found on the system. My proposition here is to error if the conditions are not met. The user can explicitly disable fibers in that case. That won't "just work", but that won't surprise anyone either.

What do you think?

c++/src/kj/CMakeLists.txt
JonasVautherin: Sure. Note that I could move the `PkgConfig` part out into a cmake module, and just do `find_package(ucontext REQUIRED)` instead, but that would only earn 2 lines here (at the cost of adding a file). It's just in case you don't want to see pkg-config there.

In this case it felt more expressive like this, but I'm happy to change.

JonasVautherin:
I managed to make the tests pass on the alpine container by changing `localhost` for `127.0.0.1`. I blindly tested it, I have no clue why the following was failing with `localhost`:

```
KJ_SYSCALL(::bind(sockfd, &addr.generic, addrlen), toString());
```

c++/src/kj/CMakeLists.txt
Sarcasm: Allow me to share my opinion but please note that I'm not a capnproto/kj maintainer, just have some opinions on CMake stuff.

In the top-level CMakeLists, I would handle the dependencies like `WITH_OPENSSL`, with `ON`, `OFF` and `AUTO` (default) options, and maybe the same naming scheme `WITH_FIBERS`?
And like `WITH_OPENSSL`, do the feature detection after the option.

https://github.com/capnproto/capnproto/blob/82b29296f76c583856511f645d9ab4e427a6d218/c%2B%2B/CMakeLists.txt#L44-L65

Then in kj's CMakeLists.txt, something like `WITH_OPENSSL`:

```cmake
  if (WITH_FIBERS)
    target_compile_definitions(kj-async PRIVATE KJ_USE_FIBERS)
    if (libucontext_FOUND)
      target_link_libraries(kj-async PUBLIC PkgConfig::libucontext)
    else()
  else()
    target_compile_definitions(kj-async PRIVATE KJ_USE_FIBERS=0)
  endif()
```

Then for CMake installation, propagate the libucontext dependencies like `WITH_OPENSSL`:

https://github.com/capnproto/capnproto/blob/82b29296f76c583856511f645d9ab4e427a6d218/c%2B%2B/cmake/CapnProtoConfig.cmake.in#L52-L63


Regarding `KJ_USE_FIBERS` being `PRIVATE`, I'm not sure it's correct:

```cmake
target_compile_definitions(kj-async PRIVATE KJ_USE_FIBERS=0) 
```

kj/async.h check/define it in its header, so I think it's part of the public API, and maybe should be propagated to consumers.
At least, the unit test file, should define it like `WITH_OPENSSL` here:

https://github.com/capnproto/capnproto/blob/82b29296f76c583856511f645d9ab4e427a6d218/c%2B%2B/src/kj/CMakeLists.txt#L273-L277

Note that using `set_source_files_properties` is wrong in the `WITH_OPENSSL` case, it would be better to `APPEND` the property with something like:

```cmake
  if (WITH_FIBERS)
    set_property(
      SOURCE async-test.c++
      APPEND PROPERTY COMPILE_DEFINITIONS KJ_USE_FIBERS
  )
  endif()
```

c++/configure.ac
kentonv: It's idiomatic with autotools to default to checking if the functionality is available and building it if it is. We already do that for zlib and openssl, for example. This is why the default for the `--with-x` flags is `check`, instead of `yes` or `no` -- `check` means "check if the library is available and use it if so".

If someone explicitly specifies `--with-fibers` (i.e. the value is `yes`, rather than `check`), and setcontext isn't available, then we should error out. But if they don't specify either `--with-fibers` or `--without-fibers` then we should enable or disable based on the result of the check.

(BTW, we might want to check for `makecontext`, `setcontext`, and/or `swapcontext` instead of `getcontext` -- I think some platforms have `getcontext` but don't support setting.)

kentonv:
Hmm, maybe the IP problem actually is our fault -- should we be passing `AI_ADDRCONFIG` in `ai_flags` to `getaddrinfo()`, to instruct it not to return IP versions that the system doesn't support? I am surprised this hasn't come up previously, though.

kentonv:
Wait, no, the man page says:

> Specifying  hints  as  NULL  is equivalent to setting ai_socktype and ai_protocol to 0; ai_family to AF_UNSPEC; and ai_flags to (AI_V4MAPPED | AI_ADDRCONFIG).

So `AI_ADDRCONFIG` is the default.

So the question is, why is `AI_ADDRCONFIG` not apparently working as expected? Is this a musl bug? Or is the system misconfigured in some way?

c++/configure.ac
JonasVautherin: Right, I was not aware of that. I made it check for all 4 symbols (though KJ does not use `swapcontext` for what I could see). It is a bit verbose, but I don't think I can ask `AC_SEARCH_LIBS` or `AC_CHECK_LIB` to check a list of symbols.

c++/src/kj/CMakeLists.txt
JonasVautherin: > Regarding KJ_USE_FIBERS being PRIVATE, I'm not sure it's correct

Right, I honestly don't know. @harrishancock any opinion?

@Sarcasm: I get your opinion about the options, but to me it feels like it's not in the scope of this PR. This PR is about adding support for musl, not improving the options, right? Just to keep it contained :innocent:

Sarcasm: > This PR is about adding support for musl, not improving the options, right?

This PR is adding a CMake option, CMake options defined by the project is a sort of public API that packagers relies on.
So, it's nice to provide stability here and IMHO `WITH_FIBERS` is a better name for 2 reasons:

1. The corresponding autotools option is `--with-fibers` (also, there is `--with-openssl`).
2. `WITH_OPENSSL` is already named like this.

```
./configure --with-openssl --with-fibers
cmake -DWITH_OPENSSL=ON -DWITH_FIBERS=ON
# VS
cmake -DWITH_OPENSSL=ON -DKJ_USE_FIBERS=ON
```

If supporting `AUTO` if too annoying for you, that can be added later on in a retro-compatible way.
This "How to" can be of help too:
- https://sarcasm.github.io/notes/dev/cmake.html#optional-dependencies
  The example, like this PR, uses `pkg_check_modules()`.
  Disclaimer: I'm the author.

The `find_dependency()` and `PUBLIC/PRIVATE` stuff seems critical to me, let's see what Harris has to say.

I don't my trying to implement my suggestions if you prefer but I don't have the environment to test for `libucontext`.

JonasVautherin: No sure, I can do it, I was not aware that `AUTO` was a thing in cmake, but you convinced me :blush:.

JonasVautherin: @Sarcasm: I think I implemented all your suggestions but the `find_dependency` one. I'm not completely sure how to do it, because `libucontext` does not provide a cmake module (that's why I use pkg-config). So `find_dependency()` (which is a wrapper around `find_package`) won't "just find it", and I'm not sure if calling pkg-config will do exactly what's needed, will it?

Any opinion there?

c++/src/kj/CMakeLists.txt
```diff
@@ -270,12 +279,18 @@ if(BUILD_TESTING)
...
         SOURCE compat/tls-test.c++
         APPEND PROPERTY COMPILE_DEFINITIONS KJ_HAS_OPENSSL
       )
     endif()
+    if(WITH_FIBERS)
+      set_property(
+        SOURCE async-test.c++
+        APPEND PROPERTY COMPILE_DEFINITIONS KJ_USE_FIBERS
+      )
+    endif()
```
JonasVautherin, line 293: This is the part failing the MSVC ci checks, but I am not sure why :man_shrugging:

c++/CMakeLists.txt
```diff
@@ -64,6 +64,46 @@ elseif (WITH_OPENSSL)
...
+set_property(CACHE WITH_FIBERS PROPERTY STRINGS AUTO ON OFF)
+
+if (WITH_FIBERS OR WITH_FIBERS STREQUAL "AUTO")
+  # Fibers need makecontext, setcontext, getcontext, swapcontext that may be in libc,
+  # or in libucontext (e.g. for musl)
+  include(CheckLibraryExists)
+  check_library_exists(c makecontext "" LIBC_HAS_MAKECONTEXT)
+  check_library_exists(c getcontext "" LIBC_HAS_GETCONTEXT)
+  check_library_exists(c setcontext "" LIBC_HAS_SETCONTEXT)
+  check_library_exists(c swapcontext "" LIBC_HAS_SWAPCONTEXT)
```
Sarcasm, line 79: Maybe you can get around a single compile check using `check_cxx_source_compiles`?

- https://cmake.org/cmake/help/latest/module/CheckCXXSourceCompiles.html#command:check_cxx_source_compiles

c++/CMakeLists.txt
```diff
@@ -64,6 +64,46 @@ elseif (WITH_OPENSSL)
...
+
+  if (LIBC_HAS_MAKECONTEXT AND LIBC_HAS_GETCONTEXT AND LIBC_HAS_SETCONTEXT AND LIBC_HAS_SWAPCONTEXT)
+    set(WITH_FIBERS ON)
+  else()
+    # Try with libucontext
+    find_package(PkgConfig)
+    if(PKG_CONFIG_FOUND)
+      pkg_check_modules(libucontext QUIET IMPORTED_TARGET libucontext)
```
Sarcasm, line 87: `IMPORTED_TARGET` requires CMake >= 3.6 (current minimum at top of this CMakeLists.txt is 3.4).

c++/CMakeLists.txt
```diff
@@ -64,6 +64,46 @@ elseif (WITH_OPENSSL)
...
+
+      if(libucontext_FOUND)
+        check_library_exists(ucontext makecontext "" UCONTEXT_HAS_MAKECONTEXT)
+        check_library_exists(ucontext getcontext "" UCONTEXT_HAS_GETCONTEXT)
+        check_library_exists(ucontext setcontext "" UCONTEXT_HAS_SETCONTEXT)
+        check_library_exists(ucontext swapcontext "" UCONTEXT_HAS_SWAPCONTEXT)
+      endif()
```
Sarcasm, line 94: Is it really necessary to check the validity of `libucontext`?

c++/src/kj/CMakeLists.txt
Sarcasm: I don't understand why either, but since `KJ_USE_FIBERS` is now `PUBLIC`, you don't need to set the source file property on the test case, this if/set_property can be removed.

Sarcasm:
> @Sarcasm: I think I implemented all your suggestions but the find_dependency one. I'm not completely sure how to do it, because libucontext does not provide a cmake module (that's why I use pkg-config). So find_dependency() (which is a wrapper around find_package) won't "just find it", and I'm not sure if calling pkg-config will do exactly what's needed, will it?
>
> Any opinion there?
> -- https://github.com/capnproto/capnproto/pull/1409#discussion_r795102771

It's annoying, I think the most correct approach is to copy the behavior of `find_dependency` (fortunately it's relatively short):

- https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/CMakeFindDependencyMacro.cmake#L33-63

Looks like the most important, is to forward the `QUIET` and `REQUIRED` of `find_package(CapnProto {REQUIRED?} {QUIET?})`.

The `_CMAKE_${dep}_TRANSITIVE_DEPENDENCY` stuff does not seem very important, IIUC it's just for https://cmake.org/cmake/help/latest/module/FeatureSummary.html.

c++/src/kj/CMakeLists.txt
JonasVautherin: Ok.

JonasVautherin:
@kentonv @Sarcasm: Respectfully, I'm slowly realizing that I am not competent enough to make that contribution properly; every change I make brings 3 new (admittedly justified) problems, and I won't spend another 20 hours on this PR because I don't need it. I was just trying to contribute because I thought I could.

Feel free to close this PR, or keep it open in case it could serve as inspiration one day, whatever the policy is. Thanks for the help, and sorry if I wasted your time :confused:.

Sarcasm:
Thanks you Jonas for your patience, the CMake stuff is just painful I understand how time consuming it can be.
I forked your PR to add/change some CMake stuff:

- https://github.com/Sarcasm/capnproto/commits/ci-add-musl

You can maybe cherry-pick these commits to this PR and see if that changes anything for the Windows build for example.
I feel like creating a new PR would lose some context.

On a side note, I have noticed the .pc files do not seem to propagate the OpenSSL dependencies, and not the KJ_USE_FIBERS usage requirements either, unlike what the CMake build tries to do.

c++/CMakeLists.txt
```diff
@@ -64,6 +64,58 @@ elseif (WITH_OPENSSL)
...
+    # Fibers need makecontext, setcontext, getcontext, swapcontext that may be in libc,
+    # or in libucontext (e.g. for musl)
+    include(CheckCXXSourceCompiles)
+
+    check_cxx_source_compiles("
+#include <ucontext.h>
+int main() {
+  ucontext_t cp;
+  getcontext(&cp);
+  setcontext(&cp);
```
JonasVautherin, line 89: @Sarcasm: I personally like the previous syntax more than this one, but I follow you if you say this is better :blush:. This said, shouldn't we test for swapcontext and makecontext, too?

Sarcasm: The reason I find this better, is that, one of the reasons CMake configure time can be slow, is because of slow features detection, so I thought a one-shot feature detection would be better than 4, but I could totally be wrong since symbol detection may not require to spawn a compiler (although I checked and it seems to invoke a compiler).

Other reasons is that the CMake configure output shows 2 lines instead of 8, and there is 1 resulting CACHE variable to check instead of 4.

You are right `swapcontext` and/or `makecontext` should probably be checked too, I was lazy (they seem more complex invoke with a function pointer).

I added a new commit that tests them:

- https://github.com/Sarcasm/capnproto/commit/4c54f395a30f8350c89eaa2a991beebb5b77fa17

JonasVautherin:
> Wait, no, the man page says:
>
> >    Specifying hints as NULL is equivalent to setting ai_socktype and ai_protocol to 0; ai_family to AF_UNSPEC; and ai_flags to (AI_V4MAPPED | AI_ADDRCONFIG).

@kentonv: So I tried to set those defaults explicitly, just to see what happens (see patch below). And it turns out that those two tests now seem to be passing, but a new one fails (see below). Could it be that the musl defaults are wrong?

```patch
diff --git a/c++/src/kj/async-io-unix.c++ b/c++/src/kj/async-io-unix.c++
index aa845bcb..b813ba2c 100644
--- a/c++/src/kj/async-io-unix.c++
+++ b/c++/src/kj/async-io-unix.c++
@@ -1208,11 +1208,16 @@ Promise<Array<SocketAddress>> SocketAddress::lookupHost(
   auto thread = heap<Thread>(kj::mvCapture(params, [outFd,portHint](LookupParams&& params) {
     FdOutputStream output((AutoCloseFd(outFd)));
 
+    struct addrinfo hints;
+    hints.ai_family = AF_UNSPEC;
+    hints.ai_socktype = 0;
+    hints.ai_flags = AI_W4MAPPED | AI_ADDRCONFIG;
+    hints.ai_protocol = 0;
     struct addrinfo* list;
     int status = getaddrinfo(
         params.host == "*" ? nullptr : params.host.cStr(),
         params.service == nullptr ? nullptr : params.service.cStr(),
-        nullptr, &list);
+        &hints, &list);
     if (status == 0) {
       KJ_DEFER(freeaddrinfo(list));
```

```
[ TEST ] kj/async-io-test.c++:387: legacy test: AsyncIo/AddressParsing
kj/async-io-unix.c++:1269: failed: DNS lookup failed.; params.host = ::; params.service = http; gai_strerror(status) = Name does not resolve
stack: 7fe19126e070
    ??:0: returning here
[ FAIL ] kj/async-io-test.c++:387: legacy test: AsyncIo/AddressParsing (23552 μs)
```

JonasVautherin:
I think that was it! POSIX defaults to `hints.ai_flags = 0`, but glibc defaults to `AI_V4MAPPED | AI_ADDRCONFIG`. So we should explicitly set it to be consistent.

c++/configure.ac
kentonv: Oh, right, I think we used to use `swapcontext` but stopped because it wasn't supported somewhere... switched to just using `setcontext`.

I don't think it's necessary to check for all of them. The existence of either `makecontext` or `setcontext` probably implies the rest. I just meant that `getcontext` itself may not necessarily imply the others, so it's not the right one to check.

c++/configure.ac
```diff
@@ -195,8 +200,63 @@ AS_IF([test "$with_openssl" != no], [
...
+  CXXFLAGS="$CXXFLAGS -DKJ_USE_FIBERS"
+], [
+  CXXFLAGS="$CXXFLAGS -DKJ_USE_FIBERS=0"
+])
+
+# CapnProtoConfig.cmake.in needs these variables,
+# we force them to NO because we don't need the CMake dependency for them,
+# the dependencies are provided by the .pc files.
+AC_SUBST(WITH_OPENSSL, NO)
+AC_SUBST(_WITH_LIBUCONTEXT, NO)
```
kentonv, line 259: Out of curiosity, why does one of these have a leading underscore and the other doesn't?

JonasVautherin: I think (@Sarcasm should confirm) that it is to say that `WITH_OPENSSL` is in the public API (that's a user-facing option), and `_WITH_LIBUCONTEXT` is meant for internal use (and therefore should not be set by the user).

c++/configure.ac
JonasVautherin: Right :facepalm: I misunderstood that.

kentonv:
LGTM, happy to merge if either @Sarcasm or @harrishancock approves the cmake part.

It would be slightly nice to rewrite the commits so that the getaddrinfo() fix is in one commit and the fibers build stuff in a separate commit, since these aren't really related. But if you don't want to do that then I can juts squash-and-merge the whole thing.

JonasVautherin:
> It would be slightly nice to rewrite the commits so that the getaddrinfo() fix is in one commit and the fibers build stuff in a separate commit

Does it work like this? I kept @Sarcasm's separate so he gets the credit for them

c++/configure.ac
Sarcasm: I confirm, but feel free to drop the leading underscore if you think it looks better.

Sarcasm:
I also think they would look nicer if there 2 separate commits.

I made the CMake commits so that you could follow the changes and cherry-pick them.
I think you can squash them with the musl support commit now.

I don't mind the attribution but if you do, you can add me as a co-author:
- https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors

kentonv:
Thanks for all the work on this! This should solve a lot of portability problems people have been having.

PR #1410, opened 2022-01-20
Disable the coroutine immediately-ready optimization before the first suspension

harrishancock:
This behavior was erroneous, and broke the guarantee that calling code could observe its own subsequent synchronous code run before newly-enqueued asynchronous code.


harrishancock:
I tried to replace the unit test with a new one which verifies the optimization fires, but it's difficult to test now. I tried a strategy where I counted the number of `EventLoop::run(1)` calls it took to advance the coroutine to a particular point, but had trouble diagnosing crashes in my attempt.

kentonv (approved):

PR #1411, opened 2022-01-20
Assert that HttpHeaders instances know about indexed ids

bcaimano:
The HttpHeaderTable::Builder class allows you to reference its future table before it is fully built. But that HttpHeaderTable must be used to make HttpHeaders instances only after HttpHeaderTable::Builder::build() is invoked under peril of index out of bounds error in debug mode or segfault in release mode. These errors can be somewhat mysterious since they usually occur on a different stack from where the HttpHeaders instance was made. This commit adds an assertion that makes this situation more evident and logs useful diagnostic info about which header id was added after the construction. Hopefully, it should downgrade a segfault to failed Promise or thrown Exception for most users.

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -528,7 +528,9 @@ struct HttpHeaderTable::IdsByNameMap {
 };
 
 HttpHeaderTable::Builder::Builder()
-    : table(kj::heap<HttpHeaderTable>()) {}
+    : table(kj::heap<HttpHeaderTable>()) {
+  table->buildStatus = BuildStatus::BUILDING;
+}
 
 HttpHeaderId HttpHeaderTable::Builder::add(kj::StringPtr name) {
   requireValidHeaderName(name);
@@ -576,7 +578,11 @@ bool HttpHeaders::isValidHeaderValue(kj::StringPtr value) {
 
 HttpHeaders::HttpHeaders(const HttpHeaderTable& table)
     : table(&table),
-      indexedHeaders(kj::heapArray<kj::StringPtr>(table.idCount())) {}
+      indexedHeaders(kj::heapArray<kj::StringPtr>(table.idCount())) {
+  KJ_ASSERT(
+      table.isReady(), "HttpHeaders object was constructed from "
+      "HttpHeaderTable that wasn't fully built yet at the time of construction");
+}
 
 void HttpHeaders::clear() {
   for (auto& header: indexedHeaders) {
--- a/c++/src/kj/compat/http.h
+++ b/c++/src/kj/compat/http.h
@@ -234,9 +234,20 @@ public:
   kj::StringPtr idToString(HttpHeaderId id) const;
   // Get the canonical string name for the given ID.
 
+  bool isReady() const;
+  // Returns true if this HttpHeaderTable either was default constructed or its Builder has
+  // invoked `build()` and released it.
+
 private:
   kj::Vector<kj::StringPtr> namesById;
   kj::Own<IdsByNameMap> idsByName;
+
+  enum class BuildStatus {
+    UNSTARTED = 0,
+    BUILDING = 1,
+    FINISHED = 2,
+  };
+  BuildStatus buildStatus = BuildStatus::UNSTARTED;
 };
 
 class HttpHeaders {
@@ -1077,10 +1088,22 @@ inline void HttpHeaderId::requireFrom(const HttpHeaderTable& table) const {
       "the provided HttpHeaderId is from the wrong HttpHeaderTable");
 }
 
-inline kj::Own<HttpHeaderTable> HttpHeaderTable::Builder::build() { return kj::mv(table); }
+inline kj::Own<HttpHeaderTable> HttpHeaderTable::Builder::build() {
+  table->buildStatus = BuildStatus::FINISHED;
+  return kj::mv(table);
+}
 inline HttpHeaderTable& HttpHeaderTable::Builder::getFutureTable() { return *table; }
 
 inline uint HttpHeaderTable::idCount() const { return namesById.size(); }
+inline bool HttpHeaderTable::isReady() const {
+  switch (buildStatus) {
+    case BuildStatus::UNSTARTED: return true;
+    case BuildStatus::BUILDING: return false;
+    case BuildStatus::FINISHED: return true;
+  }
+
+  KJ_UNREACHABLE;
+}
 
 inline kj::StringPtr HttpHeaderTable::idToString(HttpHeaderId id) const {
   id.requireFrom(*this);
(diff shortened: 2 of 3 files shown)


c++/src/kj/compat/http.c++
```diff
@@ -683,6 +683,12 @@ void HttpHeaders::add(kj::String&& name, kj::String&& value) {
 
 void HttpHeaders::addNoCheck(kj::StringPtr name, kj::StringPtr value) {
   KJ_IF_MAYBE(id, table->stringToId(name)) {
+    // Assert instead of allowing a mysterious non-debug only segfault.
+    KJ_ASSERT(
+        id->id < indexedHeaders.size(), "Attempted to use an unknown indexed header id, "
```
kentonv, line 688: If I understand correctly, the only case this fails is in fact when the header table is not fully built, and `indexedHeaders` is empty, right? Because otherwise `table` should never contain IDs that aren't in `indexedHeaders`.

So we could write this as:

```
KJ_ASSERT(!indexedHeaders.empty(), "tried to use HttpHeaderTable before it was fully built");
```

I think this error message would be a bit clearer to the reader.

kentonv: Actually that doesn't seem right, it appears `indexedHeaders` is built up as it goes and so I can't see how this assertion would ever be hit.

kentonv: Sorry I'm confusing myself.

What would actually have to happen is the `HttpHeaders` object is constructed from the table before the table is built, and then _after that_ the desired header is added to the header table builder, and then the resulting ID is used to set the header in the `HttpHeaders` object.

But that seems kind of hard to do by accident.

kentonv: So I think this is what we want here:

```
KJ_REQUIRE(id->id < indexedHeaders.size(), "HttpHeaders object was constructed from HttpHeaderTable that wasn't fully built yet");
```

bcaimano: I think it would have to be something like this:
```
KJ_ASSERT(!indexedHeaders.size() != table->idCount(), "tried to use HttpHeaderTable before it was fully built");
```
You could have added some but not all of the ids already. This is actually the failure case I believe I am experiencing. (We had added something like two header ids before we make the HttpHeaders instance and ten header ids after.)

The other thing that might be worth mentioning is that this variant of the assertion will fire even if you never would have used the unknown indexed headers. So if you register the header id "Foo" after making your `HttpHeaders` instance but never use it to parse "Foo", I think it will still fail the assertion. If you're okay with that, then I'd be happy to make the change.

jclee: If it's actually an error to call `HttpHeader`'s constructor with an unbuilt table, I suppose it might be preferable to assert that directly in the constructor, but I guess that as the system is currently implemented, that would require adding a bit of state to tables to track whether they had been built or not.

I also notice that `HttpHeaders::set()`, `get()`, and `unset()` do a similarly unchecked index into `indexedHeaders` that might warrant a similar assert.

bcaimano: > If it's actually an error to call HttpHeader's constructor with an unbuilt table, I suppose it might be preferable to assert that directly in the constructor, but I guess that as the system is currently implemented, that would require adding a bit of state to tables to track whether they had been built or not.

I think I'd call it a contract violation but I do agree that we can add an assertion with a small amount of state. It does potentially break users that otherwise would not have experienced a problem. Maybe a debug only assertion?

> I also notice that HttpHeaders::set(), get(), and unset() do a similarly unchecked index into indexedHeaders that might warrant a similar assert.

Sure, if you think it's worth it, I'm happy to add assertions there as well.

bcaimano:
Btw, this should be ready for another look. I'll leave the fixups in till the next review to make the history easier to follow.

c++/src/kj/compat/http.c++
kentonv: > t does potentially break users that otherwise would not have experienced a problem.

Ehh I think the users potentially broken are already in a precarious position here, if they're constructing headers with unbuilt header tables. So I'd be happy with a hard error in the HttpHeaders constructor in this case.

c++/src/kj/compat/http.h
```diff
@@ -1089,12 +1091,14 @@ inline kj::StringPtr HttpHeaderTable::idToString(HttpHeaderId id) const {
 
 inline kj::Maybe<kj::StringPtr> HttpHeaders::get(HttpHeaderId id) const {
   id.requireFrom(*table);
+  assertKnownIndexedHeaderId(id);
```
kentonv, line 1094: This method could previously be optimized down to an array index, but now it involves an out-of-line function call. That seems kind of sad. Can we make this a debug-only check? (Or assert in the HttpHeaders constructor instead, as discussed.)

bcaimano: Yeah, I agree that it's a bit sad. I was conflicted as well. Out of curiosity, do you happen to remember why these functions were in the header? Maybe similar hopes for inline optimization?

Anyway, since you're fine with the ctor assertion, I'll do that instead.

kentonv: `get()` is intended to be inline-able, yes. A major design goal here -- the reason we have the whole header ID registration system -- was that accessing headers should be extremely cheap, compared to typical HTTP libraries that do a name hash lookup for every access.

bcaimano:
Okay, this should be ready for another look.

c++/src/kj/compat/http.h
```diff
@@ -234,9 +234,20 @@ class HttpHeaderTable {
   kj::StringPtr idToString(HttpHeaderId id) const;
   // Get the canonical string name for the given ID.
 
+  bool isValid() const;
```
kentonv, line 237: Sorry to bikeshed but something like `isReady()` might be a clearer name...

c++/src/kj/compat/http.h
```diff
@@ -234,9 +234,20 @@ class HttpHeaderTable {
...
 private:
   kj::Vector<kj::StringPtr> namesById;
   kj::Own<IdsByNameMap> idsByName;
+
+  enum class BuildStatus {
+    UNSTARTED = 0,
+    BUILDING = 1,
+    FINISHED = 2,
+  };
+  BuildStatus buildStatus = BuildStatus::UNSTARTED;
```
kentonv, line 250: Could just be an `isReady` bool that defaults to true, but is set false by Builder's constructor, and then set true again in `build()`? The enum feels heavy to me, especially being int-sized rather than one byte. But I guess this is pretty nitpicky... I dunno.

c++/src/kj/compat/http.h
bcaimano: That's a good point, I'll switch it.

c++/src/kj/compat/http.h
bcaimano: I think this should pad to be the same size in practice.

PR #1413, opened 2022-01-21
CMake: don't overwrite tls-test.c++ COMPILE_DEFINITIONS, append to it

Sarcasm:
An potential issue I saw when re-reading the code (I was the original author of the WITH_OPENSSL CMake stuff).


harrishancock (approved):
Thanks!

PR #1417, opened 2022-01-28
Make sure to clean up properly with a pipe hits an exception to avoid inadvertently hanging

a-robinson:
The first commit is definitely needed to fix a known problem.

The second one (A) might not be needed, and (B) might go too far. All tests pass with it, including internally, but that's far from a sure sign it's ok. So I'm happy to discard it if you aren't sure either.

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -3289,6 +3289,11 @@ private:
         canceler.release();
         pipe.endState(*this);
         fulfiller.fulfill();
+      }, [this](kj::Exception&& e) {
+        canceler.release();
+        pipe.endState(*this);
+        fulfiller.reject(kj::cp(e));
+        kj::throwRecoverableException(kj::mv(e));
       }));
     }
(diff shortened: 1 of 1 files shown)


a-robinson:
I didn't see any other possible issues along these lines elsewhere in kj.

a-robinson:
Although I suppose I should also check capnp itself, let's see.

a-robinson:
Nope, don't see any issues in capnp either.

Although it looks like I shouldn't be using `auto` for parameters in the second commit given that we try to support older compiler versions here. I can clean that up if you decide you want to keep that commit, but I'll wait until then since I suspect we're gonna just discard it.

kentonv:
I guess we should drop the second commit unless we have evidence of a real problem.

Probably the _real_ right thing to do would be to put the pipe into an aborted state after any exception is seen but I'm not completely sure so let's leave as-is for now.

a-robinson:
Sounds good to me too, removed the second commit.

kentonv (approved):
LGTM assuming CI passes.

PR #1418, opened 2022-02-02
Fix bug with queue wait time metrics in TwoPartyNetwork

xortive:
Before this PR, sending a message over a two party connection with an empty queue, after not having sent a message for some time, would cause incorrect queue wait time metrics to be reported until the message is actually written.


a-robinson (approved):

a-robinson:
I'll merge this Friday if Kenton hasn't taken a look by then. I assume he doesn't really care to look at this, but am leaving some time on the off chance he does.

PR #1419, opened 2022-02-07
Fix release tests

kentonv:


a-robinson (approved):

kentonv:
The MinGW-wine problem looks like it may be a temporary package dependency breakage in Ubuntu introduced recently. I'm going to ignore it for the moment in hopes that it'll be fixed upstream. Otherwise it seems like we'd either have to switch to a different distro or disable the test.

PR #1422, opened 2022-02-09
Fix windows-sanity.h in NOGDI case, again.

### LINKED ISSUE #1421: window-sanity.h still appears inccorect
does not define ERROR at all if not previously seen, but code will still try to reference the win32::ERROR entity.
(https://github.com/capnproto/capnproto/blob/55a368beb987abf9eeb9b3843e9c5423ad37ab29/c%2B%2B/src/kj/windows-sanity.h#L52-L56)

Perhaps applying a patch similar to following would be an improvement:
```
diff --git a/c++/src/kj/windows-sanity.h b/c++/src/kj/windows-sanity.h
index 894ec1d7..f18b39e6 100644
--- a/c++/src/kj/windows-sanity.h
+++ b/c++/src/kj/windows-sanity.h
@@ -49,9 +49,13 @@
 #pragma once
 
 namespace win32 {
+#ifdef ERROR  // This could be absent if e.g. NOGDI was used.
   const auto ERROR_ = ERROR;
 #undef ERROR
   const auto ERROR = ERROR_;
+#else //if macro def missing, we still need to define the symbol.
+  const auto ERROR = 0;
+#endif
 
   typedef VOID VOID_;
 #undef VOID
```

kentonv:
Since `ERROR` is conditionally defined, we can't write `using win32::ERROR` later in the file. Use `using namespace` instead to pull in everything.

I also renamed the namespace to be more unique.

Fixes #1421.

### DIFF
--- a/c++/src/kj/windows-sanity.h
+++ b/c++/src/kj/windows-sanity.h
@@ -48,7 +48,11 @@
 // now, we use `#pragma once` to tell the compiler never to include this file again.
 #pragma once
 
-namespace win32 {
+namespace kj_win32_workarounds {
+  // Namespace containing constant definitions intended to replace constants that are defined as
+  // macros in the Windows headers. Do not refer to this namespace directly, we'll import it into
+  // the global scope below.
+
 #ifdef ERROR  // This could be absent if e.g. NOGDI was used.
   const auto ERROR_ = ERROR;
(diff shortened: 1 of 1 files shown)


harrishancock (approved):
Sorry I didn't see this!

PR #1426, opened 2022-02-20
Check ERANGE error from getcwd

### LINKED ISSUE #1425: compile: getcwd(): Result too large
Hi,

We're using `capnp` with [Bazel](https://bazel.build/), which likes to use long paths and in some situations will cause `capnp compile` to fail.

```
--stderr:
*** Fatal uncaught kj::Exception: kj/filesystem-disk-unix.c++:1717: failed: getcwd(): Result too large
thread 'main' panicked at 'schema compiler command: Error { kind: Failed, description: "Error while trying to execute `capnp compile`: Failed: Premature end of file.  Please verify that version 0.5.2 or higher of the capnp executable is installed on your system. See https://capnproto.org/install.html" }', external/raze__segments_rpc__0_0_0/rpc/build.rs:17:10
stack backtrace:
```

The offending snippet:

https://github.com/capnproto/capnproto/blob/767b8a30ee100f1c648a75c9aaf252212d904c17/c%2B%2B/src/kj/filesystem-disk-unix.c%2B%2B#L1711-L1719

Reading through the documentation for [`getcwd`](https://linux.die.net/man/3/getcwd), it seems that `ERANGE` should be checked instead of (in addition to?) `ENAMETOOLONG`. Am I correct in reading that `ENAMETOOLONG` is actually only applicable to `getwd`? I don't see any mention of `ENAMETOOLONG` in the Open Group Specification for [`getcwd`](https://pubs.opengroup.org/onlinepubs/007904975/functions/getcwd.html).

So I think there may be two fixes available.

1. Use `ERANGE` instead of `ENAMETOOLONG`.
2. Use [`std::filesystem::current_path();`](https://en.cppreference.com/w/cpp/filesystem/current_path) (introduced in C++17) as described in this [Stack Overflow answer](https://stackoverflow.com/a/2203694).

I'm not too deep into C++ so I'm unsure of which of the two proposed solutions is correct 😄 What are your thoughts?

uhthomas:
Fixes #1425

### DIFF
--- a/c++/src/kj/filesystem-disk-test.c++
+++ b/c++/src/kj/filesystem-disk-test.c++
@@ -958,17 +958,17 @@ KJ_TEST("DiskFilesystem::computeCurrentPath") {
   auto maxPathSegment = std::string(255, 'a');
 
   // Create a path which exceeds the 256 byte buffer used in
-  // computeCurrentPath. 
+  // computeCurrentPath.
   auto subdir = dir->openSubdir(Path({
     maxPathSegment,
     maxPathSegment,
     "some_path_longer_than_256_bytes"
   }), WriteMode::CREATE | WriteMode::CREATE_PARENT);
 
-  fchdir(KJ_ASSERT_NONNULL(subdir->getFd()));
+  KJ_SYSCALL(fchdir(KJ_ASSERT_NONNULL(subdir->getFd())));
 
   // Test computeCurrentPath indirectly.
-  newDiskFilesystem();
+  KJ_EXPECT(newDiskFilesystem());
 }
 
 }  // namespace
--- a/c++/src/kj/filesystem-disk-test.c++
+++ b/c++/src/kj/filesystem-disk-test.c++
@@ -946,6 +946,7 @@ KJ_TEST("DiskFile holes") {
 }
 #endif
 
+#if !_WIN32 // Only applies to Unix.
 // Ensure the current path is correctly computed.
 //
 // See issue #1425.
@@ -965,11 +966,14 @@ KJ_TEST("DiskFilesystem::computeCurrentPath") {
     "some_path_longer_than_256_bytes"
   }), WriteMode::CREATE | WriteMode::CREATE_PARENT);
 
+  auto origDir = open(".", O_RDONLY | O_DIRECTORY);
   KJ_SYSCALL(fchdir(KJ_ASSERT_NONNULL(subdir->getFd())));
+  KJ_DEFER(KJ_SYSCALL(fchdir(origDir)));
 
   // Test computeCurrentPath indirectly.
-  KJ_EXPECT(newDiskFilesystem());
+  newDiskFilesystem();
 }
+#endif
 
 }  // namespace
 }  // namespace kj
--- a/c++/src/kj/filesystem-disk-test.c++
+++ b/c++/src/kj/filesystem-disk-test.c++
@@ -23,6 +23,7 @@
 #include "test.h"
 #include "encoding.h"
 #include <stdlib.h>
+#include <string>
 #if _WIN32
 #include <windows.h>
 #include "windows-sanity.h"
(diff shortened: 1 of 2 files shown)


c++/src/kj/filesystem-disk-test.c++
```diff
@@ -945,5 +946,30 @@ KJ_TEST("DiskFile holes") {
...
+  // Paths can be PATH_MAX, but the segments which make up that path typically
+  // can't exceed 255 bytes.
+  auto maxPathSegment = std::string(255, 'a');
+
+  // Create a path which exceeds the 256 byte buffer used in
+  // computeCurrentPath. 
+  auto subdir = dir->openSubdir(Path({
+    maxPathSegment,
+    maxPathSegment,
+    "some_path_longer_than_256_bytes"
```
kentonv, line 965: Running the CI now and we'll see what it says, but I'm going to guess this part fails on Windows and maybe some other OSes. Might need to limit to Linux.

c++/src/kj/filesystem-disk-unix.c++
```diff
@@ -1705,12 +1705,15 @@ private:
       }
     }
 
+    // TODO(someday): Use MAX_PATH from <limits.h>?
```
kentonv, line 1708: I intentionally didn't want to over-allocate here.

c++/src/kj/filesystem-disk-unix.c++
```diff
@@ -1705,12 +1705,15 @@ private:
...
+    // TODO(someday): Use std::filesystem::current_path() when compiling with
```
kentonv, line 1709: KJ generally prefers to avoid the C++ standard library and go direct to OS APIs, so I'd remove this comment.

kentonv (commented):
Thanks for fixing this!

c++/src/kj/filesystem-disk-test.c++
uhthomas: Interesting thought.

I've pushed a change which addresses the unused result errors. Let's see if CI is any happier now, and if not I'm happy to restrict this to Linux.

kentonv: It actually does appear that it passed on windows-latest so maybe Windows does support long paths after all? Somehow I thought it was limited to 256 chars total for the whole path.

Actually I bet it works specifically because the KJ filesystem impl uses the crazy `\\?\` prefix that bypasses a huge amount of Win32 legacy cruft, heh.

kentonv: Wait no. The windows-latest build is broken and is not even doing a build, but reporting success. :man_facepalming:

c++/src/kj/filesystem-disk-test.c++
```diff
@@ -945,5 +946,30 @@ KJ_TEST("DiskFile holes") {
...
+
+  // Create a path which exceeds the 256 byte buffer used in
+  // computeCurrentPath.
+  auto subdir = dir->openSubdir(Path({
+    maxPathSegment,
+    maxPathSegment,
+    "some_path_longer_than_256_bytes"
+  }), WriteMode::CREATE | WriteMode::CREATE_PARENT);
+
+  KJ_SYSCALL(fchdir(KJ_ASSERT_NONNULL(subdir->getFd())));
```
kentonv, line 968: You need to arrange to change directories back at the end of the test, otherwise:
* All remaining tests (e.g. unit tests in other source files) will run in the wrong directory, possibly leading to unexpected behavior.
* `TempDir`'s destructor won't be able to delete the temporary directory... I'm kind of surprised this isn't causing a failure, actually.

`fchdir` doesn't exist on Windows, so we definitely at least have to wrap this test in `#if !_WIN32`. That's fine since it's testing for a bug that was Unix-specific anyway.

kentonv: Something like...

```
int origDir = open(cwd, O_RDONLY | O_DIRECTORY);
...
KJ_DEFER(fchdir(origDir));
```

c++/src/kj/filesystem-disk-test.c++
```diff
@@ -945,5 +946,30 @@ KJ_TEST("DiskFile holes") {
...
+
+  // Test computeCurrentPath indirectly.
+  KJ_EXPECT(newDiskFilesystem());
```
kentonv, line 971: KJ_EXPECT doesn't really make sense here since `newDiskFilesystem()` doesn't return a boolean. I think this only complies by accident.

kentonv: To be clear, you can just remove the `KJ_EXCEPT`. This function wont produce a warning if the result is unused. `fchdir` is apparently specially marked to produce such a warning.

c++/src/kj/filesystem-disk-test.c++
uhthomas: Sounds good to me! Thank you for all the help and snippets 😄

c++/src/kj/filesystem-disk-test.c++
uhthomas: Done! Thank you again for being so helpful!

c++/src/kj/filesystem-disk-test.c++
uhthomas: 😅 Restricted the test to Unix.

kentonv:
Thanks!

PR #1428, opened 2022-02-22
Provide UnbufferedGzipOutputStream to avoid extra copy

jrsala-auguration:
This is in order to save a potentially unnecessary copy into that 4kB buffer when the user already has a `BufferedOutputStream` at hand. I might have violated a few style rules, and made the interface of `GzipOutputStream` less obvious when reading the header. Also I didn't look at `AsyncGzipOutputStream`.


kentonv:
I like the objective here, but I'm finding the code very hard to follow. The class hierarchy and combined use of templates+inheritance is not a pattern I've seen before.

I would have expected we'd simply add a constructor overload to the existing stream classes that accepts a buffered stream and acts appropriately. We'd want to change the internal buffer to be dynamically allocated, so that it's not allocated when not needed. Perhaps, indeed, the constructor that takes a non-buffered stream would work by dynamically allocating a `Buffered{Input,Output}StreamWrapper` around that stream, so that the rest of the implementation can always assume a buffered stream...

jrsala-auguration:
> I like the objective here, but I'm finding the code very hard to follow.

I agree, I tried to factor out common bits into base classes, including the use of CRTP, and sometimes that's just not worth it I guess.

>The class hierarchy and combined use of templates+inheritance is not a pattern I've seen before.

That's just the CRTP as I'm sure you're aware, but if it can be avoided, it should. I just didn't find a way to make things simpler without repeating a lot of code.
 
> I would have expected we'd simply add a constructor overload to the existing stream classes that accepts a buffered stream and acts appropriately. We'd want to change the internal buffer to be dynamically allocated, so that it's not allocated when not needed. Perhaps, indeed, the constructor that takes a non-buffered stream would work by dynamically allocating a `Buffered{Input,Output}StreamWrapper` around that stream, so that the rest of the implementation can always assume a buffered stream...

That seems like the right course of action. I wasn't sure changing the internals of GzipOutputStream from having 4kB within the `GzipOutputStream` object to any other memory layout was acceptable. Chesterton's fence and all that.

Thanks for the feedback.

kentonv:
Couldn't you just have something like:

```
kj::Maybe<kj::Own<BufferedInputStreamWrapper>> ownInput;  // if we allocated it...
BufferedInputStream& input;
```

jrsala-auguration:
Hi Kenton, thanks for your response. I think this would work, and the only runtime costs would be at construction and destruction time and therefore negligible. I will try that.

jrsala-auguration:
I had to modify `BufferOutputStreamWrapper::write` in the case where it detects that a write has been made directly into its buffer. Without this change, `GzipOutputStream::pump` would fill the buffer and then ask for the next write buffer with `getWriteBuffer()` but that would return an empty `ArrayPtr<byte>` because the buffer hadn't been flushed.

An alternative would have been for `GzipOutputStream::pump` to manually invoke `BufferedOutputStreamWrapper::flush()` but that was impossible because it only has a `BufferedOutputStream&` at hand, and that type doesn't have a virtual `flush()` member function.

I think it makes sense to automatically flush the `BufferedOutputStreamWrapper` when a direct write is detected and the buffer is therefore full. This behavior makes sense because it guarantees that in the usage pattern of manual direct writes to the buffer, `getWriteBuffer()` always returns space to write. Please tell me what you think.

jrsala-auguration:
When `VectorOutputStream::getWriteBuffer()` is called, it grows the vector to make space. So for consistency it might be best to have `BufferedOutputStreamWrapper` similarly flush when `getWriteBuffer()` is called and there's no space left, instead of flushing as soon as a `write` call fills the buffer.

Issue #1429, opened 2022-02-23
`kj-test` lib built unnecessarily


matthew-limbinar:
The `kj-test` target is only needed by the `kj-tests` target, but it is always built. 

A dependency of mine uses this project (admittedly, in an unconventional way -- using CMake on Linux instead of autotools, as seems to be expected), and I occasionally get errors like these when building my own unit tests that use that dependency:
```
catch_with_main.cpp:(.text.startup+0x0): multiple definition of `main'
[...]/lib/libkj-test.a(test.c++.o):test.c++:(.text.startup+0x0): first defined here
```

kentonv:
Are you or your dependency linking against kj-test? Only unit tests that use the KJ unittest framework should be linking against this library.

PR #1430, opened 2022-02-23
1429: Make `kj-test` target build conditionally

matthew-limbinar:
Addressing issue #1429: Make this target build and install only when required. No substantive change; just relocating the code inside the `if`-statement.


kentonv:
`kj-test` is a unit test framework that is intended to be usable by projects that depend on KJ, to write their own unit tests. It should always be built and installed so that those users can use it.

I don't understand the issue you're trying to solve. `kj-test` is built as a separate library. You should only link against it if you are building a unit test using the KJ test framework. It sounds like you're somehow getting it linked in by accident in your project when you don't actually want it. You need to figure out why that's happening. It sounds like a bug in your build files or the build files of the project you depend on that itself depends on KJ.

PR #1431, opened 2022-02-27
add popFront to ProducerConsumerQueue

vaci:
I have a use case for a ProducerConsumerStack.

This patch trivially adds LIFO behaviour to kj::ProducerConsumerQueue, but a separate implementation class might be might be more appropriate?

### DIFF
--- a/c++/src/kj/async-queue.h
+++ b/c++/src/kj/async-queue.h
@@ -102,7 +102,7 @@ private:
 
 template <typename T>
 class ProducerConsumerQueue {
-  // ProducerConsumerQueue is an async FIFO queue.
+  // ProducerConsumerQueue is an async FIFO queue or LIFO stack.
 
 public:
   void push(T v) {
@@ -146,6 +146,23 @@ public:
     }
   }
 
+  Promise<T> popFront() {
+    // Eventually pop a value from the queue.
+    // Note that if your sinks lag your sources, the promise will always be ready.
+
+    if (!values.empty()) {
+      // We have at least one value, get the youngest.
+      KJ_IASSERT(waiters.empty());
+
+      auto value = kj::mv(values.front());
+      values.pop_front();
+      return kj::mv(value);
+    } else {
+      // We don't have any values, add ourselves to the waiting queue.
+      return waiters.wait();
+    }
+  }
+  
 private:
   std::list<T> values;
   WaiterQueue<T> waiters;


kentonv:
Hmm, some thoughts...
* Is it weird that WaiterQueue is still FIFO?
* This seems to commit us to using a double-ended underlying queue structure, which worries me slightly.
* For LIFO, a `kj::Vector` would actually be a better underlying data structure.
* Nitpick: I'd expect the name `popBack()` or `popLast()` rather than `popFront()`. (It looks like you chose the name to match how the underlying list is being used, but the underlying list seems to be used intuitively backwards here, with new items being added using push_front().)

vaci:
Yeah, I realised after posting that a more general "task allocator" concept would be more useful than extending this class. 
I can imagine one would want to specialise the worker and task containers, as well as the "distribution policy" for the work.

jrsala-auguration:
@kentonv On this topic, wouldn't generic async data structures come in handy for a wide range of problems?

```c++
template <typename T>
class AsyncList {
public:
  T& pushBack(const T&);
  T& pushBack(T&&);
  T& pushFront(const T&);
  T& pushFront(T&&);
  kj::Promise<T> popBack();
  kj::Promise<T> popFront();

  size_t size() const;
  void rejectAll(Exception e);

private:
  std::list<T> values;
  WaiterQueue<T> waiters;
};

template <typename T>
class AsyncQueue {
public:
  T& push(const T& v) { return list.pushBack(v); }
  T& push(T&& v) { return list.pushBack(kj::mv(v)); }
  kj::Promise<T> pop() { return list.popFront(); }
  void rejectAll(Exception e) { list.rejectAll(kj::mv(e)); }

private:
  AsyncList<T> list;
};

template <typename T>
class AsyncStack {
public:
  T& push(const T& v) { return list.pushBack(v); }
  T& push(T&& v) { return list.pushBack(kj::mv(v)); }
  kj::Promise<T> pop() { return list.popBack(); }
  void rejectAll(Exception e) { list.rejectAll(kj::mv(e)); }

private:
  AsyncList<T> list;
};

template <typename T>
using ProducerConsumerQueue = AsyncQueue<T>;
// For backwards-compatibility
```

I'm sure a `kj::Vector` can be used as well if performance matters, e.g. by allowing the type of the underlying data structure in `AsyncList` to be provided as a template parameter.

kentonv:
@jrsala-auguration I don't think it'd be quite as easy as that to make the data structure interchangable. Also in addition to FIFO vs. LIFO you might want different implementations to apply different kinds of backpressure, or other use-case-specific concerns.

Meanwhile it's really not that much code to implement these from first principles. So I usually just implement exactly what I need within the context of any particular use case...

PR #1432, opened 2022-02-28
Fix computeCurrentPath test with generic Unix test

ObsidianMinor:
Test fails in ekam since it also runs as a generic Unix test, which doesn't have O_DIRECTORY. This should fix the test in the short-term.


kentonv:
I think you can also just remove `O_DIRECTORY` from the flags, its only effect is to make `open()` fail if the path isn't a directory but that's not really necessary.

PR #1433, opened 2022-03-01
Add comment clarifying how to get unindexed headers

MellowYarker:
It's not immediately clear how one should try to access unindexed headers. This comment clarifies why we don't have a `get` method for unindexed headers, and is similar to an existing comment under the `unset` method.

### DIFF
--- a/c++/src/kj/compat/http.h
+++ b/c++/src/kj/compat/http.h
@@ -298,6 +298,12 @@ public:
 
   kj::Maybe<kj::StringPtr> get(HttpHeaderId id) const;
   // Read a header.
+  //
+  // Note that there is intentionally no method to look up a header by string name rather than
+  // header ID. The intent is that you should always allocate a header ID for any header that you
+  // care about, so that you can get() it by ID. Headers with registered IDs are stored in an array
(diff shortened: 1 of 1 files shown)


c++/src/kj/compat/http.h
```diff
@@ -298,6 +298,13 @@ class HttpHeaders {
 
   kj::Maybe<kj::StringPtr> get(HttpHeaderId id) const;
   // Read a header.
+  //
+  // We do not provide a convenience method to access an unindexed header by string name because
```
kentonv, line 302: Hmm, I'd phrase this a little differently since "indexed headers" is not a term we've used anywhere else in the documentation. How about:

> Note that there is intentionally no method to look up a header by string name rather than header ID. The intent is that you should always allocate a header ID for any header that you care about, so that you can `get()` it by ID. Headers with registered IDs are stored in an array indexed by ID, making lookup fast. Headers without registered IDs are stored in a separate list that is optimized for re-transmission of the whole list, but not for lookup.

MellowYarker: Thanks, I prefer this update as it's more descriptive.

PR #1434, opened 2022-03-06
OpenSSL 3.0

thomaslee:
Debian is preparing to transition to OpenSSL 3.0 and we ran into two issues with this in capnproto:

1. Failing tests attributable to openssl error messages changing slightly
2. Lots of "unimplemented bioctrl" warnings, mostly associated with kTLS

### DIFF
--- a/c++/src/kj/compat/tls.c++
+++ b/c++/src/kj/compat/tls.c++
@@ -369,9 +369,8 @@ private:
     switch (cmd) {
       case BIO_CTRL_FLUSH:
         return 1;
-      case BIO_CTRL_INFO:
       case BIO_CTRL_EOF:
-        // Queries for info and EOF
+        // Queries for EOF
         return 0;
 #if OPENSSL_VERSION_NUMBER >= 0x30000000L && \
         !defined(OPENSSL_IS_BORINGSSL) && \
--- a/c++/src/kj/compat/tls.c++
+++ b/c++/src/kj/compat/tls.c++
@@ -370,7 +370,10 @@ private:
       case BIO_CTRL_FLUSH:
         return 1;
       case BIO_CTRL_EOF:
-        // Queries for EOF
+        // Queries for EOF. OpenSSL sometimes polls this BIO_CTRL after
+        // attempting a read, but we don't have a good way to check for
+        // an EOF explicitly so we "lie" and say we have not reached EOF.
+        // OpenSSL can still use the read return value to indicate EOF.
         return 0;
 #ifdef BIO_CTRL_GET_KTLS_SEND
       case BIO_CTRL_GET_KTLS_SEND:
--- a/c++/src/kj/compat/tls.c++
+++ b/c++/src/kj/compat/tls.c++
@@ -175,6 +175,9 @@ public:
   kj::Promise<void> accept() {
     // We are the server. Set SSL options to prefer server's cipher choice.
     SSL_set_options(ssl, SSL_OP_CIPHER_SERVER_PREFERENCE);
+#ifdef SSL_OP_IGNORE_UNEXPECTED_EOF
+    SSL_set_options(ssl, SSL_OP_IGNORE_UNEXPECTED_EOF);
+#endif
 
     auto acceptPromise = sslCall([this]() {
       return SSL_accept(ssl);
@@ -347,7 +350,11 @@ private:
     BIO_clear_retry_flags(b);
     KJ_IF_MAYBE(n, reinterpret_cast<TlsConnection*>(BIO_get_data(b))->readBuffer
         .read(kj::arrayPtr(out, outl).asBytes())) {
-      return *n;
+      int ret = *n;
+      if (ret == 0) {
+        BIO_set_flags(b, BIO_FLAGS_IN_EOF);
+      }
+      return ret;
     } else {
       BIO_set_retry_read(b);
       return -1;
@@ -370,11 +377,7 @@ private:
       case BIO_CTRL_FLUSH:
         return 1;
       case BIO_CTRL_EOF:
-        // Queries for EOF. OpenSSL sometimes polls this BIO_CTRL after
-        // attempting a read, but we don't have a good way to check for
-        // an EOF explicitly so we "lie" and say we have not reached EOF.
-        // OpenSSL can still use the read return value to indicate EOF.
-        return 0;
+        return BIO_test_flags(b, BIO_FLAGS_IN_EOF) != 0 ? 1 : 0;
 #ifdef BIO_CTRL_GET_KTLS_SEND
       case BIO_CTRL_GET_KTLS_SEND:
       case BIO_CTRL_GET_KTLS_RECV:
--- a/c++/src/kj/compat/tls.c++
+++ b/c++/src/kj/compat/tls.c++
@@ -372,9 +372,7 @@ private:
       case BIO_CTRL_EOF:
         // Queries for EOF
         return 0;
-#if OPENSSL_VERSION_NUMBER >= 0x30000000L && \
-        !defined(OPENSSL_IS_BORINGSSL) && \
-        !defined(OPENSSL_NO_KTLS)
+#ifdef BIO_CTRL_GET_KTLS_SEND
       case BIO_CTRL_GET_KTLS_SEND:
       case BIO_CTRL_GET_KTLS_RECV:
         // Queries to determine if our BIO supports KTLS
--- a/c++/src/kj/compat/tls.c++
+++ b/c++/src/kj/compat/tls.c++
@@ -369,6 +369,18 @@ private:
     switch (cmd) {
       case BIO_CTRL_FLUSH:
         return 1;
+      case BIO_CTRL_INFO:
+      case BIO_CTRL_EOF:
+        // Queries for info and EOF
+        return 0;
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L && \
+        !defined(OPENSSL_IS_BORINGSSL) && \
+        !defined(OPENSSL_NO_KTLS)
+      case BIO_CTRL_GET_KTLS_SEND:
+      case BIO_CTRL_GET_KTLS_RECV:
+        // Queries to determine if our BIO supports KTLS
+        return 0;
+#endif
       case BIO_CTRL_PUSH:
       case BIO_CTRL_POP:
         // Informational?
--- a/c++/src/kj/compat/tls-test.c++
+++ b/c++/src/kj/compat/tls-test.c++
@@ -682,16 +682,25 @@ void expectInvalidCert(kj::StringPtr hostname, TlsCertificate cert, kj::StringPt
   KJ_EXPECT_THROW_MESSAGE(message, clientPromise.wait(test.io.waitScope));
 }
 
+// OpenSSL 3.0 changed error messages
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L && !defined(OPENSSL_IS_BORINGSSL)
+#define SSL_MESSAGE(v11, v30) v30
+#else
+#define SSL_MESSAGE(v11, v30) v11
+#endif
+
 KJ_TEST("TLS certificate validation") {
   expectInvalidCert("wrong.com", TlsCertificate(kj::str(VALID_CERT, INTERMEDIATE_CERT)),
-                    "Hostname mismatch");
+                    SSL_MESSAGE("Hostname mismatch", "hostname mismatch"));
   expectInvalidCert("example.com", TlsCertificate(VALID_CERT),
                     "unable to get local issuer certificate");
   expectInvalidCert("example.com", TlsCertificate(kj::str(EXPIRED_CERT, INTERMEDIATE_CERT)),
                     "certificate has expired");
   expectInvalidCert("example.com", TlsCertificate(SELF_SIGNED_CERT),
-                    "self signed certificate");
+                    SSL_MESSAGE("self signed certificate", "self-signed certificate"));
 }
+
+#undef SSL_MESSAGE
 
 // BoringSSL seems to print error messages differently.
 #ifdef OPENSSL_IS_BORINGSSL
@@ -852,6 +861,8 @@ KJ_TEST("TLS client certificate verification") {
     KJ_EXPECT(!id->hasCertificate());
   }
 }
+
+#undef SSL_MESSAGE
 
 class MockConnectionReceiver final: public ConnectionReceiver {
   // This connection receiver allows mocked async connection establishment without the network.


c++/src/kj/compat/tls.c++
```diff
@@ -369,6 +369,18 @@ private:
     switch (cmd) {
       case BIO_CTRL_FLUSH:
         return 1;
+      case BIO_CTRL_INFO:
+      case BIO_CTRL_EOF:
+        // Queries for info and EOF
+        return 0;
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L && \
```
kentonv, line 376: Maybe this could just be `#ifdef BIO_CTRL_GET_KTLS_SEND`? That way if/when BoringSSL adds support, we won't need to update this.

c++/src/kj/compat/tls.c++
```diff
@@ -369,6 +369,18 @@ private:
     switch (cmd) {
       case BIO_CTRL_FLUSH:
         return 1;
+      case BIO_CTRL_INFO:
```
kentonv, line 372: I don't suppose there are any, like, docs describing these?

`BIO_CTRL_INFO` and `BIO_CTRL_EOF` don't seem to be new, but I've never seen the warning log for them...

c++/src/kj/compat/tls.c++
thomaslee: Good idea, sure.

c++/src/kj/compat/tls.c++
thomaslee: > BIO_CTRL_INFO and BIO_CTRL_EOF don't seem to be new, but I've never seen the warning log for them...

I just ran the tests against both 1.1.1 and 3.0.1 again to be sure: [link](https://gist.github.com/thomaslee/a6c13708a0a0c6a6a2727e2b9e5b7261). Neither are called using Debian unstable's 1.1.1 build and only `BIO_CTRL_EOF` is called in Debian's 3.0.1. No calls to `BIO_CTRL_INFO` in either release. I grepped the full build logs just to be certain:

```
$ grep bio_ctrl capnproto*.build
capnproto_0.9.1-4_arm64.build:kj/compat/tls.c++:377: unimplemented bio_ctrl; cmd = 2
capnproto_0.9.1-4_arm64.build:kj/compat/tls.c++:377: unimplemented bio_ctrl; cmd = 2
```

So I must have messed something up when I was testing the changes and prepping the patch, looks like it might just be some subtle change in semantics around polling the EOF bit in the BIO. I'll update the PR to remove the special casing of `BIO_CTRL_INFO`.

> I don't suppose there are any, like, docs describing these?

Yeah, the OpenSSL docs are painfully terse. You probably know more about OpenSSL than I do but my understanding is that bioctrls are the BIO (OpenSSL I/O abstraction) equivalent of ioctls: openssl calls the bioctrl callback to query the state of the BIO. The best "docs" I was able to find was the openssl sources unfortunately, but you can typically tie the bioctrls back to macros in `bio.h` that *do* have some useful (if minimal) docs.

For example, it's irrelevant now that we've confirmed it's not being hit in tests but `BIO_CTRL_INFO` drives [BIO_get_mem_data](https://github.com/openssl/openssl/blob/e278f18563dd3dd67c00200ee30402f48023c6ef/include/openssl/bio.h.in#L515) which seems to be used to fetch a pointer into buffer memory in things like the [mem BIO](https://github.com/openssl/openssl/blob/4cedf30e995f9789cf6bb103e248d33285a84067/crypto/bio/bss_mem.c#L294-L300) (and/or "readable bytes" for unbuffered stuff like the [fd BIO](https://github.com/openssl/openssl/blob/4cedf30e995f9789cf6bb103e248d33285a84067/crypto/bio/bss_fd.c#L157-L159) ... seems dangerous to mix the two now that I look at it, but what do I know.). Here are the [docs for BIO_get_mem_data](https://www.openssl.org/docs/man1.1.1/man3/BIO_s_mem.html), among others.

`BIO_CTRL_EOF` unsurprisingly drives [BIO_eof](https://github.com/openssl/openssl/blob/e278f18563dd3dd67c00200ee30402f48023c6ef/include/openssl/bio.h.in#L533), so just OpenSSL polling to see if the BIO is in an "EOF" state. Some [docs for BIO_eof](https://www.openssl.org/docs/man1.1.1/man3/BIO_wpending.html).

kentonv: Well, I guess the question then, is: Are we doing anything wrong by implementing `BIO_CTRL_EOF` to always return zero? Does zero mean "I don't know", or does it mean "No, not at EOF"? If the latter, then is it bad to lie when we actually are at EOF?

thomaslee: [A full backtrace](https://gist.githubusercontent.com/thomaslee/14e7515edfa94115a8e23be3df862bf5/raw/2bee2fd4c38ba758ed8ff2e6cc741451e831f8fe/gistfile1.txt) from within bioCtrl while running capnproto tests show the `BIO_eof` checks originate from `ssl3_read_n` in both tests: [relevant openssl 3.0.1 source](https://github.com/openssl/openssl/blob/openssl-3.0.1/ssl/record/rec_layer_s3.c#L301-L303)

This particular case seems like a premature EOF check while accepting a new TLS connection. If we lie about EOF we'll just keep retrying the read even though the underlying socket might be dead. At the same time I suspect we always would have lied about the EOF check in 1.1.1 if we tickled a code path that happened to trigger that bioctrl, (though we would have emitted a warning if that happened).

So if we wanted to be safe, is there an easy way for us to detect stream EOF for a `TlsConnection` from within `bioCtrl` that might make sense here?

kentonv: I don't think we do, no. So let's go with a comment like:

> OpenSSL sometimes calls this, hoping to probe whether the stream is at EOF. We don't have any way to check this in terms of KJ interfaces. But if we lie and say "no", OpenSSL will keep trying reads, and if the read itself reports EOF, OpenSSL accepts that.

thomaslee: Hmm added the comment, but looking more closely at the [openssl code](https://github.com/openssl/openssl/blob/openssl-3.0.1/ssl/record/rec_layer_s3.c#L296-L323) without being half asleep:

1. If the read returns <= 0 and the BIO's "retriable" flag is **not** set, OpenSSL checks the EOF bit
2. If the conditions for (1) are met and the EOF bit is set, check if unexpected EOFs can be ignored. If so, continue. If not, fail hard.
3. Again check if the read returns <= 0, optionally release buffers and return the read return value

That's a bit odd because you'd expect the `BIO_read` [here](https://github.com/openssl/openssl/blob/openssl-3.0.1/ssl/record/rec_layer_s3.c#L298) to invoke the `TlsConnection::bioRead` method, right? And since we're seeing the `BIO_eof` check happen, we know we're **not** hitting the branch in `bioRead` that calls `BIO_set_retry_read`, we're hitting the branch that returns `*n`.

So assume `*n` evaluates to `-1` or `0` ... is that possible? And if so, is it a bug for us not to call `BIO_set_retry_read` in the `*n` branch for a "non-erroneous" read?

[BIO_should_retry docs](https://www.openssl.org/docs/man3.0/man3/BIO_should_retry.html)

thomaslee: No, from the linked docs:

> For example if a call to BIO_read_ex() on a socket BIO returns 0 and BIO_should_retry() is false then the cause will be that the connection closed. A similar condition on a file BIO will mean that it has reached EOF.

So yeah, I still don't fully understand the necessity of the `BIO_eof` check if the `BIO_read` return value indicates that we're in some kind of EOF/error state anyway but I think kj is doing the right thing.

thomaslee: Ah, [this commit](https://github.com/openssl/openssl/commit/d924dbf4ae127c68463bcbece04b6e06abc58928) introduced the `BIO_eof` check we're seeing. Check out the [changes to `bss_sock.c`](https://github.com/openssl/openssl/commit/d924dbf4ae127c68463bcbece04b6e06abc58928#diff-c6bb81a35c881c60f05c0dedfb7c70853a6e9699d4b3fb5100a62e303ce63659R216). Might be a hint as to how we can handle this.

If we assume a return value of zero from `TlsConnection::bioRead` implies EOF, this change still fails the test but it's maybe "correct"?:

```
tom@debian-bullseye-arm64:~/src/capnproto-debian$ git diff src/kj/compat/tls.c++
diff --git a/src/kj/compat/tls.c++ b/src/kj/compat/tls.c++
index 05bf5d5..1867544 100644
--- a/src/kj/compat/tls.c++
+++ b/src/kj/compat/tls.c++
@@ -339,7 +339,11 @@ private:
     BIO_clear_retry_flags(b);
     KJ_IF_MAYBE(n, reinterpret_cast<TlsConnection*>(BIO_get_data(b))->readBuffer
         .read(kj::arrayPtr(out, outl).asBytes())) {
-      return *n;
+      int status = *n;
+      if (status == 0) {
+        BIO_set_flags(b, BIO_FLAGS_IN_EOF);
+      }
+      return status;
     } else {
       BIO_set_retry_read(b);
       return -1;
@@ -361,6 +365,16 @@ private:
     switch (cmd) {
       case BIO_CTRL_FLUSH:
         return 1;
+      case BIO_CTRL_EOF:
+        return BIO_test_flags(b, BIO_FLAGS_IN_EOF) != 0 ? 1 : 0;
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L && \
+        !defined(OPENSSL_IS_BORINGSSL) && \
+        !defined(OPENSSL_NO_KTLS)
+      case BIO_CTRL_GET_KTLS_SEND:
+      case BIO_CTRL_GET_KTLS_RECV:
+        // Queries to determine if our BIO supports KTLS
+        return 0;
+#endif
       case BIO_CTRL_PUSH:
       case BIO_CTRL_POP:
         // Informational?
```

And the test output:

```
[ TEST ] kj/compat/tls-test.c++:1014: TLS receiver does not stall on client that disconnects before ssl handshake
kj/compat/tls.c++:448: error accepting tls connection; kj::mv(e) = kj/compat/tls.c++:66: failed: OpenSSL error; message = error:0A000126:SSL routines::unexpected eof while reading
stack: ffff7fcbca83 ffff7fcc815f ffff7fccc8fb ffff7fbfad5f ffff7fcc3960 ffff7fcbc710 ffff7fcc3150
stack: ffff7fcc42df ffff7fc09e8f ffff7fbfcd6b ffff7fc0460b aaaac47be363 aaaac48d6643 ffff7fbaf1fb ffff7fb46703 ffff7fbaff93 ffff7fbb056b ffff7fb783d3 ffff7fb46703 ffff7fb751d3 ffff7fbae44f ffff7f74a613 aaaac4608cf7
```

Think we can also be "more correct" while remaining backward compatible by setting the `SSL_OP_IGNORE_UNEXPECTED_EOF` option. Thoughts?

Also totally fine to stick with the original change since it seems to work and I think this is an edge case we're dealing with here. A few rabbit holes deep at this point. :)

thomaslee: If it's not clear from the commit this is totally what the new `BIO_CTRL_EOF` check is for btw: to bubble up the unexpected EOF in a way that gives a useful error because previously `errno` would not typically be set on an EOF and folks would get that weird "error: success" type output.

thomaslee: Confirmed tests pass using the above diff with `SSL_OP_IGNORE_UNEXPECTED_EOF` set, so that's a viable option here.

thomaslee: See a789003 for the full proposed change.

kentonv: Hmm, I think that flagging an error on unclean shutdown is a good thing, so `SSL_OP_IGNORE_UNEXPECTED_EOF` is probably not the right answer. Do tests break if we don't set that flag?

thomaslee: That's correct, tests fail because an exception is thrown:

```
failed: OpenSSL error; message = error:0A000126:SSL routines::unexpected eof while reading
```

So throwing an exception on unexpected EOF would seem to be a change in semantics, but guess that could be considered a bug & we just change the tests.

kentonv: Yes, I'd argue this was a bug before, possibly even a mild security problem (truncation attacks). We should fix the test to properly shut down or to expect the exception.

thomaslee: @kentonv I should clarify that the test failure will only occur against openssl 3.0, so we're talking about different semantics building against openssl 1.1 vs. 3.0. If you're still OK with that, I'll make it happen.

kentonv: Understood. Yes I still think the behavior is desirable. We'll have to design the test so it succeeds either way, which I suppose means that it should always do a clean shutdown.

pepijndevos:
For what it's worth I'm also in a situation where using openssl 1.1 is causing issues. So is the current status that the tests have to be updated?

thomaslee:
@pepijndevos that's correct, the tests need to be modified and I've been slow to get back to this after failing to force a clean shutdown in the tests. I'll see if I can take the route of simply expecting the exception instead a little later today.

c++/src/kj/compat/tls.c++
thomaslee: Okay looks like the exception is being routed through `TlsConnectionReceiver::taskFailed` ... my repro case on top of all the Debian build tooling is currently using 0.9.1 which is missing the new callback in `TlsConnectionReceiver` that I would be able to use to capture the exception when building against OpenSSL 3.0. Planning to backport the change introducing the callback to 0.9.1 in Debian just to get this change passing tests against OpenSSL 3.0 then I'll work those changes back to capnproto latest.

h-vetinari:
Hey all, what's the status of this PR?

When trying to package capnproto against OpenSSL 3 in [conda-forge](https://github.com/conda-forge/capnproto-feedstock/pull/32) (we're trying to get things buttoned up well before the EOL of 1.1.1 relatively soon), we're running into some test errors on unix:
```
961 test(s) passed 
2 test(s) failed 
```

In more detail:
```
[ TEST ] kj/compat/tls-test.c++:685: TLS certificate validation
kj/compat/tls-test.c++:682: failed: expected ::kj::_::hasSubstring(e->getDescription(), message); exception description didn't contain expected substring; *e = kj/compat/tls.c++:170: failed: TLS peer's certificate is not trusted; reason = hostname mismatch
stack: 7f6583692450 7f658369294a 7f65835b37e1 7f6583682ad0 56121b5700f0 56121b569e90 56121b44ffaa 56121b56b0b8
stack: 56121b56ed55 56121b57435d 56121b5744ca 7f658356d38e 7f658350e199 7f658356ef26 7f658356f8c1 7f65835465ee 7f658350e199 7f65835409d2 7f658356d180 7f6582d42554 56121b1ea09c
kj/compat/tls-test.c++:682: failed: expected ::kj::_::hasSubstring(e->getDescription(), message); exception description didn't contain expected substring; *e = kj/compat/tls.c++:170: failed: TLS peer's certificate is not trusted; reason = self-signed certificate
stack: 7f6583692450 7f658369294a 7f65835b37e1 7f6583682ad0 56121b5700f0 56121b569e90 56121b44ffaa 56121b56b0b8
stack: 56121b56ed55 56121b57435d 56121b574622 7f658356d38e 7f658350e199 7f658356ef26 7f658356f8c1 7f65835465ee 7f658350e199 7f65835409d2 7f658356d180 7f6582d42554 56121b1ea09c
[ FAIL ] kj/compat/tls-test.c++:685: TLS certificate validation (199265 μs)
[ TEST ] kj/compat/tls-test.c++:703: TLS client certificate verification
kj/compat/tls.c++:66: failed: OpenSSL error; message = error:16000069:STORE routines::unregistered scheme
stack: 7f6583687390 7f658368d3fb 7f658368d658 7f65835d7699 7f65835d75a4 56121b56c2a1 56121b56c3e2 56121b578b3e 7f658356d38e
[ FAIL ] kj/compat/tls-test.c++:703: TLS client certificate verification (161491 μs)
```

I've picked the contents of this PR into a patch on top of 0.10.3, but while the "TLS certificate validation" then passes, the client version still fails:
```
2022-12-14T00:21:39.3218924Z [ TEST ] kj/compat/tls-test.c++:712: TLS client certificate verification
2022-12-14T00:21:39.5138043Z kj/compat/tls.c++:66: failed: OpenSSL error; message = error:16000069:STORE routines::unregistered scheme
2022-12-14T00:21:39.5139064Z error:80000002:system library::No such file or directory
2022-12-14T00:21:39.5139887Z stack: 7f9e425ec3b0 7f9e425f243b 7f9e425f2698 7f9e4253c699 7f9e4253c5a4 55b5c32f12a1 55b5c32f13e2 55b5c32fdb3e 7f9e424d238e
2022-12-14T00:21:39.5135986Z [ FAIL ] kj/compat/tls-test.c++:712: TLS client certificate verification (193842 μs)
```

kentonv:
I opened #1596 to complete this work.

jakirkham:
It looks like PR ( https://github.com/capnproto/capnproto/pull/1596 ) was merged. Is this PR still needed?

jakirkham:
Also it looks like the issue @h-vetinari mentioned above is similar (the same?) as the one described in PR ( https://github.com/capnproto/capnproto/pull/1435 )

kentonv:
Yeah this PR is superseded by #1596.

PR #1435, opened 2022-03-06
More detailed OpenSSL errors

thomaslee:
While working on the OpenSSL 3.0 test failures that led to #1434 I ran into a test failure that was a bit harder to debug because there wasn't a lot of detail in the OpenSSL error reported by capnproto:

```
[ TEST ] kj/compat/tls-test.c++:710: TLS client certificate verification
kj/compat/tls.c++:66: failed: OpenSSL error; message = error:16000069:STORE routines::unregistered scheme
stack: ffff8323ba33 ffff832465ab ffff832468b3 ffff83197d27 ffff83197bff aaaabff624d3 aaaabff62663 aaaabff6bb17 ffff8312e1fb
[ FAIL ] kj/compat/tls-test.c++:710: TLS client certificate verification (30588 μs)
```

Here the `system library` is reporting a missing file/directory, but which one?

It turns out the details are hidden away in the `data` associated with the error. The same error with additional detail from `ERR_get_error_all` (the OpenSSL 3.x function that supersedes `ERR_get_error_line_data`) from an earlier incarnation of this change:

```
kj/compat/tls.c++:72: failed: OpenSSL error; message = src/kj/compat/tls.c++:329
error:16000069:STORE routines::unregistered scheme, filename=../crypto/store/store_register.c, lineno=237, func=ossl_store_get0_loader_int, data=scheme=file, flags=3
error:80000002:system library::No such file or directory, filename=../providers/implementations/storemgmt/file_store.c, lineno=269, func=file_open, data=calling stat(/usr/lib/ssl/certs), flags=3
stack: ffffbe174d3b ffffbe17fbef ffffbe17fef3 ffffbe0d0d27 ffffbe0d0bff aaaae0780553 aaaae07806e3 aaaae0789c97 ffffbe0671fb
```

We can clearly see OpenSSL was looking for `/usr/lib/ssl/certs` but it didn't exist. With this info in the build logs it was trivial to fix the broken test.

To make this kind of issue clearer/easier to debug going forward this PR outputs the additional error detail (filename, lineno, data) as part of `throwOpensslError`.

Note that it's tricky to force an error with `data` set so the test here is a little incomplete, but the implementation follows [examples found in the OpenSSL code base](https://github.com/openssl/openssl/blob/OpenSSL_1_1_1m/crypto/err/err_prn.c#L36-L42).

### DIFF
--- a/c++/src/kj/compat/tls.c++
+++ b/c++/src/kj/compat/tls.c++
@@ -57,10 +57,20 @@ void throwOpensslError() {
   // throw it.
 
   kj::Vector<kj::String> lines;
-  while (unsigned long long error = ERR_get_error()) {
+  const char *fname;
+  int lno;
+  const char *data;
+  int flags;
+  while (unsigned long long error = ERR_get_error_line_data(&fname, &lno, &data, &flags)) {
     char message[1024];
     ERR_error_string_n(error, message, sizeof(message));
-    lines.add(kj::heapString(message));
+
+    bool has_data = (flags & ERR_TXT_STRING) != 0;
+    const char *extra_prefix = has_data ? " [" : "";
+    const char *extra_suffix = has_data ? "]" : "";
+    const char *extra = has_data ? data : "";
+
+    lines.add(kj::str("[", fname, ":", lno, "] ", message, extra_prefix, extra, extra_suffix));
   }
   kj::String message = kj::strArray(lines, "\n");
   KJ_FAIL_ASSERT("OpenSSL error", message);
--- a/c++/src/kj/compat/tls-test.c++
+++ b/c++/src/kj/compat/tls-test.c++
@@ -1123,6 +1123,19 @@ KJ_TEST("TLS receiver does not stall on hung client") {
   KJ_EXPECT(!extraAcceptPromise.poll(test.io.waitScope));
 }
 
+void throwBadContext() {
+  TlsContext::Options options;
+  options.cipherList = "bogus,broken,wrong";
+
+  TlsContext ctx(kj::mv(options));
+}
+
+KJ_TEST("OpenSSL error formatting") {
+  KJ_EXPECT_THROW_MESSAGE("ssl_lib.c", throwBadContext());
+  KJ_EXPECT_THROW_MESSAGE("SSL_CTX_set_cipher_list", throwBadContext());
+  KJ_EXPECT_THROW_MESSAGE("no cipher match", throwBadContext());
+}
+
 #ifdef KJ_EXTERNAL_TESTS
 KJ_TEST("TLS to capnproto.org") {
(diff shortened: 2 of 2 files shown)


thomaslee:
Fixed the MSVC/Win2016 build failures, but I don't think this PR caused the most recent MacOS build failure.

c++/src/kj/compat/tls.c++
```diff
@@ -57,10 +57,20 @@ void throwOpensslError() {
...
     char message[1024];
     ERR_error_string_n(error, message, sizeof(message));
-    lines.add(kj::heapString(message));
+
+    bool has_data = (flags & ERR_TXT_STRING) != 0;
+    const char *extra_prefix = has_data ? " [" : "";
+    const char *extra_suffix = has_data ? "]" : "";
+    const char *extra = has_data ? data : "";
+
+    lines.add(kj::str("[", fname, ":", lno, "] ", message, extra_prefix, extra, extra_suffix));
```
kentonv, line 73: I take it `fname` is the OpenSSL source file name?

`kj::Exception` itself already stores a filename and line number, maybe we should initialize that directly? Like so:

```
kj::throwFatalException(kj::Exception(kj::Exception::Type::FAILED,
    fname, lno, kj::str(message, extra_prefix, extra, extra_suffix));
```

The one problem is how to deal with multiple errors. In that case I'd say we want the overall exception's source location to match either the first or last error (not sure which is more appropriate...) and then the rest could be in the description... maybe?

thomaslee: > I take it fname is the OpenSSL source file name?

That's correct, yep. Sometimes it's a weird relative path like `../ssl/ssl_lib.c`, seems to depend on how OpenSSL was built.

> The one problem is how to deal with multiple errors. In that case I'd say we want the overall exception's source location to match either the first or last error (not sure which is more appropriate...) and then the rest could be in the description... maybe?

Ah I think the original semantics make sense, like I'd expect the filename + lineno to match the exception "callsite" myself. If including filename + line number info from OpenSSL in the error message makes you uneasy I'm not particularly attached to it. The `data` fragment is more important IMO.

kentonv: Note the exception's stack trace will already identify the location where it is constructed, so identifying the location of the KJ assertion is actually redundant.

To me, what we're doing here is translating an openssl exception to a KJ exception, and it feels right for the filename / line number to be part of that translation.

But on a practical note, I guess the OpenSSL source location probably isn't that useful to anyone who isn't an OpenSSL developer, so maybe that's a good argument for just omitting it?

jakirkham:
This may need to merge in latest changes ( like PR https://github.com/capnproto/capnproto/pull/1596 )

PR #1438, opened 2022-03-09
allow defer to be cancelled

vaci:
I have a use case for cancelling a kj::defer(), rather like a transaction commit/rollback. I can write:

```
  bool success = false;
  KJ_DEFER({ if (!success) abort(); });
```

...but it seems a bit sad to keep track of another flag! Deferred is private, though, so not sure if this patch is the best approach.

### DIFF
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -1870,9 +1870,14 @@ public:
   KJ_DISALLOW_COPY(Deferred);
 
   // This move constructor is usually optimized away by the compiler.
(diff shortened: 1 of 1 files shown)


vaci:
Provide public access to `Deferred::cancel()`

kentonv:
You can just add a `cancel()` method to `_::Deferred`. The point of it being "private" is that people should not write code that directly instantiates `_::Deferred`, but if they called `kj::deferred()` to get the object then I'm OK with them calling a method like `cancel()` on it.

vaci:
Cool, then the first patch is sufficient!

PR #1439, opened 2022-03-09
Docs: correct target of parameter annotation

vaci:


vaci:
Oops, there's already an open PR for this fix.

kentonv:
The other one seems to be 5 years old, and not mergeable, so I'll re-open yours and merge.

PR #1441, opened 2022-03-12
Fix bug in TreeIndex erase().

### LINKED ISSUE #1436: Multiple tree indexes on a single table break error handling when inserting duplicate rows
Given two indexes, where one is a non-unique computed property:

```cpp
class UniqueKeyCallbacks {
public:
  inline const kj::String& keyForRow(const kj::String& entry) const {
    return entry;
  }

  inline kj::String& keyForRow(kj::String& entry) const {
    return entry;
  }

  template <typename KeyLike>
  inline bool matches(const kj::String& e, KeyLike&& key) const {
    return e == key;
  }

  template <typename KeyLike>
  inline bool matches(kj::String& e, KeyLike&& key) const {
    return e == key;
  }

  template <typename KeyLike>
  inline bool isBefore(const kj::String& e, KeyLike&& key) const {
    return e < key;
  }

  template <typename KeyLike>
  inline bool isBefore(kj::String& e, KeyLike&& key) const {
    return e < key;
  }

  template <typename KeyLike>
  inline auto hashCode(KeyLike&& key) const {
    return kj::hashCode(key);
  }
};

class StrLengthCallbacks {
public:
  inline const size_t keyForRow(const kj::String& entry) const {
    return entry.size();
  }

  inline size_t keyForRow(kj::String& entry) const {
    return entry.size();
  }

  template <typename KeyLike>
  inline bool matches(const kj::String& e, KeyLike&& key) const {
    return e.size() == key;
  }

  template <typename KeyLike>
  inline bool matches(kj::String& e, KeyLike&& key) const {
    return e.size() == key;
  }

  template <typename KeyLike>
  inline bool isBefore(const kj::String& e, KeyLike&& key) const {
    return e.size() < key;
  }

  template <typename KeyLike>
  inline bool isBefore(kj::String& e, KeyLike&& key) const {
    return e.size() < key;
  }
};
```

The following example throws `kj/table.c++:49: failed: inserted row already exists in table`, which is expected.

```cpp
kj::Table<
  kj::String,
  kj::HashIndex<UniqueKeyCallbacks>,
  kj::TreeIndex<StrLengthCallbacks>
> table;
table.insert(kj::str("a"));
table.insert(kj::str("ab"));
table.insert(kj::str("abc"));
table.insert(kj::str("xyz"));
```

This, however, throws `kj/common.h:1730: failed: expected index < size_; Out-of-bounds ArrayPtr access`, which appears to be an internal error.

```cpp
kj::Table<
  kj::String,
  kj::HashIndex<UniqueKeyCallbacks>,
  kj::TreeIndex<UniqueKeyCallbacks>,
  kj::TreeIndex<StrLengthCallbacks>
> table;
table.insert(kj::str("a"));
table.insert(kj::str("ab"));
table.insert(kj::str("abc"));
table.insert(kj::str("xyz"));
```

The error appears to occur within the call to `erase` here:

https://github.com/capnproto/capnproto/blob/d86ef24f070a9b1a976f38a19103b88c931b4e1d/c%2B%2B/src/kj/table.h#L453-L456

kentonv:
erase() is supposed to avoid accessing the row that is being erased. This is important in particular when rolling back an insertion due to a later index reporting an insertion failure.

Thanks @tniessen for providing a repro.

Fixes #1436

### DIFF
--- a/c++/src/kj/table-test.c++
+++ b/c++/src/kj/table-test.c++
@@ -1317,6 +1317,94 @@ KJ_TEST("insertion order index is movable") {
   KJ_EXPECT(iter == range.end());
 }
 
+// =======================================================================================
+// Test bug where insertion failure on a later index in the table would not be rolled back
+// correctly if a previous index was TreeIndex.
+
+class StringLengthCompare {
+  // Considers two strings equal if they have the same length.
+public:
+  inline size_t keyForRow(StringPtr entry) const {
+    return entry.size();
+  }
+
+  inline bool matches(StringPtr e, size_t key) const {
+    return e.size() == key;
+  }
+
+  inline bool isBefore(StringPtr e, size_t key) const {
+    return e.size() < key;
+  }
+
+  uint hashCode(size_t size) const {
+    return size;
+  }
+};
+
+KJ_TEST("HashIndex rollback on insertion failure") {
+  // Test that when an insertion produces a duplicate on a later index, changes to previous indexes
+  // are properly rolled back.
+
+  Table<StringPtr, HashIndex<StringHasher>, HashIndex<StringLengthCompare>> table;
+  table.insert("a"_kj);
+  table.insert("ab"_kj);
+  table.insert("abc"_kj);
+
+  {
+    // We use upsert() so that we don't throw an exception from the duplicate, but this exercises
+    // the same logic as a duplicate insert() other than throwing.
+    kj::StringPtr& found = table.upsert("xyz"_kj, [&](StringPtr& existing, StringPtr&& param) {
+      KJ_EXPECT(existing == "abc");
+      KJ_EXPECT(param == "xyz");
+    });
+    KJ_EXPECT(found == "abc");
+
+    table.erase(found);
+  }
+
+  table.insert("xyz"_kj);
+
+  {
+    kj::StringPtr& found = table.upsert("tuv"_kj, [&](StringPtr& existing, StringPtr&& param) {
+      KJ_EXPECT(existing == "xyz");
+      KJ_EXPECT(param == "tuv");
+    });
+    KJ_EXPECT(found == "xyz");
+  }
+}
+
+KJ_TEST("TreeIndex rollback on insertion failure") {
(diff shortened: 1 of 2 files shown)


harrishancock (approved):
Nice find.

c++/src/kj/table-test.c++
```diff
@@ -1317,6 +1317,94 @@ KJ_TEST("insertion order index is movable") {
...
 }
 
+// =======================================================================================
+// Test bug where insertion failure on a later index in the table would not be rolled back
+// correctly if a previous index was TreeIndex.
+
+class StringLengthCompare {
+  // Considers two strings equal if they have the same length.
+public:
+  inline const size_t keyForRow(StringPtr entry) const {
```
harrishancock, line 1327: The `const` in this return type is causing the builds to fail.
```
src/kj/table-test.c++:1327:10: error: 'const' type qualifier on return type has no effect [-Werror,-Wignored-qualifiers]
  inline const size_t keyForRow(StringPtr entry) const {
         ^~~~~~
```

kentonv: Copy/pasto. Removed (no other change).

kentonv:
Apparently I did something else horribly wrong here... but only with certain compiler versions/flags?

kentonv:
Duh, little fixup...

```diff
diff --git a/c++/src/kj/table.h b/c++/src/kj/table.h
index 603b0eb1..630b79c9 100644
--- a/c++/src/kj/table.h
+++ b/c++/src/kj/table.h
@@ -1542,7 +1542,7 @@ private:
   inline auto searchKeyForErase(kj::ArrayPtr<Row>& table, uint pos, Params&... params) const {
     // When erasing, the table entry for the erased row may already be invalid, so we must avoid
     // accessing it.
-    auto predicate = [&](uint i) {
+    auto predicate = [&,pos](uint i) {
       return i != pos && cb.isBefore(table[i], params...);
     };
     return SearchKeyImpl<decltype(predicate)>(kj::mv(predicate));
```

PR #1446, opened 2022-03-24
Make threads move-constructible and move-assignable

jrsala-auguration:
`std::thread`s are movable and `kj::Thread`s are internally very simple so there's no reason for them not to be as well.

This restores `thread-test.c++` in the list of test files, but I suppose there's a good reason it wasn't there in the first place so this can be removed.

For the implementation of `Thread& Thread::operator=(Thread&&)`, in the absence of a `std::swap` equivalent, I manually typed the swaps, which is verbose.

### DIFF
--- a/c++/src/kj/CMakeLists.txt
+++ b/c++/src/kj/CMakeLists.txt
@@ -241,6 +241,7 @@ if(BUILD_TESTING)
     io-test.c++
     mutex-test.c++
     time-test.c++
+    thread-test.c++
     threadlocal-test.c++
     test-test.c++
     std/iostream-test.c++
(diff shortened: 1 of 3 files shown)


kentonv:
Sorry for the slow reply, I've been out for a few weeks.

Thanks for contributing! However, the decision not to make thread objects movable was an intentional style choice. In [KJ style](https://github.com/capnproto/capnproto/blob/master/style-guide.md#value-types-vs-resource-types), we generally only make pure data structures ("value types") movable. "Resource types" like `Thread` do not have move constructors, instead if you want to be able to pass ownership around, you're expected to allocate the object on the heap and pass `Own<Thread>`.

PR #1447, opened 2022-03-25
Support for bazel build (including zlib and openssl deps)

min-tian-thomas:
Support for vscode settings with bazel

Split bazel BUILD files


min-tian-thomas:
Hi @uhthomas,

I've removed custom build for clang yet I don't have local M1 Pro build env to verify and we might have to change the workflow to add support for different platforms with Bazel to build and test everything later?

Thanks,
Min

BUILD
```diff
@@ -0,0 +1,522 @@
...
+    copts = [
+        "-Wno-sign-compare",
+        "-Wno-strict-aliasing",
+        "-Wno-shadow",
+        "-Wno-unused-parameter",
+    ],
+    defines = [
+        "KJ_USE_FUTEX=0",
+    ],
+    includes = ["c++/src"],
```
uhthomas, line 77: Generally the use of the `includes` attribute is discouraged.

We're builting capnp with Bazel internally and found that builds would fail in some cases where capnp was already installed on the host system.

Lots of

```
In file included from external/com_github_capnproto_capnproto/capnp/compiler/parser.c++:26:
In file included from external/com_github_capnproto_capnproto/capnp/compiler/parser.h:28:
In file included from external/com_github_capnproto_capnproto/capnp/compiler/error-reporter.h:24:
external/com_github_capnproto_capnproto/capnp/compiler/../common.h:148:20: error: redefinition of 'Kind_<unsigned char>'
template <> struct Kind_<uint8_t> { static constexpr Kind kind = Kind::PRIMITIVE; };
                   ^~~~~~~~~~~~~~
/usr/local/include/capnp/common.h:148:20: note: previous definition is here
template <> struct Kind_<uint8_t> { static constexpr Kind kind = Kind::PRIMITIVE; };
                   ^
```

We had to use explicit `isystem` include directives.

```
copts = [
    "-Iexternal/com_github_capnproto_capnproto",
    # ...
]
```

This shouldn't be necessary, but is because capnp is using angled brackets rather than quotes...

```
In file included from external/com_github_capnproto_capnproto/capnp/schema-loader.c++:23:
external/com_github_capnproto_capnproto/capnp/schema-loader.h:25:10: error: 'kj/memory.h' file not found with <angled> include; use "quotes" instead
#include <kj/memory.h>
         ^~~~~~~~~~~~~
         "kj/memory.h"
```

I'll create an issue for this.

BUILD
```diff
@@ -0,0 +1,522 @@
...
+        "c++/src/kj/refcount.c++",
+        "c++/src/kj/source-location.c++",
+        "c++/src/kj/string.c++",
+        "c++/src/kj/string-tree.c++",
+        "c++/src/kj/table.c++",
+        "c++/src/kj/test-helpers.c++",
+        "c++/src/kj/thread.c++",
+        "c++/src/kj/time.c++",
+        "c++/src/kj/units.c++",
+    ] + [
```
uhthomas, line 30: ?

BUILD
```diff
@@ -0,0 +1,522 @@
...
+        "c++/src/kj/table.h",
+        "c++/src/kj/thread.h",
+        "c++/src/kj/threadlocal.h",
+        "c++/src/kj/time.h",
+        "c++/src/kj/tuple.h",
+        "c++/src/kj/units.h",
+        "c++/src/kj/vector.h",
+        "c++/src/kj/windows-sanity.h",
+    ],
+    copts = [
```
uhthomas, line 68: We found that we had to explicitly target C++ 14 features with this copt:

```starlark
copts = [
    "-std=c++14",
]
```

BUILD
```diff
@@ -0,0 +1,522 @@
...
+        "-Wno-shadow",
+        "-Wno-unused-parameter",
+        "-iquote c++/src/kj",
+    ],
+    include_prefix = "capnp",
+    visibility = ["//visibility:public"],
+    deps = [
+        "//:kj",
```
uhthomas, line 341: All the deps for these targets should be revisited. This library does not directly depend on `//:kj`.

BUILD
```diff
@@ -0,0 +1,522 @@
...
+        "//:kj-async",
+    ],
+)
+
+cc_import(
```
uhthomas, line 346: Why use cc_import? Why not just use cc_library?

BUILD
```diff
@@ -0,0 +1,522 @@
...
+    include_prefix = "capnp",
+    visibility = ["//visibility:public"],
+    deps = [
+        ":capnp",
+        "//:kj",
+    ],
+)
+
+cc_binary(
+    name = "capnp_tool",
```
uhthomas, line 488: `bin` or `capnp_bin` is probably a better name. I didn't realise what this was at first because of `tool`.

BUILD
```diff
@@ -0,0 +1,522 @@
...
+cc_import(
+    name = "capnp-json_schemas",
+    hdrs = [
+        "c++/src/capnp/compat/json.h",
+    ],
+    visibility = ["//visibility:public"],
+)
+
+cc_library(
+    name = "capnp-json",
```
uhthomas, line 401: There's both hyphens and underscores. Would we be able to make these all underscores for consistency?

WORKSPACE
```diff
@@ -0,0 +1,53 @@
...
+http_archive(
+    name = "openssl",
+    build_file = "//bazel/builds:openssl.BUILD",
+    sha256 = "6b2d2440ced8c802aaa61475919f0870ec556694c466ebea460e35ea2b14839e",
+    strip_prefix = "openssl-OpenSSL_1_1_1n",
+    urls = ["https://github.com/openssl/openssl/archive/refs/tags/OpenSSL_1_1_1n.tar.gz"],
+)
+
+
+git_repository(
```
uhthomas, line 24: Avoid git_repository where possible. Prefer http_archive instead.

https://docs.bazel.build/versions/main/external.html#repository-rules

uhthomas (changes requested):

kentonv:
Hey, sorry I haven't responded to this yet, I've been on parental leave and other things and this just hasn't managed to get to the top of my inbox. But I am actually interested in Bazel support -- in fact it's probably the only other build system I'd allow into the repo. :) So I'm happy you submitted this!

One quick comment: I believe it's more idiomatic with Bazel to have a BUILD file in each directory covering the files in that directory, rather than a single top-level BUILD file, right? Can we do that? Or at least have `capnp/BUILD` and `kj/BUILD`?

uhthomas:
>I believe it's more idiomatic with Bazel to have a BUILD file in each directory

I agree with this too.

min-tian-thomas:
> Hey, sorry I haven't responded to this yet, I've been on parental leave and other things and this just hasn't managed to get to the top of my inbox. But I am actually interested in Bazel support -- in fact it's probably the only other build system I'd allow into the repo. :) So I'm happy you submitted this!
> 
> One quick comment: I believe it's more idiomatic with Bazel to have a BUILD file in each directory covering the files in that directory, rather than a single top-level BUILD file, right? Can we do that? Or at least have `capnp/BUILD` and `kj/BUILD`?

Hey @kentonv, 

I was doing this in a separate way previously, yet I was thinking people may would like to use a single BUILD file as their custom build file there to make changes easily if they want to, so I merged all of them together.

I'd prefer to split them into subdirs later once things can be merged and the build workflow can be checked later. Meanwhile, the main purpose is not just build capnp with Bazel, but also use Bazel to create code generator.

Sorry I got a bit delay to reply to both you and @uhthomas here during these busy days. Let me take some changes and make it better.

min-tian-thomas:
> > I believe it's more idiomatic with Bazel to have a BUILD file in each directory
> 
> I agree with this too.

I remember the reason why I put everything all together is because we have quote includes with different patterns, eg:
```
#include "exception.h"
#include "kj/exception.h"
#include <kj/exception.h>
```
Bazel actually suggested following the same absolute path to include, eg:
```
#include "kj/exception.h"
#include <kj/exception.h>
```
As it builds everything from execution root and add "dynamic"/"virtual" paths into -I or -iquote to build command:
```
-Ibazel-out/k8-fastbuild/bin/c++/src/kj/_virtual_includes/kj 
```
So it is not friendly to those flexible quote includes with different relative roots.

min-tian-thomas:
@uhthomas @kentonv I am looking at another PR for replacing the quotes to brackets includes. 
We don't have to replace all of them but we need to follow a pattern.
1. For the same lib, eg: inside lib kj, we can include quotes with relative paths
2. For different libs, if they want to cross link to other libs' header, they need to use `<>` and start with the same "system" include path pattern, eg: all start from `<kj/....>`
Currently the main issue there is for quote includes, various different patterns make things difficult and not very standard.

uhthomas:
@min-tian-thomas https://github.com/capnproto/capnproto/pull/1451 might be relevant?

min-tian-thomas:
> @min-tian-thomas #1451 might be relevant?
Yes, this is the thing I mentioned, we need to follow a rule for includes and this is the reason why I merged everything all together in my initial check-in.

For cc_lib, if I put something like:
```
cc_library(
name = 'kj',
srcs=[....],
hdrs=["exception.h"]
include_prefix = "kj",
)

Then I am expecting others (libs or apps) to include my lib like `#include <kj/exception.h>`
Yet we have "excpetion.h" and "kj/exception.h" or "../exception.h" 
For internals, we'd better use absolute path, but this is not something we can easily change everywhere.

**Comment by min-tian-thomas** [comment]
I've got one branch https://github.com/min-tian-thomas/capnproto/tree/bazel-support-split-builds
that split BUILD into multiple ones, however, because we need to support both `#include <kj/exception.h>` and `#include "exception.h"` we need to add `-iquote` from external usage:
```
cc_binary(
    name = "cc_example",
    srcs = ["main.cpp"],
    copts = [
        "-iquote external/capnproto/c++/src/kj",    
    ],
    deps = [":address_book"],
)
```

kentonv:
Hi @min-tian-thomas,

We recently had the need to add Bazel support for use inside Cloudflare. We appreciate your work here, but it ended up making the most sense for us to write the BUILD files ourselves targeting our specific needs. So, the repo now contains Bazel BUILD files.

Could you take a look at them and see if they fit your needs as well? If not, let's figure out what needs to be different and start a new PR to make those changes.

In the meantime, this PR seems obsolete so I'm going to close it.

PR #1451, opened 2022-04-22
Replace angle bracket includes with quotes

uhthomas:
Bazel treats angle bracket includes as system headers, and therefore
does not include them in the search path.

https://docs.bazel.build/versions/main/bazel-and-cpp.html#include-paths

https://stackoverflow.com/questions/21593/what-is-the-difference-between-include-filename-and-include-filename

Changing the angle bracket includes to quoted includes removes the need
for explicit -isystem (or -I) copts.

Commands used:

```
sd '#include <capnp(.*)>' '#include "capnp$1"' $(fd -t f)
sd '#include <kj(.*)>' '#include "kj$1"' $(fd -t f)
```

### DIFF
--- a/c++/samples/addressbook.c++
+++ b/c++/samples/addressbook.c++
@@ -37,8 +37,8 @@
 // TODO(test):  Needs cleanup.
 
 #include "addressbook.capnp.h"
-#include <capnp/message.h>
-#include <capnp/serialize-packed.h>
+#include "capnp/message.h"
+#include "capnp/serialize-packed.h"
 #include <iostream>
 
 using addressbook::Person;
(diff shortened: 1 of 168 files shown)


kentonv:
Sorry, I don't agree with this. Quotes are for paths relative to the file where the include appears. Angle brackets are for absolute paths, i.e. starting from the root of the source tree (`c++/src` in the case of this repo). If we use quotes for both, it can lead to weird ambiguities.

min-tian-thomas:
> Sorry, I don't agree with this. Quotes are for paths relative to the file where the include appears. Angle brackets are for absolute paths, i.e. starting from the root of the source tree (`c++/src` in the case of this repo). If we use quotes for both, it can lead to weird ambiguities.

Hi @kentonv, @uhthomas,

Bazel makes lib's public headers and private ones more explicit and you just pub public ones into `hdrs=[....]` and private ones as part of srcs. 
For the same lib, headers can reference with quotes with a relative path, which also works in Bazel, as lib's BUILD dir will be included with `-I` but relative path like `#include "../exception.h"` is not that friendly to Bazel people have to add special `-I` flags to copts.
For different libs that cross use other lib's public header, recommend using angle brackets with an absolute path starting from `<kj/...>` would be good enough. 
So in general for public headers using them with angle brackets with absolute paths. For private ones, use double quotes that are relative paths. 

Bazel is not that friendly to traditional double quotes including mode if they are not relative to workspace dir and this is one reason why I kept everything in one BUILD file as I got troubles with those includes issues.

min-tian-thomas:
Changing from double quotes to angle brackets for public lib headers would make sense to me. It does not bring any problem. For  double quotes includes for hdrs in the same lib or private hdrs we can visit case by case.

PR #1454, opened 2022-05-06
Implement http CONNECT

jasnell:
This is the start of an implementation of CONNECT support for kj http. ~~It's incomplete and fairly rough still. Opening now to start collecting early feedback on the approach here before spending a ton more time on it.~~

### DIFF
--- a/c++/src/kj/compat/http-test.c++
+++ b/c++/src/kj/compat/http-test.c++
@@ -58,8 +58,15 @@ namespace {
 KJ_TEST("HttpMethod parse / stringify") {
 #define TRY(name) \
   KJ_EXPECT(kj::str(HttpMethod::name) == #name); \
-  KJ_IF_MAYBE(parsed, tryParseHttpMethod(#name)) { \
-    KJ_EXPECT(*parsed == HttpMethod::name); \
+  KJ_IF_MAYBE(parsed, tryParseHttpMethodAllowingConnect(#name)) { \
+    KJ_SWITCH_ONEOF(*parsed) { \
+      KJ_CASE_ONEOF(method, HttpMethod) { \
+        KJ_EXPECT(method == HttpMethod::name); \
+      } \
+      KJ_CASE_ONEOF(method, HttpConnectMethod) { \
+        KJ_FAIL_EXPECT("http method parsed as CONNECT", #name); \
+      } \
+    } \
   } else { \
     KJ_FAIL_EXPECT("couldn't parse \"" #name "\" as HttpMethod"); \
   }
@@ -73,6 +80,10 @@ KJ_TEST("HttpMethod parse / stringify") {
   KJ_EXPECT(tryParseHttpMethod("GE") == nullptr);
   KJ_EXPECT(tryParseHttpMethod("GET ") == nullptr);
   KJ_EXPECT(tryParseHttpMethod("get") == nullptr);
+
+  KJ_EXPECT(KJ_ASSERT_NONNULL(tryParseHttpMethodAllowingConnect("CONNECT"))
+      .is<HttpConnectMethod>());
+  KJ_EXPECT(tryParseHttpMethod("connect") == nullptr);
 }
 
 KJ_TEST("HttpHeaderTable") {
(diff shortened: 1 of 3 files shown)


c++/src/kj/compat/http.h
```diff
@@ -52,10 +52,7 @@ namespace kj {
   MACRO(PURGE) \
   MACRO(OPTIONS) \
   MACRO(TRACE) \
-  /* standard methods */ \
-  /* */ \
-  /* (CONNECT is intentionally omitted since it should be handled specially in HttpServer) */ \
```
kentonv, line 57: It's kind of precarious to add CONNECT here because it means we have to be very careful that the application cannot use it. For example, `HttpClient::request()` really needs to throw if given `CONNECT` as the method, otherwise request smuggling vulnerabilities may open up. It's also dangerous if an HTTP server tries to process a `CONNECT` request as if it were a generic method.

I do see that this creates a problem for `HttpHeaders::tryParseRequest()`... we would want that to support parsing CONNECT requests, too. Maybe the solution is to extend the `RequestOrProtocolError` `OneOf` to include `ConnectRequest` as a completely separate type? That way no one can accidentally handle `CONNECT` the same as everything else.

kentonv, line 57: It's kind of precarious to add CONNECT here because it means we have to be very careful that the application cannot use it. For example, `HttpClient::request()` really needs to throw if given `CONNECT` as the method, otherwise request smuggling vulnerabilities may open up. It's also dangerous if an HTTP server tries to process a `CONNECT` request as if it were a generic method.

I do see that this creates a problem for `HttpHeaders::tryParseRequest()`... we would want that to support parsing CONNECT requests, too. Maybe the solution is to extend the `RequestOrProtocolError` `OneOf` to include `ConnectRequest` as a completely separate type? That way no one can accidentally handle `CONNECT` the same as everything else.

c++/src/kj/compat/http.h
```diff
@@ -662,9 +659,20 @@ class HttpClient {
...
   // stack-allocated).
 
-  virtual kj::Promise<kj::Own<kj::AsyncIoStream>> connect(kj::StringPtr host);
-  // Handles CONNECT requests. Only relevant for proxy clients. Default implementation throws
-  // UNIMPLEMENTED.
+  struct ConnectResponse {
+    uint statusCode;
+    kj::StringPtr statusText;
+    const HttpHeaders* headers;
+    kj::Own<kj::AsyncIoStream> io;
```
kentonv, line 666: If the server response with an error, is the connection still established? I'd expect not. Can the error response contain a regular HTTP response body? If so then I think we need a `OneOf` here to allow for that, like in `WebSocketResponse`.

kentonv, line 666: If the server response with an error, is the connection still established? I'd expect not. Can the error response contain a regular HTTP response body? If so then I think we need a `OneOf` here to allow for that, like in `WebSocketResponse`.

c++/src/kj/compat/http.h
jasnell: Makes sense. Ok, I'll take another pass.

c++/src/kj/compat/http.h
jasnell: If the server responds with an error, no the connection is not established, and yes, the error response can contain a regular HTTP response body.

Specifically from the spec:

```
   Any 2xx (Successful) response indicates that the sender (and all
   inbound proxies) will switch to tunnel mode immediately after the
   blank line that concludes the successful response's header section;
   data received after that blank line is from the server identified by
   the request-target.  Any response other than a successful response
   indicates that the tunnel has not yet been formed and that the
   connection remains governed by HTTP.
```

Which is fun given that a 204 No Content, etc will still indicate that there *will* be an ongoing payload despite the normal rules for those responses. If an error response is returned, then the connection is still an HTTP connection and HTTP rules still apply.

c++/src/kj/compat/http.h
jasnell: Adding a new option to `RequestOrProtocolError` is going to have a trickle down effect on the rest of the public API so I want to make sure it's what we want. Specifically, on the `HttpInputStream` API, the `readRequest()` method current returns a `kj::Promise<Request>`. We'll need that to learn the difference between a regular request and a connect request. There are two options:

1. We modify the `HttpInputStream::Request::method` field to be `kj::OneOf<HttpMethod, HttpConnectMethod>` (`HttpConnectMethod` is new)
2. We modify the return value of `HttpInputStream::readRequest()` to be `kj::Promise<kj::OneOf<Request, ConnectRequest>>`

Do you have a preference?

c++/src/kj/compat/http.h
kentonv: OK, so we do indeed need this to be:

`kj::OneOf<kj::Own<kj::AsyncIoStream>, kj::Own<kj::AsyncInputStream>> connectionOrErrorBody`

The latter would be used in the case of an error response.

c++/src/kj/compat/http.h
kentonv: Yes, #3 is the right answer. This is a public interface meant as a helper for certain oddball applications that aren't actually HTTP servers but want to read HTTP-shaped requests. Such applications are getting an error thrown today when the request is CONNECT, so continuing to throw an error at them is fine.

If some app actually wants to use this interface to receive CONNECT requests, then we can add a new method for them to be able to do that. But for now we should not bother.

`HttpServer` itself does not use the public `HttpInputStream` interface, so we can add CONNECT support without changing that interface.

c++/src/kj/compat/http.h
jasnell: See below. I've change the `connect()` method here to return `kj::Promise<kj::OneOf<ConnectResponse, Response>>`. Returning `Response` is an indication that the tunnel has not been established and the response is still HTTP rules. Returning `ConnectResponse` indicates that we're proceeding with the tunnel.

c++/src/kj/compat/http.h
```diff
@@ -722,9 +758,30 @@ class HttpService {
   // Request processing can be canceled by dropping the returned promise. HttpServer may do so if
   // the client disconnects prematurely.
 
-  virtual kj::Promise<kj::Own<kj::AsyncIoStream>> connect(kj::StringPtr host);
-  // Handles CONNECT requests. Only relevant for proxy services. Default implementation throws
-  // UNIMPLEMENTED.
+  class ConnectTunnel {
+  public:
+    virtual kj::Own<kj::AsyncIoStream> accept(const HttpHeaders& headers) = 0;
```
jasnell, line 763: No status code or status text here because both are largely meaningless when establishing a tunnel. Any 2xx code is interpreted as success. Any other code means we're still operating under HTTP rules, which is addressed using the `reject()` method below (which does accept statusCode and statusText.

c++/src/kj/compat/http.h
kentonv: Can we be consistent with how `openWebSocket()` works?

c++/src/kj/compat/http.h
kentonv: What if a particular application wants to use different 2xx codes to signal something to the client?

Or for example in a proxying scenario, if the origin server returned 247 (made up number), I don't think we'd want the proxy to silently change the code to 200.

So I think we need to support specifying the status code here, unfortunately (and throw if it isn't 2xx).

c++/src/kj/compat/http.h
jasnell: Yep. Will get that updated next week.

c++/src/kj/compat/http.h
jasnell: Ok. Not certain how likely that is but there's no harm in it.

jasnell:
Ok, this is updated per the feedback and should be compiling now. Still need to add tests, docs, and generally verify that things are working. Want to make sure first that the API is what you'd want to see @kentonv

c++/src/kj/compat/http.h
```diff
@@ -858,6 +919,14 @@ WebSocketPipe newWebSocketPipe();
 // end. No buffering occurs -- a message send does not complete until a corresponding receive
 // accepts the message.
 
+struct ConnectPipe {
+  kj::Own<kj::AsyncIoStream> ends[2];
+};
+
+ConnectPipe newConnectPipe();
```
kentonv, line 926: How is this different from `kj::newTwoWayPipe()`?

c++/src/kj/compat/http.c++
```diff
@@ -3714,13 +4076,96 @@ public:
...
+            -> ConnectResponse {
+      KJ_SWITCH_ONEOF(responseOrProtocolError) {
+        KJ_CASE_ONEOF(response, HttpHeaders::Response) {
+          auto& responseHeaders = httpInput.getHeaders();
+          if (response.statusCode < 200 || response.statusCode >= 300) {
+            // Any statusCode that is not in the 2xx range in interpreted
+            // as an HTTP response. Any status code in the 2xx range is
+            // interpreted as a successful CONNECT response.
+            closed = true;
+            kj::Own<kj::AsyncInputStream> body = heap<NullInputStream>();
```
kentonv, line 4110: Don't we want to return the actual response body here?

c++/src/kj/compat/http.c++
```diff
@@ -3714,13 +4076,96 @@ public:
...
+              httpOutput.flush());
+
+          ConnectResponse result {
+            .statusCode = response.statusCode,
+            .statusText = response.statusText,
+            .headers = &responseHeaders,
+            .connectionOrBody = kj::mv(io),
+          };
+
+          if (fastCaseCmp<'c', 'l', 'o', 's', 'e'>(
```
kentonv, line 4133: I don't think `Connection: close` is meaningful for a successful CONNECT response? Obviously the connection can't be used for any more HTTP requests either way.

c++/src/kj/compat/http.c++
```diff
@@ -3714,13 +4076,96 @@ public:
...
+          } else {
+            // Another request was already queued after this one, so we don't want to watch for
+            // stream closure because we're fully expecting another response.
+          }
+          return result;
+        }
+        KJ_CASE_ONEOF(protocolError, HttpHeaders::ProtocolError) {
+          closed = true;
+          auto response = handleProtocolError(protocolError);
+          kj::Own<kj::AsyncInputStream> body = heap<NullIoStream>();
```
kentonv, line 4147: NullIoStream -> NullInputStream?

c++/src/kj/compat/http.c++
```diff
@@ -3714,13 +4076,96 @@ public:
...
+              .statusCode = response.statusCode,
+              .statusText = response.statusText,
+              .headers = &responseHeaders,
+              .connectionOrBody = kj::mv(body),
+            };
+          }
+
+          auto& stream = *ownStream;
+          kj::Own<kj::AsyncIoStream> io = heap<ConnectPipeEnd>(
+              kj::Own<kj::AsyncIoStream>(&stream, kj::NullDisposer::instance),
```
kentonv, line 4121: I don't think you can just use the underlying `stream` here because you'll miss any bytes that were left over in the header buffer.

c++/src/kj/compat/http.c++
```diff
@@ -4077,6 +4556,28 @@ public:
...
   }
 
+  kj::Promise<ConnectResponse> connect(kj::StringPtr host, const HttpHeaders& headers) override {
+    auto hostCopy = kj::str(host);
+    auto headersCopy = headers.clone();
+
+    // TODO(now): Verify that this is correct.
+    auto hostHeader = KJ_REQUIRE_NONNULL(headersCopy.get(HttpHeaderId::HOST),
+                                         "Host header is required for connect");
+    auto parsed = Url(kj::str("http"),
```
kentonv, line 4566: Maybe easier to do:

```
Url parsed;
parsed.scheme = kj::str("http");
parsed.host = kj::str(hostHeader)`
```

c++/src/kj/compat/http.c++
```diff
@@ -4077,6 +4556,28 @@ public:
     return getClient(parsed).openWebSocket(path, headersCopy);
   }
 
+  kj::Promise<ConnectResponse> connect(kj::StringPtr host, const HttpHeaders& headers) override {
```
kentonv, line 4559: I think `connect()` may not be meaningful for `NetworkHttpClient`.

`NetworkHttpClient` is an `HttpClient` which connects directly to the server specified in the URL. I.e. it's used specifically when there is no proxy. The CONNECT method, meanwhile, only makes sense with proxies.

If anything, `NetworkHttpClient::connect()` should actually make a TCP connection to the given host, and not use HTTP at all. I am not sure, though, if that's actually useful to implement, vs. just always throwing an exception.

c++/src/kj/compat/http.c++
```diff
@@ -4746,6 +5271,153 @@ private:
     kj::Own<kj::PromiseFulfiller<HttpClient::WebSocketResponse>> fulfiller;
     kj::Promise<void> task = nullptr;
   };
+
+  class DelayedEofIoStream final: public kj::AsyncIoStream {
```
kentonv, line 5275: `tryPumpFrom()` is missing from this class.

c++/src/kj/compat/http.c++
```diff
@@ -4746,6 +5271,153 @@ private:
...
+    void shutdownWrite() override {
+      inner->shutdownWrite();
+    }
+
+  private:
+    kj::Own<kj::AsyncIoStream> inner;
+    kj::Maybe<kj::Promise<void>> completionTask;
+
+    template <typename T>
+    kj::Promise<T> wrap(T requested, kj::Promise<T> innerPromise) {
```
kentonv, line 5315: I wonder if we could reuse `DelayedEofInputStream` rather than reproduce this logic. You could set up a `DelayedEofInputStream` wrapping the input side of `inner`, forward all read calls to that, while forwarding write calls directly to `inner`.

c++/src/kj/compat/http.c++
```diff
@@ -5049,6 +5762,49 @@ private:
...
+          auto maybeService = factory(suspendable);
+
+          if (suspended) {
+            return false;
+          }
+
+          auto service = KJ_ASSERT_NONNULL(kj::mv(maybeService),
+              "SuspendableHttpServiceFactory did not suspend, but returned nullptr.");
+
+          // TODO(perf): If the client disconnects, should we cancel the response? Probably, to
```
kentonv, line 5780: It looks like you copied this comment from the non-connect path. I think the comment is actually obsolete -- we actually cancel out the whole loop if `whenWriteDisconnected()` resolves.

c++/src/kj/compat/http.c++
```diff
@@ -5049,6 +5762,49 @@ private:
...
+              return kj::mv(promise);
+            }
+
+            // TODO(now): Needed?
+            if (httpOutput.isBroken()) {
+              return false;
+            }
+
+            return httpOutput.flush().then([]() mutable -> kj::Promise<bool> {
+              // TODO(now): Determine if there is any reasonable path to reusing
```
kentonv, line 5802: Certainly not if the connection was accepted.

If it was rejected, you could arguably accept other requests after sending the rejection. However, this may be asking for trouble if the client gets confused and thinks a connection was opened. I think we should make sure to send `Connection: close` on all rejections of CONNECT requests, and then always close the connection.

c++/src/kj/compat/http.c++
```diff
@@ -5204,65 +5960,77 @@ private:
...
+          // No entity-body.
+          httpOutput.finishBody();
+          return heap<HttpNullEntityWriter>();
+        } else KJ_IF_MAYBE(s, expectedBodySize) {
+          return heap<HttpFixedLengthEntityWriter>(httpOutput, *s);
+        } else {
+          return heap<HttpChunkedEntityWriter>(httpOutput);
+        }
+      }
+      KJ_CASE_ONEOF(m, HttpConnectMethod) {
```
kentonv, line 6017: I think this shouldn't be possible to reach? We are in the `HttpService::Response::send()` callback here, but on CONNECT requests we don't give `HttpService::Response` to the application in the first place, so it can' be calling this.

c++/src/kj/compat/http.c++
```diff
@@ -5204,65 +5960,77 @@ private:
...
 
   kj::Own<WebSocket> acceptWebSocket(const HttpHeaders& headers) override {
     auto& requestHeaders = httpInput.getHeaders();
     KJ_REQUIRE(requestHeaders.isWebSocket(),
         "can't call acceptWebSocket() if the request headers didn't have Upgrade: WebSocket");
 
-    auto method = KJ_REQUIRE_NONNULL(currentMethod, "already called send()");
+    auto method = KJ_REQUIRE_NONNULL(
+          KJ_REQUIRE_NONNULL(currentMethod, "already called send()")
+              .tryGet<HttpMethod>(), "WebSockets are not supported when using CONNECT");
```
kentonv, line 6033: The very next check below is verifying the method was GET. Maybe that check should be modified instead of adding an extra check here?

c++/src/kj/compat/http.c++
jasnell: Yep, not meaningful at all. The question is, however, should we just ignore this case or treat it as an error?

c++/src/kj/compat/http.c++
jasnell: Just removing this implementation as suggested.

c++/src/kj/compat/http.c++
jasnell: For now, going to just ignore the header here if it exists.

c++/src/kj/compat/http.c++
```diff
@@ -5380,6 +6085,43 @@ private:
...
+    KJ_REQUIRE(method.is<HttpConnectMethod>(), "Only use accept() with CONNECT requests.");
+
+    KJ_REQUIRE(statusCode >= 200 && statusCode < 300, "the statusCode must be 2xx for accept.");
+    tunnelRejected = nullptr;
+
+    httpOutput.writeHeaders(headers.serializeResponse(statusCode, statusText));
+
+    auto releasedBuffer = httpInput.releaseBuffer();
+    return heap<ConnectPipeEnd>(
+        kj::Own<kj::AsyncIoStream>(&stream, kj::NullDisposer::instance),
```
kentonv, line 6102: Do we want to do something like the `deferNoteClosed` mechanism that we do in `acceptWebSocket()`?

c++/src/kj/compat/http.c++
```diff
@@ -3714,18 +4077,105 @@ public:
...
+    closeWatcherTask = nullptr;
+
+    // Mark upgraded for now even though the tunnel could fail, because we can't allow pipelined
+    // requests in the meantime.
+    upgraded = true;
+
+    kj::StringPtr connectionHeaders[HttpHeaders::CONNECTION_HEADERS_COUNT];
+    kj::String lengthStr;
+
+    // TODO(now): Make sure that forbidden headers are not included...
```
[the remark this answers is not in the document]
jasnell: There are none. I left that as a reminder to go double check.

c++/src/kj/compat/http-test.c++
```diff
@@ -4694,5 +4705,295 @@ KJ_TEST("HttpServer.listenHttp() doesn't prematurely terminate if an accepted co
...
+  HttpHeaderTable clientHeaders;
+  // Initiates a CONNECT with the echo server. Once established, sends a bit of data
+  // and waits for it to be echoed back.
+  client->connect("https://example.org"_kj, HttpHeaders(clientHeaders))
+      .then([&](HttpClient::ConnectResponse response) {
+    KJ_ASSERT(response.statusCode == 200);
+    KJ_ASSERT(response.statusText == "OK"_kj);
+
+    auto& io = KJ_ASSERT_NONNULL(response.connectionOrBody.tryGet<kj::Own<kj::AsyncIoStream>>());
+    io->write("hello", 5).poll(waitScope);
```
kentonv, line 4827: Why `.poll()` rather than `.wait()`? This will cancel the promise if it doesn't complete immediately, which seems weird.

c++/src/kj/compat/http.c++
```diff
@@ -1729,94 +1825,110 @@ static_assert(!fastCaseCmp<'f','O','o','B'>("FooB1"), "");
...
-    if (type->startsWith("multipart/byteranges")) {
-      KJ_FAIL_REQUIRE(
-          "refusing to handle multipart/byteranges response without transfer-encoding nor "
-          "content-length due to ambiguity between RFC 2616 vs RFC 7230.");
+      // #7
+      return kj::heap<HttpConnectionCloseEntityReader>(*this);
+    }
+    KJ_CASE_ONEOF(m, HttpConnectMethod) {
+      // While the CONNECT method does not generally have a body entity to read in the
+      // normal case, when a CONNECT request fails and the response is an error, the
```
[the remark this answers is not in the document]
kentonv: You've factored out some of the inner logic here but the outer logic is still largely copied from above. I don't think this logic needs to differ for connect vs. non-connect at all, except obviously for the place where it checks if the method is HEAD.

c++/src/kj/compat/http.c++
```diff
@@ -5380,6 +6085,50 @@ private:
...
+    KJ_REQUIRE(statusCode < 200 || statusCode >= 300, "the statusCode must not be 2xx for reject.");
+    tunnelRejected = Maybe<kj::Promise<bool>>(true);
+
+    // We are always going to send a Connection: Close with rejections, just to be safe.
+    kj::StringPtr connectionHeaders[HttpHeaders::CONNECTION_HEADERS_COUNT];
+    connectionHeaders[HttpHeaders::BuiltinIndices::CONNECTION] = "close";
+
+    httpOutput.writeHeaders(headers.serializeResponse(
+        statusCode, statusText, connectionHeaders));
+    return heap<HttpPassthroughWriter>(httpOutput);
```
kentonv, line 6130: I think that `reject()` should take `kj::Maybe<uint64_t> expectedBodySize` like `send()` does, and construct the appropriate kind of body stream. Even though we're going to close the connection anyway, it's still polite to delimit the body as we would normally.

c++/src/kj/compat/http.c++
```diff
@@ -1768,55 +1842,123 @@ kj::Own<kj::AsyncInputStream> HttpInputStreamImpl::getEntityBody(
...
+    if (fastCaseCmp<'c','h','u','n','k','e','d'>(te.cStr())) {
       // #3¶1
       return kj::heap<HttpChunkedEntityReader>(*this);
-    } else if (fastCaseCmp<'i','d','e','n','t','i','t','y'>(te->cStr())) {
+    } else if (fastCaseCmp<'i','d','e','n','t','i','t','y'>(te.cStr())) {
       // #3¶2
       KJ_REQUIRE(type != REQUEST, "request body cannot have Transfer-Encoding other than chunked");
       return kj::heap<HttpConnectionCloseEntityReader>(*this);
-    } else {
-      KJ_FAIL_REQUIRE("unknown transfer encoding", *te) { break; }
```
kentonv, line 1779: Note that this `{ break; }` was likely important, since otherwise this exception (which is easily triggered by a remote party) would abort the process when exceptions are disabled. Another reason we should keep the code as it was.

c++/src/kj/compat/http.c++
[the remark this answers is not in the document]
jasnell: It's not clear what change you're suggesting here.

c++/src/kj/compat/http-test.c++
jasnell: ah right, I mean to leave a comment about this but forgot. Using wait() here is causing an error to be reported saying there's nothing to wait on. Wasn't quite sure what was going on here yet.

c++/src/kj/compat/http.c++
jasnell: I was curious about that. I notice there's not a similar `{ break; }` on the `Content-Length` check below. Wouldn't that have the same issue potentially?

c++/src/kj/compat/http.c++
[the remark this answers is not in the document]
jasnell: Actually, nevermind, I got it :-)

c++/src/kj/compat/http.c++
```diff
@@ -1729,33 +1826,58 @@ static_assert(!fastCaseCmp<'f','O','o','B'>("FooB1"), "");
...
+  //
+  // TODO(now): We need to decide if we are going to allow for CONNECT request
+  // message payloads or not. Per the spec, "A payload within a CONNECT request
+  // message has no defined semantics; sending a payload body on a CONNECT request
+  // might cause some existing implementations to reject the request."
+  //
+  // Here, we just choose to ignore any delimination of any payload and act as if
+  // whatever is received after the headers is just going to be part of the
+  // CONNECT tunnel. But is that safe? Should we instead active refuse to process
+  // CONNECT requests with a payload?
```
jasnell, line 1877: Check out the comment here. We need to decide how we want to handle the possibility of a CONNECT request having a payload.

jasnell: Ok, updated to have the server check on the presence of `Content-Length` and `Transfer-Encoding` headers in the request and to reject those with a `400 Bad Request` automatically.

c++/src/kj/compat/http.c++
```diff
@@ -1729,33 +1826,49 @@ static_assert(!fastCaseCmp<'f','O','o','B'>("FooB1"), "");
...
     } else if (statusCode == 204 || statusCode == 304) {
       // No body.
       return kj::heap<HttpNullEntityReader>(*this, uint64_t(0));
     }
   }
 
-  // #2 deals with the CONNECT method which is handled separately.
+  // For CONNECT requests messages, we will ignore both the Transfer-Encoding
+  // and the Content-Length here. The HttpServer will reject such requests
+  // so ignoring it here should be safe.
```
kentonv, line 1868: I would not special-case CONNECT here, but just let the rest of the logic play out. If the caller is already preventing content-length and transfer-encoding headers from getting here, we end up in the same place. If the caller later changes to remove that prohibition, then this code will produce the correct kind of stream (but it's then up to the caller to use it correctly).

c++/src/kj/compat/http.c++
```diff
@@ -1729,33 +1826,49 @@ static_assert(!fastCaseCmp<'f','O','o','B'>("FooB1"), "");
...
+  if (type == REQUEST && isConnectRequest) {
+    return kj::heap<HttpConnectionCloseEntityReader>(*this);
```
kentonv, line 1870: `HttpConnectionCloseEntityReader` is actually the wrong type here. There should be no body, so `HttpNullEntityReader` is what you want to return. Note that the logic below, if you just let it run, will produce the correct result when it hits `#6`, so that's what I think we should do.

c++/src/kj/compat/http.c++
```diff
@@ -1729,33 +1826,49 @@ static_assert(!fastCaseCmp<'f','O','o','B'>("FooB1"), "");
...
-        //   Propagate this fact with a 0 expected body length.
+        // HACK: Neither Content-Length nor Transfer-Encoding header in response to HEAD
+        //   request. Propagate this fact with a 0 expected body length.
         length = uint64_t(0);
       }
       return kj::heap<HttpNullEntityReader>(*this, length);
+    } else if (isConnectRequest && statusCode >= 200 && statusCode < 300) {
+      // #2 For any 2xx response to a CONNECT request... a client MUST ignore any
+      // Content-Length or Transfer-Encoding header fields received.
+      return kj::heap<HttpConnectionCloseEntityReader>(*this);
```
kentonv, line 1859: I think the way things are written, we never take this branch, because `HttpClientImpl` will never call `getEntityBody()` for CONNECT responses with 2xx status codes. But if we did take this branch, then returning `HttpConnectionCloseEntityReader` seems wrong, since this interferes with constructing an actual CONNECT stream. Maybe we should just assert-fail here?

c++/src/kj/compat/http.c++
```diff
@@ -3714,6 +4011,91 @@ public:
...
+      KJ_SWITCH_ONEOF(responseOrProtocolError) {
+        KJ_CASE_ONEOF(response, HttpHeaders::Response) {
+          auto& responseHeaders = httpInput.getHeaders();
+          if (response.statusCode < 200 || response.statusCode >= 300) {
+            // Any statusCode that is not in the 2xx range in interpreted
+            // as an HTTP response. Any status code in the 2xx range is
+            // interpreted as a successful CONNECT response.
+            closed = true;
+            return HttpClient::ConnectResponse {
+              .statusCode = response.statusCode,
```
kentonv, line 4047: Unfortunately, designated initializers are a C++20 feature, and we still target C++14 in this repo. This is why the MSVC build is failing. (GCC and Clang supported this syntax before it was officially in the standard.)

kentonv:
Other than the MSVC build failure and the test TODOs, looks good.

jasnell:
Squashed the fixup commits up to this point. Will be adding more for the additional tests that are still pending.

jasnell:
Hmm, I was trying to look up to see why the linux-bazel build failed here and can't seem to find the log file anywhere. Any clues?

kentonv:
@jasnell It looks like CI doesn't publish the logs. Maybe try running locally and see if it reproduces?

@mikea Is there a way to get the test failure logs from bazel CI?

mikea:
> @mikea Is there a way to get the test failure logs from bazel CI?

Yes, sure: https://github.com/capnproto/capnproto/pull/1514

jasnell:
Ok, not quite sure if this is a problem with the impl, the test, or the bazel build... this test works find in the other test variations and only fails with bazel...

```
 TEST ] http-test.c++:5669: CONNECT Server cancels read w/client
kj/compat/http-test.c++:5694: failed: expected ex.getDescription() == "abortRead() has been called" [::write(fd, buffer, size): Broken pipe == abortRead() has been called]
stack: 55c8c699bccf 55c8c69db5f6 55c8c69cb375 7f4a5d0ba074 7f4a5d0c02bd 55c8c69e201f 55c8c69ef3ff 55c8c699bfb3 55c8c69db546 55c8c69cb1f6 7f4a5d0ba074 7f4a5d0c02bd 55c8c69e201f 55c8c69ef3ff 55c8c699c438 7f4a5cfb6a7c 7f4a5cfbcacb
    ??:0: returning here
    ??:0: returning here
    ??:0: returning here
    ??:0: returning here
    ??:0: returning here
    ??:0: returning here
    ??:0: returning here
    ??:0: returning here
    ??:0: returning here
    ??:0: returning here
    ??:0: returning here
    ??:0: returning here
    ??:0: returning here
    ??:0: returning here
    ??:0: returning here
    ??:0: returning here
    ??:0: returning here
[ FAIL ] http-test.c++:5669: CONNECT Server cancels read w/client (6484 μs)
```

@kentonv @mikea ... any suggestions?

kentonv:
@jasnell Take a look at `http-socketpair-test.c++`. I think your test is failing because the error message from a broken connection is different when it's a real socket.

It looks like `http-socketpair-test.c++` is currently only actually executed in the Ekam and Bazel builds, not in automake or cmake.

Have you not been developing this under the Workers build, with Ekam?

jasnell:
> Have you not been developing this under the Workers build, with Ekam?

No no, not yet. For this particular piece I've been building independently to make sure that just standalone things work.

Ok, somehow I managed to completely miss that `KJ_HTTP_TEST_USE_OS_PIPE` define.

jasnell:
ok, passing everything exception mingw

c++/src/kj/compat/http-test.c++
```diff
@@ -5691,7 +5691,11 @@ KJ_TEST("CONNECT Server cancels read w/client") {
 
     bool failed = false;
     io->write("hello", 5).catch_([&failed](kj::Exception&& ex) {
+#if KJ_HTTP_TEST_USE_OS_PIPE
+      KJ_ASSERT(ex.getDescription() == "::write(fd, buffer, size): Broken pipe");
```
kentonv, line 5695: Perhaps it would be best to check for a DISCONNECTED exception and not rely on specific message text.

jasnell: Done. Also squashed out the fixup commit.

Issue #976, opened 2020-04-19
'fire-and-forget' type of calls


fvannee:
I've read a few of these issues about this already, but I was still wondering some things about adding some sort of 'fire-and-forget' type of call. I've created a simple server where clients can subscribe for events. They do this by passing a capability object to a `subscribe` call to the server. The capability has some sort of `onEvent` method that the client implements. Whenever an event occurs, the server will call this `onEvent` RPC method for every client that has subscribed.
Now, in my test case:
- Events occur frequently (hundreds of times per second)
- Multiple clients are subscribed to the server (10+)
- It's pretty much useless for the server to know anything about what the client did with the message. The calls return `void`. It's the client responsibility to process them in time. If it can't , it may as well be disconnected.

What I've noticed when I started increasing the message rate and/or client count in some simple examples, was that there was that the CPU load of  the server increased dramatically (more than I would expect in such situations). Some `perf` tests seemed to point to the fact that the server would spend most of its time handling the `return void` messages from all the clients it's been sending events to.

Now it's obviously possible to do this kind of load on a different connection where we don't use any of the RPC protocol features, but just use the serialization of capnp. However, when trying out something like that, I actually find myself trying to replicate many of the (simpler) features that the RPC protocol provides. I end up writing all the boilerplate code as well for simulating 'rpc' calls over a socket. That's because the `onEvent` call is just one of the methods, but the server provides other interfaces too that are just much more easy to specify in capnp rpc language. The issue is I can't mix the behavior. It's either use RPC and forget about these events, or use events but implement the whole RPC stuff myself.

Now I think most of these issues would be solved if you could mark a method in an interface as 'fire-and-forget'. Specifying you just want to call this method, but you don't expect anything to be returned (not even a void message). What are your thoughts on this? Or am I missing an obvious way to accomplish something like this already?

kentonv:
So, first of all, the RPC implementation is actually not very well-optimized... in fact I've spent no time at all trying to optimize it, so there's probably tons of low-hanging fruit. In the past it hasn't been an issue, but for some recent use cases in my day job this is starting to be a problem, so I suspect we'll be doing some optimization in the next few months.

My usual problem with fire-and-forget calls is: where do exceptions go? But I suppose in the case of a client that connects to a server and requests a subscription, if the client callback throws an exception, the server would presumably respond by having the original `subscribe()` call fail, propagating the same exception. So I guess you can argue in this case, the client might as well short-circuit the whole thing and report the exception to itself.

[...]

PR #1460, opened 2022-05-19
Realtime stream

JonasVautherin:
Implementing the idea of a `realtime stream`, as discussed in #976. The dynamic capabilities part and the tests are missing, but I figured that I would already ask for review with this milestone; no need to go further if I'm going in the wrong direction :slightly_smiling_face:.

Similar to the `stream` keyword, this can be used to signal a `realtime stream`:

```
write @0 (bytes :Data) -> realtime stream;
```

A "realtime stream" defines a `Call` that does not expect a `Return` (so the return type is always `void`). The idea is that some `VatNetwork` implementation could leverage this to send "unreliable" messages over e.g. UDP. Over a reliable transport (e.g. TCP), it just means that the RPC system does not keep track of those messages (which may be desirable in some situations?).

A "realtime call" is defined by setting `isRealtime` on the `Call`, and by using a `QuestionId` in the range `[2^31, 2^32)`. The question id just increments from 2^31 until it reaches 2^32 - 1, at which point it starts from 2^31 again. It is assumed that this is safe because that huge interval will never be full of in-flight messages.

The approach in this PR is backwards-compatible. If the caller sends a realtime call, but the callee does not know about it, then the callee will answer with a `Return`, and the caller will respond with a `Finish` (as usual). The only difference being that the caller, in this case, will not be keeping track of realtime messages.

Note that I tend to use "realtime" and "realtime stream" interchangeably (depending on what feels more readable). It feels right since a "realtime stream" is just a bunch of "realtime calls", without any flow control.

I tested the code with a small example and checked the traffic with wireshark and the [capnp wireshark plugin](https://github.com/kaos/wireshark-plugins) (which works fine, but not perfectly, as can be seen in the screenshots below). The example sends 2 realtime calls, then one `done()`, then again 3 realtime calls, and `done()` again.

The first example below shows traffic when both the client and the server know the concept of `realtime stream`:
realtime_stream

It is a bit difficult to see here, but the point is that we see `finish(0)` and `finish(1)` for the boostrap and `done()` calls, but no `finish(2^31)` for the realtime calls (because no returns were sent). Compare this to the same example, but when the server does not know about the concept of "realtime stream" (i.e. it knows that call as something like `write @0 (bytes :Data) -> ();` instead of `write @0 (bytes :Data) -> realtime stream;`).

backward_compatible

This time we see that the server sends `Return` messages for realtime calls (with an id in the 2^31 range), and the realtime-aware client still answers with `Finish`, to be backwards-compatible.


c++/src/capnp/realtime.capnp
```diff
@@ -0,0 +1,48 @@
+# Copyright (c) 2022 Cloudflare, Inc. and contributors
```
JonasVautherin, line 1: I did not really know what to put there, I just copied from `stream.capnp`. Let me know if that should be something other than `Cloudflare` :sweat_smile:

c++/src/capnp/rpc.c++
```diff
@@ -2052,7 +2121,7 @@ private:
                    kj::Array<kj::Maybe<kj::Own<ClientHook>>> capTableArray,
                    const AnyPointer::Reader& params,
                    bool redirectResults, kj::Own<kj::PromiseFulfiller<void>>&& cancelFulfiller,
-                   uint64_t interfaceId, uint16_t methodId)
+                   uint64_t interfaceId, uint16_t methodId, bool realtime = false)
```
JonasVautherin, line 2124: This is probably not the most elegant solution, but my alternative was to refactor the `Context` to have e.g. `RpcCallContext` and `RpcRealtimeCallContext`. But I wanted to keep the changes in this PR minimal, and before I do the bigger change I would like to have the opinion from a maintainer :blush:

JonasVautherin:
I realise that this is not an easy PR to review, but I would really appreciate if I could get any feedback at all :blush:.

Even just "this looks completely wrong, just give up" or "looks promising, but it will take time until someone has time to look into it" :innocent:.

kentonv:
Hey, sorry for the radio silence here. I've seen this, it sits in my Sisyphean inbox and I keep thinking "OK, as soon as I'm done reviewing all the stuff my team needs me to review then I can look at this" but inevitably, there is always something my team needs me to review. :(

This is still in my inbox and I continue to endeavor to actually clear my inbox but it's hard to say when, exactly, I'll manage to do that...

JonasVautherin:
Sure, no worries! :blush:

risa2000:
@kentonv I have been recently looking for something exactly like this, so I wonder, if it is/was accepted already, or what is the plan for merging it into the main branch?

JonasVautherin:
@risa2000: this was never reviewed, and now it's conflicting with the main branch. At this point I am not sure if I should fix the conflicts, or just give up because it hasn't gotten any attention in almost a year :man_shrugging:.

kentonv:
Ugh has it been that long?

I'm really sorry about this. I feel pretty bad that you spent time to write this, which looks like a useful feature, and the only reason it's not merged is because I haven't looked at it.

The thing is, I do dozens of code reviews every week -- mostly for my job, of course. At this moment there are 14 reviews in my inbox. Getting through those is hard, and most weeks there are several PRs that I don't look closely enough at because I just don't have time. I haven't reached inbox zero in years.

When I persistently have a dozen code reviews from my team waiting for me, it feels really hard to justify spending time on Cap'n Proto PRs that my team doesn't need. It's one thing when it's small typo fixes or whatever that I can just click merge on without thinking but when it's a PR that I know I'm going to have to spend an hour or two to understand... that hour or two never seems to present itself. And I know I cannot just mindlessly merge this one since it touches deep details of the RPC system, which is critical for my day job.

I haven't simply closed this PR because I do actually want to review it at some point. It looks like a useful feature, based on a design I suggested. I keep thinking "oh, after this next product launch gets out the door I'll have some more time". But maybe it would be more realistic for me to admit that I cannot realistically promise to review every change people offer and maybe this one is never going to reach the top of the priority queue unless it turns out to be needed by Cloudflare... :/

JonasVautherin:
@kentonv: I did realize that, and even though it hurts me to have spent so much time on it to be completely ignored, I do understand. You don't owe me anything. I promised myself that I would stop trying to contribute after this PR is done. The thing is that I cannot start using CapnProto without this feature (I can't reasonably maintain a fork).

This said:

* Instead of committing for a 2h review, do you think you could spend 10 minutes skimming through it? You may be surprised: it's mostly boilerplate (mimicking the stream feature). The actual changes in the RPC protocol are very few lines.
* Is there something I could do to make it more appealing? Like write a sample app to help run it, or point you to the exact lines that are not boilerplate? I'm pretty sure you could have a person you trust confirm that the boilerplate is boilerplate, so that you would only have to focus on the actual protocol changes (which would take you a few minutes, it's really a couple `if` statements and a list :sweat_smile:).

I started resolving the conflicts, and though I need to rewrite the `capnpc-c++.c++` code (which is generating boilerplate), the rest stays valid. I am just having a hard time finding motivation to finish it, because probably it won't get reviewed. But if you tell me that you can commit for 10 minutes to skim through it + a quick feedback, I'm happy to fix the conflicts.

ocdtrekkie:
@kentonv Is there anyone else working on capnp who would be interested/willing/qualified to review and is trusted enough to improve bus factor here?

I think a shockingly large number of problems in the world could be solved by multiplying Kenton Varda, but unfortunately as long as that remains in the realm of science fiction, more mundane solutions may be required.

kentonv:
Sorry @JonasVautherin but your description of the boilerplate actually makes me less comfortable. If you found yourself cloning a lot of the code for streams and renaming it to "realtime" then I have to wonder if there's a missing abstraction that would allow less code duplication. Like, why is `RealtimeRequest` needed? Couldn't you reuse `StreamingRequest`, since it's the same API? It seems like the realtime-ness of the request could just be a flag passed down somewhere? In fact, there's a new structure called `CallHints` (introduced after you originally opened this PR), could "realtime" just be a new boolean in that structure?

Even putting the boilerplate aside, eliding messages (which is the objective of this PR, right?) has complex implications and no matter how you implemented it I expect to need to think carefully about whether it's safe.

kentonv:
@ocdtrekkie There really isn't anyone I trust to review the addition of major new features that affect the internals of the RPC system, no.

JonasVautherin:
> If you found yourself cloning a lot of the code for streams and renaming it to "realtime" then I have to wonder if there's a missing abstraction that would allow less code duplication.

It's been a long time, but I'm certain I thought about that, and chose/had to "duplicate" it. Mostly because it comes from [`realtime.capnp`](https://github.com/capnproto/capnproto/pull/1460/files#diff-f3d0b209b9f0ab9133a94e3c9f40312f0eeb1866e8d374778b6fd4121517a7e6R29), which generates code which is used by the boilerplate. So I did not stupidly copy-paste until it worked, I carefully thought about it and it seemed like the cleaner way (and maybe you disagree, but that's another question). There is a ton of very similar "boilerplate" for `Call` and `Stream` already (that you wrote), exactly for this reason. And you did not find a way to have "stream" be a boolean in "call" :wink:.

Also there are places where it called for some small refactoring, but I went for the slightly more verbose version to reduce the diff. I learned when contributing to Signal Messenger years ago that refactoring in a PR is bad when the reviewer is a superstar with no time (in that case it was Moxie) :sweat:.

Today I have been working in fixing the conflicts, and this time tried to reduce the diff again (in capnpc-c++.c++). I could refactor to have fewer lines, but the diff would be bigger and the review more risky for you. I haven't pushed the changes yet, though (I'm scared to destroy the PR with the merge and definitely bury my chances to get a review some day :see_no_evil:).

> In fact, there's a new structure called CallHints (introduced after you originally opened this PR), could "realtime" just be a new boolean in that structure?

I could look into that, but I am not convinced and obviously it did not exist when I started in December 2021. Now I must say I'm reluctant to look into it because you sound like you clearly won't skim through it unless it becomes a need for CloudFlare (in which case I guess you would just write it yourself).

Should I just trash this and live with the fact that I lost my time?

kentonv:
> There is a ton of very similar "boilerplate" (at least that's how I call it) for Call and Stream already (that you wrote), exactly for this reason. And you did not find a way to have "stream" be a boolean in "call".

That's because they have different signatures (returning a response vs. `Promise<void>`). But regular-stream and realtime-stream have the same signature.

> Should I just trash this and live with the fact that I lost my time?

That might be best. I don't want to ask you to do a lot of rewriting just to have me fail at reviewing again. :/

JonasVautherin:
> That's because they have different signatures (returning a response vs. Promise<void>). But regular-stream and realtime-stream have the same signature.

But still you would have a separate [`realtime.capnp`](https://github.com/capnproto/capnproto/pull/1460/files#diff-f3d0b209b9f0ab9133a94e3c9f40312f0eeb1866e8d374778b6fd4121517a7e6R29) file? That's really the only way I found to add the `realtime` keyword in the parser.

JonasVautherin:
> In fact, there's a new structure called CallHints (introduced after you originally opened this PR), could "realtime" just be a new boolean in that structure?

@kentonv: seems like it was not so hard to adapt to `CallHints`, and it saved me ~150 of lines in the diff.

Let me know if there are other obvious things that I could improve :blush:.

c++/src/capnp/rpc.capnp
```diff
@@ -443,6 +443,11 @@ struct Call {
   # The call parameters.  `params.content` is a struct whose fields correspond to the parameters of
   # the method.
 
+  isRealtime @11 :Bool = false;
+  # Indicates whether or not the caller cares about receiving a `Return` for this message.  Note that
```
kentonv, line 447: I think we want this to be defined more like:

> Indicates that it is better not to deliver this message at all, than to deliver it late. The semantics are similar to UDP. Implementations are encouraged to discard realtime messages when there is not enough bandwidth to send them immediately (e.g. when they'd otherwise have to buffer the message to send later). Realtime messages can also be delivered out-of-order.
>
> Moreover, since the caller is expected to have some other means of detecting whether messages were received, the callee should not bother sending a `Return` message. For backwards compatibility with implementations that don't understand this flag, if the callee does send a `Return`, the caller must send a `Finish`; otherwise, the caller must NOT send a `Finish`. Callers generally won't keep any record of what messages were sent in the past, so to handle this backwards-compatibility scenario, callers should choose question IDs cleverly such that they can recognize based on the ID alone that a `Return` was for a realtime stream.
>
> A realtime message cannot contain capabilities, because there would be no way to know whether the capability was received, and thus no way to know whether it is still in use.
>
> An elaborate transport implementation could implement realtime calls using UDP. However, even TCP transports can take advantage of this flag to avoid excess buffering and unnecessary messages.

c++/src/capnp/capability.h
```diff
@@ -686,6 +691,9 @@ class RequestHook {
   virtual kj::Promise<void> sendStreaming() = 0;
   // Send a streaming call.
 
+  virtual kj::Promise<void> sendRealtime() = 0;
```
kentonv, line 694: I think `sendRealtime()` isn't needed anymore with the `CallHints` approach? A `CallHints` should have been specified when creating the `RequestHook`, therefore we can just use `sendStreaming()` and it'll know whether or not to treat as realtime.

c++/src/capnp/rpc.c++
```diff
@@ -3134,9 +3221,26 @@ private:
         // ahead and delete it from the table.
         questions.erase(ret.getAnswerId(), *question);
       }
-
     } else {
-      KJ_FAIL_REQUIRE("Invalid question ID in Return message.") { return; }
+      auto questionId = ret.getAnswerId();
+      if (questionId >= 1 << 31) {
```
kentonv, line 3226: At present this branch will never be reached because the `questions.isHigh(questionId)` check earlier on checks for exactly the same condition. This was added late last year to support the `onlyPromisePipeline` flag. When this flag is used, the caller is saying that the callee need not send any `Return` message because the call will only be used to perform promise pipelining. We end up with a similar situation where we don't know if we'll ever receive a `Return` so we need to be able to handle it without keeping any memory of the ID. So I used the same high-bit trick. Unfortunately, the semantics are slightly different than realtime messages: a `Finish` message is still sent, when the caller is done pipelining. Hence, when a caller receives a `Return` message with a question ID with the high bit set, it simply ignores the message, because it expects that a `Finish` will be sent whenever the pipeline is dropped. But for realtime messages, different behavior is needed: the caller must _always_ send a `Finish` message.

Unfortunately I think what we have to do here is split the "high ID" space further. I think we should use the _low_ bit to distinguish between these two modes. That way, we still only need one counter. The counter increments by 2 for each message sent, and then the low bit is set to 1 for realtime or 0 for onlyPromisePipelining.

kentonv (commented):
Instead of introducing the new `-> realtime stream` grammar and the `RealtimeResult` type, could we make this an annotation applied to regular `-> stream`? I think that would reduce the complexity of this PR quite a bit and I think it works because (1) the API signature is the same between regular and realtime streams, (2) gracefully degrading to regular streaming is probably acceptable behavior most of the time.

JonasVautherin:
@kentonv: I applied your changes, and that makes the PR much smaller! I hope it goes in the right direction :blush:.

> Instead of introducing the new -> realtime stream grammar and the RealtimeResult type, could we make this an annotation applied to regular -> stream?

It does reduce the complexity indeed, but now a `.capnp` file using the `$Cxx.realtime` annotation won't build with a version of capnp that does not know the annotation. Unless I missed something?

JonasVautherin:
Is it fine if I squash the whole commit history? I think it would be squashed if it got merged, but I wouldn't want to complicate the review process (in case squashing would break someone's review UI/workflow).

c++/src/capnp/rpc.c++
```diff
@@ -3030,6 +3079,16 @@ private:
 
     // No more using `call` after this point, as it now belongs to the context.
 
+    if (call.getIsRealtime()) {
+      auto cancelPaf = kj::newPromiseAndFulfiller<void>();
+      auto promiseAndPipeline = startCall(
```
kentonv, line 3084: Problem: `hints.isRealtime` may be true here, but `startCall()` does not know whether it's making a streaming call. If `capability` itself turns out to be an outgoing RPC client, i.e. the call is going to be proxied over another connection, then ultimately `RpcRequest::send()` will be invoked to send the call. The outgoing call will have `isRealtime = true` due to the call hint, but because `send()` is used to send it, a regular question table entry will be created, with a regular low-numbered question ID. Since it's a realtime call, the recipient may never send a `Return` back, so the question table entry will leak.

To fix this, I think `RpcRequest::send()` needs to check if the `isRealtime` bit is set and, if so, delegate to `sendStreaming()` (or maybe directly to `sendRealtimeInternal()` depending on what makes the code simpler).

Please be sure to add a test for this case, that is, arrange a proxy so that a capability crosses two connections, and then try doing realtime calls on it. You'll need some way to detect if question IDs are being leaked... maybe you can add a `RpcSystem::countQuestionsForTest()` that returns a count of all outstanding questions on all connections?

c++/src/capnp/rpc.c++
```diff
@@ -3147,6 +3206,19 @@ private:
       // that we already removed it and re-allocated the ID to something else. So, we should ignore
       // the `Return`. But we might want to make note to stop using these hints, to protect against
       // the (again, remote) possibility of our ID space wrapping around and leading to confusion.
+      if (questions.isLow(questionId)) {
+        // The low bit of the questionId is set, meaning that it is a realtime message. In this case,
+        // we must send a Finish message.
```
kentonv, line 3211: You should check if `noFinishNeeded` is set on the return message. If so, you don't need to send a Finish! The `noFinishNeeded` bit is commonly set when a return message contains no capabilities, which should always be the case for streaming calls since they don't return anything at all.

kentonv (commented):
Thanks, this is much simpler now.

When addressing these comments, please make sure to make your changes as separate commits so that I can see what changed -- don't squash them into the original commit. Once approved we can squash before merging.

kentonv:
Watch out for this:

Screenshot from 2023-06-17 15-14-10

I've found people commonly miss this little misfeature of the github UI and forget to address the hidden comments... :/

JonasVautherin:
@kentonv: This is addressing many of your points, but I'm stuck with a couple. I have free time in the next few weeks, so I would be glad if you could give me some pointers for them (without necessarily having to review all the changes) :blush:.

* First, I did not update the branch because I'm scared I will mess up the GitHub WebUI, but they compile on `master` and on `v2` (I am running the tests on `master` though). Let me know what you want me to do (stay like this, rebase on `master`, or on `v2`?).
* I would rather have the `realtime` annotation in `stream.capnp` instead of `c++.capnp`, but I did not manage to make it work. Somehow `capnpc` seems to ignore it. Is there something special I have to do when moving the annotation from `c++.capnp` to `stream.capnp`? It seems like there is some kind of optimization that prevents stream.capnp from being loaded if no stream is defined, maybe I missed something along those lines for capnpc?
* I am not sure how to translate `Promise<void>` to `RemotePromise<AnyPointer>`, but I answered in your comment above.
* Lastly, do I understand correctly that the Question IDs of the realtime streams purposely get leaked? I mean if we never receive a Finish, then we never erase the element. Is that correct or did I miss something?

I hope this goes in the right direction! :crossed_fingers:

c++/src/capnp/c++.capnp
```diff
@@ -46,3 +46,5 @@ annotation allowCancellation(interface, method, file) :Void;
 # If your code is not cancellation-safe, then allowing cancellation might give a malicious client
 # an easy way to induce use-after-free or other bugs in your server, by requesting cancellation
 # when not expected.
+
+annotation realtime(method) :Void;
```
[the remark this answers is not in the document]
JonasVautherin: I agree, but I did not manage to put it in `stream.capnp`. I tried just moving it there, and also importing `namespace` from `c++.capnp` and calling `$namespace("capnp::annotations")`, but the result is always the same: capnpc compiles fine, but it just ignores the `realtime` annotation.

I am wondering if `stream.capnp` is "optimized away" when compiling capnpc, or something like this? Or do you think it should "just work" by moving this line to `stream.capnp` and I missed something else?

c++/src/capnp/rpc.c++
```diff
@@ -3732,6 +3819,24 @@ public:
...
 
+  kj::Promise<void> sendRealtime(kj::Own<OutgoingRpcMessage> message) override {
+    KJ_SWITCH_ONEOF(state) {
+      KJ_CASE_ONEOF(ignored, Running) {
+          if (isReady()) {
+            message->sendRealtime();
+            return kj::READY_NOW;
+          } else if (!hasLoggedDroppedRealtimeMessage) {
+            KJ_DBG("Dropping realtime message(s) because of link congestion");
+            hasLoggedDroppedRealtimeMessage = true;
```
JonasVautherin, line 3830: @kentonv: I don't really like this, but I wanted to show a proposition for logging it. But I guess it would be better to just silently drop the message here, especially given that the returned promise will show that the message got dropped, so the app can log it if necessary.

What do you think?

c++/src/capnp/rpc.c++
```diff
@@ -1748,6 +1757,13 @@ private:
     }
 
     RemotePromise<AnyPointer> send() override {
+      //if (callBuilder.getIsRealtime()) {
+      //  // In a situation involving a call being proxied over another connection, it could
+      //  // happen that send() is called for a realtime stream. In that case, redirect to
+      //  // sendStreaming() directly.
+      //  return RemotePromise<AnyPointer>(sendStreaming(), getDisabledPipeline());
+      //}
```
JonasVautherin, line 1765: @kentonv: This does not compile, and I am a bit stuck. It is related to [your comment here](https://github.com/capnproto/capnproto/pull/1460/files#r1233122531).

It seemed fine to just delegate to `sendStreaming()` here, but I am not sure how to convert the `Promise<void>` of `sendStreaming()` into a `RemotePromise<AnyPointer>` (because it requires a `Response`).

Any insight here? :pray:

c++/src/capnp/rpc.c++
JonasVautherin: Nevermind, I'm pretty sure that logging here was a bad idea. I removed it.

c++/src/capnp/rpc.c++
```diff
@@ -1748,6 +1764,21 @@ private:
...
+        auto streamPromise = sendStreaming().then([]() {
+          auto response = kj::Own<RpcResponse>();
+          auto reader = response->getResults();
+          return Response<AnyPointer>(reader, kj::mv(response));
+        });
+
+        return RemotePromise<AnyPointer>(
+            kj::mv(streamPromise),
+            AnyPointer::Pipeline(getDisabledPipeline()));
+      }
```
JonasVautherin, line 1780: I am trying to fix the unit test with this, but then I get a segmentation fault [here](https://github.com/capnproto/capnproto/blob/e8c7e3227b8087f0cc2634df12503a023ab11481/c%2B%2B/src/capnp/rpc.c%2B%2B#L2648) on the `connectionState->connection.is<Connected>()` call. Not sure if I am doing something wrong here, or rather [there](https://github.com/capnproto/capnproto/blob/e8c7e3227b8087f0cc2634df12503a023ab11481/c%2B%2B/src/capnp/rpc.c%2B%2B#L3104-L3113) (because it seems like the segmentation fault happens soon after that `return` I make there...).

c++/src/capnp/rpc.c++
```diff
@@ -3030,6 +3101,17 @@ private:
...
 
+    if (call.getIsRealtime()) {
+      auto promiseAndPipeline = startCall(
+          call.getInterfaceId(), call.getMethodId(), kj::mv(capability), context->addRef(), hints);
+      tasks.add(promiseAndPipeline.promise
+          .then([](){}, [](kj::Exception&& exception) {
+            KJ_LOG(ERROR, exception);
+          })
+      );
+      return;
```
JonasVautherin, line 3112: @kentonv: does that seem correct? I have the feeling that this causes the segmentation fault I mention [in the comment above](https://github.com/capnproto/capnproto/pull/1460/files#r1610736509). Somehow it feels like I should be attaching the context around here before returning? Not sure.

JonasVautherin: I think I was right in that I was missing the `.attach(kj::mv(context))` above, resulting in a segfault :+1:

c++/src/capnp/rpc.c++
JonasVautherin: Note: I fixed the segmentation fault that was indeed coming from somewhere else (:tada:), but somehow my unit test is still failing. I'll have to investigate another day, it's getting late here.

JonasVautherin: Argh I don't know. I guess probably `auto response = kj::Own<RpcResponse>();` is the reason for the segmentation fault somewhere down the line.

I'm giving up.

JonasVautherin:
I'm done. It's been more than two years, I spent a ridiculous amount of time on that thing, I don't know how the hell I should redirect `send()` to `sendStreaming()` and it most certainly won't magically come to me overnight.

Let's officially close it so that I stop hoping that maybe I'll finish it someday. It just won't happen.

PR #1463, opened 2022-05-23
CI: alpine needs bash to run the tests

JonasVautherin:
The musl build fails in https://github.com/capnproto/capnproto/pull/1460 because alpine doesn't have bash by default. I don't know why it was working before, but this fixes it.


JonasVautherin:
Oh, closing since it's fixed again in https://github.com/capnproto/capnproto/pull/1466

PR #1465, opened 2022-05-24
Work around getcontext() being broken on manylinux2014-x86.

### LINKED ISSUE #1464: Unimplemented: getcontext(context): Function not implemented
I realized in https://github.com/capnproto/capnproto/pull/1460 that one of the manylinux CI tests fails with [the following error](https://github.com/capnproto/capnproto/runs/6564241356?check_suite_focus=true):

```
[ TEST ] kj/async-test.c++:1066: cancel a fiber
kj/async.c++:1366: unimplemented: getcontext(context): Function not implemented
```

However, [the test for `makecontext`](https://github.com/capnproto/capnproto/runs/6564241356?check_suite_focus=true#step:4:253) seems to find it:

```
checking for library containing makecontext... none required
```

Just to be sure, I tried to replace `makecontext` with `getcontext` [here](https://github.com/JonasVautherin/capnproto/runs/6564083277?check_suite_focus=true#step:4:253), and it fails with the same errors.

Any idea why it works for all the other tests (including another manylinux dockcross image) and not for this one, given that autoconf apparently finds `makecontext`/`getcontext` in glibc?

Just trying to fix the CI test :innocent:.

kentonv:
Fixes #1464.

### DIFF
--- a/c++/src/kj/async-test.c++
+++ b/c++/src/kj/async-test.c++
@@ -30,6 +30,7 @@
 #endif
 
 #if KJ_USE_FIBERS && !_WIN32
+#include <errno.h>
 #include <ucontext.h>
 #endif
 
--- a/c++/src/kj/async-test.c++
+++ b/c++/src/kj/async-test.c++
@@ -29,6 +29,10 @@
 #include <pthread.h>
 #endif
 
+#if KJ_USE_FIBERS && !_WIN32
+#include <ucontext.h>
+#endif
+
 namespace kj {
 namespace {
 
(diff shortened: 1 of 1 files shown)


JonasVautherin (approved):
Seems like it fixes it! Just missing an include, I believe :innocent:

c++/src/kj/async-test.c++
```diff
@@ -29,6 +29,10 @@
 #include <pthread.h>
 #endif
 
+#if KJ_USE_FIBERS && !_WIN32
+#include <ucontext.h>
```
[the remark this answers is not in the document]
kentonv: Weird that it worked for me locally.

c++/src/kj/async-test.c++
```diff
@@ -29,6 +29,11 @@
 #include <pthread.h>
 #endif
 
+#if KJ_USE_FIBERS && !_WIN32
```
kentonv, line 32:
```suggestion
#if KJ_USE_FIBERS && __linux__
```

c++/src/kj/async-test.c++
```diff
@@ -39,6 +44,34 @@ TEST(Async, GetFunctorStartAddress) {
...
 #endif
 
+bool isLibcContextHandlingKnownBroken() {
+  // manylinux2014-x86's libc implements getcontext() to fail with ENOSYS. This is flagrantly
+  // against spec: getcontext() is not a syscall and is documented as never failing. Our configure
+  // script cannot detect this problem because it would require actually executing code to see
+  // what happens, which wouldn't work when cross-compiling. It would have been so much better if
+  // they had removed the symbol from libc entirely. But as a work-around, we will skip the tests
+  // when libc is broken.
+#if KJ_USE_FIBERS && !_WIN32
```
kentonv, line 54:
```suggestion
#if KJ_USE_FIBERS && __linux__
```

kentonv (commented):
Gotta target the work-around to Linux so we don't have to deal with convincing Mac to compile it correctly.

kentonv:
I'm going to assume the musl failure -- where it complaints that `bash` is missing -- is an unrelated problem. Yay CI whack-a-mole.

JonasVautherin:
Yeah that would be this one: https://github.com/capnproto/capnproto/pull/1463

PR #1466, opened 2022-05-24
Fix linux-musl build

kentonv:
Apparently `bash` is no longer installed by default?


kentonv:
Oops, #1463 already fixes this.

PR #1469, opened 2022-05-25
Optimize ArrayPtr::operator== for integral types

tniessen:
Make `ArrayPtr<T>::operator==` much faster when `T` is an integral type, including `const char`, which is used by `StringPtr`.

### DIFF
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -573,6 +573,19 @@ template <typename T, typename U> struct IsSameType_ { static constexpr bool val
 template <typename T> struct IsSameType_<T, T> { static constexpr bool value = true; };
 template <typename T, typename U> constexpr bool isSameType() { return IsSameType_<T, U>::value; }
 
+template <typename T> constexpr bool isIntegral() { return false; }
+template <> constexpr bool isIntegral<char>() { return true; }
+template <> constexpr bool isIntegral<signed char>() { return true; }
+template <> constexpr bool isIntegral<short>() { return true; }
+template <> constexpr bool isIntegral<int>() { return true; }
+template <> constexpr bool isIntegral<long>() { return true; }
+template <> constexpr bool isIntegral<long long>() { return true; }
+template <> constexpr bool isIntegral<unsigned char>() { return true; }
+template <> constexpr bool isIntegral<unsigned short>() { return true; }
+template <> constexpr bool isIntegral<unsigned int>() { return true; }
+template <> constexpr bool isIntegral<unsigned long>() { return true; }
+template <> constexpr bool isIntegral<unsigned long long>() { return true; }
+
 template <typename T>
 struct CanConvert_ {
   static int sfinae(T);
@@ -609,52 +622,6 @@ constexpr bool canMemcpy() {
 #define KJ_ASSERT_CAN_MEMCPY(T) \
   static_assert(kj::canMemcpy<T>(), "this code expects this type to be memcpy()-able");
 #endif
-
-namespace _ {  // private
-#if defined(__clang__) && defined(__is_identifier)
-#define KJ_HAS_BUILTIN_TYPE_TRAIT(x) (!__is_identifier(x))
-#elif defined(__has_builtin)
-#define KJ_HAS_BUILTIN_TYPE_TRAIT(x) __has_builtin(x)
-#else
-#define KJ_HAS_BUILTIN_TYPE_TRAIT(x) 0
-#endif
-
-#if KJ_HAS_BUILTIN_TYPE_TRAIT(__is_integral)
-template <typename T>
-constexpr bool canMemcmpArrayOf() {
-  return __is_integral(T);
-}
-#else
-template <typename T, typename U>
-constexpr bool isSameType() {
-#if KJ_HAS_BUILTIN_TYPE_TRAIT(__is_same)
-  return __is_same(T, U);
-#elif KJ_HAS_BUILTIN_TYPE_TRAIT(__is_same_as) || __GNUC__
-  return __is_same_as(T, U);
-#else
-  return false;
-#endif
-}
-
-#undef KJ_HAS_BUILTIN_TYPE_TRAIT  // Don't leak this to user code.
-
-template <typename T>
-constexpr bool canMemcmpArrayOfHelper() {
-  return isSameType<T, const bool>()
-      || isSameType<T, const char>()
-      || isSameType<T, const signed char>()   || isSameType<T, const unsigned char>()
-      || isSameType<T, const short int>()     || isSameType<T, const unsigned short int>()
-      || isSameType<T, const int>()           || isSameType<T, const unsigned int>()
-      || isSameType<T, const long int>()      || isSameType<T, const unsigned long int>()
-      || isSameType<T, const long long int>() || isSameType<T, const unsigned long long int>();
-}
-
-template <typename T>
-constexpr bool canMemcmpArrayOf() {
-  return canMemcmpArrayOfHelper<const T>();
-}
-#endif
-}  // namespace _ (private)
 
 template <typename T>
 class Badge {
@@ -1812,7 +1779,7 @@ public:
 
   inline bool operator==(const ArrayPtr& other) const {
     if (size_ != other.size_) return false;
-    if (_::canMemcmpArrayOf<T>()) {
+    if (isIntegral<RemoveConst<T>>()) {
       return std::memcmp(ptr, other.ptr, size_ * sizeof(T)) == 0;
     }
     for (size_t i = 0; i < size_; i++) {
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -99,6 +99,7 @@ KJ_BEGIN_HEADER
 #endif
 
 #include <stddef.h>
+#include <cstring>
 #include <initializer_list>
 
 #if __linux__ && __cplusplus > 201200L
@@ -608,6 +609,52 @@ constexpr bool canMemcpy() {
 #define KJ_ASSERT_CAN_MEMCPY(T) \
   static_assert(kj::canMemcpy<T>(), "this code expects this type to be memcpy()-able");
 #endif
+
+namespace _ {  // private
+#if defined(__clang__) && defined(__is_identifier)
+#define KJ_HAS_BUILTIN_TYPE_TRAIT(x) (!__is_identifier(x))
+#elif defined(__has_builtin)
+#define KJ_HAS_BUILTIN_TYPE_TRAIT(x) __has_builtin(x)
+#else
+#define KJ_HAS_BUILTIN_TYPE_TRAIT(x) 0
+#endif
+
+#if KJ_HAS_BUILTIN_TYPE_TRAIT(__is_integral)
+template <typename T>
+constexpr bool canMemcmpArrayOf() {
+  return __is_integral(T);
+}
+#else
+template <typename T, typename U>
+constexpr bool isSameType() {
+#if KJ_HAS_BUILTIN_TYPE_TRAIT(__is_same)
+  return __is_same(T, U);
+#elif KJ_HAS_BUILTIN_TYPE_TRAIT(__is_same_as) || __GNUC__
+  return __is_same_as(T, U);
+#else
+  return false;
+#endif
+}
+
+#undef KJ_HAS_BUILTIN_TYPE_TRAIT  // Don't leak this to user code.
+
+template <typename T>
+constexpr bool canMemcmpArrayOfHelper() {
+  return isSameType<T, const bool>()
+      || isSameType<T, const char>()
+      || isSameType<T, const signed char>()   || isSameType<T, const unsigned char>()
+      || isSameType<T, const short int>()     || isSameType<T, const unsigned short int>()
+      || isSameType<T, const int>()           || isSameType<T, const unsigned int>()
+      || isSameType<T, const long int>()      || isSameType<T, const unsigned long int>()
+      || isSameType<T, const long long int>() || isSameType<T, const unsigned long long int>();
+}
+
+template <typename T>
+constexpr bool canMemcmpArrayOf() {
+  return canMemcmpArrayOfHelper<const T>();
+}
+#endif
+}  // namespace _ (private)
 
 template <typename T>
 class Badge {
@@ -1765,6 +1812,9 @@ public:
 
   inline bool operator==(const ArrayPtr& other) const {
     if (size_ != other.size_) return false;
+    if (_::canMemcmpArrayOf<T>()) {
+      return std::memcmp(ptr, other.ptr, size_ * sizeof(T)) == 0;
+    }
     for (size_t i = 0; i < size_; i++) {
       if (ptr[i] != other[i]) return false;
     }
--- a/c++/src/kj/common-test.c++
+++ b/c++/src/kj/common-test.c++
@@ -811,6 +811,10 @@ KJ_TEST("ArrayPtr operator ==") {
              ArrayPtr<const char* const>({"foo", "baz"})));
   KJ_EXPECT((ArrayPtr<const StringPtr>({"foo", "bar"}) !=
              ArrayPtr<const char* const>({"foo"})));
+
+  // operator== should not use memcmp for double elements.
+  double d[1] = { nan() };
+  KJ_EXPECT(ArrayPtr<double>(d, 1) != ArrayPtr<double>(d, 1));
 }
 
 KJ_TEST("kj::range()") {
--- a/c++/src/kj/units.h
+++ b/c++/src/kj/units.h
@@ -74,19 +74,6 @@ class Bounded;
 template <uint value>
 class BoundedConst;
 
-template <typename T> constexpr bool isIntegral() { return false; }
-template <> constexpr bool isIntegral<char>() { return true; }
-template <> constexpr bool isIntegral<signed char>() { return true; }
-template <> constexpr bool isIntegral<short>() { return true; }
-template <> constexpr bool isIntegral<int>() { return true; }
-template <> constexpr bool isIntegral<long>() { return true; }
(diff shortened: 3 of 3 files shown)


kentonv:
I wonder if this is actually faster. I would guess that any reasonably-smart compiler should be able to optimize the simple loop. In fact, it might be able to optimize better than memcmp() since memcmp() has to be designed for bytes whereas the loop can compare a whole integer at a time.

The one thing that might actually make the loop faster is if we could hint to the compiler that it's safe to preload later array elements before it has checked equality of previous elements -- i.e. this doesn't risk a segfault, which otherwise the compiler has no way to know, and I don't think memcmp() actually gives it this information.

tniessen:
> I would guess that any reasonably-smart compiler should be able to optimize the simple loop.

I am using [this Compiler Explorer source](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGEgJykrgAyeAyYAHI%2BAEaYxBIAzKQADqgKhE4MHt6%2BASlpGQKh4VEssfFcSXaYDplCBEzEBNk%2BflyB1bUC9Y0ExZExcYm2DU0tue0jvf2l5YkAlLaoXsTI7BzmCWHI3lgA1CYJbgoE6FhUAHQIh9gmGgCCm9u7mAdHJ8RhwFc3d48PBEwLGSBkBbzcBAAnslGKxXgAVX4PHZMBQKPb3YjEJiQ5QEYgHADsVgeyS80VoeGQID%2Bez2YUp4T2aAYJ0wqmSBMx2Nx%2BIg8xAe2SfIYXlotGFxEWe3SAC9MAB9CAaeZEqyEgAitPpDEZrxZbI5XKxOLxxAgWB2UJhEFF4sl8wFQpFYol%2BOlcsVytVJmJvq1DzpDLC%2BoEhs5GJNvPN8IAVM6yDK8PKFQQk/KnZKIA7SOmvZ6fX7Ndrg0yDYCjZGeWaIHG9rFgGFc3XXPNtXS6YKsw2wh7k17XHsALT1zCNhiF9UB%2B507mm/GHNyIhK3MwANj2qBh2KIxEOGotNVo1swttdDsn7YTbw1eztbr3CRJM47noVN72GkOz47e2ImAIFYGD2WMCAQPAFG/bV/T%2BEtdRDZkwwrCM3zTT1%2BUQ1k019Cw/wAoC8wVb8iWnIN4LLJC9kRddN23JhdxMABWKxGIPVCdSwVRVXLNUr3/QDiGAyUmIsMJOKYrUn2g4tAx1PUqPMDctzieiSBEiSIHYsT2UvWS6X4gjhOY7TVAkqDZJgh44Pk6JUE8WiVIYhJJIPHi52jRTNzAuJuMonCrzwKg9k0/t3zAMB9y8hA4guN9VQMwS9ioMQlHMx4zC2Khzj2bAInuCxgmwBUAFlsGKtxiuUAKgogBUFQg%2BrBDHbFaFrR1eL0jsEuAtgWGQYFs3xXNUG84gLklXM3xAvNUCoNqfWcyKvykzrLPShJXECq9%2BAJEKUzTPAP2W3DDsXQjiLwaxrF0l9f0C4KjNEiS9nCyKRuivdjIk%2BL8MS5LaFSlbbpIvjfuA/EvEwNK6TWv5OTwAA3ej2G1OthKBul2LfNKYKBv4wjTfhUAgdyzUXeFsCEeEFXhABNZRbhXPYmFzUmFyOCmqZp%2BnGewesbv0sHmZvSLohxmT7g4RZaE4RjeD8DgtFIVBODcK7LBlZZVleTYeFIAhNClxYAGsQASBILnNq3rZttd9E4SR5cN5XOF4BQQA0fXDcWOBYCQNBgToOJyEoAPkiD%2BIUSMKQNE9mhjzid2IGiZ3ojCRpIU4PWA7YQQAHldUzxXeCwFhDGAcRi9IfB/1qBHMHdqv2RqLxASz3gCcwGWq8paJsWISEPCwZ38TwFh28WKgDGABQADU8EwAB3POYQVvX%2BEEEQxHYKQZEERQVHUKvdC4fRy5QdWbF793IEWLdHDDTghzz5KTmHDVOepumGf3ZAEEad%2BuV8qFRKmVCqyg9huy7jUB%2BfgICuHGG0MwQQGDoBmIMCodtUjpFgYgiQyDsGFAYOgsoQwuB206LAnoYxPCtHwbYaBXQGDUL6CGWYZCKGjGaLQiYyCTjTDYRgiQa5FgKC1msPQ%2BJMDrB4NLWWTsq4qw4KoAAHGuIca5JDMgMEYPYUgLgaAMcFNWlhrC5lwIQEgBwMqnz2B4QO9ACS63mLwA2xdHSkFNubS2NtfHmztt3R2pAFZKyUW7D2Xt3GkF9ogFAqAHHBwoBAMOEcUA6OADHOOdBATECTinKuadmAD3bqQHOjACAF1oEXJWpdy6VxqXgWujh66NyVs3ZArcZEdyat3JWvd%2B6DwwOsJWo9x7cClnwaec8F7L1XiUjewhRDiF3gsg%2Bahna6GQeki%2BpjLD6DwNEG%2B/JlbCkyI3Z%2Br80xDg/pTL%2BPNf7/2IFA%2BwsCXCoLwVwZBIRBGkPiGYQk%2BQcGZA%2BQQgosCSFzH%2BQwl5dQuEguhTA2FAiShCKhfwmhOQkFTCaBCoY/zRHiJ3vrf8XS5EcDlsE52SjVHqM0do8uejJAGKMRAExVg9k5XwLuaxCRbH2PDo4nlZgXGRK0B4rxFs/F%2BPthwIJITeBhNsBEtxYqyVmAUaE12oqjakHrrkzIIBJBAA%3D) as a test.
- Neither clang nor GCC seem to optimize the loop much, and produce code that compares each element with no vectorization as far as I can tell.
- With `memcmp`, gcc's output uses `memcmp` and clang's output uses `bcmp@PLT`.

> it might be able to optimize better than memcmp() since memcmp() has to be designed for bytes whereas the loop can compare a whole integer at a time.

That might be. It'd surprise me if libc's `memcmp()` didn't have at least the same optimizations when the pointer is aligned though, probably even vectorization etc.

kentonv:
Wow, weird, I wonder why the compiler doesn't just optimize the loop to memcmp() itself (or bcmp(), as it were), if it's so much faster.

That said your benchmark tests very very large arrays. I would guess the more typical case for operator== would be comparing arrays of size 10 or so. Performance profile might be entirely different there.

But I won't actually make you test that, unless you want to. I'm convinced enough that this is probably beneficial.

c++/src/kj/common.h
```diff
@@ -609,6 +610,52 @@ constexpr bool canMemcpy() {
...
+#endif
+
+#if KJ_HAS_BUILTIN_TYPE_TRAIT(__is_integral)
+template <typename T>
+constexpr bool canMemcmpArrayOf() {
+  return __is_integral(T);
+}
+#else
+template <typename T, typename U>
+constexpr bool isSameType() {
```
kentonv, line 629: There's already a `kj::isSameType()` defined earlier in this file. :)

c++/src/kj/common.h
```diff
@@ -609,6 +610,52 @@ constexpr bool canMemcpy() {
...
+#elif defined(__has_builtin)
+#define KJ_HAS_BUILTIN_TYPE_TRAIT(x) __has_builtin(x)
+#else
+#define KJ_HAS_BUILTIN_TYPE_TRAIT(x) 0
+#endif
+
+#if KJ_HAS_BUILTIN_TYPE_TRAIT(__is_integral)
+template <typename T>
+constexpr bool canMemcmpArrayOf() {
+  return __is_integral(T);
```
kentonv, line 625: IMO it's not really worth doing all the checks to use __is_integral() if we still have to have the fallback for other compilers... I'd just use the fallback on all compilers. The benefit of __is_integral() would be if we could use it everywhere and avoid the added code complexity.

c++/src/kj/common.h
```diff
@@ -609,6 +610,52 @@ constexpr bool canMemcpy() {
...
+      || isSameType<T, const signed char>()   || isSameType<T, const unsigned char>()
+      || isSameType<T, const short int>()     || isSameType<T, const unsigned short int>()
+      || isSameType<T, const int>()           || isSameType<T, const unsigned int>()
+      || isSameType<T, const long int>()      || isSameType<T, const unsigned long int>()
+      || isSameType<T, const long long int>() || isSameType<T, const unsigned long long int>();
+}
+
+template <typename T>
+constexpr bool canMemcmpArrayOf() {
+  return canMemcmpArrayOfHelper<const T>();
```
kentonv, line 654: FWIW we have a style pattern of using a `_` suffix to name an internal helper for a template. So I'd rename `canMemcmpArrayOfHelper` to `canMemcmpArray_`.

Although I'm not sure a helper function is needed, you could also do:

```
template <typename T>
constexpr bool canMemcmpArrayOfHelper() {
  typedef const T U;
  return isSameType<U, const bool>()
      || isSameType<U, const char>()
      ...
```

Or alternatively:

```
template <typename T>
constexpr bool canMemcmpArrayOfHelper() {
  typedef RemoveConst<T> U;
  return isSameType<U, bool>()
      || isSameType<U, char>()
      ...
```

kentonv: Oh... there's also an `isIntegral` template defined in `c++/src/kj/units.h`, maybe we should just move it here?

tniessen:
> That said your benchmark tests very very large arrays. I would guess the more typical case for operator== would be comparing arrays of size 10 or so. Performance profile might be entirely different there.
> 
> But I won't actually make you test that, unless you want to. I'm convinced enough that this is probably beneficial.

@kentonv This is more difficult to test because the number of CPU cycles spent on the actual comparison is very small in each iteration, but let's give it a try because I am curious.

(Still using `clang version 14.0.1` on `x86_64-pc-linux-gnu`.)

---

<details>
<summary><code>benchmark.c++</code> (added <code>testEq16</code> and <code>testEq10Char</code>)</summary>

```cpp
#include <cstdio>
#include <kj/array.h>
#include <kj/debug.h>
#include <kj/time.h>

template <typename T>
int testEq64M(const kj::MonotonicClock& clock, kj::TimePoint& start, kj::TimePoint& end) {
  kj::Array<T> a = kj::heapArray<T>(64 * 1024 * 1024 / sizeof(T));
  kj::Array<T> b = kj::heapArray<T>(64 * 1024 * 1024 / sizeof(T));

  start = clock.now();
  int ret = !(a == b);
  end = clock.now();

  return ret;
}

template <typename T>
int testEq16(const kj::MonotonicClock& clock, kj::TimePoint& start, kj::TimePoint& end) {
  constexpr size_t n = 64 * 1024 * 1024 / 16;
  kj::Array<kj::Array<T>> arrays = kj::heapArray<kj::Array<T>>(n);
  for (size_t i = 0; i < arrays.size(); i++) {
    arrays[i] = kj::heapArray<T>(16 / sizeof(T));
  }

  int ret = 0;
  start = clock.now();
  for (size_t i = 0; i < n; i++) {
    ret += arrays[0] == arrays[i];
  }
  end = clock.now();

  return ret != n;
}

int testEq10Char(const kj::MonotonicClock& clock, kj::TimePoint& start, kj::TimePoint& end) {
  constexpr size_t n = 64 * 1024 * 1024 / 10;
  kj::Array<kj::Array<char>> arrays = kj::heapArray<kj::Array<char>>(n);
  for (size_t i = 0; i < arrays.size(); i++) {
    arrays[i] = kj::heapArray<char>(10);
  }

  int ret = 0;
  start = clock.now();
  for (size_t i = 0; i < n; i++) {
    ret += arrays[0] == arrays[i];
  }
  end = clock.now();

  return ret != n;
}

int main(int argc, char** argv) {
  KJ_REQUIRE(argc == 2);
  const kj::MonotonicClock& clock = kj::systemPreciseMonotonicClock();
  kj::StringPtr testCase(argv[1]);

  kj::TimePoint start = clock.now(), end = clock.now();
  int ret;

  if (testCase == "64M-eq-char"_kj) {
    ret = testEq64M<char>(clock, start, end);
  } else if (testCase == "64M-eq-uint16"_kj) {
    ret = testEq64M<uint16_t>(clock, start, end);
  } else if (testCase == "64M-eq-uint32"_kj) {
    ret = testEq64M<uint32_t>(clock, start, end);
  } else if (testCase == "64M-eq-uint64"_kj) {
    ret = testEq64M<uint64_t>(clock, start, end);
  } else if (testCase == "16-eq-char"_kj) {
    ret = testEq16<char>(clock, start, end);
  } else if (testCase == "16-eq-uint16"_kj) {
    ret = testEq16<uint16_t>(clock, start, end);
  } else if (testCase == "16-eq-uint32"_kj) {
    ret = testEq16<uint32_t>(clock, start, end);
  } else if (testCase == "16-eq-uint64"_kj) {
    ret = testEq16<uint64_t>(clock, start, end);
  } else if (testCase == "10-eq-char"_kj) {
    ret = testEq10Char(clock, start, end);
  } else {
    KJ_UNREACHABLE;
  }

  long duration = (end - start) / kj::NANOSECONDS;
  printf("%ld", duration);
  return ret;
}
```
</details>

<details>
<summary><code>benchmark.sh</code> (added new <code>modes</code>)</summary>

```bash
#!/bin/bash
set -euo pipefail

modes="64M-eq-char 64M-eq-uint16 64M-eq-uint32 64M-eq-uint64 16-eq-char 16-eq-uint16 16-eq-uint32 16-eq-uint64 10-eq-char"

for ref in "$@"; do
  (
    cd capnproto
    git checkout "$ref"
    cd c++
    autoreconf -i
    ./configure
    make -j12 check
  )

  $CXX -Ofast -Icapnproto/c++/src -o "benchmark-$ref" benchmark.c++ capnproto/c++/.libs/libkj.a
done

echo -n "i"
for mode in $modes; do
  for ref in "$@"; do
    echo -n ",$mode.$ref"
  done
done
echo
for i in $(seq 1 1000); do
  echo -n "$i"
  for mode in $modes; do
    for ref in "$@"; do
      echo -n ","
      "./benchmark-$ref" $mode
    done
  done
  echo
done
```
</details>

Invoked as:

```bash
export CXX=clang++-14
export CC=clang-14
./benchmark.sh master optimized-arrayptr-operatoreq
```

Raw output: [`data-v2.csv` (GitHub Gist)](https://gist.github.com/tniessen/76ca1bfa3a911c367dae22524d8da432)

---

Analysis:

```R
data <- read.csv('data-v2.csv')

library(ggplot2)

reshape.mode <- function(value, type, memcmp) {
  i <- data$i
  value <- value
  reshaped <- data.frame(i, value, type, memcmp)
}
```

## 64 MiB

```R
reshaped64M <- rbind(
  reshape.mode(data$X64M.eq.char.master, 'char', FALSE),
  reshape.mode(data$X64M.eq.char.optimized.arrayptr.operatoreq, 'char', TRUE),
  reshape.mode(data$X64M.eq.uint16.master, 'uint16', FALSE),
  reshape.mode(data$X64M.eq.uint16.optimized.arrayptr.operatoreq, 'uint16', TRUE),
  reshape.mode(data$X64M.eq.uint32.master, 'uint32', FALSE),
  reshape.mode(data$X64M.eq.uint32.optimized.arrayptr.operatoreq, 'uint32', TRUE),
  reshape.mode(data$X64M.eq.uint64.master, 'uint64', FALSE),
  reshape.mode(data$X64M.eq.uint64.optimized.arrayptr.operatoreq, 'uint64', TRUE)
)

ggplot(reshaped64M, aes(value, type, fill=memcmp)) +
  geom_boxplot() +
  expand_limits(x = 0) +
  ggtitle('64 MiB')
```

<p align="center">

</p>

## 16 bytes

```R
reshaped16 <- rbind(
  reshape.mode(data$X16.eq.char.master, 'char', FALSE),
  reshape.mode(data$X16.eq.char.optimized.arrayptr.operatoreq, 'char', TRUE),
  reshape.mode(data$X16.eq.uint16.master, 'uint16', FALSE),
  reshape.mode(data$X16.eq.uint16.optimized.arrayptr.operatoreq, 'uint16', TRUE),
  reshape.mode(data$X16.eq.uint32.master, 'uint32', FALSE),
  reshape.mode(data$X16.eq.uint32.optimized.arrayptr.operatoreq, 'uint32', TRUE),
  reshape.mode(data$X16.eq.uint64.master, 'uint64', FALSE),
  reshape.mode(data$X16.eq.uint64.optimized.arrayptr.operatoreq, 'uint64', TRUE)
)

ggplot(reshaped16, aes(value, type, fill=memcmp)) +
  geom_boxplot() +
  expand_limits(x = 0) +
  ggtitle('16 bytes')
```

<p align="center">

</p>

## 10 chars

```R
reshaped10 <- rbind(
  reshape.mode(data$X10.eq.char.master, 'char', FALSE),
  reshape.mode(data$X10.eq.char.optimized.arrayptr.operatoreq, 'char', TRUE)
)

ggplot(reshaped10, aes(value, type, fill=memcmp)) +
  geom_boxplot() +
  expand_limits(x = 0) +
  ggtitle('10 chars')
```

<p align="center">

</p>

---

Unsurprisingly, the loop beats `memcmp` for very small `uint64_t` arrays, but that's probably a very rare use case.

`memcmp` appears to be faster even for strings of lengths 10 and 16. (But take these micro-benchmarks with a grain of salt.)

c++/src/kj/common.h
tniessen: I've moved `isIntegral` to `common.h` but it does not support `bool` or `const` types. I don't know if I can change that without affecting other parts of KJ that rely on `isIntegral`, so I'll ignore `bool` and added a workaround for `const` to `operator==`.

kentonv (approved):

PR #1471, opened 2022-05-27
Rename AnyPointer -> Capability

JonasVautherin:
I ran through this TODO that it seems can now be fulfilled :smile:


JonasVautherin:
Note that I'm not entirely sure this doesn't break anything. I ran the tests and the calculator example, and everything seems fine, but let me know if I should try something else.

I was expecting that somewhere, I could remove a cast from `AnyPointer` to `Capability`, but I did not find that in the codebase (doesn't mean there is none).

kentonv:
Yeah, joins actually aren't even implemented so this isn't used anywhere yet. But thanks for addressing the TODO, it makes the proposed protocol easier to read at least.

PR #1473, opened 2022-05-31
Support OneOf for up to 50 variants

tniessen:


kentonv:
"50 should be enough for anyone."

PR #1474, opened 2022-06-01
Introduce a debugging aid to detect when KJ I/O objects are held in the wrong places.

kentonv:
KJ I/O objects cannot cross threads. This implies that objects which can cross threads must never own KJ I/O objects, execpt perhaps through specific mechanisms designed to allow it. This commit introduces some utilities that can be used to detect when KJ I/O objects are held in the wrong places.

### DIFF
--- a/c++/src/kj/async.h
+++ b/c++/src/kj/async.h
@@ -64,6 +64,54 @@ using PromiseForResult = _::ReducePromises<_::ReturnType<Func, T>>;
 // Func itself returns a promise, the promises are joined, so you never get Promise<Promise<T>>.
 
 // =======================================================================================
+
+class AsyncObject {
+  // You may optionally inherit privately from this to indicate that the type is a KJ async object,
+  // meaning it deals with KJ async I/O making it tied to a specific thread and event loop. This
+  // enables some additional debug checks, but does not otherwise have any effect on behavior as
+  // long as there are no bugs.
+  //
+  // (We prefer inheritance rather than composition here because inheriting an empty type adds zero
+  // size to the derived class.)
+
+public:
+  ~AsyncObject() noexcept;
+};
+
+class DisallowAsyncDestructorsScope {
+  // Create this type on the stack in order to specify that during its scope, no KJ async objects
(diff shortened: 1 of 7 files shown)


c++/src/kj/async.h
```diff
@@ -63,6 +63,49 @@ using PromiseForResult = _::ReducePromises<_::ReturnType<Func, T>>;
...
+
+  friend class AsyncObject;
+};
+
+class AllowAsyncDestructorsScope {
+  // Negates the effect of DisallowAsyncDestructorsScope.
+
+public:
+  AllowAsyncDestructorsScope();
+  ~AllowAsyncDestructorsScope();
```
harrishancock, line 103: Nit: I'd suggest `KJ_DISALLOW_COPY` on all of these new types. Granted, it's hard to imagine a circumstance where someone would accidentally move one of these...

c++/src/kj/async.h
```diff
@@ -63,6 +63,49 @@ using PromiseForResult = _::ReducePromises<_::ReturnType<Func, T>>;
 // T.  If T is void, then the promise is for the result of calling Func with no arguments.  If
 // Func itself returns a promise, the promises are joined, so you never get Promise<Promise<T>>.
 
+// =======================================================================================
+
+class AsyncObject {
+  // You may optionally inherit privately from this to indicate that the type is a KJ async object,
```
harrishancock, line 69: Any reason for inheritance versus composition? I.e.,
```cpp
class IoThingy {
  kj::AsyncObject asyncDestructorGuard;
};
```
I like composition for this use case because, at least to me, it reinforces the optional nature of the debugging aid.

harrishancock (approved):

c++/src/kj/async.h
kentonv: Yes: Inheritance of an empty class adds zero bytes to the derived class's size. Composition adds at least one byte and up to 8 depending on padding.

c++/src/kj/async.h
kentonv: Good idea, done.

PR #1475, opened 2022-06-06
Support `permessage-deflate` WebSocket Extension

MellowYarker:
## WIP
**Edit**
The first 3 commits were essentially:
1. build the extension header for the request (drop everything but `permessage-deflate` offers)
2. parse the offers as the server + build a response
3. parse the servers response as the client

I wasn't sure how I wanted to move the code since a lot of it could be shared, so rather than doing fixups I made a 4th commit for refactoring everything. This was... not a great idea... since now commits 1-3 are pretty outdated.

I'm considering: 
- leaving it as is (first 3 commits are helpful indications of where each step of negotiation takes place), or
- closing the PR and opening a new one with a single squashed commit. 

@kentonv do you have a preference either way?

---

Since the first chunk of this feature involves parsing the `Sec-WebSocket-Extensions` header, I figured it might be better to get eyes on it before moving past extension negotiation. Once we're happy with how negotiation is handled, I'll move on to compression.

As of now, we can:
- Drop all `Sec-WebSocket-Extensions` aside from `permessage-deflate` before sending the request to the server.
- Parse `permessage-deflate` offers as the server, populate a struct `CompressionParameters`, then build a response with agreed parameters.
- Parse `permessage-deflate` in a response (as a client), comparing the clients offer with what the server agreed to support, then populate `CompressionParameters`.
- Handle invalid offers/agreements, i.e. decline offers as the server/fail connections as the client.

Would be nice to add some tests for the string parsing code, though we indirectly test this by testing `Sec-WebSocket-Extensions` negotiation.

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -5350,6 +5350,128 @@ private:
     }
   }
 
+  struct CompressionParameters {
+    // These are the parameters for `Sec-WebSocket-Extensions` permessage-deflate extension.
+    // Since we cannot distinguish the client/server in `upgradeToWebSocket`, we use the prefixes
+    // `inbound` and `outbound` instead.
+    bool outbound_no_context_takeover;
+    bool inbound_no_context_takeover;
+    kj::Maybe<size_t> outbound_max_window_bits;
+    kj::Maybe<size_t> inbound_max_window_bits;
+  };
+
+  kj::Maybe<CompressionParameters> validateRequestCompressionParameters(
+      kj::ArrayPtr<kj::String> parameters) {
+    if (parameters.size() > 5) {
+      // `permessage-deflate` has 4 optional parameters. If we've seen all 4 parameters, then any
+      // additional parameters are either repeats, or invalid, so there's no need to process them.
+      // `parameters` starts with `permessage-deflate`, so we should have 5 elements at most.
+      return nullptr;
+    }
+
+    kj::Maybe<CompressionParameters> compressionParameters = CompressionParameters {
+        .outbound_no_context_takeover = false,
+        .inbound_no_context_takeover = false,
+        .outbound_max_window_bits = nullptr,
+        .inbound_max_window_bits = nullptr };
+    auto& config = KJ_REQUIRE_NONNULL(compressionParameters);
+
+    for (const auto &i : kj::indices(parameters)) {
+      if (i == 0) {
+        // Skip first element, which is `permessage-deflate`.
+        continue;
+      }
+      if (parameters[i].startsWith("server_max_window_bits=")) {
+        // TODO (now): RFC seems to imply that if the param has a value, it will not have
+        // whitespace between the `=`. Also, for permessage-deflate specifically, we should
+        // not expect any double quotations, since the values for x_max_window_bits is a digit.
+        if (config.outbound_max_window_bits != nullptr) {
+          // We've seen this parameter already, decline.
+          compressionParameters = nullptr;
+          break;
+        }
+        auto windowSize = kj::str(
+            parameters[i].slice("server_max_window_bits="_kj.size(), parameters[i].size()));
+
+        size_t window;
+        kj::Maybe<kj::Exception> maybeException = kj::runCatchingExceptions([&]() {
+          window = windowSize.parseAs<size_t>();
+        });
+        KJ_IF_MAYBE(e, maybeException) {
+          // Received invalid ABNF, expected 1*DIGIT -- decline.
+          compressionParameters = nullptr;
+          break;
+        }
+        config.outbound_max_window_bits = window;
+      } else if (parameters[i].startsWith("client_max_window_bits")) {
+        if (config.inbound_max_window_bits != nullptr) {
+          // We've seen this parameter already, decline.
+          compressionParameters = nullptr;
+          break;
+        }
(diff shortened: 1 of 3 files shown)


c++/src/kj/compat/http.c++
```diff
@@ -1045,6 +1045,65 @@ kj::String HttpHeaders::toString() const {
...
+  advancedBy = input.findFirst(c);
+  KJ_IF_MAYBE(index, advancedBy) {
+    // We found a delimiter, move to 1 past it, then skip over the whitespace.
+    auto startOfNext = const_cast<char*>((input.begin() + *index + 1));
+    result = skipSpace(startOfNext);
+  }
+  return result;
+}
+
+kj::Vector<kj::String> splitPartsWithCondition(const kj::StringPtr input, const char& delim,
```
MellowYarker, line 1066: To give some context for why I've included the ability to skip if a precondition isn't met: when we're parsing the headers as a client (prior to sending the request to the server), we only want to retain `permessage-deflate` offers, so we might as well do this in 1 pass.

MellowYarker:
Fixup + force push was to silence unused variable warnings that caused some checks to fail.

c++/src/kj/compat/http.c++
MellowYarker: Additionally, `kj/compat/url.c++` has a few `split` variants, but extensions have the abnf 
```
Sec-WebSocket-Extensions = extension-list
         extension-list = 1#extension
         extension = extension-token *( ";" extension-param )
```
so this is a valid `extension-list`: `,,,          ,permessage-deflate;client_max_window_bits`. Since we can have 1+ commas after an extension, as well as optional linear whitespace, it seemed easier to have a function that just gives us what we care about so we can focus on processing that.

c++/src/kj/compat/http.c++
```diff
@@ -3490,7 +3549,260 @@ WebSocketPipe newWebSocketPipe() {
...
+        WebSocketCompressionUtil::CompressionParameters {};
+    auto& config = KJ_REQUIRE_NONNULL(compressionParameters);
+
+    for (const auto &i : kj::indices(parameters)) {
+      if (i == 0) {
+        // Skip first element, which is `permessage-deflate`.
+        continue;
+      }
+      if (parameters[i].startsWith("server_max_window_bits=")) {
+        // TODO (now): RFC seems to imply that if the param has a value, it will not have
```
MellowYarker, line 3594: Not completely certain about this bit. RFC-6455 [section 9.1](https://datatracker.ietf.org/doc/html/rfc6455#section-9.1) says extension parameters have the following abnf:
```
extension-param = token [ "=" (token | quoted-string) ]
```
So it's a bit unclear to me (for now) whether `token *LWS "=" *LWS (token | quoted-string)` is valid. Ex. can we have `client_max_window_bits       =   10`? It isn't necessarily a problem since if we fail to determine the value, we'll default to the max window as the server, and `min(offer, max_window)` as the client.

MellowYarker: Fixup https://github.com/capnproto/capnproto/pull/1475/commits/7da637bd2b26d3205108b11851cf3532b5425bc5

c++/src/kj/compat/http.c++
```diff
@@ -3490,7 +3549,287 @@ WebSocketPipe newWebSocketPipe() {
 // =======================================================================================
 
 namespace {
+  class WebSocketCompressionUtil {
```
MellowYarker, line 3552: I'm considering moving/splitting this code in the following way:
1. Move `CompressionParameters` struct out of `WebSocketCompressionUtil` into `http.h` as a standalone struct. This would mean it could be added as a parameter to `newWebSocket()` in `http.h`.
2. Moving all the static methods out of the class but keeping them in an anonymous namespace. I don't really see a reason to have a class anymore. Previously, I was trying to limit access to the methods to `HttpServer::Connection` and `HttpClientImpl`, but it feels like keeping this within file scope is good enough.

MellowYarker: Fixup https://github.com/capnproto/capnproto/pull/1475/commits/c9860e420142adb8b0cef6eda5996a47d670f8a2. I figure it'd be nice to change the function/struct names a bit to associate them to websocket stuff, assuming this is a better direction.

MellowYarker:
Renamed some functions https://github.com/capnproto/capnproto/pull/1475/commits/604b632590092e6e491986392fedaa0b534af6db, going to rebase and force push. I should have just done fixups on the first three commits rather than adding a 4th commit since I see how that complicates the review... lesson learned.

MellowYarker:
- https://github.com/capnproto/capnproto/pull/1475/commits/83645294e4041516ea2170c09ee4d1f5f08b54be found that we were setting the wrong variable to `nullptr`.
- https://github.com/capnproto/capnproto/pull/1475/commits/c6d53ec7fa213097118506f89efec8dcf200f86b combined `validateRequestCompressionParameters` and `validateResponseCompressionParameters` into one function.

MellowYarker:
- https://github.com/capnproto/capnproto/pull/1475/commits/e06bcf9c45a4259d2465a190c47f76873ce46ab1 added check for invalid range of LZ77 window bits.
- https://github.com/capnproto/capnproto/pull/1475/commits/cab196ee616953073e31a99d8bfa070441c3356d formatting fixes.

PR #1478, opened 2022-06-09
Fix ASAN problems under Clang 15.

kentonv:
It appears ASAN now by default tries to detect stack-use-after-return. This breaks our assumptions in requireOnStack() and totally breaks fibers.

For requireOnStack() we can just skip the check in this case.

For fibers, we need to implement the ASAN hints to tell it when we're switching fibers.


kentonv:
Hmm this might not be clang 15 related, but rather something else that changed in V8 10.3?

Anyway, these hints have been supported for a long time but suddenly they are required.

PR #1480, opened 2022-06-14
Support `permessage-deflate` WebSocket Extension

MellowYarker:
Closed the old PR because I messed up the git history and didn't want to make review unnecessarily challenging.

This is going to be fairly dense, so I think it's better to break it into parts.

### Phase 1 - Negotiating the Extension
This is all that's been done so far. Commits 1-3 represent a stage in extension negotiation, that is:
1. build the extension header for the request
2. parse the offers as the server + build a response
3. parse the servers response as the client

I added a test in commit 4, but since we're doing a lot of string parsing I think more tests are needed.

As of now, we:
- Drop all `Sec-WebSocket-Extensions` aside from `permessage-deflate` before sending the request to the server. The client retains information from its highest preference offer (ex. # bits to use for `client_max_window_bits`) so it can compare with what the server agrees to support.
- Parse `permessage-deflate` offers as the server, populate a struct `CompressionParameters`, then build a response with agreed parameters.
- Parse `permessage-deflate` in a response (as a client), potentially comparing the information the client retained with what the server agreed to support. The client also populates its own `CompressionParameters`.
- Handle invalid offers/agreements, i.e. decline offers as the server/fail connections as the client (could use tests here!)
- Handle whitespace between tokens correctly (implied `*LWS`) (could use tests here too, though the existing test somewhat covers)

One thing on my mind is I'm not sure what the best way to organize the code is. I've put new functions in an anonymous namespace so that they can't be accessed outside the file (we only really need these for this file anyways) since it looks like that's done elsewhere. AFAIK, this means we can only test the string parsing indirectly by passing `Sec-WebSocket-Extensions` headers in tests that create websocket client/servers.
### Phase 3 - Using Compression Parameters

### DIFF
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -1753,6 +1753,15 @@ public:
     return other.size_ <= size_ && slice(0, other.size_) == other;
   }
 
+  inline Maybe<size_t> findFirst(const T& match) const {
+    for (size_t i = 0; i < size_; i++) {
+      if (ptr[i] == match) {
+        return i;
+      }
+    }
+    return nullptr;
+  }
+
   inline ArrayPtr<PropagateConst<T, byte>> asBytes() const {
     // Reinterpret the array as a byte array. This is explicitly legal under C++ aliasing
     // rules.
@@ -1798,6 +1807,46 @@ private:
   T* ptr;
   size_t size_;
 };
+
+template <>
+inline Maybe<size_t> ArrayPtr<const char>::findFirst(const char& c) const {
+  const char* pos = reinterpret_cast<const char*>(memchr(ptr, c, size_));
+  if (pos == nullptr) {
+    return nullptr;
+  } else {
+    return pos - ptr;
+  }
+}
+
+template <>
+inline Maybe<size_t> ArrayPtr<char>::findFirst(const char& c) const {
+  char* pos = reinterpret_cast<char*>(memchr(ptr, c, size_));
+  if (pos == nullptr) {
+    return nullptr;
+  } else {
+    return pos - ptr;
+  }
+}
+
+template <>
+inline Maybe<size_t> ArrayPtr<const byte>::findFirst(const byte& c) const {
+  const byte* pos = reinterpret_cast<const byte*>(memchr(ptr, c, size_));
+  if (pos == nullptr) {
+    return nullptr;
+  } else {
+    return pos - ptr;
+  }
+}
+
+template <>
+inline Maybe<size_t> ArrayPtr<byte>::findFirst(const byte& c) const {
+  byte* pos = reinterpret_cast<byte*>(memchr(ptr, c, size_));
+  if (pos == nullptr) {
+    return nullptr;
+  } else {
+    return pos - ptr;
+  }
+}
 
 template <typename T>
 inline constexpr ArrayPtr<T> arrayPtr(T* ptr KJ_LIFETIMEBOUND, size_t size) {
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -100,6 +100,7 @@ KJ_BEGIN_HEADER
 
 #include <stddef.h>
 #include <initializer_list>
+#include <string.h>
 
 #if __linux__ && __cplusplus > 201200L
 // Hack around stdlib bug with C++14 that exists on some Linux systems.
@@ -1747,6 +1748,9 @@ public:
   inline ArrayPtr slice(size_t start, size_t end) {
     KJ_IREQUIRE(start <= end && end <= size_, "Out-of-bounds ArrayPtr::slice().");
     return ArrayPtr(ptr + start, end - start);
+  }
+  inline bool startsWith(const ArrayPtr<const T>& other) const {
+    return other.size_ <= size_ && slice(0, other.size_) == other;
   }
 
   inline ArrayPtr<PropagateConst<T, byte>> asBytes() const {
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -3489,6 +3489,330 @@ WebSocketPipe newWebSocketPipe() {
 
 // =======================================================================================
 
+namespace _ { // private implementation details
+
+kj::ArrayPtr<const char> splitNext(kj::ArrayPtr<const char>& cursor, char delimiter) {
+  // Consumes and returns the next item in a delimited list.
+  //
+  // If a delimiter is found:
+  //  - `cursor` is updated to point to the rest of the string after the delimiter.
+  //  - The text before the delimiter is returned.
+  // If no delimiter is found:
+  //  - `cursor` is updated to an empty string.
+  //  - The text that had been in `cursor` is returned.
+  //
+  // (It's up to the caller to stop the loop once `cursor` is empty.)
+  KJ_IF_MAYBE(index, cursor.findFirst(delimiter)) {
+    auto part = cursor.slice(0, *index);
+    cursor = cursor.slice(*index + 1, cursor.size());
+    return part;
+  }
+  kj::ArrayPtr<const char> result(kj::mv(cursor));
+  cursor = nullptr;
+
+  return result;
+}
+
+void stripLeadingAndTrailingSpace(ArrayPtr<const char>& str) {
+  // Remove any leading/trailing spaces from `str`, modifying it in-place.
+  while (str.size() > 0 && (str[0] == ' ' || str[0] == '\t')) {
+    str = str.slice(1, str.size());
+  }
+  while (str.size() > 0 && (str.back() == ' ' || str.back() == '\t')) {
+    str = str.slice(0, str.size() - 1);
+  }
+}
+
+kj::Vector<kj::ArrayPtr<const char>> splitParts(kj::ArrayPtr<const char> input, char delim) {
+  // Given a string `input` and a delimiter `delim`, split the string into a vector of substrings,
+  // separated by the delimiter. Note that leading/trailing whitespace is stripped from each element.
+  kj::Vector<kj::ArrayPtr<const char>> parts;
+
+  while (input.size() != 0) {
+    auto part = splitNext(input, delim);
+    stripLeadingAndTrailingSpace(part);
+    parts.add(kj::mv(part));
+  }
+
+  return parts;
+}
+
+kj::Array<KeyMaybeVal> toKeysAndVals(const kj::ArrayPtr<kj::ArrayPtr<const char>>& params) {
+  // Given a collection of parameters (a single offer), parse the parameters into <key, MaybeValue>
+  // pairs. If the parameter contains an `=`, we set the `key` to everything before, and the `value`
+  // to everything after. Otherwise, we set the `key` to be the entire parameter.
+  // Either way, both the key and value (if it exists) are stripped of leading & trailing whitespace.
+  auto result = kj::heapArray<KeyMaybeVal>(params.size());
+  size_t count = 0;
+  for (const auto& param : params) {
+    kj::ArrayPtr<const char> key;
+    kj::Maybe<kj::ArrayPtr<const char>> value;
+
+    KJ_IF_MAYBE(index, param.findFirst('=')) {
+      // Found '=' so we have a value.
+      key = param.slice(0, *index);
+      stripLeadingAndTrailingSpace(key);
+      value = param.slice(*index + 1, param.size());
+      KJ_IF_MAYBE(v, value) {
+        stripLeadingAndTrailingSpace(*v);
+      }
+    } else {
+      key = kj::mv(param);
+    }
+
+    result[count].key = kj::mv(key);
+    result[count].val = kj::mv(value);
+    ++count;
+  }
+  return kj::mv(result);
+}
+
+struct ParamType {
+  enum { CLIENT, SERVER } side;
+  enum { NO_CONTEXT_TAKEOVER, MAX_WINDOW_BITS } property;
+};
+
+inline kj::Maybe<ParamType> parseKeyName(kj::ArrayPtr<const char>& key) {
+  // Returns a `ParamType` struct if the `key` is valid and nullptr if invalid.
+
+  if (key == "client_no_context_takeover"_kj) {
+    return ParamType { ParamType::CLIENT, ParamType::NO_CONTEXT_TAKEOVER };
+  } else if (key == "server_no_context_takeover"_kj) {
+    return ParamType { ParamType::SERVER, ParamType::NO_CONTEXT_TAKEOVER };
+  } else if (key == "client_max_window_bits"_kj) {
+    return ParamType { ParamType::CLIENT, ParamType::MAX_WINDOW_BITS };
+  } else if (key == "server_max_window_bits"_kj) {
+    return ParamType { ParamType::SERVER, ParamType::MAX_WINDOW_BITS };
+  }
+  return nullptr;
+}
+
+kj::Maybe<UnverifiedConfig> populateUnverifiedConfig(kj::Array<KeyMaybeVal>& params) {
+  // Given a collection of <key, MaybeValue> pairs, attempt to populate an `UnverifiedConfig` struct.
+  // If the struct cannot be populated, we return null.
+  //
+  // This function populates the struct with what it finds, it does not perform bounds checking or
+  // concern itself with valid `Value`s (so long as the `Value` is non-empty).
+  //
+  // The following issues would prevent a struct from being populated:
+  //  Key issues:
+  //    - `Key` is invalid (see `parseKeyName()`).
+  //    - `Key` is repeated.
+  //  Value issues:
+  //    - Got a `Value` when none was expected (only the `max_window_bits` parameters expect values).
+  //    - Got an empty `Value` (0 characters, or all whitespace characters).
+
+  if (params.size() > 4) {
+    // We expect 4 `Key`s at most, having more implies repeats/invalid keys are present.
+    return nullptr;
+  }
+
+  UnverifiedConfig config;
+
+  for (auto& param : params) {
+    KJ_IF_MAYBE(paramType, parseKeyName(param.key)) {
+      // `Key` is valid, but we still want to check for repeats.
+      const auto& side = paramType->side;
+      const auto& property = paramType->property;
+
+      if (property == ParamType::NO_CONTEXT_TAKEOVER) {
+        auto& takeOverSetting = (side == ParamType::CLIENT) ?
+            config.clientNoContextTakeover : config.serverNoContextTakeover;
+
+        if (takeOverSetting == true) {
+          // This `Key` is a repeat; invalid config.
+          return nullptr;
+        }
+
+        if (param.val != nullptr) {
+          // The `x_no_context_takeover` parameter shouldn't have a value; invalid config.
+          return nullptr;
+        }
+
+        takeOverSetting = true;
+      } else if (property == ParamType::MAX_WINDOW_BITS) {
+        auto& maxBitsSetting =
+            (side == ParamType::CLIENT) ? config.clientMaxWindowBits : config.serverMaxWindowBits;
+
+        if (maxBitsSetting != nullptr) {
+          // This `Key` is a repeat; invalid config.
+          return nullptr;
+        }
+
+        KJ_IF_MAYBE(value, param.val) {
+          if (value->size() == 0) {
+            // This is equivalent to `x_max_window_bits=`, since we got an "=" we expected a token
+            // to follow.
+            return nullptr;
+          }
+          maxBitsSetting = param.val;
+        } else {
+          // We know we got this `max_window_bits` parameter in a Request/Response, and we also know
+          // that it didn't include an "=" (otherwise the value wouldn't be null).
+          // It's important to retain the information that the parameter was received *without* a
+          // corresponding value, as this may determine whether the offer is valid or not.
+          //
+          // To retain this information, we'll set `maxBitsSetting` to be an empty ArrayPtr so this
+          // can be dealt with properly later.
+          maxBitsSetting = ArrayPtr<const char>();
+        }
+      }
+    } else {
+      // Invalid parameter.
+      return nullptr;
+    }
+  }
+  return kj::mv(config);
+}
+
+kj::Maybe<CompressionParameters> validateCompressionConfig(UnverifiedConfig&& config,
+    bool isAgreement) {
+  // Verifies that the `config` is valid depending on whether we're validating a Request (offer) or
+  // a Response (agreement). This essentially consumes the `UnverifiedConfig` and converts it into a
+  // `CompressionParameters` struct.
+  CompressionParameters result;
+
+  KJ_IF_MAYBE(serverBits, config.serverMaxWindowBits) {
+    if (serverBits->size() == 0) {
+      // This means `server_max_window_bits` was passed without a value. Since a value is required,
+      // this config is invalid.
+      return nullptr;
+    } else {
+      KJ_IF_MAYBE(bits, kj::str(*serverBits).tryParseAs<size_t>()) {
+        if (*bits < 8 || 15 < *bits) {
+          // Out of range -- invalid.
+          return nullptr;
+        }
+        if (isAgreement) {
+          result.inboundMaxWindowBits = *bits;
+        } else {
+          result.outboundMaxWindowBits = *bits;
+        }
+      } else {
+        // Invalid ABNF, expected 1*DIGIT.
+        return nullptr;
+      }
+    }
+  }
+
+  KJ_IF_MAYBE(clientBits, config.clientMaxWindowBits) {
+    if (clientBits->size() == 0) {
+      if (!isAgreement) {
+        // `client_max_window_bits` does not need to have a value in an offer, let's set it to 15
+        // to get the best level of compression.
+        result.inboundMaxWindowBits = 15;
+      } else {
+        // `client_max_window_bits` must have a value in a Response.
+        return nullptr;
+      }
+    } else {
+      KJ_IF_MAYBE(bits, kj::str(*clientBits).tryParseAs<size_t>()) {
+        if (*bits < 8 || 15 < *bits) {
+          // Out of range -- invalid.
+          return nullptr;
+        }
+        if (isAgreement) {
+          result.outboundMaxWindowBits = *bits;
+        } else {
+          result.inboundMaxWindowBits = *bits;
+        }
+      } else {
+        // Invalid ABNF, expected 1*DIGIT.
+        return nullptr;
+      }
+    }
+  }
+
+  if (isAgreement) {
+    result.outboundNoContextTakeover = config.clientNoContextTakeover;
+    result.inboundNoContextTakeover = config.serverNoContextTakeover;
+  } else {
+    result.inboundNoContextTakeover = config.clientNoContextTakeover;
+    result.outboundNoContextTakeover = config.serverNoContextTakeover;
+  }
+  return kj::mv(result);
+}
+
+inline kj::Maybe<CompressionParameters> tryExtractParameters(
+    kj::Vector<kj::ArrayPtr<const char>>& configuration,
+    bool isAgreement) {
+  // If the `configuration` is structured correctly and has no invalid parameters/values, we will
+  // return a populated `CompressionParameters` struct.
+  if (configuration.size() == 1) {
+    // Plain `permessage-deflate`.
+    return CompressionParameters{};
+  }
+  auto params = configuration.slice(1, configuration.size());
+  auto keyMaybeValuePairs = toKeysAndVals(params);
+  // Parse parameter strings into parameter[=value] pairs.
+  auto maybeUnverified = populateUnverifiedConfig(keyMaybeValuePairs);
+  KJ_IF_MAYBE(unverified, maybeUnverified) {
+    // Parsing succeeded, i.e. the parameter (`key`) names are valid and we don't have
+    // values for `x_no_context_takeover` parameters (the configuration is structured correctly).
+    // All that's left is to check the `x_max_window_bits` values (if any are present).
+    KJ_IF_MAYBE(validConfig, validateCompressionConfig(kj::mv(*unverified), isAgreement)) {
+      return kj::mv(*validConfig);
+    }
+  }
+  return nullptr;
+}
+
+kj::Vector<CompressionParameters> findValidExtensionOffers(StringPtr offers) {
+  // A function to be called by the client that wants to offer extensions through
+  // `Sec-WebSocket-Extensions`. This function takes the value of the header (a string) and
+  // populates a Vector of all the valid offers.
+  kj::Vector<CompressionParameters> result;
+
+  auto extensions = splitParts(offers, ',');
+
+  for (const auto& offer : extensions) {
+    auto splitOffer = splitParts(offer, ';');
+    if (splitOffer.front() != "permessage-deflate"_kj) {
+      continue;
+    }
+    KJ_IF_MAYBE(validated, tryExtractParameters(splitOffer, false)) {
+      // We need to swap the inbound/outbound properties since `tryExtractParameters` thinks we're
+      // parsing as the server (`isAgreement` is false).
+      auto tempCtx = validated->inboundNoContextTakeover;
+      validated->inboundNoContextTakeover = validated->outboundNoContextTakeover;
+      validated->outboundNoContextTakeover = tempCtx;
+      auto tempWindow = validated->inboundMaxWindowBits;
+      validated->inboundMaxWindowBits = validated->outboundMaxWindowBits;
+      validated->outboundMaxWindowBits = tempWindow;
+      result.add(kj::mv(*validated));
+    }
+  }
+
+  return kj::mv(result);
+}
+
+kj::String generateExtensionRequest(const ArrayPtr<CompressionParameters>& extensions) {
+  // Build the `Sec-WebSocket-Extensions` request from the validated parameters.
+  constexpr auto EXT = "permessage-deflate"_kj;
+  auto offers = kj::heapArray<String>(extensions.size());
+  size_t i = 0;
+  for (const auto& offer : extensions) {
+    offers[i] = kj::str(EXT);
+    if (offer.outboundNoContextTakeover) {
+      offers[i] = kj::str(offers[i], "; client_no_context_takeover");
+    }
+    if (offer.inboundNoContextTakeover) {
+      offers[i] = kj::str(offers[i], "; server_no_context_takeover");
+    }
+    if (offer.outboundMaxWindowBits != nullptr) {
+      auto w = KJ_ASSERT_NONNULL(offer.outboundMaxWindowBits);
+      offers[i] = kj::str(offers[i], "; client_max_window_bits=", w);
+    }
+    if (offer.inboundMaxWindowBits != nullptr) {
+      auto w = KJ_ASSERT_NONNULL(offer.inboundMaxWindowBits);
+      offers[i] = kj::str(offers[i], "; server_max_window_bits=", w);
+    }
+    ++i;
+  }
+  return kj::strArray(offers, ", ");
+}
+
+} // namespace _ (private)
 namespace {
 
 class HttpClientImpl final: public HttpClient,
@@ -3626,6 +3950,18 @@ public:
     connectionHeaders[HttpHeaders::BuiltinIndices::UPGRADE] = "websocket";
     connectionHeaders[HttpHeaders::BuiltinIndices::SEC_WEBSOCKET_VERSION] = "13";
     connectionHeaders[HttpHeaders::BuiltinIndices::SEC_WEBSOCKET_KEY] = keyBase64;
+
+    kj::Maybe<kj::String> offeredExtensions;
+    if (settings.enableWebSocketCompression) {
+      KJ_IF_MAYBE(value, headers.get(HttpHeaderId::SEC_WEBSOCKET_EXTENSIONS)) {
+        // Strip all `Sec-WebSocket-Extensions` except for `permessage-deflate`.
+        auto extensions = _::findValidExtensionOffers(*value);
+        if (extensions.size() > 0) {
+          connectionHeaders[HttpHeaders::BuiltinIndices::SEC_WEBSOCKET_EXTENSIONS] =
+              offeredExtensions.emplace(_::generateExtensionRequest(extensions.asPtr()));
+        }
+      }
+    }
 
     httpOutput.writeHeaders(headers.serializeRequest(HttpMethod::GET, url, connectionHeaders));
 
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -3812,6 +3812,45 @@ kj::String generateExtensionRequest(const ArrayPtr<CompressionParameters>& exten
   return kj::strArray(offers, ", ");
 }
 
+kj::Maybe<CompressionParameters> tryParseExtensionOffers(StringPtr offers) {
+  // Given a string of offers, accept the first valid offer by returning a `CompressionParameters`
+  // struct. If there are no valid offers, return `nullptr`.
+  auto splitOffers = splitParts(offers, ',');
+
+  for (const auto& offer : splitOffers) {
+    auto splitOffer = splitParts(offer, ';');
+
+    if (splitOffer.front() != "permessage-deflate"_kj) {
+      // Extension token was invalid.
+      continue;
+    }
+    KJ_IF_MAYBE(config, tryExtractParameters(splitOffer, false)) {
+      return kj::mv(*config);
+    }
+  }
+  return nullptr;
+}
+
+kj::String generateExtensionResponse(const CompressionParameters& parameters) {
+  // Build the `Sec-WebSocket-Extensions` response from the agreed parameters.
+  kj::String response = kj::str("permessage-deflate");
+  if (parameters.inboundNoContextTakeover) {
+    response = kj::str(response, "; client_no_context_takeover");
+  }
+  if (parameters.outboundNoContextTakeover) {
+    response = kj::str(response, "; server_no_context_takeover");
+  }
+  if (parameters.inboundMaxWindowBits != nullptr) {
+    auto w = KJ_REQUIRE_NONNULL(parameters.inboundMaxWindowBits);
+    response = kj::str(response, "; client_max_window_bits=", w);
+  }
+  if (parameters.outboundMaxWindowBits != nullptr) {
+    auto w = KJ_REQUIRE_NONNULL(parameters.outboundMaxWindowBits);
+    response = kj::str(response, "; server_max_window_bits=", w);
+  }
+  return kj::mv(response);
+}
+
 } // namespace _ (private)
 namespace {
 
@@ -5618,12 +5657,26 @@ private:
       return sendWebSocketError("Missing Sec-WebSocket-Key");
     }
 
+    kj::Maybe<CompressionParameters> acceptedParameters;
+    kj::String agreedParameters;
+    if (server.settings.enableWebSocketCompression) {
+      KJ_IF_MAYBE(value, requestHeaders.get(HttpHeaderId::SEC_WEBSOCKET_EXTENSIONS)) {
+        KJ_IF_MAYBE(config, _::tryParseExtensionOffers(*value)) {
+          acceptedParameters = kj::mv(*config);
+        }
+      }
+    }
+
     auto websocketAccept = generateWebSocketAccept(key);
 
     kj::StringPtr connectionHeaders[HttpHeaders::WEBSOCKET_CONNECTION_HEADERS_COUNT];
     connectionHeaders[HttpHeaders::BuiltinIndices::SEC_WEBSOCKET_ACCEPT] = websocketAccept;
     connectionHeaders[HttpHeaders::BuiltinIndices::UPGRADE] = "websocket";
     connectionHeaders[HttpHeaders::BuiltinIndices::CONNECTION] = "Upgrade";
+    KJ_IF_MAYBE(parameters, acceptedParameters) {
+      agreedParameters = _::generateExtensionResponse(*parameters);
+      connectionHeaders[HttpHeaders::BuiltinIndices::SEC_WEBSOCKET_EXTENSIONS] = agreedParameters;
+    }
 
     httpOutput.writeHeaders(headers.serializeResponse(
         101, "Switching Protocols", connectionHeaders));
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -3851,6 +3851,62 @@ kj::String generateExtensionResponse(const CompressionParameters& parameters) {
   return kj::mv(response);
 }
 
+kj::OneOf<CompressionParameters, kj::Exception> tryParseExtensionAgreement(
+    const Maybe<CompressionParameters>& clientOffer,
+    StringPtr agreedParameters) {
+  // Like `tryParseExtensionOffers`, but called by the client when parsing the server's Response.
+  // If the client must decline the agreement, we want to provide some details about what went wrong
+  // (since the client has to fail the connection).
+  constexpr auto FAILURE = "Server failed WebSocket handshake: "_kj;
+  auto e = KJ_EXCEPTION(FAILED);
+
+  if (clientOffer == nullptr) {
+    // We've received extensions when we did not send any in the first place.
+    e.setDescription(
+        kj::str(FAILURE, "added Sec-WebSocket-Extensions when client did not offer any."));
+    return kj::mv(e);
+  }
+
+  auto offers = splitParts(agreedParameters, ',');
+  if (offers.size() != 1) {
+    constexpr auto EXPECT = "expected exactly one extension (permessage-deflate) but received "
+                            "more than one."_kj;
+    e.setDescription(kj::str(FAILURE, EXPECT));
+    return kj::mv(e);
+  }
+  auto splitOffer = splitParts(offers.front(), ';');
+
+  if (splitOffer.front() != "permessage-deflate"_kj) {
+    e.setDescription(kj::str(FAILURE, "response included a Sec-WebSocket-Extensions value that was "
+                                      "not permessage-deflate."));
+    return kj::mv(e);
+  }
+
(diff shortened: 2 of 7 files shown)


c++/src/kj/compat/http.c++
```diff
@@ -3489,6 +3489,318 @@ WebSocketPipe newWebSocketPipe() {
...
+        }
+      } else {
+        parts.add(kj::str(cursor));
+      }
+    }
+
+    return parts;
+  }
+
+  inline kj::Maybe<kj::Exception> validateTokenTail(const StringPtr s) {
```
MellowYarker, line 3549: Going to give some context on this as well. Suppose I wanted to verify the parameter `client_max_window_bits`. Previously, I was only checking if `parameter.startsWith("client_max_window_bits")`, which obviously doesn't account for anything _after_the token.

Currently, the `validateCompressionParameters()` function relies on `startsWith`, but it also uses `validateTokenTail()` to be certain there is no invalid stuff _after_ the token. It really boils down to this:
- `permessage-deflate` and `x_no_context_takeover` can be followed by `*LWS`.
- `x_max_window_bits` can be followed by:
    - `*LWS`
    - `*LWS` then an `"="`

We're just verifying that the stuff _after_ the token follows these rules.

c++/src/kj/compat/http.c++
```diff
@@ -3973,6 +4015,75 @@ public:
...
+                // that we did not ask for.
+                auto msg = kj::str(
+                    failed,
+                    "response included a Sec-WebSocket-Extensions value that was not permessage-deflate.");
+                return settings.errorHandler.orDefault(*this).handleWebSocketProtocolError({
+                  502, "Bad Gateway", msg, nullptr
+                });
+              }
+              // Only thing left to do is verify the parameters of our single extension.
+              auto parameters = splitPartsWithCondition(finalOffer, ';');
```
kentonv, line 4054: As with the server side, I'd like to see the parsing code separated from the negotiation logic. We should parse down to a struct first, then in a separate function decide if the options in that struct are acceptable.

kentonv (commented):
Sorry for the long delay reviewing this, I was sick last week and it took a while to catch up.

The code seems generally correct, I mostly have comments on how it could be structured more clearly. I think this is important with this kind of ad hoc parsing code to simplify as much as possible since it's easy to miss subtle bugs otherwise.

FWIW, if you wanted to expose some of the internal functions in order to test them directly, you could do that by putting them in the `kj::_` namespace, which is intended to be private. You could also declare them in a separate header called `http-internal.h` or something, which wouldn't be installed.

c++/src/kj/compat/http.c++
```diff
@@ -3489,6 +3489,318 @@ WebSocketPipe newWebSocketPipe() {
...
+        if (serverBits != nullptr) {
+          // We've seen this parameter already, decline.
+          compressionParameters = nullptr;
+          break;
+        }
+
+        KJ_SWITCH_ONEOF(extractMaxBits(parameters[i])) {
+          KJ_CASE_ONEOF(maybeWindow, kj::Maybe<size_t>) {
+            KJ_IF_MAYBE(window, maybeWindow) {
+              serverBits = *window;
```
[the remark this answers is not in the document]
MellowYarker: Previously, if the server saw `8`, it would reply with `9`. This is wrong, we should reply with `8` and then pass `9` to zlib: https://github.com/capnproto/capnproto/pull/1480/commits/417cbbb073af297d355372b1ee514cb43c1f6adf

c++/src/kj/compat/http.c++
```diff
@@ -5360,12 +5578,51 @@ private:
       return sendWebSocketError("Missing Sec-WebSocket-Key");
     }
 
+    kj::Maybe<CompressionParameters> acceptedParameters;
+    kj::String agreedParameters;
+    KJ_IF_MAYBE(value, requestHeaders.get(HttpHeaderId::SEC_WEBSOCKET_EXTENSIONS)) {
+
+      // TODO (now): It would be nice to use the splitPartsWithCondition() function here, but as the
+      // server, we only want to accept the first compatible extension we see, so it would likely be
```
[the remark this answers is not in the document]
MellowYarker: My understanding is we only want to support `permessage-deflate` (at least for now). The entire config fits in one extension offer and offers are ordered by preference, so if we find a `permessage-deflate` offer we can accept, then we can ignore everything after it.

https://github.com/capnproto/capnproto/pull/1480/commits/001782c4ab2b7ab5a98b232ee9213eed29a635ab

c++/src/kj/compat/http.c++
```diff
@@ -3489,6 +3489,63 @@ WebSocketPipe newWebSocketPipe() {
...
+    KJ_IF_MAYBE(index, advancedBy) {
+      // We found a delimiter, move to 1 past it, then skip over the whitespace.
+      auto startOfNext = const_cast<char*>((input.begin() + *index + 1));
+      result = skipSpace(startOfNext);
+    }
+    return result;
+  }
+
+  kj::Vector<kj::String> splitPartsWithCondition(const kj::StringPtr input, const char& delim,
+      kj::Maybe<StringPtr> condition = nullptr) {
```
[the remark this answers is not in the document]
MellowYarker: Thanks, this is a neat observation!

https://github.com/capnproto/capnproto/pull/1480/commits/f191e739ffd7c8de375d9a5531329eb35575836d

MellowYarker:
Adding a comment to track moving functions into the `kj::_` namespace:
https://github.com/capnproto/capnproto/pull/1480/commits/69681059dbb116cfce031fa6fff749af339ec9a4 fixup for commit 1
https://github.com/capnproto/capnproto/pull/1480/commits/a37b6b62484c0078e21f96fc8073ac5579c143d6 fixup for commit 2
https://github.com/capnproto/capnproto/pull/1480/commits/485f99d4142130c470a45e08aa2a1b9a137912bc fixup for commit 3

c++/src/kj/compat/http.c++
```diff
@@ -3489,6 +3489,63 @@ WebSocketPipe newWebSocketPipe() {
 
 // =======================================================================================
 
+namespace {
+  inline kj::Maybe<StringPtr> advanceToNext(const StringPtr input, const char& c,
```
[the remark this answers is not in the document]
MellowYarker: https://github.com/capnproto/capnproto/pull/1480/commits/3c453a343709638e0e71c39edaf6a5e275ebbc50
https://github.com/capnproto/capnproto/pull/1480/commits/88433ba10c2feaa6aa04fed31622e9fdd8d0e10c
https://github.com/capnproto/capnproto/pull/1480/commits/665778f7ff4d7406da8d5667789ab411617a97ad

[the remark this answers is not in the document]
MellowYarker: https://github.com/capnproto/capnproto/pull/1480/commits/94133b9f4be1cabcabedb933bbb3232db91191d2

c++/src/kj/compat/http.c++
```diff
@@ -3489,6 +3489,63 @@ WebSocketPipe newWebSocketPipe() {
...
+    advancedBy = input.findFirst(c);
+    KJ_IF_MAYBE(index, advancedBy) {
+      // We found a delimiter, move to 1 past it, then skip over the whitespace.
+      auto startOfNext = const_cast<char*>((input.begin() + *index + 1));
+      result = skipSpace(startOfNext);
+    }
+    return result;
+  }
+
+  kj::Vector<kj::String> splitPartsWithCondition(const kj::StringPtr input, const char& delim,
```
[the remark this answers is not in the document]
MellowYarker: Resolved here: https://github.com/capnproto/capnproto/pull/1480#discussion_r898227956

c++/src/kj/compat/http.c++
```diff
@@ -3627,6 +3684,27 @@ public:
...
+      auto extensions = splitPartsWithCondition(*value, ',', permitedExtension);
+
+      // Construct our final extensions string.
+      auto empty = true;
+      for (const auto& ext : extensions) {
+        if (empty) {
+          offeredExtensions = kj::str(ext);
+          empty = false;
+        } else {
+          offeredExtensions = kj::str(KJ_REQUIRE_NONNULL(offeredExtensions), ", ", ext);
```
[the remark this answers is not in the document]
MellowYarker: Thanks, technically the application code could add a bunch of junk `permessage-deflate` offers (such as repeating the same offer `n` times) so the less work we do the better.

https://github.com/capnproto/capnproto/pull/1480/commits/b0f5c7890336613f23a6f038c0028a42284097da

kentonv:
Oops, I left a bunch of comments on fixups and they don't appear to have made it to this thread:

https://github.com/capnproto/capnproto/commit/02bb99b030d30030ebe1b76502a6a0372185cf7a

https://github.com/capnproto/capnproto/commit/69681059dbb116cfce031fa6fff749af339ec9a4#commitcomment-76366939

c++/src/kj/compat/http.c++
```diff
@@ -729,8 +729,17 @@ void HttpHeaders::takeOwnership(HttpHeaders&& otherHeaders) {
 
 // -----------------------------------------------------------------------------
 
-static inline char* skipSpace(char* p) {
+static inline char* skipSpace(char* p, kj::Maybe<size_t> end = nullptr) {
```
kentonv, line 732: I think reusing this function may be more trouble than it's worth at this point, if you have to make modifications like this.

Instead you can implement `stripLeadingSpace()` like:

```
void stripLeadingSpace(ArrayPtr<const char>& str) {
  while (str.size() > 0 && (str[0] == ' ' || str[0] == '\t')) {
    str = str.slice(1, str.size());
  }
}
```

Overall keeping it separate ends up being less code anyway.

c++/src/kj/common.h
```diff
@@ -1748,6 +1749,9 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
     KJ_IREQUIRE(start <= end && end <= size_, "Out-of-bounds ArrayPtr::slice().");
     return ArrayPtr(ptr + start, end - start);
   }
+  inline bool startsWith(const ArrayPtr<const T>& other) const {
+    return other.size_ <= size_ && memcmp(ptr, other.ptr, other.size_) == 0;
```
kentonv, line 1753: `memcmp()` is only correct here if `T` is a basic integral type.

Please do:

```
  inline bool startsWith(const ArrayPtr<const T>& other) const {
    return other.size_ <= size_ && slice(0, other.size_) == other;
  }
```

We already have an optimization so that `operator==` will use `memcmp()` when `T` is integral.

MellowYarker: https://github.com/capnproto/capnproto/pull/1480/commits/3331537fc308364e6062575a5270b35aad53dc89

MellowYarker:
Added a small fixup (https://github.com/capnproto/capnproto/pull/1480/commits/d7deceb124faa72596d51aa48d9d291d6852f883) to correct the window size we choose as the client (we will handle window sizes of 8 later when we actually deal with `zlib`)

c++/src/kj/compat/http.c++
```diff
@@ -3545,6 +3545,224 @@ namespace {
...
+  inline kj::Maybe<kj::Exception> validateTokenTail(const StringPtr s) {
+    // Assert that the extension/parameter token does not include invalid characters between
+    // the end of the valid token and the start of the separator/EOL.
+    //
+    // Suppose END refers to one past the last character we would want to process, for example:
+    //  - if `s` == "client_max_window_bits = 10", then END would be the index of "=".
+    //  - if `s` == "permessage-deflate1234JUNK_VALUES", then END would be s.size()
+    //
+    // We are matching this regex:
+    // ^((client|server)((_max_window_bits\s*(=.*)?)|(_no_context_takeover\s*))$)|(permessage-deflate\s*$)
```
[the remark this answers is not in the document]
MellowYarker: Here's a summary since it feels like a lot changed:
1. Given a collection of `kj::String`s (each element is the result of splitting an offer by `;`), we parse into `key[=value]` pairs (this is the `keyMaybeVal` struct + `toKeysAndVals()` function).
2. Next, try to build an `UnverifiedConfig` struct (basically a structured version of param names). This could fail if the structure of the offer is invalid, ex:
    - Too many parameters (we can have 4 at most).
    - Invalid parameter names (we make sure the `key` part of `key[=value]` is valid here.
    - `x_no_context_takeover` parameters have values (i.e. we got `x_no_context_takeover=blah`), when they shouldn't.
    - See the `populateUnverifiedConfig()` function for details.
3. Finally, we convert from an `UnverifiedConfig` struct to a `CompressionParameters` struct. During this process (`validateCompressionConfig()` function) we check the values (if any are present) for the `x_max_window_bits` parameters.

Had to do this in a few parts to avoid merge conflicts :upside_down_face: 

1. Added the structs and functions mentioned above: https://github.com/capnproto/capnproto/pull/1480/commits/c808f9e04cf590fd6fc37220d143b663917bac30
2. Use new functions when parsing as the server: https://github.com/capnproto/capnproto/pull/1480/commits/92c59bd53040e49b86dfef51a13623131381454d
3. Use new functions when parsing as the client: https://github.com/capnproto/capnproto/pull/1480/commits/565bd44992aa488af22291dd5e220de214923108

I'm going to rebase, then delete the _old_ parsing functions and add that fixup below.

MellowYarker:
Removed tests for functions that were deleted + added tests for the new functions: https://github.com/capnproto/capnproto/pull/1480/commits/b996417081f821ae92b90210ca1be277fceb4932.

I want to add tests for the following functions:
- `toKeysAndVals()`
- `populateUnverifiedConfig()`

Also need to clean up the tests more generally.

MellowYarker:
Cut down line length for formatting reasons: https://github.com/capnproto/capnproto/pull/1480/commits/e57d440d1751dfde9a906d4c00444dbdfb3feecc.

c++/src/kj/compat/http.c++
```diff
@@ -3545,6 +3545,224 @@ namespace {
...
+
+    size_t charsBetweenTokenAndEnd;
+    size_t tokenLength; // Length of the valid token.
+
+    size_t clientOrServerPrefixLen = 7; // Number of chars in "x_" where x is "client" or "server".
+    const StringPtr maxWindowBits = "max_window_bits"_kj;
+    const StringPtr noCtxTakeover = "no_context_takeover"_kj;
+    const StringPtr permessageExt = "permessage-deflate"_kj;
```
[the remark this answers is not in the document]
MellowYarker: Got rid of this here: https://github.com/capnproto/capnproto/pull/1480#discussion_r903127931

c++/src/kj/compat/http.c++
MellowYarker: See: https://github.com/capnproto/capnproto/pull/1480#discussion_r903127931

c++/src/kj/compat/http.c++
```diff
@@ -3677,6 +4027,109 @@ public:
...
+              // single (valid) `permessage-deflate` extension.
+              if (offeredExtensions == nullptr) {
+                // We've received extensions when we did not send any in the first place.
+                auto msg = kj::str(
+                    failed, "added Sec-WebSocket-Extensions when client did not offer any.");
+                return settings.errorHandler.orDefault(*this).handleWebSocketProtocolError({
+                  502, "Bad Gateway", msg, nullptr
+                });
+              }
+              auto offers = _::splitPartsWithCondition(*agreedParameters, ',');
```
kentonv, line 4044: There's a whole bunch of string twiddling here that feels redundant.

Couldn't you reuse `retainClientCompressorConfig()` (or a very similar function) to parse the response from the server, and then compare the parsed structures, rather than carefully pick through the text again here?

c++/src/kj/compat/http.c++
```diff
@@ -5282,12 +5735,56 @@ private:
       return sendWebSocketError("Missing Sec-WebSocket-Key");
     }
 
+    kj::Maybe<CompressionParameters> acceptedParameters;
+    kj::String agreedParameters;
+    KJ_IF_MAYBE(value, requestHeaders.get(HttpHeaderId::SEC_WEBSOCKET_EXTENSIONS)) {
+
+      kj::StringPtr permittedExtension = "permessage-deflate"_kj;
+      auto extensions = _::splitPartsWithCondition(*value, ',', permittedExtension);
```
kentonv, line 5743: More string twiddling, should reuse a common `parseWebSocketExtensions()` function.

c++/src/kj/compat/http.c++
```diff
@@ -3627,6 +3959,23 @@ public:
...
+    kj::Maybe<kj::String> offeredExtensions;
+    kj::Maybe<CompressionParameters> clientOffer;
+    KJ_IF_MAYBE(value, headers.get(HttpHeaderId::SEC_WEBSOCKET_EXTENSIONS)) {
+      // Strip all `Sec-WebSocket-Extensions` except for `permessage-deflate`.
+      kj::StringPtr permittedExtension = "permessage-deflate"_kj;
+      auto extensions = _::splitPartsWithCondition(*value, ',', permittedExtension);
+
+      // Construct our final extensions string.
+      if (extensions.size() > 0) {
+        clientOffer = _::retainClientCompressorConfig(extensions.front());
```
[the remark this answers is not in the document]
MellowYarker: The client can offer the same extension multiple times, but offers are ordered by preference. For example, a client might offer `server_max_window_bits`, a parameter the server is not required to support. The offer will be declined if the server doesn't support the parameter, and if no other offer (of lower preference to the client) is provided, then the server won't agree to use any compression. There's an example in the RFC [here](https://datatracker.ietf.org/doc/html/rfc7692#section-7.1.3) (page 18).

I made it only retain the first one since it would be the highest preference option for the client. It wouldn't really make sense for the client to provide different _values_ across offers since values aren't the cause of extensions being declined (unless they're out of range). By values I mean the value 10 in `client_max_window_bits=10`.

IMO, the "most correct" behavior would be to save the config of the first _valid_ offer. Here's some reasons why:
1. The first offer might be invalid, in which case we currently do not retain anything about the client's offer. Retaining the first _valid_ offer fixes this.
2. The `client_x` params are either accepted or ignored. Their presence cannot result in the server outright declining the offer (in the way the `server_x` params might), so there's no reason for a `client_x` param present in offer `n` to be excluded from any lower preference (subsequent) offers `n+k` where `0 < k`.

To be a bit more concrete, let's say the client provided 3 valid offers that include:
1. `client_max_window_bits=14`
2. `client_max_window_bits=10`
3. `client_max_window_bits=15`

respectively. We'll retain the first one, so `clientOffer = 14`.

If the server agrees to offer (1), then all is good and we'll use `14`. If the server agrees to offer (2), then we'll pick `min(clientOffer, serverAgreement) = min(14, 10) = 10`, which is fine. If the server agrees to offer (3), we'll pick `min(14, 15) = 14`, which might seem weird, but is still valid because technically the server "agrees" by responding with a value less than or equal to that which the client offered. If the server doesn't respond with `client_max_window_bits` at all (but still agrees to use compression), we'll just use `clientOffer=14` (which I'd argue is fine since the client gave multiple values, so it doesn't care for any one in particular).

c++/src/kj/compat/http.h
```diff
@@ -534,6 +534,53 @@ class EntropySource {
...
+namespace _ { // private implementation details
+
+kj::ArrayPtr<const char> splitNext(StringPtr& cursor, char delimiter);
+
+void stripLeadingSpace(ArrayPtr<const char>& str);
+
+kj::Vector<kj::String> splitPartsWithCondition(StringPtr input, char delim,
+    StringPtr condition = nullptr);
+
+struct keyMaybeVal {
```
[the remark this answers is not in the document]
MellowYarker: https://github.com/capnproto/capnproto/pull/1480/commits/96eaa6d754e2053b518bc273f2c9912bb155ec4c

c++/src/kj/compat/http.h
```diff
@@ -534,6 +534,53 @@ class EntropySource {
   virtual void generate(kj::ArrayPtr<byte> buffer) = 0;
 };
 
+struct CompressionParameters {
+  // These are the parameters for `Sec-WebSocket-Extensions` permessage-deflate extension.
+  // Since we cannot distinguish the client/server in `upgradeToWebSocket`, we use the prefixes
+  // `inbound` and `outbound` instead.
+  bool outbound_no_context_takeover = false;
```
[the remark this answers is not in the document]
MellowYarker: `UnverifiedConfig`: https://github.com/capnproto/capnproto/pull/1480/commits/23fe90eaaf67bed6c955948e876cfaf825228498

`CompressionParameters`: `TODO`

c++/src/kj/compat/http.h
```diff
@@ -534,6 +534,53 @@ class EntropySource {
...
+  bool inbound_no_context_takeover = false;
+  kj::Maybe<size_t> outbound_max_window_bits = nullptr;
+  kj::Maybe<size_t> inbound_max_window_bits = nullptr;
+};
+
+namespace _ { // private implementation details
```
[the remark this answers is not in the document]
MellowYarker: https://github.com/capnproto/capnproto/pull/1480/commits/775d0fca81a72911862814a61237668e8cc65682
https://github.com/capnproto/capnproto/pull/1480/commits/785ee7e3371181f08fb48df658b29fa7f57718c9
https://github.com/capnproto/capnproto/pull/1480/commits/06c8d4d36971154d4c489b4a5434959334387ca5

c++/src/kj/compat/http.c++
```diff
@@ -3489,6 +3489,338 @@ WebSocketPipe newWebSocketPipe() {
...
+
+    KJ_IF_MAYBE(index, param.findFirst('=')) {
+      // Found '=' so we have a value.
+      key = param.slice(0, *index);
+      value = param.slice(*index + 1, param.size());
+    } else {
+      key = param.asArray();
+    }
+
+    result[count] = keyMaybeVal{.key = kj::mv(key), .val = kj::mv(value)};
```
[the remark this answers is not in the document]
MellowYarker: https://github.com/capnproto/capnproto/pull/1480/commits/ca8bd4bc78b1f976aa9257c96b28341721f1e34e

Am I correct in expecting `kj::heapArray<KeyMaybeVal>(params.size());` to default initialize each element in the heap allocated array? Judging by the method body of the constructor and this comment in the body of `heapArrayBuilder`:
```
  // Like `heapArray<T>()` but does not default-construct the elements.  You must construct them
  // manually by calling `add()`.
```
I suspect it's fine to assign to the elements of the struct without explicitly initializing them.

c++/src/kj/compat/http.c++
```diff
@@ -3489,6 +3489,338 @@ WebSocketPipe newWebSocketPipe() {
...
+
+  ParamType paramType;
+
+  if (key.startsWith(client)) {
+    paramType.isClientParam = true;
+  } else if (!key.startsWith(server)) {
+    // `Key` must start with either "client" or "server" -- invalid.
+    return nullptr;
+  }
+  key = key.slice(client.size(), key.size());
```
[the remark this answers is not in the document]
MellowYarker: https://github.com/capnproto/capnproto/pull/1480/commits/a451522fca9207d6bcef083beb7360b956373b39

Had to keep it a bit longer to check trailing characters (since trailing whitespace is allowed).

c++/src/kj/compat/http.c++
```diff
@@ -3489,6 +3489,338 @@ WebSocketPipe newWebSocketPipe() {
...
+
+struct ParamType {
+  bool isClientParam = false;
+  // `isClientParam` is true if `key` starts with `client`, false otherwise (implies server).
+  bool isCtxTakeover = false;
+  // `isCtxTakeover` is true if `key` is a `no_context_takeover` parameter, false otherwise
+  // (implies max_window_bits).
+};
+
+inline kj::Maybe<ParamType> isValidKeyName(kj::ArrayPtr<const char>& key) {
```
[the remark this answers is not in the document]
MellowYarker: https://github.com/capnproto/capnproto/pull/1480/commits/3b396f1e82b37d335d743022dd6f89b1cf5878ee

c++/src/kj/compat/http.c++
```diff
@@ -3489,6 +3489,338 @@ WebSocketPipe newWebSocketPipe() {
...
+    result[count] = keyMaybeVal{.key = kj::mv(key), .val = kj::mv(value)};
+    ++count;
+  }
+  return kj::mv(result);
+}
+
+struct ParamType {
+  bool isClientParam = false;
+  // `isClientParam` is true if `key` starts with `client`, false otherwise (implies server).
+  bool isCtxTakeover = false;
```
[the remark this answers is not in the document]
MellowYarker: https://github.com/capnproto/capnproto/pull/1480/commits/3bb5c03e162bba868bacc5ec636a5a556266ddbb

c++/src/kj/compat/http.c++
```diff
@@ -3489,6 +3489,338 @@ WebSocketPipe newWebSocketPipe() {
...
+
+  UnverifiedConfig config;
+
+  for (auto& param : params) {
+    auto paramType = isValidKeyName(param.key);
+    if (paramType == nullptr) {
+      return nullptr;
+    }
+    // `Key` is valid, but we still want to check for repeats.
+    auto isClientParam = KJ_REQUIRE_NONNULL(paramType).isClientParam;
```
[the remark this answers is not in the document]
MellowYarker: https://github.com/capnproto/capnproto/pull/1480#discussion_r907875999

c++/src/kj/compat/http.c++
```diff
@@ -3489,6 +3489,338 @@ WebSocketPipe newWebSocketPipe() {
...
+};
+
+inline kj::Maybe<ParamType> isValidKeyName(kj::ArrayPtr<const char>& key) {
+  // Returns a `ParamType` struct if the `key` is valid and nullptr if invalid.
+  //
+  // The following `key`s are valid, assume x = "client" or "server".
+  //  - x_no_context_takeover
+  //  - x_max_window_bits
+  // Trailing whitespace is allowed -- any other character invalidates the `key`.
+  const auto client = "client"_kj;
```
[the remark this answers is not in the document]
MellowYarker: https://github.com/capnproto/capnproto/pull/1480#discussion_r907861324

c++/src/kj/compat/http.c++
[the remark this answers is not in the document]
MellowYarker: Did this in 2 parts since I found a small bug.
Renamed a parameter in `validateCompressionConfig` that caused confusion: https://github.com/capnproto/capnproto/pull/1480/commits/f6c0e5009063f1dd09414e97a0f47e51f6a5e5be
Made `retainClientCompressorConfig` retain the first valid offer + bugfix: https://github.com/capnproto/capnproto/pull/1480/commits/67c479ad0d1c51e1d34df55015774692567f8d0b

Brief description of the bug:
> We were validating the client's offers as though they were in a Response, in other words, as though they were agreed parameters.

This is wrong because a client can make an offer such as `client_max_window_bits; client_no_context_takeover`. If the max bits parameter has no value in an offer, it's just a hint to the server that the client supports it, but this parameter must have a value _in a Response_. Since we were previously validating the offer as though it was an agreement, `validateCompressionParameters` would return `nullptr` it got the offer above, which meant we wouldn't retain `client_no_context_takeover`. The new behavior is to record `client_max_window_bits = 15` and `client_no_context_takeover = true`.

c++/src/kj/compat/http.c++
```diff
@@ -3627,6 +3959,23 @@ public:
...
+      kj::StringPtr permittedExtension = "permessage-deflate"_kj;
+      auto extensions = _::splitPartsWithCondition(*value, ',', permittedExtension);
+
+      // Construct our final extensions string.
+      if (extensions.size() > 0) {
+        clientOffer = _::retainClientCompressorConfig(extensions.front());
+        offeredExtensions = kj::strArray(extensions, ", ");
+      }
+      KJ_IF_MAYBE(extensions, offeredExtensions) {
+        connectionHeaders[HttpHeaders::BuiltinIndices::SEC_WEBSOCKET_EXTENSIONS] = *extensions;
```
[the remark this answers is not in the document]
MellowYarker: https://github.com/capnproto/capnproto/pull/1480/commits/74e9db442bbddef576cac1139dd3eb32d82a78ef

Whoops, this ended up here because we switched from a for-loop to `kj::strArray`. I think `.emplace()` will help me solve another issue I've been trying to figure out so thanks for mentioning it!

c++/src/kj/compat/http.c++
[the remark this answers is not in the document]
kentonv: But, like, shouldn't we remove all invalid offers from the list before sending them to the server? Otherwise the server might accept an offer that we're not prepared to support?

c++/src/kj/compat/http.c++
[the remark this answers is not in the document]
kentonv: Yes. Alternatively, you could keep the syntax you had before but just remove `.key =` and `.val =`, so the properties are matched up by order rather than by name.

c++/src/kj/compat/http.c++
```diff
@@ -3489,6 +3489,338 @@ WebSocketPipe newWebSocketPipe() {
...
+  // Returns a `ParamType` struct if the `key` is valid and nullptr if invalid.
+  //
+  // The following `key`s are valid, assume x = "client" or "server".
+  //  - x_no_context_takeover
+  //  - x_max_window_bits
+  // Trailing whitespace is allowed -- any other character invalidates the `key`.
+  size_t keySize = 0;
+  ParamType paramType;
+
+  if (key.startsWith("client_no_context_takeover"_kj)) {
```
kentonv, line 3583: Why `startsWith` and not `==`?

kentonv: I guess because you're checking for trailing whitespace below, but couldn't you trip the trailing whitespace _before_ comparing against all the names? That seems more intuitive.

c++/src/kj/compat/http.c++
MellowYarker: The tricky thing is that the way the server should interpret and accept/decline offers is different from how the client should interpret and accept/decline an agreement. They can share _some_ code but having them share a lot might make it more difficult to follow.

Helper functions + moving stuff on server side: https://github.com/capnproto/capnproto/pull/1480/commits/f29cbd448330fe515f6617bc9c9be69b85c1a351
Another helper + moving stuff on client side: https://github.com/capnproto/capnproto/pull/1480/commits/5dbccbcde5b7daebc5a03c57b642b9a2b9dd22d6
Use helper functions in `retainClientCompressorConfig`: https://github.com/capnproto/capnproto/pull/1480/commits/14168b64a6ce0e4f9d1b03eeaab6567d27a1fccc

`TODO:` One thing I want to come back to is `isValidExtensionName()`, since it checks if the extension `startsWith("permessage-deflate")` even though every time we call it (except for [here](https://github.com/capnproto/capnproto/pull/1480/commits/5dbccbcde5b7daebc5a03c57b642b9a2b9dd22d6#diff-09d28b46bd98866f20dd7241e02c70338a76fdb59a483ed02d4c16bbe6bcc3a1R3908)) we already know it does (we passed a condition to `splitPartsWithCondition()`). I've got to run atm but so I'll figure it out tomorrow morning.

c++/src/kj/compat/http.c++
MellowYarker: https://github.com/capnproto/capnproto/pull/1480#discussion_r910503319

Really boils down to: client expects one config that the server agreed to (agreed parameters), and if the client declines this agreement we have to completely fail the connection. OTOH if the server declined _all_ the extensions that the client offered, it would still return a websocket, though the response would not include the `Sec-WebSocket-Extensions` header. 

The relevant commit to this comment is the middle of the three in the comment I linked.

c++/src/kj/compat/http.c++
```diff
@@ -3627,6 +3959,21 @@ public:
...
+    kj::Maybe<kj::String> offeredExtensions;
+    kj::Maybe<CompressionParameters> clientOffer;
+    KJ_IF_MAYBE(value, headers.get(HttpHeaderId::SEC_WEBSOCKET_EXTENSIONS)) {
+      // Strip all `Sec-WebSocket-Extensions` except for `permessage-deflate`.
+      kj::StringPtr permittedExtension = "permessage-deflate"_kj;
+      auto extensions = _::splitPartsWithCondition(*value, ',', permittedExtension);
+
+      // Construct our final extensions string.
+      if (extensions.size() > 0) {
+        clientOffer = _::retainClientCompressorConfig(extensions.slice(0, extensions.size()));
```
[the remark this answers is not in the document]
MellowYarker: Added as part of other fixup: https://github.com/capnproto/capnproto/commit/14168b64a6ce0e4f9d1b03eeaab6567d27a1fccc#diff-09d28b46bd98866f20dd7241e02c70338a76fdb59a483ed02d4c16bbe6bcc3a1R4068

c++/src/kj/compat/http.c++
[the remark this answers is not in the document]
MellowYarker: Hmmm, I guess that depends on what you mean by "accept an offer that we're not prepared to support". We already drop every extension that doesn't start with `permessage-deflate`. Assuming that's fine, if the offer is invalid then the server should drop it, but even if the server doesn't follow the spec, our client will parse the response and fail to connect if it finds anything invalid. We could definitely drop invalid offers on the client side before sending them off, it would just require a little more processing on our end.

c++/src/kj/compat/http.c++
```diff
@@ -4084,95 +4129,26 @@ public:
...
-                auto msg = kj::str(
-                    failed,
-                    "the Sec-WebSocket-Extensions header in the Response included "
-                    "an invalid value.");
-                return settings.errorHandler.orDefault(*this).handleWebSocketProtocolError({
-                  502, "Bad Gateway", msg, nullptr
-                });
               }
-              auto& agreement = KJ_REQUIRE_NONNULL(compressionParameters);
+              auto& agreement = KJ_ASSERT_NONNULL(compressionParameters);
```
MellowYarker, line 4142: Seems like this line is causing some trouble: `C2065: 'compressionParameters': undeclared identifier`.

Just noticed the "_one `KJ_SWITCH_ONEOF()` per block_" in C++14, will fix this in the next fixup.

c++/src/kj/compat/http.c++
kentonv: You don't need `findFirst()`. You can implement stripping whitespace from the back of the array the same way you implement stripping from the front:

```
while (str.size() > 0 && (str.back() == ' ' || str.back() == '\t')) {
  str = str.slice(0, str.size() - 1);
}
```

c++/src/kj/compat/http.c++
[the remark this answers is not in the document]
kentonv: What if the application tries to set some option that is specified by a future version of the spec which we don't support yet, and the server accepts that option? The server will then agree to an offer that we can't actually support, and the connection will fail. That doesn't seem like what we want.

The correct behavior here is to filter out all offers that we don't support.

c++/src/kj/compat/http.c++
```diff
@@ -3489,6 +3489,431 @@ WebSocketPipe newWebSocketPipe() {
...
+      // Invalid parameter.
+      return nullptr;
+    }
+  }
+  return kj::mv(config);
+}
+
+inline kj::Maybe<size_t> extractBitsFromValue(ArrayPtr<const char>&& bits) {
+  kj::Maybe<size_t> window;
+  kj::Maybe<kj::Exception> maybeException = kj::runCatchingExceptions([&]() {
```
kentonv, line 3693: Catching all exceptions like this and assuming the exception must be due to a parse error is not very good form. What if some other assertion fails in here? Also, throwing and catching an exception is very expensive.

Instead, I'd recommend using `strtoul` directly, so that you can check if the parse failed.

Alternatively, maybe we should add `tryParseAs` methods to `StringPtr`, which return a `Maybe` rather than throwing... this has come up a lot.

c++/src/kj/compat/http.c++
MellowYarker: https://github.com/capnproto/capnproto/pull/1480/commits/b7bfdca9454aad3e6b78367b1ba3186f385622bc

c++/src/kj/compat/http.c++
MellowYarker: Added `tryParseAs` to `StringPtr`: https://github.com/capnproto/capnproto/pull/1480/commits/181ab0032975fc4db4079ba3c60fd5e9cf04822d.
Also updated this code to use it: https://github.com/capnproto/capnproto/pull/1480/commits/113ae4304be821d5b3479064462a1279d8183f69

c++/src/kj/compat/http.c++
```diff
@@ -3489,6 +3489,358 @@ WebSocketPipe newWebSocketPipe() {
...
+      validated->inboundMaxWindowBits = validated->outboundMaxWindowBits;
+      validated->outboundMaxWindowBits = tempWindow;
+      result.add(kj::mv(*validated));
+    }
+  }
+
+  return kj::mv(result);
+}
+
+kj::String generatedExtensionRequest(const ArrayPtr<CompressionParameters>& extensions) {
```
MellowYarker, line 3817: Should rename this to `generateExtensionRequest` and add it to `http.h` so it can be tested.

c++/src/kj/compat/http.c++
```diff
@@ -3489,6 +3489,358 @@ WebSocketPipe newWebSocketPipe() {
...
+    // values for `x_no_context_takeover` parameters (the configuration is structured correctly).
+    // All that's left is to check the `x_max_window_bits` values (if any are present).
+    KJ_IF_MAYBE(validConfig, validateCompressionConfig(kj::mv(*unverified), isAgreement)) {
+      return kj::mv(*validConfig);
+    }
+  }
+  return nullptr;
+}
+
+kj::Vector<CompressionParameters> findValidExtensionOffers(StringPtr offers) {
```
MellowYarker, line 3787: Should add this to `http.h` so we can test it.

c++/src/kj/compat/http.c++
```diff
@@ -3880,6 +3880,74 @@ kj::String generateExtensionResponse(const CompressionParameters& parameters) {
   return kj::mv(response);
 }
 
+kj::Maybe<CompressionParameters> retainClientCompressorConfig(const ArrayPtr<String>& offers) {
```
MellowYarker, line 3883: Should delete this as it is no longer needed.

c++/src/kj/compat/http.c++
```diff
@@ -3489,6 +3489,466 @@ WebSocketPipe newWebSocketPipe() {
...
+    } else {
+      // Invalid parameter.
+      return nullptr;
+    }
+  }
+  return kj::mv(config);
+}
+
+inline kj::Maybe<size_t> extractBitsFromValue(ArrayPtr<const char>&& bits) {
+  stripLeadingSpace(bits);
```
kentonv, line 3676: It looks like you already strip leading space before setting `maxBitsSetting` above, so this is redundant?

c++/src/kj/compat/http.c++
```diff
@@ -3489,6 +3489,466 @@ WebSocketPipe newWebSocketPipe() {
...
+  // We also want to strip the whitespace from the back, since `tryParseAs()` gets upset if we have
+  // something like "10   ".
+  while (bits.size() > 0 && (bits.back() == ' ' || bits.back() == '\t')) {
```
kentonv, line 3679: Strip-whitespace-from-end now appears in two places, maybe it should be factored out?

Actually, should `stripLeadingSpace()` be changed to `stripSpace()` and strip both ends? It seems like there are no use cases where you want to strip one end but keep whitespace on the other?

c++/src/kj/compat/http.c++
```diff
@@ -3489,6 +3489,466 @@ WebSocketPipe newWebSocketPipe() {
...
+  } else {
+    result.inboundNoContextTakeover = config.clientNoContextTakeover;
+    result.outboundNoContextTakeover = config.serverNoContextTakeover;
+  }
+  return kj::mv(result);
+}
+
+inline bool isValidExtensionName(StringPtr extension, StringPtr permitted) {
+  // Ensures the extension name is not followed by invalid characters.
+  // Assumes that `extension` starts with `permitted` (caller's responsibility).
```
kentonv, line 3755: This assumption feels unexpected and error-prone.

I guess you designed it this way because `splitPartsWithCondition` already verifies a prefix match and so comparing those characters again would be a waste. However, it's kind of difficult to verify that all the call sites have really satisfied this. `retainClientCompressorConfig()` in particular doesn't actually contain the prefix check, but I guess it's assuming the caller has already filtered offers to the right prefix. But this all feels like something where bugs could be introduced later.

An alternative would be: Change `splitPartsWithCondition` to be just `splitParts()`, always returning all parts, with both leading and trailing whitespace already stripped. Then when you're parsing the list you do something like (pseudocode):

```
for (auto& extension: splitParts(extensions, ',')) {
  auto parts = splitParts(extension, ';');
  if (part[0] != "permessage-deflate"_kj) {
    // skip unknown extension
    continue;
  }

  // ... handle extension ...
}
```

c++/src/kj/compat/http.c++
kentonv: This still seems weird, I don't see why nested switch statements would be a problem, but whatever, this seems fine.

kentonv (commented):
Can we extend `HttpServerSettings` and `HttpClientSettings` with a boolean flag which enables the new compression handling, defaulting off? That would allow us to merge this PR to handle options parsing without waiting for the whole implementation to be done, and it would also give us a way to test compression support more carefully without rolling it out to the whole world all at once.

c++/src/kj/compat/http.c++
MellowYarker: https://github.com/capnproto/capnproto/pull/1480/commits/74580a60b546c599bf2aab71e2f86b52c0bbc8e5

c++/src/kj/compat/http.c++
MellowYarker: Yeah, this turned out to be a better fix than I expected. Whitespace is now stripped only in `splitPartsWithCondition` or `toKeysAndVals`: https://github.com/capnproto/capnproto/pull/1480/commits/7a2088e97181782d4ef87378f1f82ae7fff64f94 (i.e. at the beginning of parsing).

c++/src/kj/compat/http.c++
MellowYarker: https://github.com/capnproto/capnproto/pull/1480#discussion_r916229192

MellowYarker:
> Can we extend `HttpServerSettings` and `HttpClientSettings` with a boolean flag which enables the new compression handling, defaulting off? That would allow us to merge this PR to handle options parsing without waiting for the whole implementation to be done, and it would also give us a way to test compression support more carefully without rolling it out to the whole world all at once.

- Added bool flags to `Settings` + client side: https://github.com/capnproto/capnproto/pull/1480/commits/94408b52cdeec7abac8f9d9f2cbecba0c6619e6c
- Server: https://github.com/capnproto/capnproto/pull/1480/commits/063c892f7b39480eefd3fb901542ce7286e59fd1
- Client resp handling: https://github.com/capnproto/capnproto/pull/1480/commits/c43202318890c0d67aee161586b1083ee3bb9929
- Test: https://github.com/capnproto/capnproto/pull/1480/commits/3ba2a4c58b56cddb61c2b6e6224debc4ff55d638

c++/src/kj/compat/http.c++
```diff
@@ -3677,6 +4095,41 @@ public:
...
+                    *agreedParameters);
+                if (parseResult.is<kj::Exception>()) {
+                  return settings.errorHandler.orDefault(*this).handleWebSocketProtocolError({
+                    502, "Bad Gateway", parseResult.get<kj::Exception>().getDescription(), nullptr});
+                }
+
+                compressionParameters = kj::mv(parseResult.get<CompressionParameters>());
+                auto& agreement = KJ_ASSERT_NONNULL(compressionParameters);
+                KJ_IF_MAYBE(original, clientOffer) {
+                  // Client offered `client_x` parameters, let's compare them with the response.
```
MellowYarker, line 4112: Nit: this comment should probably be updated. Due to changes in how we parse the clients request (before sending it to the server), `clientOffer` will only be `null` if the client _does not_ send any `permessage-deflate` extensions. That is, we enter this `KJ_IF_MAYBE` even if the client offers `permessage-deflate` without any `client_x` parameters.

Also, changing `KJ_IF_MAYBE` to `const auto& original = KJ_ASSERT_NONNULL(clientOffer)` would make the intention more clear, since we always expect `clientOffer` to be `!null` at this point (we throw in `tryParseExtensionAgreement` if `offeredExtensions` is `null`).

Actually, we don't even need to pass `offeredExtensions` to this `.then()` now that `clientOffer` has changed in this way. We could just pass `clientOffer` to `tryParseExtensionAgreement` (instead of `offeredExtensions`) and move the stuff in the block starting at line 4111 into `tryParseExtensionAgreement`.

MellowYarker: Basically this: https://github.com/capnproto/capnproto/pull/1480/commits/03b65e6f59754f7faa1041fdcf0d629aa4ccb7f3, if this is alright I'll rebase and we can merge, otherwise I'll revert the fixup.

MellowYarker:
@kentonv gave [this suggestion](https://github.com/capnproto/capnproto/pull/1480#discussion_r898243240) a go.
- Added `findFirst` to `ArrayPtr`: https://github.com/capnproto/capnproto/pull/1480/commits/2d00f029ff48cbef504d7c05bdc14d864e80cc9e, 
- Switch from `String` to `ArrayPtr<const char>`: https://github.com/capnproto/capnproto/pull/1480/commits/2670c19857fa85fea34a414ffd314ebcff43881b

c++/src/kj/common.h
```diff
@@ -1748,6 +1749,18 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
     KJ_IREQUIRE(start <= end && end <= size_, "Out-of-bounds ArrayPtr::slice().");
     return ArrayPtr(ptr + start, end - start);
   }
+  inline bool startsWith(const ArrayPtr<const T>& other) const {
+    return other.size_ <= size_ && slice(0, other.size_) == other;
+  }
+
+  inline Maybe<size_t> findFirst(char c) const {
```
kentonv, line 1756: This only makes sense if `T` is `char`. If the array is some other type, this is a weird operation that would likely confuse people. For example, if you had an `ArrayPtr<int>` and tried to call `array.findFirst(1)`, it would surprisingly not find the first element with value 1, but would instead find the first individual byte with value 1.

If you want to add `findFirst` I think its parameter needs to be type `const T&`, and you'd have to implement it using a generic `for` loop that tests each element for equality with `==`. You could then implement template explicit specializations for `char` and `byte` that use `memchr`.

c++/src/kj/common.h
```diff
@@ -1753,13 +1753,14 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
...
   }
 
-  inline Maybe<size_t> findFirst(char c) const {
-    const char* pos = reinterpret_cast<const char*>(memchr(ptr, c, size_));
-    if (pos == nullptr) {
-      return nullptr;
-    } else {
-      return pos - ptr;
+  inline Maybe<size_t> findFirst(const T& match) const {
+    for (int i = 0; i < size_; i++) {
```
kentonv, line 1757: Please use `size_t`, not `int`.

c++/src/kj/common.h
```diff
@@ -1753,13 +1753,14 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
...
+      // TODO(now): Do we need to consider if `ptr + i` == nullptr?
```
kentonv, line 1758: How could that be?

c++/src/kj/common.h
```diff
@@ -1753,13 +1753,14 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
...
+      if (*(ptr + i) == match) {
```
kentonv, line 1759: `*(ptr + i)` can also be written `ptr[i]` -- they mean the same thing but the latter is shorter.

kentonv: In fact, fun fact: `*(i + ptr)` is obviously the same as `*(ptr + i)`. And so `i[ptr]` is the same as `ptr[i]`.

c++/src/kj/common.h
MellowYarker: https://github.com/capnproto/capnproto/pull/1480/commits/8a731e27692bcd083d7a4e6052789b23a0db99df

It's been a while, got a bit rusty so I had to write stuff out :grin:. My main concern was a situation where the elements themselves were pointers, so `ArrayPtr<int*>`, with the array of pointers `{0x100, 0x104, nullptr, 0x112}`. I forgot that the underlying `ptr` of `ArrayPtr<T>` is a `T*`, so it would be an `int **`, and `*(ptr + 2) == nullptr` (I was concerned we'd by mistake dereference the `nullptr`).

c++/src/kj/common.h
MellowYarker: That feels so wrong but... also makes sense?

```
~ cat wat.c
#include <stdio.h>

int main() {
	int arr[] = {1, 2};
	int *p = arr;
	printf("%d\n", p[0]);
	printf("%d\n", 0[p]);
	
	if (p[1] == 1[p] && &p[1] == &1[p]) {
		printf("wtf\n");
	} else {
		printf("gcc disagrees\n");
	}
	return 0;
}
~ ./wat
1
1
wtf
~
```
:upside_down_face:

c++/src/kj/compat/http.c++
```diff
@@ -3677,6 +4110,19 @@ public:
...
+            if (settings.enableWebSocketCompression) {
+              KJ_IF_MAYBE(agreedParameters, responseHeaders.get(
+                  HttpHeaderId::SEC_WEBSOCKET_EXTENSIONS)) {
+                auto parseResult = _::tryParseExtensionAgreement(clientOffer,
+                    *agreedParameters);
+                if (parseResult.is<kj::Exception>()) {
+                  return settings.errorHandler.orDefault(*this).handleWebSocketProtocolError({
+                    502, "Bad Gateway", parseResult.get<kj::Exception>().getDescription(), nullptr});
+                }
+              }
```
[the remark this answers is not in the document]
kentonv: It's too late for fixups since this was merged. You could submit a bufix PR, or you could just include the fix as part of your next implementation PR. I assume this bug isn't actually affecting anyone since it's guarded by `enableWebSocketCompression` which no one should be enabling yet (since it's not actually implemented yet).

PR #1482, opened 2022-06-17
Add Android to CI

JonasVautherin:
Is there any interest in that? I was trying it for myself and figured I could share it here, just in case. Feel free to close if it is superfluous :+1:.

Two things:
1. It doesn't run the tests, because the tests need to run `capnp` to generate code (e.g. tests.capnp.c++, I believe), and I would have to jump through hoops to get the `capnp` binary in the dockcross image (and anyway I would need an emulator to run them). So those CI tests just verify that capnproto _builds_ on Android.
2. I added a `--disable-tests` options to the configure script for that reason. `-DBUILD_TESTING` was already there in CMake.

I also have experience with cross-compiling for iOS using CMake, so let me know if there is interest in that, I could open a PR to show how I do it. Though I don't think I would know how to do it with Autoconf/Automake :thinking:.

### DIFF
--- a/.github/workflows/quick-test.yml
+++ b/.github/workflows/quick-test.yml
@@ -64,11 +64,41 @@ jobs:
       - uses: actions/checkout@v2
       - name: install dockcross
         run: |
-            docker run ${{ matrix.docker-run-args }} --rm dockcross/${{ matrix.cross-compiler }} > ./dockcross
-            chmod +x ./dockcross
+            docker run ${{ matrix.docker-run-args }} --rm dockcross/${{ matrix.cross-compiler }} > ./dockcross-${{ matrix.cross-compiler }}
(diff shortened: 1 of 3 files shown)


kentonv:
Hi Jonas,

Hate to tell you this now but we actually already have Android in CI:

https://github.com/capnproto/capnproto/blob/master/.github/workflows/release-test.yml#L94

It's part of the release tests, which means we don't run it on every PR, but it's there, and we even run the tests in the simulator.

We do not have iOS covered in CI. That seems like it would be nice to add. I don't know a single thing about iOS development (and I don't own any Apple hardware) so help would definitely be appreciated on that one.

JonasVautherin:
Oh, I had not realized. It's only running for armeabi-v7a, but it does have the emulator (I may get inspiration from that part for another project :grin:). 

Well, no worries! :blush:

PR #1484, opened 2022-06-17
Add kj::VectorOutputStream::clear()

jrsala-auguration:
My use case is that I want to repeatedly `capnp::writePackedMessage` so I'd like to use and reuse a `kj::VectorOutputStream` for that. If you have a better solution I'll take it but this seemed reasonable.


kentonv:
Seems reasonable. Thanks!

PR #1485, opened 2022-06-20
using absolute includes across library boundary

mikea:
bazel doesn't support relative includes across library boundaries, thus makes it impossible to defined componentized build.


kentonv (approved):
(waiting for CI before merging)

kentonv:
Apparently this broke the Windows builds, which use cmake. You might be able to reproduce with cmake on Linux?

PR #1488, opened 2022-06-21
Allow unused parameter in KJ_STRINGIFY

rebeccasenger:
The parameter `l` is not actually used anywhere in KJ_STRINGIFY function. This change allows the `l` parameter to be unused.

Original compilation error message:
```
source-location.h:104:68: error: unused parameter ‘l’ [-Werror=unused-parameter]
 KJ_UNUSED static kj::String KJ_STRINGIFY(const NoopSourceLocation& l) { 
 ```


kentonv:
This was fixed by #1477. I suppose I should do a point release with a cherry-pick...

PR #1490, opened 2022-06-23
Schema loader for

alexrobomind:
Currently, when using the dynamic API via SchemaLoader, it is possible to convert dynamic objects obtained with a SchemaLoader to C++ objects (as long as the conversion is registered via loadCompiledTypeAndDependencies). However, the opposite direction, passing a dynamic object obtained from a statically-typed C++ object into structs / capabilities from the SchemaLoader, is currently not available.

This is because the SchemaLoader maintains its own copy of the schema, and therefore the C++ object's schema will compare unequal to the types requested by other dynamic structs and/or capabilities.

This pull requests adds a new method, SchemaLoader::getSchemaFor<StructOrCapability>, which constructs a capnp::Schema object equivalent to the static schema in the loader. This is achieved simply by combining SchemaLoader::Impl::loadNative and SchemaLoader::Impl::makeBranded (which is done in a new method added to SchemaLoader:: Impl).

A small test is also added. I tested compilation and the test itself (it is part of capnp-tests the way I wrote it, right?) locally on MSVC 19, CLANG 15 (WSL) and GCC 9 (WSL).

### DIFF
--- a/c++/src/capnp/schema-loader-test.c++
+++ b/c++/src/capnp/schema-loader-test.c++
@@ -400,6 +400,19 @@ TEST(SchemaLoader, LoadStreaming) {
   KJ_EXPECT(results.getShortDisplayName() == "StreamResult", results.getShortDisplayName());
 }
 
+TEST(SchemaLoader, SchemaFor) {
+  SchemaLoader loader;
+  
+  using TargetCls = test::TestLists;
+  
+  Schema nativeSchema = Schema::from<TargetCls>();
+  Schema loaderFromProto = loader.load(nativeSchema.getProto());
+  Schema loaderFromSchemaFor = loader.getSchemaFor<TargetCls>();
+  
+  EXPECT_FALSE(nativeSchema == loaderFromProto);
+  EXPECT_TRUE(loaderFromSchemaFor == loaderFromProto);
+}
+
 }  // namespace
 }  // namespace _ (private)
 }  // namespace capnp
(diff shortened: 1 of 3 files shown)


alexrobomind:
After sleeping over this, I noticed I might not have fully thought this through. The way this is written right now, it binds the brand parameters to the RawBrandedSchema pointer of the C++ struct. So I would need to recursively process all the brand parameters first. I would stick close to the implementation of makeBranded which loads from a Brand::Reader.

Do I need to perform any sorts of validations? Particularly, do I need to do recursion checks of some sort? I would think this to be unneccessary, as I only process builtin schemas from C++ code here.

kentonv:
Oof, it's been quite a few years since I wrote this code and it's going to take me a while to page back in how all this works in order to review this.

As an alternative, would it work to use `loader.loadCompiledTypeAndDependencies<T>()` to make sure the type is loaded, and then just call `loader.get(capnp::typeId<T>())` to get the schema by its type ID?

alexrobomind:
For parameterless types, that would work. However, the code I am using also has some branded types (which are really important), so I would have to figure out how to reconstruct schema::Brand instances from the Schema class (which uses its own Brand representation) to pass into the SchemaLoader. There is one hurdle here:

The Schema class gives no list of potential scopes to query for brand bindings. I'm not sure whether I can just rely on climbing up the node hierarchy, as some of the parent nodes might not be loaded by the schema loader (loadCompiledTypeAndDependencies doesn't seem to load parent nodes). To be sure that I get the right scope binding list (if I don't match the exact list in RawBrandedSchema, then Schema::operator== will complain), I would have to put a method into Schema that would return this information (that would go into a separate PR, but that would be extremely simple to review).

Beyond that, I should be able to figure this out completely on my side. The idea of this PR was just based on the observation that the low-level infrastructure seemed remarkably simple to extend here, and that someone else might need this in the future. But if the demand isn't there, I'm totally happy to handle this in the library I'm building.

Thanks for your response. If you would be fine with the above approach, then simply close this PR.

PS: I still am working on the fixes for the DynamicStruct::Pipeline issues that I talked about. I just had to put it a bit back in the queue.

PR #1493, opened 2022-07-05
Added method to list brand binding scopes

alexrobomind:
As mentioned in #1490, in order to accurately reconstruct brand bindings in a way that will lead to identically comparing schemas, I need to be able to get an exact list of binding scopes. This PR adds a new method to the schema class that lists all the brand binding scopes of the underlying RawBrandedSchema. This should hopefully be both simpler to review and a lower-risk addition, as it only reads information instead of mutating the schema loader's state.

### DIFF
--- a/c++/src/capnp/schema.c++
+++ b/c++/src/capnp/schema.c++
@@ -265,6 +265,18 @@ Schema::BrandArgumentList Schema::getBrandArgumentsAtScope(uint64_t scopeId) con
   return BrandArgumentList(scopeId, raw->isUnbound());
 }
 
+kj::Array<uint64_t> Schema::listBrandArgumentScopes() const {
+  KJ_REQUIRE(getProto().getIsGeneric(), "Not a generic type.", getProto().getDisplayName());
+  
+  auto result = kj::heapArray<uint64_t>(raw->scopeCount);
+  for(auto iScope : kj::indices(result)) {
+	result[iScope] = raw->scopes[iScope].typeId;
+  }
+  
+  return result;
+}
+	
+
 StructSchema Schema::asStruct() const {
   KJ_REQUIRE(getProto().isStruct(), "Tried to use non-struct schema as a struct.",
              getProto().getDisplayName()) {
--- a/c++/src/capnp/schema.c++
+++ b/c++/src/capnp/schema.c++
@@ -265,8 +265,9 @@ Schema::BrandArgumentList Schema::getBrandArgumentsAtScope(uint64_t scopeId) con
   return BrandArgumentList(scopeId, raw->isUnbound());
 }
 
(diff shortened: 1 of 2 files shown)


c++/src/capnp/schema.c++
```diff
@@ -265,6 +265,18 @@ Schema::BrandArgumentList Schema::getBrandArgumentsAtScope(uint64_t scopeId) con
   return BrandArgumentList(scopeId, raw->isUnbound());
 }
 
+kj::Array<uint64_t> Schema::listBrandArgumentScopes() const {
+  KJ_REQUIRE(getProto().getIsGeneric(), "Not a generic type.", getProto().getDisplayName());
+  
+  auto result = kj::heapArray<uint64_t>(raw->scopeCount);
+  for(auto iScope : kj::indices(result)) {
+	result[iScope] = raw->scopes[iScope].typeId;
+  }
```
kentonv, line 274: Style nitpicking.

```suggestion
  for (auto iScope: kj::indices(result)) {
    result[iScope] = raw->scopes[iScope].typeId;
  }
```

c++/src/capnp/schema.c++
```diff
@@ -265,6 +265,18 @@ Schema::BrandArgumentList Schema::getBrandArgumentsAtScope(uint64_t scopeId) con
   return BrandArgumentList(scopeId, raw->isUnbound());
 }
 
+kj::Array<uint64_t> Schema::listBrandArgumentScopes() const {
+  KJ_REQUIRE(getProto().getIsGeneric(), "Not a generic type.", getProto().getDisplayName());
```
kentonv, line 269: Maybe this should just return an empty array (`return nullptr`)?

c++/src/capnp/schema.h
```diff
@@ -125,6 +125,8 @@ class Schema {
   class BrandArgumentList;
   BrandArgumentList getBrandArgumentsAtScope(uint64_t scopeId) const;
   // Gets the values bound to the brand parameters at the given scope.
+  
+  kj::Array<uint64_t> listBrandArgumentScopes() const;
```
kentonv, line 129: Bikeshedding...

```suggestion
  kj::Array<uint64_t> getGenericScopeIds() const;
  // Returns the type IDs of all parent scopes that have generic parameters, to which this type is
  // subject.
```

c++/src/capnp/schema.c++
alexrobomind: I was simply following the existing API laid out by getBrandArgumentsAtScope. Returning an empty array makes sense .

alexrobomind:
All suggestions & comments implemented.

kentonv:
Ugghhh forgot to squash+merge. Oh well.

PR #1496, opened 2022-07-12
[bazel] compile capnp

mikea:
Provides cc_capnp_library for capnp compilation.

### DIFF
--- a/c++/src/capnp/BUILD.bazel
+++ b/c++/src/capnp/BUILD.bazel
@@ -1,0 +1,245 @@
+load("@capnp-cpp//src/capnp:cc_capnp_library.bzl", "cc_capnp_library")
+
+cc_library(
+    name = "capnp",
+    srcs = [
+        "any.c++",
+        "arena.c++",
+        "blob.c++",
+        "c++.capnp.c++",
+        "dynamic.c++",
+        "layout.c++",
+        "list.c++",
+        "message.c++",
+        "schema.c++",
+        "schema.capnp.c++",
+        "schema-loader.c++",
+        "serialize.c++",
+        "serialize-packed.c++",
+        "stream.capnp.c++",
+        "stringify.c++",
+    ],
+    hdrs = [
+        "any.h",
+        "arena.h",
+        "blob.h",
+        "c++.capnp.h",
+        "capability.h",
+        "common.h",
+        "dynamic.h",
+        "endian.h",
+        "generated-header-support.h",
+        "layout.h",
+        "list.h",
+        "membrane.h",
+        "message.h",
+        "orphan.h",
+        "pointer-helpers.h",
+        "pretty-print.h",
+        "raw-schema.h",
+        "schema.capnp.h",
+        "schema.h",
+        "schema-lite.h",
+        "schema-loader.h",
+        "schema-parser.h",
+        "serialize.h",
+        "serialize-async.h",
+        "serialize-packed.h",
+        "serialize-text.h",
+        "stream.capnp.h",
+    ],
+    include_prefix = "capnp",
+    visibility = ["//visibility:public"],
+    deps = [
+        "//src/kj:kj-async",
+    ],
+)
+
+cc_library(
+    name = "capnp-rpc",
+    srcs = [
+        "capability.c++",
+        "dynamic-capability.c++",
+        "ez-rpc.c++",
+        "membrane.c++",
+        "persistent.capnp.c++",
+        "reconnect.c++",
+        "rpc.c++",
+        "rpc.capnp.c++",
+        "rpc-twoparty.c++",
+        "rpc-twoparty.capnp.c++",
+        "serialize-async.c++",
+    ],
+    hdrs = [
+        "ez-rpc.h",
+        "persistent.capnp.h",
+        "reconnect.h",
+        "rpc.capnp.h",
+        "rpc.h",
+        "rpc-prelude.h",
+        "rpc-twoparty.capnp.h",
+        "rpc-twoparty.h",
+    ],
+    include_prefix = "capnp",
+    visibility = ["//visibility:public"],
+    deps = [
+        ":capnp",
+    ],
+)
+
+cc_library(
+    name = "capnpc",
+    srcs = [
+        "compiler/compiler.c++",
+        "compiler/error-reporter.c++",
+        "compiler/generics.c++",
+        "compiler/grammar.capnp.c++",
+        "compiler/lexer.c++",
+        "compiler/lexer.capnp.c++",
+        "compiler/node-translator.c++",
+        "compiler/parser.c++",
+        "compiler/type-id.c++",
+        "schema-parser.c++",
+        "serialize-text.c++",
+    ],
+    hdrs = [
+        "compiler/compiler.h",
+        "compiler/error-reporter.h",
+        "compiler/generics.h",
+        "compiler/grammar.capnp.h",
+        "compiler/lexer.capnp.h",
+        "compiler/lexer.h",
+        "compiler/module-loader.h",
+        "compiler/node-translator.h",
+        "compiler/parser.h",
+        "compiler/resolver.h",
+        "compiler/type-id.h",
+    ],
+    include_prefix = "capnp",
+    visibility = ["//visibility:public"],
+    deps = [
+        ":capnp",
+    ],
+)
+
+cc_binary(
+    name = "capnp_tool",
+    srcs = [
+        "compiler/capnp.c++",
+        "compiler/module-loader.c++",
+    ],
+    visibility = ["//visibility:public"],
+    deps = [
+        ":capnpc",
+        "//src/capnp/compat:json",
+    ],
+)
+
+cc_binary(
+    name = "capnpc-c++",
+    srcs = [
+        "compiler/capnpc-c++.c++",
+    ],
+    visibility = ["//visibility:public"],
+    deps = [
+        ":capnpc",
+    ],
+)
+
+cc_binary(
+    name = "capnpc-capnp",
+    srcs = [
+        "compiler/capnpc-capnp.c++",
+    ],
+    visibility = ["//visibility:public"],
+    deps = [
+        ":capnpc",
+    ],
+)
+
+# capnp files that are implicitly available for import to any .capnp.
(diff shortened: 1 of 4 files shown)


c++/src/capnp/BUILD.bazel
```diff
@@ -0,0 +1,245 @@
...
+    ],
+)
+
+filegroup(
+    name = "testdata",
+    srcs = glob(["testdata/**/*"]),
+)
+
+cc_capnp_library(
+    name = "test_capnp",
```
uhthomas, line 189: the convention is capnp_test, or capnp_test_library I believe

c++/src/capnp/cc_capnp_library.bzl
```diff
@@ -0,0 +1,114 @@
...
+        includes.append(src_prefix)
+
+    system_include = _workspace_path(ctx.attr._capnp_system.label, "src")
+
+    out_dir = ctx.var["GENDIR"]
+    if src_prefix != "":
+        out_dir = out_dir + "/" + src_prefix
+
+    cc_out = "-o%s:%s" % (ctx.executable._capnpc_cxx.path, out_dir)
+    args = ["compile", "--verbose", cc_out] + \
```
uhthomas, line 34: prefer ctx.actions.args()

c++/src/capnp/cc_capnp_library.bzl
```diff
@@ -0,0 +1,114 @@
+"""Bazel rule to compile .capnp files into c++."""
+
+capnp_provider = provider("Capnproto Provider", fields = {
+    "includes": "includes for this target (transitive)",
+    "inputs": "src + data for the target",
+    "src_prefix": "src_prefix of the target",
+})
+
+def _workspace_path(label, path):
```
uhthomas, line 9: why is this needed? shouldn't the path generated by bazel be fine?

c++/src/capnp/cc_capnp_library.bzl
```diff
@@ -0,0 +1,114 @@
...
+_capnp_gen = rule(
+    attrs = {
+        "srcs": attr.label_list(allow_files = True),
+        "deps": attr.label_list(providers = [capnp_provider]),
+        "data": attr.label_list(allow_files = True),
+        "outs": attr.output_list(),
+        "src_prefix": attr.string(),
+        "_capnpc": attr.label(executable = True, allow_single_file = True, cfg = "exec", default = "@capnp-cpp//src/capnp:capnp_tool"),
+        "_capnpc_cxx": attr.label(executable = True, allow_single_file = True, cfg = "exec", default = "@capnp-cpp//src/capnp:capnpc-c++"),
+        "_capnpc_capnp": attr.label(executable = True, allow_single_file = True, cfg = "exec", default = "@capnp-cpp//src/capnp:capnpc-capnp"),
```
uhthomas, line 67: maybe use toolchains instead?

c++/src/capnp/cc_capnp_library.bzl
mikea: This is needed to generate -I directives when capnp is an external repo. Then this essentially adds "external/" prefix to paths where needed.

c++/src/capnp/cc_capnp_library.bzl
mikea: Frankly, after setting up c++ toolchain it seems like way too much work for very little benefit.

c++/src/capnp/cc_capnp_library.bzl
uhthomas: Bazel 6 removes the external/ stuff. Given this is a rule I really don't see why the file can't be passed to ctx.actions.args() and expanded automatically by Bazel.

c++/src/capnp/cc_capnp_library.bzl
uhthomas: The c++ toolchain stuff is a very different and tedious beast compared to regular toolchains. A capnp toolchain would be very simple and very beneficial. I can make a mock-up tomorrow if you like.

uhthomas:
also please add an empty `BUILD.bazel` file to `c++/`. Otherwise errors like:

```
(15:37:58) ERROR: /private/var/tmp/_bazel_thomas/796dd52d4d981101108548a9446ddca6/external/crate_index__segments-rpc-0.0.0/BUILD.bazel:5:22: no such package '@com_github_capnproto_capnproto//': The repository '@com_github_capnproto_capnproto' could not be resolved: Repository '@com_github_capnproto_capnproto' is not defined and referenced by '@crate_index__segments-rpc-0.0.0//:segments-rpc_build_script'
```

will occur.

c++/src/capnp/cc_capnp_library.bzl
mikea: > Given this is a rule I really don't see why the file can't be passed to ctx.actions.args() and expanded automatically by Bazel.

This doesn't get files as input, but directories. E.g. src-prefix is not a file but part of the name of the file, etc. I don't see how args will help.

mikea:
> If we're thinking of providing native rules, maybe we should also look at https://github.com/kgreenek/rules_capnproto? It seems quite fleshed out.

I feel that implementation is needlessly complicated and will definitely require lots of wrangling to adopt to this and runtime situation. I also think we prefer the original implementation for the speed.

c++/src/capnp/cc_capnp_library.bzl
mikea: I gave it a bit of thought and I don't think toolchain is necessary in this case. These files build capnp from these sources and use it. The goal is not to provide a separate independent implementation of cc_capnp_library. I don't think we need a toolchain complexity here.

mikea:
Comments address. PTAL @uhthomas

c++/src/capnp/cc_capnp_library.bzl
uhthomas: Given

```
"_capnp_system": attr.label(default = "@capnp-cpp//src/capnp:capnp_system_library"),
```

Can't this be changed to `attr.label(allow_files = True)` and then use `ctx.files._capnp_system[0].dirname`? I just don't see how appending the workspace root makes any sense within an action.

mikea: src_prefix is not like this anyway.

mikea: but done for capnp_system

build/.cmake/api/v1/query/client-vscode/query.json
```diff
@@ -0,0 +1 @@
+{"requests":[{"kind":"cache","version":2},{"kind":"codemodel","version":2},{"kind":"toolchains","version":1}]}
```
uhthomas, line 1: ?

c++/src/capnp/cc_capnp_library.bzl
```diff
@@ -0,0 +1,119 @@
...
+        name = name + "_gen",
+        srcs = srcs,
+        deps = [s + "_gen" for s in deps],
+        data = data,
+        outs = hdrs + srcs_cpp,
+        src_prefix = src_prefix,
+        visibility = visibility,
+        target_compatible_with = target_compatible_with,
+    )
+    native.cc_library(
```
uhthomas, line 111: Why not just use [CcInfo](https://docs.bazel.build/versions/main/integrating-with-rules-cc.html#implementing-starlark-rules-that-depend-on-c-rules-andor-that-c-rules-can-depend-on)?

example from rules_go: https://github.com/bazelbuild/rules_go/blob/df02d01fb37701ff826697271220f043b364d1c1/go/private/rules/binary.bzl#L72

c++/src/capnp/cc_capnp_library.bzl
```diff
@@ -0,0 +1,119 @@
...
+
+def cc_capnp_library(
+        name,
+        srcs = [],
+        data = [],
+        deps = [],
+        src_prefix = "",
+        visibility = None,
+        target_compatible_with = None,
+        **kargs):
```
uhthomas, line 83: kwargs?

c++/src/capnp/cc_capnp_library.bzl
```diff
@@ -0,0 +1,119 @@
...
+        "data": attr.label_list(allow_files = True),
+        "outs": attr.output_list(),
+        "src_prefix": attr.string(),
+        "_capnpc": attr.label(executable = True, allow_single_file = True, cfg = "exec", default = "@capnp-cpp//src/capnp:capnp_tool"),
+        "_capnpc_cxx": attr.label(executable = True, allow_single_file = True, cfg = "exec", default = "@capnp-cpp//src/capnp:capnpc-c++"),
+        "_capnpc_capnp": attr.label(executable = True, allow_single_file = True, cfg = "exec", default = "@capnp-cpp//src/capnp:capnpc-capnp"),
+        "_capnp_system": attr.label(default = "@capnp-cpp//src/capnp:capnp_system_library"),
+    },
+    output_to_genfiles = True,
+    implementation = _impl,
```
uhthomas, line 72: prefer `_capnp_gen_impl` as convention

c++/src/capnp/BUILD.bazel
```diff
@@ -0,0 +1,245 @@
+load(":cc_capnp_library.bzl", "cc_capnp_library")
```
uhthomas, line 1: prefer absolute paths (//c++/src/capnp:cc_capnp_library)

c++/src/capnp/cc_capnp_library.bzl
mikea: I'm not sure what that is, it will take me some time to figure it out. Can I postpone this to a future PR?

Also what are the benefits of it? It is a cc_library anyway.

mikea:
@uhthomas I can't get rid of remaining workspace path because of src/prefix, otherwise I think I addressed everything. Is there anything else you'd like me to change?

@kentonv could you take a quick look too?

kkoenig (approved):

kkoenig:
> @uhthomas I can't get rid of remaining workspace path because of src/prefix, otherwise I think I addressed everything. Is there anything else you'd like me to change?
> 
> @kentonv could you take a quick look too?

I took a look

PR #1501, opened 2022-07-18
[bazel] adding remaining capnp tests

mikea:
Fixing relative include path across component boundariess.

### DIFF
--- a/c++/src/capnp/reconnect.h
+++ b/c++/src/capnp/reconnect.h
@@ -21,7 +21,7 @@
 
 #pragma once
 
-#include "capability.h"
+#include <capnp/capability.h>
 #include <kj/function.h>
 
 CAPNP_BEGIN_HEADER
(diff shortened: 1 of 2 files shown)


kkoenig (approved):

c++/src/capnp/reconnect.h
```diff
@@ -21,7 +22,7 @@
 
 #pragma once
 
-#include "capability.h"
+#include <capnp/capability.h>
```
kentonv, line 25: Isn't this in the same library?

mikea: It is `capnp-rpc`, not `capnp`.

kentonv: Oh I see, capability.h is in the base library while capability.c++ is in the RPC library, due to the weird compile-time vs. link-time dependency differences.

Issue #1498, opened 2022-07-13
CAPNP_LITE still includes schema defs in binary


Vadmeme:
On release builds in gcc and msvc the compiler does not optimize away the binary data describing the protocol if you have `CAPNP_LITE` defined.

The issue seems to be caused by the `capnp::_::AlignedData` structure not being within the bounds of the `CAPNP_LITE` check

```cpp
static const ::capnp::_::AlignedData<37> b_deadbeefcafebabe = {
  {   /* omitted */ }
};
::capnp::word const* const bp_deadbeefcafebabe = b_deadbeefcafebabe.words;
#if !CAPNP_LITE
static const uint16_t m_deadbeefcafebabe[] = {0};
static const uint16_t i_deadbeefcafebabe[] = {0};
const ::capnp::_::RawSchema s_deadbeefcafebabe = {
  0xdeadbeefcafebabe, b_deadbeefcafebabe.words, 37, nullptr, m_deadbeefcafebabe,
  0, 1, i_deadbeefcafebabe, nullptr, nullptr, { &s_deadbeefcafebabe, nullptr, nullptr, 0, 0, nullptr }
};
```

It should be solvable quite easily by moving up the `#if !CAPNP_LITE` to also include the schema data in [c++/src/capnp/compiler/capnpc-c++.c++](https://github.com/capnproto/capnproto/blob/045d5ff0e50cd044c1f05925789d5c3e46d96d21/c%2B%2B/src/capnp/compiler/capnpc-c%2B%2B.c%2B%2B#L2779)

kentonv:
This is intended. `CAPNP_LITE` was specifically designed to trim away features that, at the time, we couldn't get to compile correctly in MSVC. Since then, MSVC has improved and lite mode is essentially obsolete.

Lite mode was not designed as a way to reduce binary size. If this is your goal, my recommendation is to use a linker that does tree shaking / dead code elimination to remove symbols that aren't referenced anywhere. This should remove all the embedded schemas, assuming you in fact don't use them.

Vadmeme:
Thank you for your clarification

Lite mode is quite helpful for sensitive protocols where having information leak (such as field names) would cause problems. 

I ask you to consider implementing such a change as it shouldnt affect any existing users while adding extra use cases.

For now I will look into linker options to strip such data like you suggested.

kentonv:
[...]

PR #1502, opened 2022-07-19
Implement $Cxx.omitSchemas annotation

Vadmeme:
Implements a `$Cxx.omitSchemas` annotation that strips schema information from the binary as discussed in https://github.com/capnproto/capnproto/issues/1498


c++/src/capnp/c++.capnp
```diff
@@ -24,3 +24,4 @@ $namespace("capnp::annotations");
 
 annotation namespace(file): Text;
 annotation name(field, enumerant, struct, enum, interface, method, param, group, union): Text;
+annotation omitSchemas(file): Void;
```
kentonv, line 27: Let's add some comments (and a trailing newline).

```suggestion

annotation omitSchemas(file): Void;
# Requests that the generated code does not embed the types' schemas.
# This means you will not be able to use <capnp/schema.h> or <capnp/dynamic.h> APIs with this type.
# By extension, that means you will not be able to stringify messages nor use the JSON codec.
# However, the generated code will be somewhat smaller and will not include easily-extractable
# type information, which may be important in some use cases.
#
# All files that import your file will also be required to set this option if they are used to
# generate C++ code.
```

c++/src/capnp/compiler/capnpc-c++.c++
```diff
@@ -3045,7 +3054,12 @@ private:
           "#if CAPNP_VERSION != ", CAPNP_VERSION, "\n"
           "#error \"Version mismatch between generated code and library headers.  You must "
               "use the same version of the Cap'n Proto compiler and library.\"\n"
-          "#endif\n"
+          "#endif\n",
+          omitSchemas ? kj::strTree(
+                  "#if !CAPNP_LITE\n",
+                  "#error \"$Cxx.omitSchemas requires that you compile using CAPNP_LITE\"\n"
```
kentonv, line 3060: Why does it require this? I would expect it to work fine without CAPNP_LITE (as long as you don't try to use the schema or dynamic APIs with the type, and all dependent types also omit schemas). Is there a particular error you're seeing?

kentonv:
Let's make sure this is tested. I think you could just apply the option to `test-import2.capnp`, to verify it still compiles. You might want to add some declarations in that file to cover enums, interfaces, etc. to make sure everything is covered.

c++/src/capnp/compiler/capnpc-c++.c++
Vadmeme: You are correct, I added this before I fully understood the generated code, After running some tests it seems it is not required.

Vadmeme:
I did the simple test of adding the annotation to the existing `test-import2.capnp` file
I also added some extra declarations to handle interfaces/enums/structs as per your suggestion.

c++/src/capnp/c++.capnp
```diff
@@ -24,3 +24,13 @@ $namespace("capnp::annotations");
...
+
+annotation omitSchemas(file): Void;
+# Requests that the generated code does not embed the types' schemas.
+# This means you will not be able to use <capnp/schema.h> or <capnp/dynamic.h> APIs with this type.
+# By extension, that means you will not be able to stringify messages nor use the JSON codec.
+# However, the generated code will be somewhat smaller and will not include easily-extractable
+# type information, which may be important in some use cases.
+#
+# All files that import your file will also be required to set this option if they are used to
+# generate C++ code.
```
kentonv, line 36: Missing newline at end of file.

(This may seem nitpicky, but it's problematic mostly because so many tools complain about it, like the review tools here. The reason they all complain is because some old Unix standard actually says that text files must end with a newline, though obviously nothing enforces that in reality.)

Annoyingly it doesn't look like GitHub's suggested changes interface lets me suggest adding the newline... it complains that I haven't changed anything, unless I had two newlines...

c++/src/capnp/compiler/capnpc-c++.c++
```diff
@@ -3045,7 +3054,7 @@ private:
           "#if CAPNP_VERSION != ", CAPNP_VERSION, "\n"
           "#error \"Version mismatch between generated code and library headers.  You must "
               "use the same version of the Cap'n Proto compiler and library.\"\n"
-          "#endif\n"
+          "#endif\n",
```
kentonv, line 3057: Line was changed unnecessarily.

```suggestion
          "#endif\n"
```

c++/src/capnp/compiler/capnpc-c++.c++
```diff
@@ -2985,6 +2986,14 @@ private:
...
     kj::Vector<kj::ArrayPtr<const char>> namespaceParts;
     kj::String namespacePrefix;
+    bool omitSchemas = false;
+
+    for(auto annotation : node.getAnnotations()) {
+        if(annotation.getId() == OMIT_SCHEMAS_ANNOTATION_ID) {
+            omitSchemas = true;
+            break;
+        }
+    }
```
kentonv, line 2996: Please match prevailing formatting style.

```suggestion
    for (auto annotation : node.getAnnotations()) {
      if (annotation.getId() == OMIT_SCHEMAS_ANNOTATION_ID) {
        omitSchemas = true;
        break;
      }
    }
```

c++/src/capnp/test-import2.capnp
```diff
@@ -21,12 +21,36 @@
...
+  one @1;
+  two @2;
+  three @3;
+  four @4;
+}
+
+struct TestStructure {
+  enumValue @0 :TestEnum;
+  anInterface @1 :TestExtendingInterface;
+}
```
kentonv, line 56: Missing newline at end of file.

c++/src/capnp/test-import2.capnp
```diff
@@ -21,12 +21,36 @@
...
 
+$Cxx.omitSchemas;
+
 struct TestImport2 {
   foo @0 :Import3.TestAllTypes;
   bar @1 :Import1.Node;
   baz @2 :Import2.TestImport;
 }
+
+interface TestInterface {
```
kentonv, line 37: Let's give these slightly more unique names for the file.

```suggestion
interface TestImport2Interface {
```

c++/src/capnp/test-import2.capnp
```diff
@@ -21,12 +21,36 @@
...
+  testInterfaceMethod @0 () -> (result :Bool);
+}
+
+interface TestExtendingInterface extends(TestInterface) {
```
kentonv, line 41:
```suggestion
interface TestImport2ExtendingInterface extends(TestImport2Interface) {
```

c++/src/capnp/test-import2.capnp
```diff
@@ -21,12 +21,36 @@
...
+  testInterfaceMethodTwo @0 () -> (result :Bool);
+}
+
+enum TestEnum {
```
kentonv, line 45:
```suggestion
enum TestImport2Enum {
```

c++/src/capnp/test-import2.capnp
```diff
@@ -21,12 +21,36 @@
...
+  zero @0;
+  one @1;
+  two @2;
+  three @3;
+  four @4;
+}
+
+struct TestStructure {
```
kentonv, line 53:
```suggestion
struct TestImport2Structure {
```

kentonv (commented):
Thanks for adding tests! Just a bunch of style nitpicks...

Vadmeme:
Implemented style changes and added newlines

kentonv:
Looks good, thanks! Bazel failure seems unrelated.

mikea:
> Bazel failure seems unrelated.

I think this change actually broke bazel build. I don't know why yet, but it passes on a previous commit. @kentonv @Vadmeme

mikea:
Other builds fail because of the same error:


```
/usr/bin/ld: CMakeFiles/capnp-heavy-tests.dir/test_capnp/capnp/test-import2.capnp.c++.o:(.data.rel.ro+0x0): undefined reference to `capnp::schemas::s_cfdb43ee5505db69'
/usr/bin/ld: CMakeFiles/capnp-heavy-tests.dir/test_capnp/capnp/test-import2.capnp.c++.o:(.data.rel.ro+0x8): undefined reference to `capnp::schemas::s_92adc979fd930df5'
make[5]: Leaving directory '/home/runner/work/capnproto/capnproto/c++/distcheck-cmake'
/usr/bin/ld: CMakeFiles/capnp-heavy-tests.dir/test_capnp/capnp/test-import2.capnp.c++.o:(.data.rel.ro+0x10): undefined reference to `capnp::schemas::s_9b550ca7cf5da1c1'
make[4]: Leaving directory '/home/runner/work/capnproto/capnproto/c++/distcheck-cmake'
/usr/bin/ld: CMakeFiles/capnp-heavy-tests.dir/test_capnp/capnp/test-import2.capnp.c++.o:(.data.rel.ro+0x18): undefined reference to `capnp::schemas::s_f2756f068e360761'
make[3]: Leaving directory '/home/runner/work/capnproto/capnproto/c++/distcheck-cmake'
make[2]: Leaving directory '/home/runner/work/capnproto/capnproto/c++/distcheck-cmake'
/usr/bin/ld: CMakeFiles/capnp-heavy-tests.dir/test_capnp/capnp/test-import2.capnp.c++.o:(.data.rel.ro+0x20): undefined reference to `capnp::schemas::s_951d576f783c952c'
make[1]: Leaving directory '/home/runner/work/capnproto/capnproto/c++'
/usr/bin/ld: CMakeFiles/capnp-heavy-tests.dir/test_capnp/capnp/test-import2.capnp.c++.o:(.data.rel.ro+0x28): undefined reference to `capnp::schemas::s_aa793a6361fa6e16'
/usr/bin/ld: CMakeFiles/capnp-heavy-tests.dir/test_capnp/capnp/test-import2.capnp.c++.o:(.data.rel.ro+0x30): undefined reference to `capnp::schemas::s_98e7c001808198cd'
/usr/bin/ld: CMakeFiles/capnp-heavy-tests.dir/test_capnp/capnp/test-import2.capnp.c++.o:(.data.rel.ro+0x38): undefined reference to `capnp::schemas::s_f6bd77f100ecb0ff'
```

Vadmeme:
I personally am on windows and i dont think the current bazel build supports it, so i cannot run local testing and all my information is gathered from reading github action logs.

It appears that the problem is on the linker level as the object file already exists, if i had to guess it is because `CAPNP_DECLARE_SCHEMA`  declares the schema values as `extern` so the linker tries to find them in external sources and cannot.

A possible workaround I can think of is instead of fully stripping schema data to emit a zero length array of data.

kentonv:
@mikea Sorry, when I looked at the log I was confused by all of the sign-compare warnings and didn't see the link errors. We should disable that warning (it's disabled in the other builds).

All of the CI builds other than Bazel passed. (Well, and MinGW which has been broken for a while.)

kentonv:
Weirdly the cmake build fails for me locally (although the automake build does not, and obviously neither failed in CI).

kentonv:
I've reverted the change for now.

@Vadmeme we need a version that actually removes the declarations from the header file, I think.

Vadmeme:
I looked into simply removing the `CAPNP_DECLARE_SCHEMA` from the header however this caused alot of problems with the schemas being referenced by other parts of capnp that i feel would be a bad idea to mess with.

An alternative is what i have suggested above of replacing the buffer with schema data with a zero sized placeholder as shown below. I have already implemented this and it seems to work and pass all test cases and should not cause any linker issues.

However before i make another PR id like to make sure this workaround is something you would be open too.

```cpp
static const ::capnp::_::AlignedData<0> b_f6bd77f100ecb0ff = {{}};
::capnp::word const* const bp_f6bd77f100ecb0ff = b_f6bd77f100ecb0ff.words;
#if !CAPNP_LITE
static const ::capnp::_::RawSchema* const d_f6bd77f100ecb0ff[] = {
  &s_a0a8f314b80b63fd,
  &s_bc55b08b672b5d97,
  &s_e682ab4cf923a417,
};
static const uint16_t m_f6bd77f100ecb0ff[] = {1, 2, 0};
static const uint16_t i_f6bd77f100ecb0ff[] = {0, 1, 2};
const ::capnp::_::RawSchema s_f6bd77f100ecb0ff = {
  0xf6bd77f100ecb0ff, b_f6bd77f100ecb0ff.words, 0, d_f6bd77f100ecb0ff, m_f6bd77f100ecb0ff,
  3, 3, i_f6bd77f100ecb0ff, nullptr, nullptr, { &s_f6bd77f100ecb0ff, nullptr, nullptr, 0, 0, nullptr }
};
#endif  // !CAPNP_LITE
```

kentonv:
I don't like the empty buffer approach as it means any attempt to use the schemas (e.g. stringifying a message) would compile just fine and fail at runtime. I'd rather the error be at compile time.

Vadmeme:
The only other option i see is manually editing everything that exposed an `encodedSchema` method in `generated-header-support.h` and adding a "_STRIPPED" variant to that exposes `encodedSchema` that is just a `static_assert(false)` or something similar.

kentonv:
Ideally I think we don't want to declare `encodedSchema()` at all if there's no schema. Does anything break if we omit it entirely?

Maybe what we need to do here is split the macros like `CAPNP_DECLARE_STRUCT_HEADER` into one macro that declares the non-schema-related stuff, and another that declares the schema stuff. When `omitSchemas` is enabled, we only use the former macro.

Vadmeme:
What you suggested would be ideal but i personally do not know enough about the inner workings of the generated headers to implement it yet. 

Nothing SHOULD break if `encodedSchema()` does not exist as its only use is in `schemaProto()` and thats only used  if you use schemas.

PR #1504, opened 2022-07-21
Several small random changes

kentonv:
These were things I needed while working on a project.

### DIFF
--- a/c++/src/kj/async-io-unix.c++
+++ b/c++/src/kj/async-io-unix.c++
@@ -1128,58 +1128,6 @@ private:
   } addr;
 
   struct LookupParams;
-  class LookupReader;
-};
-
-class SocketAddress::LookupReader {
-  // Reads SocketAddresses off of a pipe coming from another thread that is performing
-  // getaddrinfo.
-
-public:
-  LookupReader(kj::Own<Thread>&& thread, kj::Own<AsyncInputStream>&& input,
-               _::NetworkFilter& filter)
-      : thread(kj::mv(thread)), input(kj::mv(input)), filter(filter) {}
-
-  ~LookupReader() {
-    if (thread) thread->detach();
-  }
-
-  Promise<Array<SocketAddress>> read() {
-    return input->tryRead(&current, sizeof(current), sizeof(current)).then(
-        [this](size_t n) -> Promise<Array<SocketAddress>> {
-      if (n < sizeof(current)) {
-        thread = nullptr;
-        // getaddrinfo()'s docs seem to say it will never return an empty list, but let's check
-        // anyway.
-        KJ_REQUIRE(addresses.size() > 0, "DNS lookup returned no permitted addresses.") { break; }
-        return addresses.releaseAsArray();
-      } else {
-        // getaddrinfo() can return multiple copies of the same address for several reasons.
-        // A major one is that we don't give it a socket type (SOCK_STREAM vs. SOCK_DGRAM), so
-        // it may return two copies of the same address, one for each type, unless it explicitly
-        // knows that the service name given is specific to one type.  But we can't tell it a type,
-        // because we don't actually know which one the user wants, and if we specify SOCK_STREAM
-        // while the user specified a UDP service name then they'll get a resolution error which
-        // is lame.  (At least, I think that's how it works.)
-        //
-        // So we instead resort to de-duping results.
-        if (alreadySeen.insert(current).second) {
-          if (current.parseAllowedBy(filter)) {
-            addresses.add(current);
-          }
-        }
-        return read();
-      }
-    });
-  }
-
-private:
-  kj::Own<Thread> thread;
-  kj::Own<AsyncInputStream> input;
-  _::NetworkFilter& filter;
-  SocketAddress current;
-  kj::Vector<SocketAddress> addresses;
-  std::set<SocketAddress> alreadySeen;
 };
 
 struct SocketAddress::LookupParams {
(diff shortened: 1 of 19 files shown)


kentonv:
Ugh the sendfile optimization has a bug that should be caught in the test but it turns out that on all the 64-bit systems sendfile() is just always returning EINVAL and we're falling back to regular copying? WTF? Only ManyLinux 32-bit actually caught the problem.

kentonv:
sendfile() fixup for posterity: 151f354cebc060cb04b91288c4664d52521fbbff

c++/src/kj/async-io-unix.c++
```diff
@@ -228,17 +239,77 @@ public:
...
+      KJ_SYSCALL_HANDLE_ERRORS(n = sendfile(fd, fileFd, &offset, requested)) {
+        case EINVAL:
+        case ENOSYS:
+          // Fall back to regular pump
+          return unoptimizedPumpTo(input, *this, amount, soFar);
+
+        case EAGAIN:
+          return observer.whenBecomesWritable()
+              .then([this, &input, fileFd, amount, soFar]() {
+            return pumpFromFile(input, fileFd, amount, soFar);
```
jasnell, line 286: As absolutely unlikely as it will be to get `EAGAIN` multiple times in a row, the possibility is still non-zero. Something to guard against too much recursion might be good. But also happy like this.

c++/src/kj/async-io.c++
```diff
@@ -2768,6 +2769,36 @@ String CapabilityStreamNetworkAddress::toString() {
   return kj::str("<CapabilityStreamNetworkAddress>");
 }
 
+Promise<size_t> FileInputStream::tryRead(void* buffer, size_t minBytes, size_t maxBytes) {
+  size_t result = file.read(offset, arrayPtr(reinterpret_cast<byte*>(buffer), maxBytes));
```
jasnell, line 2773: Any intention here of paying attention to the `minBytes` contract here? If not, a comment here might be helpful. If yes, then a TODO comment here would be good.

c++/src/kj/compat/tls.c++
```diff
@@ -100,6 +100,37 @@ inline void ensureOpenSslInitialized() {
...
+  for (auto c: addr) {
+    if (c == ':') {
+      isPossiblyIp4 = false;
+      ++colonCount;
+    } else if (c == '.') {
+      isPossiblyIp6 = false;
+      ++dotCount;
+    } else if ('0' <= c && c <= '9') {
+      // Digit is valid for ipv4 or ipv6.
+    } else if ('a' <= c && c <= 'f') {
```
jasnell, line 117: Ignoring upper case hex digits?

c++/src/kj/compat/tls.c++
```diff
@@ -100,6 +100,37 @@ inline void ensureOpenSslInitialized() {
...
+      // Nope.
+      return false;
+    }
+  }
+
+  // An IPv4 address has 3 dots. (Yes, I'm aware that technically IPv4 addresses can be formatted
+  // with fewer dots, but it's not clear that we actually want to support TLS authentication of
+  // non-canonical address formats, so for now I'm not. File a bug if you care.) An IPv6 address
+  // has at least 2 and as many as 7 colons.
+  return (isPossiblyIp4 && dotCount == 3)
```
jasnell, line 130: I've been considering the possibility of moving the whatwg url parsing algorithm into kj as part of kj::Url... in which case we can likely use it's ip parsing algorithms to support those other formats if we wanted later. I agree for now that this is more than sufficient.

c++/src/kj/async-io-unix.c++
kentonv: From the man page:

`EAGAIN Nonblocking I/O has been selected using O_NONBLOCK and the write would block.`

`EAGAIN` here very specifically means that the socket is not writable because the send buffer is full. So we wait until it becomes writable, and try again. Since we're using edge triggering, it is technically possible that we could receive a queued writability edge notification that is no longer accurate, but we would not receive a never-ending stream of these, barring some sort of [deep kernel bug](https://sandstorm.io/news/2015-04-08-osx-security-bug).

Note also that this is promise recursion, and it's a tail call, so this does not result in a growing chain.

jasnell: Yep, I'm just in an overly paranoid mood at the moment ;-) ... as I said, happy with this as is.

jasnell (approved):
Couple of minor comments, otherwise LGTM

c++/src/kj/async-io.c++
kentonv: This implementation correctly handles `minBytes`. The requirement is that we only return fewer than `minBytes` on EOF, and a file read will only return fewer bytes than requested on EOF.

kentonv: (I'll add a comment.)

c++/src/kj/compat/tls.c++
kentonv: Good catch.

c++/src/kj/compat/tls.c++
kentonv: I am not sure I want the whatwg algorithm in `kj::Url`. I prefer that `kj::Url` decodes %-escapes correctly. :)

Also note that `kj::Network::parseAddress()` actually does parse any IP address format correctly, using `inet_pton()`, but as the comment says it's not clear to me we actually want to support weird formats here.

c++/src/kj/compat/tls.c++
kentonv: fb616dec535b49055ba35f11142e112a44cbbb5f

c++/src/kj/async-io.c++
kentonv: 5cd52440cc39cb061a525f382d02dc4b53c6dcad

c++/src/kj/compat/tls.c++
jasnell: Fair enough!

c++/src/kj/async-io.c++
jasnell: Awesome ok.

jasnell (approved):

PR #1505, opened 2022-07-22
Websocket Compression (Phase Two)

MellowYarker:
This PR implements sending and receiving compressed messages. All `permessage-deflate` parameters are supported.

I've written an `ew-test-bin` for testing and it seems like things are working. I compared my compressed results with a few examples in the compression RFC and things matched up so that's a good sign. That said, I want to add to our existing KJ tests, though I suspect it might require modifying the fake server a bit.

There's still a couple `KJ_DBG`s sprinkled around for debugging reasons. The `Context` class is based on a similar class used by our `CompressionStream` implementation.


MellowYarker:
Added a test that attempts to use:
- 2 deflate blocks in a message
- a deflate block without compression
- a deflate block with `BFINAL` set to 1

Our implementation does not send these types of messages, so these examples determine whether our implementation can understand payloads sent from other implementations.

I'm still not sure what's happening with failing builds, it looks like `zlib.h` is found, but then fails to link?

c++/src/kj/compat/http.c++
```diff
@@ -2594,12 +2663,167 @@ private:
...
+      // Since Zlib buffers the writes, we want to continue processing until there's nothing left.
+      Result result;
+      KJ_IF_MAYBE(exception, kj::runCatchingExceptions([this, &result]() {
+        result = pumpOnce();
+      })) {
+        return kj::mv(*exception);
+      }
+
+      if (result.buffer.size() > 0) {
+        output.addAll(result.buffer);
```
kentonv, line 2791: We will need to bail out when the total number of decompressed bytes exceeds the message size limit. Otherwise, it's possible to construct a malicious message whose compressed size is very small, but which uncompresses to a huge amount of data, consuming all available memory.

c++/src/kj/compat/http.c++
```diff
@@ -2594,12 +2663,167 @@ private:
...
+      }
+
+      if (ctx.avail_in == 0 || result.processResult == Z_STREAM_END) {
```
kentonv, line 2794: `avail_in` being zero doesn't necessarily mean we're done. It may be that zlib has consumed all of the input data into its internal state, but still has a few output bytes that didn't fit in the last buffer. We have to keep giving it new output buffers to write to until it actually signals to us that it has finished.

c++/src/kj/compat/http.c++
```diff
@@ -2594,12 +2663,167 @@ private:
...
     static constexpr byte FIN_MASK = 0x80;
     static constexpr byte RSV_MASK = 0x70;
+    static constexpr byte RSV1_MASK = 0x40;
     static constexpr byte OPCODE_MASK = 0x0f;
 
     static constexpr byte USE_MASK_MASK = 0x80;
     static constexpr byte PAYLOAD_LEN_MASK = 0x7f;
   };
 
+  class Context {
```
kentonv, line 2673: Could this class be simplified by using `kj::GzipOutputStream` on top of `kj::VectorOutputStream` or the like? Especially perhaps if #1428 got finished and merged, which avoids the need for `GzipOutputStream` to hold its own buffer (which is something you're currently doing anyway in this class, but which I commented we might want to avoid...).

c++/src/kj/compat/http.c++
```diff
@@ -2667,6 +2894,34 @@ private:
...
+        KJ_SWITCH_ONEOF(compressor.processMessage(message)) {
+          KJ_CASE_ONEOF(e, kj::Exception) {
+            // TODO(now): What should happen here? We failed to compress...
+            KJ_DBG(e);
+          }
+          KJ_CASE_ONEOF(compressed, kj::Vector<kj::byte>) {
+            KJ_DBG("[SEND] Compressed size", compressed.size());
+            auto& innerMessage = compressedMessage.emplace(compressed.releaseAsArray());
+            message = innerMessage.slice(0, innerMessage.size() - 4);
+            // Strip 0x00 0x00 0xFF 0xFF off the tail.
```
kentonv, line 2919: Should we assert that the buffer actually ends with that?

Once #1504 is in, you should even be able to write this as:

`KJ_ASSERT(innerMessage.endsWith({0x00, 0x00, 0xFF, 0xFF}));`

c++/src/kj/compat/http.c++
```diff
@@ -2667,6 +2894,34 @@ private:
 
     Mask mask(maskKeyGenerator);
 
+    bool useCompression = false;
+    kj::Maybe<kj::Array<byte>> compressedMessage;
```
kentonv, line 2898: You need to make sure this gets `.attach()`ed to the write promise below (just like `ownMessage`), otherwise if the message doesn't get fully written to the socket buffer on the first syscall, it'll be used after free on subrequest writes.

kentonv (commented):
It occurs to me that we're going to want to let the application selectively opt into / out of compression when calling `HttpService::Response::acceptWebSocket()`.

Specifically, when we're proxying a WebSocket through Workers, we probably want to make sure the eyeball -> worker connection uses exactly the same parameters as the worker -> origin connection, so that we are able to pump the bytes through.

Speaking of which, you'll need to update `WebSocketImpl::tryPumpFrom` to make sure it doesn't allow optimized pumping if the compression parameters differ.

c++/src/kj/compat/http.c++
```diff
@@ -2303,8 +2311,14 @@ public:
     kj::Array<byte> message;           // space to allocate
     byte* payloadTarget;               // location into which to read payload (size is payloadLen)
     if (isFin) {
-      // Add space for NUL terminator when allocating text message.
-      size_t amountToAllocate = payloadLen + (opcode == OPCODE_TEXT && isFin);
+      size_t amountToAllocate;
+      if (recvHeader.isCompressed()) {
+        // Add 4 since we append 0x00 0x00 0xFF 0xFF.
```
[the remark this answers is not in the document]
MellowYarker: https://github.com/capnproto/capnproto/pull/1505/commits/8a9cc53f2a178a1ea6e07b229341f3b761b67612

c++/src/kj/compat/http.c++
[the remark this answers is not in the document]
MellowYarker: Went with `ZlibContext` since `Compression` has been thrown around a lot already: https://github.com/capnproto/capnproto/pull/1505/commits/408d58a67849795a076ef5653bb89d10e37f0822

c++/src/kj/compat/http.c++
```diff
@@ -2594,12 +2663,167 @@ private:
...
+    };
+
+    struct Result {
+      int processResult = 0;
+      kj::ArrayPtr<const byte> buffer;
+    };
+
+    Context(Mode mode, const CompressionParameters& config) : mode(mode) {
+      int result = Z_OK;
+      int windowBits;
```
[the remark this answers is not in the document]
MellowYarker: Thanks! https://github.com/capnproto/capnproto/pull/1505/commits/f1d41fa531442900f2c0f99ab544172452da2a34

c++/src/kj/compat/http.c++
```diff
@@ -2594,12 +2663,167 @@ private:
...
+              windowBits,
+              8,  // memLevel = 8 is the default
+              Z_DEFAULT_STRATEGY);
+          break;
+        case Mode::DECOMPRESS:
+          windowBits = -config.inboundMaxWindowBits.orDefault(15);
+          // We use negative values because we want to use raw inflate.
+          result = inflateInit2(&ctx, windowBits);
+          break;
+        default:
```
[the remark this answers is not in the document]
MellowYarker: Thanks for pointing this out, especially the Cap'n Proto part. I've removed it from two other places as well.
https://github.com/capnproto/capnproto/pull/1505/commits/5d269c59dc1939527618cfb5b3f446f51363cc89
https://github.com/capnproto/capnproto/pull/1505/commits/8aba4f4ffcc4fd1a6e4d3837509aa5ad7b5ee5c7

MellowYarker:
Just changed the return type to `void` and assert that we get the correct return value: https://github.com/capnproto/capnproto/pull/1505/commits/ee92d4420c2931a029d409c9f4c4fe9e087c49d9

c++/src/kj/compat/http.c++
```diff
@@ -2594,12 +2663,167 @@ private:
...
+    static constexpr byte RSV1_MASK = 0x40;
     static constexpr byte OPCODE_MASK = 0x0f;
 
     static constexpr byte USE_MASK_MASK = 0x80;
     static constexpr byte PAYLOAD_LEN_MASK = 0x7f;
   };
 
+  class Context {
+  // `Context` is the WebSocket's interface to Zlib's compression/decompression functions.
+  // Depending on the `mode`, `Context` will act as a compressor or a decompressor.
```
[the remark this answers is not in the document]
MellowYarker: https://github.com/capnproto/capnproto/commit/408d58a67849795a076ef5653bb89d10e37f0822#diff-09d28b46bd98866f20dd7241e02c70338a76fdb59a483ed02d4c16bbe6bcc3a1R2677-R2678

c++/src/kj/compat/http.c++
```diff
@@ -2594,12 +2663,167 @@ private:
...
+          break;
+        case Mode::DECOMPRESS:
+          inflateEnd(&ctx);
+          break;
+      }
+    }
+
+    KJ_DISALLOW_COPY(Context);
+
+    void setInput(const void* in, size_t size) {
```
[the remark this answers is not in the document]
MellowYarker: https://github.com/capnproto/capnproto/pull/1505/commits/7186db5714156866a075018cc83d55de9b34e8d6

c++/src/kj/compat/http.c++
```diff
@@ -2594,12 +2663,167 @@ private:
...
+      ctx.next_in = const_cast<byte*>(reinterpret_cast<const byte*>(in));
+      ctx.avail_in = size;
+    }
+
+    Result pumpOnce() {
```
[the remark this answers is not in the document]
MellowYarker: Moved code: https://github.com/capnproto/capnproto/pull/1505/commits/1ce40efbb88401a15d828fafb2ac37e96848c875
Added comment: https://github.com/capnproto/capnproto/pull/1505/commits/a70f2bdfe94adb1f016422e989b86005e11b6507

c++/src/kj/compat/http.c++
```diff
@@ -2594,12 +2663,167 @@ private:
...
+      setInput(message.begin(), message.size());
+
+      kj::Vector<kj::byte> result;
+      KJ_IF_MAYBE(e, processLoop(result)) {
+        return kj::mv(*e);
+      }
+      return kj::mv(result);
+    }
+
+    kj::Maybe<kj::Exception> processLoop(kj::Vector<kj::byte>& output) {
```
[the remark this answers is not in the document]
MellowYarker: https://github.com/capnproto/capnproto/pull/1505/commits/02ba81c2c5e30763d2feb1deb5fdb3989a22eb1e

c++/src/kj/compat/http.c++
```diff
@@ -2359,9 +2373,58 @@ public:
...
+            tail[2] = 255;
+            tail[3] = 255;
+            // We have to append 0x00 0x00 0xFF 0xFF to the message before inflating.
+            if (config->inboundNoContextTakeover) {
+              // We must reset context on each message.
+              decompressor.reset();
+            }
+            KJ_SWITCH_ONEOF(decompressor.processMessage(message)) {
+              KJ_CASE_ONEOF(e, kj::Exception) {
+                // TODO(now): What should happen here? We failed to decompress...
```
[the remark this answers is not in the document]
MellowYarker: https://github.com/capnproto/capnproto/pull/1505/commits/e620fbdd9e3220879bfa76c73a0724d4b57d5f4a

c++/src/kj/compat/http.c++
```diff
@@ -2594,12 +2663,167 @@ private:
...
+      KJ_IF_MAYBE(e, processLoop(result)) {
+        return kj::mv(*e);
+      }
+      return kj::mv(result);
+    }
+
+    kj::Maybe<kj::Exception> processLoop(kj::Vector<kj::byte>& output) {
+      // Since Zlib buffers the writes, we want to continue processing until there's nothing left.
+      Result result;
+      KJ_IF_MAYBE(exception, kj::runCatchingExceptions([this, &result]() {
```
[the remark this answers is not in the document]
MellowYarker: https://github.com/capnproto/capnproto/commit/e620fbdd9e3220879bfa76c73a0724d4b57d5f4a

c++/src/kj/compat/http.c++
```diff
@@ -2594,12 +2663,167 @@ private:
...
+
+      if (ctx.avail_in == 0 || result.processResult == Z_STREAM_END) {
+        // We're finished processing.
+        if (!(result.processResult == Z_OK || result.processResult == Z_STREAM_END)) {
+          return KJ_EXCEPTION(FAILED, "Failed to process message.");
+        }
+        return nullptr;
+      }
+
+      return processLoop(output);
```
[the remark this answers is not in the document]
MellowYarker: Yeah I was wondering about this, the `CompressionStream` code returned a promise but it felt weird to keep the recursion after that was removed. Thanks for clarifying: https://github.com/capnproto/capnproto/pull/1505/commits/57744f736869c9041df1ccf86b6bd6f42cec7d1c

c++/src/kj/compat/http.c++
MellowYarker: I tried to rely on `Z_STREAM_END` only but we would end up in a scenario where we'd get `Z_BUF_ERROR` because the input was empty. Resolved it by checking for `Z_BUF_ERROR` showing up along with an empty output buffer. https://github.com/capnproto/capnproto/pull/1505/commits/6b4ff9d4b80d4af8a3cea510229e42fa260c2699

c++/src/kj/compat/http.c++
```diff
@@ -2594,12 +2663,167 @@ private:
...
+          KJ_REQUIRE(result == Z_OK || result == Z_BUF_ERROR || result == Z_STREAM_END,
+                      "Decompression failed.");
+          break;
+        default:
+          KJ_UNREACHABLE;
+      }
+
+      return Result {
+        .processResult = result,
+        .buffer = kj::arrayPtr(buffer, sizeof(buffer) - ctx.avail_out),
```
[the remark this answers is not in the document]
MellowYarker: https://github.com/capnproto/capnproto/pull/1505/commits/89889eae6c2d5251bdddd0b846e09702284448b9

c++/src/kj/compat/http.c++
MellowYarker: https://github.com/capnproto/capnproto/pull/1505/commits/bff88eec0bc71fb2270d5b632d4ce2c21527a8cf

c++/src/kj/compat/http.c++
MellowYarker: https://github.com/capnproto/capnproto/pull/1505/commits/2188f68f065b2eceaf2b52ff748056fa4e72d3fd

MellowYarker: Hmm, on second thought this looks wrong. If the client splits the message into `n` frames, then each time we add a fragment and call `receive()` we're passing a smaller `maxSize` until we get the final frame. At this point, we start decompressing the message, but `maxSize` isn't the original `maxSize`, it just tells us how large the remaining payload can be :thinking:. Will need to rethink this.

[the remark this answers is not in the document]
MellowYarker: Went with the `Vector<Array<byte>>` approach: https://github.com/capnproto/capnproto/pull/1505/commits/7bdcb02a85b396b7edfdc88ceffd24ad9fd5ab41.

c++/src/kj/compat/http.c++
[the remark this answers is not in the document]
MellowYarker: Added a bit more to the comment after modifying the condition under which we are done processing: https://github.com/capnproto/capnproto/pull/1505/commits/ce1efdc72caf230834c44cfd949576cd1dddf583

c++/src/kj/compat/http.c++
kentonv: Hmmm, that doesn't really sound right. Actually, I found this in `zlib.h`:

```
    If the parameter flush is set to Z_SYNC_FLUSH, all pending output is
  flushed to the output buffer and the output is aligned on a byte boundary, so
  that the decompressor can get all input data available so far.  (In
  particular avail_in is zero after the call if enough output space has been
  provided before the call.) Flushing may degrade compression for some
  compression algorithms and so it should be used only when necessary.  This
  completes the current deflate block and follows it with an empty stored block
  that is three bits plus filler bits to the next byte, followed by four bytes
  (00 00 ff ff).
```

This seems to say that checking `avail_in == 0` is actually sufficient. So maybe you were right initially. Sorry for the false alarm.

c++/src/kj/compat/http.c++
```diff
@@ -2594,12 +2653,205 @@ private:
...
+    }
+
+    kj::Vector<Result> processLoop(kj::Maybe<size_t> maxSize) {
+      // Since Zlib buffers the writes, we want to continue processing until there's nothing left.
+      kj::Vector<Result> output;
+      size_t totalBytesProcessed = 0;
+      for (;;) {
+        Result result = pumpOnce();
+
+        size_t status = result.processResult;
```
MellowYarker, line 2829: This should be an `int` (or `auto`).

c++/src/kj/compat/http.c++
```diff
@@ -2594,12 +2653,205 @@ private:
...
+        memcpy(&processedMessage[currentIndex], part.buffer.begin(), part.size);
+        // We need to use `part.size` to determine the number of useful bytes, since data after
+        // `part.size` is unused (and probably junk).
+        currentIndex += part.size;
+      }
+
+      if (addNullTerminator) {
+        processedMessage.back() = '\0';
+      }
+
```
kentonv, line 2763: At the end of code like this I like to include an assertion that we didn't overrun or underrun the buffer:

`KJ_ASSERT(currentIndex == processedMessage.size());`

c++/src/kj/compat/http.c++
```diff
@@ -2594,12 +2653,205 @@ private:
...
+      return kj::mv(processedMessage);
+    }
+
+    void reset() {
+      // Resets the (de)compression context. This should only be called when the (de)compressor uses
+      // client/server_no_context_takeover.
+      switch (mode) {
+        case Mode::COMPRESS: {
+          int result = deflateReset(&ctx);
+          KJ_ASSERT(result == Z_OK, "Deflate reset failed with ", result);
```
kentonv, line 2773: Remember that the assert macros do not just concatenate the parameters. This one will end up producing an error like:

`failed: expected result == Z_OK [-2 == 0]; Deflate reset failed with ; result = -2`

I would probably change the string part of the message to `deflateReset() failed`.

c++/src/kj/compat/http.c++
MellowYarker: I think this needs to be investigated a bit more. Earlier today I noticed [chromium](https://github.com/chromium/chromium/blob/c676681f0040f5b310f8949bd44d332b472a712d/net/websockets/websocket_deflater.cc#L73-L113) uses both `Z_NO_FLUSH` and `Z_SYNC_FLUSH`. I don't know for certain _yet_ [1], but I suspect that flushing to the output buffer when using `Z_NO_FLUSH` means zlib will not end the deflate block each time the 4KB buffer is full (whereas `Z_SYNC_FLUSH` would).

**If so**, I figure the drawback of using `Z_SYNC_FLUSH` is that for messages that are still large _after_ compression, we will have multiple deflate blocks (including empty blocks after each), whereas for `Z_NO_FLUSH`, the entire compressed message will fit in a single deflate block. Assuming a 1MB message limit (taking this from `receive()` in `http.h`) and a buffer of 4096 bytes, the largest uncompressed message would fit in about ~244 buffers. However, the compressed message would need to be smaller, so the upper bound on number of buffers (deflate blocks) is probably 200-220? That's likely less than 1KB of extra data going over the wire (worst case) per message.

Otherwise, it's not immediately obvious to me why we would want to use `Z_NO_FLUSH` in our case since:
1. We only start compressing/decompressing once the entire input (message) is available. If a message were to only _partially_ fill up the `output` buffer, it's fine, since we don't expect any more input (until we work on the next message, which is separate).
2. `Z_NO_FLUSH` seems to force a flush if the `output` buffer would be totally filled (`avail_out == 0`), so
i. In the case that we have a large message, we're forcing a flush either way.
ii. In the case that the message fits in the buffer, we end up calling `Z_SYNC_FLUSH` either way.

Anything obvious I'm missing here? I've messed around a bit with these two flush types a bit, but I still need to try sending some _really_ large messages to see if there's any significant difference.

[1]: `zlib.h` hasn't really explained flushing vs. finishing blocks, etc. I'll look again, but I think it might be easier to just check actual program output.

kentonv: Chromium uses both `Z_NO_FLUSH` and `Z_SYNC_FLUSH` because it doesn't necessarily write the whose message at once. Presumably `AddBytes()` can be called multiple times to add more data. If they flushed after every call to `AddBytes()`, then the compressor would be unable to take advantage of redundancies that cross a buffer boundary, because it is forced to completely process each buffer before it knows what comes in the next buffer.

In our case, though, we always present the _entire_ message to zlib at one time. So we can always pass `Z_SYNC_FLUSH`.

Note that if zlib stops because it doesn't have enough space in the output buffer, then it doesn't matter what flush mode you specified. The flush only actually applies if zlib was able to consume all input and write out everything it wanted to output.

To put it another way:
* You should use `Z_NO_FLUSH` to tell zlib: "This is not the whole message. There will be another buffer of input after this one. So you don't have to write out everything necessary to decode it yet since there's more data coming anyway."
* You use `Z_SYNC_FLUSH` to say "This is the last (or only) buffer in the message. Please make sure that you output all bytes needed so that the other end will be able to full decompress the message."

c++/src/kj/compat/http.c++
MellowYarker: https://github.com/capnproto/capnproto/pull/1505/commits/6a0e04cc8a99acec3fc99af92f8bfd9988e577cf

c++/src/kj/compat/http.c++
MellowYarker: Thanks for clarifying! https://github.com/capnproto/capnproto/pull/1505/commits/6a0e04cc8a99acec3fc99af92f8bfd9988e577cf

c++/src/kj/compat/http.c++
MellowYarker: https://github.com/capnproto/capnproto/pull/1505/commits/b2d1fada2df91e6d925096b79a4df2b627867bdc

c++/src/kj/compat/http.c++
MellowYarker: Will keep that in mind, sounds like an easy way to prevent headaches: https://github.com/capnproto/capnproto/pull/1505/commits/e6972f939854ae79daa8869ff63a1e91f48b469d

c++/src/kj/compat/http.c++
MellowYarker: https://github.com/capnproto/capnproto/pull/1505/commits/f3d82e6c7227a4966032bf7af07274d0aa1aee58

c++/src/kj/compat/http.c++
```diff
@@ -2359,9 +2374,47 @@ public:
           // Shouldn't get here; handled above.
           KJ_UNREACHABLE;
         case OPCODE_TEXT:
+          KJ_IF_MAYBE(config, compressionConfig) {
+            KJ_DBG("[RECV TEXT] Compressed size", message.size());
+            auto& decompressor = KJ_ASSERT_NONNULL(decompressionContext);
+            auto tail = message.slice(message.size() - 4, message.size());
```
jasnell, line 2380: Might be worth a short comment here that the extra 4 bytes were added above in the `if (recvHeader.isCompressed()) {` check.

c++/src/kj/compat/http.c++
```diff
@@ -2703,12 +2762,207 @@ private:
...
+          // It's the maxSize of this current `receieve()` call (consider multi-frame messages).
+          KJ_IF_MAYBE(maxMsgSize, maxSize) {
+            // This is only non-null for `receive` calls, so we must be decompressing. We don't want
+            // the decompressed message to OOM us, so let's make sure it's not too big.
+            KJ_REQUIRE(totalBytesProcessed < *maxMsgSize,
+                "Decompressed WebSocket message is too large");
+          }
+        }
+
+        if ((ctx.avail_in == 0 && ctx.avail_out != 0) || status == Z_STREAM_END) {
```
kentonv, line 2953: My reading of zlib.h is that you don't need `&& ctx.avail_out != 0` here -- it says that when using `Z_SYNC_FLUSH`, zlib will only allow `ctx.avail_in` to be zero if it actually managed to write everything, otherwise it will keep `avail_in` non-zero. Does that seem right?

I think `|| status == Z_STREAM_END` may also be redundant, as my reading of zlib.h is that `Z_STREAM_END` will only be returned if `Z_FINISH` was requested, which we never do.

c++/src/kj/compat/http.c++
MellowYarker: https://github.com/capnproto/capnproto/pull/1505/commits/96a1e0d46eeef2f33abfa788011e5707adbfc361

c++/src/kj/compat/http.c++
MellowYarker: This first one was more so just defensive coding, I feel like that's probably the case, but then I saw:
```
  If deflate returns with avail_out == 0, this function must be called again
  with the same value of the flush parameter and more output space (updated
  avail_out), until the flush is complete (deflate returns with non-zero
  avail_out).  In the case of a Z_FULL_FLUSH or Z_SYNC_FLUSH, make sure that
  avail_out is greater than six to avoid repeated flush markers due to
  avail_out == 0 on return.
```
A (maybe weird) case that came to my mind was if we could flush the entire compressed input into 4096 bytes, but then not have space for the `0x00 0x00 0xFF 0xFF`. Would zlib set `ctx.avail_in` to 4, or 0? (I suspect it's _probably_ going to keep it at 4, but since I wasn't totally sure, I put the extra check to adhere to the comment above). I'll re-read the header file to try to convince myself of one or the other.

As for the second thing, if we exclude `|| status == Z_STREAM_END`, then [this test](https://github.com/capnproto/capnproto/pull/1505/files#diff-bad743492083c5cd840b54442f4c32da3375e01f43c2fa021c927a4588f5a47eR3030) will hang (we get `Z_STREAM_END` but never terminate the infinite loop). Apparently, implementations can send payloads [that set](https://datatracker.ietf.org/doc/html/rfc7692#section-7.2.3.4) the `BFINAL` bit of the deflate block header, which seems to override the flush type.

c++/src/kj/compat/http.c++
MellowYarker: I think this is a fairly non-invasive way to get back the original maximum size: https://github.com/capnproto/capnproto/pull/1505/commits/3eaf8d6672c45ab1ea11b72edbaa50c8f7141724

MellowYarker:
> It occurs to me that we're going to want to let the application selectively opt into / out of compression when calling HttpService::Response::acceptWebSocket().
>
> Specifically, when we're proxying a WebSocket through Workers, we probably want to make sure the eyeball -> worker connection uses exactly the same parameters as the worker -> origin connection, so that we are able to pump the bytes through.

I'm a bit confused by  _selectively opt in/out_. Does this only have consequences for when a direct pump cannot be achieved, or are there other situations as well? In other words, does preventing direct pump when configs differ (as is done below) achieve this?
> Speaking of which, you'll need to update WebSocketImpl::tryPumpFrom to make sure it doesn't allow optimized pumping if the compression parameters differ.
- https://github.com/capnproto/capnproto/pull/1505/commits/5b628aa6144e633fa8c27ffeab8bf70898650e6c
- Added a second fixup to silence build failures: https://github.com/capnproto/capnproto/pull/1505/commits/a6ea0a4b750dc585058255ff47a0b3b319bb8b8b

c++/src/kj/compat/http.c++
```diff
@@ -2359,9 +2373,58 @@ public:
           // Shouldn't get here; handled above.
           KJ_UNREACHABLE;
         case OPCODE_TEXT:
+          KJ_IF_MAYBE(config, compressionConfig) {
+            KJ_DBG("[RECV TEXT] Compressed size", message.size());
```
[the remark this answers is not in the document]
MellowYarker: https://github.com/capnproto/capnproto/pull/1505/commits/2e625934efb94ed25264bfb0688f9980547a8cd6

MellowYarker:
Brief summary on build failures and what I think is happening:
- linux-musl looks like it doesn't install `zlib` while building, so I figure we can fix that by including it [here](https://github.com/capnproto/capnproto/blob/52028fe3e0dd83e82857069bce48ea890f5aad4c/.github/workflows/quick-test.yml#L19)?
- MacOS finds `zlib.h` but we've got undefined symbols so we must be failing to link. Should we add [this](https://github.com/capnproto/capnproto/blob/e8946760a94e9720859979877d60eb73d882c984/c%2B%2B/src/kj/CMakeLists.txt#L218-L221) to [kj-http](https://github.com/capnproto/capnproto/blob/e8946760a94e9720859979877d60eb73d882c984/c%2B%2B/src/kj/CMakeLists.txt#L175)?
- Both MSVC's find zlib (as part of Miniconda?), but we get: **fatal error C1083: Cannot open include file: `zlib.h`: no such file or directory** in `http.c++` when we try to include. I guess it's just having trouble finding the library, so probably linking like I mentioned above for MacOS would fix this?

More generally speaking, I see that `gzip.c++` is contained within a preprecessor condition `#if KJ_HAS_ZLIB`. I'm wondering if all the websocket compression code that references `zlib` names should also be wrapped in that?

c++/src/kj/compat/http.c++
kentonv: Ugh, that comment about `avail_out == 0` seems to directly contradict the earlier comment saying "(In particular avail_in is zero after the call if enough output space has been provided before the call.)" Now I don't know what to believe. Maybe zlib guarantees that `avail_in` and `avail_out` are never both left as zero? Need tests for this, but it's tricky since you need to come up with test input that leads to the compressed message ending exactly at a buffer boundary, as well as another test input where it ends one byte past a buffer boundary.

Regarding `Z_STREAM_END`, I guess I was thinking specifically on the sending (compressing) side, you will never see it. On the receiving (decompressing) side, I suppose we need to handle it, but it sounds like it should be considered an error? Can we assert-fail?

c++/src/kj/compat/http.c++
```diff
@@ -2411,15 +2419,24 @@ public:
 
     kj::Array<byte> message;           // space to allocate
     byte* payloadTarget;               // location into which to read payload (size is payloadLen)
+    kj::Maybe<size_t> maxMessageSize;  // maximum message size (maxSize from first `receive()` call)
```
kentonv, line 2422: Bikeshed: `originalMaxSize` would make more sense to me.

c++/src/kj/compat/http.c++
```diff
@@ -2703,12 +2783,205 @@ private:
...
+      size_t totalBytesProcessed = 0;
+      for (;;) {
+        Result result = pumpOnce();
+
+        auto status = result.processResult;
+        auto bytesProcessed = result.size;
+        if (bytesProcessed > 0) {
+          output.add(kj::mv(result));
+          totalBytesProcessed += bytesProcessed;
+          KJ_IF_MAYBE(maxMsgSize, maxSize) {
```
kentonv, line 2964: In this `KJ_IF_MAYBE`, the unwrapped name `maxMsgSize` is a variation of the wrapped name `maxSize` that doesn't really express what the difference is in a meaningful way.

I haven't come up with a great answer for what to do in these situations, but I don't like these random variations. Instead, I've mostly settled on using a single-letter name for the unwrapped version. So here I'd do `KJ_IF_MAYBE(m, maxSize)`. It's not beautiful but at least by being a consistent convention across the codebase, it's easy for readers to understand what they're reading.

c++/src/kj/compat/http.c++
```diff
@@ -2637,6 +2713,10 @@ private:
       return bytes[0] & FIN_MASK;
     }
 
+    bool isCompressed() const {
+      return bytes[0] & RSV1_MASK;
+    }
+
     bool hasRsv() const {
```
kentonv, line 2720: Hmm, it occurs to me that I probably defined this `hasRsv()` so that I could verify that it was always zero but I never actually wrote that check. Maybe we should changes this to `hasRsv2or3()` and actually check it somewhere...

kentonv:
>> It occurs to me that we're going to want to let the application selectively opt into / out of compression when calling HttpService::Response::acceptWebSocket().
>>
>> Specifically, when we're proxying a WebSocket through Workers, we probably want to make sure the eyeball -> worker connection uses exactly the same parameters as the worker -> origin connection, so that we are able to pump the bytes through.
>
> I'm a bit confused by  _selectively opt in/out_. Does this only have consequences for when a direct pump cannot be achieved, or are there other situations as well? In other words, does preventing direct pump when configs differ (as is done below) achieve this?

What I meant here is: Imagine the eyeball requests compression. The worker forwards the request to the origin, including the proposal for compression. The origin, however, doesn't implement compression, and does not return an extensions header. Now the worker calls `acceptWebSocket()` on the eyeball request, giving the headers that the origin sent, which has no reply to the extension request.

KJ, at this point, might kick in and say "Oh, I know how to implement compression!" and will decide to prepare its own extension response saying this, thus establishing a compressed WebSocket with the eyeball. However, by doing so, now we cannot optimize the pump to origin, because we have to compress/decompress every frame. We want to avoid this.

So, I think we need to make sure that, at least as used in Workers, KJ respects exactly the response headers that the application gives it. If the application does not respond to the extensions request, then KJ should not use compression.

But that behavior is only what we want in Workers. It is _not_ what we'd want in a regular KJ-based application server, where implementing compression is probably desirable.

So I think our HttpServerOptions needs three modes: No compression, automatic compression, and manual compression. In manual compression mode, KJ never completes the extension handshake itself, it only passes exactly the headers the application gives it and implements that mode (or fails if the application specifies a mode that KJ can't handle).

kentonv:
> More generally speaking, I see that gzip.c++ is contained within a preprecessor condition #if KJ_HAS_ZLIB. I'm wondering if all the websocket compression code that references zlib names should also be wrapped in that?

I think so, yes.

MellowYarker:
Rebase + force push, no code changes.

c++/src/kj/compat/http.c++
MellowYarker: https://github.com/capnproto/capnproto/pull/1505/commits/d1fddc92daf46d4ee4813eb50005597f7fbb53d3

c++/src/kj/compat/http.c++
MellowYarker: https://github.com/capnproto/capnproto/pull/1505/commits/c9f2965ffc7a2db2d0989f45520a4f17df25003f

c++/src/kj/compat/http.c++
MellowYarker: We `KJ_REQUIRE` that they're 0 in `receive`, and `KJ_ASSERT` that they're 0 in `send`: https://github.com/capnproto/capnproto/pull/1505/commits/0ca87c76dfb6f70c8b2455998e1302b6f62392ea.

MellowYarker:
These fixups
- https://github.com/capnproto/capnproto/pull/1505/commits/08a684372c2b81bc61ac072a1480909c198785c4
- https://github.com/capnproto/capnproto/pull/1505/commits/15096ece50753033b33d77045a13baff5d937915 

will intentionally break the linux-bazel and MacOS builds respectively. I want to be sure they break _before_  pushing preprocessor changes so we know whether all the zlib stuff is skipped for these builds.

c++/src/kj/compat/http.c++
```diff
@@ -2295,12 +2298,25 @@ class WebSocketImpl final: public WebSocket {
...
+                kj::Maybe<CompressionParameters> compressionConfig = nullptr,
                 kj::Array<byte> buffer = kj::heapArray<byte>(4096),
                 kj::ArrayPtr<byte> leftover = nullptr,
                 kj::Maybe<kj::Promise<void>> waitBeforeSend = nullptr)
       : stream(kj::mv(stream)), maskKeyGenerator(maskKeyGenerator),
+        compressionConfig(kj::mv(compressionConfig)),
         sendingPong(kj::mv(waitBeforeSend)),
-        recvBuffer(kj::mv(buffer)), recvData(leftover) {}
+        recvBuffer(kj::mv(buffer)), recvData(leftover) {
+    #if KJ_HAS_ZLIB
```
kentonv, line 2309: We normally don't intent preprocessor directives at all, though the code that they surround is indented as it would be if the preprocessor directives weren't there.

(Occasionally if preprocessor directives are deeply nested and not interleaved with regular code, I might indent them... not in places like this though.)

c++/src/kj/compat/http.c++
```diff
@@ -2295,12 +2298,25 @@ class WebSocketImpl final: public WebSocket {
...
+    KJ_IF_MAYBE(config, this->compressionConfig) {
+      compressionContext.emplace(ZlibContext::Mode::COMPRESS, *config);
+      decompressionContext.emplace(ZlibContext::Mode::DECOMPRESS, *config);
+    }
+    #else
+    this->compressionConfig = nullptr;
```
kentonv, line 2315: I think we should be KJ_REQUIREing that this is null, rather than changing it to null? If it is non-null then we've already negotiated compression settings we can't support.

On another note, usually if I find myself using `this->blah` in a constructor to avoid confusion with a parameter, I instead rename the parameter to `blahParam` so that `blah` unambiguously refers to the member.

kentonv (commented):
> Edit 5: That seems to have done it. My sense is that we should change BUILD_KJ_GZIP to something like BUILD_WITH_ZLIB since we rely on it for both kj_http and kj_gzip. Thoughts?

Makes sense. Or maybe KJ_HAS_ZLIB to match the preprocessor define?

c++/src/kj/compat/http.c++
MellowYarker: Here's an interesting result. I've generated a random sequence of 8192 lowercase characters. This is what happens when sending and receiving from a KJ websocket:

### SEND
```
[sendImpl] about to process message, input message size: 8192
[processLoop] starting processLoop
[processLoop] avail_in at start: 8192
[processLoop] iteration: 0
   [processLoop] bytes processed this run: 4096
   [processLoop] avail_in: 0
   [processLoop] avail_out: 0
[processLoop] iteration: 1
   [processLoop] bytes processed this run: 1022
   [processLoop] avail_in: 0
   [processLoop] avail_out: 3074
[sendImpl] Done Processing Message, output size: ; innerMessage.size() = 5118
```
### RECEIVE
```
[receive] about to process TEXT
[processLoop] starting processLoop
[processLoop] avail_in at start: 5118
[processLoop] iteration: 0
    [processLoop] bytes processed this run: 4096
    [processLoop] avail_in: 2559
    [processLoop] avail_out: 0
[processLoop] iteration: 1
    [processLoop] bytes processed this run: 4096
    [processLoop] avail_in: 0
    [processLoop] avail_out: 0
[processLoop] iteration: 2
    [processLoop] bytes processed this run: 0
    [processLoop] avail_in: 0
    [processLoop] avail_out: 4096
[receive] done processing TEXT, output size: 8193
```

Note that the output on the `receive` size is larger than the original input by 1 character because we append the null terminator.

This seems to conflict with what the zlib header says:
```
DEFLATE...
  If the parameter flush is set to Z_SYNC_FLUSH, all pending output is
  flushed to the output buffer and the output is aligned on a byte boundary, so
  that the decompressor can get all input data available so far.  (In
  particular avail_in is zero after the call if enough output space has been
  provided before the call.
```

My reading of this (specifically `avail_in` and `avail_out` being 0 in the compression (send) example) is that zlib ends up reading all of `avail_in` into its internal state, but can't flush everything to the output buffer in one go. So as long as we get `avail_out == 0`, we need to flush again, i.e. we should be checking for `avail_out != 0` if we're using `Z_SYNC_FLUSH`.

---
As for `Z_STREAM_END`, I agree we wouldn't expect to see it when sending, and _realistically_ we wouldn't ever expect to see it when receiving (since it doesn't seem likely someone is using an implementation that doesn't allow sending empty uncompressed blocks). My sense is that we should be treating it as any other message, so it worries me a bit that `zlib` throws `Z_STREAM_END` at me when it sees a block with `bfinal` set. I suspect it shouldn't be breaking like this because [the rfc says](https://datatracker.ietf.org/doc/html/rfc7692#section-7.2.3.4):
```
   The last 1 octet (0x00) contains the header
   bits with "BFINAL" set to 0 and "BTYPE" set to 00, and 5 padding bits
   of 0.  This octet is necessary to allow the payload to be
   decompressed in the same manner as messages flushed using DEFLATE
   blocks with "BFINAL" unset.
```
In this case it's referring to the payload `0xf3 0x48 0xcd 0xc9 0xc9 0x07 0x00 0x00` (first 7 octets are the message, last is the octet described above).

c++/src/kj/compat/http.c++
MellowYarker: I might be totally off here and/or overthinking this, but suppose I wanted to use KJ websockets for proxying and I didn't build with `zlib`. Let's also say I still want the client/server to use compression. In this case, couldn't the client/server technically use compression while the proxy just does an optimized pump? In that case, doesn't it sort of make sense to actually allow non-null `compressionConfig`s?

I guess what I'm asking is would it make more sense to:
- Not force `compressionConfig` to be null here
- `KJ_REQUIRE` the config to be null if we try to _unoptimized_ pump using websocket compression (when zlib hasn't been linked)

Also noted for the param, I'll fix that. Thanks for mentioning because IIRC I got confused by why I had to use `this`.

c++/src/kj/compat/http.c++
MellowYarker: https://github.com/capnproto/capnproto/pull/1505/commits/0da5d412d313a2c3ca628fbc3c6b29720f61cd16

Realized I linked the fixup to the wrong commit (previously, not this fixup ^) so I'm going to move these preprocessor directives to the correct commit next.

c++/src/kj/compat/http.c++
MellowYarker: Param name fixes:
- https://github.com/capnproto/capnproto/pull/1505/commits/58d08a749eb2d5d5db75dc304d99729d4f89562c
- https://github.com/capnproto/capnproto/pull/1505/commits/427686e6e01636dc3171b7d61fe508460c545232

MellowYarker:
I've reordered the commits and moved the zlib build commit up before the commit that introduces zlib related code: https://github.com/capnproto/capnproto/pull/1505/commits/ac560be7e691a633916f6309f37c6b0acd48f3e8. Hopefully this fixes the build issue, if not, then it might have to do with the `BUILD_KJ_GZIP -> KJ_HAS_ZLIB` rename.

MellowYarker:
Rebased to get commits from https://github.com/capnproto/capnproto/pull/1504 as I wanted to finish a TODO related to that PR.

c++/src/kj/compat/http.c++
MellowYarker: Followed up on this now that #1504 was merged: https://github.com/capnproto/capnproto/pull/1505/commits/4b3520dc75dc38872ec187f5734895d614beb254

MellowYarker:
> So, I think we need to make sure that, at least as used in Workers, KJ respects exactly the response headers that the application gives it. If the application does not respond to the extensions request, then KJ should not use compression.
> 
> But that behavior is only what we want in Workers. It is _not_ what we'd want in a regular KJ-based application server, where implementing compression is probably desirable.
> 
> So I think our HttpServerOptions needs three modes: No compression, automatic compression, and manual compression. In manual compression mode, KJ never completes the extension handshake itself, it only passes exactly the headers the application gives it and implements that mode (or fails if the application specifies a mode that KJ can't handle).

I think this (https://github.com/capnproto/capnproto/pull/1505/commits/d05c6a48b71a899418b29e6e8aee8986e77e93ef) implements what you've described. One immediate consequence of this change is that if we use `MANUAL_COMPRESSION` in workers (without changing anything else), then when a worker behaves as the websocket server, it'll ignore all websocket compression requests (since it'll look at the `headers` passed to `acceptWebSocket(headers)` rather than the `requestHeaders`).

I was a bit hesitant to push this because I don't see an obvious way to fix that particular issue, so really just looking for clarification here.

c++/src/kj/compat/http.c++
kentonv: We should throw an exception and terminate the connection on `BFINAL`, it is an error for the other end to send that.

c++/src/kj/compat/http.c++
kentonv: > Let's also say I still want the client/server to use compression. In this case, couldn't the client/server technically use compression while the proxy just does an optimized pump? In that case, doesn't it sort of make sense to actually allow non-null compressionConfigs?

In that case you would need to make sure the code throws an exception if you try to do anything other than an optimized pump. This does not seem worth the effort to me. I think it's fine if we say you can't have compression at all unless you compile with zlib. Compiling with zlib is not a high barrier anyway.

kentonv:
> One immediate consequence of this change is that if we use MANUAL_COMPRESSION in workers (without changing anything else), then when a worker behaves as the websocket server, it'll ignore all websocket compression requests

Hmm yeah that's a problem. It seems like we really want to specify this when we call `acceptWebSocket()`, at which point we know if we're proxying or not. The answer to this might be to set manual mode, but also have KJ provide a separate function that can be called to ask it to choose the appropriate response to a given extension request. Then the workers runtime will call that to set the header in the case that the WebSocket is being terminated locally.

Note that we need the same enum on the client side. In automatic mode, KJ should automatically offer the extensions it supports. In manual mode, the application needs to pass the header.

BTW, I think `NO_COMPRESSION` is the same as setting `enableWebSocketCompression = false`? Can we remove the boolean and just use the enum?

MellowYarker:
Removed `enableWebSocketCompression` `bool` from both client and server: https://github.com/capnproto/capnproto/pull/1505/commits/b66dac90e501d401d822e5359a984a65f0aa2267
Set `AUTOMATIC_COMPRESSION` in tests: https://github.com/capnproto/capnproto/pull/1505/commits/fd9b618719fa05a870c0935313f0e3b7bd0d9b76

Going to rebase, then reorder commits so the compression modes is one of the earlier commits.

c++/src/kj/compat/http.c++
```diff
@@ -6402,8 +6726,16 @@ private:
 
     kj::Maybe<CompressionParameters> acceptedParameters;
     kj::String agreedParameters;
-    if (server.settings.enableWebSocketCompression) {
-      KJ_IF_MAYBE(value, requestHeaders.get(HttpHeaderId::SEC_WEBSOCKET_EXTENSIONS)) {
+    if (server.settings.webSocketCompressionMode != HttpServerSettings::NO_COMPRESSION) {
+      auto mode = server.settings.webSocketCompressionMode;
+      auto& target = mode == HttpServerSettings::AUTOMATIC_COMPRESSION ?
+          requestHeaders : (headers.isWebSocket() ? headers : requestHeaders);
```
MellowYarker, line 6732: @kentonv curious what you think of this. I think we can determine if we're proxying by checking if `headers.isWebSocket()` is true (when we're in `MANUAL_COMPRESSION` mode). I ran some tests against EW and found that when acting as a server, `headers.isWebSocket() == false`, and when proxying, `headers.isWebSocket() == true`. Would this be preferable to:
> Hmm yeah that's a problem. It seems like we really want to specify this when we call acceptWebSocket(), at which point we know if we're proxying or not. The answer to this might be to set manual mode, but also have KJ provide a separate function that can be called to ask it to choose the appropriate response to a given extension request. Then the workers runtime will call that to set the header in the case that the WebSocket is being terminated locally.

Here's the actual fixup: https://github.com/capnproto/capnproto/pull/1505/commits/a6d42c2c4af81e4394bfa652e78c15ae7df2549d, I just commented on file so it would show up in the PR.

c++/src/kj/compat/http.c++
MellowYarker: https://datatracker.ietf.org/doc/html/rfc7692#section-7.2.1 suggests it's not an error.
```
- An endpoint MAY use DEFLATE blocks of any type.

- An endpoint MAY use both DEFLATE blocks with the "BFINAL" bit set
  to 0 and DEFLATE blocks with the "BFINAL" bit set to 1.

- When any DEFLATE block with the "BFINAL" bit set to 1 doesn't end
  at a byte boundary, an endpoint MUST add minimal padding bits of 0
  to make it end at a byte boundary.  The next DEFLATE block follows
  the padded data if any.
```

https://datatracker.ietf.org/doc/html/rfc7692#section-7.2.3.4 provides an example where `BFINAL = 1`

So I think we would be deviating from the spec if we threw an exception here.

c++/src/kj/compat/http.c++
MellowYarker: https://github.com/capnproto/capnproto/pull/1505/commits/6dc671574cb7516a353781c17fafa11556d72932

c++/src/kj/compat/http.h
```diff
@@ -844,8 +844,12 @@ struct HttpClientSettings {
   // Customize how protocol errors are handled by the HttpClient. If null, HttpClientErrorHandler's
   // default implementation will be used.
 
-  bool enableWebSocketCompression = false;
-  // We won't attempt to parse `Sec-WebSocket-Extensions` by default.
+  enum WebSocketCompressionMode {
+    NO_COMPRESSION,
+    MANUAL_COMPRESSION,    // TODO(now) add comment.
```
kentonv, line 849: `TODO(now)`

c++/src/kj/compat/http.c++
kentonv: Oh, ok, then we need to do what the spec says to do. Sorry, I assumed the spec said never to send this. Why would the client ever send BFINAL? Would it only happen on the last message of the stream?

c++/src/kj/compat/http.c++
kentonv: I don't think this is a reliable way to detect proxying. I also don't think we should be detecting proxying. The point here is that some applications want to dictate exactly what compression options are used and some don't. Proxying is just a common reason why an app might want control, but it's entirely possible that a proxy might _want_ to add/remove compression, and that a non-proxy might want to avoid it for other reasons.

kentonv:
> If all this is true, why would we need MANUAL_COMPRESSION for the client?

Let's say that the eyeball opens a connection to Workers and doesn't specify support for compression. The Worker proxies this connection to origin. With automatic compression on HttpClient, KJ will kick in here and add compression to the origin connection, which then means that we cannot do an optimized pump.

> Automatic: take the Sec-WebSocket-Extensions header given to us by the caller and run it through the checks from the previous PR before sending off the validated extensions to the server

No no, that's what manual mode should do. In Automatic mode, KJ ignores any `Sec-WebSocket-Extensions` from the app and instead always replaces it with a header requesting compression.

kentonv:
Same on the server side: In automatic mode, `acceptWebSocket()` should ignore the `Sec-WebSocket-Extensions` header passed from the application and always send back a header based on what the client had sent, accepting whichever client-requested extensions KJ supprots. In manual mode, `acceptWebSocket()` would take the `Sec-WebSocket-Extensions` header specified by the app, filter it down to the subset that are supported both by KJ and by the client (according to the request headers), and then send that back.

Note that in manual mode (on either side), if the application does not send a `Sec-WebSocket-Extensions` header, then no compression will be applied, regardless of whether KJ or the peer otherwise supports it.

c++/src/kj/compat/http.c++
MellowYarker: Gotcha, I've removed this fixup and force pushed.

c++/src/kj/compat/http.h
MellowYarker: https://github.com/capnproto/capnproto/pull/1505/commits/6cf36ab95562af81c9fa9fb415d151fb732a4fc6

MellowYarker:
> > Automatic: take the Sec-WebSocket-Extensions header given to us by the caller and run it through the checks from the previous PR before sending off the validated extensions to the server
> 
> No no, that's what manual mode should do. In Automatic mode, KJ ignores any `Sec-WebSocket-Extensions` from the app and instead always replaces it with a header requesting compression.

Yeah this makes sense, sorry about that, I got confused because I was too focused on the server side and proxying. I think this handles the client side code: https://github.com/capnproto/capnproto/pull/1505/commits/7c6e5360a15a5df349c136b0949f89281c12bf0e. I've fixed the setting in test as well https://github.com/capnproto/capnproto/pull/1505/commits/5fe77bcd7f0499c82bcb0f769df731e075fe5544

MellowYarker:
> Same on the server side: In automatic mode, `acceptWebSocket()` should ignore the `Sec-WebSocket-Extensions` header passed from the application and always send back a header based on what the client had sent, accepting whichever client-requested extensions KJ supprots. In manual mode, `acceptWebSocket()` would take the `Sec-WebSocket-Extensions` header specified by the app, filter it down to the subset that are supported both by KJ and by the client (according to the request headers), and then send that back.
> 
> Note that in manual mode (on either side), if the application does not send a `Sec-WebSocket-Extensions` header, then no compression will be applied, regardless of whether KJ or the peer otherwise supports it.

Alright, I've made it so in automatic mode we ignore the `headers` passed by the application, and in manual mode we start with the config manually passed via `headers` and filter it down to a subset supported by the client: https://github.com/capnproto/capnproto/pull/1505/commits/b70ceab8122403794c871cffb8aa70cc27fd7789.

Couple things to note:
- There's a few `TODO`s in there concerning the `server_x` parameters. The issue being if a client requests a `server_x` param and the application explicitly does not offer this, we could either:
  - Support it anyways because we can do that in KJ (I think this is probably _not_ what we want, it would make proxying more confusing IMO).
  - Reject the offer because when a client sends `server_x` parameters, it's not a hint, it's a restriction.

- Assuming we want to reject, I'll need to change the `tryParseExtensionOffers()` function (or maybe duplicate it) so that we get an array of all the fully parsed + validated offers from the client. Then, we could loop through each offer, and if the offer includes a `server_x` parameter (and the application does not support it) then we can drop it and go to the next offer. This is why the RFC specifies that a client can provide multiple offers, see [7.1.3](https://datatracker.ietf.org/doc/html/rfc7692#section-7.1.3) for an example of a client sending one offer with `server_max_window_bits` and another without it.

`acceptWebSocket` is getting kind of crowded now, so I might start splitting the part where we handle `MANUAL_COMPRESSION` into functions. Also still need to handle the case where we want to terminate the websocket server in a worker, though I think I'm going to do that last.

MellowYarker:
> The answer to this might be to set manual mode, but also have KJ provide a separate function that can be called to ask it to choose the appropriate response to a given extension request. Then the workers runtime will call that to set the header in the case that the WebSocket is being terminated locally.

Just collecting some thoughts here, 
- We're using `MANUAL_COMPRESSION` in workers (to support proxying), 
- we know at the time that we call `acceptWebSocket()` whether we're proxying or not, and
- if we're terminating the websocket in a worker (server side), then we really just want it to operate in `AUTOMATIC_COMPRESSION` mode.

So all this considered, wouldn't it make sense to add a `bool` param to `acceptWebSocket()` that allows us to override manual mode and use automatic instead? It would default to false but if we know we want to ignore `headers`, then it seems fairly straightforward to just say "use automatic compression for this call".

c++/src/kj/compat/http.c++
jrsala-auguration: @kentonv Hi, I'm the author of #1428 and as far as I'm concerned it's finished unless you have further comments on it which I'll be happy to look at. We've been using that code in our branch of Cap'n Proto for months now and it's working fine.

c++/src/kj/compat/http.c++
```diff
@@ -4345,16 +4753,25 @@ public:
...
-          clientOffer = extensions.front();
-          connectionHeaders[HttpHeaders::BuiltinIndices::SEC_WEBSOCKET_EXTENSIONS] =
-              offeredExtensions.emplace(_::generateExtensionRequest(extensions.asPtr()));
-        }
+        extensions = _::findValidExtensionOffers(*value);
       }
+    } else if (compressionMode == HttpClientSettings::AUTOMATIC_COMPRESSION) {
+      // If AUTOMATIC_COMPRESSION is enabled, we send `Sec-WebSocket-Extensions: permessage-deflate`
+      // to the server and ignore the `headers` provided by the caller.
+      constexpr auto default_config = "permessage-deflate";
```
kentonv, line 4767: Should use camel case: `defaultConfig`

However, do we really want to initialize the config by parsing a string? Couldn't you do `extensions.add(CompressionParameters())` instead and avoid the parsing?

c++/src/kj/compat/http.c++
```diff
@@ -4345,16 +4753,25 @@ public:
...
+      extensions = _::findValidExtensionOffers(default_config);
+    }
+
+    if (extensions.size() > 0) {
+      clientOffer = extensions.front();
```
kentonv, line 4772: I still don't really understand why we allow multiple extension offers to be sent, but then primarily care about the first one. Sorry if I asked this already... what's the use case for sending multiple offers? And why do we specifically remember the first offer we sent and not the others?

c++/src/kj/compat/http.c++
```diff
@@ -6402,12 +6822,35 @@ private:
...
         KJ_IF_MAYBE(config, _::tryParseExtensionOffers(*value)) {
           acceptedParameters = kj::mv(*config);
         }
       }
+    } else if (compressionMode == HttpServerSettings::MANUAL_COMPRESSION) {
+      // If MANUAL_COMPRESSION is enabled, we use the `headers` passed in by the application, and
+      // try to find a configuration that respects both the server's preferred configuration,
+      // as well as the client's requested configuration.
+      //
+      // TODO(now): Manual Mode and terminating in worker
```
kentonv, line 6840: What if the caller of `acceptWebSocket()` simply always sets `Sec-WebSocket-Extension: permessage-deflate`? If the client didn't ask for `permessage-deflate`, then KJ will filter it out, right? So would this be an effective way to accept whatever the client offered?

If that doens't work, then I think we should expose a utility function like `kj::choosePreferredWebSocketExtensions(kj::StringPtr clientOffer)`. When a WebSocket is terminating in a worker, the Workers Runtime could call this function to construct the Sec-WebSocket-Extensions header.

c++/src/kj/compat/http.c++
MellowYarker: > Why do we allow multiple offers to be sent?

Right, so it really boils down to the fact that the server might reject the client's requested parameters. If we don't have fallback options, the connection won't use compression at all.

Let's say we're the client and we send `permessage-deflate; server_max_window_bits = 10`. The server gets this and doesn't recognize `server_max_window_bits`, so it rejects the offer and replies without the `Sec-WebSocket-Extensions` header (i.e. we establish a regular websocket connection). Here's an example from the rfc (https://datatracker.ietf.org/doc/html/rfc7692#section-7.1.3)

```
        Sec-WebSocket-Extensions:
           permessage-deflate;
           client_max_window_bits; server_max_window_bits=10

   This extension negotiation offer might be rejected by the server
   because the server doesn't support the "server_max_window_bits"
   extension parameter in an extension negotiation offer.  This is fine
   if the client cannot receive messages compressed using a larger
   sliding window size, but if the client just prefers using a small
   window but wants to fall back to the "permessage-deflate" without the
   "server_max_window_bits" extension parameter, the client can make an
   offer with the fallback option like this:

       Sec-WebSocket-Extensions:
           permessage-deflate;
           client_max_window_bits; server_max_window_bits=10,
           permessage-deflate;
           client_max_window_bits

   The server can accept "permessage-deflate" by picking any supported
   one from the listed offers.  To accept the first option, for example,
   the server may send back a response as follows:

       Sec-WebSocket-Extensions:
           permessage-deflate; server_max_window_bits=10

   To accept the second option, for example, the server may send back a
   response as follows:

       Sec-WebSocket-Extensions: permessage-deflate
```
---

> Why do we remember the first offer we sent and not the others?

Couple things to note here:
1. Offers are ordered by preference.
2. We can assume that offer `n` has more parameters than offer `n + 1`.
3. We only want to retain `client_no_context_takeover` and `client_max_window_bits` (the client's compressor config).

So first, why retain this info at all? Basically, the `client_x` parameters are hints to the server about how the client will configure its _compressor_. The server isn't required to acknowledge these hints (by replying with the same parameters in the response), so if we have a situation where the client sends `client_no_context_takeover` and the server ignores it, then the only way the client can use `client_no_context_takeover` is if it remembered that it sent it in the first place.

As for why we only retain the first one, see (1) and (2). The assumption is that if a client wants to use any `client_x` parameters, then offer 1 will have all the `client_x` params that they'd want to use.

kentonv: OK, I get it now. Thanks.

Maybe worthwhile to explain this in a code comment.

c++/src/kj/compat/http.c++
MellowYarker: I think my assumption was if someone wanted to use `AUTOMATIC_COMPRESSION` but also add parameters then it'd be easier to just modify the string, but I guess in that case it would make more sense to use `MANUAL_COMPRESSION`.

c++/src/kj/compat/http.c++
MellowYarker: IMO, this is fundamentally an issue on the workers (application) side of things -- we're in manual mode but we aren't passing headers to `acceptWebSocket()`. I think we should determine if we're proxying, then populate `headers` (with the client's requested `Sec-WebSocket-Extensions` value) if we're terminating in a worker. That would resolve the issue.

### Terminating websocket in a worker
This is the current state of things:
1. Client offers `permessage-deflate`
2. worker gets the request, we start building the response and see we want to accept a websocket
3. `headers` passed to `acceptWebSocket()` is empty
5. Because we're in manual mode, we look at `headers` and see it doesn't want to use compression so we reply to the client without `Sec-WebSocket-Extensions` (even if the client requested it).

We don't want to set `Sec-WebSocket-Extensions: permessage-deflate` because that would prevent us from ever allowing the `server_x` parameters (and we would always ignore the `client_x` params). I think the most appropriate thing to do is just use the exact header that the client provided.
### Proxying
Proxying should work correctly, since we set `headers` to be the response headers from the origin server.

c++/src/kj/compat/http.c++
MellowYarker: I'm not really sure to be honest :slightly_frowning_face:, if zlib see's `BFINAL` it ends the stream so it's no longer useful. If we're using `inbound_no_context_takeover` then we just reset the stream for each message, in which case its fine, but if we're using the same stream for the entire connection, then it only makes sense to send this for the final message.

Messages are typically flushed by including empty deflate blocks without compression, but the RFC says implementations that use a compression platform that doesn't support empty deflate blocks without compression can use `BFINAL` instead.

```
   On platforms on which the flush method using an empty DEFLATE block
   with no compression is not available, implementors can choose to
   flush data using DEFLATE blocks with "BFINAL" set to 1.
```

The only thing that's clear to me is we should still be able to decompress the message. It's unclear whether it should end the stream:
```
       0xf3 0x48 0xcd 0xc9 0xc9 0x07 0x00 0x00

   This is the payload of a message containing "Hello" compressed using
   a DEFLATE block with "BFINAL" set to 1.  The first 7 octets
   constitute a DEFLATE block with "BFINAL" set to 1 and "BTYPE" set to
   01 containing "Hello".  The last 1 octet (0x00) contains the header
   bits with "BFINAL" set to 0 and "BTYPE" set to 00, and 5 padding bits
   of 0.  This octet is necessary to allow the payload to be
   decompressed in the same manner as messages flushed using DEFLATE
   blocks with "BFINAL" unset
```

It's really not obvious what the answer is here, I don't think zlib provides a way to override the default behavior of ending the stream on receiving a `BFINAL`. From what I read online, most implementations will not use `BFINAL` and none of the browsers send it.

c++/src/kj/compat/http.c++
MellowYarker: https://github.com/capnproto/capnproto/pull/1505/commits/c350aaf975f89d5e16aaa613c3535ce5d21112d1

c++/src/kj/compat/http.c++
MellowYarker: https://github.com/capnproto/capnproto/pull/1505/commits/c350aaf975f89d5e16aaa613c3535ce5d21112d1

c++/src/kj/compat/http.h
```diff
@@ -1273,6 +1287,12 @@ kj::String generateExtensionRequest(const ArrayPtr<CompressionParameters>& exten
 
 kj::Maybe<CompressionParameters> tryParseExtensionOffers(StringPtr offers);
 
+kj::Maybe<CompressionParameters> tryParseAllExtensionOffers(StringPtr offers,
+    CompressionParameters manualConfig);
+
+kj::Maybe<CompressionParameters> compareClientAndServerConfigs(CompressionParameters requestConfig,
```
MellowYarker, line 1293: I should add some tests for these new functions, will do that now.

MellowYarker: https://github.com/capnproto/capnproto/pull/1505/commits/245d063b095a6c799764b5305cc9f1ca9a979c5d

Test found a subtle bug so I'm happy about that.

c++/src/kj/compat/http.h
```diff
@@ -844,8 +844,12 @@ struct HttpClientSettings {
   // Customize how protocol errors are handled by the HttpClient. If null, HttpClientErrorHandler's
   // default implementation will be used.
 
-  bool enableWebSocketCompression = false;
-  // We won't attempt to parse `Sec-WebSocket-Extensions` by default.
+  enum WebSocketCompressionMode {
+    NO_COMPRESSION,
+    MANUAL_COMPRESSION,    // Let's the application decide the compression configuration (if any).
```
MellowYarker, line 849: `s/Let's/Lets`

MellowYarker: https://github.com/capnproto/capnproto/pull/1505/commits/9e8b6ff72a83acdb1276403631c4411596ebde22

c++/src/kj/compat/http.c++
kentonv: Well if the other side sent BFINAL, doesn't that imply that they reset their compressor? You can't sent more bytes from the same compression context after BFINAL, right? So resetting the decompressor seems like the right answer to me.

MellowYarker: https://github.com/capnproto/capnproto/pull/1505/commits/050d7dd2a9b15504635c87b15a158258df367494

kentonv:
It seems like everything has been addressed. Questions:
- Is there anything else you planned to do before merging?
- Did you want to test the BFINAL handling in some way?
- Have you done any integration testing with a real browser? I'm OK with merging this before that but wanted to know what the plan is.

MellowYarker:
> Is there anything else you planned to do before merging?

Not on the KJ side, no.
> Did you want to test the BFINAL handling in some way?

We test that we can receive the message and decompress it [here](https://github.com/capnproto/capnproto/pull/1505/files#diff-bad743492083c5cd840b54442f4c32da3375e01f43c2fa021c927a4588f5a47eR3123), but I can push a change to make sure we can read new messages after resetting the decompressor on receiving `BFINAL`. I've got this locally and it seems to work so no harm in adding that quickly.
> Have you done any integration testing with a real browser? I'm OK with merging this before that but wanted to know what the plan is.

I haven't done that, but I used python's `zlib` to compress some data and then decompressed it using a KJ websocket (as in literally copy/pasted the bytes into a kj-test). I've got some tests on the Workers side too, for example, sending a larger (~8192 byte) message and having it echoed back. I'd like to run some tests on an actual metal though.

kentonv:
You could test against a local preview instance... write a little worker that returns some HTML that opens a WebSocket.

But I'll go ahead and merge this and we can address problems in future PRs.

PR #1506, opened 2022-07-26
KJ_EXPECT_EXIT and KJ_EXPECT_SIGNAL fork tests

mikea:


kentonv (commented):
Ugh I apparently wrote these comments a few days ago and failed to submit them.

c++/src/kj/test-helpers.c++
```diff
@@ -153,5 +153,39 @@ bool expectFatalThrow(kj::Maybe<Exception::Type> type, kj::Maybe<StringPtr> mess
...
 
+bool expectDeath(Maybe<int> statusCode, Function<void()> code) {
+#if _WIN32
+  // We don't support death tests on Windows due to lack of efficient fork.
+  return true;
+#else
+  pid_t child;
+  KJ_SYSCALL(child = fork());
+  if (child == 0) {
+    KJ_DEFER(_exit(0));
```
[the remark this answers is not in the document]
mikea: done

c++/src/kj/test-helpers.c++
```diff
@@ -153,5 +153,39 @@ bool expectFatalThrow(kj::Maybe<Exception::Type> type, kj::Maybe<StringPtr> mess
...
+  if (WIFEXITED(status)) {
+    KJ_IF_MAYBE(s, statusCode) {
+      KJ_EXPECT(WEXITSTATUS(status) == *s);
+      return WEXITSTATUS(status) == *s;
+    } else {
+      KJ_EXPECT(WEXITSTATUS(status) != 0);
+      return WEXITSTATUS(status) != 0;
+    }
+  } else if (WIFSIGNALED(status)) {
+    KJ_FAIL_EXPECT("subprocess crashed without throwing exception", WTERMSIG(status));
```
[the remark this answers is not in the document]
mikea: done

c++/src/kj/test-helpers.c++
```diff
@@ -153,5 +153,39 @@ bool expectFatalThrow(kj::Maybe<Exception::Type> type, kj::Maybe<StringPtr> mess
...
+
+  if (WIFEXITED(status)) {
+    KJ_IF_MAYBE(s, statusCode) {
+      KJ_EXPECT(WEXITSTATUS(status) == *s);
+      return WEXITSTATUS(status) == *s;
+    } else {
+      KJ_EXPECT(WEXITSTATUS(status) != 0);
+      return WEXITSTATUS(status) != 0;
+    }
+  } else if (WIFSIGNALED(status)) {
```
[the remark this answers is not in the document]
mikea: done

mikea:
All comments addressed, added more test cases. Please take a look.

c++/src/kj/test-helpers.c++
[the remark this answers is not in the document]
kentonv: It looks like we're now overloading status codes and signals. So if I'm expecting `SIGABRT` but the process actually exits with code 6 then it'll be considered a pass.

Shouldn't we distinguish between these? Like KJ_EXPECT_EXIT vs. KJ_EXPECT_SIGNAL or something?

c++/src/kj/test-helpers.c++
```diff
@@ -153,5 +154,44 @@ bool expectFatalThrow(kj::Maybe<Exception::Type> type, kj::Maybe<StringPtr> mess
...
+bool expectDeath(Maybe<int> statusCode, Function<void()> code) {
+#if _WIN32
+  // We don't support death tests on Windows due to lack of efficient fork.
+  return true;
+#else
+  pid_t child;
+  KJ_SYSCALL(child = fork());
+  if (child == 0) {
+    resetSignalHandlers();
+    code();
```
kentonv, line 166: It looks like now, if `code()` throws an exception, the exception will propagate to the caller where it could be caught, leaving the child process to continue executing the test in its fork. That seems weird. I think we need to make sure `expectDeath()` never allows an exception to propagate out. The easiest way to do that is to mark it `noexcept`, in which case the terminate handler will always be invoked.

mikea:
@kentonv I've separated signal and exit expectation in KJ_EXPECT_EXIT and KJ_EXPECT_SIGNAL macros. PTAL.

c++/src/kj/test.h
```diff
@@ -128,6 +141,17 @@ bool expectFatalThrow(Maybe<Exception::Type> type, Maybe<StringPtr> message,
 // fork() is not available, this always returns true.
 #endif
 
+bool expectExit(Maybe<int> statusCode, Function<void()> code);
```
kentonv, line 144: It looks like the first parameter to both of these functions can just be `int` instead of `Maybe<int>` since the corresponding macros never pass `nullptr`?

kentonv, line 144: You can use `FunctionParam` instead of `Function` since the function will always be called synchronously.

c++/src/kj/test-helpers.c++
```diff
@@ -153,5 +153,65 @@ bool expectFatalThrow(kj::Maybe<Exception::Type> type, kj::Maybe<StringPtr> mess
...
+  if (WIFEXITED(status)) {
+    KJ_IF_MAYBE(s, statusCode) {
+      KJ_EXPECT(WEXITSTATUS(status) == *s);
+      return WEXITSTATUS(status) == *s;
+    } else {
+      KJ_EXPECT(WEXITSTATUS(status) != 0);
+      return WEXITSTATUS(status) != 0;
+    }
+  } else {
+    KJ_FAIL_EXPECT("subprocess didn't exit", status);
```
kentonv, line 180: The `status` value produced by `waitpid()` is pretty hard for a human to interpret, since it's a bitfield of a bunch of stuff. Could we check `WIFSIGNALED()` here and if so produce an error like "code raised a signal instead of exiting; signal = 11 (Segmentation Fault)"? (You can use `strsignal()` from the C library to convert the signal number to a name.)

Similarly below if we are expecting a signal and instead the code exits, we should say what the exit code was.

c++/src/kj/test-helpers.c++
kentonv: You marked this resolved but it doesn't look like it was?

c++/src/kj/test.h
mikea: You can pass `nullptr` in manually in case where you don't know the exit code or signal for sure. I've added couple tests with nullptr too.

c++/src/kj/test-helpers.c++
mikea: sorry, marked them as `noexcept`.

mikea:
@kentonv All done. PTAL.

c++/src/kj/test-helpers.c++
```diff
@@ -153,5 +153,72 @@ bool expectFatalThrow(kj::Maybe<Exception::Type> type, kj::Maybe<StringPtr> mess
...
+      KJ_EXPECT(WEXITSTATUS(status) == *s);
+      return WEXITSTATUS(status) == *s;
+    } else {
+      KJ_EXPECT(WEXITSTATUS(status) != 0);
+      return WEXITSTATUS(status) != 0;
+    }
+  } else {
+    if (WIFSIGNALED(status)) {
+      KJ_FAIL_EXPECT("subprocess didn't exit but triggered a signal", strsignal(WTERMSIG(status)));
+    }
```
kentonv, line 182: `EXPECT` macros do not throw an exception, they just note the failure and keep executing. So you need `} else {` here, or just return false.

Actually on that topic, it's a bit weird that these functions return a boolean in addition to the expectation failures inside. So when there's a failure, actually two failures will be reported. However, this code review has gone on too long already so I'd say don't worry about it, we can clean that up some other time. Plus it looks like `KJ_EXPECT_THROW` has some similar problems.

mikea: fixed.

c++/src/kj/exception.c++
```diff
@@ -714,6 +714,27 @@ retry:
   return filename;
 }
 
+void resetSignalHandlers() {
```
kentonv, line 717: As seen in the CI failures, this function doesn't compile on Windows.

mikea: Ugh... Sorry I missed it since there are few failing build on master.
Added ifdefs around signal stuff (still keeping `set_terminate(null)` on Windows)

kentonv: Maybe we should name it `resetCrashHandlers()` then since "signals" are a Unix thing?

I fixed MinGW last week so there should be no failing quick tests on master.

PR #1508, opened 2022-07-27
Introduce the ability to share the same signalfd among many threads.

kentonv:
It turns out that creating a signalfd is weirdly expensive. But, due to their unintuitive behavior of always checking the signals of the calling thread, not the creating thread, we can actually share just one among many threads, as long as those threads are waiting on the same signals.

We should probably stop using signalfd altogether, maybe falling back to the old longjmp() trick we use on other platforms, but that's a bigger change than I want to make right now.


jasnell (approved):
LGTM

PR #1510, opened 2022-07-29
Buffer writes using `evalLast` in TwoPartyVatNetwork

xortive:
I'm not sure if we want this to be a default, or configurable behavior -- If your event loop was backed up enough for this to cause problems, I think that you'd be hurting from queueing writes regardless. And even if so, you could run into stack overflow if you queue enough messages, which this PR should also stop from happening.

First commit is same as (https://github.com/capnproto/capnproto/pull/1509) but I can't set the merge target to a branch in my own repo for a PR in this one.

### DIFF
--- a/c++/src/capnp/rpc-twoparty.h
+++ b/c++/src/capnp/rpc-twoparty.h
@@ -79,6 +79,7 @@ public:
   // clock is used for calculating the oldest queued message age, which is a useful metric for
   // detecting queue overload
 
+  ~TwoPartyVatNetwork() noexcept(false);
   KJ_DISALLOW_COPY(TwoPartyVatNetwork);
 
   kj::Promise<void> onDisconnect() { return disconnectPromise.addBranch(); }
@@ -90,7 +91,7 @@ public:
   // Get the number of bytes worth of outgoing messages that are currently queued in memory waiting
   // to be sent on this connection. This may be useful for backpressure.
 
-  size_t getCurrentQueueCount() { return currentQueueCount; }
+  size_t getCurrentQueueCount() { return queuedMessages.size(); }
   // Get the count of outgoing messages that are currently queued in memory waiting
   // to be sent on this connection. This may be useful for backpressure.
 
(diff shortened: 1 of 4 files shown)


c++/src/capnp/rpc-twoparty.h
```diff
@@ -105,8 +105,22 @@ class TwoPartyVatNetwork: public TwoPartyVatNetworkBase,
   kj::Promise<kj::Own<TwoPartyVatNetworkBase::Connection>> accept() override;
 
 private:
-  class OutgoingMessageImpl;
   class IncomingMessageImpl;
+  class OutgoingMessageImpl final
```
kentonv, line 109: You don't need to move this class definition to the header if you instead declare `TwoPartyVatNetwork`'s destructor out-of-line.

c++/src/capnp/rpc.c++
```diff
@@ -2458,10 +2458,10 @@ private:
         handleBootstrap(kj::mv(message), reader.getBootstrap());
         break;
 
-      case rpc::Message::CALL:
+      case rpc::Message::CALL: {
```
kentonv, line 2461: Unnecessary diff?

c++/src/capnp/serialize-async.c++
```diff
@@ -366,6 +366,45 @@ kj::Promise<void> writeMessages(
...
+      // We have a message with FDs attached. We need to write any bare messages we've accumulated,
+      // if any, then write the message with FDs, then continue on with any remaining messages.
+
+      if (messages.size() > 1) {
+        remainingMessages = messages.slice(1, messages.size());
+      }
+
+      return writeMessage(messages[0].fds, messages[0].segments);
+    } else {
+      kj::Vector<kj::ArrayPtr<const kj::ArrayPtr<const word>>> bareMessages;
```
kentonv, line 384: Slight optimization...

```suggestion
      kj::Vector<kj::ArrayPtr<const kj::ArrayPtr<const word>>> bareMessages(messages.size());
```

PR #1511, opened 2022-07-31
Stop using signalfd on Linux.

kentonv:
Instead, use `epoll_pwait()` to wait on epoll events and signals at the same time.

Reasons are explained in the comment in `UnixEventPort::wait()`:

```
    // Note: Once upon a time, we used signalfd for this. However, this turned out to be more
    // trouble than it was worth. Some problems with signalfd:
    // - It required opening an additional file descriptor per thread.
    // - If the set of interesting signals changed, the signalfd would have to be updated before
    //   calling epoll_wait(), which was an extra syscall.
    // - When a signal arrives, it requires extra syscalls to read the signal info from the
    //   signalfd, as well as code to translate from signalfd_siginfo to siginfo_t, which are
    //   different for some reason.
    // - signalfd suffers from surprising lock contention during epoll_wait or when the signalfd's
    //   mask is updated in programs with many threads. Because the lock is a spinlock, this
    //   could consume exorbitant CPU.
    // - When a signalfd is in an epoll, it will be flagged readable based on signals which are
    //   pending in the process/thread which called epoll_ctl_add() to register the signalfd.
    //   This is mostly fine for our usage, except that it breaks one useful case that otherwise
    //   works: many servers are designed to "daemonize" themselves by fork()ing and then having
    //   the parent process exit while the child thread lives on. In this case, if a UnixEventPort
    //   had been created before daemonizing, signal handling would be forever broken in the child.
```

### DIFF
--- a/c++/src/kj/async-unix.c++
+++ b/c++/src/kj/async-unix.c++
@@ -379,6 +379,12 @@ UnixEventPort::UnixEventPort()
   event.events = EPOLLIN;
   event.data.u64 = 0;
   KJ_SYSCALL(epoll_ctl(epollFd, EPOLL_CTL_ADD, eventFd, &event));
+
+  // Get the current signal mask, from which we'll compute the appropriate mask to pass to
+  // epoll_pwait() on each loop. (We explicitly memset to 0 first to make sure we can compare
+  // this against another mask with memcmp() for debug purposes.)
+  memset(&originalMask, 0, sizeof(originalMask));
+  KJ_SYSCALL(sigprocmask(0, nullptr, &originalMask));
 }
 
 UnixEventPort::~UnixEventPort() noexcept(false) {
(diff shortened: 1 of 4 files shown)


c++/src/kj/async-unix.c++
```diff
@@ -54,6 +53,43 @@ bool tooLateToSetReserved = false;
...
+// returns.
+
+}  // namespace
+
+void UnixEventPort::signalHandler(int, siginfo_t* siginfo, void*) noexcept {
+  // Since this signal handler is *only* called during `epoll_pwait()`, we aren't subject to the
+  // usual signal-safety concerns. We can treat this more like a callback. So, we can just call
+  // gotSignal() directly, no biggy.
+
+  // Note that, if somehow the signal hanlder is invoked when *not* running `epoll_pwait()`, then
```
harrishancock, line 73:
```suggestion
  // Note that, if somehow the signal handler is invoked when *not* running `epoll_pwait()`, then
```

c++/src/kj/async-unix.c++
```diff
@@ -451,202 +487,111 @@ Promise<void> UnixEventPort::FdObserver::whenWriteDisconnected() {
...
 
-  struct epoll_event events[16];
-  int n = epoll_wait(epollFd, events, kj::size(events), timeout);
   if (n < 0) {
     int error = errno;
     if (error == EINTR) {
-      // We can't simply restart the epoll call because we need to recompute the timeout. Instead,
-      // we pretend epoll_wait() returned zero events. This will cause the event loop to spin once,
-      // decide it has nothing to do, recompute timeouts, then return to waiting.
+      // We received a singal. The signal handler may have queued an event to the event loop. Even
```
harrishancock, line 577:
```suggestion
      // We received a signal. The signal handler may have queued an event to the event loop. Even
```

c++/src/kj/async-unix.c++
```diff
@@ -666,6 +611,60 @@ bool UnixEventPort::doEpollWait(int timeout) {
   return woken;
 }
 
+bool UnixEventPort::poll() {
+  // Unfortunately, epoll_pwait() with a timeout of zero will never deliver actually deliver any
```
harrishancock, line 615:
```suggestion
  // Unfortunately, epoll_pwait() with a timeout of zero will never actually deliver any
```

harrishancock (approved):
LGTM!

PR #1512, opened 2022-08-01
[bazel] using correct warning settings

mikea:


mikea:
needs some more work.

PR #1517, opened 2022-08-05
Support encoding raw JSON text

ObsidianMinor:
Useful in cases where you have an already serialized JSON value that
you want to write without deserializing and serializing it again.


c++/src/capnp/compat/json.capnp
```diff
@@ -44,6 +44,11 @@ struct Value {
     # "binary" and "date" types in text, since JSON has no analog of these. This is basically the
     # reason this extension exists. We do NOT recommend using `call` unless you specifically need
     # to be compatible with some silly format that uses this syntax.
+
+    raw @7 :Text;
+    # Used to indicate that the text should be written directly to the output without
+    # modifications. Use this if you have an already serialized JSON value and don't want
+    # to feel the cost of deserializing the value just to serialize it again.
```
kentonv, line 51: I'd add:

> The parser will never produce a `raw` value -- this is only useful for serialization.

and maybe

> WARNING: You must ensure that the value is valid stand-alone JSON. It will not be verified. Invalid JSON could make the whole message unparseable. Worse, a malicious raw value could perform JSON injection attacks. Make sure that the value was produced by a trustworthy JSON encoder.

PR #1521, opened 2022-08-09
Implement buffered form of AsyncInputStream/AsyncIoStream

xortive:
I wanted to get this up earlier than later to get feedback and also to figure out what to do about `skip()` not having an equivalent for async streams -- the omission seems deliberate so I didn't want to just add it in.

I'd still like to add some tests even though this is a straightforward copy of the sync stream implementation, since that one doesn't have tests aside from being used to implement `TestPipe` in `serialize-packed-test`.

I couldn't come up with a way to implement a buffered form of `AsyncCapabilityStream`, since there's no knowledge of message boundaries at the byte stream level, so handling the case where you complete a partial read of a control message range and then read the control messages of the following range isn't possible as there's no way to know the boundaries.


kentonv:
For `AsyncCapabilityStream`: If, when reading into the buffer, you see any FDs, then you know that those FDs are attached to some suffix of the buffer, since the kernel won't read past the end of where those FDs are attached in a single read. I think everything should work if we simply assume the FDs are attached to the _last_ byte that was read. So when the caller reads that byte, we need to deliver the FDs to them.

c++/src/kj/async-io.c++
```diff
@@ -79,6 +79,122 @@ Maybe<Own<AsyncInputStream>> AsyncInputStream::tryTee(uint64_t) {
...
+ArrayPtr<const byte> AsyncBufferedInputStream::getReadBuffer() {
+  auto result = tryGetReadBuffer();
+  KJ_REQUIRE(result.size() > 0, "Premature EOF");
+  return result;
+}
+
+AsyncBufferedInputStreamWrapper::AsyncBufferedInputStreamWrapper(AsyncInputStream& inner,
+    ArrayPtr<byte> buffer) : inner(inner),
+    ownedBuffer(buffer == nullptr ? heapArray<byte>(8192) : nullptr),
+    buffer(buffer == nullptr ? ownedBuffer : buffer) {}
```
kentonv, line 93: It took me a while to understand the indentation here.

```suggestion
AsyncBufferedInputStreamWrapper::AsyncBufferedInputStreamWrapper(
    AsyncInputStream& inner, ArrayPtr<byte> buffer)
    : inner(inner),
      ownedBuffer(buffer == nullptr ? heapArray<byte>(8192) : nullptr),
      buffer(buffer == nullptr ? ownedBuffer : buffer) {}
```

c++/src/kj/async-io.h
```diff
@@ -136,6 +136,60 @@ class AsyncOutputStream: private AsyncObject {
...
 
+class AsyncBufferedInputStream {
+  // Does not inherit from InputStream so this can act as a mixin to allow implementing
+  // AsyncBufferedIoStreamWrapper
+
+public:
+  virtual ~AsyncBufferedInputStream() noexcept(false);
+
+  ArrayPtr<const byte> getReadBuffer();
+  // TODO(now): No skip for AsyncInputStream -- I guess we could implement an async form?
```
kentonv, line 147: You could add `skip()` specifically to the buffered interface, and specify for now that it's an error to skip more than the current buffer. We could consider adding it to `AsyncInputStream` later but I'm not sure it's worth it.

c++/src/kj/async-io.h
```diff
@@ -136,6 +136,60 @@ class AsyncOutputStream: private AsyncObject {
...
+  //
+  // If the second parameter is non-null, the stream uses the given buffer instead of allocating
+  // its own.  This may improve performance if the buffer can be reused.
+
+  AsyncBufferedInputStreamWrapper(AsyncBufferedInputStreamWrapper&& other);
+  KJ_DISALLOW_COPY(AsyncBufferedInputStreamWrapper);
+
+  ~AsyncBufferedInputStreamWrapper() noexcept(false);
+
+  // implements AsyncBufferedInputStream ----------------------------------
```
kentonv, line 179: I think `tryGetLength()` and `pumpTo()` should be implemented... shouldn't be too hard. `pumpTo()` should fall back to the underlying stream's `pumpTo()` as soon as the buffer is empty.

c++/src/kj/async-io.h
```diff
@@ -136,6 +136,60 @@ class AsyncOutputStream: private AsyncObject {
   // multiple times without canceling the previous promises.
 };
 
+class AsyncBufferedInputStream {
```
kentonv, line 139: Shouldn't this inherit `AsyncInputStream`?

c++/src/kj/async-io.h
```diff
@@ -168,6 +222,41 @@ class AsyncIoStream: public AsyncInputStream, public AsyncOutputStream {
   // isn't wrapping a file descriptor.
 };
 
+class AsyncBufferedIoStreamWrapper: public AsyncBufferedInputStream, public AsyncIoStream {
```
kentonv, line 225: This inheritance hierarchy won't work very well, because you're actually inheriting two copies of `AsyncInputStream`. You could theoretically fix that with virtual inheritance, but that can lead to other issues.

Another weird thing about this class is that if we later introduce a buffered output stream, someone might expect that this class will add buffering in both directions.

Perhaps we should simply change `AsyncIoMessageStream` to optionally accept the input and output as separate objects.

c++/src/kj/async-io.h
```diff
@@ -136,6 +136,60 @@ class AsyncOutputStream: private AsyncObject {
...
+public:
+  explicit AsyncBufferedInputStreamWrapper(AsyncInputStream& inner, ArrayPtr<byte> buffer = nullptr);
+  // Creates a buffered stream wrapping the given non-buffered stream.  No guarantee is made about
+  // the position of the inner stream after a buffered wrapper has been created unless the entire
+  // input is read.
+  //
+  // If the second parameter is non-null, the stream uses the given buffer instead of allocating
+  // its own.  This may improve performance if the buffer can be reused.
+
+  AsyncBufferedInputStreamWrapper(AsyncBufferedInputStreamWrapper&& other);
```
kentonv, line 174: Why is there a move constructor? That seems unexpected for this kind of class.

c++/src/kj/async-io.h
```diff
@@ -136,6 +136,60 @@ class AsyncOutputStream: private AsyncObject {
...
+  KJ_DISALLOW_COPY(AsyncBufferedInputStreamWrapper);
+
+  ~AsyncBufferedInputStreamWrapper() noexcept(false);
+
+  // implements AsyncBufferedInputStream ----------------------------------
+  ArrayPtr<const byte> tryGetReadBuffer() override;
+
+  // Implements AsyncInputStream
+  Promise<size_t> tryRead(void* buffer, size_t minBytes, size_t maxBytes) override;
+  void registerAncillaryMessageHandler(Function<void(ArrayPtr<AncillaryMessage>)> fn) override;
```
kentonv, line 184: I wonder if we actually want to implement this. The semantics for a buffered reader are weird, since you'll get ancillary messages for bytes you haven't actually read yet. The application can always call the method directly on the underlying stream object if it understands and is comfortable with this.

c++/src/kj/async-io.h
xortive: `AsyncBufferedInputStream` deliberately doesn't inherit from `AsyncInputStream` to solve the issue of two copies.

I agree about the buffered output stream problem though, and I did think about it but couldn't really come up with a great solution -- I considered was bringing over the `AsyncIoStream` implementation that lets you pass in the input/output sides to the constructor from one of the KJ tests, but this has the problem of not being able to implement the methods that are defined on `AsyncIoStream` itself.

There's also the problem of buffering for `AsyncCapabilityStream` since there is no `AsyncCapabilityInputStream`, if we do figure out a way to implement that -- we'd need to something there to have an equivalent of passing separate input/output objects for `AsyncCapabilityMessageStream`.

c++/src/kj/async-io.h
xortive: I wanted to use this class with `kj::attachVal` to tie together ownership of the inner stream / buffer where applicable. I guess you can `kj::heap` the wrapper if that seems weird, or implement the pattern where there are varations of the constructor that own the stream and/or buffer and add ownStream/ownBuffer class members.

kentonv: Yeah I would have a constructor that allows passing in an owned buffer.

It feels generally weird for a type that has async methods to be movable, since it's possibly ambiguous whether you're allowed to move it while a promise is outstanding. KJ style guide says that resource types shouldn't be movable for reasons like this.

c++/src/kj/async-io.h
kentonv: Oof, `AsyncCapabilityStream` is indeed a problem.

I missed that `AsyncBufferedInputStream` doesn't subclass `AsyncInputStream`, but that creates a different problem... you wouldn't be able to use `kj::dynamicDowncastIfAvailable` to discover if a particular `AsyncInputStream` implements `AsyncBufferedInputStream`.

xortive:
> For `AsyncCapabilityStream`: If, when reading into the buffer, you see any FDs, then you know that those FDs are attached to some suffix of the buffer, since the kernel won't read past the end of where those FDs are attached in a single read. I think everything should work if we simply assume the FDs are attached to the _last_ byte that was read. So when the caller reads that byte, we need to deliver the FDs to them.

It seems weird to not be able to guarantee that you'll get control messages when you read the first byte of a range, only at some point during reading the range. That's probably OK for capnp messages (we'd need to adjust the way `AsyncMessageReader` works) but it feels very odd.

c++/src/kj/async-io.h
xortive: I wonder if just sticking with the `MessageStream` approach might be simpler at this point. Obviously I'm biased because I don't want to throw away code, but it's difficult to come up with a nice interface here that works for normal byte streams and capability streams at this level of abstraction... If you don't want to commit to that interface in capnp, perhaps we can keep it as part of the runtime?

kentonv:
> It seems weird to not be able to guarantee that you'll get control messages when you read the first byte of a range, only at some point during reading the range. That's probably OK for capnp messages (we'd need to adjust the way AsyncMessageReader works) but it feels very odd.

I think it's something we'll just have to document. Essentially: "If the sending side attaches FDs to a byte range that is more than 1 byte in size, the receiving end may receive those FDs as part of any read in that range, not necessarily on the first bytes. However, the receiving end does guarantee that it won't read past the end of the range before the FDs have been delivered, unless `minBytes` forces it to."

Then for capnp messages, we could consider adjusting the writing side instead of the reading side, so that FDs are always attached to the first word. That does mean we have to split such message writes into two syscalls rather than one, but they are uncommon enough that it's probably not a big deal.

c++/src/kj/async-io.h
kentonv: I guess I should take another look at that version and see if I'm OK with it.

PR #1526, opened 2022-08-16
Add missing tests to bazel build

kkoenig:


kentonv:
Seems reasonable but I don't totally understand the details. Why were these tests missing in the first place?

mikea:
> Why were these tests missing in the first place?

I think I just missed some when copying/pasting lists.

PR #1528, opened 2022-08-17
implement buffered forms of `AsyncIoMessageStream`/`AsyncCapabilityMessageStream`

xortive:
Capnp's async message reading was particularly syscall heavy, as it read the first segment, segment tables, and each segment in separate calls. This PR introduces buffered readers, that read a batch of messages at once which significantly reduces the number of syscalls required to read messages.

I fixed some of the nits from the first review of this and the flaky test. I'm a bit unhappy that this implementation doesn't make use of `scratchSpace` and is lacking smarts on reusing the read buffer, but I don't think fixing that is going to massively change the structure of the PR so I'm putting up what I have for now.

### DIFF
--- a/c++/src/capnp/serialize-async.c++
+++ b/c++/src/capnp/serialize-async.c++
@@ -31,8 +31,12 @@
 #endif
 
 #include "serialize-async.h"
+#include "serialize.h"
 #include <kj/debug.h>
 #include <kj/io.h>
+#include <kj/encoding.h>
+#include <kj/buffering.h>
+#include <deque>
 
 namespace capnp {
 
@@ -539,6 +543,155 @@ kj::Promise<MessageReaderAndFds> MessageStream::readMessage(
         KJ_UNREACHABLE;
       }
   });
+}
+
+template<typename Stream>
+class BufferedAsyncMessageStream final : public BufferedMessageStream {
+public:
+  explicit BufferedAsyncMessageStream(Stream& stream,
+      BufferedStreamOptions options)
+      : buffer(stream, kj::FdRangeInputBufferOptions{
+            options.bufferSize,
+            options.maxFdsPerMessage }),
+        stream(stream),
+        options(options) {}
+
+  // Implements MessageStream
+  kj::Promise<kj::Maybe<MessageReaderAndFds>> tryReadMessage(
+      kj::ArrayPtr<kj::AutoCloseFd> fdSpace,
+      ReaderOptions options = ReaderOptions(), kj::ArrayPtr<word> scratchSpace = nullptr) override;
+  kj::Promise<void> writeMessage(
+      kj::ArrayPtr<const int> fds,
+      kj::ArrayPtr<const kj::ArrayPtr<const word>> segments) override;
+  kj::Promise<void> writeMessages(
+      kj::ArrayPtr<kj::ArrayPtr<const kj::ArrayPtr<const word>>> messages) override;
+  kj::Maybe<int> getSendBufferSize() override;
+  kj::Promise<void> end() override;
+
+  // Implements BufferedMessageStream
+  void retainLastMessageLongTerm() override;
+
+private:
+  kj::FdRangeInputBuffer<Stream> buffer;
+  kj::Maybe<kj::Own<kj::RetainableRange>> lastRead;
+
+  Stream& stream;
+  BufferedStreamOptions options;
+};
+
+kj::Own<BufferedMessageStream> makeBufferedStream(kj::AsyncIoStream& ioStream,
+    BufferedStreamOptions options) {
+  options.bufferSize = kj::max(sizeof(word), options.bufferSize / sizeof(word) * sizeof(word));
+  return kj::heap<BufferedAsyncMessageStream<kj::AsyncIoStream>>(ioStream, options);
+}
+
+kj::Own<BufferedMessageStream> makeBufferedStream(kj::AsyncCapabilityStream& capStream,
+    BufferedStreamOptions options) {
+  options.bufferSize = kj::max(sizeof(word), options.bufferSize / sizeof(word) * sizeof(word));
+  return kj::heap<BufferedAsyncMessageStream<kj::AsyncCapabilityStream>>(capStream, options);
+}
+
+kj::ArrayPtr<const word> asWords(kj::ArrayPtr<const byte> bytes) {
+  return kj::ArrayPtr<const word>(
+      reinterpret_cast<const word*>(bytes.begin()), bytes.size() / sizeof(word));
+}
+
+namespace {
+  template<typename Stream>
+  kj::Promise<kj::Own<kj::RetainableRange>> finishMessage(
+      typename kj::FdRangeInputBuffer<Stream>::Handle handle) {
+    auto bytes = handle.peek();
+    auto words = asWords(bytes);
+
+    auto expectedSize = expectedSizeInWordsFromPrefix(words);
+    auto expectedSizeInBytes = expectedSize * sizeof(word);
+
+    if (expectedSize <= words.size()) {
+      // Complete message, split it off
+      return handle.split(expectedSizeInBytes);
+    } else {
+      // Incomplete message, finish it
+      auto min = expectedSizeInBytes - bytes.size();
+      return handle.append(min).then(
+          [handle, min](size_t n) mutable {
+        if (n < min) {
+          kj::throwRecoverableException(KJ_EXCEPTION(DISCONNECTED, "Premature EOF.", n, min));
+          KJ_UNREACHABLE;
+        }
+        return finishMessage<Stream>(handle);
+      });
+    }
+  }
+}
+
+template<typename Stream>
+kj::Promise<kj::Maybe<MessageReaderAndFds>> BufferedAsyncMessageStream<Stream>::tryReadMessage(
+    kj::ArrayPtr<kj::AutoCloseFd> passedInFdSpace,
+    ReaderOptions readerOptions, kj::ArrayPtr<word> scratchSpace) {
+  auto status = buffer.tryStartRead();
+  KJ_SWITCH_ONEOF(status) {
+    KJ_CASE_ONEOF(_, kj::EndOfFile) {
+      return nullptr;
+    }
+    KJ_CASE_ONEOF(prom, kj::Promise<void>) {
+      return prom.then([this, passedInFdSpace, readerOptions, scratchSpace]() mutable {
+        return tryReadMessage(passedInFdSpace, readerOptions, scratchSpace);
+      });
+    }
+    KJ_CASE_ONEOF(handle, typename kj::FdRangeInputBuffer<Stream>::Handle) {
+      return finishMessage<Stream>(handle).then(
+            [this, readerOptions](kj::Own<kj::RetainableRange> result)
+            -> kj::Maybe<MessageReaderAndFds> {
+        lastRead = kj::addRef(*result);
+
+        auto fds = result->getFds();
+        auto reader = kj::attachVal(
+            FlatArrayMessageReader(asWords(result->getBytes()), readerOptions),
+            kj::mv(result));
+
+        return MessageReaderAndFds { kj::mv(reader), fds };
+      });
+    }
+  }
+  KJ_UNREACHABLE;
+}
+
+template<>
+kj::Promise<void> BufferedAsyncMessageStream<kj::AsyncIoStream>::writeMessage(
+    kj::ArrayPtr<const int> fds,
+    kj::ArrayPtr<const kj::ArrayPtr<const word>> segments) {
+  return capnp::writeMessage(stream, segments);
+}
+
+template<>
+kj::Promise<void> BufferedAsyncMessageStream<kj::AsyncCapabilityStream>::writeMessage(
+    kj::ArrayPtr<const int> fds,
+    kj::ArrayPtr<const kj::ArrayPtr<const word>> segments) {
+  return capnp::writeMessage(stream, fds, segments);
+}
+
+template<typename Stream>
+kj::Promise<void> BufferedAsyncMessageStream<Stream>::writeMessages(
+    kj::ArrayPtr<kj::ArrayPtr<const kj::ArrayPtr<const word>>> messages) {
+  return capnp::writeMessages(stream, messages);
+}
+
+template<typename Stream>
+kj::Maybe<int> BufferedAsyncMessageStream<Stream>::getSendBufferSize() {
+  return capnp::getSendBufferSize(stream);
+}
+
+template<typename Stream>
+kj::Promise<void> BufferedAsyncMessageStream<Stream>::end() {
+  stream.shutdownWrite();
+  return kj::READY_NOW;
+}
+
+template<typename Stream>
+void BufferedAsyncMessageStream<Stream>::retainLastMessageLongTerm() {
+  KJ_IF_MAYBE(lr, lastRead) {
+    (*lr)->retainLongTerm();
+  }
 }
 
 }  // namespace capnp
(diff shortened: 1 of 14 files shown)


c++/src/capnp/serialize-async.c++
```diff
@@ -502,4 +504,288 @@ kj::Promise<MessageReaderAndFds> MessageStream::readMessage(
...
+    kj::ArrayPtr<kj::byte> leftover, kj::Array<kj::AutoCloseFd> fds, size_t size) {
+  kj::ArrayPtr<word> leftoverWords(reinterpret_cast<word*>(leftover.begin()),
+      leftover.size() / sizeof(word));
+
+  auto& incomplete = incompleteRead.emplace(IncompleteRead {
+    kj::heapArray<byte>(leftover.size()),
+    capnp::expectedSizeInWordsFromPrefix(leftoverWords) * sizeof(word)
+  });
+  memcpy(incomplete.bytes.begin(), leftover.begin(), leftover.size());
+  incompleteReadFds = kj::mv(fds);
```
kentonv, line 720: Looking at all the private helper methods called by the template function (`tryGetIncompleteRead`, `tryGetBufferedMessage`, `bufferCompleteMessage`, `bufferIncompleteMessage`), it looks like the only actual differences in the implementation between the two classes are this line of code, and the body of `tryGetBufferedMessage`. Is that right? Can we inline more of this code into the shared template function to make it more readable?

c++/src/capnp/serialize-async.h
```diff
@@ -139,6 +141,166 @@ class AsyncCapabilityMessageStream final: public MessageStream {
...
+  };
+
+  std::deque<MessageReaderAndOwnedFds> bufferedReaders;
+
+  struct IncompleteRead {
+    kj::Array<kj::byte> bytes;
+    size_t expectedSize;
+  };
+
+  kj::Maybe<IncompleteRead> incompleteRead;
```
kentonv, line 269: Can you add some comments here explaining what all these pieces of state represent?

It's hard to follow the body of `tryReadMessageInternal` without first understanding the starting state.

c++/src/capnp/serialize-async.c++
```diff
@@ -502,4 +506,334 @@ kj::Promise<MessageReaderAndFds> MessageStream::readMessage(
...
+      buffer = ownBuffer.asPtr();
+    }
+
+    memcpy(static_cast<void*>(buffer.begin()), static_cast<void*>(i->bytes.begin()), prefixSize);
+
+    // We want to start the read after the prefix, and so we must adjust start, minBytes, and maxBytes
+    // to account for the prefix.
+    start = buffer.asBytes().begin() + prefixSize;
+    minBytes = i->expectedSizeInWords * sizeof(word) - prefixSize;
+    maxBytes = bufferSizeInWords * sizeof(word) - prefixSize;
```
kentonv, line 636: Why `- prefixSize` here? Don't we want to fill the buffer if we can?

kentonv (commented):
Sorry for the slow review, somehow I keep not seeing this in my inbox.

c++/src/capnp/serialize-async.c++
```diff
@@ -502,4 +506,334 @@ kj::Promise<MessageReaderAndFds> MessageStream::readMessage(
...
+  } else {
+    if (scratchSpace.size() >= options.defaultReadSizeInWords) {
+      buffer = scratchSpace;
+    } else {
+      ownBuffer = kj::heapArray<word>(options.defaultReadSizeInWords);
```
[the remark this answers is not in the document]
kentonv: > better for processor cache hit rate

Copying all the messages out of the read buffer into separate message objects proactively seems worse for the cache to me. In this case, those separate objects will be cold in cache by the time you get to actually processing them. And since those objects are not contiguous chunks of memory, but rather a few different C++ objects malloc()ed in different places, there would be more cache misses in processing them.

That said, performance at this level is a mysterious thing and hard to predict confidently. We'd need to test to be sure.

> What if we returned MessageReaders that pointed into the read buffer, and said that any call to tryReadMessage would then invalidate any previously returned MessageReaders

Hmm. Well, this would make things slower if you _weren't_ using buffered reads, as the copies are unnecessary in that case. It also potentially makes things slower in the case of a `Call` or `Return` message that is much larger than the buffer; in this case, presumably the `MessageReader` would already read directly into a buffer dedicated to the message, and so making a copy later is not necessary.

One hacky optimization we could do here -- without an interface change -- is we could say that if a single message in the buffer occupies more than n% of the buffer, we're willing to transfer ownership of the buffer to it (or refcounted, if there is more than one), rather than make a copy. This wastes a bit of memory but saves copies of larger messages. Small messages, meanwhile, are probably not a big deal to copy.

With that said, I could be convinced to accept an interface change if it shows big wins. But I think I'd prefer a change where the RPC system has a way to request ownership of the message, and the MessageReader then decides whether or not to make a copy, as appropriate for its implementation.

c++/src/kj/buffering.c++
```diff
@@ -0,0 +1,324 @@
+#include "buffering.h"
+#include <kj/debug.h>
+
+namespace kj {
+
+namespace _ {
+  static inline uint lg(uint value) {
```
bcaimano, line 7: nit: Would you please use a more informative name for this function?

c++/src/kj/buffering.h
```diff
@@ -0,0 +1,203 @@
...
+  // A handle to access a partial range of bytes in an `FdRangeInputBuffer`, that may be used to
+  // retain the range for long-term ownership.
+
+public:
+  KJ_DISALLOW_COPY(RetainableFdBufferRange);
+
+  ArrayPtr<byte> getBytes();
+  ArrayPtr<AutoCloseFd> getFds();
+
+  void retain();
```
bcaimano, line 30: nit: `assertOwnership()` might be more descriptive?

c++/src/kj/buffering.c++
```diff
@@ -0,0 +1,324 @@
...
+      fds(mv(fds)) {}
+
+ArrayPtr<byte> RetainableFdBufferRange::getBytes() { return bytes; }
+ArrayPtr<AutoCloseFd> RetainableFdBufferRange::getFds() { return fds; }
+
+void RetainableFdBufferRange::retain() {
+  KJ_SWITCH_ONEOF(parentOrRetainedBytes) {
+    KJ_CASE_ONEOF(parent, FdRangeInputBuffer*) {
+      Array<byte> ownBytes;
+      if (bytes.size() > parent->bytes.size() * 0.8) {
```
bcaimano, line 41: Hmmm, I'd take this `0.8` ratio as an option. In practice, this will probably be a useful knob. That said, can you make the parent responsible for this decision instead of this class?

c++/src/kj/buffering.h
```diff
@@ -0,0 +1,203 @@
...
+  // `amount` bytes are read, such that any FDs read after `amount` will be part of a new FD range.
+
+ SplittableFdBufferRange(Badge<FdRangeInputBuffer>,
+      OneOf<FdRangeInputBuffer*, Own<FdRangeInputBuffer>> parentOrOverflow, ArrayPtr<byte> bytes);
+
+private:
+  friend class FdRangeInputBuffer;
+
+  struct Empty{}; // Used for when an overflow buffer is split away;
+  OneOf<FdRangeInputBuffer*, Own<FdRangeInputBuffer>, Empty> parentOrOverflow;
```
bcaimano, line 98: Would you please make named structs for these `OneOf` options? I understand that these types aren't actually convertable to each other, but they feel dangerously close.

c++/src/kj/buffering.h
```diff
@@ -0,0 +1,203 @@
...
+  struct FdRange {
+    ArrayPtr<byte> bytes;
+    Array<AutoCloseFd> fds;
+  };
+  struct TwoFdRange {
+    FdRange one;
+    FdRange two;
+  };
+
+  OneOf<Empty, FdRange, TwoFdRange> state = Empty{};
```
bcaimano, line 195: I don't think we care how many FDs are in the range as long as we have an index of each range. I'm picturing a bunch of small messages with a fd each. Would this work with a vector of FdRange instead?

c++/src/kj/buffering.h
bcaimano, line 98: So I'm not absolutely certain I understand this, but I'm not in love with the concept of an overflow buffer. Could we do something like this:
- Truncate our current buffer to the end of the last finished message/read.
- Allocate a new buffer, copy everything after the last finished message/read and put it in a list of buffers along with the previous one.
- Continue using the tail buffer in the list.

bcaimano (commented):
Okay, I've reviewed through the non-test code once. I very much suspect this is the sort of PR where I'm gonna have to take a second look. The general concept seems reasonable.

c++/src/capnp/serialize-async.c++
```diff
@@ -541,4 +545,181 @@ kj::Promise<MessageReaderAndFds> MessageStream::readMessage(
...
+// * on read, to use for overflow space
+//   This seems like it's of dubious usefulness
+// * on retain -- this would allow allocating space inside of the rpc call context to hold small messages
+//   This would be useful, and hits the more common case
+//
+
+// Returned kj::Own<MessageReader>s need to keep the buffer around
+// "retaining" a message
+
+class BufferedAsyncMessageStream final : public BufferedMessageStream {
```
bcaimano, line 565: I'm guessing this lives in serialize-async.c++ because buffering.c++ is in kj, right? Just checking because it feels natural to have it next to the buffering implementation it uses.

bcaimano, line 565: I'd highly recommend templating this class on StreamT. We already have private implementation, we can simply instantiate two different specializations in the two factory functions.

c++/src/capnp/serialize-async.c++
```diff
@@ -541,4 +545,181 @@ kj::Promise<MessageReaderAndFds> MessageStream::readMessage(
...
+      // Incomplete message, finish it
+      return range->appendFromStream(stream, expectedSizeInBytes - bytes.size())
+          .then([stream, range = kj::mv(range)]() mutable {
+        return finishMessage(stream, kj::mv(range));
+      });
+    }
+  }
+}
+
+kj::Promise<kj::Maybe<MessageReaderAndFds>> BufferedAsyncMessageStream::tryReadMessage(
```
bcaimano, line 637: This is pretty much the only function that doesn't delegate to the underlying stream. I think we should be able to do something like `return buffer.tryReadMessage(...);`. I'd guess you don't do this because the `FdRangeInputBuffer` is in kj and that would require knowing about `capnp::MessageReaderAndFds`. I really think it's worth moving `buffering.c++` to capnp if so.

I think it's worth a lot to be able to view the entire read message implementation in one place instead of splitting it up like this.

c++/src/kj/buffering.h
```diff
@@ -0,0 +1,202 @@
...
+  // It is assumed that when two FD ranges are buffered, once FDs are associated with the second range,
+  // then a complete "message" is present in some prefix of the buffer, and this prefix is >= the
+  // size of the first FD range. The caller must split this prefix out of the buffer
+  // before attempting to read any more bytes into the buffer.
+
+public:
+  KJ_DISALLOW_COPY(FdRangeInputBuffer);
+  FdRangeInputBuffer(size_t size, size_t maxFds);
+
+  OneOf<EmptyFdBufferRange, Own<SplittableFdBufferRange>> getRange();
```
bcaimano, line 173: I'd like us to reconceptualize this as a generator class. Essentially, we should be able to say `getNextMessageRange()` without making the caller deal with a `OneOf`. We should be able to do the split for callers and not have a tightly coupled relationship with downstream code. Put another way, this doesn't have to be isomorphic to the input end of `AsyncIoMessageStream` but I do think it is an async input stream of some sort.

c++/src/kj/buffering.c++
```diff
@@ -0,0 +1,350 @@
...
+    Own<void> ownBytes, ArrayPtr<byte> bytes, Array<AutoCloseFd> fds)
+    : parentOrRetainedBytes(mv(ownBytes)),
+      bytes(bytes),
+      fds(mv(fds)) {}
+
+ArrayPtr<byte> RetainableFdBufferRange::getBytes() { return bytes; }
+ArrayPtr<AutoCloseFd> RetainableFdBufferRange::getFds() { return fds; }
+
+void RetainableFdBufferRange::retain() {
+  KJ_SWITCH_ONEOF(parentOrRetainedBytes) {
```
bcaimano, line 38: Okay, upon review, I still feel like there's a world where we can avoid large copies altogether. I notice we're using `FlatArrayMessageReader` in serialize-async.c++ but we don't have to be. If you read a message into two refcounted buffers, you can stitch them together into one message reader. (You might need to copy one segment of the message into a smaller "transition" buffer.)

I'm pushing on this because both allocating a large buffer and copying a large amount of data can be a "stop the world" moment for low level IO code. Anything we do to reduce that will be worth it.

bcaimano (commented):
Alright, did a second pass with some note taking.

The way I see it, we have three pieces here:
- The actual MessageStream implementation, this should be entirely a passthru to other types in my opinion.
- The wrapper around an input stream that chunks up into byte + fd units, this is essentially an async input stream to me and I'd like the semantics to reflect that.
- The buffer management code that is used both inside the async input stream and the message reader/stream, this is where the fun stuff is.

I am concerned at how tightly coupled the code is while being spread across a collection of types. I suspect that we can only have one or two complicated types and a bunch of less coupled wrappers and have commented to that effect in appropriate places. I also strongly suspect that we can be more efficient about our buffer allocation and copying.

c++/src/kj/buffering.h
```diff
@@ -0,0 +1,237 @@
...
+  void handleReadResult(size_t byteCount, Array<AutoCloseFd> fds);
+
+  Own<RetainableRange> split(size_t splitAt);
+
+  friend class ::kj::RetainableRange;
+};
+
+} // namespace _
+
+template<typename Stream>
```
kentonv, line 76: I don't think this should be a template. It looks like the purpose is just to support AsyncInputStream + AsyncCapabilityStream. I would recommend using a `OneOf` or `Maybe<AsyncCapabilityStream&>` instead. Using a template for this is overkill: it's quite expensive in terms of code bloat (which often translates into poor performance due to instruction cache inefficiency) and templated code is generally harder to work with (e.g. since the IDE doesn't know what the types are, compiler errors are confusing, etc.).

c++/src/kj/buffering.c++
```diff
@@ -0,0 +1,248 @@
...
+
+  auto bytes = getByteRange();
+
+  auto spaceNeeded = bytes.size() + min;
+
+  if (spaceNeeded > allBytes.size()) {
+    // We need an overflow buffer. Allocate one (sized to the next power of 2 after spaceNeeded),
+    // copy our buffer into it, and empty the state of our old parent buffer.
+
+    auto overflow = heapArray<byte>(nextPowerOf2(spaceNeeded));
```
kentonv, line 74: It looks like you are growing the buffer to accommodate the largest single message it ever receives (plus rounding that up to a power of 2).

I don't think that's the right approach. If a 64MB message is received once you might end up with a 64MB buffer long-term. Though I guess more likely any 64MB message is likely to be retained, so you'd end up transferring away ownership of the buffer (hopefully), though then the power-of-2 round-up is wasted.

I would suggest instead that the buffer size should stay static, and whenever a message larger than the buffer is read, you immediately allocate a new buffer specifically for that message and read into that. Then you don't need to do anything if the message is retained later, as it already owns the buffer.

The benefit of buffered reads is when we can read many _small_ messages at once. It's better to bypass the buffer when reading a very large message.

kentonv (commented):
This isn't a complete review, but I wanted to post the comments I have so far.

But, please hold off on doing anything on this PR for a bit. I'm worried that we're playing "Bring me a rock" here. I am going to dig into this a bit next week and try to get a better handle on where we are going with this.

kentonv:
Superseded by #1572.

PR #1530, opened 2022-08-25
Changes shouldResolveBeforeRedirecting default to false in membranes.

dom96:
As discussed with Kenton, the `shouldResolveBeforeRedirecting` feature shouldn't be enabled by default. This is a simple PR to flip the default. It is a breaking change but it's unlikely to be used in the wild and I am preparing a PR for sandstorm to ensure things continue working there.

### DIFF
--- a/c++/src/capnp/membrane-test.c++
+++ b/c++/src/capnp/membrane-test.c++
@@ -129,7 +129,7 @@ public:
     });
   }
   
-  bool shouldResolveBeforeRedirecting() { return true; }
+  bool shouldResolveBeforeRedirecting() override { return true; }
 
 private:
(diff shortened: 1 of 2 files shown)


kentonv:
Need to update the unit test.

c++/src/capnp/membrane-test.c++
```diff
@@ -128,6 +128,8 @@ public:
       return fork.addBranch();
     });
   }
+  
+  bool shouldResolveBeforeRedirecting() { return true; }
```
kentonv, line 132:
```suggestion
  bool shouldResolveBeforeRedirecting() override { return true; }
```

PR #1531, opened 2022-08-25
Improve description of schema mismatch/corruption errors.

dom96:
* "Message contained out-of-bounds struct pointer." indicates corruption
* "Message contains non-struct pointer where struct pointer was expected." indicates a schema mismatch

So we might as well include those in the error message to improve clarity.

### DIFF
--- a/c++/src/capnp/layout.c++
+++ b/c++/src/capnp/layout.c++
@@ -321,6 +321,13 @@ static_assert(unboundAs<size_t>(POINTERS * BITS_PER_POINTER / BITS_PER_BYTE / BY
               sizeof(WirePointer),
     "BITS_PER_POINTER is wrong.");
 
+#define OUT_OF_BOUNDS_ERROR_DETAIL \
+    "This usually indicates that " \
+    "the input data was corrupted, used a different encoding than specified (e.g. " \
+    "packed vs. non-packed), or was not a Cap'n Proto message to begin with. Note " \
+    "that this error is NOT due to a schema mismatch; the input is invalid " \
+    "regardless of schema."
+
 namespace {
 
 static const union {
@@ -578,7 +585,8 @@ struct WireHelpers {
       const word* ptr = ref->farTarget(segment);
       auto padWords = (ONE + bounded(ref->isDoubleFar())) * POINTER_SIZE_IN_WORDS;
       KJ_REQUIRE(boundsCheck(segment, ptr, padWords),
-                 "Message contains out-of-bounds far pointer.") {
+                 "Message contains out-of-bounds far pointer. "
+                 OUT_OF_BOUNDS_ERROR_DETAIL) {
         return nullptr;
       }
 
@@ -790,7 +798,8 @@ struct WireHelpers {
     switch (ref->kind()) {
       case WirePointer::STRUCT: {
         KJ_REQUIRE(boundsCheck(segment, ptr, ref->structRef.wordSize()),
-                   "Message contained out-of-bounds struct pointer.") {
+                   "Message contained out-of-bounds struct pointer. "
+                   OUT_OF_BOUNDS_ERROR_DETAIL) {
           return result;
         }
         result.addWords(ref->structRef.wordSize());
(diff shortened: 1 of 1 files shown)


c++/src/capnp/layout.c++
```diff
@@ -790,7 +790,7 @@ struct WireHelpers {
     switch (ref->kind()) {
       case WirePointer::STRUCT: {
         KJ_REQUIRE(boundsCheck(segment, ptr, ref->structRef.wordSize()),
-                   "Message contained out-of-bounds struct pointer.") {
+                   "Message contained out-of-bounds struct pointer. Is your memory corrupted?") {
```
kentonv, line 793: I think we could expand this out a bit...

```suggestion
                   "Message contained out-of-bounds struct pointer. This usually indicates that "
                   "the input data was corrupted, used a different encoding that specified (e.g. "
                   "packed vs. non-packed), or was not a Cap'n Proto message to begin with. Note "
                   "that this error is NOT due to a schema mismatch; the input is invalid "
                   "regardless of schema.") {
```

Perhaps all the text after the first sentence should be factored out into a macro.

c++/src/capnp/layout.c++
```diff
@@ -1131,7 +1131,7 @@ struct WireHelpers {
     word* oldPtr = followFars(oldRef, refTarget, oldSegment);
 
     KJ_REQUIRE(oldRef->kind() == WirePointer::STRUCT,
-        "Message contains non-struct pointer where struct pointer was expected.") {
+        "Schema mismatch: Message contains non-struct pointer where struct pointer was expected.") {
```
kentonv, line 1134: Aren't there some other errors in this file that should be marked as "schema mismatch"? I'd expect to see e.g. the case where a list is expected but a non-list pointer is found.

dom96:
Improved far more error messages now. Hope I got them right :)

c++/src/capnp/layout.c++
```diff
@@ -578,7 +578,8 @@ struct WireHelpers {
       const word* ptr = ref->farTarget(segment);
       auto padWords = (ONE + bounded(ref->isDoubleFar())) * POINTER_SIZE_IN_WORDS;
       KJ_REQUIRE(boundsCheck(segment, ptr, padWords),
-                 "Message contains out-of-bounds far pointer.") {
+                 "Message contains out-of-bounds far pointer."
```
kentonv, line 581: Minor, but it looks like none of these messages include a space between the terminating period of the specific message and `OUT_OF_BOUNDS_ERROR_DETAIL`. E.g. this message will contain `far pointer.This usually`.

c++/src/capnp/layout.c++
```diff
@@ -701,7 +702,7 @@ struct WireHelpers {
             WirePointer* elementTag = reinterpret_cast<WirePointer*>(ptr);
 
             KJ_ASSERT(elementTag->kind() == WirePointer::STRUCT,
-                  "Don't know how to handle non-STRUCT inline composite.");
+                  "Schema mismatch: Don't know how to handle non-STRUCT inline composite.");
```
kentonv, line 705: This one isn't a schema mismatch. In fact `zeroObject()` doesn't care about schemas.

c++/src/capnp/layout.c++
```diff
@@ -841,22 +845,24 @@ struct WireHelpers {
...
+                       OUT_OF_BOUNDS_ERROR_DETAIL) {
               return result;
             }
 
             const WirePointer* elementTag = reinterpret_cast<const WirePointer*>(ptr);
             auto count = elementTag->inlineCompositeListElementCount();
 
             KJ_REQUIRE(elementTag->kind() == WirePointer::STRUCT,
-                       "Don't know how to handle non-STRUCT inline composite.") {
+                       "Schema mismatch: Don't know how to handle non-STRUCT inline composite.") {
```
kentonv, line 857: Same here, not a schema mismatch, the underlying data is not valid under any schema.

c++/src/capnp/layout.h
```diff
@@ -1261,6 +1261,13 @@ inline OrphanBuilder& OrphanBuilder::operator=(OrphanBuilder&& other) {
   return *this;
 }
 
+#define OUT_OF_BOUNDS_ERROR_DETAIL \
```
kentonv, line 1264: Let's put this in `layout.c++` instead of the header. All Cap'n Proto users inlcude this header so this pollutes the macro space, but this is only legitimately used in `layout.c++` anyway.

PR #1535, opened 2022-09-07
Prefer __is_trivially_constructible etc. intrinsics

ecatmur:
__has_trivial_constructor etc. are deprecated since Clang 15: https://releases.llvm.org/15.0.0/tools/clang/docs/ReleaseNotes.html#non-comprehensive-list-of-changes-in-this-release

> Some type-trait builtins, such as `__has_trivial_assign`, have been documented as deprecated for a while because their semantics don’t mix well with post-C++11 type-traits. Clang now emits deprecation warnings for them under the flag `-Wdeprecated-builtins`.

### DIFF
--- a/c++/src/capnp/dynamic.c++
+++ b/c++/src/capnp/dynamic.c++
@@ -1573,12 +1573,12 @@ DynamicValue::Builder::Builder(Builder& other) {
       // Unfortunately canMemcpy() doesn't work on these types due to the use of
       // DisallowConstCopy, but __has_trivial_destructor should detect if any of these types
       // become non-trivial.
-      static_assert(__has_trivial_destructor(Text::Builder) &&
-                    __has_trivial_destructor(Data::Builder) &&
-                    __has_trivial_destructor(DynamicList::Builder) &&
-                    __has_trivial_destructor(DynamicEnum) &&
-                    __has_trivial_destructor(DynamicStruct::Builder) &&
-                    __has_trivial_destructor(AnyPointer::Builder),
+      static_assert(KJ_HAS_TRIVIAL_DESTRUCTOR(Text::Builder) &&
+                    KJ_HAS_TRIVIAL_DESTRUCTOR(Data::Builder) &&
+                    KJ_HAS_TRIVIAL_DESTRUCTOR(DynamicList::Builder) &&
+                    KJ_HAS_TRIVIAL_DESTRUCTOR(DynamicEnum) &&
+                    KJ_HAS_TRIVIAL_DESTRUCTOR(DynamicStruct::Builder) &&
+                    KJ_HAS_TRIVIAL_DESTRUCTOR(AnyPointer::Builder),
                     "Assumptions here don't hold.");
       break;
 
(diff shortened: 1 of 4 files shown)


ecatmur:
Think I've got it right this time.

I know it looks weird using `__is_trivially_constructible` alongside `__has_trivial_destructor` but that's what libstdc++ does: https://github.com/gcc-mirror/gcc/blob/16e2427f50c208dfe07d07f18009969502c25dc8/libstdc%2B%2B-v3/include/std/type_traits#L1185

kentonv:
Is there any subtle difference between the meanings of these intrinsics? Or is it purely a rename?

We've had trouble in the past where GCC and Clang disagreed on the exact definition of an intrinsic and it resulted in differing ABIs between the compilers, which was kinda bad.

ecatmur:
Ugh, this is a hot mess. No, they aren't exact equivalents; consider:

```c++
#include <type_traits>
struct A {
    A() = default;
    ~A();
};
struct B { B(int = 0) noexcept; };
struct C { private: ~C() = default; };
char s[] = {
    '0' + __is_trivially_constructible(A),
    '0' + __has_trivial_constructor(A),
#if __clang__ or __GNUC__ >= 11
    '0' + __is_nothrow_constructible(B),
#else
    '0' + std::is_nothrow_constructible_v<B>,
#endif
    '0' + __has_nothrow_constructor(B),
#if __clang__
    '0' + __is_trivially_destructible(C),
#else
    '0' + std::is_trivially_destructible_v<C>,
#endif
    '0' + __has_trivial_destructor(C),
};
```

This outputs `"011001"` on clang, but `"011101"` on gcc. So, this PR:
- alters behavior (between versions) for types like `A` that are not trivially destructible (probably not an ABI break, since HeapArrayDisposer::Allocate_ is private and its methods should get inlined)
- makes clang more consistent with gcc for types like `B` that have noexcept default ctors with default arguments (this is a good thing?)
- breaks compatibility for types like `C` that have inaccessible/deleted trivial dtors (but they probably don't work with capnp/kj anyway)

So while putting in a warning suppression instead would be viable, there would still be the inconsistency between clang and gcc over `B`. In other words, just because clang and gcc use the same name for an intrinsic doesn't mean that the behavior is the same; sometimes you need to use differently-named intrinsics to get the same behavior.

btw, libstdc++'s implementation of `std::is_$foo_destructible` is utterly horrible, gcc doesn't even have an intrinsic for `__is_destructible` so they use sfinae. https://github.com/gcc-mirror/gcc/blob/16e2427f50c208dfe07d07f18009969502c25dc8/libstdc%2B%2B-v3/include/std/type_traits#L806

kentonv:
Hmm so the ABI problem we've had before is described here:

https://github.com/capnproto/capnproto/blob/master/c%2B%2B/src/kj/common.h#L485

But I don't see anything in your PR that looks like it would cause an ABI incompatibility if compilers disagree on the intrinsic. So I think I'm going to merge and hope for the best. :)

It turns out we just started seeing these deprecation warnings in our own builds so, nice timing I guess. Thanks!

PR #1536, opened 2022-09-08
use size when converting between std::string and kj::String

vaci:
Avoids a call to ::strlen(), but will now copy any embedded null characters and beyond.


kentonv:
Thanks! Not sure why I didn't do it that way in the first place.

PR #1537, opened 2022-09-08
HttpServer: Avoid flushing broken streams when handling errors

jclee:
Fixes a race condition where a client disconnection could cause a "failed" exception to be thrown after its "disconnected" exception had already been handled.

### DIFF
--- a/c++/src/kj/compat/http-test.c++
+++ b/c++/src/kj/compat/http-test.c++
@@ -5527,6 +5527,151 @@ KJ_TEST("HttpServer.listenHttp() doesn't prematurely terminate if an accepted co
   KJ_ASSERT(!promise.poll(waitScope));
 }
 
+KJ_TEST("HttpServer handles disconnected exception for clients disconnecting after headers") {
+  // This test case reproduces a race condition where a client could disconnect after the server
+  // sent response headers but before it sent the response body, resulting in a broken pipe
+  // "disconnected" exception when writing the body.  The default handler for application errors
+  // tells the server to ignore "disconnected" exceptions and close the connection, but code
+  // after the handler exercised the broken connection, causing the server loop to instead fail
+  // with a "failed" exception.
+
+  KJ_HTTP_TEST_SETUP_IO;
+  kj::TimerImpl timer(kj::origin<kj::TimePoint>());
+  auto pipe = KJ_HTTP_TEST_CREATE_2PIPE;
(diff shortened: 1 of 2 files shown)


c++/src/kj/compat/http.c++
```diff
@@ -6906,19 +6904,27 @@ private:
...
-                  .then([]() { return false; });  // loop ends after flush
+  kj::Promise<bool> finishSendingError(kj::Promise<void> promise) {
+    return promise.then([this]() -> kj::Promise<void> {
+      if (httpOutput.isBroken()) {
+        // Skip flush for broken streams, since it will throw an exception that may be worse than
+        // the one we just handled.
+        return kj::READY_NOW;
+      } else {
+        return httpOutput.flush();
+      }
```
jclee, line 6926: I'm not sure if this is necessarily the correct solution -- there are a lot of other ways to suppress the error, and doing it this way may hide stream errors originating in the handlers themselves.

But I do notice that there a couple other places in `HttpServer::Connection` that are also using the pattern of skipping the `flush()` call if `isBroken()` is true, so I assume it's at least a reasonable thing to do in particular situations.

c++/src/kj/compat/http-test.c++
```diff
@@ -5527,6 +5527,151 @@ KJ_TEST("HttpServer.listenHttp() doesn't prematurely terminate if an accepted co
...
+    Promise<uint64_t> pumpTo(AsyncOutputStream& output, uint64_t amount) override {
+      return inner.pumpTo(output, amount);
+    }
+
+    Promise<void> write(const void* buffer, size_t size) override {
+      int writeId = writeCount++;
+      if (writeId == 0) {
+        // Allow first write (headers) to succeed.
+        auto promise = inner.write(buffer, size);
+        inner.shutdownWrite();
```
jclee, line 5578: I don't fully understand why, but it seemed that this call to `shutdownWrite()` was necessary for the race condition failure to manifest in `http-test`.  Without it, the failure only manifested when run as part of `http-socketpair-test`.  I assume there's some aspect of native pipes that makes it necessary.

c++/src/kj/compat/http.c++
kentonv: This seem pretty reasonable to me. I think this is a place where there isn't any perfect answer.

PR #1540, opened 2022-09-15
Add `getPreferredExtensions()` to WebSocket class

MellowYarker:
As mentioned in the PR to enable compression in Workers, I suspect we don't need to extract the extensions from the websocket. Happy to do that if I'm mistaken, though in that case I may have a follow up question regarding the fact that a KJ websocket is unaware whether it's on the client/server side. This makes it a bit tricky to determine if `inbound` maps to client or server (unless we know for certain whether our websocket is client or server side).

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -2637,6 +2637,54 @@ public:
 
   uint64_t receivedByteCount() override { return receivedBytes; }
 
+  kj::Maybe<kj::String> getPreferredExtensions(ExtensionsContext ctx) override {
+     if (maskKeyGenerator == nullptr) {
+      // `this` is the server side of a websocket.
+      if (ctx == ExtensionsContext::REQUEST) {
+        // The other WebSocket is (going to be) the client side of a WebSocket, i.e. this is a
+        // proxying pass-through scenario. Optimization is possible. Confusingly, we have to use
+        // generateExtensionResponse() (even though we're generating headers to be passed in a
+        // request) because this is the function that correctly maps our config's inbound/outbound
+        // to client/server.
+        KJ_IF_MAYBE(c, compressionConfig) {
+          return _::generateExtensionResponse(*c);
+        } else {
+          return kj::String(nullptr);  // recommend no compression
+        }
+      } else {
+        // We're apparently arranging to pump from the server side of one WebSocket to the server
+        // side of another; i.e., we are a server, we have two clients, and we're trying to pump
+        // between them. We cannot optimize this case, because the masking requirements are
+        // different for client->server vs. server->client messages. Since we have to parse out
+        // the messages anyway there's no point in trying to match extensions, so return null.
+        return nullptr;
+      }
+    } else {
+      // `this` is the client side of a websocket.
+      if (ctx == ExtensionsContext::RESPONSE) {
+        // The other WebSocket is (going to be) the server side of a WebSocket, i.e. this is a
+        // proxying pass-through scenario. Optimization is possible. Confusingly, we have to use
+        // generateExtensionRequest() (even though we're generating headers to be passed in a
+        // response) because this is the function that correctly maps our config's inbound/outbound
+        // to server/client.
+        KJ_IF_MAYBE(c, compressionConfig) {
+          CompressionParameters arr[1]{*c};
+          return _::generateExtensionRequest(arr);
+        } else {
+          return kj::String(nullptr);  // recommend no compression
+        }
+      } else {
+        // We're apparently arranging to pump from the client side of one WebSocket to the client
+        // side of another; i.e., we are a client, we are connected to two servers, and we're
+        // trying to pump between them. We cannot optimize this case, because the masking
+        // requirements are different for client->server vs. server->client messages. Since we have
+        // to parse out the messages anyway there's no point in trying to match extensions, so
+        // return null.
+        return nullptr;
+      }
+    }
+  }
+
 private:
   class Mask {
   public:
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -2638,7 +2638,7 @@ public:
   uint64_t receivedByteCount() override { return receivedBytes; }
 
   kj::Maybe<kj::String> getPreferredExtensions(ExtensionsContext ctx) override {
-     if (maskKeyGenerator == nullptr) {
+    if (maskKeyGenerator == nullptr) {
       // `this` is the server side of a websocket.
       if (ctx == ExtensionsContext::REQUEST) {
         // The other WebSocket is (going to be) the client side of a WebSocket, i.e. this is a
--- a/c++/src/kj/compat/http.h
+++ b/c++/src/kj/compat/http.h
@@ -655,10 +655,9 @@ public:
   // certain extensions (e.g. compression settings), then this method returns what those extensions
   // are. For example, matching extensions between standard WebSockets allows pumping to be
   // implemented by pumping raw bytes between network connections, without reading individual frames.
-
+  //
   // A null return value indicates that there is no preference. A non-null return value containing
   // an empty string indicates a preference for no extensions to be applied.
-
 };
 
 class HttpClient {
--- a/c++/src/kj/compat/http.h
+++ b/c++/src/kj/compat/http.h
@@ -643,6 +643,22 @@ public:
 
   virtual uint64_t sentByteCount() = 0;
   virtual uint64_t receivedByteCount() = 0;
+
+  enum ExtensionsContext {
+    // Indicate whether a Sec-WebSocket-Extension header should be rendered for use in request
+    // headers or response headers.
+    REQUEST,
+    RESPONSE
+  };
+  virtual kj::Maybe<kj::String> getPreferredExtensions(ExtensionsContext ctx) { return nullptr; }
+  // If pumpTo() / tryPumpFrom() is able to be optimized only if the other WebSocket is using
+  // certain extensions (e.g. compression settings), then this method returns what those extensions
+  // are. For example, matching extensions between standard WebSockets allows pumping to be
+  // implemented by pumping raw bytes between network connections, without reading individual frames.
+
+  // A null return value indicates that there is no preference. A non-null return value containing
+  // an empty string indicates a preference for no extensions to be applied.
+
 };
 
 class HttpClient {
--- a/c++/src/kj/compat/http-test.c++
+++ b/c++/src/kj/compat/http-test.c++
@@ -2391,6 +2391,51 @@ void testWebSocketTwoMessageCompression(kj::WaitScope& waitScope, HttpHeaderTabl
 #endif // KJ_HAS_ZLIB
 
 #if KJ_HAS_ZLIB
+void testWebSocketOptimizePumpProxy(kj::WaitScope& waitScope, HttpHeaderTable& headerTable,
+                                    kj::HttpHeaderId extHeader, kj::StringPtr extensions,
+                                    HttpClient& client) {
+  // Suppose we are proxying a websocket conversation between a client and a server.
+  // This looks something like: CLIENT <--> (proxyServer <==PUMP==> proxyClient) <--> SERVER
+  //
+  // We want to enable optimizedPumping from the proxy's server (which communicates with the client),
+  // to the proxy's client (which communicates with the origin server).
+  //
+  // For this to work, proxyServer's inbound settings must map to proxyClient's outbound settings
+  // (and vice versa). In this case, `ws` is `proxyClient`, so we want to take `ws`'s compression
+  // configuration and pass it to `proxyServer` in a way that would allow for optimizedPumping.
+
+  kj::HttpHeaders headers(headerTable);
+  headers.set(extHeader, extensions);
+  auto response = client.openWebSocket("/websocket", headers).wait(waitScope);
+
+  KJ_EXPECT(response.statusCode == 101);
+  KJ_EXPECT(response.statusText == "Switching Protocols", response.statusText);
+  KJ_EXPECT(KJ_ASSERT_NONNULL(response.headers->get(extHeader)).startsWith("permessage-deflate"));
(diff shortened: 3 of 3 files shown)


kentonv:
What we really need here is a method that returns the exact settings that should be used to enable optimized pumping. Simply knowing whether the WebSocket is `WebSocketImpl` is not sufficient. In fact, we technically don't care what implementation it is at all -- the real question we're trying to answer is "what settings should I use on another WebSocketImpl in order to enable optimized pumping from this one".

c++/src/kj/compat/http.h
```diff
@@ -643,6 +643,7 @@ class WebSocket {
 
   virtual uint64_t sentByteCount() = 0;
   virtual uint64_t receivedByteCount() = 0;
+  virtual kj::Maybe<kj::String> getPreferredExtensions() { return nullptr; }
```
kentonv, line 646: This needs some doc comments:

> If pumpTo() / tryPumpFrom() is able to be optimized only if the other WebSocket is using certain extensions (e.g. compression settings), then this method returns what those extensions are. For example, matching extensions between standard WebSockets allows pumping to be implemented by pumping raw bytes between network connections, without reading individual frames.
>
> A null return value indicates that there is no preference. A non-null return value containing an empty string indicates a preference for no extensions to be applied.

And now I see the point I think you were trying to make about distinguishing the outgoing and incoming directions. Technically we could end up pumping between two client-side or two server-side WebSockets. Those cases actually already can't support optimized pumping anyway, at least at present, due to the masking requirements that differ in the two directions.

Perhaps we should add to the comment:

> The extensions are rendered assuming this WebSocket represents the client side of a connection to a remote server. A proxy service could then use this string to fill in the `Sec-WebSocket-Extensions` header of a response to its own client. Calling this method on the server side of a WebSocket doesn't make sense and will not return a useful result.

This is a weird wart, though. I wonder if we can do better by either:
a. Have `getPreferredExtensions()` take a parameter indicating whether the extensions should be rendered for the client->server or server->client direction.
b. Have `getPreferredExtensions()` return a structured representation of the extensions that refers to inbound/outbound rather than client/server, and leave it up to the caller to stringify this appropriately for the context.

Thougths?

c++/src/kj/compat/http.c++
```diff
@@ -2637,6 +2637,13 @@ public:
 
   uint64_t receivedByteCount() override { return receivedBytes; }
 
+  kj::Maybe<kj::String> getPreferredExtensions() override {
+    KJ_IF_MAYBE(config, compressionConfig) {
+      return _::generateExtensionResponse(*config);
+    }
+    return kj::str("");
```
kentonv, line 2644: You can avoid an allocation here with:

```suggestion
    return kj::String(nullptr);
```

c++/src/kj/compat/http.h
MellowYarker: Yeah, I was thinking it would probably be easier for the caller to specify whether it wanted `client->server` or `server->client`. This function has access to other functions that already stringify compression parameters so I'd rather use those as they've been tested.

c++/src/kj/compat/http.h
```diff
@@ -643,6 +643,20 @@ class WebSocket {
 
   virtual uint64_t sentByteCount() = 0;
   virtual uint64_t receivedByteCount() = 0;
+
+  enum Role {
```
MellowYarker, line 647: I was between `Direction` and `Role`, went with the shorter one. Open to a better name!

kentonv: Hmm, at present it's not really clear which enum value I'm supposed to use for any particular context.

Let's change it to:

```
  enum ExtensionsContext {
    // Indicate whether a Sec-WebSocket-Extension header should be rendered for use in request
    // headers or response headers.
    REQUEST,
    RESPONSE
  };
```

c++/src/kj/compat/http.h
```diff
@@ -643,6 +643,20 @@ class WebSocket {
...
+  };
+  virtual kj::Maybe<kj::String> getPreferredExtensions(Role role) { return nullptr; }
+  // If pumpTo() / tryPumpFrom() is able to be optimized only if the other WebSocket is using
+  // certain extensions (e.g. compression settings), then this method returns what those extensions
+  // are. For example, matching extensions between standard WebSockets allows pumping to be
+  // implemented by pumping raw bytes between network connections, without reading individual frames.
+
+  // A null return value indicates that there is no preference. A non-null return value containing
+  // an empty string indicates a preference for no extensions to be applied.
+
```
kentonv, line 659:
```suggestion
  //
  // A null return value indicates that there is no preference. A non-null return value containing
  // an empty string indicates a preference for no extensions to be applied.
```

MellowYarker:
Synced with master + rebased.

c++/src/kj/compat/http.c++
```diff
@@ -2637,6 +2637,18 @@ public:
 
   uint64_t receivedByteCount() override { return receivedBytes; }
 
+  kj::Maybe<kj::String> getPreferredExtensions(ExtensionsContext ctx) override {
+    KJ_IF_MAYBE(config, compressionConfig) {
+      if (ctx == ExtensionsContext::RESPONSE) {
```
kentonv, line 2642: Should we go ahead and have this method return null in the case that the context suggests a reverse of direction? E.g. pumping from one client to another, rather than proxying client-to-server? optimizedPump won't work in that case anyway so there's no reason to specify a preference.

kentonv:
Can you cover this in the unit test? Just a couple lines added to an existing test case to check the output of this method would be fine.

c++/src/kj/compat/http.c++
MellowYarker: I'm having some trouble seeing how we could know this information from within this method. I'd think the caller would need to know ahead of time whether it was trying to get the client-to-server config, or whether it was trying to pump from one client to another, no?

kentonv: You can tell based on the presence or absence of `maskKeyGenerator` whether the WebSocket object is the client side vs. the server side of a connection. See `tryPumpFrom()`, where it will refuse to do an optimized pump unless exactly one of the two sides has a `maskKeyGenerator`. Since optimized pumping will be refused in that case anyway, we should return null here.

c++/src/kj/compat/http.c++
```diff
@@ -2637,6 +2637,31 @@ public:
 
   uint64_t receivedByteCount() override { return receivedBytes; }
 
+  kj::Maybe<kj::String> getPreferredExtensions(ExtensionsContext ctx) override {
+    if ((maskKeyGenerator == nullptr && ctx == ExtensionsContext::REQUEST) ||
+        (maskKeyGenerator != nullptr && ctx == ExtensionsContext::RESPONSE)) {
```
kentonv, line 2642: I think we're confusing the meaning of REQUEST and RESPONSE here.

If `maskKeyGenerator` is non-null, then this WebSocket object is the client side of a connection (since only the client needs to apply a mask to outgoing messages).

Meanwhile, if the caller passed `RESPONSE`, that means they are requesting to construct an extension header for use in a response, i.e. to create the server side of a connection.

That's the case we want to support -- coupling a client-side and a server-side WebSocket. But it looks like the code here will return null in that case?

Or am I reading this wrong?

c++/src/kj/compat/http.c++
```diff
@@ -2637,6 +2637,31 @@ public:
...
+      // We can only direct pump from a client to a server (and vice versa).
+      // If the caller attempts to get preferred extensions for pumping from client to client,
+      // or from server to server, then we will return nullptr, indicating no preferred extensions.
+      return nullptr;
+    }
+    KJ_IF_MAYBE(config, compressionConfig) {
+      if (ctx == ExtensionsContext::RESPONSE) {
+        // We are a server websocket, so our
```
kentonv, line 2650: This comment is a little confusing. The caller is asking to construct the headers for a new server-side websocket, buht that doesn't mean "we" are a server websocket -- in fact, "we" must be a client websocket for this to work?

c++/src/kj/compat/http.c++
MellowYarker: Right, sorry, got mixed up thinking about what the caller wants vs. what we want to extract from `this` websocket. It felt weird that `ctx` being `REQUEST` would result in us calling `generateExtensionResponse()`.

Just for posterity, what we want is:
**Caller**: "I want to pump from `this` to `other`, where `other` will be returned in a `RESPONSE`.
**Callee**:
1. Make sure that `this` is a client (non-null `maskKeyGenerator`), as `other` is a server-side connection.
2. Get `this` websocket's extensions. Since we're a client, `out->client`, so use `generateExtensionRequest`.
3. Return `permessage-deflate; ...`

**Caller**: Interprets `permessage-deflate; ...` parameters in the opposite direction, i.e. `out->server`, so we can `optimizedPump`.

kentonv: > It felt weird that ctx being REQUEST would result in us calling generateExtensionResponse().

Hmm, is that right? `generateExtensionResponse()` formats a response header, right? So that's what we should use if the client requests RESPONSE?

I think it was just this check that was backwards here, but the code below was correct originally (aside from the comment being confusing)?

MellowYarker: > `generateExtensionResponse()` formats a response header, right?

Yes, but more specifically it assumes that the `config` passed to it belongs to a server-side websocket. It assumes this because we map `inbound` parameters to `client`, and `outbound` to `server`.

> So that's what we should use if the client requests RESPONSE?

Bit unclear here, is `this` websocket a "client", i.e. we want to pump to a server, or did you mean "caller" in place of "client"?

---
TL;DR of the stuff below is: when the caller passes `RESPONSE` to `getPreferredExtensions()`, is the caller saying: 
```
a) I intend to pump from `this` websocket to a server-side websocket (implies `this` must be a client), or
b) I want `this` websocket's config to be parsed as though `this` websocket is server-side.
```
---
(This is a bit long, but I want to be clear so it's easier to find where the confusion is).
I think the main goal here is that we have some websocket, and we want to convert its `inboundX` parameters to `clientX` or `serverX`. Let's say we have a client-side websocket `proxyClient`, which has the following compression config:
```
CompressionParameters {
    outboundNoContextTakeover = false,
    inboundNoContextTakeover = true,
    outboundMaxWindowBits = 10,
    inboundMaxWindowBits = 13
};
```
We would want this to be converted to: 
```
permessage-deflate; server_no_context_takeover; client_max_window_bits=10; server_max_window_bits=13
```
since `outbound->client` and `inbound->server` for client-side websockets.

Now let's say we want to have another websocket, `proxyServer`, which is a server-side websocket. We want to have the final connection look like this:
```
eyeball <-----> proxyServer <==optimizedPumpTo==> proxyClient <-----> server

     outboundNoContextTakeover = true          outboundNoContextTakeover = false
     inboundNoContextTakeover = false          inboundNoContextTakeover = true
     outboundMaxWindowBits = 13                outboundMaxWindowBits = 10
     inboundMaxWindowBits = 10                 inboundMaxWindowBits = 13
```

To enable `optimizedPumpTo`, we need to ensure `proxyServer` uses extension parameters that are compatible with `proxyClient`. We would do this by passing the exact header value `permessage-deflate; server_no_context_takeover; client_max_window_bits=10; server_max_window_bits=13` to `acceptWebSocket`, because `acceptWebSocket` will map `server->outbound` and `client->inbound` for us.

**This is probably where the confusion is**:
As I see the _current_ code, calling `proxyClient.getPreferredExtensions(RESPONSE)` is the same as saying
> I want to convert `proxyClient`s compression config to a `Sec-WebSocket-Extensions` header, and I will use this header to create a websocket (`proxyServer`) that will act as a server-side websocket (used in a `RESPONSE`) that can `optimizedPumpTo` `proxyClient`.

This is based on what you said above:
> Meanwhile, if the caller passed `RESPONSE`, that means they are requesting to construct an extension header for use in a response, i.e. to create the server side of a connection.

Did you mean passing `RESPONSE` tells `getPreferredExtensions` to generate extensions as though `this` websocket is server-side?

So all things considered, `proxyClient.getPreferredExtensions(RESPONSE)` results in us converting `proxyClient`'s `CompressionParameters` to `permessage-deflate; server_no_context_takeover; client_max_window_bits=10; server_max_window_bits=13` (this is done by calling `generateExtensionRequest`, since `proxyClient` is client-side). We then pass this header value to `acceptWebSocket()`, which gives us `proxyServer`, which has the following `CompressionParameters`:
```
CompressionParameters {
    outboundNoContextTakeover = true,
    inboundNoContextTakeover = false,
    outboundMaxWindowBits = 13,
    inboundMaxWindowBits = 10
};
```
Which is the opposite of `proxyClient` (which is what we want).

kentonv: > Yes, but more specifically it assumes that the config passed to it belongs to a server-side websocket. It assumes this because we map inbound parameters to client, and outbound to server.

Ohhhhhh! I get it.

c++/src/kj/compat/http.c++
```diff
@@ -2637,6 +2637,31 @@ public:
...
+        return _::generateExtensionResponse(*config);
+      } else {
+        // `this` is a client websocket, so our
+        // - "inbound" maps to "server", and our
+        // - "outbound" maps to "client".
+        CompressionParameters arr[1]{*config};
+        return _::generateExtensionRequest(arr);
+      }
+    }
+    return kj::String(nullptr);
```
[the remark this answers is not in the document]
MellowYarker: Cool, I'm happy with this, really clears up what's going on. Apologizes for the confusion earlier!

c++/src/kj/compat/http.c++
```diff
@@ -2637,6 +2637,54 @@ public:
 
   uint64_t receivedByteCount() override { return receivedBytes; }
 
+  kj::Maybe<kj::String> getPreferredExtensions(ExtensionsContext ctx) override {
+     if (maskKeyGenerator == nullptr) {
```
kentonv, line 2641:
```suggestion
    if (maskKeyGenerator == nullptr) {
```

kentonv:
Sorry this took so long. Seems it's more subtle than it looks!

PR #1541, opened 2022-09-15
llvm_symbolizer address mode

mikea:
Introducing new TRY_LLVM_SYMBOLIZER stack trace mode. In this mode, if LLVM_SYMBOLIZER env var is defined, will use its value as a path to the llvm-symbolizer tool, fork and exec it to symbolize the stack trace.

### DIFF
--- a/c++/src/kj/exception.c++
+++ b/c++/src/kj/exception.c++
@@ -48,6 +48,7 @@
 #include <stdint.h>
 #ifndef _WIN32
 #include <sys/mman.h>
+#include <sys/wait.h>
 #endif
 #include "io.h"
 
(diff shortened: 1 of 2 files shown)


mikea:
I'm not entirely sure how to trigger this mode, so that there's a realistic chance of us enabling it in the runtime. 
I have introduced new `TRY_LLVM_SYMBOLIZER` mode since this is the best idea I had.

Should I write a test that depends on LLVM_SYMBOLIZER var too to test this behavior? @kentonv

c++/src/kj/exception.c++
```diff
@@ -254,9 +255,138 @@ ArrayPtr<void* const> getStackTrace(ArrayPtr<void*> space, uint ignoreCount) {
...
+  }
+}
+#endif
+
 String stringifyStackTrace(ArrayPtr<void* const> trace) {
   if (trace.size() == 0) return nullptr;
-  if (getExceptionCallback().stackTraceMode() != ExceptionCallback::StackTraceMode::FULL) {
+  auto stackTraceMode = getExceptionCallback().stackTraceMode();
+#if (__linux__ || __APPLE__ || __CYGWIN__) && !__ANDROID__
+  if (stackTraceMode == ExceptionCallback::StackTraceMode::TRY_LLVM_SYMBOLIZER) {
```
kentonv, line 385: Hmm. If it's necessary to set a custom stack trace mode anyway (requiring code added to programs that use this), maybe we should instead provide a way to specify a callback to symbolize stack traces, rather than add the llvm-symbolizer-specific logic to KJ?

Perhaps we could even use that rust addr2line library in-process rather than shelling out.

It would be especially nice if we could make it so simply linking against that library will cause KJ to use it, without any other code changes needed in the binary. That way we can link it into all our KJ tests...

(Another possible path is to automatically assume `TRY_LLVM_SYMBOLIZER` here but I worry there are security implications to us happily invoking whatever binary happens to be specified by some environment variable...)

mikea:
not needed after https://github.com/capnproto/capnproto/commit/46f54a2c027b6da987fcbe8e6971cf14a84213cc

PR #1543, opened 2022-09-18
Allow file IDs to be omitted when using `capnp eval`.

kentonv:
Assigning file IDs is a pointless exercise when using capnp files as configuration. Typically this involve files that don't even define any types, just constants. The ID does not affect the output of `capnp eval` in any way. So, we should not require file IDs to be present in this case.

Also expose this option on `SchemaParser`, for applications which use that class to parse config.


kentonv:
@jasnell approved the downstream PR; I'm assuming that implies approval of this.

PR #1545, opened 2022-09-20
[bazel] Add missing async-coroutine-test -> kj-test dependency

harrishancock:
Without this, I got a bunch of linker errors.


mikea (approved):
I think the title is slightly incorrect. Otherwise LGTM.

PR #1546, opened 2022-09-20
Implement asynchronous environments

harrishancock:
NOTE: This is a working rough draft, but it contains some hacks that I'm not super happy about, specifically around making sure user destructors are run under environments. Nevertheless, review feedback would be useful at this point.

This commit adds a new concept to the kj-async library: Environments.

Environments are ambiently available storage, of a type specified by the user, which "follow" user callbacks scheduled via functions/syntax like evalLater(), promise.then(), startFiber(), co_await, and more.

Environments do not cross threads. Type safety is as safe as kj::downcast<T>() makes it -- in debug mode, you'll probably get an assertion if you set an environment of type T and try to get it later as type U -- but in release mode you're on your own.

### DIFF
--- a/c++/src/kj/async-inl.h
+++ b/c++/src/kj/async-inl.h
@@ -342,11 +342,11 @@ public:
 
 }  // namespace _
 
-template <typename E, typename T, typename... Args>
-Promise<T> runInEnvironment(E&& environment, FunctionParam<Promise<T>(Args...)> func) {
+template <typename E, typename Func>
+PromiseForResult<Func, void> runInEnvironment(E&& environment, Func&& func) {
   auto holder = refcounted<_::Environment::Holder<Decay<E>>>(fwd<E>(environment));
   _::Environment::Scope scope(*holder);
-  return evalNow(func);
+  return evalNow(fwd<Func>(func));
 }
 
 template <typename T>
--- a/c++/src/kj/async-inl.h
+++ b/c++/src/kj/async-inl.h
@@ -300,35 +300,66 @@ protected:
 
 // -------------------------------------------------------------------
 
-class Environment: public kj::Refcounted {
-  // Base class of Environment::Holder<T>. This class also provides access to the current
-  // EventLoop's reference to the currently-active Environment.
+class EnvironmentSet {
+  // A collection of Environments.
+  //
+  // This class also provides access to the current EventLoop's reference to the currently-active
+  // EnvironmentSet.
 
 public:
   class Scope;
+  // Put a `Scope(set)` on the stack to make an environment current. Requires a current EventLoop.
+
+  static Maybe<EnvironmentSet&> tryGetCurrent();
+  // Call `EnvironmentSet::tryGetCurrent()` to access the current EnvironmentSet, if any.
+
+  static Maybe<EnvironmentSet> tryCaptureCurrent();
+  // Return an EnvironmentSet containing the same environments as the current EnvironmentSet, or
+  // null if there is no current EnvironmentSet.
+
+  explicit EnvironmentSet(void* typeKey, Own<Environment> environment);
+  // Clone the current EnvironmentSet, adding `environment` to the set. If an Environment already
+  // exists in the EnvironmentSet for `typeKey`, `environment` takes its place in the newly-
+  // constructed EnvironmentSet.
+
+  EnvironmentSet(EnvironmentSet&&) = default;
+  EnvironmentSet& operator=(EnvironmentSet&&) = default;
+
+  ~EnvironmentSet() noexcept(false);
+
+  Maybe<Environment&> tryGetEnvironment(void* typeKey);
+  // Return the specific Environment associated with `typeKey`.
+
+private:
+  struct Impl;
+  Own<Impl> impl;
+  explicit EnvironmentSet(Own<Impl>);
+};
+
+class EnvironmentSet::Scope {
+  // Construct one of these on the stack whenever you need to push an Environment reference onto the
+  // Environment stack.
+
+public:
+  explicit Scope(Maybe<EnvironmentSet&> environmentSet);
+  ~Scope() noexcept(false);
+  KJ_DISALLOW_COPY(Scope);
+
+private:
+  Maybe<EnvironmentSet&> previous;
+};
+
+class Environment: public kj::Refcounted {
+  // Base class of Environment::Holder<T>.
+
+public:
   template <typename T> class Holder;
 
   KJ_DISALLOW_COPY(Environment);
   ~Environment() noexcept(false) = 0;
 
-  static Maybe<Own<Environment>> tryCaptureCurrent();
-  static Maybe<Environment&> tryGetCurrent();
-
 private:
   Environment() = default;
-};
-
-class Environment::Scope {
-  // Construct one of these on the stack whenever you need to push an Environment reference onto the
-  // Environment stack.
-
-public:
-  explicit Scope(Maybe<Environment&> environment);
-  ~Scope() noexcept(false);
-  KJ_DISALLOW_COPY(Scope);
-
-private:
-  Maybe<Environment&> previous;
 };
 
 template <typename T>
@@ -338,14 +369,20 @@ public:
   Holder(Args&&... args): environment(fwd<Args>(args)...) {}
 
   T environment;
+
+  static void* typeKey() {
+    static bool uniqueAddress;
+    return &uniqueAddress;
+  }
 };
 
 }  // namespace _
 
 template <typename E, typename Func>
 PromiseForResult<Func, void> runInEnvironment(E&& environment, Func&& func) {
-  auto holder = refcounted<_::Environment::Holder<Decay<E>>>(fwd<E>(environment));
-  _::Environment::Scope scope(*holder);
+  using Holder = _::Environment::Holder<Decay<E>>;
+  _::EnvironmentSet set(Holder::typeKey(), refcounted<Holder>(fwd<E>(environment)));
+  _::EnvironmentSet::Scope scope(set);
   return evalNow(fwd<Func>(func));
 }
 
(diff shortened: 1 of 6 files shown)


harrishancock:
One open question I had while implementing this was, should we make it possible to override environments with a null environment? Perhaps a `runInEnvironment(nullptr_t, Func)` overload?

c++/src/kj/async.h
```diff
@@ -458,6 +458,21 @@ PromiseForResult<Func, void> evalLast(Func&& func) KJ_WARN_UNUSED_RESULT;
 // callback enqueues new events, then latter callbacks will not execute until those events are
 // drained.
 
+template <typename E, typename T, typename... Args>
+kj::Promise<T> runInEnvironment(E&& environment, FunctionParam<kj::Promise<T>(Args...)> func);
```
kentonv, line 462: Since this is a template, there's no need for `FunctionParam`. The second param can just be `Func&& func`, where `Func` is a template parameter.

This has two benefits:
1. No need to manually cast to `FunctionParam` at call sites to make template matching work.
2. Avoids forcing the callbacks to go through a virtual function invocation.

Note when you invoke `evalNow` in the implementation, it should be `evalNow(kj::fwd<Func>(func))`.

c++/src/kj/async.h
```diff
@@ -458,6 +458,21 @@ PromiseForResult<Func, void> evalLast(Func&& func) KJ_WARN_UNUSED_RESULT;
...
+//
+// Note that Environments never implicitly cross threads.
+
+template <typename T>
+T& getEnvironment();
+template <typename T>
+kj::Maybe<T&> tryGetEnvironment();
+// If the currently-executing code was called by `runInEnvironment()`, or if the currently-executing
+// code was scheduled asynchronously by code called by `runInEnvironment()`, return the current
+// environment downcast as a value of type T.
```
kentonv, line 474: It feels concerning that if you try to get the environment as the wrong type here, you'll get UB (or an exception in debug mode). This sort of means that there has to be agreement among all code in a process that there is one true environment type. If that's the case then maybe we should be defining the type in KJ.

At the same time, I noticed that AsyncLocalContext in node and similar API proposals actually support multiple overlapping async contexts. Maybe we want to do the same here. I would key it by type. So if you have one set of code using environments of type T and another set using type U, calling each other, they won't interfere.

I think we can do this without RTTI. If you have a function like:

```
template <typename T>
void* getEnvironmentKey() {
  static const bool b;
  return &b;
}
```

This should return a unique value for each type.

c++/src/kj/async-inl.h
```diff
@@ -300,6 +300,77 @@ class PromiseNode: private AsyncObject {
...
+  KJ_DISALLOW_COPY(Environment);
+  ~Environment() noexcept(false) = 0;
+
+  static Maybe<Own<Environment>> tryCaptureCurrent();
+  static Maybe<Environment&> tryGetCurrent();
+
+private:
+  Environment() = default;
+
+  static thread_local Maybe<Environment&> current;
```
kentonv, line 320: Two problems with this `thread_local`:
* Defining it in a header accessible from multiple translation units forces the compiler to add extra boilerplate when accessing it. Defining it as a global with static linkage in the `.c++` file reduces this boilerplate, and since you are only actually accessing it from inside the `.c++` file anyway, there's no downside.
* Defining it as `Maybe<Environment&>` may make the compiler believe this is a non-trivial value with construction and destruction requirements. I'm not sure if it's smart enough to see that the constructor and destructor are actually trivial. We definitely don't want the compiler to believe it has to arrange callbacks for thread start/stop. I would define it as `Environment*` instead to make things obvious for the compiler.

kentonv: Actually does this `thread_local` need to exist at all or can you store the information in the `EventLoop` which already has a `thread_local` pointing to it?

c++/src/kj/async.h
harrishancock: I like that idea, too, will give it a shot.

c++/src/kj/async-inl.h
kentonv: Hmm, detached promises are actually owned by the EventLoop. However, the EventLoop is only "current" for the thread during the time when the `WaitScope` exists, which is destroyed before the `EventLoop` is destroyed, so possibly before any detached promises are destroyed.

There's actually a `WaitScope::cancelAllDetached()` which exists specifically to work around this, destroying detached promises while the `EventLoop` is still current.

But why does this matter for environments exactly? Why does the promise's destructor cane if the event loop is still current?

harrishancock: > But why does this matter for environments exactly? Why does the promise's destructor cane if the event loop is still current?

It determines whether we can provide the following guarantee: "All user code scheduled asynchronously by `f()` from within `kj::runInEnvironment(E(e), f)` will have access to `e` by calling `kj::getEnvironment<E>()`." The "user code scheduled asynchronously" that I'm talking about includes things like `.then()` continuation destructors, `.attach()` attachment destructors, coroutine frame object destructors, etc.

If we don't require the event loop to be current, then we must accept that we also don't have access to the current environment ref (the line of code we're commenting on, now moved to the EventLoop). I think this means that `kj::getEnvironment<E>()` must throw if there's no current event loop, breaking that guarantee.

harrishancock: I made an attempt at moving the ref to the EventLoop in the latest three commits. The first one does the move, the second adds `cancelAllDetached()` to `~WaitScope()` to fix the issues with detached promises. The third fixes some bugs in my Fiber changes that were surfaced by the move to the EventLoop.

Regarding `cancelAllDetached()`, that was my first hacky impulse, but perhaps instead we should make a WaitScope in EventLoop's destructor for the purpose of dropping the daemons?

harrishancock: > perhaps instead we should make a WaitScope in EventLoop's destructor for the purpose of dropping the daemons?

I did this.

harrishancock:
I think I've addressed all feedback so far.

kentonv:
Have you done any benchmarking to verify that this doesn't add too much overhead to the promise system?

harrishancock:
Overhead seems significant.

I wrote an evalLater()-based loop that adds 10 million integers and does not use `tryGetEnvironment()`, and observed that this branch is roughly 11% slower than the master branch in an optimized build.

For a similar evalLater()-based loop that _does_ use `tryGetEnvironment(), I observed that having an environment causes the loop to run roughly 10% slower than not having an environment. The no-environment case was itself roughly 4% slower than the evalLater() loop that doesn't use `tryGetEnvironment()`.

The 11% penalty for code that doesn't care about environments at all seems most concerning.

### Before/after this PR

With this diff:
```
diff --git a/c++/src/kj/async-test.c++ b/c++/src/kj/async-test.c++
index a7b9bc5a..4edf3c6e 100644
--- a/c++/src/kj/async-test.c++
+++ b/c++/src/kj/async-test.c++
@@ -1553,5 +1553,20 @@ KJ_TEST("retryOnDisconnect") {
   }
 }

+kj::Promise<uint> asyncAccumulate(uint z, uint ttl) {
+  if (ttl == 0) return z;
+  z += ttl;
+  return evalLater([z, ttl]() {
+    return asyncAccumulate(z, ttl - 1);
+  });
+}
+
+KJ_TEST("benchmark: Pre-environment evalLater()") {
+  EventLoop eventLoop;
+  WaitScope waitScope(eventLoop);
+
+  asyncAccumulate(0, 10'000'000).wait(waitScope);
+}
+
 }  // namespace
 }  // namespace kj
```

This is a typical result from an optimized build of the master branch:
```
[ PASS ] async-test.c++:1564: benchmark: Pre-environment evalLater() (1497997 μs)
```

And this from an optimized build of this branch:
```
[ PASS ] async-test.c++:1564: benchmark: Pre-environment evalLater() (1668736 μs)
```

### With/without `runInEnvironment()`

With this diff:
```
diff --git a/c++/src/kj/async-environment-test.c++ b/c++/src/kj/async-environment-test.c++
index 997d6857..bae112a2 100644
--- a/c++/src/kj/async-environment-test.c++
+++ b/c++/src/kj/async-environment-test.c++
@@ -351,5 +351,31 @@ KJ_TEST("Environments are available in suspended coroutine destructors") {

 #endif  // KJ_HAS_COROUTINE

+kj::Promise<uint> asyncAccumulate(uint z, uint ttl) {
+  if (ttl == 0) return z;
+  z += ttl;
+  return evalLater([z, ttl]() {
+    return asyncAccumulate(z, ttl - tryGetEnvironment<uint>().orDefault(1));
+  });
+}
+
+KJ_TEST("benchmark: Environment-less evalLater()") {
+  EventLoop eventLoop;
+  WaitScope waitScope(eventLoop);
+
+  asyncAccumulate(0, 10'000'000).wait(waitScope);
+}
+
+KJ_TEST("benchmark: Environment-ful evalLater()") {
+  EventLoop eventLoop;
+  WaitScope waitScope(eventLoop);
+
+  runInEnvironment(uint(1), []() {
+    return asyncAccumulate(0, 10'000'000);
+  }).wait(waitScope);
+}
+
 }  // namespace
 }  // namespace kj
```

These are typical results from an optimized build:

```
[ PASS ] async-environment-test.c++:362: benchmark: Environment-less evalLater() (1706716 μs)
[ PASS ] async-environment-test.c++:369: benchmark: Environment-ful evalLater() (1932639 μs)
```

Sarcasm:
I'm wondering, what is the purpose of the asynchronous environments?
I think I understand how it works and how it can be used, I guess it can be useful for different things but I'm wondering about the concrete use case(s) you have in mind by implementing it.

kentonv:
Can we re-test this with http-over-capnp-perf-test to get a better idea of the real performance overhead?

@Sarcasm Most use cases revolve around debugging and tracing. E.g. in an async server handling multiple requests concurrently, when we log an error, we might want to log which request we were running on behalf of.

harrishancock:
Superseded by https://github.com/capnproto/capnproto/pull/1546.

PR #1548, opened 2022-09-21
mark stringifyStackTrace() as weak

mikea:
Rather then implementing all possible solutions, make it overridable.

### DIFF
--- a/c++/src/kj/BUILD.bazel
+++ b/c++/src/kj/BUILD.bazel
@@ -166,8 +166,8 @@ cc_test(
         "//conditions:default": ["@platforms//:incompatible"],
     }),
     deps = [
-        "//src/kj/compat:kj-http",
         ":kj-test",
+        "//src/kj/compat:kj-http",
     ],
 )
 
@@ -210,4 +210,14 @@ cc_test(
         ":kj-async",
         ":kj-test",
     ],
+)
+
+cc_test(
+    name = "exception-override-symbolizer-test",
+    srcs = ["exception-override-symbolizer-test.c++"],
+    deps = [
+        ":kj",
+        ":kj-test",
+    ],
+    linkstatic = True,
 )
--- a/c++/src/kj/exception-override-symbolizer-test.c++
+++ b/c++/src/kj/exception-override-symbolizer-test.c++
@@ -1,0 +1,49 @@
+// Copyright (c) 2022 Cloudflare, Inc. and contributors
+// Licensed under the MIT License:
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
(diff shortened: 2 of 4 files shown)


c++/src/kj/exception-override-symbolizer-test.c++
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) 2013-2014 Sandstorm Development Group, Inc. and contributors
```
kentonv, line 1:
```suggestion
// Copyright (c) 2022 Cloudflare, Inc. and contributors
```

c++/src/kj/exception.h
```diff
@@ -369,7 +369,7 @@ KJ_NOINLINE ArrayPtr<void* const> getStackTrace(ArrayPtr<void*> space, uint igno
 // is never exactly equal to `space.begin()` due to this effect, even if `ignoreCount` is zero
 // since `getStackTrace()` needs to ignore its own internal frames).
 
-String stringifyStackTrace(ArrayPtr<void* const>);
+String stringifyStackTrace(ArrayPtr<void* const>) __attribute__((weak));
```
kentonv, line 372: Hmm, shouldn't the attribute go on the implementation, not the prototype? Otherwise how does the compiler know which implementation is the weak one?

mikea:
FYI I have decided against enabling this test in CMake/Make builds since those link all tests together and this new test spikes the symbolizer, which is not a nice dev experience.

c++/src/kj/exception.c++
```diff
@@ -254,6 +254,9 @@ ArrayPtr<void* const> getStackTrace(ArrayPtr<void*> space, uint ignoreCount) {
 #endif
 }
 
+#if __GNUC__ && !(__MINGW32__ || __MINGW64__)
```
kentonv, line 257: Maybe:

```suggestion
#if __GNUC__ && !_WIN32
```

? Since I think Windows in general doesn't support this.

c++/src/kj/BUILD.bazel
```diff
@@ -211,3 +211,14 @@ cc_test(
...
+
+cc_test(
+    name = "exception-override-symbolizer-test",
+    srcs = ["exception-override-symbolizer-test.c++"],
+    deps = [
+        ":kj",
+        ":kj-test",
+    ],
+    linkstatic = True,
+    # linkshared = True,
```
kentonv, line 223: Delete commented line?

c++/src/kj/exception.c++
```diff
@@ -254,6 +254,9 @@ ArrayPtr<void* const> getStackTrace(ArrayPtr<void*> space, uint ignoreCount) {
...
+__attribute__((weak))
```
kentonv, line 258:
```suggestion
// Allow dependents to override the implementation of stack symbolication by making it a weak
// symbol. We prefer weak symbols over some sort of callback registration mechanism becasue this
// allows an alternate symbolication library to be easily linked into tests without changing the
// code of the test.
__attribute__((weak)) 
```

kentonv:
> FYI I have decided against enabling this test in CMake/Make builds since those link all tests together and this new test spikes the symbolizer, which is not a nice dev experience.

Agreed!

PR #1551, opened 2022-09-23
Support C++20 coroutines

harrishancock:
This avoids deprecation warnings when using modern clang builds.

Still need to test this in Windows.


kentonv:
Well that was easy.

PR #1552, opened 2022-09-29
Fix utf-8 escape sequences parsing for json strings

WowVeryLogin:
Hex value was converted wrongfully to a code-point value. Test added.


harrishancock (approved):

kentonv:
Thanks. That's a doozy, ugh.

PR #1555, opened 2022-10-02
Add support for kqueue in UnixEventPort.

kentonv:
This will hopefully benefit MacOS. However, I do not have a Mac, and Apple absolutely prohibits MacOS running in a VM on non-Mac hardware. FreeBSD works perfectly in my VM. So I tested this on FreeBSD.

In the meantime, I also fixed all the other tests that were failing on FreeBSD.

I guess we'll see how Mac does in CI.


kentonv:
Woweee it works! (Except for the, uh, various signal-handling issues that I'm handwaving away, but yeah.)

c++/src/kj/async-unix.c++
```diff
@@ -671,7 +712,440 @@ bool UnixEventPort::poll() {
...
+  inline ChildExitPromiseAdapter(PromiseFulfiller<int>& fulfiller,
+                                 UnixEventPort& eventPort, Maybe<pid_t>& pid)
+      : eventPort(eventPort), pid(pid), fulfiller(fulfiller) {
+    pid_t p = KJ_ASSERT_NONNULL(pid);
+
+    struct kevent event;
+    EV_SET(&event, p, EVFILT_PROC, EV_ADD | EV_CLEAR, NOTE_EXIT, 0, this);
+    KJ_SYSCALL(kevent(eventPort.kqueueFd, &event, 1, nullptr, 0, nullptr));
+
+    // Check for race where child had already existed before the event was waiting.
```
harrishancock, line 1011: existed, or exited?

harrishancock (approved):
Nice!

heatd:
Hi guys,

REALLY cool issue you found on FreeBSD ZFS (lurked capnproto at the really right time :P). After talking with a FreeBSD pal I submitted an issue to OpenZFS (https://github.com/openzfs/zfs/issues/13991). In the meanwhile, if you really want to keep the test for FreeBSD you can try and squeeze in a `sleep(5);` which in my tests usually allows it to update `st_blocks.`

kentonv:
@heatd Hah, glad someone was able to track that down a little further! I'd rather not have tests sleeping for long periods so I think we'll just live with the work-around. TBH this test is so unreliable across platforms and filesystems that I should probably remove it altogether...

PR #1557, opened 2022-10-03
kj::Vector<T>::reserve() should always grow by at least doubling the …

harrishancock:
…current capacity


harrishancock:
Rebased on master.

PR #1558, opened 2022-10-03
Remove mvcapture

JonasVautherin:
Because it felt like it would not hurt to try removing this pre-C++14 workaround :innocent::

```cpp
template <typename Func, typename MovedParam>
inline CaptureByMove<Func, Decay<MovedParam>> mvCapture(MovedParam&& param, Func&& func) {
  // Hack to create a "lambda" which captures a variable by moving it rather than copying or
  // referencing.  C++14 generalized captures should make this obsolete, but for now in C++11 this
  // is commonly needed for Promise continuations that own their state.  Example usage:
  //
  //    Own<Foo> ptr = makeFoo();
  //    Promise<int> promise = callRpc();
  //    promise.then(mvCapture(ptr, [](Own<Foo>&& ptr, int result) {
  //      return ptr->finish(result);
  //    }));

  return CaptureByMove<Func, Decay<MovedParam>>(kj::fwd<Func>(func), kj::mv(param));
}
```

### DIFF
--- a/c++/src/capnp/capability.c++
+++ b/c++/src/capnp/capability.c++
@@ -239,8 +239,7 @@ public:
         .detach([](kj::Exception&&) {});  // ignore exceptions
 
     // Now the other branch returns the response from the context.
-    auto promise = forked.addBranch().then(kj::mvCapture(context,
-        [](kj::Own<LocalCallContext>&& context) {
+    auto promise = forked.addBranch().then([context=kj::mv(context)]() mutable {
       // force response allocation
       auto reader = context->getResults(MessageSize { 0, 0 }).asReader();
 
@@ -258,7 +257,7 @@ public:
       } else {
         return kj::mv(KJ_ASSERT_NONNULL(context->response));
       }
-    }));
+    });
 
     // We return the other branch.
     return RemotePromise<AnyPointer>(
@@ -422,11 +421,11 @@ public:
 
     // Create a promise for the call initiation.
     kj::ForkedPromise<kj::Own<CallResultHolder>> callResultPromise =
-        promiseForCallForwarding.addBranch().then(kj::mvCapture(context,
-        [=](kj::Own<CallContextHook>&& context, kj::Own<ClientHook>&& client){
+        promiseForCallForwarding.addBranch().then(
+        [=,context=kj::mv(context)](kj::Own<ClientHook>&& client) mutable {
           return kj::refcounted<CallResultHolder>(
               client->call(interfaceId, methodId, kj::mv(context)));
-        })).fork();
+        }).fork();
 
     // Create a promise that extracts the pipeline from the call initiation, and construct our
     // QueuedPipeline to chain to it.
@@ -606,11 +605,11 @@ public:
     // We have to fork this promise for the pipeline to receive a copy of the answer.
     auto forked = promise.fork();
 
-    auto pipelinePromise = forked.addBranch().then(kj::mvCapture(context->addRef(),
-        [=](kj::Own<CallContextHook>&& context) -> kj::Own<PipelineHook> {
+    auto pipelinePromise = forked.addBranch()
+        .then([=,context=context->addRef()]() mutable -> kj::Own<PipelineHook> {
           context->releaseParams();
           return kj::refcounted<LocalPipeline>(kj::mv(context));
-        }));
+        });
 
     auto tailPipelinePromise = context->onTailCall().then([](AnyPointer::Pipeline&& pipeline) {
       return kj::mv(pipeline.hook);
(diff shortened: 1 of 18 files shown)


c++/src/capnp/capability.c++
```diff
@@ -606,11 +605,12 @@ public:
     // We have to fork this promise for the pipeline to receive a copy of the answer.
     auto forked = promise.fork();
 
-    auto pipelinePromise = forked.addBranch().then(kj::mvCapture(context->addRef(),
-        [=](kj::Own<CallContextHook>&& context) -> kj::Own<PipelineHook> {
+    auto contextRef = context->addRef();
+    auto pipelinePromise = forked.addBranch()
+        .then([=,context=kj::mv(contextRef)]() mutable -> kj::Own<PipelineHook> {
```
kentonv, line 610:
```suggestion
    auto pipelinePromise = forked.addBranch()
        .then([=,context=context->addRef()]() mutable -> kj::Own<PipelineHook> {
```

c++/src/capnp/ez-rpc.c++
```diff
@@ -177,10 +177,11 @@ Capability::Client EzRpcClient::importCap(kj::StringPtr name) {
   KJ_IF_MAYBE(client, impl->clientContext) {
     return client->get()->restore(name);
   } else {
-    return impl->setupPromise.addBranch().then(kj::mvCapture(kj::heapString(name),
-        [this](kj::String&& name) {
+    auto nameCopy = kj::heapString(name);
+    return impl->setupPromise.addBranch().then(
+        [this,name=kj::mv(nameCopy)]() {
```
kentonv, line 182:
```suggestion
    return impl->setupPromise.addBranch().then(
        [this,name=kj::heapString(name)]() {
```

c++/src/kj/async-io-unix.c++
```diff
@@ -1747,9 +1747,10 @@ public:
       : lowLevel(parent.lowLevel), filter(allow, deny, parent.filter) {}
 
   Promise<Own<NetworkAddress>> parseAddress(StringPtr addr, uint portHint = 0) override {
-    return evalLater(mvCapture(heapString(addr), [this,portHint](String&& addr) {
+    auto addrCopy = heapString(addr);
+    return evalLater([this,portHint,addr=kj::mv(addrCopy)]() {
```
kentonv, line 1751:
```suggestion
    return evalLater([this,portHint,addr=heapString(addr)]() {
```

c++/src/kj/async.h
```diff
@@ -570,21 +570,6 @@ class CaptureByMove {
   MovedParam param;
 };
 
-template <typename Func, typename MovedParam>
-inline CaptureByMove<Func, Decay<MovedParam>> mvCapture(MovedParam&& param, Func&& func) {
```
kentonv, line 574: I suspect some dependents may use this, we should keep it but mark it deprecated.

kentonv (commented):
Hey thanks, this is a nice cleanup! I intended do to this eventually but never got around to it.

I tweaked a couple places where I don't think it's necessary to use an intermediate local variable. I think we also have to keep `mvCapture` defined for the moment; see comment.

c++/src/kj/async.h
JonasVautherin: Makes sense! I could not use `KJ_DEPRECATED` in the function declaration so I had to add a definition above, not sure if there is a cleaner way...

JonasVautherin:
> I intended do to this eventually but never got around to it.

Happy to help where I can :innocent:

kentonv:
Tip with code reviews: I often like to use the "compare" button next to the force-push notification to see what has changed since my last review. But in this case, it looks like you both made the changes I suggested _and_ rebased on master at the same time. Unfortunately that means the "compare" button isn't very useful as it mixes both sets of changes together. I recommend doing separate force-pushes to address code review comments, vs. to rebase on master. (Or avoid the latter if there's no conflicts to resolve.)

Not a big deal this time, just something to keep in mind.

Thanks for the cleanup!

JonasVautherin:
You mean doing it in two steps, like so?

1. Add the changes in one (or more) _new_ commits
2. After you approve the changes, cleanup the history

I was trying to keep the history clean so that you don't have to ask for a cleanup and come back later, but if it helps with the review, I'll do that next time of course!

kentonv:
No, that's not what I was saying at all. I definitely appreciate having a clean commit history.

I'm talking about the button that appears on the right of this line:

Screenshot from 2022-10-06 18-43-39

This button would normally let me review what changed, even after you've rewritten history. But if you simultaneously rewrite your commit history of your own changes, _and_ you rebase on the latest master branch, then these changes get mixed together and the button is less useful. If you click on it above, you'll see there's a bunch of changes in there that aren't yours...

JonasVautherin:
Ooooh, I get it now, sorry about that! I don't use the compare button when reviewing myself (I marked the files as "viewed", and then usually it works fairly well). But it actually seems pretty cool, I'll start using the compare button too :sunglasses:.

Duly noted, won't happen again :blush:.

PR #1137, opened 2020-12-21
Adding WITH_ZLIB Cmake option

ScatteredRay:
WITH_ZLIB function can be used to disable zlib even if it's detected, it's modeled off of WITH_OPENSSL

PR #1560, opened 2022-10-04
cmake: add ZLIB option

JonasVautherin:
Similar, but better than #1137 which is generating an empty `libkj-gzip.a`, whereas this one is just not building it at all if `ZLIB` is disabled (just like the Makefile does, I checked). I added the author @ScatteredRay as a co-author here :+1:.

@harrishancock had a mention there:

> On a related note, do we need to add a find_dependency(ZLIB) call to capnproto/c++/cmake/CapnProtoConfig.cmake.in? Similar to: https://github.com/capnproto/capnproto/blob/master/c%2B%2B/cmake/CapnProtoConfig.cmake.in#L52-L63

I think this is a valid point, but I would need to test it to be sure. I will open another PR if I get to that. This PR has no impact on it however (it just feels like a somehow related bug in CapnProtoConfig.cmake.in), so IMO it could be merged as-is.

### DIFF
--- a/c++/CMakeLists.txt
+++ b/c++/CMakeLists.txt
@@ -46,6 +46,10 @@ set(WITH_OPENSSL "AUTO" CACHE STRING
 # define list of values GUI will offer for the variable
 set_property(CACHE WITH_OPENSSL PROPERTY STRINGS AUTO ON OFF)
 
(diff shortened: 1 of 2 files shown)


harrishancock (approved):
Thanks!

PR #1561, opened 2022-10-07
Import coCapture() from workerd

harrishancock:
Since the defect this feature works around applies to all lambda coroutines, it makes sense to live in libkj.


kentonv (approved):
I wonder if this is the most efficient possible implementation. It's certainly elegant, but a custom promise node might involve fewer allocations?

We can always optimize later, of course. It's not like we use this in many places.

harrishancock:
> I wonder if this is the most efficient possible implementation. It's certainly elegant, but a custom promise node might involve fewer allocations?

I think there's only one extra allocation: for the wrapping coroutine frame. I think any PromiseNode-based implementation would also need at least one allocation. It'd be nice if we could combine that with the inner coroutine frame's allocation, but I spent some time thinking about how to do that, and it seems impossible -- and I don't think I've seen any of the other hacks on the interwebs able to solve it that way either.

On the other hand, I think clang theoretically has all the information it needs to perform heap elision on the inner frame, since the inner coroutine frame's lifetime is provably bounded by the outer coroutine's frame. So, this might actually be the most efficient possible implementation.

I wrote a little toy example to explore it, and eventually got clang to elide coroutine frame allocations: https://godbolt.org/z/69jrTz4j5 However, I noticed that if I change `final_suspend()` to return `suspend_always` (which is what KJ coroutines return), it degrades to performing two allocations again. I _think_ I made our `final_suspend()` return `suspend_always` because it was most convenient for us implementation-wise. Maybe revisiting that could help us optimize further.

kentonv:
Oh, I didn't know "heap elision" was a thing! Yes, we should do whatever we can to make that work...

c++/src/kj/async.h
```diff
@@ -585,6 +585,114 @@ inline CaptureByMove<Func, Decay<MovedParam>> mvCapture(MovedParam&& param, Func
...
+  //
+  // For situations where you are trying to capture a specific local variable, kj::mvCapture() can
+  // also be useful:
+  // ```
+  // kj::Promise<void> reactToPromise(kj::Promise<MyType> promise) {
+  //   BigA a;
+  //   TinyB b;
+  //
+  //   doSomething(a, b);
+  //   return promise.then(kj::mvCapture(b, [](TinyB b, MyType type) -> kj::Promise<void> {
```
JonasVautherin, line 684: Out of curiosity, is there a need for `kj::mvCapture` (which got deprecated [recently](https://github.com/capnproto/capnproto/pull/1558)) instead of using a C++14 generalized lambda?

```
return promise.then([b=kj::mv(b)](MyTypetype) -> kj::Promise<void> {
```

kentonv: Yes, this whole comment block is discussing a problem with C++20 coroutines (`co_await`/`co_return`) when combined with lambads -- things captured in the `[]` capture block typically no longer exist after the first `co_await`, but can still be referenced, leading to undefined behavior. Parameters, however, do not suffer this problem, since they are copied into the coroutine frame. So it turns out `kj::mvCapture` is, unintentionally, a possible work-around for this problem, since it uses parameters instead of `[]`. But `coCapture` is probably a better solution.

PR #1562, opened 2022-10-08
Various CMake improvements for WITH_ZLIB=OFF and WITH_OPENSSL=OFF

JonasVautherin:
* Fix a bug where heavy-tests do not build if `WITH_ZLIB=OFF` (because they were still linking kj-gzip which is not built in that case).
* Have `CapnProtoConfig.cmake` find ZLIB if `WITH_ZLIB=ON` since it's a transitive dependency (just like OpenSSL).
* Do not build `kj-tls` if `WITH_OPENSSL=OFF`. I don't see a use-case where one would want that, and it ends up in `undefined reference` errors at link time instead of complaining at configure time.

@harrishancock: FYI


harrishancock (approved):
Thanks!

PR #1566, opened 2022-10-24
Tinker with performance

kentonv:
In this PR:
* Set up slightly better tooling for working on Cap'n Proto with Ekam independently of any dependent project.
* Create a benchmark covering KJ HTTP and http-over-capnp.
* Improve performance by like 15% with a couple tweaks.

### DIFF
--- a/c++/src/kj/async-inl.h
+++ b/c++/src/kj/async-inl.h
@@ -345,7 +345,7 @@ public:
   }
 
   template <typename T, typename... Params>
-  static kj::Own<T, PromiseDisposer> alloc(Params&&... params) {
+  static kj::Own<T, PromiseDisposer> alloc(Params&&... params) noexcept {
     // Implements allocPromise().
     T* ptr;
     if (sizeof(T) > sizeof(PromiseArena)) {
@@ -353,6 +353,10 @@ public:
       ptr = new T(kj::fwd<Params>(params)...);
     } else {
       // Start a new arena.
+      //
+      // NOTE: As in append() (below), we don't implement exception-safety because it causes code
+      //   bloat and these constructors probably don't throw. Instead this function is noexcept, so
+      //   if a constructor does throw, it'll crash rather than leak memory.
       auto* arena = new PromiseArena;
       ptr = reinterpret_cast<T*>(arena + 1) - 1;
       ctor(*ptr, kj::fwd<Params>(params)...);
@@ -1590,8 +1594,8 @@ namespace _ {  // (private)
 class XThreadEvent: public PromiseNode,    // it's a PromiseNode in the requesting thread
                     private Event {        // it's an event in the target thread
 public:
-  XThreadEvent(ExceptionOrValue& result, const Executor& targetExecutor, void* funcTracePtr,
-               SourceLocation location);
+  XThreadEvent(ExceptionOrValue& result, const Executor& targetExecutor, EventLoop& loop,
+               void* funcTracePtr, SourceLocation location);
 
   void tracePromise(TraceBuilder& builder, bool stopAtNextEvent) override;
 
@@ -1692,8 +1696,8 @@ template <typename Func, typename = _::FixVoid<_::ReturnType<Func, void>>>
 class XThreadEventImpl final: public XThreadEvent {
   // Implementation for a function that does not return a Promise.
 public:
-  XThreadEventImpl(Func&& func, const Executor& target, SourceLocation location)
-      : XThreadEvent(result, target, GetFunctorStartAddress<>::apply(func), location),
+  XThreadEventImpl(Func&& func, const Executor& target, EventLoop& loop, SourceLocation location)
+      : XThreadEvent(result, target, loop, GetFunctorStartAddress<>::apply(func), location),
         func(kj::fwd<Func>(func)) {}
   ~XThreadEventImpl() noexcept(false) { ensureDoneOrCanceled(); }
   void destroy() override { dtor(*this); }
@@ -1720,8 +1724,8 @@ template <typename Func, typename T>
 class XThreadEventImpl<Func, Promise<T>> final: public XThreadEvent {
   // Implementation for a function that DOES return a Promise.
 public:
-  XThreadEventImpl(Func&& func, const Executor& target, SourceLocation location)
-      : XThreadEvent(result, target, GetFunctorStartAddress<>::apply(func), location),
+  XThreadEventImpl(Func&& func, const Executor& target, EventLoop& loop, SourceLocation location)
+      : XThreadEvent(result, target, loop, GetFunctorStartAddress<>::apply(func), location),
         func(kj::fwd<Func>(func)) {}
   ~XThreadEventImpl() noexcept(false) { ensureDoneOrCanceled(); }
   void destroy() override { dtor(*this); }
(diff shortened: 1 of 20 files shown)


kentonv:
It seems that the bazel build depends on a specific release of zlib which has been pulled. :grimacing:

c++/src/kj/async.c++
```diff
@@ -120,9 +120,20 @@ KJ_THREADLOCAL_PTR(DisallowAsyncDestructorsScope) disallowAsyncDestructorsScope
 
 }  // namespace
 
-AsyncObject::~AsyncObject() noexcept {
+AsyncObject::~AsyncObject() {
   // Since the method is noexcept, this will abort the process.
```
JonasVautherin, line 124: Nitpick: It is not `noexcept` anymore

kentonv:
I've extended this with some optimization of PromiseNode allocation.

This produces a few percent improvement, but not as much as I had hoped for. It seems like the time spent in malloc() and free() is cut in half, though, which is cool. We might be able to tune the size of PromiseArena, although in my brief experiments doubling the size and halving the size from what it is now both seemed to make things worse.

Overall this PR achieves 15%-20% perf gain on the benchmark at this point. (It'll be interesting to see how it impacts workerd overall...)

c++/src/kj/async.c++
```diff
@@ -1201,36 +1186,33 @@ XThreadPaf::XThreadPaf()
...
+    executor.impl->state.when([&](auto&) {
+      return state == FULFILLED || state == DISPATCHED;
     }, [&](Executor::Impl::State& exState) {
-      if (obj->state == FULFILLED) {
+      if (state == FULFILLED) {
         // The object is on the queue but was not yet dispatched. Remove it.
-        exState.fulfilled.remove(*obj);
+        exState.fulfilled.remove(*this);
       }
     });
```
jclee, line 1212: I might be mistaken, but I could have sworn that there's a compiler that complains if you try to implicitly use `this` in a lambda without explicitly capturing it.

c++/src/kj/async-inl.h
```diff
@@ -306,18 +314,97 @@ class PromiseNode: private AsyncObject {
...
+    if (arena == nullptr ||
+        reinterpret_cast<byte*>(next.get()) - reinterpret_cast<byte*>(arena) < sizeof(T)) {
+      // No arena available, or not enough space. Start new arena.
+      return alloc<T>(kj::mv(next), kj::fwd<Params>(params)...);
+    } else {
+      // Append to arena.
+      //
+      // NOTE: When we call ctor(), it takes ownership of `next`, so we shouldn't assume `next`
+      //   still exists after it returns. So we have to remove ownership of the arena before that.
+      //   In theory if we wanted this bo exception-safe, we'd also have to arrange to delete
```
jclee, line 383: s/`bo`/`to be` ?

c++/src/kj/async-inl.h
```diff
@@ -306,18 +314,97 @@ class PromiseNode: private AsyncObject {
...
+    // Implements allocPromise().
+    T* ptr;
+    if (sizeof(T) > sizeof(PromiseArena)) {
+      // Node too big, fall back to regular heap allocation.
+      ptr = new T(kj::fwd<Params>(params)...);
+    } else {
+      // Start a new arena.
+      auto* arena = new PromiseArena;
+      ptr = reinterpret_cast<T*>(arena + 1) - 1;
+      ctor(*ptr, kj::fwd<Params>(params)...);
```
jclee, line 358: Not sure if this is a concern, but if ctor throws, would this leak the arena?

kentonv: Hmm actually this does seem to be a problem. Although none of the constructors called by `append()` will throw, there are some constructors called by `alloc()` than can throw (rejecting invalid arguments). In fact, the reason I have not marked `alloc()` as `noexcept` is because it broke a test that was explicitly checking for such an exeption.

I guess I should probably mark `alloc()` as `noexcept` as well, and accept that the exception in question becomes fatal. I think that's probably better than the code bloat needed to conditionally delete the arena on exceptions.

c++/src/kj/async-inl.h
```diff
@@ -134,6 +134,11 @@ class TraceBuilder {
   void** limit;
 };
 
+struct alignas(void*) PromiseArena {
```
jclee, line 137: I suppose it couldn't hurt to add `final` here, in case it maybe helps the compiler work out sizeof(PromiseArena) earlier?

c++/src/kj/async-inl.h
jclee: I suppose another possibility is just to have an arena always preallocated, and only claim and reallocate it if there is no exception.  But I imagine that has its own set of pitfalls.

c++/src/kj/async-inl.h
```diff
@@ -134,6 +134,11 @@ class TraceBuilder {
...
+  // Space in which a chain of promises may be allocated. See PromiseDisposer.
+  byte bytes[1024];
```
jclee, line 139: I guess if certain applications have deeper promise chains or intrinsically bigger promise nodes, there may be some benefit to having a dynamic promise chain that allocates a new arena twice as large, each time it runs out.  Although the presence of deep promise chains suggests the application might have other problems.

It occurs to me that if the promises in the chain are particularly lightweight, a fair proportion of the arena will be devoted to storing its own address, which could alternately be encoded as a small offset from the promise object itself.  ...I guess that fact might be useful in an alternate formulation of promises, but presumably not worth sacrificing simplicity or natural C++ implementation in this one.

bcaimano:
Most of the optimizations seem pretty reasonable and familiar to me! I do have some questions about how much we're actually getting from the arena optimizations but I think that's more of a matter of tuning allocations and reusing slots than anything else.

c++/src/kj/async-inl.h
```diff
@@ -306,18 +314,97 @@ class PromiseNode: private AsyncObject {
...
+
+  friend class PromiseDisposer;
 };
 
 class PromiseDisposer {
 public:
   static void dispose(PromiseNode* node) {
+    PromiseArena* arena = node->arena;
     node->destroy();
+    delete arena;
```
harrishancock, line 344: Maybe worth a comment that `delete` is safe to call on `nullptr`? I had to look it up to be sure.

harrishancock (approved):
Very nice! I just had some nits.

I'm looking forward to rebasing my async environments PR on this.

c++/src/kj/async-inl.h
kentonv: I don't think that makes a difference, since `PromiseArena` is not polymorphic (has no virtual methods), `final` doesn't enable any particular optimizations.

I did wonder for a bit if this has an effect on the compiler's ability to generate sized deletes (call the two-argument operator delete that takes the size as a parameter), but I couldn't actually get Clang to produce such calls in godbolt at all. Anyway, I don't think it should make a difference since it would be UB to delete a pointer to a superclass of an object if that superclass doesn't have a virtual destructor.

(Incidentally, I also tried directly calling two-argument operator delete, and even directly calling tcmalloc's underlying functions, but these both made things slower. I think the compiler must be able to optimize around operators new and delete when it actually understands them...)

c++/src/kj/async-inl.h
kentonv: I think a bigger potential use for dynamic-sized arenas is that we could maybe be smarter about guessing how big an arena is needed in any particular use case. However, I haven't thought of any clever heuristics for that yet. I think for now having a static size probably helps by reducing code footprint a bit.

c++/src/kj/async-inl.h
```diff
@@ -239,6 +236,29 @@ class PromiseNode: private AsyncObject {
   //    handle to delete themselves.
   // 2. XThreadEvents sometimes leave it up to a different thread to actually delete the object.
 
+private:
+  PromiseArena* arena = nullptr;
+  // If non-null, then this PromiseNode is the last node allocated within the given arena, and
```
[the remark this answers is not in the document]
kentonv: meh, only sort of

c++/src/kj/async-inl.h
```diff
@@ -217,13 +217,10 @@ class Event: private AsyncObject {
...
-  // A Promise<T> contains a chain of PromiseNodes tracking the pending transformations.
-  //
-  // To reduce generated code bloat, PromiseNode is not a template.  Instead, it makes very hacky
-  // use of pointers to ExceptionOrValue which actually point to ExceptionOr<T>, but are only
-  // so down-cast in the few places that really need to be templated.  Luckily this is all
-  // internal implementation details.
+class PromiseArenaMember {
+  // An object that is allocated in a PromiseArena. `PromiseNode` inherits this, and most
+  // arena-allocated objects are `PromiseNode` subclasses, but `TaskSet::Task` and potentially
+  // other objects that commonly live on the end of a promise chain can also leverage this.
```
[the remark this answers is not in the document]
kentonv: Yeah the refcounting makes it difficult... I would probably have to re-implement refcounting rather than use kj::Refcounted. That said, that does seem like an interesting thing to explore... maybe I will later.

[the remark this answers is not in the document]
kentonv: Went ahead and implemented this. Worth another couple of percentage points!

bcaimano (commented):
Wamp, apologies, I didn't notice that the review I put up didn't submit with my comment. Happily, nothing I had to point out is blocking in the least.

PR #1567, opened 2022-11-03
Implements NetworkHttpClient::Connect and fixes for EOF

dom96:
While implementing TCP sockets for workerd I ran into a few things which I fix in this PR:

* NetworkHttpClient::Connect wasn't implemented (Kenton suggested we implement this as a direct TCP connection, I left a link to where he suggested this in a code comment). It works well for direct (non-proxied) connections in workerd.
* Reads of zero bytes (indicating a socket disconnect) weren't working. I narrowed it down to two things: BlockedPumpFrom and DelayedEofIoStream.
  * ~~The former seems to not be handling EOF correctly, though perhaps it is meant to block on EOF by design as well and I should instead fix the code so it's not used in the HTTP CONNECT code path?~~
  * ~~The latter is that ConnectResponseImpl uses a DelayedEofIoStream wrapper, I am not sure why this is used but removing it resolves the EOF issues.~~
  * The solution to the DelayedEofIoStream problem is avoiding the adapter and creating my own `ConnectResponse` which does not use `DelayedEofIoStream` (done in workerd)
  * As for the BlockedPumpFrom, it turned out that was fine, the problem was in HttpServiceAdapter::connect. This is fixed in this PR.

### Test Plan

```
$ bazel test //src/kj/compat:http-test
$ bazel test //src/kj/compat:http-socketpair-test
```

### DIFF
--- a/c++/src/kj/compat/http-test.c++
+++ b/c++/src/kj/compat/http-test.c++
@@ -5201,6 +5201,36 @@ KJ_TEST("HttpClient concurrency limiting") {
 }
 
 #if KJ_HTTP_TEST_USE_OS_PIPE
+// This test relies on access to the network.
+KJ_TEST("NetworkHttpClient connect impl") {
+  KJ_HTTP_TEST_SETUP_IO;
+  auto listener1 = io.provider->getNetwork().parseAddress("localhost", 0)
+      .wait(io.waitScope)->listen();
+
+  auto ignored KJ_UNUSED = listener1->accept().then([](Own<kj::AsyncIoStream> stream) {
+    auto buffer = kj::str("test");
+    return stream->write(buffer.cStr(), buffer.size()).attach(kj::mv(stream), kj::mv(buffer));
+  }).eagerlyEvaluate(nullptr);
+
+  HttpClientSettings clientSettings;
+  kj::TimerImpl clientTimer(kj::origin<kj::TimePoint>());
+  HttpHeaderTable headerTable;
+  auto client = newHttpClient(clientTimer, headerTable,
+      io.provider->getNetwork(), nullptr, clientSettings);
+  auto connection = client->connect(
+      kj::str("localhost:", listener1->getPort()), HttpHeaders(headerTable));
+  connection.then([](HttpClient::ConnectResponse response) {
+    auto& conn = KJ_ASSERT_NONNULL(response.connectionOrBody.tryGet<kj::Own<kj::AsyncIoStream>>());
+    auto buf = kj::heapArray<char>(4);
+    return conn->tryRead(buf.begin(), 1, buf.size()).then([buf = kj::mv(buf)](size_t count) {
+      KJ_ASSERT(count == 4);
+      KJ_ASSERT(kj::str(buf.asChars()) == "test");
+    }).attach(kj::mv(conn));
+  }).wait(io.waitScope);
+}
+#endif
+
+#if KJ_HTTP_TEST_USE_OS_PIPE
 // TODO(someday): Implement mock kj::Network for userspace version of this test?
 KJ_TEST("HttpClient multi host") {
   auto io = kj::setupAsyncIo();
@@ -5927,7 +5957,6 @@ public:
 
 private:
   HttpHeaderTable& headerTable;
-  kj::Maybe<kj::Own<kj::AsyncIoStream>> stream;
 };
 
 KJ_TEST("Simple CONNECT Server works") {
@@ -6240,7 +6269,7 @@ public:
       statusCode,
       statusText,
       headersCopy.get(),
-      pipe.ends[0].attach(kj::addRef(*this)),
+      pipe.ends[0].attach(kj::addRef(*this), kj::mv(headersCopy)),
     });
     return kj::mv(pipe.ends[1]);
   }
@@ -6277,8 +6306,8 @@ KJ_TEST("CONNECT Service Adapter handles stream closure correctly") {
   auto tunnel = kj::refcounted<TestConnectResponseImpl>(kj::mv(paf.fulfiller));
   // The below promise is fulfilled when the read below finishes successfully.
   auto readPaf = kj::newPromiseAndFulfiller<void>();
-  httpService->connect("https://example.org"_kj, HttpHeaders(clientHeaders), *tunnel).detach(
-      [](kj::Exception err) { KJ_DBG(err); KJ_FAIL_ASSERT(false); });
+  auto connectPromise = httpService->connect(
+      "https://example.org"_kj, HttpHeaders(clientHeaders), *tunnel);
   paf.promise.then([&](HttpClient::ConnectResponse response) -> kj::Promise<void> {
     KJ_ASSERT(response.statusCode == 200);
     KJ_ASSERT(response.statusText == "OK"_kj);
@@ -6291,10 +6320,12 @@ KJ_TEST("CONNECT Service Adapter handles stream closure correctly") {
     return io->tryRead(&buffer, 1, 1).then([&](size_t count) {
       KJ_ASSERT(count == 0, "Expecting the stream to get disconnected.");
       readPaf.fulfiller->fulfill();
-    }).attach(kj::mv(io), kj::mv(buffer));
+    }).attach(kj::mv(io));
   }).wait(waitScope);
 
+  connectPromise.wait(waitScope);
   KJ_ASSERT(readPaf.promise.poll(waitScope));
+
   listenTask.wait(waitScope);
 }
 
--- a/c++/src/kj/compat/http-test.c++
+++ b/c++/src/kj/compat/http-test.c++
@@ -5905,6 +5905,31 @@ private:
   HttpServer server;
 };
 
+class ConnectCloseService final: public HttpService {
+  // A simple CONNECT server that will accept a connection then immediately
+  // shutdown the write side of the AsyncIoStream to simulate socket disconnection.
+public:
+  ConnectCloseService(HttpHeaderTable& headerTable)
+      : headerTable(headerTable) {}
+
+  kj::Promise<void> request(
+      HttpMethod method, kj::StringPtr url, const HttpHeaders& headers,
+      kj::AsyncInputStream& requestBody, Response& response) override {
+    KJ_UNIMPLEMENTED("Regular HTTP requests are not implemented here.");
+  }
+
+  kj::Promise<void> connect(
+      kj::StringPtr host, const HttpHeaders& headers, ConnectResponse& tunnel) override {
+    auto io = tunnel.accept(200, "OK", HttpHeaders(headerTable));
+    io->shutdownWrite();
+    return kj::READY_NOW;
+  }
+
+private:
+  HttpHeaderTable& headerTable;
+  kj::Maybe<kj::Own<kj::AsyncIoStream>> stream;
+};
+
 KJ_TEST("Simple CONNECT Server works") {
   KJ_HTTP_TEST_SETUP_IO;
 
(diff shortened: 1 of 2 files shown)


c++/src/kj/async-io.c++
```diff
@@ -750,7 +750,7 @@ private:
           pipe.endState(*this);
         }
 
-        if (actual >= minBytes) {
+        if (actual >= minBytes || actual == 0) {
```
kentonv, line 753: This isn't correct, because EOF does not propagate through a pipe. If `tryPumpFrom()` on the write end reaches EOF (of whatever other stream it's pulling from), the write end of the pipe is still open. You could call `tryPumpFrom()` again to pump from a different source, or call write() to push individual buffers of data.

The way EOF would usually propagate is that the promise returned by `tryPumpFrom()` would complete, and then whatever was waiting on that would destroy the write end of the pipe, which then causes the read end to receive EOF.

c++/src/kj/compat/http.c++
```diff
@@ -6186,14 +6199,11 @@ private:
 
       auto pipe = newTwoWayPipe();
 
-      kj::Own<kj::AsyncIoStream> io =
-          kj::heap<DelayedEofIoStream>(kj::mv(pipe.ends[0]), task.attach(kj::addRef(*this)));
```
kentonv, line 6190: What we really need here is for DelayedEofIoStream to inject the delay before the stream is ended in _both_ directions, that is, delay the second EOF that passes over it. However, there currently isn't really a way to do that, because sending an EOF is not an async operation... that's something I've been meaning to fix, by adding an `AsyncOutputStream::end()` method. Until then, I don't really know what to do here.

If you can avoid using the adapter entirely, that would probably be best for now.

c++/src/kj/compat/http.c++
```diff
@@ -5418,7 +5418,20 @@ public:
...
+        return address->connect().then([&headers](auto connection) {
+          return ConnectResponse {
+            .statusCode = 200,
+            .statusText = "OK",
+            .headers = &headers,
+            .connectionOrBody = kj::mv(connection),
+          };
+        });
+      }
+    );
```
kentonv, line 5434: Lambda body should always be indented one stop more than the surrounding body.

```suggestion
    return network.parseAddress(host)
        .then([&headers](auto address) {
      return address->connect().then([&headers](auto connection) {
        return ConnectResponse {
          .statusCode = 200,
          .statusText = "OK",
          .headers = &headers,
          .connectionOrBody = kj::mv(connection),
        };
      });
    });
```

kentonv: If this seems confusing, consider how you'd indent an `if()` statement where the condition wraps across multiple lines. Do you write:

```
if (really + long + condition ==
  blah + blah) {
    blockContents();
  }
```

Or do you write:

```
if (really + long + condition ==
    blah + blah) {
  blockContents();
}
```

c++/src/kj/compat/http.c++
```diff
@@ -5418,7 +5418,20 @@ public:
...
+    return network.parseAddress(host).then(
+      [&headers](auto address) {
+        return address->connect().then([&headers](auto connection) {
+          return ConnectResponse {
+            .statusCode = 200,
+            .statusText = "OK",
+            .headers = &headers,
+            .connectionOrBody = kj::mv(connection),
+          };
+        });
```
kentonv, line 5432: I think a `.attach(kj::mv(address))` is needed here.

c++/src/kj/compat/http.c++
```diff
@@ -5418,7 +5418,20 @@ public:
...
-    KJ_UNIMPLEMENTED("connect() is not implemented for NetworkHttpClient");
+    // We want to connect directly instead of going through a proxy here.
+    // https://github.com/capnproto/capnproto/pull/1454#discussion_r900414879
+    return network.parseAddress(host).then(
+      [&headers](auto address) {
+        return address->connect().then([&headers](auto connection) {
+          return ConnectResponse {
+            .statusCode = 200,
+            .statusText = "OK",
+            .headers = &headers,
```
kentonv, line 5429: This is supposed to be the response headers (for an HTTP CONNECT method), but you're returning the request headers as the response headers.

You should be returning an empty headers object. Technically the object should be owned by `connectionOrBody`, which you could accomplish by `connection.attach()`. I guess the headers object itself has to be heap-allocated so that the reference isn't invalidated when you pass it into `attach()`.

dom96:
Tests added

c++/src/kj/compat/http-test.c++
```diff
@@ -5905,6 +5905,31 @@ private:
...
+  kj::Promise<void> connect(
+      kj::StringPtr host, const HttpHeaders& headers, ConnectResponse& tunnel) override {
+    auto io = tunnel.accept(200, "OK", HttpHeaders(headerTable));
+    io->shutdownWrite();
+    return kj::READY_NOW;
+  }
+
+private:
+  HttpHeaderTable& headerTable;
+  kj::Maybe<kj::Own<kj::AsyncIoStream>> stream;
```
harrishancock, line 5961: Nit: unused member?

c++/src/kj/compat/http-test.c++
```diff
@@ -6188,6 +6213,91 @@ KJ_TEST("CONNECT Server cancels read w/client") {
...
+        kj::throwRecoverableException(kj::mv(exception));
+      }
+    });
+  }
+
+  kj::Own<kj::AsyncIoStream> accept(
+      uint statusCode, kj::StringPtr statusText, const kj::HttpHeaders& headers) override {
+    KJ_REQUIRE(statusCode >= 200 && statusCode < 300, "the statusCode must be 2xx for accept");
+
+    auto headersCopy = kj::heap(headers.clone());
```
harrishancock, line 6266: Does this need to be attached anywhere?

c++/src/kj/compat/http-test.c++
```diff
@@ -6188,6 +6213,91 @@ KJ_TEST("CONNECT Server cancels read w/client") {
...
+  auto client = newHttpClient(table, *pipe.ends[1]);
+  auto httpService = newHttpService(*client);
+
+  HttpHeaderTable clientHeaders;
+  auto paf = kj::newPromiseAndFulfiller<kj::HttpClient::ConnectResponse>();
+  auto tunnel = kj::refcounted<TestConnectResponseImpl>(kj::mv(paf.fulfiller));
+  // The below promise is fulfilled when the read below finishes successfully.
+  auto readPaf = kj::newPromiseAndFulfiller<void>();
+  httpService->connect("https://example.org"_kj, HttpHeaders(clientHeaders), *tunnel).detach(
+      [](kj::Exception err) { KJ_DBG(err); KJ_FAIL_ASSERT(false); });
```
harrishancock, line 6312: Leftover KJ_DBG. Also, why detach? Can we wait() this promise at the end, instead?

c++/src/kj/compat/http-test.c++
```diff
@@ -6188,6 +6213,91 @@ KJ_TEST("CONNECT Server cancels read w/client") {
...
+
+    auto& io = KJ_ASSERT_NONNULL(response.connectionOrBody.tryGet<kj::Own<kj::AsyncIoStream>>());
+    char buffer;
+    // Attempt a read. It should disconnect. We are verifying here that the stream is shutdown
+    // correctly, if it doesn't this will block forever (and KJ will give us a
+    // "this thread would hang forever" error message).
+    return io->tryRead(&buffer, 1, 1).then([&](size_t count) {
+      KJ_ASSERT(count == 0, "Expecting the stream to get disconnected.");
+      readPaf.fulfiller->fulfill();
+    }).attach(kj::mv(io), kj::mv(buffer));
```
harrishancock, line 6325: In this case, attaching the buffer does not prolong its lifetime. ;) Maybe the `char buffer` could live at the next scope out, or be heap-allocated?

c++/src/kj/compat/http-test.c++
```diff
@@ -5200,6 +5200,37 @@ KJ_TEST("HttpClient concurrency limiting") {
   KJ_EXPECT(callbackEvents == kj::ArrayPtr<const CallbackEvent>({ {0, 0} }));
 }
 
+KJ_TEST("NetworkHttpClient connect impl") {
+  {
+    KJ_HTTP_TEST_SETUP_IO;
+  }
```
harrishancock, line 5206: Since this test relies on having access to the network, I think it should follow the same pattern as the ones below it:
```
#if KJ_HTTP_TEST_USE_OS_PIPE
// A comment that this test relies on having access to a real network.
KJ_TEST(...) {
  auto io = kj::setupAsyncIo();
  // ...
```

c++/src/kj/compat/http-test.c++
```diff
@@ -5200,6 +5200,37 @@ KJ_TEST("HttpClient concurrency limiting") {
...
+  auto io = kj::setupAsyncIo();
+  auto listener1 = io.provider->getNetwork().parseAddress("localhost", 0)
+      .wait(io.waitScope)->listen();
+
+  auto errHandler = [](kj::Exception err) { KJ_DBG(err); KJ_FAIL_ASSERT(false); };
```
harrishancock, line 5211: Leftover KJ_DBG.

c++/src/kj/compat/http-test.c++
```diff
@@ -5200,6 +5200,37 @@ KJ_TEST("HttpClient concurrency limiting") {
...
+  listener1->accept().then([](Own<kj::AsyncIoStream> stream) {
+    auto buffer = kj::str("test");
+    return stream->write(buffer.cStr(), buffer.size()).attach(kj::mv(stream));
```
harrishancock, line 5214: Also needs to attach `buffer`, I think.

c++/src/kj/compat/http-test.c++
```diff
@@ -5200,6 +5200,37 @@ KJ_TEST("HttpClient concurrency limiting") {
...
+  }).detach(errHandler);
```
harrishancock, line 5215: Why detach?

c++/src/kj/compat/http-test.c++
```diff
@@ -5200,6 +5200,37 @@ KJ_TEST("HttpClient concurrency limiting") {
...
+
+  HttpClientSettings clientSettings;
+  kj::TimerImpl clientTimer(kj::origin<kj::TimePoint>());
+  HttpHeaderTable headerTable;
+  auto client = newHttpClient(clientTimer, headerTable,
+      io.provider->getNetwork(), nullptr, clientSettings);
+  auto connection = client->connect(
+      kj::str("localhost:", listener1->getPort()), HttpHeaders(headerTable));
+  connection.then([](HttpClient::ConnectResponse response) {
+    auto& io = KJ_ASSERT_NONNULL(response.connectionOrBody.tryGet<kj::Own<kj::AsyncIoStream>>());
```
harrishancock, line 5225: Nit: shadows `io`, of `setupAsyncIo()` fame.

c++/src/kj/compat/http-test.c++
```diff
@@ -5200,6 +5200,37 @@ KJ_TEST("HttpClient concurrency limiting") {
...
+    auto buf = kj::heapArray<char>(4);
+    return io->tryRead(buf.begin(), 1, buf.size()).then([buf = kj::mv(buf)](size_t count) {
+      KJ_ASSERT(count == 4);
+      KJ_ASSERT(kj::str(buf.asChars()) == "test");
+    }).attach(kj::mv(io), kj::mv(buf));
```
harrishancock, line 5230: `buf` has already been captured by the lambda at this point.

c++/src/kj/compat/http-test.c++
dom96: I based on this code which doesn't attach it: https://github.com/capnproto/capnproto/blob/master/c++/src/kj/compat/http.c++#L6185. 

Should both be changed?

c++/src/kj/compat/http-test.c++
dom96: It shouldn't really matter in this case. The callback doesn't read `buffer` anyway, so I'll just remove it from the attach statement.

c++/src/kj/compat/http-test.c++
harrishancock: Hmm, I suspect so. FYI @jasnell

c++/src/kj/compat/http-test.c++
harrishancock: > The callback doesn't read buffer anyway

But the implementation of `tryRead()` could write to it if the test were to fail for some reason. Such failures probably won't ever manifest as a segfault, but to guarantee we see such failures as `"Expecting the stream to get disconnected."` assertion failures I think technically the buffer should be valid for `tryRead()`.

c++/src/kj/compat/http-test.c++
```diff
@@ -5200,6 +5200,36 @@ KJ_TEST("HttpClient concurrency limiting") {
...
+// This test relies on access to the network.
+KJ_TEST("NetworkHttpClient connect impl") {
+  KJ_HTTP_TEST_SETUP_IO;
+  auto listener1 = io.provider->getNetwork().parseAddress("localhost", 0)
+      .wait(io.waitScope)->listen();
+
+  auto ignored KJ_UNUSED = listener1->accept().then([](Own<kj::AsyncIoStream> stream) {
+    auto buffer = kj::str("test");
+    return stream->write(buffer.cStr(), buffer.size()).attach(kj::mv(stream), kj::mv(buffer));
+  }).eagerlyEvaluate([](kj::Exception err) { throw err; });
```
harrishancock, line 5213: I think `.eagerlyEvaluate(nullptr)` has the same effect.

dom96:
Last update was a rebase.

c++/src/kj/compat/http-test.c++
dom96: Okay, attached the headers in both cases.

harrishancock (approved):

c++/src/kj/compat/http.c++
```diff
@@ -5418,7 +5418,20 @@ public:
...
+    // We want to connect directly instead of going through a proxy here.
+    // https://github.com/capnproto/capnproto/pull/1454#discussion_r900414879
+    return network.parseAddress(host)
+        .then([this](auto address) {
+      return address->connect().then([this](auto connection) {
+        auto emptyHeaders = kj::heap<kj::HttpHeaders>(responseHeaderTable);
+        return ConnectResponse {
+          200, // .statusCode
+          "OK", // .statusText
+          emptyHeaders, // .headers
```
kentonv, line 5430: You are using `emptyHeaders` here, but on the next line -- which is part of the same expression -- you are moving `emptyHeaders` away. Does C++ guarantee that these actions will execute in the order you expect? I'm not sure. I would normally copy a pointer to a local variable in the line before, to avoid the ambiguity.

c++/src/kj/compat/http.c++
```diff
@@ -6187,7 +6200,8 @@ private:
       auto pipe = newTwoWayPipe();
 
       kj::Own<kj::AsyncIoStream> io =
-          kj::heap<DelayedEofIoStream>(kj::mv(pipe.ends[0]), task.attach(kj::addRef(*this)));
+          kj::heap<DelayedEofIoStream>(kj::mv(pipe.ends[0]),
+              task.attach(kj::addRef(*this), kj::mv(headersCopy)));
```
kentonv, line 6204: You are moving away `headersCopy` here but then you return it again below. You'll end up returning a null pointer. I guess we don't have any tests that actually examine the returned headers?

dom96: Related discussion: https://github.com/capnproto/capnproto/pull/1567#discussion_r1020415427.

Can you confirm that this definitely needs to be attached to something? if so what can I attach it to in this code?

dom96: Makes sense, I'll fix this once the http connect pipeline PR is merged.

c++/src/kj/compat/http.c++
dom96: Just for future reference: This was resolved as part of the pipelining PR.

c++/src/kj/compat/http.c++
dom96: Just for future reference: This was resolved as part of the pipelining PR.

PR #1570, opened 2022-11-09
Two RPC optimizations

kentonv:
1. When the client does not expect an RPC call to return any capabilities, set a flag to disable pipelining. This allows a lot of bookkeeping to be skipped. (This is the first 5 commits.) ~5% gain on full-RPC benchmark, 10% on local calls.

2. When a response contains no capabilities, set a hint flag telling the caller that no `Finish` message is required, and clean up the answer table entry immediately. This didn't noticeably improve CPU performance in the in-memory RPC test, but it does eliminate a lot of network messages, which would probably have a big impact if an actual network were involved in the test.

NOTE: This is based on top of #1569. It edits some of the same code so I didn't want to separate them out and re-merge later.

cc @xortive


kentonv:
Will assign @xortive as reviewer once he has accepted the repo invite.

Issue #1573, opened 2022-11-13
Issue with coroutine header when using Clang + libstdc++


chronoxor:
Clang compiler finds <coroutine> from gcc and finishes with compilation error:
```
-- The C compiler identification is Clang 10.0.0
-- The CXX compiler identification is Clang 10.0.0
...
In file included from /home/runner/work/CppSerialization/CppSerialization/modules/capnproto/c++/src/kj/async-prelude.h:38:
/usr/lib/gcc/x86_64-linux-gnu/10/../../../../include/c++/10/coroutine:295:2: error: "the coroutine header requires -fcoroutines"
#error "the coroutine header requires -fcoroutines"
 ^
```

To fix the issue we need some more checks in async-prelude.h:
```
#if __cpp_coroutines && (__cpp_impl_coroutine >= 201902L) && __has_include(<coroutine>)
// C++20 Coroutines detected.
#include <coroutine>
...
#elif __cpp_coroutines && (__cpp_coroutines >= 201703L) && __has_include(<experimental/coroutine>)
// Coroutines TS detected.
#include <experimental/coroutine>
...
#endif
```

The first check would cover gcc + libstdc++ and clang (from version 14)
The second check would cover clang + libc++ (up until version 13)

chronoxor:
Fixed!

PR #1574, opened 2022-11-13
Issue with coroutine header when using Clang + libstdc++

chronoxor:
To fix the [issue](https://github.com/capnproto/capnproto/issues/1573) we need some more checks in async-prelude.h:
```
#if __cpp_coroutines && (__cpp_impl_coroutine >= 201902L) && __has_include(<coroutine>)
// C++20 Coroutines detected.
#include <coroutine>
...
#elif __cpp_coroutines && (__cpp_coroutines >= 201703L) && __has_include(<experimental/coroutine>)
// Coroutines TS detected.
#include <experimental/coroutine>
...
#endif
```

The first check would cover gcc + libstdc++ and clang (from version 14)
The second check would cover clang + libc++ (up until version 13)

### DIFF
--- a/c++/src/kj/async-prelude.h
+++ b/c++/src/kj/async-prelude.h
@@ -33,12 +33,12 @@
 // TODO(someday): Support coroutines with -fno-exceptions.
 #if !KJ_NO_EXCEPTIONS
 #ifdef __has_include
-#if __cpp_coroutines && __has_include(<coroutine>)
+#if __cpp_coroutines && (__cpp_impl_coroutine >= 201902L) && __has_include(<coroutine>)
 // C++20 Coroutines detected.
 #include <coroutine>
 #define KJ_HAS_COROUTINE 1
 #define KJ_COROUTINE_STD_NAMESPACE std
-#elif __cpp_coroutines && __has_include(<experimental/coroutine>)
+#elif __cpp_coroutines && (__cpp_coroutines >= 201703L) && __has_include(<experimental/coroutine>)
 // Coroutines TS detected.
 #include <experimental/coroutine>
 #define KJ_HAS_COROUTINE 1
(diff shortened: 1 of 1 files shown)


c++/src/kj/async-prelude.h
```diff
@@ -33,12 +33,12 @@
 // TODO(someday): Support coroutines with -fno-exceptions.
 #if !KJ_NO_EXCEPTIONS
 #ifdef __has_include
-#if __cpp_coroutines && __has_include(<coroutine>)
+#if __cpp_coroutines && (__cpp_impl_coroutine >= 201902L) && __has_include(<coroutine>)
```
harrishancock, line 36: Looking at https://en.cppreference.com/w/cpp/feature_test, I don't actually see `__cpp_coroutines` at all. I wonder if that feature macro is technically defined just for the Coroutines TS, and we should only be relying on `__cpp_impl_coroutine` here to detect C++20 coroutines? E.g., I think this would work:

```cpp
#if (__cpp_impl_coroutine >= 201902L) && __has_include(<coroutine>)
// ...
#elif (__cpp_coroutines >= 201703L) && __has_include(<experimental/coroutine>)
// ...
#endif
```

Does that make sense to you?

chronoxor: Will check

chronoxor: Your fix works fine as well! I've updated the PR

chronoxor:
@harrishancock I've updated PR with your suggestion and checked the fix on Linux (gcc, clang), MacOS (clang), Windows (VS-2022, MinGW, MSYS2). Think we're ready to merge it in master.

harrishancock (approved):

harrishancock:
Thanks, I appreciate it!

PR #1576, opened 2022-11-14
Fixes confusing restrictPeers semantics with multiple named filters.

dom96:
Fixes https://github.com/cloudflare/workerd/issues/62

### DIFF
--- a/c++/src/kj/async-io-test.c++
+++ b/c++/src/kj/async-io-test.c++
@@ -1190,6 +1190,54 @@ KJ_TEST("NetworkFilter") {
   }
 
   {
+    _::NetworkFilter filter({"local", "public"}, {}, base);
+
+    // Public
+    KJ_EXPECT(allowed4(filter, "8.8.8.8"));
+
+    // Local
+    KJ_EXPECT(allowed4(filter, "127.0.0.1"));
+    KJ_EXPECT(allowed4(filter, "0.0.0.0"));
+    KJ_EXPECT(allowed4(filter, "localhost"));
+
+    // Private
+    KJ_EXPECT(!allowed6(filter, "fc00::1234"));
+    KJ_EXPECT(!allowed4(filter, "192.168.0.1"));
+  }
+
+  { // https://github.com/cloudflare/workerd/issues/62
+    _::NetworkFilter filter({"public", "private", "local", "network"}, {}, base);
+
+    // Public
+    KJ_EXPECT(allowed4(filter, "8.8.8.8"));
+
+    // Local
+    KJ_EXPECT(allowed4(filter, "127.0.0.1"));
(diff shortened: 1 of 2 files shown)


kentonv:
Can you please expand your commit and PR description to explain what was confusing and what this changes?

kentonv:
If I understand the problem correctly by reading the code, I think this may fix e.g. `allow({"public", "private"})`, but it does NOT fix e.g. `allow({"public", "127.0.0.1"})`

dom96:
Good point :)

kentonv:
I think the allow and deny lists might need to be more than just CIDRs. We may need to have explicit representations of "public", "private", etc. in the list, which are interpreted when the filter is run, rather than trying to compile them down to CIDRs in advance.

dom96:
I solved it a little bit differently, the hashing is probably overkill as well, but let me know what you think (and whether I missed any other edge cases)

kentonv:
I don't think this solves the issue. To be a bit more specific, imagine:

`allow({"public", "192.168.1.25/32"})`

This says: Allow public IPs, and also allow _specifically_ 192.168.1.25 (the `/32` means all 32 bits of the address must match).

The "private" network includes the CIDR `192.168.0.0/16`. Your code as-is will still end up adding that CIDR to the deny list, since it isn't explicitly in the allow list. But since `192.168.1.25` is a member of that CIDR, the IP will be denied, despite being individually in the allow list, because denies override allows.

Perhaps another way we could solve this is to change the logic from "deny always trumps allow" to "most-specific match wins". However, changing the semantics of a security API doesn't seem like a good idea.

So I think we need to go with the approach of saying that an individual entry in the allow list can actually be a "compound" entry that has a nested list of positive and negative CIDRs. We would then use this to implement "public" and "network" aliases.

dom96:
So I'm not 100% sure what you mean by "compound" entries.

The way I see it to support the above example what we need is a "soft deny", so "public" adds an allow for public CIDRs and a "soft deny" for private CIDRs. That way the "soft deny" won't override the allow for the `192.168.1.25/32` CIDR. This sounds similar to what you describe: 

> Perhaps another way we could solve this is to change the logic from "deny always trumps allow" to "most-specific match wins". However, changing the semantics of a security API doesn't seem like a good idea.

But you also say that we don't want to do this.

In any case, could we merge the initial fix since it's simple and already resolves at least the specific bug that was reported?

harrishancock (approved):

harrishancock:
@kentonv can we merge and file issues describing what needs to happen to fix `restrictPeers()` the rest of the way? This looks like a safe incremental improvement to me.

kentonv:
@harrishancock This is not hard to do right. Can we please do it right? Once something is merged no one ever comes back to it.

PR #1578, opened 2022-11-16
Include error code when (de)compression fails

MellowYarker:
Saw an event where decompression failed, but since we didn't log the error code we can't determine what went wrong.


harrishancock (approved):

harrishancock:
@kentonv this had two failing builds due to a network connectivity issue I think while running `apt-get update`. I re-ran the builds, and now they're failing due to .deb unavailability:
```
E: Package 'g++-7' has no installation candidate
```
and
```
E: Package 'clang-9' has no installation candidate
```

That can't have anything to do with the content of this PR, but I'm mystified by why some of our later builds aren't running into the same issue, which I'd expect if these packages don't exist in the runner images anymore. Perhaps the initial `apt-get update` failure poisoned some cache?

kentonv:
The build failures are likely because this PR doesn't include #1588, which fixed CI after GitHub updated the Ubuntu image to a new major version.

PR #1579, opened 2022-11-16
Implement pipelined connect() on HttpClient

jasnell:
Implements a variation on connect() that allows for optimistic piplining of outbound writes while the CONNECT tunnel is still being established. Useful for reducing latency. Best implemented as a separate method since not all uses will want to, or be capable of, making use of pipelining.

Work in progress: Needs tests

### DIFF
--- a/c++/src/kj/compat/http-test.c++
+++ b/c++/src/kj/compat/http-test.c++
@@ -5217,16 +5217,15 @@ KJ_TEST("NetworkHttpClient connect impl") {
   HttpHeaderTable headerTable;
   auto client = newHttpClient(clientTimer, headerTable,
       io.provider->getNetwork(), nullptr, clientSettings);
-  auto connection = client->connect(
+  auto request = client->connect(
       kj::str("localhost:", listener1->getPort()), HttpHeaders(headerTable));
-  connection.then([](HttpClient::ConnectResponse response) {
-    auto& conn = KJ_ASSERT_NONNULL(response.connectionOrBody.tryGet<kj::Own<kj::AsyncIoStream>>());
-    auto buf = kj::heapArray<char>(4);
-    return conn->tryRead(buf.begin(), 1, buf.size()).then([buf = kj::mv(buf)](size_t count) {
-      KJ_ASSERT(count == 4);
-      KJ_ASSERT(kj::str(buf.asChars()) == "test");
-    }).attach(kj::mv(conn));
-  }).wait(io.waitScope);
+
+  auto buf = kj::heapArray<char>(4);
+  return request.connection->tryRead(buf.begin(), 1, buf.size())
+      .then([buf = kj::mv(buf)](size_t count) {
+    KJ_ASSERT(count == 4);
+    KJ_ASSERT(kj::str(buf.asChars()) == "test");
+  }).attach(kj::mv(request.connection)).wait(io.waitScope);
 }
 #endif
 
@@ -5799,11 +5798,13 @@ public:
     KJ_UNIMPLEMENTED("Regular HTTP requests are not implemented here.");
   }
 
-  kj::Promise<void> connect(kj::StringPtr host, const HttpHeaders& headers,
-                           ConnectResponse& tunnel) override {
+  kj::Promise<void> connect(kj::StringPtr host,
+                            const HttpHeaders& headers,
+                            kj::AsyncIoStream& connection,
+                            ConnectResponse& response) override {
     connectCount++;
-    auto io = tunnel.accept(statusCodeToSend, "OK", HttpHeaders(headerTable));
-    return io->pumpTo(*io).ignoreResult().attach(kj::mv(io));
+    response.accept(statusCodeToSend, "OK", HttpHeaders(headerTable));
+    return connection.pumpTo(connection).ignoreResult();
   }
 
 private:
@@ -5828,10 +5829,12 @@ public:
     KJ_UNIMPLEMENTED("Regular HTTP requests are not implemented here.");
   }
 
-  kj::Promise<void> connect(kj::StringPtr host, const HttpHeaders& headers,
-                            ConnectResponse& tunnel) override {
+  kj::Promise<void> connect(kj::StringPtr host,
+                            const HttpHeaders& headers,
+                            kj::AsyncIoStream& connection,
+                            ConnectResponse& response) override {
     connectCount++;
-    auto out = tunnel.reject(statusCodeToSend, "Failed"_kj, HttpHeaders(headerTable), 4);
+    auto out = response.reject(statusCodeToSend, "Failed"_kj, HttpHeaders(headerTable), 4);
     return out->write("boom", 4).attach(kj::mv(out));
   }
 
@@ -5853,11 +5856,12 @@ public:
     KJ_UNIMPLEMENTED("Regular HTTP requests are not implemented here.");
   }
 
-  kj::Promise<void> connect(kj::StringPtr host, const HttpHeaders& headers,
-                           ConnectResponse& tunnel) override {
-    auto io = tunnel.accept(200, "OK", HttpHeaders(headerTable));
-
-    // Return an immediately resolved promise and drop the io
+  kj::Promise<void> connect(kj::StringPtr host,
+                            const HttpHeaders& headers,
+                            kj::AsyncIoStream& connection,
+                            ConnectResponse& response) override {
+    response.accept(200, "OK", HttpHeaders(headerTable));
+    // Return an immediately resolved promise and drop the connection
     return kj::READY_NOW;
   }
 
@@ -5878,12 +5882,14 @@ public:
     KJ_UNIMPLEMENTED("Regular HTTP requests are not implemented here.");
   }
 
-  kj::Promise<void> connect(kj::StringPtr host, const HttpHeaders& headers,
-                           ConnectResponse& tunnel) override {
-    auto io = tunnel.accept(200, "OK", HttpHeaders(headerTable));
+  kj::Promise<void> connect(kj::StringPtr host,
+                            const HttpHeaders& headers,
+                            kj::AsyncIoStream& connection,
+                            ConnectResponse& response) override {
+    response.accept(200, "OK", HttpHeaders(headerTable));
 
     auto msg = "hello"_kj;
-    auto promise KJ_UNUSED = io->write(msg.begin(), 5);
+    auto promise KJ_UNUSED = connection.write(msg.begin(), 5);
 
     // Return an immediately resolved promise and drop the io
     return kj::READY_NOW;
@@ -5909,11 +5915,12 @@ private:
     KJ_UNIMPLEMENTED("Regular HTTP requests are not implemented here.");
   }
 
-  kj::Promise<void> connect(kj::StringPtr host, const HttpHeaders& headers,
-                           ConnectResponse& tunnel) override {
-    auto io = tunnel.accept(200, "OK", HttpHeaders(tunneledService.table));
-
-    return server.listenHttp(kj::mv(io));
+  kj::Promise<void> connect(kj::StringPtr host,
+                            const HttpHeaders& headers,
+                            kj::AsyncIoStream& connection,
+                            ConnectResponse& response) override {
+    response.accept(200, "OK", HttpHeaders(tunneledService.table));
+    return server.listenHttp(kj::Own<kj::AsyncIoStream>(&connection, kj::NullDisposer::instance));
   }
 
   class SimpleHttpService final: public HttpService {
@@ -5948,10 +5955,12 @@ public:
     KJ_UNIMPLEMENTED("Regular HTTP requests are not implemented here.");
   }
 
-  kj::Promise<void> connect(
-      kj::StringPtr host, const HttpHeaders& headers, ConnectResponse& tunnel) override {
-    auto io = tunnel.accept(200, "OK", HttpHeaders(headerTable));
-    io->shutdownWrite();
+  kj::Promise<void> connect(kj::StringPtr host,
+                            const HttpHeaders& headers,
+                            kj::AsyncIoStream& connection,
+                            ConnectResponse& response) override {
+    response.accept(200, "OK", HttpHeaders(headerTable));
+    connection.shutdownWrite();
     return kj::READY_NOW;
   }
 
@@ -6009,21 +6018,20 @@ KJ_TEST("Simple CONNECT Client/Server works") {
   HttpHeaderTable clientHeaders;
   // Initiates a CONNECT with the echo server. Once established, sends a bit of data
   // and waits for it to be echoed back.
-  client->connect("https://example.org"_kj, HttpHeaders(clientHeaders))
-      .then([&](HttpClient::ConnectResponse response) {
-    KJ_ASSERT(response.statusCode == 200);
-    KJ_ASSERT(response.statusText == "OK"_kj);
+  auto request = client->connect("https://example.org"_kj, HttpHeaders(clientHeaders));
 
-    auto& io = KJ_ASSERT_NONNULL(response.connectionOrBody.tryGet<kj::Own<kj::AsyncIoStream>>());
+  request.status.then([io=kj::mv(request.connection)](auto status) mutable {
+    KJ_ASSERT(status.statusCode == 200);
+    KJ_ASSERT(status.statusText == "OK"_kj);
 
     auto promises = kj::heapArrayBuilder<kj::Promise<void>>(2);
     promises.add(io->write("hello", 5));
     promises.add(expectRead(*io, "hello"_kj));
-
-    kj::joinPromises(promises.finish()).wait(waitScope);
-
-    io->shutdownWrite();
-    expectEnd(*io).wait(waitScope);
+    return kj::joinPromises(promises.finish())
+        .then([io=kj::mv(io)]() mutable {
+      io->shutdownWrite();
+      return expectEnd(*io).attach(kj::mv(io));
+    });
   }).wait(waitScope);
 
   listenTask.wait(waitScope);
@@ -6087,22 +6095,21 @@ KJ_TEST("CONNECT Client (204 status)") {
   HttpHeaderTable clientHeaders;
   // Initiates a CONNECT with the echo server. Once established, sends a bit of data
   // and waits for it to be echoed back.
-  client->connect("https://example.org"_kj, HttpHeaders(clientHeaders))
-      .then([&](HttpClient::ConnectResponse response) {
-    KJ_ASSERT(response.statusCode == 204);
-    KJ_ASSERT(response.statusText == "OK"_kj);
+  auto request = client->connect("https://example.org"_kj, HttpHeaders(clientHeaders));
 
-    auto& io = KJ_ASSERT_NONNULL(response.connectionOrBody.tryGet<kj::Own<kj::AsyncIoStream>>());
+  request.status.then([io=kj::mv(request.connection)](auto status) mutable {
+    KJ_ASSERT(status.statusCode == 204);
+    KJ_ASSERT(status.statusText == "OK"_kj);
 
     auto promises = kj::heapArrayBuilder<kj::Promise<void>>(2);
     promises.add(io->write("hello", 5));
     promises.add(expectRead(*io, "hello"_kj));
 
-    kj::joinPromises(promises.finish()).wait(waitScope);
-
-    io->shutdownWrite();
-    expectEnd(*io).wait(waitScope);
-
+    return kj::joinPromises(promises.finish())
+        .then([io=kj::mv(io)]() mutable {
+      io->shutdownWrite();
+      return expectEnd(*io).attach(kj::mv(io));
+    });
   }).wait(waitScope);
 
   listenTask.wait(waitScope);
@@ -6144,6 +6151,7 @@ KJ_TEST("CONNECT Server rejected") {
   KJ_ASSERT(service.connectCount == 1);
 }
 
+#ifndef KJ_HTTP_TEST_USE_OS_PIPE
 KJ_TEST("CONNECT Client rejected") {
   KJ_HTTP_TEST_SETUP_IO;
 
@@ -6160,22 +6168,24 @@ KJ_TEST("CONNECT Client rejected") {
   auto client = newHttpClient(table, *pipe.ends[1]);
 
   HttpHeaderTable clientHeaders;
-  client->connect("https://example.org"_kj, HttpHeaders(clientHeaders))
-      .then([&](HttpClient::ConnectResponse response) {
-    KJ_ASSERT(response.statusCode == 400);
-    KJ_ASSERT(response.statusText == "Failed"_kj);
+  auto request = client->connect("https://example.org"_kj, HttpHeaders(clientHeaders));
 
-    auto& output = KJ_ASSERT_NONNULL(response.connectionOrBody
-        .tryGet<kj::Own<kj::AsyncInputStream>>());
+  request.status.then([](auto status) mutable {
+    KJ_ASSERT(status.statusCode == 400);
+    KJ_ASSERT(status.statusText == "Failed"_kj);
 
-    expectRead(*output, "boom"_kj).wait(waitScope);
-    expectEnd(*output);
+    auto& errorBody = KJ_ASSERT_NONNULL(status.errorBody);
+
+    return expectRead(*errorBody, "boom"_kj).then([&errorBody=*errorBody]() {
+      return expectEnd(errorBody);
+    }).attach(kj::mv(errorBody));
   }).wait(waitScope);
 
   listenTask.wait(waitScope);
 
   KJ_ASSERT(service.connectCount == 1);
 }
+#endif
 
 KJ_TEST("CONNECT Server cancels read") {
   KJ_HTTP_TEST_SETUP_IO;
@@ -6206,6 +6216,7 @@ KJ_TEST("CONNECT Server cancels read") {
   listenTask.wait(waitScope);
 }
 
+#ifndef KJ_HTTP_TEST_USE_OS_PIPE
 KJ_TEST("CONNECT Server cancels read w/client") {
   KJ_HTTP_TEST_SETUP_IO;
 
@@ -6220,114 +6231,26 @@ KJ_TEST("CONNECT Server cancels read w/client") {
   auto listenTask = server.listenHttp(kj::mv(pipe.ends[0]));
 
   auto client = newHttpClient(table, *pipe.ends[1]);
+  bool failed = false;
 
   HttpHeaderTable clientHeaders;
-  client->connect("https://example.org"_kj, HttpHeaders(clientHeaders))
-      .then([&](HttpClient::ConnectResponse response) {
-    KJ_ASSERT(response.statusCode == 200);
-    KJ_ASSERT(response.statusText == "OK"_kj);
+  auto request = client->connect("https://example.org"_kj, HttpHeaders(clientHeaders));
 
-    auto& io = KJ_ASSERT_NONNULL(response.connectionOrBody.tryGet<kj::Own<kj::AsyncIoStream>>());
+  request.status.then([&failed, io=kj::mv(request.connection)](auto status) mutable {
+    KJ_ASSERT(status.statusCode == 200);
+    KJ_ASSERT(status.statusText == "OK"_kj);
 
-    bool failed = false;
-    io->write("hello", 5).catch_([&failed](kj::Exception&& ex) {
+    return io->write("hello", 5).catch_([&](kj::Exception&& ex) {
       KJ_ASSERT(ex.getType() == kj::Exception::Type::DISCONNECTED);
       failed = true;
-    }).wait(waitScope);
-
-    KJ_ASSERT(failed, "the write promise should have failed");
-
-  }).wait(waitScope);
-
-  listenTask.wait(waitScope);
-}
-
-class TestConnectResponseImpl final: public kj::HttpService::ConnectResponse, public kj::Refcounted {
-public:
-  TestConnectResponseImpl(kj::Own<kj::PromiseFulfiller<kj::HttpClient::ConnectResponse>> fulfiller)
-      : fulfiller(kj::mv(fulfiller)) {}
-
-  void setPromise(kj::Promise<void> promise) {
-    task = promise.eagerlyEvaluate([this](kj::Exception&& exception) {
-      if (fulfiller->isWaiting()) {
-        fulfiller->reject(kj::mv(exception));
-      } else {
-        kj::throwRecoverableException(kj::mv(exception));
-      }
-    });
-  }
-
-  kj::Own<kj::AsyncIoStream> accept(
-      uint statusCode, kj::StringPtr statusText, const kj::HttpHeaders& headers) override {
-    KJ_REQUIRE(statusCode >= 200 && statusCode < 300, "the statusCode must be 2xx for accept");
-
-    auto headersCopy = kj::heap(headers.clone());
-
-    auto pipe = kj::newTwoWayPipe();
-
-    fulfiller->fulfill(kj::HttpClient::ConnectResponse {
-      statusCode,
-      statusText,
-      headersCopy.get(),
-      pipe.ends[0].attach(kj::addRef(*this), kj::mv(headersCopy)),
-    });
-    return kj::mv(pipe.ends[1]);
-  }
-
-  kj::Own<kj::AsyncOutputStream> reject(
-      uint statusCode, kj::StringPtr statusText, const kj::HttpHeaders& headers,
-      kj::Maybe<uint64_t> expectedBodySize) override {
-    KJ_ASSERT(false, "Not testing this branch.");
-    KJ_UNREACHABLE;
-  }
-private:
-  kj::Own<kj::PromiseFulfiller<kj::HttpClient::ConnectResponse>> fulfiller;
-  kj::Promise<void> task = nullptr;
-};
-
-KJ_TEST("CONNECT Service Adapter handles stream closure correctly") {
-  KJ_HTTP_TEST_SETUP_IO;
-
-  auto pipe = KJ_HTTP_TEST_CREATE_2PIPE;
-
-  kj::TimerImpl timer(kj::origin<kj::TimePoint>());
-
-  HttpHeaderTable table;
-  ConnectCloseService service(table);
-  HttpServer server(timer, table, service);
-
-  auto listenTask = server.listenHttp(kj::mv(pipe.ends[0]));
-
-  auto client = newHttpClient(table, *pipe.ends[1]);
-  auto httpService = newHttpService(*client);
-
-  HttpHeaderTable clientHeaders;
-  auto paf = kj::newPromiseAndFulfiller<kj::HttpClient::ConnectResponse>();
-  auto tunnel = kj::refcounted<TestConnectResponseImpl>(kj::mv(paf.fulfiller));
-  // The below promise is fulfilled when the read below finishes successfully.
-  auto readPaf = kj::newPromiseAndFulfiller<void>();
-  auto connectPromise = httpService->connect(
-      "https://example.org"_kj, HttpHeaders(clientHeaders), *tunnel);
-  paf.promise.then([&](HttpClient::ConnectResponse response) -> kj::Promise<void> {
-    KJ_ASSERT(response.statusCode == 200);
-    KJ_ASSERT(response.statusText == "OK"_kj);
-
-    auto& io = KJ_ASSERT_NONNULL(response.connectionOrBody.tryGet<kj::Own<kj::AsyncIoStream>>());
-    char buffer;
-    // Attempt a read. It should disconnect. We are verifying here that the stream is shutdown
-    // correctly, if it doesn't this will block forever (and KJ will give us a
-    // "this thread would hang forever" error message).
-    return io->tryRead(&buffer, 1, 1).then([&](size_t count) {
-      KJ_ASSERT(count == 0, "Expecting the stream to get disconnected.");
-      readPaf.fulfiller->fulfill();
     }).attach(kj::mv(io));
   }).wait(waitScope);
 
-  connectPromise.wait(waitScope);
-  KJ_ASSERT(readPaf.promise.poll(waitScope));
+  KJ_ASSERT(failed, "the write promise should have failed");
 
   listenTask.wait(waitScope);
 }
+#endif
 
 KJ_TEST("CONNECT Server cancels write") {
   KJ_HTTP_TEST_SETUP_IO;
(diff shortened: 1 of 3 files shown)


kentonv:
Can we change the existing `connect()` method to use this signature? I'm reasonably certain no one other than us is using it (after all, it was just recently implemented) so I am not worried about breakage.

jasnell:
I'm not yet convinced that pipelining is something we always want to do. It's possible a proxy may not support it, or if it does, there may be limitations such that a caller may want to be able to fallback. For instance, if the proxy implements a failover mechanism and is only able to buffer a certain amount of data for the failover, then a caller might opt out of the pipelining.

If we're not concerned about breaking changes, we certainly could *combine* the two signatures... essentially add a new argument to `connect()` and have it return `kj::OneOf<...>`? What would you think of that?

kentonv:
If the application doesn't want to pipeline, it can wait for the status promise before performing any writes. Won't that have the same effect? I don't see why we need to maintain two separate methods for this.

Keep in mind that if there are two methods -- or one method that has two distinct modes using a `OneOf` -- then every `WorkerInterface` implementation has to implement both of them, which seems like a big pain.

> For instance, if the proxy implements a failover mechanism and is only able to buffer a certain amount of data for the failover

In this case wouldn't the proxy just stop reading from the socket when it has reached its buffer limit?

jasnell:
Ideally yeah, it would stop. I just don't like making assumptions about what impls would ideally do ;-) ... As I said, I don't feel strongly about it and a single method is certainly better than two. So if you're good with the breaking API change I'll do that.

jasnell:
@dom96  .. just a heads up, this change is likely to impact your workerd socket api work as well since it is changing the existing `connect()` API

c++/src/kj/compat/http.c++
```diff
@@ -4093,6 +4093,103 @@ private:
...
+      inner = heap<AsyncIoStreamWithInitialBuffer>(
+          kj::mv(inner),
+          kj::mv(buffer.buffer), buffer.leftover);
+    }, [this](kj::Exception&& exception) {
+      // Set the permanent exception to catch new calls...
+      permanentException = kj::cp(exception);
+      // Trigger the canceler to interupt existing calls...
+      canceler.cancel(kj::cp(exception));
+      // Return the exception to error out the write guard...
+      return kj::mv(exception);
```
kentonv, line 4188: Since the success continuation returns void, the return value of the error continuation will be ignored, i.e. this exception will get swallowed. This is interesting, I don't think I've seen this before but it seems like a pretty big gotcha.

You need to do: `kj::throwRecoverableException(kj::mv(exception))`

c++/src/kj/compat/http.c++
```diff
@@ -4093,6 +4093,103 @@ private:
...
+
+  void shutdownWrite() override {
+    KJ_IF_MAYBE(exception, permanentException) {
+      throw *exception;
+    }
+    inner->shutdownWrite();
+  }
+
+  kj::Maybe<kj::Promise<uint64_t>> tryPumpFrom(AsyncInputStream& input, uint64_t amount) override {
+    KJ_IF_MAYBE(exception, permanentException) {
```
kentonv, line 4144: There's a lot of effort and bookkeeping overhead introduced here to make sure that writes end up propagating the status exception. I'm not really sure it's needed. You could say it's the app's responsibility to wait on the status and cancel any writes if it fails. This kind of bookkeeping has a cost, it's the kind of thing I've been removing from the RPC system in order to make capnp RPC faster.

c++/src/kj/compat/http.c++
```diff
@@ -4093,6 +4093,103 @@ private:
...
+    return guard.addBranch().then([this, &output, amount] {
+      return inner->pumpTo(output, amount);
+    });
+  }
+
+  // AsyncOutputStream
+
+  void shutdownWrite() override {
+    KJ_IF_MAYBE(exception, permanentException) {
+      throw *exception;
```
kentonv, line 4138: We don't use bare `throw` in KJ, use `kj::throwRecoverableException` or `kj::throwFatalException` instead (these two functions do the same thing when compiling with exceptions enabled, but different things when exceptions are disabled).

c++/src/kj/compat/http.c++
```diff
@@ -4093,6 +4093,103 @@ private:
...
+      kj::Promise<HttpInputStreamImpl::ReleasedBuffer> guard)
+      : inner(kj::mv(inner)),
+        guard(handleGuard(kj::mv(guard))) {}
+
+  // AsyncInputStream
+  Promise<size_t> tryRead(void* buffer, size_t minBytes, size_t maxBytes) override {
+    KJ_IF_MAYBE(exception, permanentException) {
+      return kj::cp(*exception);
+    }
+    return guard.addBranch().then([this, buffer, minBytes, maxBytes] {
```
kentonv, line 4116: Can you follow the pattern of `PromisedAsyncIoStream`, which avoids the overhead of `addBranch()` in the case that the promise has already resolved?

c++/src/kj/compat/http.h
```diff
@@ -718,16 +718,18 @@ class HttpClient {
...
+    };
+
+    kj::Promise<Status> status;
+    // If the CONNECT request is successful, the status will indicate a 2xx status code
+    // and the errorBody will be nullptr. If the CONNECT request fails, then the errorBody
+    // will be a kj::AsyncInputStream used to provide the payload (if any) of the error.
+    kj::Own<kj::AsyncIoStream> connection;
   };
 
   virtual kj::Promise<ConnectResponse> connect(kj::StringPtr host, const HttpHeaders& headers);
```
kentonv, line 735: Shouldn't this return `ConnectResponse` now, without the outer Promise?

(Maybe we rename it to `ConnectRequest` since it's very similar to the `Request` type that `request()` returns?)

c++/src/kj/compat/http.c++
```diff
@@ -6312,9 +6396,14 @@ public:
     KJ_REQUIRE(!headers.isWebSocket(), "WebSocket upgrade headers are not permitted in a connect.");
     return client.connect(host, headers)
         .then([&tunnel](HttpClient::ConnectResponse&& response) -> kj::Promise<void> {
-      KJ_SWITCH_ONEOF(response.connectionOrBody) {
-        KJ_CASE_ONEOF(io, kj::Own<kj::AsyncIoStream>) {
-          auto io2 = tunnel.accept(response.statusCode, response.statusText, *response.headers);
+      // TODO(someday): The adapter here currently does not support pipelining. We have to
```
kentonv, line 6399: This seems like it will defeat the purpose in `workerd`, since all subrequest channels there are represented as `WorkerInterface` which inherits `HttpService`, not `HttpClient`. So you will always end up using the adapter...

c++/src/kj/compat/http.c++
```diff
@@ -4956,9 +5054,12 @@ public:
 
     auto id = ++counter;
 
-    return httpInput.readResponseHeaders().then(
-        [this,id](HttpHeaders::ResponseOrProtocolError&& responseOrProtocolError)
-            -> HttpClient::ConnectResponse {
+    auto guard = kj::newPromiseAndFulfiller<HttpInputStreamImpl::ReleasedBuffer>();
```
kentonv, line 5057: It looks like if the caller drops the `status` promise without waiting for it, then the operation will be canceled, dropping this fulfiller, causing the promise to throw, so the stream will end up broken. I think waiting on `status` should be optional; if either `status` or `connection` is kept alive, the connection attempt should continue. So you'll need to `fork()` or `split()` a promise here instead of using PromiseAndFulfiller.

c++/src/kj/compat/http.c++
jasnell: Yeah, I was debating that. The reason I went with this approach was to actually keep the app from having to go through the trouble of doing this book keeping since it is likely just as much overhead and trouble there. That said, I'm definitely not that happy with this as is.

c++/src/kj/compat/http.h
jasnell: It certainly could. The only reason it's returning a promise is the `httpOutput.flush()` but we can switch it around to return a `PromisedAsyncIoStream` that waits on that

c++/src/kj/compat/http.c++
jasnell: Oof, yeah, good point.

c++/src/kj/compat/http.h
jasnell: This ends up being a much larger change than what I wanted to make here. If we want to change this signature, my preference is to do that as a separate PR.

c++/src/kj/compat/http.c++
```diff
@@ -4093,6 +4094,85 @@ private:
...
+  kj::Own<kj::AsyncIoStream> inner;
+  kj::ForkedPromise<void> guard;
+  bool guardReleased = false;
+
+  kj::ForkedPromise<void> handleGuard(
+      kj::Promise<kj::Maybe<HttpInputStreamImpl::ReleasedBuffer>> guard) {
+    return guard.then([this](kj::Maybe<HttpInputStreamImpl::ReleasedBuffer> buffer) mutable {
+      guardReleased = true;
+      KJ_IF_MAYBE(b, buffer) {
+        // We only need to replace the inner stream if a buffer is provided.
```
kentonv, line 4167: I suppose you could also check if b->leftover is empty, and if so, don't bother adding the wrapper. This is probably a common case with CONNECT: the headers arrive before the initial data, so there's no buffer.

c++/src/kj/compat/http.c++
```diff
@@ -6310,32 +6370,59 @@ public:
...
+      auto maybeCanceler = heap<MaybeCanceler>(kj::mv(response.status));
+
+      // This operates optimistically. In order to support pipelining, we have to
+      // accept the tunnel early, even if we're not yet certain that the tunnel can
+      // actually be established. If the response.status promise rejects or returns
+      // a failure, then we will interrupt the pipelines we establish below.
+      auto io = kj::mv(response.connection);
+      HttpHeaderTable headerTable {};
+      HttpHeaders responseHeaders(headerTable);
+      auto io2 = tunnel.accept(200, "OK"_kj, responseHeaders);
```
kentonv, line 6385: Isn't it a problem that you're always sending empty response headers here?

I think we need to change `HttpService::connect()` to support pipelining. It would need to take an `AsyncIoStream&` representing the connection as a parameter. We would document that `tunnel.accept()` must be called before writing to this stream (or, perhaps, that writes will be blocked until `accept()` has been called).

c++/src/kj/compat/http.h
kentonv: How so? It looks like the biggest thing is you need the write direction on `AsyncIoStreamWithDeferredInput` to be blocked by its own guard promise (which waits for `httpOutput.flush()`), but that's not that complicated?

I'd rather group all the breaking API changes together here...

jasnell: It's not a matter of complexity as much as size of the change. Changing the signature touches a lot more code than is strictly necessary to enable the pipelining. But ok. I'll have to work on it next week tho.

c++/src/kj/compat/http.c++
jasnell: The empty headers is certainly not ideal but I wouldn't call it a problem. We currently don't do anything with the response headers. That said, I'm not all that happy with this particular bit yet.

c++/src/kj/compat/http.c++
jasnell: Ah right, yeah good idea

c++/src/kj/compat/http.h
kentonv: This change already changes the signature, though? Wouldn't it touch the same code that was already touched?

c++/src/kj/compat/http.c++
kentonv: Your specific use case might not care about the headers, but we're implementing a general library here. If response headers can't be produced correctly then the API should be changed so that it doesn't pretend to produce them at all. But I think passing in the stream to `connect()` will end up with cleaner code anyway, e.g. implementing the HttpClient adapter should be easier if it doesn't have to wait for the service to call `accept()`.

c++/src/kj/compat/http.c++
```diff
@@ -6368,61 +6489,54 @@ public:
...
-              .attach(kj::mv(maybeCanceler));
-    });
+        auto out = tunnel.reject(status.statusCode, status.statusText, *status.headers);
+        return io1.pumpTo(*out).then([](uint64_t) -> kj::Promise<void> {
+          return kj::READY_NOW;
+        }).attach(kj::mv(out));
+      }
+    }).eagerlyEvaluate(nullptr));
+
+    return kj::joinPromises(promises.finish()).attach(kj::mv(io1), kj::mv(io2));
```
kentonv, line 6535: I'm a bit concerned that if the pump in one direction throws an error, it'll get held in limbo waiting for the other direction to resolve. This seems like it could easily hang.

However, I notice the handling of WebSockets appears to suffer from the same problem. Maybe it's a bug we should fix for both in a later PR? Maybe add a `// TODO(bug)` for now?

c++/src/kj/compat/http.c++
```diff
@@ -7126,40 +7244,53 @@ private:
         "received bad WebSocket handshake", errorMessage));
   }
 
-  kj::Own<kj::AsyncIoStream> accept(uint statusCode, kj::StringPtr statusText,
-                                    const HttpHeaders& headers) override {
+  kj::Own<kj::AsyncIoStream> getTunnelStream() {
```
kentonv, line 7247: Nitpick: Maybe `getConnectStream()`? The word "tunnel" feels misplaced to me.

c++/src/kj/compat/http.h
```diff
@@ -717,22 +717,18 @@ class HttpClient {
   // `url` and `headers` need only remain valid until `openWebSocket()` returns (they can be
   // stack-allocated).
 
-  struct ConnectResponse {
+  struct ConnectRequest {
     struct Status {
       uint statusCode;
       kj::StringPtr statusText;
       const HttpHeaders* headers;
-      kj::Maybe<kj::Own<kj::AsyncInputStream>> errorBody;
```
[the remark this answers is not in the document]
jasnell: Really because it's not really necessary and removing it simplifies some things (such as attaching the `ConnectionCounter` elsewhere.

c++/src/kj/compat/http.c++
```diff
@@ -6231,80 +6341,91 @@ private:
...
-      task = promise.eagerlyEvaluate([this](kj::Exception&& exception) {
-        if (fulfiller->isWaiting()) {
-          fulfiller->reject(kj::mv(exception));
-        } else {
-          kj::throwRecoverableException(kj::mv(exception));
-        }
-      });
+    ~ConnectResponseImpl() noexcept(false) {
+      if (fulfiller->isWaiting()) {
+        fulfiller->reject(KJ_EXCEPTION(DISCONNECTED, "the connect request was incomplete"));
```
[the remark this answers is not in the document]
jasnell: I think updating the others would be better, yes. Not in this PR tho.

c++/src/kj/compat/http.c++
```diff
@@ -6231,80 +6341,91 @@ private:
...
-                                      kj::StringPtr statusText,
-                                      const HttpHeaders& headers) override {
-      KJ_REQUIRE(statusCode >= 200 && statusCode < 300, "the statusCode must be 2xx for accept");
-
-      auto headersCopy = kj::heap(headers.clone());
-
-      auto pipe = newTwoWayPipe();
-
-      kj::Own<kj::AsyncIoStream> io =
-          kj::heap<DelayedEofIoStream>(kj::mv(pipe.ends[0]),
```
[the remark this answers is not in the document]
jasnell: I didn't remove it yet for precisely this reason. I wasn't sure if we were going to need it. Going to need to think through this a bit more tho.

c++/src/kj/compat/http.c++
```diff
@@ -6231,80 +6341,91 @@ private:
...
+
+    kj::Own<kj::AsyncIoStream> getTunnelStream() {
+      KJ_ASSERT(!tunnelStreamDetached, "the tunnel stream was already detached");
+      tunnelStreamDetached = true;
+      return kj::mv(kj::get<1>(streamAndFulfiller));
+    }
+
+    kj::Own<kj::AsyncOutputStream> getRejectionStream() {
+      auto out = kj::Own<kj::AsyncOutputStream>(
+          kj::get<0>(streamAndFulfiller).get(), kj::NullDisposer::instance);
```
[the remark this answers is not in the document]
jasnell: Yeah, this part is super weird as currently written and I'm not very happy with it yet.

c++/src/kj/compat/http.c++
jasnell: Yeah, I actually meant to add a comment here asking about that, I'm glad you mentioned it because I wasn't certain.

c++/src/kj/compat/http.h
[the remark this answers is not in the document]
jasnell: However, looking at the rest of the comments here... we might indeed need to bring it back. We'll need to think about how that impacts things that need to be *attached* to the streams. `ConnectionCounter`, for example. I presume that we'd have to transition that into a `Refcounted` so that we can add a reference to both the status promise (to be attached to `errorBody` if/once that exists) and another to `connection`. It gets a bit convoluted in here. Trying to work through this case is the main reason I wanted to see if we could get by without having `errorBody` at all.

c++/src/kj/compat/http.c++
```diff
@@ -5866,24 +5960,40 @@ public:
...
+    //    writes until the status is indicated by calling accept/reject.
+    auto promise = service.connect(hostCopy,
+                                   *headersCopy,
+                                   tunnel->getTunnelStream(),
+                                   *tunnel)
+        .attach(kj::mv(hostCopy), kj::mv(headersCopy), kj::mv(tunnel));
+
+    return ConnectRequest {
+      forked.addBranch(),
+      pipe.ends[1].attach(kj::mv(promise)),
```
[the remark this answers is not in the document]
jasnell: I still really have not internalized when `eagerlyEvaluate()` is needed, sigh. Thank you for catching that. I'll get it added.

As for the rest... yep... this particular bit is really tricky. For now, I think adding a `TODO(bug)` comment is probably best until the `Promise` returning `end()` is available. I was staring at this bit for a long time yesterday trying to figure out a better way to do this piece without making larger changes and really couldn't come up with anything.

c++/src/kj/compat/http.c++
```diff
@@ -6231,80 +6341,91 @@ private:
...
+      return out.attach(kj::addRef(*this));
+    }
+
+    void respond(uint statusCode, kj::StringPtr statusText, const HttpHeaders& headers) {
+      KJ_REQUIRE(statusTextCopy == nullptr, "the connect request already has a response");
+      statusTextCopy = kj::str(statusText);
+      headersCopy = kj::heap(headers.clone());
+      fulfiller->fulfill(HttpClient::ConnectRequest::Status {
+        statusCode,
+        KJ_ASSERT_NONNULL(statusTextCopy),
```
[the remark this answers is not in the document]
jasnell: Agreed. Do you have a preference for either (a) or (b)? I think I'd be good with either.

c++/src/kj/compat/http.c++
```diff
@@ -6368,61 +6489,54 @@ public:
...
+    // This operates optimistically. In order to support pipelining, we connect the
+    // input and outputs streams immediately, even if we're not yet certain that the
+    // tunnel can actually be established.
+    auto promises = kj::heapArrayBuilder<kj::Promise<void>>(2);
+    auto io1 = kj::mv(request.connection);
+    auto io2 = kj::mv(connection);
+
+    // Here, we immediately set up the outbound pipe... that is, piping from
+    // the passed in connection to the newly created request.connection.
+    // However, we do not set up the pipeline in the other direction until
```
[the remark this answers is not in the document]
jasnell: Yeah, I went around a few times on this and definitely agree that the `errorBody` approach was cleaner.

c++/src/kj/compat/http.c++
```diff
@@ -5785,14 +5880,13 @@ private:
...
-      kj::Promise<ConnectResponse>&& promise,
+  static ConnectRequest attachCounter(
+      ConnectRequest&& request,
       ConnectionCounter&& counter) {
-    return promise.then([counter = kj::mv(counter)](HttpClient::ConnectResponse&& response)
-        mutable -> HttpClient::ConnectResponse {
-      response.connection = response.connection.attach(kj::mv(counter));
-      return kj::mv(response);
-    });
+    // TODO(someday): Should the counter also be attached to the status promise?
```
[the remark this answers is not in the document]
jasnell: If we bring back `errorBody` is this still true? If the status indicates rejected, the `errorBody` effectively replaces the `connection`, which may be dropped. If we don't attach the counter to `errorBody` (by passing a reference into the promise continuation and attaching to `errorBody` if given), then we risk decrementing to counter prematurely. Is that ok?

c++/src/kj/compat/http.c++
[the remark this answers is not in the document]
jasnell: Went ahead and implemented (a).

c++/src/kj/compat/http.c++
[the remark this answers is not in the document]
kentonv: Ah, interesting. I guess the fully correct thing to do would indeed be to refcount and attach it to both. However, for our use case, I don't think counting error bodies is really needed. In Workers, we don't plan to expose CONNECT directly to applications. We do intend to use CONNECT under the hood to implement TCP connections. However, an error response to a `CONNECT` request would be a response from our internal proxy. We would not expose this response to the application anyway, and we'd always consume said error proactively.

So it _is_ important to count successfully-established connections against the connection limit, but it is _not_ important to count error responses, at least for our use case.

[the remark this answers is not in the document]
jasnell: Yep. I think going back to your point about making sure it's right for general purpose (and not just our internal use case), I think I'll go ahead and make ConnectionCounter refcounted since we really can't be certain if someone may want to count error responses (even if we don't).

c++/src/kj/compat/http.c++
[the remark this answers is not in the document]
kentonv: So, for HTTP requests, the idea behind making the headers non-owned is that the headers object would actually be able to point directly into HttpOutputStream's buffer, so the header content never needed to be copied.

I guess that optimization might be difficult to achieve with `CONNECT` when pipelining is supported, though, since the buffer is discarded on the first read, which the client might want to issue before the status promise even resolves. So maybe it's for the best to just not attempt that...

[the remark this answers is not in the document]
jasnell: I think there might be a path to getting there but it's likely not worth spinning wheels on it right now. Just having `Status` own those right now is straightforward. We can revisit this later. I'll add a TODO comment

c++/src/kj/compat/http.c++
kentonv: I'm not entirely clear on the right answer either.

Consider the case where the client has established a connection to the server, and the down-direction pump (from server to client) throws an exception. What should happen?

* We could immediately cancel the up-direction pump as well, and fail out everything.

* We could let the up direction continue, but try to make sure that any `read()`s issued by the client throw an exception indicating the down direction is broken.

The problem right now is we don't do _either_. The `read()` will just hang.

I would say the second option seems better, but I don't think we have any way to implement it with the current stream APIs. What you want here is an `abortWrite()` call, which causes `read()` on the other end of the pipe to throw. But `AsyncIoStream` only has `abortRead()` (causes `write()` on the other end to throw) and `shutdownWrite()` (which signals a clean EOF to the reader -- not what we want!).

Perhaps in addition to adding an `end()` method I really need to add an `abortWrite()` method. But as of now, we don't have that.

So perhaps the only option now is to cancel the opposite direction when one direction fails. I guess the easiest way to do that is using a `kj::Canceler`. It's an annoying amount of bookkeeping, though. It would be nice to fix this properly by adding a `kj::joinPromisesFailfast` to supplement (or perhaps replace) `kj::joinPromises`, and then use that here. (This new function would avoid the lazy evaluation problem, too.)

kentonv: But yeah I'm also OK with `TODO(bug)` for now.

c++/src/kj/compat/http.c++
```diff
@@ -6346,12 +6346,15 @@ private:
...
 
     ~ConnectResponseImpl() noexcept(false) {
-      if (fulfiller->isWaiting()) {
-        fulfiller->reject(KJ_EXCEPTION(DISCONNECTED, "the connect request was incomplete"));
-      }
-      if (streamAndFulfiller.fulfiller->isWaiting()) {
-        streamAndFulfiller.fulfiller->reject(
-            KJ_EXCEPTION(DISCONNECTED, "the connect request was incomplete"));
+      if (fulfiller->isWaiting() || streamAndFulfiller.fulfiller->isWaiting()) {
+        auto ex = KJ_EXCEPTION(DISCONNECTED,
```
kentonv, line 6368: This actually seems more like a FAILED than a DISCONNECTED to me. The server implementation is buggy if it didn't call the callback.

jasnell: Makes sense

c++/src/kj/compat/http.c++
```diff
@@ -6409,6 +6428,17 @@ private:
...
 
+    void handleException(kj::Exception&& ex) {
+      // Log the exception...
+      KJ_LOG(ERROR, "Error in HttpClientAdapter connect()", kj::cp(ex));
+      // Reject the status promise if it is still pending...
+      if (fulfiller->isWaiting()) {
+        fulfiller->reject(kj::cp(ex));
+      }
+      // Cause the stream's read() to throw...
+      kj::throwRecoverableException(kj::mv(ex));
```
kentonv, line 6428: It looks like throwing from here doesn't actually do anything, because the promise that calls this is simply `attach()`ed to the pipe end and no one ever listens for the result.

c++/src/kj/compat/http.c++
```diff
@@ -5978,16 +5978,35 @@ public:
...
+        .eagerlyEvaluate([&response=*response](kj::Exception&& ex) {
+      // A few things need to happen here.
+      //   1. We'll log the exception.
+      //   2. We'll break the pipe.
+      //   3. We'll reject the status promise if it is still pending.
+      //
+      // We'll do all of this within the ConnectResponseImpl, however, since it
+      // maintains the state necessary here.
+      response.handleException(kj::mv(ex));
+    }).attach(kj::mv(hostCopy), kj::mv(headersCopy), kj::mv(response), kj::mv(connectStream));
```
kentonv, line 5989: Note that lazy evaluation applies to destroying these attachments, too. They won't be destroyed until the pipe end that the promise is attached to is destroyed.

One thing you could do is, instead of using `.attach()`, actually capture all these values by move into the `eagerlyEvaluate()` callback. That way they will all be destroyed after that callback returns (or if it's canceled). It also conveniently avoids allocation of an extra `AttachmentPromiseNode`.

c++/src/kj/compat/http.c++
jasnell: Yep, once I was through adding the `errorBody` back in and reworking a bunch of this I was going to ask what the best way to handle this particular error is as I'm not really sure about the best way to propagate it. Specifically, given how things are wired up here, what is the best way to propagate the error to reads on the connection stream? I was thinking a kj::Canceler but is there another way?

c++/src/kj/compat/http.c++
jasnell: oh, right, good point... and good idea.

c++/src/kj/compat/http.c++
```diff
@@ -5974,7 +5973,7 @@ public:
     // 2. Create a promise/fulfiller pair for the status. The promise will be
     //    returned with the ConnectResponse, the fulfiller will be held by the
     //    ConnectResponseImpl.
-    auto paf = kj::newPromiseAndFulfiller<ConnectRequest::Status>();
+    auto paf = kj::newPromiseAndFulfiller<kj::Own<ConnectRequest::Status>>();
     auto forked = paf.promise.fork();
```
kentonv, line 5969: It looks like only one branch is ever created off this fork? Is the fork actually needed?

c++/src/kj/compat/http.h
```diff
@@ -718,13 +718,24 @@ class HttpClient {
   // stack-allocated).
 
   struct ConnectRequest {
-    struct Status {
+    struct Status : public kj::Refcounted {
```
kentonv, line 721: Converting this struct into a heap-allocated refcounted type is kind of sad. I guess you did this to allow for `fork()`ing the promise, but as I noticed above, it doesn't look like the `fork()` is actually needed. Can we remove it and then turn this back into a value type?

c++/src/kj/compat/http.c++
```diff
@@ -5796,13 +5796,15 @@ private:
   uint concurrentRequests = 0;
   kj::Function<void(uint runningCount, uint pendingCount)> countChangedCallback;
 
-  std::queue<kj::Own<kj::PromiseFulfiller<ConnectionCounter>>> pendingRequests;
+  std::queue<kj::Own<kj::PromiseFulfiller<kj::Own<ConnectionCounter>>>> pendingRequests;
   // TODO(someday): want maximum cap on queue size?
 
-  struct ConnectionCounter final {
+  struct ConnectionCounter final: public kj::Refcounted {
```
kentonv, line 5802: This commit seems to add a lot of additional heap allocation which is kind of sad.

TBH I would be just fine with not counting `errorBody` here. I don't think this is a case where being "correct for general use" is important because, to be honest, `ConcurrencyLimitingHttpClient` was build for our very specific use case in the first place, and probably shouldn't even be in KJ. I'd be OK with moving this entire class into the Workers codebase...

c++/src/kj/compat/http.h
jasnell: Yeah I'm not happy with this either. Once I finish working through adding errorBody back in I'm going to change this up. I'm pretty sure the forks aren't needed now.

c++/src/kj/compat/http.c++
jasnell: Indeed, it is pretty gross. If we can move `ConcurrencyLimitingHttpClient` then definite +1 to only attaching the counter to the connection stream. I'll do that here then maybe next week I'll open a separate set of PRs that moves it.

c++/src/kj/compat/http.c++
```diff
@@ -6177,74 +6338,118 @@ private:
...
+    StreamsAndFulfiller streamAndFulfiller;
+    bool connectStreamDetached = false;
+
+    StreamsAndFulfiller initStreamsAndFulfiller(kj::Own<kj::AsyncIoStream> stream) {
+      auto paf = kj::newPromiseAndFulfiller<void>();
+      auto wrapped = kj::Own<kj::AsyncIoStream>(stream.get(), kj::NullDisposer::instance);
+      auto guarded = kj::heap<AsyncIoStreamWithGuards>(
+          kj::mv(wrapped),
+          kj::Maybe<HttpInputStreamImpl::ReleasedBuffer>(nullptr),
+          kj::mv(paf.promise)).attach(kj::addRef(*this));
```
dom96, line 6400: This code is called in the constructor of `ConnectResponseImpl` which means that if `ConnectResponseImpl` is allocated using `kj::refcounted` we end up with an error here because the refcounted didn't finish incrementing the ref count (or at least that's what I believe happens).

In any case I get a crash here because of that. Wonder why the tests didn't pick this up.

jasnell: Ugh... doh, that's right.  Not sure why it didn't pick it up. Very odd. Ok, I'll update this so that setting the stream happens in a separate call after allocating the instance.

kentonv:
I have no idea what has changed in this PR since the last review. All record of what I reviewed before is now gone. :(

kentonv:
Maybe this is it? https://github.com/capnproto/capnproto/compare/105517a..120a18a7b26a95e6d3628c8f0bbd322de4b5084d

c++/src/kj/compat/http.c++
```diff
@@ -4956,9 +5136,12 @@ public:
 
     auto id = ++counter;
 
-    return httpInput.readResponseHeaders().then(
-        [this,id](HttpHeaders::ResponseOrProtocolError&& responseOrProtocolError)
-            -> HttpClient::ConnectResponse {
+    auto paf = kj::newPromiseAndFulfiller<kj::Maybe<HttpInputStreamImpl::ReleasedBuffer>>();
```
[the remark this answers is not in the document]
kentonv: Yeah `split()` is pretty much the only place where I willingly use `Tuple`...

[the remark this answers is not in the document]
jasnell: Updated, but can you verify the use here. Specifically, I'm returning exceptions in two cases as part of the tuple to cause the read guard to be rejected and want to be certain that's the correct approach using tuple and split.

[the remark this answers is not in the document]
kentonv: Yeah I guess that works. It's a little funny that you end up returning a `Promise` that always ends up being immediately fulfilled but I guess that's the only way to make one branch of the split error without erroring the other branch.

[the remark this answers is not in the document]
jasnell: Yeah, that's precisely the kind of weirdness I was trying to avoid by eliminating the split but ah well... we'll go with this unless you happen to think of a better suggestion

kentonv:
LGTM, please squash.

I guess we need to prepare changes in the Workers codebase to account for the interface change before we merge?

jasnell:
Had to fix a minor error in the http-tests so force-pushed a new commit. No other changes were made. This is almost ready to go, just need to do a bit more testing

jasnell:
Ok, PRs in Workerd and Workers are ready to go and have been tested in CI and lab. I think we're good to get this landed, after which we'll get the workerd PR landed, followed by the internal Workers PR

c++/src/kj/compat/http.h
```diff
@@ -717,20 +717,42 @@ class HttpClient {
...
+             kj::Own<HttpHeaders> headers,
+             kj::Maybe<kj::Own<kj::AsyncInputStream>> errorBody = nullptr)
+          : statusCode(statusCode),
+            statusText(kj::mv(statusText)),
+            headers(kj::mv(headers)),
+            errorBody(kj::mv(errorBody)) {}
+
+      Status(Status&&) = default;
+      Status& operator=(Status&&) = default;
+      KJ_DISALLOW_COPY(Status);
```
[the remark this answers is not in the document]
jasnell: I'll remove them. That said, Even if not strictly necessary I do tend to prefer to be more explicit.

c++/src/kj/compat/http.c++
```diff
@@ -4666,6 +4806,40 @@ public:
   }
 };
 
+class ReadOnlyIoStream final: public kj::AsyncIoStream {
```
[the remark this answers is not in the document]
jasnell: Ah right, meant to pull this back out but forgot.

jasnell:
image

Wtf Github... you silly.

c++/src/kj/compat/http.c++
```diff
@@ -4093,6 +4096,143 @@ private:
...
+    return guard.then([this](kj::Maybe<HttpInputStreamImpl::ReleasedBuffer> buffer) mutable {
+      readGuardReleased = true;
+      KJ_IF_MAYBE(b, buffer) {
+        if (b->leftover.size() > 0) {
+          // We only need to replace the inner stream if a non-empty buffer is provided.
+          inner = heap<AsyncIoStreamWithInitialBuffer>(
+              kj::mv(inner),
+              kj::mv(b->buffer), b->leftover);
+        } else {
+          inner = inner.attach(kj::mv(b->buffer));
```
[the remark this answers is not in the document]
harrishancock: > Possibly, I suppose?

FWIW, I only mention it for the future, if/when we try to optimize away the extra heap allocations.

harrishancock (approved):

jasnell:
@kentonv or @harrishancock ... can one of you get this merged now so I can proceed with getting the dependent PRs merged

c++/src/kj/compat/http.c++
```diff
@@ -6177,74 +6291,115 @@ private:
...
     }
 
-  private:
-    kj::Own<kj::PromiseFulfiller<HttpClient::ConnectResponse>> fulfiller;
-    kj::Promise<void> task = nullptr;
+    void respond(uint statusCode,
+                 kj::StringPtr statusText,
+                 const HttpHeaders& headers,
+                 kj::Maybe<kj::Own<kj::AsyncInputStream>> errorBody = nullptr) {
+      KJ_REQUIRE(fulfiller->isWaiting(), "the connect request already has a response");
```
kentonv, line 6388: This is not an appropriate way to check if the fulfiller was already fulfilled -- this will also fail if the associated promise has been dropped (and thus is no longer waiting on the fulfiller). We should remove this line.

dom96: Nice find! Can confirm, this resolves the issue I was seeing

c++/src/kj/compat/http-test.c++
```diff
@@ -6220,114 +6231,26 @@ KJ_TEST("CONNECT Server cancels read w/client") {
...
-    KJ_ASSERT(response.statusText == "OK"_kj);
-
-    auto& io = KJ_ASSERT_NONNULL(response.connectionOrBody.tryGet<kj::Own<kj::AsyncIoStream>>());
-    char buffer;
-    // Attempt a read. It should disconnect. We are verifying here that the stream is shutdown
-    // correctly, if it doesn't this will block forever (and KJ will give us a
-    // "this thread would hang forever" error message).
-    return io->tryRead(&buffer, 1, 1).then([&](size_t count) {
-      KJ_ASSERT(count == 0, "Expecting the stream to get disconnected.");
-      readPaf.fulfiller->fulfill();
```
dom96, line 6322: Looks like this test was removed. Was this intentional? It was added to ensure the CONNECT implementation handles stream closure correctly.

c++/src/kj/compat/http.c++
```diff
@@ -6307,31 +6462,69 @@ public:
...
+      connection.shutdownWrite();
+    }).eagerlyEvaluate(nullptr));
+
+    promises.add(request.status.then(
+        [&response,&io=*io,&connection,fulfiller=kj::mv(paf.fulfiller)]
+        (HttpClient::ConnectRequest::Status status) mutable -> kj::Promise<void> {
+      if (status.statusCode >= 200 && status.statusCode < 300) {
+        // Release the read guard!
+        fulfiller->fulfill(kj::Maybe<HttpInputStreamImpl::ReleasedBuffer>(nullptr));
+        response.accept(status.statusCode, status.statusText, *status.headers);
```
harrishancock, line 6503: Do we need a `return kj::READY_NOW` statement here to avoid the `KJ_UNREACHABLE` below?

dom96: I've added it in https://github.com/capnproto/capnproto/pull/1586

PR #1580, opened 2022-11-16
Fix restrictPeers allowing public+private, network+local, public+local.

kentonv:
These were buggy because allowing "public" and "network" were incorrectly implemented by adding 0.0.0.0/0 to the allow list and then adding non-public and non-network ranges to the deny list. If a separate rule tried to allow those ranges, the deny rules would take precedence since they had the same specificity.

Instead we special-case "public" and "network" as boolean flags instead of trying to represent them as CIDR unions.

Fixes https://github.com/cloudflare/workerd/issues/62

cc @dom96 (GitHub won't let me set him as reviewer)

### DIFF
--- a/c++/src/kj/async-io-internal.h
+++ b/c++/src/kj/async-io-internal.h
@@ -25,6 +25,7 @@
 #include "vector.h"
 #include "async-io.h"
 #include <stdint.h>
+#include "one-of.h"
 
 KJ_BEGIN_HEADER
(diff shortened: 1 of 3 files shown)


dom96 (approved):
Looks good!

harrishancock (approved):

Warfields (approved):
:+1:

PR #1582, opened 2022-11-20
Add capnpc-json for dumping code generator requests in json format

ObsidianMinor:
When bootstrapping a new Cap'n Proto implementation in another language, it can be useful to make the first code-generator in the same language as the language you're planning to write the code generator in (of course). But until the schema.capnp file has been bootstrapped at least once for the library being developed, this is difficult.

You may bootstrap a simple compiler in another language like C++, but then you won't be able to reuse any of the code and be forced to rewrite from scratch in the target language once up and running.

It'd be useful if another included output format existed like capnpc-capnp, to dump the input code generator request in another format that's easier to bootstrap in the target language. JSON is decently popular enough that it's well supported basically everywhere.

This PR adds capnpc-json, an included output plugin that simply dumps the entire code generator request in JSON format to stdout.

### DIFF
--- a/c++/src/capnp/BUILD.bazel
+++ b/c++/src/capnp/BUILD.bazel
@@ -157,6 +157,18 @@ cc_binary(
     ],
 )
 
+cc_binary(
+    name = "capnpc-json",
+    srcs = [
+        "compiler/capnpc-json.c++",
+    ],
+    visibility = ["//visibility:public"],
+    deps = [
+        ":capnpc",
+        "//src/capnp/compat:json",
+    ],
+)
+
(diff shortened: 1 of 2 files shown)


ObsidianMinor:
Note using `capnp convert` _does not work_ for bootstrapping schema.capnp since it includes AnyPointer fields which JsonCodec doesn't handle by default and `convert` doesn't include a handler for. This output does handle AnyPointer by encoding the value to base64.

kentonv:
Why not add a flag to `capnp convert` controlling how JSON handles AnyPointer fields?

PR #1583, opened 2022-11-22
Add more tests for BufferedMessageStream.

kentonv:
These come from PR #1528, but are being used to test the code from #1572.

Sorry, I meant to do this before merging #1572 but then forgot.

The only changes I made are:
* Implement `makeBufferedStream` in the test file itself.
* Replace use of `BufferedStreamOptions` with bare sizes (in words rather than bytes). Note that in the test that used a 7-byte buffer, I had to increase it to 1 word even, since buffer sizes are always in words now. Not sure if this defeats the purpose of the test.

### DIFF
--- a/.github/workflows/quick-test.yml
+++ b/.github/workflows/quick-test.yml
@@ -11,7 +11,7 @@ on:
 
 jobs:
   Linux-musl:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     container: alpine:latest
     steps:
       - uses: actions/checkout@v2
@@ -20,7 +20,7 @@ jobs:
       - name: super-test
         run: ./super-test.sh quick
   Linux:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     strategy:
       fail-fast: false
       matrix:
@@ -35,7 +35,7 @@ jobs:
         run: |
             ./super-test.sh quick ${{ matrix.compiler }}
   Linux-lock-tracking:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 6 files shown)


kentonv:
Apparently the linux-musl build broke just this afternoon as I was trying to debug the mac build. Suddenly the compiler is reporting a new (bogus) warning, which is being treated as an error.

The mac build, meanwhile... fails until I add this KJ_DBG line, then it succeeds? WTF?

kentonv:
I think GitHub may have switched `ubuntu-latest` from 20.04 to 22.04 just in the last... 15 minutes.

kentonv:
Of course, now everything is broken.

kentonv:
OK, so...

`sendmsg()` is failing with `EMSGSIZE` on MacOS only. I figured this was because the one test was writing messages with FDs attached that were too big, so I tried adding code to detect when FDs are attached to a big message and only attach them to the start of the message instead. This didn't solve the problem. But when I added a non-side-effecting KJ_DBG in there to get some information, suddenly the problem went away.

Looking closer at the tests, I noticed:
* They use threads.
* They use `PipeWithSmallBuffer`, which intentionally creates a socketpair with a really small kernel-sized buffer. Per the comments, MacOS, unlike Linux, will honor extremely small buffer size requests -- 127 bytes, in this case.

My guess is that the small buffer is part of the problem here, but also, because of the use of threads, a small delay introduced by KJ_DBG can give the other thread enough time to read from the buffer, which makes the problem go away.

It's not clear to me why these tests are using `PipeWithSmallBuffer`. In fact, using this seems like it would actually interfere with testing BufferedMessageStream, since you can't effectively fill the userspace buffer if the kernel buffer is tiny. Did the tests use `PipeWithSmallBuffer` just because other tests in the file did, or is the small kernel buffer actually important to the test?

It looks like the pattern of using threads may also have come from other tests in the file. IMO the use of threads isn't helpful here; it probably just makes the tests less predictable. In fact, the tests would probably be better off not even using a kernel socketpair, but rather a `kj::newCapabilityPipe()`.

Anyway... I've spent too many hours trying to debug this and am going to move on for now.

PR #1586, opened 2022-11-24
Implements HTTP Connect in http-over-capnp.

dom96:
### Test Plan

In internal repo for ASAN:

```
$ bazel test //src/ew/tests:sockets/sockets-tcp-test.ew-test-bin@force-process-sandboxing
$ bazel test @capnp-cpp//src/capnp/compat:http-over-capnp-test --config=asan --test_timeout 5
```

### DIFF
--- a/c++/src/capnp/compat/http-over-capnp-test.c++
+++ b/c++/src/capnp/compat/http-over-capnp-test.c++
@@ -648,5 +648,195 @@ KJ_TEST("HttpService isn't destroyed while call outstanding") {
   KJ_EXPECT(!destroyed);
 }
 
+
+class ConnectWriteCloseService final: public kj::HttpService {
+  // A simple CONNECT server that will accept a connection, write some data and close the
+  // connection.
+public:
+  ConnectWriteCloseService(kj::HttpHeaderTable& headerTable)
+      : headerTable(headerTable) {}
+
+  kj::Promise<void> request(
+      kj::HttpMethod method, kj::StringPtr url, const kj::HttpHeaders& headers,
+      kj::AsyncInputStream& requestBody, kj::HttpService::Response& response) override {
+    KJ_UNIMPLEMENTED("Regular HTTP requests are not implemented here.");
+  }
+
+  kj::Promise<void> connect(
+      kj::StringPtr host, const kj::HttpHeaders& headers, kj::AsyncIoStream& io,
+      kj::HttpService::ConnectResponse& response) override {
+    response.accept(200, "OK", kj::HttpHeaders(headerTable));
+    return io.write("test", 4).then([&io]() mutable {
+      io.shutdownWrite();
+    });
+  }
+
+private:
+  kj::HttpHeaderTable& headerTable;
+};
+
+class ConnectRejectService final: public kj::HttpService {
+  // A simple CONNECT server that will reject a connection.
+public:
+  ConnectRejectService(kj::HttpHeaderTable& headerTable)
+      : headerTable(headerTable) {}
+
+  kj::Promise<void> request(
+      kj::HttpMethod method, kj::StringPtr url, const kj::HttpHeaders& headers,
+      kj::AsyncInputStream& requestBody, kj::HttpService::Response& response) override {
+    KJ_UNIMPLEMENTED("Regular HTTP requests are not implemented here.");
+  }
+
+  kj::Promise<void> connect(
+      kj::StringPtr host, const kj::HttpHeaders& headers, kj::AsyncIoStream& io,
+      kj::HttpService::ConnectResponse& response) override {
+    auto body = response.reject(500, "Internal Server Error", kj::HttpHeaders(headerTable), 5);
+    return body->write("Error", 5).attach(kj::mv(body));
+  }
+
+private:
+  kj::HttpHeaderTable& headerTable;
+};
+
+KJ_TEST("HTTP-over-Cap'n-Proto Connect with close") {
+  kj::EventLoop eventLoop;
+  kj::WaitScope waitScope(eventLoop);
+
+  auto pipe = kj::newTwoWayPipe();
+
+  kj::TimerImpl timer(kj::origin<kj::TimePoint>());
+
+  ByteStreamFactory streamFactory;
+  kj::HttpHeaderTable::Builder tableBuilder;
+  HttpOverCapnpFactory factory(streamFactory, tableBuilder);
+  kj::Own<kj::HttpHeaderTable> table = tableBuilder.build();
+  ConnectWriteCloseService service(*table);
+  kj::HttpServer server(timer, *table, service);
+
+  auto listenTask = server.listenHttp(kj::mv(pipe.ends[0]));
+
+  auto client = newHttpClient(*table, *pipe.ends[1]);
+
+  capnp::HttpService::Client httpService = factory.kjToCapnp(newHttpService(*client));
+  auto frontCapnpHttpService = factory.capnpToKj(httpService);
+
+  struct ResponseImpl final: public kj::HttpService::ConnectResponse {
+    kj::Own<kj::PromiseFulfiller<kj::HttpClient::ConnectRequest::Status>> fulfiller;
+    ResponseImpl(kj::Own<kj::PromiseFulfiller<kj::HttpClient::ConnectRequest::Status>> fulfiller)
+      : fulfiller(kj::mv(fulfiller)) {}
+    void accept(uint statusCode, kj::StringPtr statusText, const kj::HttpHeaders& headers) override {
+      KJ_REQUIRE(statusCode >= 200 && statusCode < 300, "the statusCode must be 2xx for accept");
+      fulfiller->fulfill(
+        kj::HttpClient::ConnectRequest::Status(
+          statusCode,
+          kj::str(statusText),
+          kj::heap(headers.clone()),
+          nullptr
+        )
+      );
+    }
+
+    kj::Own<kj::AsyncOutputStream> reject(
+        uint statusCode,
+        kj::StringPtr statusText,
+        const kj::HttpHeaders& headers,
+        kj::Maybe<uint64_t> expectedBodySize) override {
+      KJ_UNREACHABLE;
+    }
+  };
+
+  auto clientPipe = kj::newTwoWayPipe();
+  auto paf = kj::newPromiseAndFulfiller<kj::HttpClient::ConnectRequest::Status>();
+  ResponseImpl response(kj::mv(paf.fulfiller));
+
+  auto promise = frontCapnpHttpService->connect(
+      "https://example.org"_kj, kj::HttpHeaders(*table), *clientPipe.ends[0],
+      response).attach(kj::mv(clientPipe.ends[0]));
+
+  paf.promise.then(
+      [io = kj::mv(clientPipe.ends[1])](auto status) mutable {
+    KJ_ASSERT(status.statusCode == 200);
+    KJ_ASSERT(status.statusText == "OK"_kj);
+
+    auto buf = kj::heapArray<char>(4);
+    return io->tryRead(buf.begin(), 4, 4).then(
+        [buf = kj::mv(buf), io = kj::mv(io)](size_t count) mutable {
+      KJ_ASSERT(count == 4, "Expecting the stream to read 4 chars.");
+      return io->tryRead(buf.begin(), 1, 1).then(
+          [buf = kj::mv(buf)](size_t count) mutable {
+        KJ_ASSERT(count == 0, "Expecting the stream to get disconnected.");
+      }).attach(kj::mv(io));
+    });
+  }).wait(waitScope);
+
+  listenTask.wait(waitScope);
+}
+
+
+KJ_TEST("HTTP-over-Cap'n-Proto Connect Reject") {
+  kj::EventLoop eventLoop;
+  kj::WaitScope waitScope(eventLoop);
+
+  auto pipe = kj::newTwoWayPipe();
+
+  kj::TimerImpl timer(kj::origin<kj::TimePoint>());
+
+  ByteStreamFactory streamFactory;
+  kj::HttpHeaderTable::Builder tableBuilder;
+  HttpOverCapnpFactory factory(streamFactory, tableBuilder);
+  kj::Own<kj::HttpHeaderTable> table = tableBuilder.build();
+  ConnectRejectService service(*table);
+  kj::HttpServer server(timer, *table, service);
+
+  auto listenTask = server.listenHttp(kj::mv(pipe.ends[0]));
+
+  auto client = newHttpClient(*table, *pipe.ends[1]);
+
+  capnp::HttpService::Client httpService = factory.kjToCapnp(newHttpService(*client));
+  auto frontCapnpHttpService = factory.capnpToKj(httpService);
+
+  struct ResponseImpl final: public kj::HttpService::ConnectResponse {
+    kj::Own<kj::PromiseFulfiller<kj::Own<kj::AsyncInputStream>>> fulfiller;
+    ResponseImpl(kj::Own<kj::PromiseFulfiller<kj::Own<kj::AsyncInputStream>>> fulfiller)
+      : fulfiller(kj::mv(fulfiller)) {}
+    void accept(uint statusCode, kj::StringPtr statusText, const kj::HttpHeaders& headers) override {
+      KJ_UNREACHABLE;
+    }
+
+    kj::Own<kj::AsyncOutputStream> reject(
+        uint statusCode,
+        kj::StringPtr statusText,
+        const kj::HttpHeaders& headers,
+        kj::Maybe<uint64_t> expectedBodySize) override {
+      KJ_ASSERT(statusCode == 500);
+      KJ_ASSERT(statusText == "Internal Server Error");
+      KJ_ASSERT(expectedBodySize.orDefault(5));
+      auto pipe = kj::newOneWayPipe();
+      fulfiller->fulfill(kj::mv(pipe.in));
+      return kj::mv(pipe.out);
+    }
+  };
+
+  auto clientPipe = kj::newTwoWayPipe();
+  auto paf = kj::newPromiseAndFulfiller<kj::Own<kj::AsyncInputStream>>();
+  ResponseImpl response(kj::mv(paf.fulfiller));
+
+  auto promise = frontCapnpHttpService->connect(
+      "https://example.org"_kj, kj::HttpHeaders(*table), *clientPipe.ends[0],
+      response).attach(kj::mv(clientPipe.ends[0]));
+
+  paf.promise.then(
+      [](auto body) mutable {
+    auto buf = kj::heapArray<char>(5);
+    return body->tryRead(buf.begin(), 5, 5).then(
+        [buf = kj::mv(buf), body = kj::mv(body)](size_t count) mutable {
+      KJ_ASSERT(count == 5, "Expecting the stream to read 5 chars.");
+    });
+  }).attach(kj::mv(promise)).wait(waitScope);
+
+  listenTask.wait(waitScope);
+}
+
+
 }  // namespace
 }  // namespace capnp
--- a/c++/src/capnp/compat/http-over-capnp.c++
+++ b/c++/src/capnp/compat/http-over-capnp.c++
@@ -323,6 +323,56 @@ private:
   // Must check state->assertNotCanceled() before using this.
 };
 
+class HttpOverCapnpFactory::ConnectClientRequestContextImpl final
+    : public capnp::HttpService::ConnectClientRequestContext::Server {
+public:
+  ConnectClientRequestContextImpl(HttpOverCapnpFactory& factory,
+      kj::HttpService::ConnectResponse& connResponse)
+      : factory(factory), connResponse(connResponse) {}
+
+  kj::Promise<void> startConnect(StartConnectContext context) override {
+    KJ_REQUIRE(!sent, "already called startConnect() or startError()");
+    sent = true;
+
+    auto params = context.getParams();
+    auto resp = params.getResponse();
+
+    auto headers = factory.headersToKj(resp.getHeaders());
+    connResponse.accept(resp.getStatusCode(), resp.getStatusText(), headers);
+
+    return kj::READY_NOW;
+  }
+
+  kj::Promise<void> startError(StartErrorContext context) override {
+    KJ_REQUIRE(!sent, "already called startConnect() or startError()");
+    sent = true;
+
+    auto params = context.getParams();
+    auto resp = params.getResponse();
+
+    auto headers = factory.headersToKj(resp.getHeaders());
+
+    auto bodySize = resp.getBodySize();
+    kj::Maybe<uint64_t> expectedSize;
+    if (bodySize.isFixed()) {
+      expectedSize = bodySize.getFixed();
+    }
+
+    auto stream = connResponse.reject(
+        resp.getStatusCode(), resp.getStatusText(), headers, expectedSize);
+
+    context.initResults().setBody(factory.streamFactory.kjToCapnp(kj::mv(stream)));
+
+    return kj::READY_NOW;
+  }
+
+private:
+  HttpOverCapnpFactory& factory;
+  bool sent = false;
+
+  kj::HttpService::ConnectResponse& connResponse;
+};
+
 class HttpOverCapnpFactory::KjToCapnpHttpServiceAdapter final: public kj::HttpService {
 public:
   KjToCapnpHttpServiceAdapter(HttpOverCapnpFactory& factory, capnp::HttpService::Client inner)
@@ -415,6 +465,44 @@ public:
           [](auto& pipeline) { return pipeline.ignoreResult(); });
     }
   }
+
+  kj::Promise<void> connect(
+      kj::StringPtr host, const kj::HttpHeaders& headers, kj::AsyncIoStream& connection,
+      ConnectResponse& tunnel) override {
+    auto rpcRequest = inner.connectRequest();
+    auto downPipe = kj::newOneWayPipe();
+    rpcRequest.setHost(host);
+    rpcRequest.setDown(factory.streamFactory.kjToCapnp(kj::mv(downPipe.out)));
+
+    auto builder = capnp::Request<
+        capnp::HttpService::ConnectParams,
+        capnp::HttpService::ConnectResults>::Builder(rpcRequest);
+    rpcRequest.adoptHeaders(factory.headersToCapnp(headers,
+        Orphanage::getForMessageContaining(builder)));
+    rpcRequest.setContext(
+        kj::heap<ConnectClientRequestContextImpl>(factory, tunnel));
+    RemotePromise<capnp::HttpService::ConnectResults> pipeline = rpcRequest.send();
+
+    // We read from `downPipe` (the other side writes into it.)
+    auto downPumpTask = downPipe.in->pumpTo(connection)
+        .then([&connection, down = kj::mv(downPipe.in)](uint64_t) -> kj::Promise<void> {
+      connection.shutdownWrite();
+      return kj::NEVER_DONE;
+    });
+    // We write to `up` (the other side reads from it).
+    auto up = pipeline.getUp();
+    kj::Own<kj::AsyncOutputStream> upStream = factory.streamFactory.capnpToKj(up);
+    auto upPumpTask = connection.pumpTo(*upStream)
+        .then([up = kj::mv(up), upStream = kj::mv(upStream)](uint64_t) mutable
+        -> kj::Promise<void> {
+      return kj::NEVER_DONE;
+    });
+
+    return pipeline.ignoreResult()
+        .attach(kj::mv(downPumpTask))
+        .attach(kj::mv(upPumpTask));
+  }
+
 
 private:
   HttpOverCapnpFactory& factory;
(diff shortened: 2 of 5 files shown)


c++/src/kj/compat/http.c++
```diff
@@ -6199,15 +6199,17 @@ private:
...
 
-      kj::Own<kj::AsyncIoStream> io =
-          kj::heap<DelayedEofIoStream>(kj::mv(pipe.ends[0]),
-              task.attach(kj::addRef(*this), kj::mv(headersCopy)));
+      // kj::Own<kj::AsyncIoStream> io =
+      //     kj::heap<DelayedEofIoStream>(kj::mv(pipe.ends[0]),
+      //         task.attach(kj::addRef(*this), kj::mv(headersCopy)));
 
-      fulfiller->fulfill(HttpClient::ConnectResponse {
+      // TODO: Should we make the above customisable?
```
dom96, line 6206: Running into the same thing as in https://github.com/capnproto/capnproto/pull/1567/files#r1013083036 here.

I worked around this by creating my own custom implementation of `HttpService::ConnectResponse` which is essentially a clone of this code but without the `DelayedEofIoStream` (you can see that here in my workerd PR: https://github.com/cloudflare/workerd/pull/162/files#diff-4e2699e03d914d2336177c99e165ed8e7f912f9bb61af226be284e62c3324f6eR12).

So since I would have to do the same thing here, I'm wondering whether I can instead add a boolean flag to disable the `DelayedEofIoStream`. That way saving code duplication.

Thinking about it some more though... I'm probably better off moving the `TCPConnectResponseImpl` here.

dom96: This is what I ended up doing. @kentonv any problems with doing it this way?

c++/src/capnp/compat/http-over-capnp.c++
```diff
@@ -688,6 +753,61 @@ public:
...
+    auto host = params.getHost();
+    auto headers = factory.headersToKj(params.getHeaders());
+    auto upPipe = kj::newOneWayPipe();
+
+    PipelineBuilder<ConnectResults> pb;
+    pb.setUp(factory.streamFactory.kjToCapnp(kj::mv(upPipe.out)));
+
+    context.setPipeline(pb.build());
+
+    auto client = kj::newHttpClient(*inner);
```
kentonv, line 767: Hmm I hate to say this but we really shouldn't be using a client adapter here. The http-over-capnp interface is really most analogous to `HttpService` rather than `HttpClient`, so trying to implement it in terms of `HttpClient` is probably making the code below significantly more complex than it would otherwise be. Additionally, using a client adapter adds a fair amount of overhead inside the adapter, as well as at least one known bug (the fact that it cancels the service prematurely at the end of the stream).

If you look at `request()`/`startRequest()`, they are implemented directly in terms of `HttpService`, we should use the same model here.

Sorry, I really should have noticed this when this was a draft... I remember you had a few problems that involved the client adapter but for some reason I didn't connect the dots.

c++/src/capnp/compat/http-over-capnp.c++
```diff
@@ -323,6 +323,37 @@ private:
...
+    sent = true;
+
+    auto params = context.getParams();
+    auto resp = params.getResponse();
+
+    auto headers = factory.headersToKj(resp.getHeaders());
+    if (resp.getIsAccept()) {
+      connResponse.accept(resp.getStatusCode(), resp.getStatusText(), headers);
+    } else {
+      connResponse.reject(resp.getStatusCode(), resp.getStatusText(), headers);
```
kentonv, line 344: This returns a stream, to which you're supposed to write the error body, but instead you're discarding it.

Looking at the server side, it appears that you are writing the error body into the socket itself. I guess that was based on an earlier version of James' PR, that took that approach. However, we decided this was confusing -- the application is not expecting an error body (e.g., some HTML error page) to appear on the connection stream where it was expecting to speak some other protocol (e.g., PostgreSQL protocol). We need to keep these streams separtae.

`ConnectClientRequestContext` should probably have a second method for this purpose, like:

```
startError @0 (response :HttpResponse) -> (body :ByteStream);
```

Alternatively, perhaps `ConnectClientRequestContext` should actually be merged with `ClientRequestContext`. For the error case, you would use the regular `startRespnose()`. `startConnect()` would then be similar to `startWebSocket()` -- a method that is only called if the request was a valid connect request.

kentonv (changes requested):

c++/src/capnp/compat/http-over-capnp.c++
dom96: I will keep the request contexts separate, mainly because there are two different HttpService::Response types for each (`connect` uses `HttpService::ConnectResponse`).

dom96:
Updated and ready for review. There are a number of things I am unsure about, in particular the "reject" handling, so there are likely a few things that are wrong here but hopefully you can give me tips and after another iteration this should be good to go :)

c++/src/capnp/compat/http-over-capnp.c++
```diff
@@ -323,6 +323,48 @@ private:
...
+    KJ_REQUIRE(!sent, "already called startConnect() or startError()");
+    sent = true;
+
+    auto params = context.getParams();
+    auto resp = params.getResponse();
+
+    auto headers = factory.headersToKj(resp.getHeaders());
+    auto stream = connResponse.reject(resp.getStatusCode(), resp.getStatusText(), headers);
+
+    context.initResults().setBody(factory.streamFactory.kjToCapnp(kj::mv(stream)));
```
kentonv, line 356: Be sure to pass a value for `expectedBodySize`, which you should be able to derive from resp.getBodySize().

c++/src/capnp/compat/http-over-capnp.c++
```diff
@@ -688,6 +825,80 @@ public:
...
+        auto rpcResponse = req.initResponse();
+        rpcResponse.setStatusCode(status.statusCode);
+        rpcResponse.setStatusText(status.statusText);
+        rpcResponse.adoptHeaders(factory.headersToCapnp(
+            headers, Orphanage::getForMessageContaining(rpcResponse)));
+
+        return req.send().then(
+            [this, status = kj::mv(status)](auto resp) mutable -> kj::Promise<void> {
+          KJ_IF_MAYBE(errStream, status.errorBody) {
+            KJ_DBG("Got error body");
```
kentonv, line 868: Oops. :)

c++/src/capnp/compat/http-over-capnp.c++
```diff
@@ -688,6 +825,80 @@ public:
...
+                          connectStream=kj::mv(pipe.ends[0])](kj::Exception&& ex) mutable {
+        response->handleException(kj::mv(ex), kj::mv(connectStream));
+      });
+
+    return paf.promise.then(
+        [this, upPipe = kj::mv(upPipe), down = params.getDown(), context = kj::mv(context),
+        io = kj::mv(pipe.ends[1])](auto status) mutable {
+      auto headers = status.headers->clone();
+
+      if (status.statusCode < 200 || status.statusCode > 299) {
```
kentonv, line 857: Maybe this should be `KJ_IF_MAYBE(errorBody, status.errorBody) {`? Seems a bit safer than relying on status code conventions, plus you can avoid the check later.

c++/src/capnp/compat/http-over-capnp.c++
```diff
@@ -688,6 +825,80 @@ public:
...
+        auto req = context.getParams().getContext().startErrorRequest();
+        auto rpcResponse = req.initResponse();
```
kentonv, line 859: Be sure to initialize `bodySize`, too, based on `errorBody->tryGetLength()`.

kentonv (commented):
Can you please add tests to `http-over-capnp-test.c++`?

c++/src/capnp/compat/http-over-capnp.c++
kentonv: (This comment was intended to be on the call to `connResponse.reject()`, two lines earlier.)

c++/src/capnp/compat/http-over-capnp.c++
```diff
@@ -578,6 +654,67 @@ public:
...
+public:
+  HttpOverCapnpConnectResponseImpl(
+      kj::Own<kj::PromiseFulfiller<kj::HttpClient::ConnectRequest::Status>> fulfiller)
+      : fulfiller(kj::mv(fulfiller)) {}
+
+  ~HttpOverCapnpConnectResponseImpl() noexcept(false) {
+    if (fulfiller->isWaiting()) {
+      auto ex = KJ_EXCEPTION(FAILED,
+          "service's connect() implementation never called accept() nor reject()");
+      if (fulfiller->isWaiting()) {
```
kentonv, line 667: This `if()` looks redundant, since you just checked the same thing two lines earlier?

c++/src/capnp/compat/http-over-capnp.c++
```diff
@@ -688,6 +825,80 @@ public:
...
+
+    auto promise = inner->connect(host, headers, *pipe.ends[0], *response)
+        .eagerlyEvaluate([response=kj::mv(response),
+                          host=kj::mv(host),
+                          headers=kj::mv(headers),
+                          connectStream=kj::mv(pipe.ends[0])](kj::Exception&& ex) mutable {
+        response->handleException(kj::mv(ex), kj::mv(connectStream));
+      });
+
+    return paf.promise.then(
```
kentonv, line 852: You should really be doing `return promise` here. Exceptions thrown by the service promise need to propagate out of the RPC. Currently you are just attaching the promise to another promise, which means exceptions will go nowhere, and in fact trailing non-promise-returning `.then()`s won't actually execute. This could be causing some of your hang problems.

I would get rid of `paf`, and move the code below that handles pumping the response to happen directly inside `HttpOverCapnpConnectResponseImpl`. Follow the model used in `HttpServiceResponseImpl`; this should be very similar. (The success case should look similar to the WebSocket case, which involves pumps in both directions.)

kentonv: Oh, I missed the `eagerlyEvaluate` above, which I suppose addresses the error handling and lazy evaluation issues.

It still feels weird that this is designed differently from the HTTP case... but I can't think of a specific reason it wouldn't work I guess.

c++/src/capnp/compat/http-over-capnp.c++
dom96: Isn't it possible to get no error body but still have non 2xx status code?

c++/src/capnp/compat/http-over-capnp.c++
```diff
@@ -688,6 +831,88 @@ public:
...
+      if (status.statusCode < 200 || status.statusCode > 299) {
+        auto req = context.getParams().getContext().startErrorRequest();
+        auto rpcResponse = req.initResponse();
+        rpcResponse.setStatusCode(status.statusCode);
+        rpcResponse.setStatusText(status.statusText);
+        rpcResponse.adoptHeaders(factory.headersToCapnp(
+            headers, Orphanage::getForMessageContaining(rpcResponse)));
+
+        // Set the body size if the error body exists.
+        auto bodySize = 0;
```
kentonv, line 872: This will declare `bodySize` to have type `int`, because that's the compiler's best guess at what type you intend `0` to have. Technically, though, HTTP response bodies can be multiple gigabytes, so this could overflow. You need to explicitly specify `uint64_t` here.

More abstractly speaking, initializing an `auto` from any literal value is weird. What is the type of, say, `123`? C language rules will tell you it is `int`, but it's kind of arbitrary. From a pure type theory perspective, the type of the literal `123` is "the set of values containing only a single member, which is the value 123". So from a theoretical perspective it looks like you're declaring a variable that cannot have any other value than the one you assigned...

c++/src/capnp/compat/http-over-capnp.c++
```diff
@@ -688,6 +831,88 @@ public:
...
+        KJ_IF_MAYBE(errStream, status.errorBody) {
+          KJ_IF_MAYBE(size, errStream->get()->tryGetLength()) {
+            bodySize = *size;
+          }
+        }
+        rpcResponse.getBodySize().setFixed(bodySize);
```
kentonv, line 878: Looks like if `tryGetLength()` returns `nullptr` (meaning "unknown"), you are doing `setFixed(0)`, which means the length is known to be zero. You should be doing `getBodySize().setUnknown()`. Or, since "unknown" is the default, you could just move this line into the block above, so that the value is left as "unknown" otherwise. (And then you don't need the `bodySize` local.)

c++/src/capnp/compat/http-over-capnp.c++
```diff
@@ -578,6 +662,84 @@ public:
...
+        headers, Orphanage::getForMessageContaining(rpcResponse)));
+
+    kj::Own<kj::AsyncOutputStream> stream = factory.streamFactory.capnpToKj(request.getDown());
+    auto promises = kj::heapArrayBuilder<kj::Promise<void>>(3);
+    promises.add(req.send().ignoreResult());
+    // We write to the `down` pipe.
+    promises.add(httpProxyStream->pumpTo(*stream).then([req = kj::mv(request)](uint64_t) mutable {
+      auto endReq = req.getDown().endRequest();
+      return endReq.send().ignoreResult();
+    }).eagerlyEvaluate(nullptr));
```
kentonv, line 691: We should probably make sure that `down` gets dropped in the error case, which means you should attach `stream` before the `eagerlyEvaluate()` here, rather than after the `joinPromises()` later. That way it doesn't wait for the other direction to complete before dropping.

Separately, you should also call `httpProxyStream->abortRead()` on errors, otherwise the server app's attempts to write to the stream will just hang once the pump has stopped.

c++/src/capnp/compat/http-over-capnp.c++
```diff
@@ -578,6 +662,84 @@ public:
...
+public:
+  HttpOverCapnpConnectResponseImpl(
+      HttpOverCapnpFactory& factory, capnp::HttpService::ConnectParams::Reader request,
+      capnp::HttpService::ConnectClientRequestContext::Client context,
+      kj::Own<kj::AsyncInputStream> upStream, kj::Own<kj::AsyncIoStream> httpProxyStream) :
+      context(context), request(request), factory(factory), upStream(kj::mv(upStream)),
+      httpProxyStream(kj::mv(httpProxyStream)) {}
+
+  void accept(uint statusCode, kj::StringPtr statusText, const kj::HttpHeaders& headers) override {
+    KJ_REQUIRE(statusCode >= 200 && statusCode < 300, "the statusCode must be 2xx for accept");
```
kentonv, line 675: In theory it's not really necessary to validate the status code in http-over-capnp. It's necessary in the actual HTTP protocol implementation, where status codes matter. But in http-over-capnp, it's already possible to distinguish success and error cases based on how the protocol is used, independently of the status code. http-over-capnp is really an RPC version of the `kj::HttpService` interface, _not_ of the actual HTTP protocol, and as a result it can actually express things that wouldn't be possible in real HTTP, and that's OK. We don't actually validate status codes anywhere else in http-over-capnp, we just pass them through.

So, I would remove these checks.

kentonv:
Also this still needs unit tests...

c++/src/capnp/compat/http-over-capnp.c++
dom96: Maybe this is the source of my issues

dom96:
Ready for another review

c++/src/capnp/compat/http-over-capnp.c++
```diff
@@ -688,6 +830,69 @@ public:
     });
   }
 
+  kj::Promise<void> connect(ConnectContext context) override {
+    auto params = context.getParams();
+    auto host = params.getHost();
+    auto headers = factory.headersToKj(params.getHeaders());
+    auto pipe = kj::newTwoWayPipe();
+
+    class EOFDetector final: public kj::AsyncOutputStream {
```
harrishancock, line 839: Nit: KJ style would be EofDetector

c++/src/capnp/compat/http-over-capnp.c++
```diff
@@ -323,6 +323,56 @@ private:
   // Must check state->assertNotCanceled() before using this.
 };
 
+class HttpOverCapnpFactory::ConnectClientRequestContextImpl final
+    : public capnp::HttpService::ConnectClientRequestContext::Server {
+public:
+  ConnectClientRequestContextImpl(HttpOverCapnpFactory& factory,
+                           kj::HttpService::ConnectResponse& connResponse)
```
harrishancock, line 330: Nit: indentation

c++/src/capnp/compat/http-over-capnp-test.c++
```diff
@@ -648,5 +648,112 @@ KJ_TEST("HttpService isn't destroyed while call outstanding") {
...
+      KJ_ASSERT(count == 4, "Expecting the stream to read 4 chars.");
+      return io->tryRead(buf.begin(), 1, 1).then(
+          [&readPaf, buf = kj::mv(buf)](size_t count) mutable {
+        KJ_ASSERT(count == 0, "Expecting the stream to get disconnected.");
+        readPaf.fulfiller->fulfill();
+      }).attach(kj::mv(io));
+    });
+  }).wait(waitScope);
+
+  KJ_ASSERT(readPaf.promise.poll(waitScope));
```
harrishancock, line 752: I'm not sure this is proving anything that the .wait() above hasn't already proven.

c++/src/capnp/compat/http-over-capnp.c++
```diff
@@ -688,6 +830,69 @@ public:
...
+
+    PipelineBuilder<ConnectResults> pb;
+    auto eofWrapper = kj::heap<EOFDetector>(kj::mv(pipe.ends[1]));
+    auto up = factory.streamFactory.kjToCapnp(kj::mv(eofWrapper));
+    pb.setUp(kj::cp(up));
+
+    context.setPipeline(pb.build());
+    context.initResults(capnp::MessageSize { 4, 1 }).setUp(kj::mv(up));
+
+    class FinalHttpOverCapnpConnectResponseImpl final: public HttpOverCapnpConnectResponseImpl {
```
harrishancock, line 884: I'm curious why this class is necessary?

harrishancock (commented):
Left some non-blocking nits, but I'd note that we probably also want a unit test for the reject() code path.

c++/src/capnp/compat/http-over-capnp.c++
dom96: Good question and I am not sure myself. I followed how `request` above was implemented and it's using a similar class.

c++/src/capnp/compat/http-over-capnp-test.c++
```diff
@@ -648,5 +648,194 @@ KJ_TEST("HttpService isn't destroyed while call outstanding") {
...
+    void accept(uint statusCode, kj::StringPtr statusText, const kj::HttpHeaders& headers) override {
+      KJ_UNREACHABLE;
+    }
+
+    kj::Own<kj::AsyncOutputStream> reject(
+        uint statusCode,
+        kj::StringPtr statusText,
+        const kj::HttpHeaders& headers,
+        kj::Maybe<uint64_t> expectedBodySize) override {
+      KJ_ASSERT(statusCode == 500);
```
dom96, line 811: Two things I noticed while writing this test:

* A KJ_ASSERT failure here does not show up in test logs (all I get is "failed: Nothing to wait for; this thread would hang forever." which isn't ideal).
* Hopefully not a blocker for this PR but the `expectedBodySize` appears to be getting eaten by https://github.com/capnproto/capnproto/blob/master/c%2B%2B/src/kj/compat/http.c%2B%2B#L6510

dom96:
Added a reject test as well.

c++/src/capnp/compat/http-over-capnp.c++
```diff
@@ -578,6 +663,67 @@ public:
...
+        headers, Orphanage::getForMessageContaining(rpcResponse)));
+
+    auto errorBody = kj::mv(pipe.in);
+    // Set the body size if the error body exists.
+    KJ_IF_MAYBE(size, errorBody->tryGetLength()) {
+      rpcResponse.getBodySize().setFixed(*size);
+    }
+
+    KJ_IF_MAYBE(size, expectedBodySize) {
+      rpcResponse.getBodySize().setFixed(*size);
```
harrishancock, line 708: It looks like we're setting this twice. I think we could get rid of this latter `KJ_IF_MAYBE` block by passing `expectedBodySize` to the one-way pipe on line 692 (`kj::newOneWayPipe(expectedBodySize)`), then relying on the `errorBody->tryGetLength()` check above.

c++/src/capnp/compat/http-over-capnp-test.c++
harrishancock: Hmm, the first point, about KJ_ASSERT, seems worth understanding. :(

Regarding `expectedBodySize`, I wonder if the linked line needs to call `errorBody->tryGetLength()` and pass its result to `reject()`? Though for this particular `expectedBodySize` (line 810 above), we should probably pass it to the `kj::newOneWayPipe()` on line 813 below.

kentonv: KJ_ASSERT throws an exception. The exception could be caught somewhere, or (more likely) could be held up in a promise that nothing is waiting on, or is one branch of a `kj::joinPromises()` and the other branch is hanging.

> Hopefully not a blocker for this PR but the expectedBodySize appears to be getting eaten by

Looks like it should be `errorBody->tryGetLength()`. Can you go ahead and fix that before we forget?

c++/src/capnp/compat/http-over-capnp.c++
kentonv: I agree.

(Note that as currently written, `errorBody->tryGetLength()` will always return null, because no value was passed to the pipe constructor.)

c++/src/capnp/compat/http-over-capnp.c++
```diff
@@ -416,6 +466,41 @@ public:
...
+        Orphanage::getForMessageContaining(builder)));
+    rpcRequest.setContext(
+        kj::heap<ConnectClientRequestContextImpl>(factory, tunnel));
+    RemotePromise<capnp::HttpService::ConnectResults> pipeline = rpcRequest.send();
+
+    auto promises = kj::heapArrayBuilder<kj::Promise<void>>(2);
+    // We read from `downPipe` (the other side writes into it.)
+    auto downPumpTask = downPipe.in->pumpTo(connection).then([&connection](uint64_t) {
+      connection.shutdownWrite();
+    }).attach(kj::mv(downPipe.in)).eagerlyEvaluate(nullptr);
```
kentonv, line 490: Since you're no longer joining promises, and you've passed `nullptr` to `eagerlyEvaluate()`, any errors from `downPumpTask` won't be reported anywhere.

The reason `eagerlyEvaluate()` takes a parameter is actually to remind you to handle errors. The parameter is a callback function for the error case, just like `catch_()` would take. It's only appropriate to pass `nullptr` in cases where you are actually waiting on the promise later and so the errors will propagate.

Admittedly, it's not entirely clear what we should do with the error here. When there's truly nothing better to do, I generally recommend logging with `KJ_LOG(ERROR)`.

But I think we probably actually want fail-fast behavior. That is, if either the up-stream or the down-stream pump throws an exception, we want to cancel out the whole operation, propagating that exception.

Now, naively, that suggests we should just `exclusiveJoin()` all the promises together. With `exclusiveJoin()`, if any promise rejects, the others are canceled. But there's a problem, which is that we actually _don't_ want to proactively cancel the operation in the case that the pump in one direction completes successfully. One trick we can do is to make the pump promises return `kj::NEVER_DONE` at the end of the success path:

```
auto downPumpTask = downPipe.in->pumpTo(connection)
    .then([&connection, down = kj::mv(downPipe.in)](uint64_t) {
  connection.shutdownWrite();
  return kj::NEVER_DONE;
});
```

(Aside: Note that instead of using `.attach()` to attach `downPipe.in`, I captured it into the `.then()` continuation. This has basically the same effect, but avoids an extra promise node allocation, and also in ensures that the object is dropped before the NEVER_DONE wait.)

With this pattern, now we can use `exclusiveJoin()` to combine everything at the end of the method:

```
return pipeline.ignoreResult()
    .exclusiveJoin(kj::mv(downPumpTask))
    .exclusiveJoin(kj::mv(upPumpTask));
```

c++/src/capnp/compat/http-over-capnp.c++
```diff
@@ -688,6 +834,69 @@ public:
...
+    private:
+      kj::Own<kj::AsyncIoStream> inner;
+    };
+
+    kj::Own<kj::AsyncOutputStream> stream = factory.streamFactory.capnpToKj(
+        context.getParams().getDown());
+    auto promises = kj::heapArrayBuilder<kj::Promise<void>>(2);
+    // We write to the `down` pipe.
+    promises.add(pipe.ends[1]->pumpTo(*stream).then([context](uint64_t) mutable {
+      auto endReq = context.getParams().getDown().endRequest();
```
kentonv, line 875: Feels better to capture only what we need here (and capture a strong reference rather than a weak one -- capturing `context` is effectively capturing an unowned pointer).

```suggestion
    promises.add(pipe.ends[1]->pumpTo(*stream).then([down = params.getDown()](uint64_t) mutable {
      auto endReq = down.endRequest();
```

kentonv: Wait, I think this is redundant anyway. The `stream` will automatically call `end()` when you drop it. So I think this whole `.then()` should just be deleted -- the `.attach(kj::mv(stream))` will do the job.

c++/src/capnp/compat/http-over-capnp.c++
kentonv: Here you can just declare `HttpOverCapnpConnectResponseImpl` itself as `final`.

In the `request` case this wasn't possible, because `startRequest()` has its own subclass that works a bit differently.

kentonv: (And the reason things need to be declared `final` is because otherwise we get a compiler warning about calling a non-virtual destructor on a non-final polymorphic class...)

c++/src/capnp/compat/http-over-capnp.c++
```diff
@@ -688,6 +834,69 @@ public:
...
+      return endReq.send().ignoreResult();
+    }).attach(kj::mv(stream)).eagerlyEvaluate(
+        [&httpProxyStream = *pipe.ends[1]](kj::Exception&&) { httpProxyStream.abortRead(); }));
+
+    PipelineBuilder<ConnectResults> pb;
+    auto eofWrapper = kj::heap<EofDetector>(kj::mv(pipe.ends[1]));
```
kentonv, line 881: Since you're passing ownership of `pipe.ends[1]` to `EofDetector`, that means when `EofDetector` is destroyed, `pipe.ends[1]` will be destroyed. But the whole point of `EofDetector` is we expect it to be destroyed when the write direction is done, but the read direction could still be acti/e.

Unfortunately it seems like we need to add a refcounted wrapper around the pipe so that the pumps in both directions can hold references. The wrapper doesn't need to be a full `AsyncIoStream` implementation, it just needs to hold ownership. E.g.:

```
struct RefcountedWrapper: public kj::Refcounted {
  kj::Own<AsyncIoStream> stream;
};
auto& pipeRef = *pipe.ends[1];
auto refcounted = kj::refcounted<RefcountedWrapper>(kj::mv(pipe.ends[1]));
kj::Own<AsyncIoStream> ref1 = kj::attachRef(pipeRef, kj::addRef(*refcounted));
kj::Own<AsyncIoStream> ref2 = kj::attachRef(pipeRef, kj::addRef(*refcounted));
```

Now `ref1` and `ref2` are two refcounts on the same underlying stream.

c++/src/capnp/compat/http-over-capnp.c++
```diff
@@ -688,6 +834,69 @@ public:
...
+    class FinalHttpOverCapnpConnectResponseImpl final: public HttpOverCapnpConnectResponseImpl {
+    public:
+      using HttpOverCapnpConnectResponseImpl::HttpOverCapnpConnectResponseImpl;
+    };
+    auto response = kj::heap<FinalHttpOverCapnpConnectResponseImpl>(
+        factory, context.getParams().getContext());
+
+    promises.add(inner->connect(host, headers, *pipe.ends[0], *response).attach(
+        kj::mv(host), kj::mv(headers), kj::mv(response), kj::mv(pipe)));
+    return kj::joinPromises(promises.finish());
```
kentonv, line 897: Perhaps we want to use `exclusiveJoin()` here too -- again with the pump ending in `NEVER_DONE` when it succeeds.

c++/src/capnp/compat/http-over-capnp.c++
```diff
@@ -688,6 +829,80 @@ public:
...
+    };
+    auto& pipeRef = *pipe.ends[1];
+    auto refcounted = kj::refcounted<RefcountedWrapper>(kj::mv(pipe.ends[1]));;
+    kj::Own<kj::AsyncIoStream> ref1 = kj::attachRef(pipeRef, kj::addRef(*refcounted));
+    kj::Own<kj::AsyncIoStream> ref2 = kj::attachRef(pipeRef, kj::mv(refcounted));
+
+    // We write to the `down` pipe.
+    promises.add(ref1->pumpTo(*stream).ignoreResult().attach(kj::mv(stream))
+        .eagerlyEvaluate([&httpProxyStream = *ref1](kj::Exception&&) {
+          httpProxyStream.abortRead();
```
kentonv, line 883: You need to actually capture ownership of `ref1` here.

```suggestion
        .eagerlyEvaluate([httpProxyStream = kj::mv(ref1)](kj::Exception&&) mutable {
          httpProxyStream->abortRead();
```

dom96:
Rebased.

dom96:
Ready for review

PR #1588, opened 2022-11-25
Pin CI to ubuntu-20.04.

kentonv:
It looks like GitHub might have just updated ubuntu-latest to ubuntu-22.04, and everything is broken there (mostly because we try to install specific compiler version packages which are no longer available, but there is at least one real error with OpenSSL 3.0). Need to come back to this later.


kentonv:
@harrishancock Merging without review because it fixes CI which is otherwise super-broken.

PR #1592, opened 2022-12-08
Fix bug in cancellation blocking.

kentonv:
The detached promise used to make the call un-cancelable did not attach references to the objects that need to stay live until the call completes.

This bug was introduced recently with the `allowCancellation` changes.


c++/src/capnp/capability.c++
```diff
@@ -797,7 +797,8 @@ private:
       // Make sure this call cannot be canceled by forking the promise and detaching one branch.
       auto fork = result.promise.attach(context.addRef()).fork();
       result.promise = fork.addBranch();
-      fork.addBranch().detach([](kj::Exception&&) {
+      fork.addBranch().detach([ref = kj::addRef(*this), ctxRef = context.addRef()]
+                              (kj::Exception&&) {
```
jclee, line 801: I assume the ownership of the ref/ctxRef handles for the non-detached promise are covered by the default kj ownership rules.  Makes sense to me that we'd need to attach `ref`, but a little surprising that we need to attach `context.addRef()` to both the pre-forked promise and to the detached fork...  But I suppose this is to cover the case where the object needs to still be kept alive after the initial promise completes?

PR #1594, opened 2022-12-22
Introduce move-inhibiting version of KJ_DISALLOW_COPY

harrishancock:
This commit is not intended to be merged as-is, because it is currently a source-incompatible change. Rather, it is intended to prove that we have retrofitted all existing uses of KJ_DISALLOW_COPY to only appear in cases where move semantics are not required.

A later commit will rename KJ_DISALLOW_COPY to KJ_DISALLOW_COPY_AND_MOVE, and then rename KJ_DISALLOW_ONLY_COPY back to KJ_DISALLOW_COPY, making this a source-compatible change.

Edit: the later commit has been pushed, and this is ready to be merged.

### DIFF
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -182,10 +182,25 @@ typedef unsigned char byte;
 #endif
 #endif
 
-#define KJ_DISALLOW_COPY(classname) \
+#define KJ_DISALLOW_ONLY_COPY(classname) \
   classname(const classname&) = delete; \
   classname& operator=(const classname&) = delete
-// Deletes the implicit copy constructor and assignment operator.
+// Deletes the implicit copy constructor and assignment operator. This inhibits the compiler from
+// generating the implicit move constructor and assignment operator for this class, but allows the
+// code author to supply them, if they make sense to implement.
+//
+// This macro should not be your first choice. Instead, prefer using KJ_DISALLOW_COPY, and only use
+// this macro when you have determined that you must implement move semantics for your type.
+
+#define KJ_DISALLOW_COPY(classname) \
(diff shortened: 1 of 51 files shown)


c++/src/kj/common.h
```diff
@@ -178,14 +178,29 @@ typedef unsigned char byte;
 #elif __OPTIMIZE__
 #define KJ_NDEBUG
 #else
-#define KJ_DEBUG
+#define KJ_DEBUGd
```
kentonv, line 181: ??

harrishancock: d'oh

harrishancock:
(Didn't mean to move out of draft status, but now I don't know how to move it back.)

harrishancock:
Now ready for review/merge.

c++/src/kj/function.h
```diff
@@ -105,7 +105,7 @@ class Function<Return(Params...)> {
 
   // Make sure people don't accidentally end up wrapping a reference when they meant to return
   // a function.
-  KJ_DISALLOW_COPY(Function);
+  KJ_DISALLOW_ONLY_COPY(Function);
   Function(Function&) = delete;
   Function& operator=(Function&) = delete;
```
harrishancock, line 110: I wonder if these lines can be removed? And in ConstFunction.

kentonv: I think it's been like 10 years since I wrote these lines so I don't quite remember the motivation but I would guess there was one. The rules around non-const copy constructors are kind of complicated. I'd just leave them be.

kentonv (approved):

PR #1596, opened 2022-12-28
Support OpenSSL 3.0

kentonv:
This is based on #1434, but I went ahead and made KJ treat unexpected EOF as an error -- even with OpenSSL 1.x -- since this is technically important to prevent truncation attacks.

(It turns out that lots of apps than use OpenSSL 1.x are vulnerable to truncation attacks, which is why OpenSSL 3 decided to strengthen the error reporting to make it clearer that this situation _should_ be an error, but this was considered a breaking change. So KJ is not alone in having this bug, but I went ahead and fixed it even when using OpenSSL 1.x.)

NOTE: I'm flying blind here as my local Debian machine is still on OpenSSL 1.x. Gonna do a little CI-driven-development. I'll mark this no longer a draft once I get CI passing...

### DIFF
--- a/super-test.sh
+++ b/super-test.sh
@@ -403,13 +403,25 @@ if [ $IS_CLANG = yes ]; then
   # Don't fail out on this ridiculous "argument unused during compilation" warning.
   export CXXFLAGS="$CXXFLAGS -Wno-error=unused-command-line-argument"
 
-  # At the moment, only our clang-10 CI run seems to like -fcoroutines-ts. Earlier versions seem to
-  # have a library misconfiguration causing ./configure to result in the following error:
-  #   conftest.cpp:12:12: fatal error: 'initializer_list' file not found
-  #   #include <initializer_list>
-  # Let's use any clang version >= 10 so that if we move to a newer version, we'll get additional
-  # coverage by default.
-  if [ "${CXX#*-}" -ge 10 ] 2>/dev/null; then
+  # Enable coroutines if supported.
+  if [ "${CXX#*-}" -ge 14 ] 2>/dev/null; then
+    # Somewhere between version 10 and 14, Clang started supporting coroutines as a C++20 feature,
+    # and started issuing deprecating warnings for -fcoroutines-ts. (I'm not sure which version it
+    # was exactly since our CI jumped from 10 to 14, so I'm somewhat arbitrarily choosing 14 as the
+    # cutoff.)
+    export CXXFLAGS="$CXXFLAGS -std=c++20 -stdlib=libc++"
+    export LDFLAGS="-stdlib=libc++"
+
+    # TODO(someday): On Ubuntu 22.04, clang-14 with -stdlib=libc++ fails to link with libfuzzer,
+    #   which looks like it might itself be linked against libstdc++? Need to investigate.
+    unset LIB_FUZZING_ENGINE
+  elif [ "${CXX#*-}" -ge 10 ] 2>/dev/null; then
+    # At the moment, only our clang-10 CI run seems to like -fcoroutines-ts. Earlier versions seem
+    # to have a library misconfiguration causing ./configure to result in the following error:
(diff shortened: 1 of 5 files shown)


kentonv:
Well the g++-12 build proves OpenSSL 3 now works, but I have no idea why the clang-14 build isn't linking libc++ correctly.

I probably need to play with this on an Ubuntu VM / container rather than iterate in CI. Not going to do that tonight.

h-vetinari:
Thanks for tackling this! :)

Picked this PR on top of 0.10.3 in https://github.com/conda-forge/capnproto-feedstock/pull/32, but still failing on linux with:
```
[ TEST ] kj/compat/tls-test.c++:728: TLS client certificate verification
kj/compat/tls.c++:74: failed: OpenSSL error; message = error:16000069:STORE routines::unregistered scheme
stack: 7fe63300a740 7fe6330108cb 7fe633010b48 7fe632f5a699 7fe632f5a5a4 55f52f37b411 55f52f37b552 55f52f38806e 7fe632ef038e
[ FAIL ] kj/compat/tls-test.c++:728: TLS client certificate verification (152963 μs)
```
Similarly on osx:
```
[ TEST ] kj/compat/tls-test.c++:728: TLS client certificate verification
kj/compat/tls.c++:74: failed: OpenSSL error; message = error:16000069:STORE routines::unregistered scheme
stack: 1081baed9 1081c334e 1081c2ef5 1081c2530 10821f95f 10821f867 107715959 107714001 1077228ee 10831578a
[ FAIL ] kj/compat/tls-test.c++:728: TLS client certificate verification (184390 μs)
```

Both builds against openssl 1.1.1 pass the tests.

It's completely possible that some of the other intervening changes on master are also necessary (it wouldn't be a problem to wait for a new version), but I just wanted to provide some additional testing. :)

kentonv:
@h-vetinari I'm probably not going to be able to debug a problem that doesn't repro in our CI. That said, if you could do a debug build so that the stack trace is properly symbolized, that might help narrow down the issue. Also, "No such file or directory" is suspicious; if you could use `strace` to find out what file it's trying to open, that might help. It may be that OpenSSL requires some config file that's missing on your setup, although that's kind of strange since the test that is failing really shouldn't require any external files.

kentonv:
OK, got CI working. AFAICT this makes KJ work correctly with OpenSSL 3.

cc @thomaslee

super-test.sh
```diff
@@ -403,13 +403,25 @@ if [ $IS_CLANG = yes ]; then
...
-  # Let's use any clang version >= 10 so that if we move to a newer version, we'll get additional
-  # coverage by default.
-  if [ "${CXX#*-}" -ge 10 ] 2>/dev/null; then
+  # Enable coroutines if supported.
+  if [ "${CXX#*-}" -ge 14 ] 2>/dev/null; then
+    # Somewhere between version 10 and 14, Clang started supporting coroutines as a C++20 feature,
+    # and started issuing deprecating warnings for -fcoroutines-ts. (I'm not sure which version it
+    # was exactly since our CI jumped from 10 to 14, so I'm somewhat arbitrarily choosing 14 as the
+    # cutoff.)
+    export CXXFLAGS="$CXXFLAGS -std=c++20 -stdlib=libc++"
```
harrishancock, line 412: I wonder if we lose any code coverage by using `-std=c++20` instead of `gnu++20`? Maybe we don't actually use any GNU extensions that clang doesn't provide by default.

harrishancock (approved):

super-test.sh
kentonv: I would expect that if we were using any "GNU extensions" that aren't enabled for `-std=c++20`, we'd see errors, because I'm not aware of us using #ifdefs that somehow distinguish whether GNU extensions are enabled. So since we see no errors, I suspect it doesn't really make a difference?

PR #1598, opened 2023-01-02
Improve HttpServer::drain()

kentonv:


harrishancock (commented):
What happens if a server uses listenHttp(), then calls drain() after response headers are sent, but before the body has finished being sent?

kentonv:
> What happens if a server uses listenHttp(), then calls drain() after response headers are sent, but before the body has finished being sent?

Then `Connection: close` will not be added, but the server will still close the connection once the response is finished. This could theoretically lead to a problem if the client doesn't receive the EOF before they begin sending the next request. This is kind of unavoidable, though. No HTTP server lets you just keep a connection open and idle for long periods -- most of them will close the connection after a few seconds. I am not sure how browsers deal with this, exactly.

harrishancock (approved):

PR #1601, opened 2023-01-04
Fix WebSocket compression empty message bug

MellowYarker:
When sending an empty message, we would fail an assert that checked if the message tail included `0x00 0x00 0xFF 0xFF`.

### DIFF
--- a/c++/src/kj/compat/http-test.c++
+++ b/c++/src/kj/compat/http-test.c++
@@ -2307,6 +2307,10 @@ const byte WEBSOCKET_TWO_DEFLATE_BLOCKS_MESSAGE[] =
     {  0xc1, 0x0d, 0xf2, 0x48, 0x05, 0x00, 0x00, 0x00, 0xff, 0xff, 0xca, 0xc9, 0xc9, 0x07, 0x00 };
 // See this example: https://datatracker.ietf.org/doc/html/rfc7692#section-7.2.3.5
 // This uses two DEFLATE blocks in a single message.
+const byte WEBSOCKET_EMPTY_COMPRESSED_MESSAGE[] =
+    { 0xc1, 0x01, 0x00 };
+const byte WEBSOCKET_EMPTY_SEND_COMPRESSED_MESSAGE[] =
+    { 0xc1, 0x81, 12, 34, 56, 78, 0x00^12 };
 #endif // KJ_HAS_ZLIB
 
 template <size_t s>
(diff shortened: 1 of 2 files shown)


kentonv:
Just curious, do you have a reference to the part of the spec that says empty messages are an exception? Seems kind of weird to me that they wouldn't be required to be encoded just like any other message.

MellowYarker:
Oops, rebased on the wrong master :upside_down_face:

MellowYarker:
> Just curious, do you have a reference to the part of the spec that says empty messages are an exception? Seems kind of weird to me that they wouldn't be required to be encoded just like any other message.

The only things that seems relevant to me are:
- https://datatracker.ietf.org/doc/html/rfc7692#section-7.2.1
```
2.    If the resulting data does not end with an empty DEFLATE block
       with no compression (the "BTYPE" bits are set to 00), append an
       empty DEFLATE block with no compression to the tail end.
```

- https://datatracker.ietf.org/doc/html/rfc7692#section-7.2.3.6
```
   Suppose that an endpoint is sending data of unknown size.  The
   endpoint may encounter the end-of-data signal from the data source
   when its buffer for uncompressed data is empty.  In such a case, the
   endpoint just needs to send the last fragment with the FIN bit set to
   1 and the payload set to the DEFLATE block(s), which contains 0 bytes
   of data.  If the compression library being used doesn't generate any
   data when its buffer is empty, an empty uncompressed DEFLATE block
   can be built and used for this purpose as follows:

       0x00

   The single octet 0x00 contains the header bits with "BFINAL" set to 0
   and "BTYPE" set to 00, and 5 padding bits of 0.
```

When our uncompressed message length is 0, `ctx.avail_in == 0`, and so Zlib will give us `Z_BUF_ERROR` when we try to `deflate()`
>  deflate() returns Z_OK if some progress has been made (more input
  processed or more output produced), Z_STREAM_END if all input has been
  consumed and all output has been produced (only when flush is set to
  Z_FINISH), Z_STREAM_ERROR if the stream state was inconsistent (for example
  if next_in or next_out was Z_NULL or the state was inadvertently written over
  by the application), or **Z_BUF_ERROR** if no progress is possible (for example
  avail_in or avail_out was zero).  Note that Z_BUF_ERROR is not fatal, and
  deflate() can be called again with more input and more output space to
  continue compressing.

c++/src/kj/compat/http.c++
```diff
@@ -3135,10 +3135,13 @@ private:
...
-        message = innerMessage.slice(0, innerMessage.size() - 4);
-        // Strip 0x00 0x00 0xFF 0xFF off the tail.
-        // See: https://datatracker.ietf.org/doc/html/rfc7692#section-7.2.1
+        message = innerMessage;
+        if (innerMessage.size() > 0) {
+          KJ_ASSERT(innerMessage.asPtr().endsWith({0x00, 0x00, 0xFF, 0xFF}));
+          message = innerMessage.slice(0, innerMessage.size() - 4);
+          // Strip 0x00 0x00 0xFF 0xFF off the tail.
+          // See: https://datatracker.ietf.org/doc/html/rfc7692#section-7.2.1
+        }
```
MellowYarker, line 3144: With the context of the previous comment, we could change this to:

```
if (message.size() > 0) {
  KJ_ASSERT(innerMessage.asPtr().endsWith({0x00, 0x00, 0xFF, 0xFF}));
  message = innerMessage.slice(0, innerMessage.size() - 4);
  // Strip 0x00 0x00 0xFF 0xFF off the tail.
  // See: https://datatracker.ietf.org/doc/html/rfc7692#section-7.2.1
} else {
  // We didn't compress anything so we add an empty deflate block.
  message = compressedMessage.emplace(Array<kj::byte>(0x00));
}
```

The (new) test still passes, and it seems to be closer to what the spec asks for?

c++/src/kj/compat/http-test.c++
```diff
@@ -2307,6 +2307,10 @@ const byte WEBSOCKET_TWO_DEFLATE_BLOCKS_MESSAGE[] =
     {  0xc1, 0x0d, 0xf2, 0x48, 0x05, 0x00, 0x00, 0x00, 0xff, 0xff, 0xca, 0xc9, 0xc9, 0x07, 0x00 };
 // See this example: https://datatracker.ietf.org/doc/html/rfc7692#section-7.2.3.5
 // This uses two DEFLATE blocks in a single message.
+const byte WEBSOCKET_EMPTY_COMPRESSED_MESSAGE[] =
+    { 0xc1, 0x00 };
+const byte WEBSOCKET_EMPTY_SEND_COMPRESSED_MESSAGE[] =
+    { 0xc1, 0x80, 12, 34, 56, 78 };
```
MellowYarker, line 2313: With the other comment in mind (regarding RFC), I think by adding an empty deflate block `0x00`, these would become:

```
const byte WEBSOCKET_EMPTY_COMPRESSED_MESSAGE[] =
    { 0xc1, 0x01, 0x00 };
const byte WEBSOCKET_EMPTY_SEND_COMPRESSED_MESSAGE[] =
    { 0xc1, 0x81, 12, 34, 56, 78, 0x00^12 };
```

kentonv:
Hmm, the spec lines you quoted don't seem to clearly say that empty messages are acceptable. It sounds more like an empty uncompressed message should "compress" to an appropriate representation of an empty input, which itself is non-empty. I guess the text seems to be saying the appropriate representation is just a single `0x00` byte, though that seems weird to me, why wouldn't it have the usual suffix?

Have we seen browsers sending us empty "compressed" messages in practice?

MellowYarker:
> Have we seen browsers sending us empty "compressed" messages in practice?

The feature isn't enabled yet, still behind compat flag. I ran into this issue when sending an empty message from a python client to a Worker that simply echo's the message.

I've noticed the following:
- When sending an empty "compressed" message from a python client using `websockets`, or from Chrome, the Worker can process the message correctly and verify that it is in fact an empty string.
- When sending an empty "compressed" message from Firefox, we **fail** to decompress the message in the Worker (we get a `Z_DATA_ERROR` from `inflate()`).

Unfortunately, the Worker can't determine if Chrome/python's `websockets` is sending `0x00` instead of an empty message, since only KJ can see the raw bytes. I can take a look at the chromium source code later today to try and figure it out. I'll also check to see if the Firefox behavior has been mentioned before.

kentonv:
OK, I'm convinced that sending a one-byte message containing just a zero is the right answer, though if you could test it against Chrome, that would provide the definitive proof. Anyway, once you change the code to send the one-byte message I'm happy to merge the PR.

kentonv (approved):

MellowYarker:
I've confirmed that this is fixed in chrome and python's `websockets` library.

PR #1604, opened 2023-01-06
[bazel] supporting generated capnp files

mikea:
Without this fix src path for generated files is somewhere in bazel internals. We need to tell capnp to strip this part of the name.


kentonv:
Well... I _think_ I understand what this is doing... :)

PR #1605, opened 2023-01-10
Introduce RefcountedWrapper to refcount types that don't inherit Refcounted.

kentonv:
Various ad hoc versions of this have appeared all over the place, but many of them -- including the one replaced in this PR -- do more allocation than necessary. By building this into refcounted.h, we can take advantage of friendly knowledge that `Refcounted` itself implements `Disposer`, and that said disposer doesn't care what pointer is passed into it, because it knows it's supposed to dispose itself. So we can set the refcounted wrapper as the disposer on an `Own<T>` pointing at the wrapped object, avoiding the need for an additional allocation that would otherwise be needed if we used `kj::attachRef`.

@dom96 can you review? (GitHub won't let me formally set you as reviewer unless I make you a project member, but it'll still let you approve the PR.)

### DIFF
--- a/c++/src/capnp/compat/http-over-capnp.c++
+++ b/c++/src/capnp/compat/http-over-capnp.c++
@@ -873,14 +873,9 @@ public:
 
     // We want to keep the stream alive even after EofDetector is destroyed, so we need to create
     // a refcounted AsyncIoStream.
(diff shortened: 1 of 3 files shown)


jasnell (approved):

dom96 (approved):
Nice! Looks good.

Issue #474, opened 2017-05-02
Support for non-pointer generic type parameters


ajg:
(Apologies if this issue has already been addressed elsewhere.)

Is there any hope for value types to be permitted at some point as parameters to generics, or is the "pragmatic limitation" [mentioned in the documentation](https://capnproto.org/language.html#generic-types) set in stone?

Alternatively, has built-in support for standard generic containers (e.g. `Map`, `Set`, `Tuple`) allowing value types (akin to `List`) been considered?

KubaO:
I see two general approaches: both have valid use cases and could be eventually supported.

1. Pointer-based generic types - what we have now, adding pointers-to-primitives. All specializations have the same encoding, and every generic field takes up a pointer slot.

2. Substitution-based generic types - where the type parameters are substituted into the type and create a new type, laid out appropriately. I had experimented with a C++ PoC that uses a constexpr-based layout engine to lay out the specialized type at compile time: it can be made to work, so that's viable. It will also work in .Net, and other languages that support code generation, like Python and Java.

  It becomes problematic in legacy languages that don't support compile-time code generation: IEC 61131, C and Pascal at least are a problem. There, the substitutions would have to be generated by the codegen, i.e. an application-specific `.capnp` file would be needed to reify the specialized types.

For pointer-based generic types, the following levels of implementation can be imagined.

1. A generic primitive type is encoded as-if it was a single-element struct of that type. This is extremely simple to implement.

2. Primitive pointers are added that can hold, say, up to a 60-bit value within the pointer
itself (without type information), or a 55-bit value if type information is preserved. Values that
don't fit within the primitive pointer are encoded using a struct pointer.

Level 1 is the simplest to implement as it requires no encoding changes. Level 2 introduces a backwards-incompatible change, i.e. old clients won't be able to decode these pointers, but the runtime should have an option to stick with level 1 when an older client is detected (in an application-specific way).

kentonv:
@KubaO I considered both the options you suggest when designing generics, but both have pretty severe problems.

Option 1 does not live up to the expectation that an instance of a generic type should be binary-compatible with what you'd get if you manually substituted the type parameter. (And anyway, option 1 "level 1" can easily be implemented by an application today, by declaring "boxed" types wrapping every primitive type.)

[...]

PR #1608, opened 2023-01-15
Fix Generic Parameter Passthrough to List(T)

ErikMcClure:
Currently, capnproto cannot pass through a generic parameter to a `List` directly. This simple example does not compile, even without ever instantiating `Foo`, producing the error `'List(AnyPointer)' is not supported.`:

```capnp
@0xdbb9ad1f14bf0b36;

struct Foo(T) {
  entries @0 :List(T);
}
```

This error is **not** a restriction in the protocol itself (unlike #474) - it is purely a limitation of the capnproto compiler and code generation. This pull request builds off of @bb010g's [initial fix](https://github.com/bb010g/capnproto/tree/list-generic) to the core capnproto compiler by fixing the C++ codegen and adding a few tests to ensure it is actually working. Other language code generators will have to be updated to support this (I hope to work on a fix for [capnproto-rust](https://github.com/capnproto/capnproto-rust) soon).

I am not familiar with capnproto's codebase, and I am not actually sure I have fixed this correctly, but hopefully this PR makes it clear *what* needs to be fixed. I tested this with MSVC 19.34 on Windows and gcc 10.3 on WSL NixOS

In particular, one major change I had to make was removing the explicit specification of the List Kind in the generated C++ code, instead relying on the inferred constexpr kind instead. I'm not sure if this is related to older C++ compilers struggling with the inference, or if there is some other case where it is necessary, but removing the explicit kind parameter did not break any of the tests.

I also had to remove an MSVC workaround that is no longer required with recent versions of MSVC and was actually causing compilation problems. I noticed that several other MSVC workarounds are no longer necessary either, but they did not cause any compilation problems, so I did not remove them. To preserve compatibility with older MSVC versions, I only turn off the workaround if the version is greater than 19, but it might cause compile errors with earlier versions that I cannot test. Given that some of these workarounds significantly increase code complexity, it may be worth updating the minimum required MSVC version and removing ones that are no longer necessary.

I'm not particularly confident in test case coverage, as I only tested very basic usage. There may still be complex situations or nested cases where it fails, but I'm not familiar enough with capnproto to know what to test.

The nix flake file has been added purely for compatibility with NixOS - if you don't want that file in the repo, I can remove it from the PR. Most of the other changes in these files were done by @bb010g, so I might not be able to answer questions about why those changes were made.


kentonv:
Hi,

Just wanted to let you know I've seen this but my inbox is pretty backlogged at the moment. I'll try to get to this when I can.

I didn't actually remember that lists of generic types weren't implemented, but if I had to guess why, it might have to do with the fact that `List(AnyPointer)` is not permitted, and of course all generic parameters default to `AnyPointer`. The reason `List(AnyPointer)` is not permitted is mostly due to potential confusion: Normally, an `AnyPointer` can point at a struct type, but list-of-struct is NOT encoded as list-of-pointer; instead, the structs are flattened. So the "unbound" version of a generic type `Foo(T)` that contains a `List(T)` ends up being incompatible with the version that binds `T` to some struct type. This is a weird wart but not necessarily a showstopper as long as people are aware of it.

Is this issue addressed in any particular way in your PR?

ErikMcClure:
Sorry for how long this response took, but we've been trying to figure out how to deal with this problem and so far haven't come up with any easy answers to this.

As you mentioned, the fundamental issue here is that we've inadvertently exposed a second way of representing lists of structs. Because `List(T)` is represented as an `AnyPointer`, it represents a list of structs as a list of *pointers to structs*, instead of a flattened sequence of structs. This introduces an invisible type difference, because now you can have `List(T)` where `T` is `Foo`, which is incompatible with `List(Foo)`. We've decided this situation is unacceptable and needs to be dealt with, somehow, but it's not clear how.

1. One suggestion was to allow the protocol to transparently convert between lists of struct pointers and lists of flattened structs. However, this breaks forward compatibility guarantees, because a list of any memory layout other than composite can be transparently upgraded to a list of a struct with a first element given by the original list. This conflicts with the rule required to make pointer lists mutually convertible representations with composite lists. If there is a way to avoid this pitfall, we haven't been able to find one.

2. Another option is to annotate the type, requiring a naked T parameter in a list to have some kind of suffix or prefix like `@pointer` to "enable" the behavior in the PR, thus making it clear that the types are different. If the annotation isn't included, the original "AnyPointer is not allowed" error would happen. This *works* and preserves existing capnproto behavior, but requires a change to the capnproto syntax, and I have no idea what you consider an appropriate syntax extension. This would look something like:

```
struct Foo(T) {
  entries @0 :List(T@pointer);
}
```

3. The final option is one that completely solves the problem, but i don't think you're gonna like it. While I thought we could address this issue without running into the Lists of Values problem, it turns out that this is actually the same problem. The whole reason this doesn't work is because, to properly support `List(T)` without accidentally allowing two different ways of representing a list of structs, then generics need to support different layout types, as suggested in #474. This means we have to revisit whether or not forcing generics to be represented as AnyPointer is actually sustainable. However, it has been almost 6 years since that issue was filed, and given the problems we have run into, maybe it's worth revisiting.

   At the time, your primary concern was that, if the layout of a list can change based on the type put into the list, this requires potentially complex layout logic to be done at compile time by the generated code in each language. This is still true, but languages have been adding much more powerful compile time expressions that I believe make this more viable now than it was back in 2017. C++17 support is now widespread, which added `if constexpr`, and C++20 greatly expanded what you can put in `constexpr` functions, which would make it much easier to implement this. Go [now has generics](https://go.dev/doc/tutorial/generics). Rust [has proc macros](https://doc.rust-lang.org/reference/procedural-macros.html). Zig [has comptime functions](https://ziglang.org/documentation/master/#comptime).
   
I honestly think the correct solution here is to bite the bullet and solve the generics value problem for good. However, if you *really* don't want to do that, let me know which of the other two proposals you prefer here, or if you have a suggestion for how to resolve this in a way you feel is appropriate.

kentonv:
Well this is fun.

I think solving this problem for `List(T)` should be easier than solving non-pointer generic parameters in general (which I believe is intractable).

The basic idea: Normally, when a generic parameter `T` is not bound, then we treat `T` as `AnyPointer`. This causes a problem because `List(AnyPointer)` is not allowed (because it would be inconsistent for structs). I think the trick here is that when `T` is not bound, `List(T)` should be treated as `AnyList`, NOT `List(AnyPointer)`. Under C++ templates, at least, it should be straightforward to implement such a special-case -- define a template alias `ListMaybeAny<T>` that maps to `List<T>` or, if `T` is `AnyPointer`, to `AnyList`.

Technically, it could even be easy to allow `T` to be a non-pointer type in the case that it is strictly used as `List(T)` -- but I wouldn't actually make that change yet.

PR #1610, opened 2023-01-19
Fix TOCTOU bug when recursively deleting a directory with symlinks.

kentonv:
We don't recurse into symlinks, we just delete the symlink. But it's possible that at the time we checked the file type, it was an actual directory, and then someone swapped it out for a symlink before we tried to recurse into it.

Rust recently fixed the same bug, which they treated as a CVE: https://github.com/rust-lang/rust/security/advisories/GHSA-r9cc-f5pr-p3j2

This is only a security issue if the program doing the deleting is a privileged process, and it is deleting directories which are writable to an attacker.

I am not calling this a CVE in KJ because I feel quite comfortable saying that nobody is using KJ filesystem in this way today. It's likely that literally no one is using the recursive-delete feature at all, and no one except my own projects are even using this entire API.

Note: The same bug presumably exists on Windows, but it appears that fixing it would require switching to entirely different APIs, which I'm not going to do today. Instead, I've left a warning.

### DIFF
--- a/c++/src/kj/filesystem-disk-unix.c++
+++ b/c++/src/kj/filesystem-disk-unix.c++
@@ -188,7 +188,7 @@ static void rmrfChildrenAndClose(int fd) {
       if (entry->d_type == DT_DIR) {
         int subdirFd;
         KJ_SYSCALL(subdirFd = openat(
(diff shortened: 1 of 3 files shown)


harrishancock (approved):
Whoops, I didn't see this until now!

PR #1612, opened 2023-01-21
Add a ResetableRef utility to protect against UAF

jasnell:
The HttpChunkedEntityWriter and HttpFixedLengthEntityWriter both previously held a bare reference to the HttpOutputStream, and both perform additional cleanup in their destructors. If, for whatever reason, the Own<HttpChunkedEntityWriter> or Own<HttpFixedLengthEntityWriter> happened to outlive whatever thing happened to own the HttpOutputStream, a UAF crash can happen. This commit adds some safety rails around it by introducing a ResetableRef utility that will clear the ref when the owning object is destroyed. A once-per-process log entry is emitted to detect problematic usage of both HttpChunkedEntityWriter and HttpFixedLengthEntityWriter.

Eventually, we need to remove the cleanup logic from these types and implement a proper end() function for AsyncOutputStream.

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -1943,6 +1943,54 @@ kj::Own<HttpInputStream> newHttpInputStream(
 
 namespace {
 
+template <typename T>
+class ResetableRef {
+  // Utility that holds a refcounted ref to an instance of T.
+  // When multiple ResetableRefs point to the same T, and
+  // reset() is called, the reference is cleared for all instances.
+  // Only the owner (whoever holds the original ResetableRef that
+  // set the reference) is allowed to reset it.
+public:
+  ResetableRef(T& t) : inner(refcounted<Impl>(t)), owner(true) {}
+  ResetableRef(ResetableRef& other)
+      : inner(kj::addRef(*other.inner)) {}
+  ResetableRef(ResetableRef&& other)
+      : inner(mv(other.inner)), owner(owner) {
+    other.movedAway = true;
+    other.owner = false;
+  }
+  ResetableRef& operator=(ResetableRef& other) {
+    inner = kj::addRef(other.inner);
+  }
+  ResetableRef& operator=(ResetableRef&& other) {
+    inner = mv(other.inner);
+    owner = other.owner;
+    other.movedAway = true;
+    other.owner = false;
+  }
+
+  Maybe<T&> tryGet() {
+    if (!movedAway) {
+      KJ_IF_MAYBE(i, inner->ref) { return *i; }
+    }
+    return nullptr;
+  }
+
+  void reset() {
+    KJ_REQUIRE(owner, "only the owner of a ResetableRef may reset it");
+    if (!movedAway) { inner->ref = nullptr; }
+  }
+
+private:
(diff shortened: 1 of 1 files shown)


c++/src/kj/compat/http.c++
```diff
@@ -2256,27 +2362,37 @@ public:
...
       // Hey, we know exactly how large the input is, so we can write just one chunk.
 
       uint64_t length = kj::min(amount, *l);
-      inner.writeBodyData(kj::str(kj::hex(length), "\r\n"));
-      return inner.pumpBodyFrom(input, length)
+      i.writeBodyData(kj::str(kj::hex(length), "\r\n"));
+      return i.pumpBodyFrom(input, length)
           .then([this,length](uint64_t actual) {
+        auto& i = KJ_UNWRAP_OR(inner.tryGet(), {
+          logInnerClearedWarningOnce();
```
harrishancock, line 2383: I wonder if we should still throw an exception here if `actual < length`?

c++/src/kj/compat/http.c++
```diff
@@ -2120,37 +2168,56 @@ public:
...
     uint64_t size = 0;
+    auto& i = KJ_UNWRAP_OR(inner.tryGet(), {
+      logInnerClearedWarningOnce();
+      return kj::READY_NOW;
+    });
+
     for (auto& piece: pieces) size += piece.size();
 
     if (size == 0) return kj::READY_NOW;
     KJ_REQUIRE(size <= length, "overwrote Content-Length");
```
harrishancock, line 2207: Nit: I think it'd be preferable to still throw if this precondition doesn't hold. Here and above in the other overload of `write()`.

harrishancock (approved):

c++/src/kj/compat/http.c++
```diff
@@ -2120,37 +2168,56 @@ public:
...
-  HttpFixedLengthEntityWriter(HttpOutputStream& inner, uint64_t length)
+  HttpFixedLengthEntityWriter(ResetableRef<HttpOutputStream>& inner, uint64_t length)
       : inner(inner), length(length) {
-    if (length == 0) inner.finishBody();
+    if (length == 0) KJ_ASSERT_NONNULL(inner.tryGet()).finishBody();
   }
   ~HttpFixedLengthEntityWriter() noexcept(false) {
-    if (length > 0 || inner.isWriteInProgress()) {
-      inner.abortBody();
+    auto& i = KJ_UNWRAP_OR(inner.tryGet(), {
```
harrishancock, line 2176: I think all of these have to be `[&] { ... }()` to keep MSVC happy. :(

harrishancock: @jasnell apparently I have led you astray: https://github.com/capnproto/capnproto/blob/master/c%2B%2B/src/kj/common.h#L1346-L1349
> // Unfortunately, these macros seem impossible to express without using GCC's non-standard
> // "statement expressions" extension. IIFEs don't do the trick here because a lambda cannot
> // return out of the parent scope. These macros should therefore only be used in projects that
> // target GCC or GCC-compatible compilers.
:(

PR #1613, opened 2023-01-22
Fix build with gcc 13 by including <cstdint>

heirecka:
Like other versions before, gcc 13 moved some includes around and as a result <cstdint> is no longer transitively included. Explicitly include it for uint8_t.

### DIFF
--- a/c++/src/kj/string-test.c++
+++ b/c++/src/kj/string-test.c++
@@ -24,7 +24,7 @@
 #include <string>
 #include "vector.h"
 #include <locale.h>
-#include <cstdint>
+#include <stdint.h>
 
 namespace kj {
 namespace _ {  // private
--- a/c++/src/kj/string-test.c++
+++ b/c++/src/kj/string-test.c++
@@ -24,6 +24,7 @@
 #include <string>
 #include "vector.h"
 #include <locale.h>
+#include <cstdint>
 
 namespace kj {
 namespace _ {  // private


c++/src/kj/string-test.c++
```diff
@@ -24,6 +24,7 @@
 #include <string>
 #include "vector.h"
 #include <locale.h>
+#include <cstdint>
```
kentonv, line 27: In KJ code we prefer not to use the C++ wrappers for C headers, we just use the C headers directly.

```suggestion
#include <stdint.h>
```

(Technically the C++ wrapper header is supposed to define all symbols within the `std` namespace. If it actually did so, then this fix presumably wouldn't work, because we don't prefix `uint8_t` with `std::`. Of course, in practice this requirement that symbols be defined in `std::` doesn't seem to be followed -- libstdc++ goes ahead and puts everything in the global scope anyway. But in that case, what's the point of using these wrapper headers? We might as well use the original C headers. If we use the wrappers then we risk that someday we want to port the code to a platform that actually does the namespacing correctly... and it doesn't work!)

kentonv (commented):
Thanks!

kentonv:
(Will squash-merge after CI passes.)

kentonv:
Mac failure is unrelated.

kloczek:
Already since las release have been accimul;ates +200 commots https://github.com/capnproto/capnproto/compare/v0.10.3...master
I thisnk that it would be good to release new version.
Any plans to do that soon? 🤔

kentonv:
@kloczek I have a patch release candidate which includes this fix:

https://capnproto.org/capnproto-c++-0.10.4-rc1.tar.gz

There's still an issue with Clang 16 though, I'll try to solve that by the end of the week and then release this.

I won't have time to do a 0.11 release anytime soon. The problem is, releases are expected to pass a much wider range of tests than what we run in CI on every PR. For example, they're expected to work with `-fno-exceptions`. I need to go back and fix the various problems that have cropped up with that before doing a release, but I don't have time at the moment. Additionally, there is currently an obscure incompatibility between git master and the last release when using `-> stream` -- the last release had a bug where streaming calls would be prematurely canceled, but also had a separate bug where the cancellations were ignored on the server side. Unfortunately since both bugs are now fixed, it means that an old client talking to a new server will see issues. I haven't quite figured out what to do about it yet, but I think we need to resolve that before the next release.

In general (notwithstanding this unusual issue with streaming) if you use Cap'n Proto in the default configuration on one of the core platforms tested by CI, then it's totally reasonable to use Cap'n Proto from git master and forget about releases. This is what Cloudflare Workers does, for instance. Think of releases as more like LTS.

PR #1614, opened 2023-01-25
Allow streaming for queued remote RPC calls

jclee:


c++/src/capnp/rpc-twoparty-test.c++
```diff
@@ -738,6 +738,118 @@ KJ_TEST("Streaming over RPC") {
...
+      calleePromise = nullptr;
+
+      return kj::READY_NOW;
+    }
+  };
+
+  kj::EventLoop loop;
+  kj::WaitScope waitScope(loop);
+
+  // Set up a local "caller" server that eventually delegates requests for callees to a remote
```
kentonv, line 784: FWIW, I think this test could be simplified by using a PromiseFulfillerPair here, like:

```
auto paf = kj::newPromiseAndFulfiller<test::TestStreaming::Client>();
test::TestStreaming::Client promisedClient = kj::mv(paf.promise);
```

Then:

1. make a streaming call to promisedClient, verify it doesn't resolve because we're blocked on a local promise
2. resolve the paf to a remote capability, but where the remote cap's method implementation hangs or otherwise doesn't return immediately
3. now verify that the original streaming call's returned promise has resolved, and several more streaming calls can be made, until the window fills up

That said, if you'd rather just leave the test as-is and move on I'm also fine with that.

kentonv (approved):

c++/src/capnp/rpc-twoparty-test.c++
jclee: Thanks, that seems much simpler.

PR #1615, opened 2023-01-26
Change sprintf() -> snprintf() to silence warnings.

kentonv:
There were only a couple usages and they are all cases that are formatting a single integer therefore couldn't possibly overrun the buffer. Plus most are in test code. But it's easy enough to call snprintf().

This fixes the mac build, which recently started failing due to a deprecation warning for sprintf().


jclee (approved):
Looks OK to me, FWIW

PR #1616, opened 2023-01-26
Alternative version of HTTP body stream UAF protection.

kentonv:
I had an idea for an alternative way to implement the UAF protection introduced in #1612.

This approach has a few benefits:
* No extra allocation is needed.
* It applies to both input and output streams (readers and writers).
* It enforces / takes advantage of the fact that there's only one wrapper stream at a time -- no refcounting needed.
* I think the code is a little more self-contained (modulo some refactoring of the affected classes).

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -1642,7 +1642,7 @@ public:
   }
 
 protected:
-  HttpInputStreamImpl& inner;
+  HttpInputStreamImpl& getInner() { return inner; }
 
   void doneReading() {
     KJ_REQUIRE(!finished);
@@ -1653,6 +1653,7 @@ protected:
   inline bool alreadyDone() { return finished; }
 
 private:
+  HttpInputStreamImpl& inner;
   bool finished = false;
 };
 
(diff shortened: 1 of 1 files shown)


c++/src/kj/compat/http.c++
```diff
@@ -1113,11 +1113,63 @@ kj::String HttpHeaders::toString() const {
...
+  //
+  // It's easy for applications to screw up and hold on to a body stream beyond the lifetime of the
+  // underlying connection stream. This used to lead to UAF. This mixin class implements behavior
+  // that detached the wrapper if it outlives the wrapped stream, so that we log errors and
+
+public:
+  ~WrappableStreamMixin() noexcept(false) {
+    KJ_IF_MAYBE(w, currentWrapper) {
+      KJ_LOG(ERROR, "HTTP connection destroyed while HTTP body streams still exist",
+          kj::getStackTrace());
```
jasnell, line 1131: Do we still need the log? I added it in my version so we could investigate further but not sure it's something we want to keep around indefinitely?

jasnell (approved):

jasnell:
Definitely much better than my Quick Fix(tm).

c++/src/kj/compat/http.c++
kentonv: I think so, it's definitely an error on the part of the app which we'd want to know about and fix. Actually it should probably be an exception in debug mode...

PR #1617, opened 2023-01-30
Updated the C# other-lang header to be just 'C#' instead of 'C# (.NET Core)'

mesedik:
Currently, the other languages page on the capnproto website implies that the capnproto implementation for the C# languages only supports .NET Core but as of now that is not the case. [The C# implementation](https://github.com/c80k/capnproto-dotnetcore) supports the .NET Standard 2.0 and has for a while now.


kentonv:
@c80k what do you think?

c80k:
agreed

kentonv:
Sorry, this got lost in my inbox for a while.

PR #1618, opened 2023-01-31
Remove operator!= synthesized by spaceship

ecatmur:
An operator!= suppresses the reversed equality comparison candidate generation.

This is visible in clang 16 (rc0 just released) and in gcc trunk (so probably gcc 13).


risa2000:
Just for the info for anyone coming to this later: This problem popped up suddenly with the release of Microsoft Visual Studio 2022 17.6 - it seems that MSVC started finally to enforce the correct operator generation. `v0.10.4` fixed that.

PR #1620, opened 2023-02-03
Add guards against canceled HTTP reads.

kentonv:
A canceled read from an HTTP body can leave the body stream in an invalid state if some data was read but never processed. This is usually fine because reads are only canceled when the whole operation is being canceled, but let's guard against attempts to read further to be safe.

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -1714,15 +1714,21 @@ public:
   }
 
   Promise<size_t> tryRead(void* buffer, size_t minBytes, size_t maxBytes) override {
+    KJ_REQUIRE(clean, "can't read more data after a previous read didn't complete");
+    clean = false;
     return tryReadInternal(buffer, minBytes, maxBytes, 0);
   }
 
 private:
   size_t length;
+  bool clean = true;
 
   Promise<size_t> tryReadInternal(void* buffer, size_t minBytes, size_t maxBytes,
                                   size_t alreadyRead) {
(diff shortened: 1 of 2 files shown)


c++/src/kj/compat/http.c++
```diff
@@ -7110,8 +7110,14 @@ private:
...
 
                 auto dummy = kj::heap<HttpDiscardingEntityWriter>();
-                auto lengthGrace = body->pumpTo(*dummy, server.settings.canceledUploadGraceBytes)
-                    .then([this](size_t amount) {
+                auto lengthGrace = kj::evalNow([&]() {
+                  return body->pumpTo(*dummy, server.settings.canceledUploadGraceBytes);
+                }).catch_([](kj::Exception&& e) -> size_t {
+                  // Reading from the input failed in some way. This may actually be the whole
+                  // reason we got here in the first place so don't propagate this error, just
+                  // give up on discarding the input.
```
harrishancock, line 7118: I think I've seen some mysteries (like errors causing other errors) over the years that could be explained by this.

harrishancock (approved):

c++/src/kj/compat/http.c++
harrishancock: Though this might have introduced some test failures related to body-discarding: https://github.com/capnproto/capnproto/actions/runs/4079783138/jobs/7031513983#step:4:2429

kentonv: It would appear I failed to actually run Cap'n Proto's tests while developing this. :facepalm:

PR #1624, opened 2023-02-10
Give deferred function guards the ability to run before destruction

bcaimano:
This commit adds three new functions to `kj::_:Deferred`:
- A move-assignment operator to allow reuse of the same variable for deferred functors of the same type.
- A `run()` function to immediately run the guard before destruction.
- A `nullptr` assignment operator (nullification) as syntactic sugar for `run()`.

It's worth mentioning that via the magic of kj::Maybe and universal references, `Deferred` maintains the ability to hold lvalue reference to functors instead of functor instances.

### DIFF
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -1976,34 +1976,48 @@ namespace _ {  // private
 template <typename Func>
 class Deferred {
 public:
-  inline Deferred(Func&& func): func(kj::fwd<Func>(func)), canceled(false) {}
-  inline ~Deferred() noexcept(false) { if (!canceled) func(); }
+  Deferred(Func&& func): maybeFunc(kj::fwd<Func>(func)) {}
+  ~Deferred() noexcept(false) {
+    run();
+  }
   KJ_DISALLOW_COPY(Deferred);
 
-  // This move constructor is usually optimized away by the compiler.
-  inline Deferred(Deferred&& other): func(kj::fwd<Func>(other.func)), canceled(other.canceled) {
-    other.canceled = true;
+  Deferred(Deferred&&) = default;
+  // Since we use a kj::Maybe, the default move constructor does exactly what we want it to do.
+
+  void run() {
+    // Move `maybeFunc` to the local scope so that even if we throw, we destroy the functor we had.
+    auto maybeLocalFunc = kj::mv(maybeFunc);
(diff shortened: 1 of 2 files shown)


c++/src/kj/common.h
```diff
@@ -1976,22 +1976,43 @@ namespace _ {  // private
...
+  }
   KJ_DISALLOW_COPY(Deferred);
 
-  // This move constructor is usually optimized away by the compiler.
-  inline Deferred(Deferred&& other): func(kj::fwd<Func>(other.func)), canceled(other.canceled) {
-    other.canceled = true;
+  Deferred(Deferred&&) = default;
+  // Since we use a kj::Maybe, the default move operator does exactly what we want it to do.
+
+  Deferred& operator=(Deferred&& other) {
```
kentonv, line 1988: Hmm, do you have a use case in mind for this assignment operator? In the common case that `Func` is a lambda, you'd have to be assigning another defer created from the exact same lambda, which seems unlikely. I guess technically you could pass something that isn't a lambda to `kj::defer()` but I'm not sure if I've ever seen that done.

bcaimano: I definitely agree on it being for limited use. I think the main use case I was thinking of was a state machine where you'd only have one guard at a time:
```
auto makeStepGuard(auto stepIdx) {
  return [stepIdx]() {
    useArgs(stepIdx);
  };
};

kj::Promise<void> myFunc() {
  auto stepGuard = kj::defer(makeStepGuard(stepOne));
  // Do step one...
  co_await doStepOne();
  stepGuard = kj::defer(makeStepGuard(stepTwo));
  co_await doStepTwo();
  ...
}
```
I'm sure that I could phrase anything I could do with the assignment operator with scopes instead, it just might be tricky with more complicated control flow and function splits.

Re: something that isn't a lambda, I suspect we've probably passed in a `kj::Function<void()>` somewhere in our code base but yeah, gonna be pretty rare that it isn't a lambda.

kentonv: Hmm, I feel like in such a use case you should really just write a class with a destructor. `kj::defer` is no longer much more concise in this case and is kind of mind-bendy to understand. A class with a simple destructor will be more straightforward.

So I lean towards not adding these assignment operators. I agree with adding `run()`, though.

PR #1625, opened 2023-02-13
Emit a different error if CAPNP_VERSION is not defined in generated headers

bcaimano:
Some tools will ignore missing headers, so if 'capnp/generated-header-support.h' is not in the include path, we would warn about a version mismatch instead of a missing include.

### DIFF
--- a/c++/src/capnp/c++.capnp.h
+++ b/c++/src/capnp/c++.capnp.h
@@ -6,7 +6,9 @@
 #include <capnp/generated-header-support.h>
 #include <kj/windows-sanity.h>
 
-#if CAPNP_VERSION != 11000
+#ifndef CAPNP_VERSION
+#error "CAPNP_VERSION is not defined, is capnp/generated-header-support.h missing?"
+#elif CAPNP_VERSION != 11000
 #error "Version mismatch between generated code and library headers.  You must use the same version of the Cap'n Proto compiler and library."
 #endif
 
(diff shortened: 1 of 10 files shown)


kentonv:
Since this changes the generated code, you'll need to regenerate bootstraps (`c++/regenerate-bootstraps.sh`).

`bootstrap-test.ekam-rule` is probably failing but only runs under Ekam... Maybe we need to port that to other build systems.

c++/src/capnp/cc_capnp_library.bzl
```diff
@@ -121,7 +121,7 @@ def cc_capnp_library(
         name = name,
         srcs = srcs_cpp,
         hdrs = hdrs,
-        deps = deps + ["@capnp-cpp//src/capnp:capnp_runtime"],
+        deps = deps + ["@capnp-cpp//src/capnp:capnp", "@capnp-cpp//src/capnp:capnp_runtime"],
```
kentonv, line 124: What's the story behind this line?

bcaimano: Ugh, apologies, this got pulled in from the work I was building on top of this. The short answer is that we might be technically violating bazel's indirect inclusion rules: https://bazel.build/reference/be/c-cpp#hdrs. I'm uncertain and still figuring it out. In any case, shouldn't be in this PR.

PR #1626, opened 2023-02-15
Fix stringification of negative `kj::Duration`.

kentonv:


c++/src/kj/time.c++
```diff
@@ -268,6 +268,9 @@ CappedArray<char, sizeof(int64_t) * 3 + 2 + 4> KJ_STRINGIFY(Date d) {
   return kj::toCharSequence(d - UNIX_EPOCH);
 }
 CappedArray<char, sizeof(int64_t) * 3 + 2 + 4> KJ_STRINGIFY(Duration d) {
```
jclee, line 270: Not important, but I'm not actually sure exactly what the `n*3 + 2 + 4` is intended to contain... I guess `n*3` is meant to cover the digits themselves, and the `2` is maybe to cover the two-character suffix.  And the `4` to cover the decimal point, null character, and maybe some padding for safety?  Not sure if the upper-ascii micro symbol incurs extra space in some environments or something.  A moot point here, though, since there seems to be plenty of room in the digits estimation to accommodate an additional negative sign at maxint.

jclee (approved):

c++/src/kj/time.c++
```diff
@@ -268,6 +268,9 @@ CappedArray<char, sizeof(int64_t) * 3 + 2 + 4> KJ_STRINGIFY(Date d) {
...
+  bool negative = d < 0 * kj::SECONDS;
+  if (negative) d = 0 * kj::SECONDS - d;
```
jclee, line 272: Oh, I guess the negative side of the signed integer range is slightly larger than the positive one, so this might still format one value weird?

c++/src/kj/time.c++
```diff
@@ -268,6 +268,9 @@ CappedArray<char, sizeof(int64_t) * 3 + 2 + 4> KJ_STRINGIFY(Date d) {
...
+
   auto digits = kj::toCharSequence(d / kj::NANOSECONDS);
```
jclee, line 274: FWIW, it's not quite clear to me why `kj::toCharSequence()`'s existing handling of signed types isn't already adding a negative sign.  I'd expect `d / kj::NANOSECONDS` to yield an `int64_t`, which I would have thought would be covered by one of the existing `STRINGIFY_INT()` overloads.

kentonv: The problem is the logic below that chooses an appropriate unit doesn't account for the negative sign correctly.

c++/src/kj/time.c++
kentonv: No, you were right the first time -- the most-negative value is stringified incorrectly!

c++/src/kj/time.c++
kentonv: Oh yeah, the μ takes two bytes (UTF-8). I probably didn't think of that when writing this, but it does appear I added extra padding anyway. I'm going to replace `2 + 4` with `8` to be safe (and less confusing).

c++/src/kj/time.c++
jclee: Ohhh... that makes more sense, thanks.

c++/src/kj/time.c++
jclee: :open_mouth:

PR #1627, opened 2023-02-16
Work around broken open("/") under qemu-user.

kentonv:
This work-around has been used internally in the Workers Runtime codebase for a long time, but is kind of awkward to apply there. Baking it into KJ makes things cleaner.


kentonv:
Merged before review because I've been trying to merge the dependent change for weeks and all the tests passed to I did it... let me know if you have any issues.

Issue #1623, opened 2023-02-09
Waiting for executeSync to complete in Windows may hang forever


fabiorossetto:
We have observed that under certain conditions `executeSync` may hang forever in Windows. Unfortunately I could not create a simple standalone example to reproduce the issue. However, I think I know what the issue is.

When the executor is sleeping, it can be resumed by calling `wake()`. On Windows, this causes `PostQueuedCompletionStatus` to be called. This in turn causes `waitIocp` to resume from sleep. 

So the standard flow of `executeSync`, as far as I understand it is:
- Post an event to the executor event loop
- Call `EventPort::wake()` to resume the event loop
- Wait until the event has executed and the result is available

Howeve I think there's a catch. `waitIocp` is called also by `~IoPromiseAdapter()`. So if `wake()` is called while waitIocp is executing within `IoPromiseAdapter` destructor, by the time `wait()` is called the event sent by `wake` has already been consumed. Thus `wait` waits to be resumed forever and `executeSync` hangs.

Can someone have a look at the code and confirm this can happen?

kentonv:
Hi @fabiorossetto,

Yes, your analysis looks correct to me. It looks like `wait()` needs to call `receivedWake()` at the start of the function and, if it returns true, perform a poll instead and return true.

Would you care to submit a PR? You might have an easier time testing this than me as I don't actually have a Windows machine.

fabiorossetto:
I'll try to make one. I should have some time in two weeks

fabiorossetto:
I've created an MR, can you check if this is what you had in mind? https://github.com/capnproto/capnproto/pull/1631

PR #1631, opened 2023-02-23
Fix potential hang of EventPort::wait on Windows

fabiorossetto:
This commit fixes a bug observed under Windows. It can happen that wake() is concurrently called while waitIocp is executing within the IoPromiseAdapter destructor. In such case, when wait() is called, the event posted by wake() has already been consumed and thus wait() waits forever to be resumed.

See issue #1623.

### DIFF
--- a/c++/src/kj/async-win32.c++
+++ b/c++/src/kj/async-win32.c++
@@ -158,13 +158,21 @@ Own<Win32EventPort::SignalObserver> Win32IocpEventPort::observeSignalState(HANDL
 }
 
 bool Win32IocpEventPort::wait() {
-  waitIocp(timerImpl.timeoutToNextEvent(clock.now(), MILLISECONDS, INFINITE - 1)
-      .map([](uint64_t t) -> DWORD { return t; })
-      .orDefault(INFINITE));
+  const bool woke = receivedWake();
+  if(woke) {
+    waitIocp(0);
+    return true;
+  }
+  try{
+    waitIocp(timerImpl.timeoutToNextEvent(clock.now(), MILLISECONDS, INFINITE - 1)
+        .map([](uint64_t t) -> DWORD { return t; })
+        .orDefault(INFINITE));
 
-  timerImpl.advanceTo(clock.now());
-
-  return receivedWake();
+    timerImpl.advanceTo(clock.now());
+  }catch(...) {
+    
+  }
+  return woke;
 }
 
 bool Win32IocpEventPort::poll() {
--- a/c++/src/kj/async-win32.c++
+++ b/c++/src/kj/async-win32.c++
@@ -158,21 +158,18 @@ Own<Win32EventPort::SignalObserver> Win32IocpEventPort::observeSignalState(HANDL
 }
 
 bool Win32IocpEventPort::wait() {
-  const bool woke = receivedWake();
-  if(woke) {
-    waitIocp(0);
+  // It's possible that a wake event was received and discarded during ~IoPromiseAdapter. We
+  // need to check for that now. Otherwise, calling waitIocp may cause it to hang forever.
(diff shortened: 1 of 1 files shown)


c++/src/kj/async-win32.c++
```diff
@@ -158,13 +158,21 @@ Own<Win32EventPort::SignalObserver> Win32IocpEventPort::observeSignalState(HANDL
...
+  }
+  try{
+    waitIocp(timerImpl.timeoutToNextEvent(clock.now(), MILLISECONDS, INFINITE - 1)
+        .map([](uint64_t t) -> DWORD { return t; })
+        .orDefault(INFINITE));
+
+    timerImpl.advanceTo(clock.now());
+  }catch(...) {
+    
+  }
```
kentonv, line 174: I don't think we want to be hiding exceptions here?

```suggestion

  waitIocp(timerImpl.timeoutToNextEvent(clock.now(), MILLISECONDS, INFINITE - 1)
      .map([](uint64_t t) -> DWORD { return t; })
      .orDefault(INFINITE));

  timerImpl.advanceTo(clock.now());
```

c++/src/kj/async-win32.c++
```diff
@@ -158,13 +158,21 @@ Own<Win32EventPort::SignalObserver> Win32IocpEventPort::observeSignalState(HANDL
...
-      .orDefault(INFINITE));
-
-  timerImpl.advanceTo(clock.now());
-
-  return receivedWake();
+  const bool woke = receivedWake();
+  if(woke) {
+    waitIocp(0);
+    return true;
+  }
```
kentonv, line 165: So I thought about it some more, and decided the poll here doesn't actually make sense. `~IoPromiseAdapter` actually did its own poll, after all.

```suggestion
  // It's possible that a wake event was received and discarded during ~IoPromiseAdapter. We
  // need to check for that now.
  if (receivedWake()) return true;
```

c++/src/kj/async-win32.c++
```diff
@@ -158,13 +158,21 @@ Own<Win32EventPort::SignalObserver> Win32IocpEventPort::observeSignalState(HANDL
...
+  try{
+    waitIocp(timerImpl.timeoutToNextEvent(clock.now(), MILLISECONDS, INFINITE - 1)
+        .map([](uint64_t t) -> DWORD { return t; })
+        .orDefault(INFINITE));
+
+    timerImpl.advanceTo(clock.now());
+  }catch(...) {
+    
+  }
+  return woke;
```
kentonv, line 175: `woke` is always false here... so this means we're now ignoring cross-thread wakes that occurred during the wait that we just completed. It would still appear to work since the _next_ wait would see the wakeup, but we shouldn't delay the notification.

```suggestion
  return receivedWake();
```

kentonv (commented):
Thanks for working on this. Were you able to verify this actually fixes the problem you observed in production? Or is this PR just based on the theory? (I do agree it's a real bug either way, just wondering if it's definitely _the_ bug you were seeing.)

c++/src/kj/async-win32.c++
fabiorossetto: No idea why I added it...

fabiorossetto:
> Thanks for working on this. Were you able to verify this actually fixes the problem you observed in production? Or is this PR just based on the theory? (I do agree it's a real bug either way, just wondering if it's definitely _the_ bug you were seeing.)

We are working on it. I'll hope I'll be able to get you an answer soon

fabiorossetto:
Hi Kenton, from our tests it seems that the patch fixes the issue

kentonv:
Great, thanks! I'll merge this once the CI is complete.

PR #1633, opened 2023-03-02
Implements ConnectSettings in HttpClient::connect.

dom96:


c++/src/kj/compat/http.h
```diff
@@ -748,7 +748,12 @@ class HttpClient {
     kj::Own<kj::AsyncIoStream> connection;
   };
 
-  virtual ConnectRequest connect(kj::StringPtr host, const HttpHeaders& headers);
+  struct ConnectSettings {
+    bool useTls;
```
kentonv, line 752: All settings should have defaults so people don't have to manually initialize them.

```suggestion
    bool useTls = false;
```

Then you can change all the places where you've written `{ .useTls = false }` to be just `{}`.

(That in turn should, by coincidece, help fix the compile errors on Windows/MSVC, where designated initializers are not supported in C++14 mode, which is what Cap'n Proto currently uses.)

kentonv, line 752: Needs a doc comment, like:

> Requests to automatically establish a TLS session over the connection. The remote party will be expected to present a valid certificate matching the requested hostname.

c++/src/kj/compat/http.c++
```diff
@@ -5674,11 +5679,18 @@ public:
...
   }
 
-  ConnectRequest connect(kj::StringPtr host, const HttpHeaders& headers) override {
+  ConnectRequest connect(
+      kj::StringPtr host, const HttpHeaders& headers, ConnectSettings settings) override {
     // We want to connect directly instead of going through a proxy here.
     // https://github.com/capnproto/capnproto/pull/1454#discussion_r900414879
+    kj::Maybe<kj::Promise<kj::Own<kj::NetworkAddress>>> addr;
+    if (settings.useTls) {
+      addr = KJ_ASSERT_NONNULL(tlsNetwork).parseAddress(host);
```
kentonv, line 5688: This is an error on the caller's part, so should be REQUIRE instead of ASSERT.

```suggestion
      addr = KJ_REQUIRE_NONNULL(tlsNetwork, "this HttpClient doesn't support TLS")
          .parseAddress(host);
```

c++/src/kj/compat/http.c++
```diff
@@ -5210,7 +5210,8 @@ public:
     });
   }
 
-  ConnectRequest connect(kj::StringPtr host, const HttpHeaders& headers) override {
+  ConnectRequest connect(
+      kj::StringPtr host, const HttpHeaders& headers, ConnectSettings settings) override {
```
kentonv, line 5214: It appears this version of `connect()` (for `HttpClientImpl`, which makes a `CONNECT` request) ignores the `useTls` setting, which I think people would find surprising.

I think the right answer here is to allow an optional `TlsContext` to be provided when creating the client, and automatically use it to upgrade the connection once established. This would match regular HTTP proxy conventions. If the caller doesn't actually want TLS to apply at this level, and instead wants the destination proxy server to deal with it, it would be up to the caller to rewrite the setting into a header or whatever.

You could also KJ_UNIMPLEMENTED for now.

kentonv (commented):
Took a quick skim, didn't read everything yet.

dom96:
@kentonv CI green. Ready to go.

harrishancock (approved):

Warfields:
This PR has introduced a segfault

PR #1635, opened 2023-03-09
Implements StartTLS in HttpConnectSettings and NetworkHttpClient.

dom96:
* In order to access `TlsContext` from a `TlsNetwork` I needed to introduce a new SecureNetworkWrapper.
* The `tlsStarter` callback API isn't perfect, but seems to be our best idea. If anyone has others please share.

### Test Plan

```
bazel test @capnp-cpp//src/kj/compat:tls-test --config=asan --test_timeout 5
```

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -5621,10 +5621,10 @@ public:
       return c->get()->connect(host, headers, settings);
     } else {
       auto split = promise.addBranch().then(
-          [this, host=kj::str(host), headers=headers.clone(), settings]()
+          [this, host=kj::str(host), headers=headers.clone(), settings]() mutable
           -> kj::Tuple<kj::Promise<ConnectRequest::Status>,
                        kj::Promise<kj::Own<kj::AsyncIoStream>>> {
-        auto request = KJ_ASSERT_NONNULL(client)->connect(host, headers, settings);
+        auto request = KJ_ASSERT_NONNULL(client)->connect(host, headers, kj::mv(settings));
         return kj::tuple(kj::mv(request.status), kj::mv(request.connection));
       }).split();
 
@@ -5644,7 +5644,7 @@ private:
 class NetworkHttpClient final: public HttpClient, private kj::TaskSet::ErrorHandler {
 public:
   NetworkHttpClient(kj::Timer& timer, const HttpHeaderTable& responseHeaderTable,
-                    kj::Network& network, kj::Maybe<kj::Network&> tlsNetwork,
+                    kj::Network& network, kj::Maybe<kj::SecureNetworkWrapper&> tlsNetwork,
                     HttpClientSettings settings)
       : timer(timer),
         responseHeaderTable(responseHeaderTable),
@@ -5689,8 +5689,8 @@ public:
     // https://github.com/capnproto/capnproto/pull/1454#discussion_r900414879
     kj::Maybe<kj::Promise<kj::Own<kj::NetworkAddress>>> addr;
     if (settings.useTls) {
-      addr = KJ_REQUIRE_NONNULL(tlsNetwork, "this HttpClient doesn't support TLS")
-          .parseAddress(host);
+      kj::Network& tlsNet = KJ_REQUIRE_NONNULL(tlsNetwork, "this HttpClient doesn't support TLS");
+      addr = tlsNet.parseAddress(host);
     } else {
       addr = network.parseAddress(host);
     }
@@ -5709,9 +5709,30 @@ public:
       }).attach(kj::mv(address));
     }).split();
 
+    auto connection = kj::newPromisedStream(kj::mv(kj::get<1>(split)));
+
+    #if KJ_HAS_OPENSSL
+    if (!settings.useTls) {
+      KJ_IF_MAYBE(tlsNet, tlsNetwork) {
+        KJ_IF_MAYBE(tlsStarter, settings.tlsStarter) {
+          auto refConnection = kj::refcountedWrapper(kj::mv(connection));
+          connection = refConnection->addWrappedRef();
+          kj::Own<kj::AsyncIoStream> ref1 = refConnection->addWrappedRef();
+          Function<kj::Own<kj::AsyncIoStream>(kj::StringPtr)> cb =
+              [tlsNet, ref1 = kj::mv(ref1)](kj::StringPtr expectedServerHostname) mutable {
+            kj::Promise<kj::Own<kj::AsyncIoStream>> secureStream =
+                tlsNet->wrapClient(kj::mv(ref1), expectedServerHostname);
+            return kj::newPromisedStream(kj::mv(secureStream));
+          };
+          *tlsStarter = kj::mv(cb);
+        }
+      }
+    }
+    #endif
+
     return ConnectRequest {
       kj::mv(kj::get<0>(split)),
-      kj::newPromisedStream(kj::mv(kj::get<1>(split)))
+      kj::mv(connection)
     };
   }
 
@@ -5719,7 +5740,7 @@ private:
   kj::Timer& timer;
   const HttpHeaderTable& responseHeaderTable;
   kj::Network& network;
-  kj::Maybe<kj::Network&> tlsNetwork;
+  kj::Maybe<kj::SecureNetworkWrapper&> tlsNetwork;
   HttpClientSettings settings;
 
   struct Host {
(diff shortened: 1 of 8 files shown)


c++/src/kj/compat/http.c++
```diff
@@ -5709,17 +5709,34 @@ public:
       }).attach(kj::mv(address));
     }).split();
 
+    auto connection = kj::refcountedWrapper(kj::newPromisedStream(kj::mv(kj::get<1>(split))));
+
+    // XXX: I'm uncertain whether this is the best approach.
+    kj::Own<kj::AsyncIoStream> ref1 = connection->addWrappedRef();
```
dom96, line 5715: I'm uncertain about lifetimes here, but I need the tlsStarter closure to capture a reference to the AsyncIoStream so that it can upgrade it.

c++/src/kj/compat/http.c++
```diff
@@ -5709,17 +5709,36 @@ public:
...
+    KJ_IF_MAYBE(tlsNet, tlsNetwork) {
+      kj::TlsContext& tlsContext = tlsNet->getTlsContext();
+      KJ_IF_MAYBE(tlsStarter, settings.tlsStarter) {
+        Function<kj::Own<kj::AsyncIoStream>(kj::StringPtr)> cb =
+            [&tlsContext, ref1 = kj::mv(ref1)](kj::StringPtr expectedServerHostname) mutable {
+          kj::Promise<kj::Own<kj::AsyncIoStream>> secureStream =
+              tlsContext.wrapClient(kj::mv(ref1), expectedServerHostname);
+          return kj::newPromisedStream(kj::mv(secureStream));
+        };
+        *tlsStarter = kj::mv(cb);
```
jclee, line 5726: It looks like the type of `settings.tlsStarter` is `kj::Maybe<TlsStarterCallback&>`... I think this assignment will just capture a reference to the `Function` object in this scope, such that `cb` 's contents will be destroyed at the end of the scope, making the reference invalid.  Probably want the type of `settings.tlsStarter` to be `kj::Maybe<TlsStarterCallback>` instead?

Also, looks like the type of `TlsStarterCallback` itself contains a Maybe, which is possibly redundant?

dom96: The idea here is that the caller of `connect` specifies a reference to a `Function` so that `connect` can fill that with a callback. As far as I understand I am moving the `Function` into that reference here. From my testing it seems to be working but maybe I just got lucky and am missing something.

The reason `TlsStarterCallback` contains a Maybe is because the `connect` might not fill the callback (if for example there is no TLS support). This is `settings.tlsStarter` is two Maybes, one because the caller might not want the callback to be filled anywhere (in which case they set this option to `nullptr`) and the second because of what I outlined previously.

Indeed, the API is ugly. We had a discussion about how to handle this and this was what we landed on. If anyone has better ideas please feel free to share.

Good point on the assert.

c++/src/kj/compat/http.h
```diff
@@ -660,10 +661,15 @@ class WebSocket {
...
 
+using TlsStarterCallback = kj::Maybe<kj::Function<kj::Own<kj::AsyncIoStream>(kj::StringPtr)>>;
 struct HttpConnectSettings {
   bool useTls = false;
   // Requests to automatically establish a TLS session over the connection. The remote party
   // will be expected to present a valid certificate matching the requested hostname.
+  kj::Maybe<TlsStarterCallback&> tlsStarter;
+  // An optional reference to a callback. If it's non-null then it may get filled with a
+  // callback function. This callback function can then be used to initiate a TLS handshake on the
+  // connection. The returned AsyncIoStream is a secure TLS stream.
```
harrishancock, line 672: I'd suggest a couple clarifications:

1. This is an output parameter. Naming it something like `tlsStarterOutputParam` would be ugly, but make it clearer what's going on.
2. It may be worth mentioning that if `tlsStarter` is non-null, but the referent is null after `connect()` is called, that means libkj was compiled without TLS support.
3. What happens if `useTls = true`, and `tlsStarter` is non-null? Is that something we should allow?

c++/src/kj/compat/http.c++
harrishancock: A differently-ugly option could be to make `ConnectRequest::connection` a `kj::OneOf<kj::Own<kj::AsyncIoStream>, kj::Own<kj::TlsStartableAsyncIoStream>>`, the latter type growing a `startTls()` member function which is callable once and either upgrades the connection in situ, or returns a new AsyncIoStream, invalidating the old one. That could potentially make lifetimes clearer.

It'd be possible to avoid the OneOf with a separate `connectWithStartTls()` function (perhaps only available if KJ_USE_OPENSSL is true) which returns the one-and-only type.

harrishancock (commented):
Is it possible to write a test, even just a smoke test, for this?

c++/src/kj/compat/http.h
dom96: > What happens if useTls = true, and tlsStarter is non-null? Is that something we should allow?

No, in this case the tlsStarter will not be filled. I'll mention this in the comment as well.

c++/src/kj/compat/http.h
```diff
@@ -904,7 +919,7 @@ struct HttpClientSettings {
 };
 
 kj::Own<HttpClient> newHttpClient(kj::Timer& timer, const HttpHeaderTable& responseHeaderTable,
-                                  kj::Network& network, kj::Maybe<kj::Network&> tlsNetwork,
+                                  kj::Network& network, kj::Maybe<kj::TlsNetwork&> tlsNetwork,
```
kentonv, line 922: This is a backwards-incompatible API change which also narrows this function's capabilities by replacing an abstract interface with a concrete implementation. The new dependency on KJ TLS with #ifdefs that avoid it based on whether OpenSSL is available is all kind of ugly, too.

Could HttpClientSettings instead be extended with an optional callback function that could be used here, without introducing any direct dependency on `tls.h`?

kentonv: The new option could be:

`kj::Function<kj::Own<kj::AsyncIoStream>(kj::Own<kj::AsyncIoStream> inner, kj::StringPtr expectedHostname)> startTls;`

Then again, if a new option is required anyway, then perhaps it's not worth extending the default HttpClient implementation to support it. People that need this functionality might as well use a wrapper? (That's what we're going to be doing anyway.)

kentonv: Perhaps another possible approach is to not extend `HttpConnectSettings` at all, but instead pass down this callback as part of `SubrequestMetadata` in the workers runtime. Ugly in its own way but keeps the weirdness out of KJ... I dunno.

dom96: Maybe even more simply: could `HttpClientSettings` hold a `TlsContext&`? This is the only reason I need the TlsNetwork.

kentonv: That still introduces a dependency on kj-tls. I'd like to avoid that.

What you could do is factor out an abstract interface which `TlsContext` implements. Maybe it could be called `SecureNetworkWrapper`. Put it in some common location that both kj-http and kj-tls depend on... probably `async-io.h`.

c++/src/kj/compat/http.c++
```diff
@@ -5709,17 +5709,38 @@ public:
...
+        kj::TlsContext& tlsContext = tlsNet->getTlsContext();
+        KJ_IF_MAYBE(tlsStarter, settings.tlsStarter) {
+          kj::Own<kj::AsyncIoStream> ref1 = connection->addWrappedRef();
+          Function<kj::Own<kj::AsyncIoStream>(kj::StringPtr)> cb =
+              [&tlsContext, ref1 = kj::mv(ref1)](kj::StringPtr expectedServerHostname) mutable {
+            kj::Promise<kj::Own<kj::AsyncIoStream>> secureStream =
+                tlsContext.wrapClient(kj::mv(ref1), expectedServerHostname);
+            return kj::newPromisedStream(kj::mv(secureStream));
+          };
+          KJ_DBG("Setting tlsStarter");
```
kentonv, line 5726: KJ_DBG

dom96:
Removed the breaking API change.

c++/src/kj/compat/http.c++
```diff
@@ -5709,17 +5709,36 @@ public:
       }).attach(kj::mv(address));
     }).split();
 
+    auto connection = kj::refcountedWrapper(kj::newPromisedStream(kj::mv(kj::get<1>(split))));
```
harrishancock, line 5712: Nit: the overhead of the refcounted wrapper isn't needed if the conditionals below aren't taken.

harrishancock (approved):

PR #1636, opened 2023-03-13
Fulfill readAbortFulfiller in AsyncPipe destructor.

dom96:
This creates errors like: https://gist.github.com/dom96/c77bccefb73c9bd162bf5028f42853cf.

### DIFF
--- a/c++/src/kj/async-io.c++
+++ b/c++/src/kj/async-io.c++
@@ -217,6 +217,10 @@ public:
       // Don't std::terminate().
       break;
     }
+
+    KJ_IF_MAYBE(raf, readAbortFulfiller) {
+      raf->get()->fulfill();
+    }
   }
 
   Promise<size_t> tryRead(void* buffer, size_t minBytes, size_t maxBytes) override {


kentonv:
This is not the right fix. If some code is still waiting on `whenWriteDisconnected()` when the stream is destroyed, that's a UAF. You need to cancel the `whenWriteDisconnected()` promise before destroying the stream it came from.

We could arguably change this destructor so that if `fulfiller->isWaiting()` returns true, it `reject()`s it with an exception saying that a UAF has occurred.

kentonv:
FWIW this is the reasoning I followed here:

1. The PR didn't explain why this is the right fix, so I went to the code to understand what's going on.
2. Discovered that `readAbortFulfiller` is used to implement `whenWriteDisconnected()`. It seems obvious enough that this promise _should_ resolve when the stream is discraded, right? So for a moment I thought this may be the right fix.
3. I was going to ask for a unit test, but I checked if there were existing unit tests.
4. There was a test, covering exactly the situation that seemed like it would be broken. So why was that test not broken?
5. I realized at this point that `PipeReadEnd` and `TwoWayPipeEnd` are responsible for calling `abortRead()` on the underlying `AsyncPipe` when they are destroyed. At this point I remembered that `AsyncPipe` is a refcounted object, with the read end and the write end each holding a reference.
6. But this implies that if `readAbortFulfiller` was still being waited on in `AsyncPipe`'s destructor, then that's a UAF violation on the part of whoever called `whenWriteDisconnected()`.

kentonv:
BTW I'm also confused why the `abortRead()` calls in `~PipeReadEnd` and `~TwoWayPipeEnd` aren't getting the job done, regardless of the UAF issue. They should still be fulfilling the promise before the `AsyncPipe` gets destroyed, shouldn't they?

PR #1638, opened 2023-03-14
Allow standard headers to be overriden in HttpClient.

dom96:
As of right now if I set a `Connection` or an `Upgrade` header in my call to `connect` then it does not get sent to the client. This PR makes a simple change to fix that, specifically if these headers are not set in `connectionHeaders` then the custom headers passed to the HTTP call are used.

I didn't look too deeply into the repercussions of doing it this way. Will try to do that later but in the meantime maybe one of the reviewers can tell me whether this is the right approach.


dom96:
Hm, maybe I should put this behind a config option in HttpClientSettings

kentonv:
No, sorry, this introduces potential security flaws. `Connection` and `Upgrade` are headers that change the underlying behavior of the HTTP connection and need to be understood by the HTTP implementation. This is why they are not allowed to be set by apps. I am not comfortable introducing this even behind a flag because we would presumably need to turn that flag on in Workers which is the place I'm most afraid of these kinds of bugs.

There are two options:
1. Don't use the `Upgrade` header. Use a header that does not have special meaning.
2. Introduce a new option for `connect()` that says that instead of actually using the `CONNECT` method, it should use `GET` with `Upgrade`. A `101 Switching Protocols` response indicates acceptance; any other response code goes down the rejection path where we return a regular HTTP body. This implements `Upgrade` in a standards-compliant way, while giving the application a raw stream so that it can implement the upgraded protocol on its own.

Note that using `CONNECT` with `Upgrade` is explicitly not one of the options here, because there is no standard interpretation of what this should mean. IMO any protocol requiring this combination is non-standard and possibly standard-hostile and so we should not support it. (This is the argument I made on the internal thread.)

PR #1639, opened 2023-03-16
Add Doubly Linked-List that Owns its Nodes

MellowYarker:
We need a dub LL to store hibernatable websockets in the `HibernationManager` (see https://github.com/cloudflare/workerd/pull/449/commits/11b4584c1ed8cac89851a26ac123b048b7d1d71a#diff-4801bd448b3ceead290776b1dbf2469a0a860ed064e81f2e07f01c26535deaf1R166-R169).

We thought we could use `kj::List` here (since we use it elsewhere for hibernation) but our overarching reason for using a doubly linked-list is for fast deletes. If the `Item` holding a `HibernatableWebSocket` had to be "owned" by some other collection (since `kj::List::add()` takes a reference), then we would still need to remove the owned object from the other collection.

### Other Ideas:
1. Use a regular collection like a `Vector`, but rather than "removing" the entry from the collection when websockets disconnect, we can use a `kj::Maybe<kj::Own<HibernatableWebSocket>>` and set it to `nullptr` on disconnect, then remove all disconnected websockets in bulk every so often.
2. Could try to pull in another header-only implementation of a doubly linked-list from some large c++ project?


a-robinson:
Sorry, I haven't really looked at the exact needs of the code that wants to use this, but "Doubly Linked-List that Owns its Nodes" is just a `std::list`, isn't it? Use `std::list<std::unique_ptr<T>>` or `std::list<kj::Own<T>>` if you want to store heap-allocated objects tracked by a unique/owned pointer.

MellowYarker:
:man_facepalming: as dumb as it sounds I kind of assumed `std::` was taboo in the codebase, but it looks like we actually use `std::list` in a few places so... lesson learned!! Very very easy and quick fix as far as my other PRs are concerned. Thanks for the suggestion, Alex!

harrishancock:
The main issue with mixing `std` owning containers with KJ objects is that it turns exceptions thrown from element destructors into crashes, assuming the container is ever destroyed with elements still inside of it. If your use case has objects with `noexcept(false)` destructors, you might need to consider this.

PR #1640, opened 2023-03-16
Fix Bazel builds on Windows

mrbbot:
_(dependency of https://github.com/cloudflare/workerd/pull/452)_

Hey! 👋 This PR enables Cap'n Proto and KJ to be built and tested on Windows under Bazel, in addition to CMake. It also makes a few changes to enable `workerd` to be built on Windows. I've configured Bazel to use `clang-cl` as that's what `workerd` needs, but I suspect native MSVC should work.

### DIFF
--- a/c++/src/kj/async-io-win32.c++
+++ b/c++/src/kj/async-io-win32.c++
@@ -23,7 +23,7 @@
 // For Unix implementation, see async-io-unix.c++.
 
 // Request Vista-level APIs.
-#include "win32-api-version.h"
+#include <kj/win32-api-version.h>
 
 #include "async-io.h"
 #include "async-io-internal.h"
@@ -1086,7 +1086,7 @@ public:
       : lowLevel(parent.lowLevel), filter(allow, deny, parent.filter) {}
 
   Promise<Own<NetworkAddress>> parseAddress(StringPtr addr, uint portHint = 0) override {
-    return evalLater([this,portHint,addr=kj::mv(addr)]() {
+    return evalNow([&]() {
       return SocketAddress::parse(lowLevel, addr, portHint, filter);
     }).then([this](Array<SocketAddress> addresses) -> Own<NetworkAddress> {
       return heap<NetworkAddressImpl>(lowLevel, filter, kj::mv(addresses));
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -1314,7 +1314,7 @@ inline T* readMaybe(T* ptr) { return ptr; }
 
 #define KJ_IF_MAYBE(name, exp) if (auto name = ::kj::_::readMaybe(exp))
 
-#if __GNUC__
+#if __GNUC__ || __clang__
 // These two macros provide a friendly syntax to extract the value of a Maybe or return early.
 //
 // Use KJ_UNWRAP_OR_RETURN if you just want to return a simple value when the Maybe is null:
@@ -1347,6 +1347,9 @@ inline T* readMaybe(T* ptr) { return ptr; }
 // "statement expressions" extension. IIFEs don't do the trick here because a lambda cannot
 // return out of the parent scope. These macros should therefore only be used in projects that
 // target GCC or GCC-compatible compilers.
+//
+// `__GNUC__` is not defined when using LLVM's MSVC-compatible compiler driver `clang-cl` (even
+// though clang supports the required extension), hence the additional `|| __clang__`.
 
 #define KJ_UNWRAP_OR_RETURN(value, ...) \
   (*({ \
--- a/c++/src/kj/compat/tls-test.c++
+++ b/c++/src/kj/compat/tls-test.c++
@@ -637,7 +637,10 @@ KJ_TEST("TLS full duplex") {
 
   auto writeUp = writeN(*client, "foo", 10000);
   auto readDown = readN(*client, "bar", 10000);
+#if !(_WIN32 && __clang__)
+  // TODO(soon): work out why this expectation fails even with the above fix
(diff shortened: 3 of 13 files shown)


c++/src/kj/async-io-win32.c++
```diff
@@ -1086,8 +1086,9 @@ public:
       : lowLevel(parent.lowLevel), filter(allow, deny, parent.filter) {}
 
   Promise<Own<NetworkAddress>> parseAddress(StringPtr addr, uint portHint = 0) override {
-    return evalLater([this,portHint,addr=kj::mv(addr)]() {
-      return SocketAddress::parse(lowLevel, addr, portHint, filter);
+    auto addrCopy = kj::str(addr);
+    return evalLater([this,portHint,addrCopy=kj::mv(addrCopy)]() {
+      return SocketAddress::parse(lowLevel, addrCopy, portHint, filter);
```
mrbbot, line 1091: Without this change, we were seeing the following error when two `workerd` services were defined with service bindings to the same external server service:

```
workerd/jsg/util.c++:423: error: e = kj/async-io-win32.c++:813: failed: getaddrinfo(): #11001 No such host is known.; params.host = αN·Φô☻ÇN·Φô☻; params.service =
```

I guess this is due to the string `addr` is pointing to getting freed before the `evalLater` callback is invoked. It's strange this is only happening on Windows though.

kentonv, line 1091: Simpler resolution:

```suggestion
    return evalLater([this,portHint,addr=kj::str(addr)]() {
      return SocketAddress::parse(lowLevel, addr, portHint, filter);
```

kentonv: Hmm, though maybe this would be more efficient?

```suggestion
    return evalNow([&]() {
      return SocketAddress::parse(lowLevel, addr, portHint, filter);
```

c++/src/kj/common.h
```diff
@@ -1314,7 +1314,7 @@ inline T* readMaybe(T* ptr) { return ptr; }
 
 #define KJ_IF_MAYBE(name, exp) if (auto name = ::kj::_::readMaybe(exp))
 
-#if __GNUC__
+#if __GNUC__ || defined(__clang__)
```
kentonv, line 1317: In KJ preprocessor code we usually rely on the fact that undefined symbols are treated as falsy.

```suggestion
#if __GNUC__ || __clang__
```

c++/src/kj/common.h
```diff
@@ -1347,6 +1347,9 @@ inline T* readMaybe(T* ptr) { return ptr; }
 // "statement expressions" extension. IIFEs don't do the trick here because a lambda cannot
 // return out of the parent scope. These macros should therefore only be used in projects that
 // target GCC or GCC-compatible compilers.
+//
+// `__GNUC__` is not defined when using LLVM's MSVC-compatible compiler driver `clang-cl` (even
```
kentonv, line 1351: I wonder if there are a bunch of other places where we are checking for `__GNUC__` but should also check for `__clang__`, or alternately places where we check for `_MSC_VER` for MSVC-specific work-arounds that Clang doesn't need.

c++/src/kj/compat/tls-test.c++
```diff
@@ -637,7 +637,10 @@ KJ_TEST("TLS full duplex") {
 
   auto writeUp = writeN(*client, "foo", 10000);
   auto readDown = readN(*client, "bar", 10000);
+#ifndef _WIN32
```
kentonv, line 640:
```suggestion
#if !_WIN32
```

c++/src/kj/compat/tls-test.c++
```diff
@@ -637,7 +637,10 @@ KJ_TEST("TLS full duplex") {
...
+  // TODO(soon): work out why this expectation fails even with the above fix
```
kentonv, line 641: Apparently it only fails in your particular configuration, since it passes on our Windows CI. Can we narrow the check to whatever's unusual in your case rather than weaken all the tests on Windows?

c++/src/kj/exception.c++
```diff
@@ -254,7 +254,7 @@ ArrayPtr<void* const> getStackTrace(ArrayPtr<void*> space, uint ignoreCount) {
 #endif
 }
 
-#if __GNUC__ && !_WIN32
+#if (__GNUC__ && !_WIN32) || __clang__
```
kentonv, line 257: Weak symbols are supported by Clang on Windows? But not by GCC (MinGW)? That seems surprising, I didn't think Windows had weak symbols, but if it does, you'd think  both compilers would support it.

mrbbot: Seems like it. I was able to link against the custom `workerd` symbolizer once I removed some of the unsupported syscalls from it.

c++/src/kj/async-io-win32.c++
mrbbot: Is the change in semantics from `evalLater` to `evalNow` ok here? If so, I've gone ahead and implemented your 2nd suggestion. 👍

c++/src/kj/common.h
mrbbot: Potentially, this was the only one that needed to be changed for the tests to pass. 🤷

c++/src/kj/async-io-win32.c++
kentonv: This seems like an improvement. I can't figure out why it was `evalLater` to begin with. I'm also surprised that this hasn't caused issues on Unix more often. In theory you could say it's the caller's fault for not attaching the string to the promise but it seems like I've commonly forgot to do that myself.

kentonv (approved):

PR #1642, opened 2023-03-19
Allow MembranePolicy to permit FD passthrough.

kentonv:


kentonv:
For context, we have a membrane (you know the one) that's used to just to track when a particular capability is no longer in use, in order to close the underlying connection. That membrane has no need to block FD passthrough.

PR #1645, opened 2023-03-21
Implements StartTLS in HttpConnectSettings and NetworkHttpClient. (Followup)

dom96:
Corrected version of https://github.com/capnproto/capnproto/pull/1635.

For reference, we want to ensure:

* No API breakage
* That kj-http doesn't depend on kj-tls.

This new PR should now ensure this.

### Test Plan

```
bazel test @capnp-cpp//src/kj/compat:tls-test --config=asan --test_timeout 5
```

### DIFF
--- a/c++/src/kj/compat/tls-test.c++
+++ b/c++/src/kj/compat/tls-test.c++
@@ -27,6 +27,8 @@
 
 #include "tls.h"
 
+#include "http.h"
+
 #include <openssl/opensslv.h>
 
 #include <stdlib.h>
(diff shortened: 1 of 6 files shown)


harrishancock (approved):

kentonv (commented):
Sorry, I thought I posted this earlier but it was still pending.

c++/src/kj/compat/tls-test.c++
```diff
@@ -1149,6 +1151,44 @@ KJ_TEST("TLS receiver does not stall on hung client") {
   KJ_EXPECT(!extraAcceptPromise.poll(test.io.waitScope));
 }
 
+#if !_WIN32 // TODO: Investigate and fix issue on Windows.
```
kentonv, line 1154: Can you elaborate? We are going to be supporting Windows soon so ignoring it here might be an issue.

kentonv:
Looks good except we really shouldn't disregard Windows failures without at least understanding why they are happening. We are porting workerd to Windows so this could actually matter.

dom96:
@kentonv I've got a task to follow up on the Windows aspect of this, not sure why it's failing and I'd have to switch workstations to investigate so I wanted to leave it for later.

kentonv:
OK, as long as you plan to proactively follow up within the next few weeks.

kentonv (approved):

PR #1649, opened 2023-03-24
Support kj::addRef(const T&)

jrsala-auguration:
Currently, for a given `T` derived from `kj::Refcounted`, passing a `const T&` to `kj::addRef` does not compile. GCC says:

```
capnproto/c++/src/kj/refcount.h:113:28: error: invalid conversion from ‘const kj::Refcounted*’ to ‘kj::Refcounted*’ [-fpermissive]
  113 |   Refcounted* refcounted = object;
      |                            ^~~~~~
      |                            |
      |                            const kj::Refcounted*
```

But I think it should be fine to `kj::addRef` a `const T` as long as you get a `Own<const T>` in return.

### DIFF
--- a/c++/src/kj/refcount-test.c++
+++ b/c++/src/kj/refcount-test.c++
@@ -58,8 +58,8 @@ TEST(Refcount, Basic) {
 }
 
 TEST(Refcount, Const) {
-  // Ensure `kj::addRef` supports argument of type `Own<const T>` and in that
-  // case returns `Own<const T>`
+  // Ensure `kj::addRef` supports argument of type `const T&` and in that case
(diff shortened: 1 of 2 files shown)


kentonv:
Sorry, but this is intentional. In KJ style, if an operation is marked `const` or applies to a `const` reference, it is expected that this means it is safe to execute that operation in multiple threads concurrently. `kj::addRef`, however, is not thread-safe, therefore should not be allowed to apply to a const object.

You might consider using `kj::AtomicRefcounted` and `kj::atomicAddRef()` instead. These are thread-safe, and `kj::atomicAddRef()` can in fact be applied to a const object.

PR #1653, opened 2023-03-31
Implements tlsStarter for http-over-capnp.

dom96:
Implements tlsStarter support for CONNECT via http-over-capnp.

### Test Plan

```
$ bazel test @capnp-cpp//src/capnp/compat:http-over-capnp-test --test_timeout 5
$ bazel test //src/capnp/compat:http-over-capnp-test --test_timeout 5
```

and upstream tests.

### DIFF
--- a/c++/src/capnp/compat/byte-stream.c++
+++ b/c++/src/capnp/compat/byte-stream.c++
@@ -72,9 +72,10 @@ public:
                 kj::AsyncOutputStream& stream,
                 capnp::ByteStream::SubstreamCallback::Client callback,
                 uint64_t limit,
+                kj::Maybe<kj::Own<kj::TlsStarterCallback>> tlsStarter,
                 kj::PromiseFulfillerPair<void> paf = kj::newPromiseAndFulfiller<void>())
       : factory(factory),
-        state(Streaming {parent, kj::mv(ownParent), stream, kj::mv(callback)}),
+        state(Streaming {parent, kj::mv(ownParent), stream, kj::mv(callback), kj::mv(tlsStarter)}),
         limit(limit),
         resolveFulfiller(kj::mv(paf.fulfiller)),
         resolvePromise(paf.promise.fork()) {}
@@ -229,6 +230,10 @@ public:
         KJ_FAIL_REQUIRE("can't call other methods while stream is borrowed");
       }
       KJ_CASE_ONEOF(streaming, Streaming) {
+        // Revoke the TLS starter when stream is ended. This will ensure any startTls calls
+        // cannot be falsely invoked after the stream is destroyed.
+        auto drop = kj::mv(streaming.tlsStarter);
+
         auto req = streaming.callback.endedRequest(MessageSize {4, 0});
         req.setByteCount(completed);
         auto result = req.send().ignoreResult();
@@ -238,6 +243,10 @@ public:
       }
     }
     KJ_UNREACHABLE;
+  }
+
+  kj::Promise<void> startTls(StartTlsContext context) override {
+    KJ_UNIMPLEMENTED("A substream does not support TLS initiation");
   }
 
   kj::Promise<void> getSubstream(GetSubstreamContext context) override {
@@ -262,7 +271,8 @@ public:
         context.releaseParams();
         auto results = context.getResults(MessageSize { 2, 1 });
         results.setSubstream(factory.streamSet.add(kj::heap<SubstreamImpl>(
-            factory, *this, thisCap(), streaming.stream, kj::mv(callback), kj::mv(limit))));
+            factory, *this, thisCap(), streaming.stream, kj::mv(callback), kj::mv(limit),
+            kj::mv(streaming.tlsStarter))));
         state = Borrowed { kj::mv(streaming) };
         return kj::READY_NOW;
       }
@@ -278,6 +288,7 @@ private:
     capnp::ByteStream::Client ownParent;
     kj::AsyncOutputStream& stream;
     capnp::ByteStream::SubstreamCallback::Client callback;
+    kj::Maybe<kj::Own<kj::TlsStarterCallback>> tlsStarter;
   };
   struct Borrowed {
     Streaming originalState;
@@ -320,6 +331,15 @@ public:
   CapnpToKjStreamAdapter(ByteStreamFactory& factory,
                          kj::Own<kj::AsyncOutputStream> inner)
       : factory(factory),
+        state(kj::heap<PathProber>(*this, kj::mv(inner))) {
+    state.get<kj::Own<PathProber>>()->startProbing();
+  }
+
+  CapnpToKjStreamAdapter(ByteStreamFactory& factory,
+                         kj::Own<kj::AsyncOutputStream> inner,
+                         kj::Maybe<kj::Own<kj::TlsStarterCallback>> starter)
+      : factory(factory),
+        tlsStarter(kj::mv(starter)),
         state(kj::heap<PathProber>(*this, kj::mv(inner))) {
     state.get<kj::Own<PathProber>>()->startProbing();
   }
@@ -607,6 +627,10 @@ protected:
   }
 
   kj::Promise<void> end(EndContext context) override {
+    // Revoke the TLS starter when stream is ended. This will ensure any startTls calls
+    // cannot be falsely invoked after the stream is destroyed.
+    auto drop = kj::mv(tlsStarter);
+
     KJ_SWITCH_ONEOF(state) {
       KJ_CASE_ONEOF(prober, kj::Own<PathProber>) {
         return prober->whenReady().then([this, context]() mutable {
@@ -637,6 +661,30 @@ protected:
     KJ_UNREACHABLE;
   }
 
+  kj::Promise<void> startTls(StartTlsContext context) override {
+    auto params = context.getParams();
+    KJ_IF_MAYBE(s, tlsStarter) {
+      KJ_SWITCH_ONEOF(state) {
+        KJ_CASE_ONEOF(prober, kj::Own<PathProber>) {
+          return KJ_ASSERT_NONNULL(*s->get())(params.getExpectedServerHostname());
+        }
+        KJ_CASE_ONEOF(kjStream, kj::Own<kj::AsyncOutputStream>) {
+          return KJ_ASSERT_NONNULL(*s->get())(params.getExpectedServerHostname());
+        }
+        KJ_CASE_ONEOF(capnpStream, capnp::ByteStream::Client) {
+          return KJ_ASSERT_NONNULL(*s->get())(params.getExpectedServerHostname());
+        }
+        KJ_CASE_ONEOF(e, Ended) {
+          KJ_FAIL_REQUIRE("cannot call startTls on a bytestream that was ended");
+        }
+        KJ_CASE_ONEOF(b, Borrowed) {
+          KJ_FAIL_REQUIRE("can't call startTls while stream is borrowed");
+        }
+      }
+    }
+    KJ_UNREACHABLE;
+  }
+
   kj::Promise<void> getSubstream(GetSubstreamContext context) override {
     KJ_SWITCH_ONEOF(state) {
       KJ_CASE_ONEOF(prober, kj::Own<PathProber>) {
@@ -653,7 +701,8 @@ protected:
 
         auto results = context.initResults(MessageSize {2, 1});
         results.setSubstream(factory.streamSet.add(kj::heap<SubstreamImpl>(
-            factory, *this, thisCap(), *kjStream, kj::mv(callback), kj::mv(limit))));
+            factory, *this, thisCap(), *kjStream, kj::mv(callback), kj::mv(limit),
+            kj::mv(tlsStarter))));
         state = Borrowed { kj::mv(kjStream) };
         return kj::READY_NOW;
       }
@@ -678,6 +727,7 @@ protected:
 
 private:
   ByteStreamFactory& factory;
+  kj::Maybe<kj::Own<kj::TlsStarterCallback>> tlsStarter;
 
   struct Borrowed { kj::Own<kj::AsyncOutputStream> stream; };
   struct Ended {};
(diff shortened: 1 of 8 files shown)


c++/src/kj/compat/http.c++
```diff
@@ -5716,14 +5716,111 @@ public:
...
+              } else {
+                return unencryptedStream->getFd();
+              }
+            }
+
+            void startTls(
+                kj::SecureNetworkWrapper* wrapper, kj::StringPtr expectedServerHostname) {
+              kj::Promise<kj::Own<kj::AsyncIoStream>> secureStream =
+                  wrapper->wrapClient(kj::mv(unencryptedStream), expectedServerHostname);
+              encryptedStream = kj::newPromisedStream(kj::mv(secureStream));
```
kentonv, line 5809: If you assigned this into `unencryptedStream` and never touched `encryptedStream`, the code would still work exactly the same.

Maybe you should just have one member `kj::Own<kj::AsyncIoStream> inner` which you overwrite here? Then all the other methods can be simplified to just forward to `inner`, no need to check the `Maybe`.

c++/src/kj/compat/http.c++
```diff
@@ -5716,14 +5716,111 @@ public:
...
+
+            kj::Maybe<int> getFd() const override {
+              KJ_IF_MAYBE(s, encryptedStream) {
+                return s->get()->getFd();
+              } else {
+                return unencryptedStream->getFd();
+              }
+            }
+
+            void startTls(
```
kentonv, line 5805: Can we at least track whether there is an outstanding read, and throw from this method if it is called while a read is outstanding? Otherwise the outstanding read will end up reading the TLS handshake bytes, breaking everything.

You could have a method like:

```
auto trackRead() {
  KJ_ASSERT(!currentlyReading, "only one read is allowed at a time");
  currentlyReading = true;
  return kj::defer([this]() { currentlyReading = false; });
}
```

Then each read operation would do something like:

```
return inner->read(whatever).attach(trackRead());
```

Note that both `tryRead()` and `pumpTo()` need to be counted here.

Also note `AsyncInputStream` already has a rule that there can only be one read at a time, hence why a boolean is sufficient rather than a counter.

c++/src/kj/compat/http.c++
```diff
@@ -5716,14 +5716,111 @@ public:
...
+                  wrapper->wrapClient(kj::mv(unencryptedStream), expectedServerHostname);
+              encryptedStream = kj::newPromisedStream(kj::mv(secureStream));
+            }
+
+          private:
+            kj::Own<kj::AsyncIoStream> unencryptedStream;
+            kj::Maybe<kj::Own<kj::AsyncIoStream>> encryptedStream;
+          };
+
+          auto refConnection = kj::refcountedWrapper(
```
kentonv, line 5817: I don't think you really need reference counting here. It should be understood that the starter function cannot be called if the stream has already been discarded. So the function lambda only needs to capture a reference to the object, it doesn't need a strong refcount.

c++/src/kj/compat/http.c++
```diff
@@ -5716,14 +5716,111 @@ public:
     if (!connectSettings.useTls) {
```
kentonv, line 5716: I notice this is still wrapped in `#if KJ_HAS_OPENSSL`. We broke that dependency with abstraction, so you can remove that `#if`.

c++/src/capnp/compat/http-over-capnp.c++
```diff
@@ -900,15 +938,19 @@ public:
...
-    return inner->connect(host, headers, *pipe.ends[0], *response, settings).attach(
+    auto result = inner->connect(host, headers, *pipe.ends[0], *response, settings).attach(
         kj::mv(host), kj::mv(headers), kj::mv(response), kj::mv(pipe))
         .exclusiveJoin(kj::mv(pumpTask));
+
+    auto tlsStarterImpl = kj::heap<TlsStarterImpl>(kj::mv(tlsStarter));
+    pb.setTlsStarter(kj::mv(tlsStarterImpl));
+
+    context.setPipeline(pb.build());
+    context.initResults(capnp::MessageSize { 4, 1 }).setUp(kj::mv(up));
```
kentonv, line 952: It's important that the final results match the pipeline set with `setPipeline`, so please initialize `tlsStarter` in the final results here.

Don't forget to change the cap count in `MessageSize` from 1 to 2.

c++/src/capnp/compat/http-over-capnp.c++
```diff
@@ -504,6 +527,19 @@ public:
...
 
+    // We need to create a tlsStarter callback which sends a startTls request to the capnp server.
+    KJ_IF_MAYBE(tlsStarter, settings.tlsStarter) {
+      kj::Function<void(kj::StringPtr)> cb =
+          [remoteTlsStarter = pipeline.getTlsStarter()](kj::StringPtr expectedServerHostname) mutable -> void {
+        auto rpcRequest = remoteTlsStarter.startTlsRequest();
+        auto downPipe = kj::newOneWayPipe();
+        rpcRequest.setExpectedServerHostname(expectedServerHostname);
+
+        RemotePromise<capnp::TlsStarter::StartTlsResults> pipeline = rpcRequest.send();
```
kentonv, line 538: Er, this seems to be immediately dropping the promise returned by `send()`, which immediately cancels the call.

Unfortunately we might need to change the TLS starter callback to return `kj::Promise<void>` so that we rat return a promise here. And the caller will have to wait for that promise before doing any writes.

c++/src/capnp/compat/http-over-capnp.c++
```diff
@@ -373,6 +373,29 @@ private:
...
 };
 
+class HttpOverCapnpFactory::TlsStarterImpl final
+    : public TlsStarter::Server {
+public:
+  TlsStarterImpl(kj::TlsStarterCallback tlsStarter)
+      : tlsStarter(kj::mv(tlsStarter)) {}
+
+  kj::Promise<void> startTls(StartTlsContext context) override {
+    KJ_REQUIRE(!tlsStarted, "already called startTls()");
```
kentonv, line 383: We may also need to prevent this from being called if the stream has already been destroyed. I think this could be accomplished by having the `EofDetector` hold on to a reference to the callback and set a flag on it when done that revokes it.

c++/src/kj/compat/http.c++
```diff
@@ -5716,14 +5716,111 @@ public:
...
       KJ_IF_MAYBE(wrapper, settings.tlsContext) {
         KJ_IF_MAYBE(tlsStarter, connectSettings.tlsStarter) {
-          auto refConnection = kj::refcountedWrapper(kj::mv(connection));
-          connection = refConnection->addWrappedRef();
-          kj::Own<kj::AsyncIoStream> ref1 = refConnection->addWrappedRef();
-          Function<kj::Own<kj::AsyncIoStream>(kj::StringPtr)> cb =
+          class TransitionaryAsyncIoStream final: public kj::AsyncIoStream {
```
kentonv, line 5719: Nitpick: This class feels a bit big to be declared nested inside a function like this.

dom96:
Just a rebase first

c++/src/capnp/compat/http-over-capnp.c++
dom96: hm, I need to call `setTlsStarter` on both the pipeline and the final results right? Do I need a refcounted `TlsStarterImpl` or is there a better way? For now I resolved this by using a refcount.

dom96:
Resolved most comments. Test currently fails so need to look closer into that, but will get back to it tomorrow.

dom96:
Ready for review!

c++/src/capnp/compat/byte-stream.c++
```diff
@@ -240,6 +242,29 @@ public:
...
+          return req.send();
+        }
+        KJ_CASE_ONEOF(e, Ended) {
+          KJ_FAIL_REQUIRE("cannot call startTls on a bytestream that was ended");
+        }
+        KJ_CASE_ONEOF(b, Borrowed) {
+          KJ_FAIL_REQUIRE("can't call startTls while stream is borrowed");
+        }
+        KJ_CASE_ONEOF(streaming, Streaming) {
+          return KJ_ASSERT_NONNULL(*s)(params.getExpectedServerHostname());
```
kentonv, line 261: More of a REQUIRE, and should have an error message.

```suggestion
          auto& callback = KJ_REQUIRE_NONNULL(*s,
              "this ByteStream does not support startTls")
          return callback(params.getExpectedServerHostname());
```

c++/src/capnp/compat/http-over-capnp.c++
```diff
@@ -845,17 +862,28 @@ public:
...
     auto pipe = kj::newTwoWayPipe();
 
     class EofDetector final: public kj::AsyncOutputStream {
     public:
-      EofDetector(kj::Own<kj::AsyncIoStream> inner) : inner(kj::mv(inner)) {}
+      EofDetector(kj::Own<kj::AsyncIoStream> inner, kj::TlsStarterCallback& starter)
+          : inner(kj::mv(inner)), starter(starter) {}
       ~EofDetector() {
         inner->shutdownWrite();
+        // Revoke the TLS starter when stream is destroyed. This will ensure any startTls calls
```
kentonv, line 877: Could we instead pass ownership of the `TlsStarterCallback` directly to the `ByteStream` and let it take care of dropping it in end()?

Then we don't need to allocate `TlsStarterCallback` as a heap object and pass all these references around.

c++/src/capnp/compat/byte-stream.c++
```diff
@@ -1094,6 +1156,17 @@ capnp::ByteStream::Client ByteStreamFactory::kjToCapnp(kj::Own<kj::AsyncOutputSt
   return streamSet.add(kj::heap<CapnpToKjStreamAdapter>(*this, kj::mv(kjStream)));
 }
 
+capnp::ByteStream::Client ByteStreamFactory::kjToCapnp(
+    kj::Own<kj::AsyncOutputStream> kjStream, kj::Maybe<kj::TlsStarterCallback&> tlsStarter) {
+  KJ_IF_MAYBE(s, tlsStarter) {
```
kentonv, line 1161: Nitpick: Why not have `CapnpToKjStreamAdapter` have just one constructor that takes `kj::Maybe<kj::TlsStarterCallback&> tlsStarter` as the third argument, so you don't have to branch here?

c++/src/capnp/compat/http-over-capnp.c++
```diff
@@ -475,7 +475,9 @@ public:
     auto rpcRequest = inner.connectRequest();
     auto downPipe = kj::newOneWayPipe();
     rpcRequest.setHost(host);
-    rpcRequest.setDown(factory.streamFactory.kjToCapnp(kj::mv(downPipe.out)));
+    capnp::ByteStream::Client downByteStream = factory.streamFactory.kjToCapnp(
+        kj::mv(downPipe.out));
+    rpcRequest.setDown(downByteStream);
```
kentonv, line 480: Nitpick: This diff looks vestigial? `downByteStream` is not used below. You should `kj::mv(downByteStream)` if you aren't using it later; otherwise this is adding refcounts -- or revert the diff.

c++/src/kj/compat/http.c++
```diff
@@ -5546,6 +5546,71 @@ private:
...
+  void startTls(
+      kj::SecureNetworkWrapper* wrapper, kj::StringPtr expectedServerHostname) {
+    kj::Promise<kj::Own<kj::AsyncIoStream>> secureStream =
+        wrapper->wrapClient(kj::mv(inner), expectedServerHostname);
+    inner = kj::newPromisedStream(kj::mv(secureStream));
+  }
+
+private:
+  kj::Own<kj::AsyncIoStream> inner;
+  bool currentlyReading;
```
kentonv, line 5611: Did you mean to require this is false in `startTls()`? Otherwise it looks like it's not doing much.

c++/src/capnp/compat/byte-stream.c++
```diff
@@ -240,6 +245,30 @@ public:
     KJ_UNREACHABLE;
   }
 
+  kj::Promise<void> startTls(StartTlsContext context) override {
```
harrishancock, line 248: A substream allowing TLS initiation strikes me as a little weird, because from reading the code, it seems that substreams' primary use case is to optimize write operations to the parent stream. Presumably those write operations will always have completed before TLS initiation.

Is this code path exercised anywhere, do you know? If not, and if it's hard to test, and if we can't think of a case where we'd need it, maybe it should unconditionally throw?

dom96: Yeah, I think we can `throw` here.

c++/src/kj/compat/http.h
```diff
@@ -674,12 +674,158 @@ struct HttpConnectSettings {
...
-  // connection at any arbitrary point. The function returns an AsyncIoStream that is a secure
-  // TLS stream. This mechanism is required for certain protocols, more info can be found on
+  // The callback function itself can be called to initiate a TLS handshake on the connection in
+  // between write() operations. It is not allowed to initiate a TLS handshake while a write
+  // operation or a pump operation to the connection exists. Read operations are not subject to
+  // the same constraint, however: implementations are required to be able to handle TLS
+  // initiation while a read operation or pump operation from the connection exists. Once the
+  // promise returned from the callback is fulfilled, the connection has become a secure stream,
+  // and write operations are once again permitted. The StringPtr parameter to the callback,
+  // expectedServerHostname (TODO: must remain valid for the lifetime of the promise? may be
```
harrishancock, line 684: By the TODO, I meant to figure out which one is the correct statement, and use that. :)

c++/src/kj/compat/http.h
```diff
@@ -674,12 +674,158 @@ struct HttpConnectSettings {
...
+  // dropped after the function synchronously returns?).
+
+
+  // This mechanism is required for certain protocols, more info can be found on
   // https://en.wikipedia.org/wiki/Opportunistic_TLS.
 };
 
+class PausableAsyncIoStream final: public kj::AsyncIoStream {
```
harrishancock, line 692: Needs a doc comment; probably should be mentioned in the `tlsStarter` doc comment, too.

c++/src/kj/compat/http.c++
```diff
@@ -5546,6 +5558,70 @@ private:
...
+  kj::Maybe<int> getFd() const override {
+    return inner->getFd();
+  }
+
+  void startTls(
+      kj::SecureNetworkWrapper* wrapper, kj::StringPtr expectedServerHostname) {
+    // Pause any potential pending reads.
+    inner->pause();
+
+    KJ_REQUIRE(!inner->getCurrentlyReading(), "Cannot startTls while a read is pending.");
```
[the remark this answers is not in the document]
dom96: I suppose it can't hurt, I left it as-is because the initial request here (https://github.com/capnproto/capnproto/pull/1653#discussion_r1154812880) was to assert on the read and I wasn't sure if checking for writes would break something.

harrishancock (approved):

c++/src/kj/compat/http.h
```diff
@@ -674,12 +675,179 @@ struct HttpConnectSettings {
...
+  // expectedServerHostname may be dropped after the function synchronously returns.
+  //
+  // The PausableReadAsyncIoStream class defined below can be used to ensure that read operations
+  // are not pending when the tlsStarter is invoked.
+  //
+  // This mechanism is required for certain protocols, more info can be found on
   // https://en.wikipedia.org/wiki/Opportunistic_TLS.
 };
 
+class PausableReadAsyncIoStream final: public kj::AsyncIoStream {
```
kentonv, line 694: Does this need to be in the header? It seems like a private implementation detail. If it needs to be in the header it should be in `namespace _`.

PR #1655, opened 2023-04-06
Fix return type of hash operator (`bool` -> `size_t`).

legrosbuffle:
Using `bool` essentially turns the hash map into a bad linked list: lookup is `O(n)`.

### DIFF
--- a/c++/src/capnp/schema-parser.c++
+++ b/c++/src/capnp/schema-parser.c++
@@ -148,7 +148,7 @@ private:
 namespace {
 
 struct SchemaFileHash {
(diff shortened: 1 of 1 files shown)


kentonv:
Ugghhhh... :facepalm: Just like #901, except in this case we're using `std::unordered_map` rather than `kj::HashMap` so the detection I added in doesn't apply.

On the bright side, this particular hash map's performance probably never mattered that much since no one is loading millions of schema files, but still, ugh.

Thanks for catching this!

kentonv (approved):
(Will merge when CI passes.)

PR #1659, opened 2023-04-15
Resolve `StructSchema::Field` on creation to cache `getType()`

bobrik:
Field objects can be reused between messages and prior to this change the type had to be resolved each time from scratch, which quickly adds up.

We see as much as 20% of CPU time spent in `getType()` in our code that serializes lua tables to capnproto, which can be removed.

I had to move `Type` definition above in the file to allow it to be included as a field in `Field`, that is why the diff is so big.

Here's the test I used:

```capnp
@0xce6ad89533407d77;

struct Message {
    text @0 :Text;
}
```

```c++

int main()
{
    capnp::MallocMessageBuilder message;
    auto schema = capnp::Schema::from<Message>();
    auto root = message.initRoot<capnp::DynamicStruct>(schema);

    auto field = capnp::Type::from<Message>().asStruct().getFields()[0];

    auto begin = std::chrono::steady_clock::now();

    for (int i = 0; i < TIMES; i++)
    {
        root.set(field, "lol");
    }

    auto end = std::chrono::steady_clock::now();

    auto duration_us = std::chrono::duration_cast<std::chrono::milliseconds>(end - begin).count();

    std::cout << "set field " << TIMES << " times in " << duration_us << "ms" << std::endl;

    return 0;
}
```

* Before: 513ms, 26.4% of on-CPU time spent in `getType()`.
* After: 371ms, 0.0% of on-CPU time spent in `getType()`.

If we change the code to set a number instead of a string:

* Before: 257ms, 44.5% of on-CPU time spent in `getType()`.
* After: 151ms, 0.0% of on-CPU time spent in `getType()`.

It's possible that there's a cleaner way to cache `Type` in a more appropriate place, but I'm not qualified to make that change. This one does the job.


bobrik:
I think it's more than just `getType()`. Here's a flemagraph with `getStruct()` highlighted:

image

It's called from both `getType()` (left, which is addressed here) and from `getDefaultValue()` (right).

Here's the same flamegraph with the change from this PR applied:

image

In production we didn't see this `getStruct()` call as coming from `getDefaultValue()` due to it being inlined and the lack of debug info, so I didn't think it was easily optimizable:

```
(gdb) bt
#0  0x00007fb16d52de90 in capnp::_::PointerReader::getStruct(capnp::word const*) const@plt () from ..
#1  0x00007fb16d56a3ac in capnp::DynamicStruct::Builder::set(capnp::StructSchema::Field, capnp::DynamicValue::Reader const&) () from ...
```

I don't see anything else calling into `getStruct()` from `DynamicStruct::Builder::get()` directly, so that must be it. I'll look into enabling more debug info to confirm this for production, but so far it looks like a convincing theory.

I suppose we can also resolve the default value when we create `Field`, same as we do with the `Type` here. This would make `Field` usage more expensive due to upfront cost of construction, but I don't know if anyone is making `Field` in a way where it's not used in `set()` for it to matter.

bobrik:
I added a commit to cache the default value as well. Here's a flamegraph for a `uint64` field set after both changes:

image

c++/src/capnp/schema.h
```diff
@@ -349,9 +462,14 @@ class StructSchema::Field {
...
   uint index;
   schema::Field::Reader proto;
+  Type type;
+
+  static Type resolveType(StructSchema parent, uint index, schema::Field::Reader proto);
 
   inline Field(StructSchema parent, uint index, schema::Field::Reader proto)
-      : parent(parent), index(index), proto(proto) {}
+      : parent(parent), index(index), proto(proto) {
+    type = resolveType(parent, index, proto);
```
kentonv, line 478: Unfortunately I don't think pushing resolution to here solves very much. Instances of `Field` are themselves created on-demand when accessing the `StructSchema` interface, e.g. when you iterate over a `FieldList` or call `getFieldByName()`. You can see there that the `Field` is returned by value, implying that a new `Field` is created each time.

Essentially all of the schema API types are actually created on-demand. The long-lived data structures underlying them are defined in `raw-schema.h`. In order to make a meaningful difference here I think we need to change the raw schemas to maintain appropriate lookup tables.

kentonv: > Before: 257ms, 44.5% of on-CPU time spent in getType().
> After: 151ms, 0.0% of on-CPU time spent in getType().

The second line here should really measure time spent in `resolveType()`, since all the compute from `getType()` was moved there. But you're reporting that the application did actually get faster overall?

I guess these numbers imply that your application is obtaining `Field` objects and then using them multiple times, such that the memoized type is actually useful?

So maybe the change is useful as-is. But, I'm sort of concerned that there may be applications which iterate fields but don't actually call `getType()` on them. Those apps will now be slower.

bobrik: > Unfortunately I don't think pushing resolution to here solves very much. Instances of `Field` are themselves created on-demand when accessing the `StructSchema` interface, e.g. when you iterate over a `FieldList` or call `getFieldByName()`. You can see there that the `Field` is returned by value, implying that a new `Field` is created each time.

The way it works for us internally to convert lua objects into capnproto:

1. We iterate the schema recursively and remember what each name corresponds to which `Field`. The `Field` objects are cached into a hashmap to avoid calling into `getFieldByName()` for every lua object we want to serialize.
2. Whenever an object is serialized, we find a corresponding `Field` _in our cache_ and reuse that. The `Field` objects are not created here, they are retrieved from a cache.

> Essentially all of the schema API types are actually created on-demand.

To reiterate: for us this is not the case. We create `Field` objects once when we discover the schema.

> The long-lived data structures underlying them are defined in `raw-schema.h`. In order to make a meaningful difference here I think we need to change the raw schemas to maintain appropriate lookup tables.

I _think_ this would make sense for a general case, as the speedup would be available for both the way we use it and the way you envision it to be used. I'm not sure if I'm capable of making that change though, but I'll see what I can do.

> The second line here should really measure time spent in `resolveType()`, since all the compute from `getType()` was moved there. But you're reporting that the application did actually get faster overall?

Hopefully this is clear by now from the previous paragraphs, but in our usage we create `Field` once and reuse it.

I don't think it would matter if we start measuring time before this line or after:

```c++
auto field = capnp::Type::from<Message>().asStruct().getFields()[0];
```

After all, we create `Field` once and run `set()` a million times, so the time is dominated by `set()`, which is confirmed by the width of the `set()` slab on the flamegraphs. Creating a `Field` and running `resolveType()` is but a blip in this.

The whole speedup is from not running `resolveType()` / original `getType()` a million times, but running it once.

> I guess these numbers imply that your application is obtaining Field objects and then using them multiple times, such that the memoized type is actually useful?

That is correct.

> So maybe the change is useful as-is. But, I'm sort of concerned that there may be applications which iterate fields but don't actually call `getType()` on them. Those apps will now be slower.

That is a reasonable concern to have. I'm not sure if it's a common or even a reasonable usage though.

Having all the caching in `raw-schema.h` would work for either case, but as I mentioned above it's harder for me to go that deep. I rarely touch C++ to begin with.

kentonv:
Ugh, I wrote up #1697 assuming the binary search was the slow part but it wasn't. It's `proto.getSlot().getType()` that is slow -- that is, all the Cap'n Proto pointer validation.

I guess initializing this in the constructor seems like the best bet for now. As I said there's a risk some applications become slower, but I suppose most apps probably do inspect the type of each field.

c++/src/capnp/schema.h
```diff
@@ -349,9 +468,18 @@ class StructSchema::Field {
   StructSchema parent;
   uint index;
   schema::Field::Reader proto;
+  Type type;
+  schema::Value::Reader defaultValue;
+
+  static Type resolveType(StructSchema parent, uint index, schema::Field::Reader proto);
```
kentonv, line 474: This doesn't need to be static. If it weren't static then you wouldn't need to pass copies of all the members as params.

c++/src/capnp/schema.h
```diff
@@ -341,6 +454,12 @@ class StructSchema::Field {
   //
   // If the above does not make sense, you probably don't need this method.
 
+  schema::Value::Reader getDefaultValue() const {
```
kentonv, line 457: Can we name this `getDefaultValueProto()`, since it returns the capnp encoding of the default value and not a `DynamicValue`?

c++/src/capnp/schema.h
```diff
@@ -341,6 +454,12 @@ class StructSchema::Field {
...
+    KJ_IREQUIRE((getProto().which() == schema::Field::SLOT),
+              "Must check which() before get()ing a union member.");
```
kentonv, line 459: This is a weird error message here given that we're not reading a union. Maybe: `getDefaultValueProto() cannot be called on groups`

c++/src/capnp/schema.h
bobrik: Updated

c++/src/capnp/schema.h
bobrik: Updated

c++/src/capnp/schema.h
bobrik: Updated

PR #1662, opened 2023-04-19
add clang-16 build

mikea:
We're having upstream compilation problems with clang-16, adding it here so that I can write unit-tests/add fixes.

### DIFF
--- a/.github/workflows/quick-test.yml
+++ b/.github/workflows/quick-test.yml
@@ -184,7 +184,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        clang_version: [10, 14, 16]
+        clang_version: [16]
     steps:
       - uses: actions/checkout@v3
       - uses: bazelbuild/setup-bazelisk@v2
@@ -192,6 +192,7 @@ jobs:
         run: |
             export DEBIAN_FRONTEND=noninteractive
             sudo apt-get install -y build-essential git
+      # todo: replace with apt-get when clang-16 is part of ubuntu lts
       - name: install clang
         uses: egor-tensin/setup-clang@v1
         with:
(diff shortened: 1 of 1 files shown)


kentonv:
We can replace 14 with 16. The idea is to test the oldest and newest versions that we want to support, we assume everything in between will work...

kentonv:
Oh I see you are actually adding two new variants of the bazel build. Do we really need to cover so many different Clang versions?

.github/workflows/quick-test.yml
```diff
@@ -179,19 +179,23 @@ jobs:
...
     steps:
       - uses: actions/checkout@v3
       - uses: bazelbuild/setup-bazelisk@v2
       - name: install dependencies
         run: |
             export DEBIAN_FRONTEND=noninteractive
-            sudo apt-get install -y build-essential git ${{ matrix.compiler }}
+            sudo apt-get install -y build-essential git
+      - name: install clang
+        uses: egor-tensin/setup-clang@v1
```
kentonv, line 196: I'm not super excited about depending on random third-party repos, especially for something we only need temporarily (until a clang-16 package lands in Ubuntu, presumably).

kentonv: Actually I see this build is using Ubuntu 20.04, which is pretty old. Did you try 22.04? Maybe it has a clang-16 package already?

kentonv: OK I see... there is a clang-16 package but only for Ubuntu 23.04 (not LTS) so I guess indeed we'd have to wait another year.

I guess our actions don't have any secrets so there's no security concern here.

Given that we already cover clang 6 and 14 in the other builds, can we have the bazel build focus on 16? I worry we have too many builds...

mikea: done. Bazel is clang-16 only.

PR #1664, opened 2023-04-20
Fix redundant constexpr warning in generated code

tuxu:
Clang 15+ in C++17 mode (or higher) with `-Wdeprecated` issues a warning in
the generated code ([godbolt](https://godbolt.org/z/M5bvn5Y66)):

```
warning: out-of-line definition of constexpr static data member is
redundant in C++17 and is deprecated [-Wdeprecated]
```

It is rather annoying for downstream users of Capnp, since there's no
way to disable the specific warning. Instead, fix it here for code
compiled for C++17 and newer.

See [cppreference]:

> A constexpr static data member is implicitly inline and does not need
> to be redeclared at namespace scope. This redeclaration without an
> initializer (formerly required) is still permitted, but is deprecated.
> (since C++17)

[cppreference]: https://en.cppreference.com/w/cpp/language/static

--- 

Some more thoughts:
- This warning exists also with GCC 7+ ([godbolt](https://godbolt.org/z/eY9aT4sbc)) and MSVC 19.29 ([godbolt](https://godbolt.org/z/Yqz6rsoxz)).
- Is the special handling of MSVC still needed today? 
- All those redundant `#ifdef`s in the generated implementation file could be reduced to a minimum, but this would have meant a more invasive change.

### DIFF
--- a/c++/src/capnp/generated-header-support.h
+++ b/c++/src/capnp/generated-header-support.h
@@ -334,6 +334,13 @@ inline constexpr uint sizeInWords() {
 #define CAPNP_AUTO_IF_MSVC(...) __VA_ARGS__
 #endif
 
+// TODO(msvc): MSVC does not even expect constexprs to have definitions below C++17.
+#if (__cplusplus < 201703L) && !(defined(_MSC_VER) && !defined(__clang__))
+#define CAPNP_NEED_REDUNDANT_CONSTEXPR_DECL 1
+#else
+#define CAPNP_NEED_REDUNDANT_CONSTEXPR_DECL 0
+#endif
+
 #if CAPNP_LITE
 
 #define CAPNP_DECLARE_SCHEMA(id) \
@@ -349,12 +356,11 @@ inline constexpr uint sizeInWords() {
       static inline ::capnp::word const* encodedSchema() { return bp_##id; } \
     }
 
-#if _MSC_VER && !defined(__clang__)
-// TODO(msvc): MSVC doesn't expect constexprs to have definitions.
-#define CAPNP_DEFINE_ENUM(type, id)
-#else
+#if CAPNP_NEED_REDUNDANT_CONSTEXPR_DECL
 #define CAPNP_DEFINE_ENUM(type, id) \
     constexpr uint64_t EnumInfo<type>::typeId
+#else
+#define CAPNP_DEFINE_ENUM(type, id)
 #endif
 
 #define CAPNP_DECLARE_STRUCT_HEADER(id, dataWordSize_, pointerCount_) \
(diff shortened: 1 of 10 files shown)


c++/src/capnp/generated-header-support.h
```diff
@@ -349,12 +355,11 @@ inline constexpr uint sizeInWords() {
       static inline ::capnp::word const* encodedSchema() { return bp_##id; } \
     }
 
-#if _MSC_VER && !defined(__clang__)
-// TODO(msvc): MSVC doesn't expect constexprs to have definitions.
-#define CAPNP_DEFINE_ENUM(type, id)
-#else
+#if CAPNP_NEED_REDUNDANT_CONSTEXPR_DECL && (!defined(_MSC_VER) || defined(__clang__))
```
kentonv, line 358: I guess if we're keeping the `(!defined(_MSC_VER) || defined(__clang__))` condition then the comment should stay, too, to explain it?

But alternatively, maybe CAPNP_NEED_REDUNDANT_CONSTEXPR_DECL should always be 0 for MSVC? That's sort of what the comment seems to imply.

For that matter, are there actually any compilers that require constexprs to be defined? If removing these definitions entirely passes all tests then I wouldn't see a need to keep them.

If there are really some compiler versions that require the definitions in C++14 mode, I'd also be up for just switching to officially require C++17 going forward. We were probably going to do that anyway at some point.

tuxu: > I guess if we're keeping the (!defined(_MSC_VER) || defined(__clang__)) condition then the comment should stay, too, to explain it?
> But alternatively, maybe CAPNP_NEED_REDUNDANT_CONSTEXPR_DECL should always be 0 for MSVC? That's sort of what the comment seems to imply.

Right, at least the comment should stay. However, making `CAPNP_NEED_REDUNDANT_CONSTEXPR_DECL` 0 for MSVC would also remove all the other definitions, which MSVC apparently didn't mind?

> For that matter, are there actually any compilers that require constexprs to be defined? If removing these definitions entirely passes all tests then I wouldn't see a need to keep them.

C++11 requires the out-of-line definition only [if the members are ODR-used](https://stackoverflow.com/a/69456975). In a few experiments I did, the compilers are normally smart enough to optimize out any actual access of a variable. However, if you disable optimizations (`-O0`), the variable access stays and you get a linker error when compiling in C++11-mode ([godbolt](https://godbolt.org/z/qzzaKT7Kd)): 
```
main.cpp:(.text+0x22): undefined reference to `Foo::bar'
```

> If there are really some compiler versions that require the definitions in C++14 mode, I'd also be up for just switching to officially require C++17 going forward. We were probably going to do that anyway at some point.

If that's an option, I think it would be the best way going forward. It would reduce the complexity and size of the generated code quite a bit.

kentonv: To clarify, we already require C++14, not C++11, but playing with your godbolt link I see the same issue is still true in C++14. It only goes away with C++17.

Let's try making `CAPNP_NEED_REDUNDANT_CONSTEXPR_DECL` always 0 for MSVC and see if it gets through CI. If so, I'm happy to ship it.

We'll probably update our requirement to C++17 soon at which point we can delete all this code, but I won't make you do that now. We may end up wanting to cherry-pick this onto the last release branch, and we wouldn't want to update the C++ version requirements in a release branch.

tuxu: I force-pushed an update that sets `CAPNP_NEED_REDUNDANT_CONSTEXPR_DECL` to 0 unconditionally for MSVC. Hope it will get through CI! 🙂 

I also sort-of have a branch ready that deals with the C++17 upgrade and removes all out-of-line constexpr definitions.  It shouldn't be too much work to get it into a polished state, but let's get this one here merged first.

kentonv:
Thanks!

Issue #1667, opened 2023-04-24
Unable to synchronously (immediately) revoke capabilities


Zentren:
See this discussion for more detail: https://groups.google.com/g/capnproto/c/wxvQiR9D4zY

The method that `capnp::MembranePolicy` uses to revoke capabilities and requests wrapped by membranes created from it is unable to ensure that outstanding requests are immediately cancelled and not executed post-revocation. This becomes an issue when resources referenced by requests are made unavailable at revocation time, resulting in possible segfaults.

`capnp::MembranePolicy::onRevoked() -> kj::Promise<void>` paired with `Promise::exclusiveJoin` is currently used to "cancel" outstanding requests, but if the request was initiated before the promise returned by `onRevoked` in the event queue, it will still complete. We propose adding `capnp::MembranePolicy::getCanceler() -> kj::Maybe<kj::Canceler&>` which provides a synchronous API to cancel/revoke wrapped capabilities and requests. The method will return `nullptr` by default (opt-in, backward compatibility), otherwise a `kj::Canceler&` which will wrap requests and upon cancellation cancel outstanding requests and revoke capabilities. If the policy is currently revoked, calls to `getCanceler` should throw an exception, which will substitute for the response.

`kj::Canceler` needs to be updated to include some kind of observable, e.g. `onCanceled(callback) -> kj::Subscription` which will notify membranes about cancellation so that they can reject any further calls.

kentonv:
I believe #1700 may provide a simpler solution to your problem. What do you think?

Zentren:
Hi Kenton, as per reply on my own PR, I think this looks like a good solution to the class of problems we (and you it seems) have encountered :smile: Thank you, I will do some testing and thereafter we can probably close this issue as the revocable server solves this problem.

Zentren:
Presumably _MembranePolicy_ will live on as I see it dealing with _permission_ while _RevocableServer_ would deal with _access_, i.e. "you may not" vs "you cannot"?

kentonv:
MembranePolicy has a huge number of uses beyond simple revocation, so indeed it will live on. But RevocableServer is a much simpler and more direct solution to this common problem.

PR #1668, opened 2023-04-24
Added getCanceler function to capnp::MembranePolicy

Zentren:
Permits synchronous (immediate) revocation that does not rely on event loop processing.

### DIFF
--- a/c++/src/capnp/membrane-test.c++
+++ b/c++/src/capnp/membrane-test.c++
@@ -51,14 +51,38 @@ private:
   kj::StringPtr text;
 };
 
-class TestMembraneImpl final: public test::TestMembrane::Server {
+class TestMembraneImpl final: public test::TestMembrane::Server, public kj::Refcounted {
+public:
+  TestMembraneImpl(kj::Maybe<kj::Promise<void>> chainablePromise = nullptr) : chainablePromise(kj::mv(chainablePromise)) {}
+  kj::Own<TestMembraneImpl> addRef() {
+    return kj::addRef(*this);
+  }
+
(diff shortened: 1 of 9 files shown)


Zentren:
This is intended to be a draft with my experiments so far to address  #1667.

There are a number of issues that'd likely preclude merging at the moment:
1. It's not working.
2. Changes needs documentation and cleanup.
3. The membrane tests are a huge mess.
4. Inclusion of Boost (signals2) - this is because it'd take some time to create a decent subject/observer (signal/slot) impl.

We'd appreciate help with (1) as we are unsure as to why the changes to `membrane.cpp` are insufficient to cancel a chained promise inside a request.

kentonv:
I actually looked into implementing this myself, and found it trickier than expected.

Instead I created #1700, which I think is an alternate solution to your specific problem, and also a solution to a class of problems I've experienced with capnp.

If that works for you, perhaps we should just abandon this PR?

Zentren:
> I actually looked into implementing this myself, and found it trickier than expected.
> 
> Instead I created #1700, which I think is an alternate solution to your specific problem, and also a solution to a class of problems I've experienced with capnp.
> 
> If that works for you, perhaps we should just abandon this PR?

Hi Kenton, sincere apologies for the very much delayed response. I have looked at that PR and it looks like exactly the solution to the problem :smile: I will confirm later in the week after testing a bit that there aren't any strange behaviors using it live, but the pattern of use shown in the unit test looks like how we'd use it :+1:

Zentren:
I'll close the PR in the meantime as it is redundant. I'll comment on the related issue in due course as it can probably be closed.

PR #1669, opened 2023-04-26
Implement async environments with arenas

harrishancock:
Copied from the main commit:

Supersedes https://github.com/capnproto/capnproto/pull/1546.

----

This commit adds a new feature to the kj-async library: Environments.

User-facing documentation is available in "A tour of KJ", so this commit message will contain only a few implementation notes.

Our desired semantics for asynchronous environments is for them to act like thread-local storage for the abstract threads of execution formed by promise call graphs. An initial implementation leaned directly on thread-local storage by storing the active environment in the EventLoop data structure, and required every Promise constructor to look up the current EventLoop. This introduced a measurable performance regression for code which did not even use the feature.

The implementation in this commit instead avoids the constant thread-local storage lookups by piggybacking off of the recent PromiseArena optimization. Now, we only need to look up the EventLoop at the start of a new PromiseArena, and the environment set is automatically propagated to all promises which are added to the arena. In practice, these promises are always dependents of the earlier promises in the arena, and this commit requires that this practice become a rule. With this rule in place, we can assume that all promises in an arena are always dependencies of other promises in the same arena (or themself the root which captured the environment). Since all promises created within an environment must be dependents of a common ancestor, or themselves roots, our optimization has semantics which propagate the environment to at least a superset of the promises that our desired semantics would have.

The mismatch between our desired semantics and our optimization's semantics is that it may be possible for environments to propagate to dependent promises that we did not intend. That is, it is impossible for us to statically guarantee that promise arenas do not gain promises which were constructed under a different environment. Instead, we detect this circumstance at runtime in debug mode, and crash (in release mode, it would manifest as a wrong-environment bug). Additionally, we provide a couple tools -- `.adoptEnvironment()` and TaskSets -- to import and export promises across environments.

PromiseNode implementations are responsible for constructing environment scopes wherever necessary to cover user code invocations. This commit adds such scopes to all known user code invocation sites. Some rules to follow to demystify this process for future PromiseNode implementations can be found in the Async Environments comment in async-inl.h.


c++/WORKSPACE
```diff
@@ -45,7 +45,7 @@ cc_library(
 http_archive(
     name = "zlib",
     build_file_content = _zlib_build,
-    sha256 = "d14c38e313afc35a9a8760dadf26042f51ea0f5d154b0630a31da0540107fb98",
+    sha256 = "b3a24de97a8fdbc835b9833169501030b8977031bcb54b3b3ac13740f846ab30",
     strip_prefix = "zlib-1.2.13",
-    urls = ["https://zlib.net/zlib-1.2.13.tar.xz"],
+    urls = ["https://zlib.net/fossils/zlib-1.2.13.tar.gz"],
```
harrishancock, line 50: This particular diff is not required to build at present, since 1.2.13 is still the current zlib version. I initially added this diff when working off an earlier commit, and should perhaps remove it from this PR.

harrishancock:
Rebased on master.

kentonv (commented):
I didn't look at the main commit yet... meant to but ran out of time.

c++/src/kj/compat/http.c++
```diff
@@ -7397,7 +7397,7 @@ private:
       // try to find a configuration that respects both the server's preferred configuration,
       // as well as the client's requested configuration.
       //
-      // TODO(now): Manual Mode and terminating in worker
+      // TODO(soon): Manual Mode and terminating in worker
```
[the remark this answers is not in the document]
MellowYarker: Yes, I suspect the reason the todo was left in was we implemented the fix in workerd ([here](https://github.com/cloudflare/workerd/blob/f4739148bf9e7a44296294dd61e1f9b39462f699/src/workerd/api/http.c%2B%2B#L1322-L1330)) and forgot to remove it here.

This entire comment can be removed:
```
// TODO(soon): Manual Mode and terminating in worker
//    - `headers` would be empty.
//    - This means client would offer x, and the server would always reject.
//    - We want a way to explicitly tell `acceptWebSocket()` to use the `requestHeaders` in
//      this case.
```

c++/src/kj/async.h
```diff
@@ -362,6 +362,25 @@ class Promise: protected _::PromiseBase {
...
+  //   kj::runInEnvironment(environment, [&]() {
+  //     return outerPromise.adoptEnvironment()
+  //                        .then([]() {
+  //       kj::getEnvironment<double>();
+  //     });
+  //   });
+  //
+  // In the above code, if the `.adoptEnvironment()` call were missing, the call to `.then()` would
+  // throw an exception in debug mode. (In release mode, that check is omitted, and the call to
+  // `kj::getEnvironment<double>()` would throw instead.)
```
kentonv, line 382: To clarify, would it throw because it detects misuse, or would it throw because it inherited the wrong environment and `double` doesn't exist in the environment?

kjdoc/tour.md
```diff
@@ -975,6 +975,114 @@ As of this writing, KJ supports C++20 coroutines and Coroutines TS coroutines, t
 
 KJ prefers C++20 coroutines when both implementations are available.
 
+### Asynchronous environments
+
+Environments are ambiently available storage, of a type specified by the user, which "follow" user callbacks scheduled via functions/syntax like evalLater(), promise.then(), startFiber(), co_await, and more.
+
+Environments do not cross threads. Type safety is as safe as kj::downcast<T>() makes it -- in debug mode, you'll probably get an assertion if you set an environment of type T and try to get it later as type U -- but in release mode you're on your own.
```
kentonv, line 982: It's not clear here where "type safety" is involved.

Does this mean there's only one active environment at any particular time, it has one specific type, and you have to use the correct type when calling `getEnvironment()`?

I would argue it's worth making this throw in release mode too.

But meanwhile I kind of expected that the "environment" actually has a slot for every possible type. `runInEnvironment` would only replace one of these slots, leaving the rest with their existing value. This is probably what we want since there may be multiple use cases for environments within the same codebase and forcing them all to share a single environment type might be cumbersome.

harrishancock: Hmm, yes, I think this paragraph must have survived from a much earlier implementation, and I didn't revise it.

The current implementation has a slot for every possible type, using a `HashMap<void*, Own<Environment>>`. The `void*` is a "type key" implemented as the address of a static variable inside a function template parameterized by the user's concrete type, T. `Environment` is a base class whose derived implementations are holders of T. Since `runInEnvironment()` and `getEnvironment()` both know what T is, they each calculate the same type key, and thus the downcast should be safe. If we pass a type U to `getEnvironment()`, the HashMap will just tell us no value exists for that type key, so my comment here is wrong.

c++/src/kj/async-test.c++
```diff
@@ -1653,5 +1653,24 @@ KJ_TEST("constPromise") {
...
+  return evalLater([z, ttl]() {
+    return asyncAccumulate(z, ttl - 1);
+  });
+}
+
+KJ_TEST("benchmark: Pre-environment evalLater()") {
+  EventLoop eventLoop;
+  WaitScope waitScope(eventLoop);
+
+  asyncAccumulate(0, 10'000'000).wait(waitScope);
```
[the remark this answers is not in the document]
harrishancock: Hmm, I'll take a look. I'd also be happy just to drop this little test, too -- the hard-coded count is a bit unfortunate,.

c++/src/kj/async-inl.h
```diff
@@ -233,21 +420,52 @@ class PromiseArenaMember {
...
   // allocated this way are required to have `PromiseNode` itself as their leftmost inherited type,
-  // so that the pointers match. Thus, the space in `arena` from its start to the location of the
-  // `PromiseNode` is known to be available for subsequent allocations (which should then take
+  // so that the pointers match. Thus, the space in `arena->space` from its start to the location of
+  // the `PromiseNode` is known to be available for subsequent allocations (which should then take
   // ownership of the arena).
 
+  struct NoEnv {};
+
+  OneOf<NoEnv, Maybe<EnvironmentSet>, Maybe<EnvironmentSet&>> environmentSet = NoEnv();
```
kentonv, line 458: This is adding three pointers to the size of every `PromiseArenaMember`. If we could manage to optimize it down to `Maybe<EnvironmentSet&>`, then it would be adding only one pointer, which could be a pretty significant difference. The main thing standing in the way seems to be the "standalone" PromiseArenaMembers that don't end up allocated inside an arena. Looking at this, uh, I suspect this code path already leaks memory before your change (my fault). It looks like they will be allocated on the heap and never freed at present. Presumably we don't today have any PromiseArenaMembers that take this code path, or if we do, we have bigger problems?

Of course, even better would be if this didn't have to be tracked on every single node at all. This might be possible. The purpose of this field is only to set the correct environment during the destructor, right? But I think we can say that it's the parent PromiseNode's responsibility to set the correct environment before destroying children. In fact that's probably largely satisfied already, when a parent PromiseNode exists. The only problem case is when there is no parent, i.e. the PromiseNode is directly held by the application via a Promise. Here, we cannot expect the application to set the correct environment when dropping the Promise. HOWEVER, in this particular case, `arena` should always be non-null! So we can get the environment off of the arena! (Well, expect for the stand-alone promise memory leak issue discussed above... which needs to be solved some other way anyway.)

c++/src/kj/async-inl.h
```diff
@@ -377,13 +614,32 @@ class PromiseDisposer {
...
+      //   Promise<void> promise2 = nullptr;
+      //   runInEnvironment([&]() {
+      //     promise1.then([]() {});  // ERROR: should have used .adoptEnvironment()
+      //     promise2 = READY_NOW;
+      //   });
+      //   promise2.then([]() {});  // ERROR: should have returned the promise
       //
+      // Unfortunately, we alert the code author to these mistakes by crashing, because
+      // `appendPromise()` is noexcept. Oh well.
+      requireEnvironmentSetEqual(next->getEnvironmentSet(), EnvironmentSet::tryGetCurrent());
```
kentonv, line 640: If we reduced this to an operator== (see comment on implementation), it might be cheap enough to do the check in release mode. And then, instead of throwing when it fails, we could actually add the missing node to the chain, maybe?

PR #1670, opened 2023-05-01
Implement brotli-based streams

fhanau:
This PR adds support for the Brotli compression algorithm as part of the kj/compat streams interface. The interface and test cases are based on those for gzip.
Potential caveats:
- Build system integration is limited to bazel so far. While the brotli streams don't add much to build time and are compiled into its own library we could turn building brotli off by default depending on how much usage of brotli we expect and only run the tests in CI (i.e. only define `--//src/kj:brotli=True` there).
- The build system has also been restructured to allow dependent projects to pull in brotli via capnproto, this should make the process of updating or switching brotli versions easier for dependents. Build-related scripts have been moved to the new directory `c++/build/`.
- The approach to the decoding window size needs to be carefully chosen, I will look into the current approach again to make sure it's sane. Currently the decoding window size is hardcoded through `KJ_BROTLI_MAX_DEC_WBITS`, I can make it configurable if we anticipate that there are use cases where memory usage is more constrained/where we expect inputs using the max window size.


c++/src/kj/compat/gzip.h
```diff
@@ -29,12 +29,14 @@ KJ_BEGIN_HEADER
 
 namespace kj {
 
+#define KJ_GZ_BUF_SIZE 4096
```
kentonv, line 32: Prefer constants over macros:

```suggestion
constexpr size_t KJ_GZ_BUF_SIZE = 4096;
```

And can this be moved inline the `_` namespace? I don't think this constant is intended to be public.

c++/src/kj/compat/gzip.h
```diff
@@ -29,12 +29,14 @@ KJ_BEGIN_HEADER
...
+
 namespace _ {  // private
 
 class GzipOutputContext final {
 public:
   GzipOutputContext(kj::Maybe<int> compressionLevel);
-  ~GzipOutputContext() noexcept(false);
+  ~GzipOutputContext();
```
kentonv, line 39: Why did you remove `noexcept(false)` here? Per KJ style, all destructors should have this annotation.

https://github.com/capnproto/capnproto/blob/master/style-guide.md#exceptions-usage

kentonv (commented):
Is the brotli code largely a clone of the zlib code? Would it make sense to define an abstract interface for a zlib-like compression library in order to allow these to share code?

c++/src/kj/compat/gzip.h
fhanau: Adjusted it – I wasn't aware that the style guide requires `noexcept(false)`, I assumed it wasn't needed here as the destructor only calls a C cleanup function.

fhanau:
> Is the brotli code largely a clone of the zlib code? Would it make sense to define an abstract interface for a zlib-like compression library in order to allow these to share code?

The brotli API has been inspired by zlib's API, but has marked differences that make it difficult to share code, i.e. having different parameters for flushing, different types for configuration parameters and return values, as well as behaving differently in how much input data they consume at once.

I ended up defining a template class for brotli and zlib OutputStream/AsyncOutputStream, turns out this can be done pretty concisely by relying on `GzipOutputContext` and `BrotliOutputContext`. For the input streams, this would be much more difficult as they interact with the API more closely; defining an abstract class for compression operations would also be pretty complex as there's no direct match between zlib and brotli for some operations.

fhanau:
I initially tried using polymorphism with the Gzip and Brotli classes inherit from a base class, but that wasn't very helpful for avoiding code duplication since several of the operations depend on the compressor-specific flush parameters or the compression context class. I looked into it again and it could be done better than what I had in mind by having an abstract class for `GzipContext`/`BrotliContext` too, but there would still be overhead to define the different classes and handle the flush parameters, making the code less readable.
Templated code can cause increases in compile time and code size, especially if a class is instantiated with a number of different parameters. In this case it should be fine as there's just one instantiation for gzip and one for brotli, the size of `//src/kj/compat:gzip-test` goes from 652KB to 654KB when switching to the templated approach which feels reasonable.

I'm not sure if there's an intuitive way to split up the changes since it's mostly new code. It's certainly a lot for one commit – please let me know if you find anything else unclear or difficult to review.

kentonv:
Sorry but I'm not into the template approach.

I think the way to deal with the param types being different would be to define a separate, common enum that is used as the parameter to `pumpOnce()`, and have the respective implementations convert that to their preferred type via a lookup table.

That said, let's just go back to not trying to share implementations. I don't want this PR to end up bogged down on trying to get that right.

fhanau:
> Sorry but I'm not into the template approach.
> 
> I think the way to deal with the param types being different would be to define a separate, common enum that is used as the parameter to `pumpOnce()`, and have the respective implementations convert that to their preferred type via a lookup table.
> 
> That said, let's just go back to not trying to share implementations. I don't want this PR to end up bogged down on trying to get that right.

Addressed in the latest fixup.

c++/src/kj/compat/brotli.c++
```diff
@@ -0,0 +1,355 @@
...
+    auto chunk = get<1>(result);
+    if (chunk.size() > 0) {
+      inner.write(chunk.begin(), chunk.size());
+    }
+  } while (ok);
+}
+
+// =======================================================================================
+
+BrotliAsyncInputStream::BrotliAsyncInputStream(AsyncInputStream& inner, kj::Maybe<int> _windowBits)
```
harrishancock, line 238: Nit: `windowBitsParam` would be a bit more conventional

c++/src/kj/compat/brotli.c++
```diff
@@ -0,0 +1,355 @@
...
+    : inner(inner), next_in(nullptr), available_in(0) {
+  windowBits = _windowBits.orDefault(_::KJ_BROTLI_MAX_DEC_WBITS);
+  KJ_REQUIRE(windowBits >= BROTLI_MIN_WINDOW_BITS && windowBits <= BROTLI_MAX_WINDOW_BITS,
+      "invalid brotli window size (window bits", windowBits, ")");
+  ctx = BrotliDecoderCreateInstance(nullptr, nullptr, nullptr);
```
harrishancock, line 243: Nit: `windowBits` and `ctx` could be initialized in the constructor initializer list

c++/src/kj/compat/brotli.c++
```diff
@@ -0,0 +1,355 @@
...
+  } else {
+    BROTLI_BOOL result = BrotliEncoderCompressStream(cctx, flush, &available_in, &next_in,
+                                                     &available_out, &next_out, nullptr);
+    KJ_REQUIRE(result == BROTLI_TRUE, "brotli compression failed");
+  }
+
+  // The 'ok' parameter represented by the first parameter of the tuple indicates that pumpOnce()
+  // should be called again as more output data can be produced. This is the case when the stream
+  // is not finished and there is either pending output data (that didn't fit into the buffer) or
+  // input that has not been processed yet.
```
harrishancock, line 130: We tend not to use tuples in return types, because they're not very self-documenting. Instead, I'd probably express the return type as something like this, integrating your comment here.
```cpp
struct PumpOnceResult {
  bool callAgain;
  // If true, `pumpOnce()` should be called again, as more output data can be produced. This is the case when ...

  kj::ArrayPtr<const byte> output;
  // A view into BrotliOutputContext's internal buffer containing the encoded/decoded output.
};
```

I see that the tuple pattern was copied from the gzip implementation, so I'll leave it up to you if you want to change this, change both, or leave as-is.

c++/src/kj/compat/brotli.c++
```diff
@@ -0,0 +1,355 @@
...
+#include "brotli.h"
+#include <kj/debug.h>
+
+namespace kj {
+
+namespace _ {  // private
+
+// Check number of window bits used by the stream, see RFC 7932 section 9.1 for the specification.
+// Adapted from FL's ngx_brotli.
+static int get_brotli_window_bits(uint8_t peek) {
```
harrishancock, line 33: Non-blocking nit: we tend to prefer enclosing things in an anonymous namespace (`namespace { ... }`) instead of `static` qualifiers. I think they're exactly equivalent, though.

c++/src/kj/compat/brotli.c++
```diff
@@ -0,0 +1,355 @@
...
+
+#include "brotli.h"
+#include <kj/debug.h>
+
+namespace kj {
+
+namespace _ {  // private
+
+// Check number of window bits used by the stream, see RFC 7932 section 9.1 for the specification.
+// Adapted from FL's ngx_brotli.
```
harrishancock, line 32: Three things:
- How about, "an internal Cloudflare codebase"? I worry that mentioning ngx_brotli at all would be misleading, unless we plan on upstreaming our fork's changes.
- Since it's so short, I think we might as well KJ-style it, with doc comment after the prototype, camelCase, and 2-space indentation.
- I'd love a link to https://www.rfc-editor.org/rfc/rfc7932.html#section-9.1 in the comment.

c++/src/kj/compat/brotli.c++
```diff
@@ -0,0 +1,355 @@
...
+void BrotliOutputContext::setInput(const void* in, size_t size) {
+  next_in = reinterpret_cast<const byte*>(in);
+  available_in = size;
+}
+
+kj::Tuple<bool, kj::ArrayPtr<const byte>> BrotliOutputContext::pumpOnce(BrotliEncoderOperation flush) {
+  byte* next_out = buffer;
+  size_t available_out = sizeof(buffer);
+  // Brotli does not accept a null input pointer; make sure there is a valid pointer even if we are
+  // not actually reading from it.
```
harrishancock, line 98: When would this occur?

kentonv:
I'm happy to defer to Harris's review for the rest of this. Sorry for the run-around regarding factoring out common implementation, I probably shouldn't have suggested that.

c++/src/kj/compat/brotli.h
```diff
@@ -0,0 +1,188 @@
...
+
+namespace _ {  // private
+// Use a window size of (1 << 19) = 512K by default. Higher values improve compression on longer
+// streams but increase memory usage.
+constexpr size_t KJ_BROTLI_DEFAULT_WBITS = 19;
+
+// Maximum window size for streams to be decompressed, streams with larger windows are rejected.
+// The default of (1 << BROTLI_DEFAULT_WINDOW) = 4MB should be useful to limit memory usage, but
+// also means that some valid streams will be rejected.
+constexpr size_t KJ_BROTLI_MAX_DEC_WBITS = BROTLI_DEFAULT_WINDOW;
```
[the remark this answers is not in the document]
fhanau: I would prefer to keep the constants here for visibility – the default values should be reasonable but there may be use cases where a different window size or compression level is neded.

c++/src/kj/compat/brotli.c++
```diff
@@ -0,0 +1,355 @@
...
+    // Emulate zlib's behavior of using -1 to signify the default quality
+    if (*level == -1) {*level = KJ_BROTLI_DEFAULT_QTY;}
+    KJ_REQUIRE(*level >= BROTLI_MIN_QUALITY && *level <= BROTLI_MAX_QUALITY,
+        "invalid brotli compression level", *level);
+    windowBits = _windowBits.orDefault(_::KJ_BROTLI_DEFAULT_WBITS);
+    KJ_REQUIRE(windowBits >= BROTLI_MIN_WINDOW_BITS && windowBits <= BROTLI_MAX_WINDOW_BITS,
+        "invalid brotli window size (window bits", windowBits, ")");
+    cctx = BrotliEncoderCreateInstance(nullptr, nullptr, nullptr);
+    KJ_REQUIRE(cctx, "brotli state allocation failed");
+    // TODO(performance): Make window size configurable from the stream interface and consider
```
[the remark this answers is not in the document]
fhanau: You're right, with the parameterized window size this is no longer needed.

c++/src/kj/compat/brotli.c++
```diff
@@ -0,0 +1,355 @@
...
+    // setting up BROTLI_PARAM_LGBLOCK if memory usage becomes an issue
+    KJ_ASSERT(BrotliEncoderSetParameter(cctx, BROTLI_PARAM_QUALITY, *level) == BROTLI_TRUE);
+    KJ_ASSERT(BrotliEncoderSetParameter(cctx, BROTLI_PARAM_LGWIN, windowBits) == BROTLI_TRUE);
```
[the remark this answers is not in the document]
fhanau: This should be unlikely – `BrotliEncoderSetParameter` should always work as long as the encoder state is initialized and the parameters are in the correct range, which is checked above.

c++/src/kj/compat/brotli.c++
```diff
@@ -0,0 +1,355 @@
...
+      available_in = amount;
+    }
+  }
+
+  byte* next_out = out;
+  size_t available_out = maxBytes;
+  // Check window bits
+  if (firstInput) {
+    firstInput = false;
+    int streamWbits = _::get_brotli_window_bits(next_in[0]);
```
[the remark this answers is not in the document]
fhanau: I added `availableIn > 0` to the if condition in all spots where the window bits are checked – this ensures that if `pumpOnce()` is called before an input is set it will effectively perform a no-op and we only do the window bits check if data is available.

c++/src/kj/compat/brotli.c++
```diff
@@ -0,0 +1,355 @@
...
+}  // namespace _ (private)
+
+// =======================================================================================
+
+BrotliInputStream::BrotliInputStream(InputStream& inner, kj::Maybe<int> _windowBits)
+    : inner(inner), next_in(nullptr), available_in(0) {
+  windowBits = _windowBits.orDefault(_::KJ_BROTLI_MAX_DEC_WBITS);
+  KJ_REQUIRE(windowBits >= BROTLI_MIN_WINDOW_BITS && windowBits <= BROTLI_MAX_WINDOW_BITS,
+      "invalid brotli window size (window bits", windowBits, ")");
+  ctx = BrotliDecoderCreateInstance(nullptr, nullptr, nullptr);
```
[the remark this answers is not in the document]
fhanau: I added windowBits to the initializer list – I'd prefer having ctx in the body as we might otherwise leak memory if the windowBits check fails.

c++/src/kj/compat/brotli.c++
fhanau: Done – since there is no prototype should I keep the doc comment as-is right?

c++/src/kj/compat/brotli.c++
fhanau: I mostly added this for defensive purposes, but it should be possible if pump() is called before `setInput()`.

c++/src/kj/compat/brotli.c++
harrishancock: Ah, no, sorry -- the doc comment should always go below the function prototype, if there is one, or the definition's signature, if there is no prototype:
```cpp
int getBrotliWindowSize() {
  // Check number of window bits ...
```

c++/src/kj/compat/brotli.c++
```diff
@@ -0,0 +1,365 @@
...
+#include <kj/debug.h>
+
+namespace kj {
+
+namespace _ {  // private
+
+// Check number of window bits used by the stream, see RFC 7932
+// (https://www.rfc-editor.org/rfc/rfc7932.html#section-9.1) for the specification.
+// Adapted from an internal Cloudflare codebase.
+int getBrotliWindowBits(kj::byte peek) {
```
harrishancock, line 34: This needs to be enclosed in `namespace { }` now that it's non-`static`, to make it invisible outside this translation unit.
```cpp
namespace {

int getBrotliWindowBits(kj::byte peek) {
  // Check number of window bits ...
  // ...
}

}  // namespace
```

(It's exactly equivalent to adding the `static` keyword to the function signature, meaning my suggestion to change it was very much a nitpick.)

harrishancock (approved):

PR #1671, opened 2023-05-03
Implement `joinPromisesFailFast()`

harrishancock:
@dom96 and I were recently bitten rather hard by a latent bug in `HttpServiceAdapter::connect()`, in which pump exceptions do not eagerly cancel the returned promise, but instead hang in limbo.

We know roughly what a solution should look like: a fail-fast form of `joinPromises()`. I took a stab at implementing such a thing in this PR, which I called `joinPromisesFailFast()`.

A subsequent commit in the PR uses `joinPromisesFailFast()` to solve a few of the places we know this bug exists. Notably, `HttpServiceAdapter::connect()` required slight adjustments in order to keep tests passing, because exceptions from the pump promises must not cancel the returned promise in the event of an error response.

### DIFF
--- a/c++/src/kj/async-test.c++
+++ b/c++/src/kj/async-test.c++
@@ -667,36 +667,134 @@ TEST(Async, ExclusiveJoin) {
 }
 
 TEST(Async, ArrayJoin) {
-  EventLoop loop;
-  WaitScope waitScope(loop);
+  for (auto specificJoinPromisesOverload: {
+    +[](kj::Array<kj::Promise<int>> promises) { return joinPromises(kj::mv(promises)); },
+    +[](kj::Array<kj::Promise<int>> promises) { return joinPromisesFailFast(kj::mv(promises)); }
+  }) {
+    EventLoop loop;
+    WaitScope waitScope(loop);
 
-  auto builder = heapArrayBuilder<Promise<int>>(3);
-  builder.add(123);
-  builder.add(456);
-  builder.add(789);
+    auto builder = heapArrayBuilder<Promise<int>>(3);
+    builder.add(123);
+    builder.add(456);
+    builder.add(789);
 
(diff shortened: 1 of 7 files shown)


c++/src/kj/async-test.c++
```diff
@@ -699,6 +699,47 @@ TEST(Async, ArrayJoinVoid) {
...
+    KJ_EXPECT(!promise.poll(waitScope));
+    KJ_EXPECT(tasksCompleted == i);
+
+    fulfillers[i++]->reject(KJ_EXCEPTION(FAILED, "Test exception"));
+
+    // The first rejection makes the promise ready.
+    KJ_EXPECT_THROW_MESSAGE("Test exception", promise.wait(waitScope));
+  }
+}
+
```
dom96, line 742: Might be worth creating the same test but for `joinPromises` instead, both to show their differences clearly but also to ensure coverage.

dom96 (commented):
Nice! I do wonder though: anything stopping us from making joinPromisesFailFast the default?

kentonv (approved):

harrishancock:
> Nice! I do wonder though: anything stopping us from making joinPromisesFailFast the default?

I wish we could, but it's very likely to be a breaking change somewhere -- it in fact turned out to be so for the code in `HttpServiceAdapter::connect()`, which was what I specifically wanted to change.

For now, I added a TODO comment to `joinPromises()` that we should deprecate it -- I don't want to do that in this PR, though.

c++/src/kj/async-test.c++
harrishancock: I agree; I made a version of this test for `joinPromises()`, and also refactored the existing `joinPromises()` happy-path tests to run against `joinPromisesFailFast()`.

PR #1672, opened 2023-05-09
Include additional ws compression error reason

MellowYarker:
Turns out there's a "reason" string for what went wrong when attempting to inflate the compressed data; we should add this to the error.


MellowYarker:
From `zlib.h`'s comments on `inflate()`:

> [returns] ... Z_DATA_ERROR if the input data was
  corrupted (input stream not conforming to the zlib format or incorrect check
  value, in which case strm->msg points to a string with a more specific
  error)

We've seen `Z_DATA_ERROR` but can't determine what's wrong without seeing the reason.

MellowYarker:
@kentonv when you've got some time (there's no rush on this). The reason I didn't also add it to the compression error is `zlib.h` doesn't say it sets `msg` if `deflate()` fails.

PR #1674, opened 2023-05-10
Cleanup: Only log HTTP-over-RPC errors at INFO level

bcaimano:
These errors only arise when an existing client is no longer able to receive a response. In practice, that sort of error is always uninteresting.

### DIFF
--- a/c++/src/capnp/compat/http-over-capnp.c++
+++ b/c++/src/capnp/compat/http-over-capnp.c++
@@ -616,19 +616,16 @@ public:
       hasBody = *s > 0;
     }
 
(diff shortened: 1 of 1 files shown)


kentonv (approved):
Seem reasonable.

In production Sentry I only see this error happening for `startWebSocket()` and only due to the client being a Durable Object that has become overloaded. Are you aware of anywhere else this is logging currently?

kentonv:
(Still waiting for CI, it apparently had a hiccup the first time.)

kentonv:
Oh heh I see we discussed this three months ago, OK makes sense.

PR #1675, opened 2023-05-11
Add missing AsyncIoStream::getFd() overrides

jp4a50:
A couple of trivial `getFd()` overrides were missing.

You could perhaps argue that `getFd()` should be pure virtual to catch such missed overrides.

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -4124,8 +4124,13 @@ public:
                                    leftoverBackingBuffer(kj::mv(leftoverBackingBuffer)),
                                    leftover(leftover) {}
 
+  // AsyncIoStream
   void shutdownWrite() override {
     stream->shutdownWrite();
+  }
+
+  kj::Maybe<int> getFd() const override {
+    return stream->getFd();
   }
 
   // AsyncInputStream
@@ -4254,6 +4259,20 @@ public:
         writeGuard(handleWriteGuard(kj::mv(writeGuard))),
         tasks(*this) {}
 
+
+  // AsyncIoStream
+  void shutdownWrite() override {
+    if (writeGuardReleased) {
+      inner->shutdownWrite();
+    } else {
+      tasks.add(writeGuard.addBranch().then([this]() { inner->shutdownWrite(); }));
+    }
+  }
+
+  Maybe<int> getFd() const override {
+    return inner->getFd();
+  }
+
   // AsyncInputStream
   Promise<size_t> tryRead(void* buffer, size_t minBytes, size_t maxBytes) override {
     if (readGuardReleased) {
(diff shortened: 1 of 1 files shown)


c++/src/kj/compat/http.c++
```diff
@@ -4124,10 +4124,15 @@ public:
                                    leftoverBackingBuffer(kj::mv(leftoverBackingBuffer)),
                                    leftover(leftover) {}
 
+  // AsyncIoStream
   void shutdownWrite() override {
     stream->shutdownWrite();
   }
 
+  kj::Maybe<int> getFd() const override {
+    return stream->getFd();
```
harrishancock, line 4133: I'm not sure if it makes sense to implement `getFd()` for this class, since it's wrapping not just another stream, but a buffer as well. I guess it depends on what sort of semantics we want to provide `getFd()`. If we want to be able to guarantee that the returned FD will produce the same bytes that reading from this AsyncIoStream implementation would, then that might be a problem.

jp4a50: In practice (in the usages of AsyncIoStreamWithInitialBuffer that I can see) I think the data in the leftover backing buffer has always previously been read from the `stream`'s FD, but it's true that the design of this class does not guarantee this.

Based on the documentation of `AsyncIoStream::getFd` which is:

>  // Get the underlying Unix file descriptor, if any. Returns nullptr if this object actually
>  // isn't wrapping a file descriptor.

I definitely wouldn't expect this to *always* return empty (which it does if we don't implement this).

An alternative could be to only return a non-empty FD if the `leftoverBackingBuffer` has been fully read/emptied. That would satisfy your criterion but only from the point at which `getFd()` starts returning non-empty. I'd be happy enough with that as a compromise since it would then at least sometimes be useful from a debugging perspective.

jp4a50: These stream classes are all actually internal to `http.c++` so we could make it more explicit that any backing buffer provided should satisfy your proposed criterion (as all usages currently seemingly do).

harrishancock: I'd be curious what @kentonv thinks.

kentonv: I agree with @harrishancock. The contract of `getFd()` is that if it returns an FD, then the stream itself does not implement any special behavior on top of that FD. For example, read()ing directly from the FD should produce identical results to read()ing from the stream.

That is not the case here.

In practice, `AsyncFdStream` (in `async-io.c++`) is probably the only stream that should ever implement `getFd()` to return non-null.

kentonv:
As described above I don't think this is correct behavior.

What is the use case you're trying to solve?

jp4a50:
> As described above I don't think this is correct behavior.
> 
> What is the use case you're trying to solve?

I was just trying to use getFd to discover what FD was being used under the hood by different connections for debugging purposes. Specifically I was trying to ensure that a particular websocket connection was closing its underlying FD when it was destroyed.

I can live without these changes, just thought they would be useful for debugging.

kentonv:
That said I could see an argument for a `debugInfo()` method that returns a string describing the stream, that's something I could probably have used many times in the past. I often find myself using `typeid(*ptr).name()` to find out what kind of stream I'm even dealing with...

PR #1678, opened 2023-05-16
No schema on dynamic set of generic

reissGRVS:
These tests demonstrate that you cannot set values for a list of enums via dynamic if it is wrapped in a generic.

### DIFF
--- a/c++/src/capnp/dynamic-test.c++
+++ b/c++/src/capnp/dynamic-test.c++
@@ -473,6 +473,40 @@ TEST(DynamicApi, SetDataFromText) {
   EXPECT_EQ(data("foo"), root.get("dataField").as<Data>());
 }
 
+void wrappedEnumListTest(std::string field)
+{
+  MallocMessageBuilder builder;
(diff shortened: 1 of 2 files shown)


c++/src/capnp/dynamic-test.c++
```diff
@@ -473,6 +473,40 @@ TEST(DynamicApi, SetDataFromText) {
...
+  wrapper.set("value", {TestEnum::BAR, TestEnum::FOO});
+  checkList<TestEnum>(wrapper.get("value"), {TestEnum::BAR, TestEnum::FOO});
+}
+
+TEST(DynamicApi, SetWrappedEnumListFromNative) {
+  wrappedEnumListTest("wrappedEnumList");
+}
+
+TEST(DynamicApi, SetGenericWrappedEnumListFromNative) {
+  wrappedEnumListTest("genericWrappedEnumList");
```
reissGRVS, line 490: This test fails due to a null schema ptr when calling Type::asEnum().
The rest of the tests work correctly.

kentonv:
Fixed in #1699.

reissGRVS:
> Fixed in #1699.

Nice thanks!

PR #1683, opened 2023-05-25
Add new type ConstString

ObsidianMinor:
This type is useful for cases where you want to store a borrowed StringPtr or an owned String and don't care about mutability of the String itself.

### DIFF
--- a/c++/src/kj/string.h
+++ b/c++/src/kj/string.h
@@ -23,13 +23,16 @@
 
 #include <initializer_list>
 #include "array.h"
+#include "kj/common.h"
 #include <string.h>
 
 KJ_BEGIN_HEADER
 
 namespace kj {
   class StringPtr;
+  class LiteralStringConst;
   class String;
+  class ConstString;
 
   class StringTree;   // string-tree.h
 }
@@ -49,6 +52,8 @@ constexpr kj::StringPtr operator "" _kj(const char* str, size_t n);
 // In other cases, there should be no difference between initializing a StringPtr from a regular
 // string literal vs. one with _kj (assuming the compiler is able to optimize away strlen() on a
 // string literal).
+
+constexpr kj::LiteralStringConst operator "" _kjc(const char* str, size_t n);
 
 namespace kj {
 
@@ -76,6 +81,7 @@ public:
   inline StringPtr(const char* begin KJ_LIFETIMEBOUND, const char* end KJ_LIFETIMEBOUND): StringPtr(begin, end - begin) {}
   inline StringPtr(String&& value KJ_LIFETIMEBOUND) : StringPtr(value) {}
   inline StringPtr(const String& value KJ_LIFETIMEBOUND);
+  inline StringPtr(const ConstString& value KJ_LIFETIMEBOUND);
   StringPtr& operator=(String&& value) = delete;
   inline StringPtr& operator=(decltype(nullptr)) {
     content = ArrayPtr<const char>("", 1);
@@ -162,12 +168,18 @@ public:
   Maybe<T> tryParseAs() const;
   // Same as parseAs, but rather than throwing an exception we return NULL.
 
+  template <typename... Attachments>
+  ConstString attach(Attachments&&... attachments) const KJ_WARN_UNUSED_RESULT;
+  ConstString attach() const KJ_WARN_UNUSED_RESULT;
+  // Like ArrayPtr<T>::attach(), but instead promotes a StringPtr into a ConstString. Generally the
+  // attachment should be an object that somehow owns the String that the StringPtr is pointing at.
+
 private:
   inline explicit constexpr StringPtr(ArrayPtr<const char> content): content(content) {}
+  friend constexpr StringPtr (::operator "" _kj)(const char* str, size_t n);
+  friend class LiteralStringConst;
 
   ArrayPtr<const char> content;
-
-  friend constexpr kj::StringPtr (::operator "" _kj)(const char* str, size_t n);
   friend class SourceLocation;
 };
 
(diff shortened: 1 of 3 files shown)


ObsidianMinor:
Not sure what kind of tests should be written or if any existing APIs should be adjusted.

c++/src/kj/string.h
```diff
@@ -296,6 +306,105 @@ class String {
...
+// A ConstString can be made by the explicit move from String constructor, or by providing a
+// StringPtr.
+
+class ConstString {
+public:
+  ConstString() = default;
+  inline ConstString(decltype(nullptr)): content(nullptr) {}
+  inline ConstString(const char* value, size_t size, const ArrayDisposer& disposer);
+  // Does not copy.  `size` does not include NUL terminator, but `value` must be NUL-terminated.
+  inline ConstString(kj::StringPtr ptr);
```
kentonv, line 325: I don't think we should so easily implicitly convert a StringPtr -- which may have limited lifetime -- to `ConstString` -- which seems like a type that is intended to own its backing buffer (like String does). This could easily lead to bugs.

I am comfortable with _specifically_ allowing string _literals_ to implicitly convert to `ConstString`. We could either do that by somehow making the `_kj` suffix magically able to convert to `ConstString` or, if that's too hard, by introducing a new `_kjc` suffix.

If we really need to give people a way to wrap StringPtrs that aren't literals, I would extend the `attach()` family of methods and have a `ConstString StringPtr::attach(...)`, where if you pass zero arguments, then it uses NullArrayDisposer (without allocating anything).

c++/src/kj/string.h
```diff
@@ -296,6 +306,105 @@ class String {
   Array<char> content;
 };
 
+// =======================================================================================
+// ConstString -- A NUL-terminated Array<const char> containing UTF-8 text.
```
kentonv, line 310: Let's describe this as:

> Same as `String`, but the backing buffer is const. This has the useful property that it can reference a string literal without allocating a copy. Any `String` can also convert (by move) to `ConstString`, transferring ownership of the buffer.

c++/src/kj/string.h
ObsidianMinor: > I am comfortable with *specifically* allowing string literals to implicitly convert to ConstString. We could either do that by somehow making the _kj suffix magically able to convert to ConstString

I'm not experienced enough in C++ to know what this magic might look like

> or, if that's too hard, by introducing a new _kjc suffix.

Couple of weird things with a specific suffix.

1. The operator can't be a `constexpr`, so it can't be used in other `constexpr`s
2. It can't be a `constexpr` because `kj::ConstString` isn't a literal type (because `kj::Array` isn't)

kentonv: How about this:

* Define a new subclass of StringPtr called LiteralStringPtr.
* Literals with the `_kj` suffix now evaluate to `LiteralStringPtr`. Indeed, this should be the only way to create a `LiteralStringPtr`.
* `LiteralStringPtr` defines an `operator ConstString()`, so it can also implicitly cast to `ConstString`.

c++/src/kj/string.h
```diff
@@ -76,6 +79,8 @@ class StringPtr {
   inline StringPtr(const char* begin KJ_LIFETIMEBOUND, const char* end KJ_LIFETIMEBOUND): StringPtr(begin, end - begin) {}
   inline StringPtr(String&& value KJ_LIFETIMEBOUND) : StringPtr(value) {}
   inline StringPtr(const String& value KJ_LIFETIMEBOUND);
+  inline StringPtr(const ConstString& value KJ_LIFETIMEBOUND);
+  constexpr StringPtr(const LiteralStringPtr& ptr);
```
ObsidianMinor, line 83: Should this be defined as an operator on `LiteralStringPtr`? I don't know what difference exists between `operator T()` vs `T(const U&)`

c++/src/kj/string.h
ObsidianMinor: This messes up some existing code, like any code that expects a `Promise<StringPtr>` or possibly other generic types (`Promise<LiteralStringPtr>` won't automagically cast to a `Promise<StringPtr>`)

ObsidianMinor: Perhaps we could define another type called `LiteralStringConst` with the `_kjc` suffix?

c++/src/kj/string.h
kentonv: You don't actually need either one because LiteralStringPtr is a subclass of StringPtr so the existing copy constructor for StringPtr will work fine.

c++/src/kj/string.h
```diff
@@ -162,12 +167,17 @@ class StringPtr {
   Maybe<T> tryParseAs() const;
   // Same as parseAs, but rather than throwing an exception we return NULL.
 
+  template <typename... Attachments>
+  ConstString attach(Attachments&&... attachments) const KJ_WARN_UNUSED_RESULT;
+  ConstString attach() const KJ_WARN_UNUSED_RESULT;
+  // Like ArrayPtr<T>::attach(), but instead promotes a StringPtr into a ConstString. Generally the
+  // attachment should be an opject that somehow owns the String that the StringPtr is pointing at.
```
kentonv, line 174:
```suggestion
  // attachment should be an object that somehow owns the String that the StringPtr is pointing at.
```

c++/src/kj/string.h
```diff
@@ -600,40 +725,72 @@ inline ArrayPtr<const char> StringPtr::slice(size_t start, size_t end) const {
   return content.slice(start, end);
 }
 
+inline LiteralStringConst::operator ConstString() const {
+  return ConstString(cStr(), size(), NullArrayDisposer::instance);
```
kentonv, line 729: This is nitpicky but `cStr()` is really meant to mean "Give me a null-terminated C string, possibly truncating the string if it contains NUL characters." What you want here is "Give me a pointer to the underlying buffer." The correct method for that is `begin()`.

Admittedly, these methods have the same underlying implementation.

c++/src/kj/string.h
```diff
@@ -600,40 +725,72 @@ inline ArrayPtr<const char> StringPtr::slice(size_t start, size_t end) const {
   return content.slice(start, end);
 }
 
+inline LiteralStringConst::operator ConstString() const {
+  return ConstString(begin(), size(), NullArrayDisposer::instance);
+}
+
+inline ConstString StringPtr::attach() const {
+  // This is meant as a roundabout way to make a ConstString from a StringPtr
+  return ConstString(cStr(), size(), NullArrayDisposer::instance);
```
kentonv, line 734: You missed a spot.

kentonv: I do see that my own code elsewhere does the same thing though, so maybe I'm being too nitpicky.

PR #1684, opened 2023-05-25
Exposes CidrRange as a public API.

dom96:
Need this upstream.


kentonv:
Can you give me some background on what you need this for?

Can't say I'm super-excited about this being a public part of KJ. If it must be, we should probably put it into a separate hedaer, async-io.h is too big and is included by just about everyone.

dom96:
Sent you a link in chat. Shall I create a dedicated header for it or is there a better existing place for it?

kentonv:
Assuming we decide we actually want to go this route, I guess I'd suggest a `cidr.h` header.

PR #1685, opened 2023-05-30
Pausable async stream fixes

dom96:
Tested upstream.

Refactoring and change to fix some fatal crashes.

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -5460,7 +5460,12 @@ kj::Promise<size_t> PausableReadAsyncIoStream::tryRead(
 
 kj::Promise<size_t> PausableReadAsyncIoStream::tryReadImpl(
     void* buffer, size_t minBytes, size_t maxBytes) {
-  return inner->tryRead(buffer, minBytes, maxBytes).attach(trackRead());
+  // Hack: evalNow used here because `newAdaptedPromise` has a bug. We may need to change
+  // `PromiseDisposer::alloc` to not be `noexcept` but in order to do so we'll need to benchmark
+  // its performance.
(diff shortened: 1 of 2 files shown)


c++/src/kj/compat/http.c++
```diff
@@ -5401,6 +5453,96 @@ _::Deferred<kj::Function<void()>> PausableReadAsyncIoStream::trackWrite() {
...
+kj::Promise<size_t> PausableReadAsyncIoStream::tryRead(
+    void* buffer, size_t minBytes, size_t maxBytes) {
+  return kj::newAdaptedPromise<size_t, PausableRead>(*this, buffer, minBytes, maxBytes);
+}
+
+kj::Promise<size_t> PausableReadAsyncIoStream::tryReadImpl(
+    void* buffer, size_t minBytes, size_t maxBytes) {
+  return kj::evalNow([&]() -> kj::Promise<size_t> {
+    return inner->tryRead(buffer, minBytes, maxBytes).attach(trackRead());
+  });
```
jasnell, line 5465: Deferring the `trackRead()` into the `kj::evalNow()` might be mildly problematic here? Specifically, it would allow scheduling multiple concurrent `tryReadImpl()` calls whereas currently those would fail synchronously. Not sure if it actually matters tho.

```suggestion
  return kj::evalNow([&, track=trackRead()]() -> kj::Promise<size_t> {
    return inner->tryRead(buffer, minBytes, maxBytes).attach(kj::mv(track));
  });
```

harrishancock: `evalNow()` executes its callback synchronously, so I don't think there's any significant difference in the code. You might be thinking of `evalLater()`?

harrishancock (approved):

PR #1688, opened 2023-06-01
capnp rpc: do not report shutdown error after read exception

mikea:
Read failure probably means broken transport. Do not report shutdown exception in such case.

### DIFF
--- a/c++/src/capnp/rpc.c++
+++ b/c++/src/capnp/rpc.c++
@@ -503,18 +503,23 @@ public:
     auto shutdownPromise = dyingConnection->shutdown()
         .attach(kj::mv(dyingConnection))
         .then([]() -> kj::Promise<void> { return kj::READY_NOW; },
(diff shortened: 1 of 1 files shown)


c++/src/capnp/rpc.c++
```diff
@@ -2766,6 +2774,9 @@ private:
         tasks.add(KJ_EXCEPTION(DISCONNECTED, "Peer disconnected."));
         return false;
       }
+    }, [this](kj::Exception&& exception) {
+      receiveIncomingMessageError = true;
+      return kj::mv(exception);
```
kentonv, line 2779: I think this is coercing the exception value to `bool` rather than propagating it, because the first callback passed to `then()` returns `bool` -- the second callback's return value always gets coerced to the same type as the first callback's. (Though, I'm surprised that doesn't lead to a compiler error, exceptions really shouldn't be convertible to `bool`.)

Try this instead:

```
kj::throwRecoverableException(kj::mv(exception));
return false;
```

kentonv (approved):

PR #1695, opened 2023-06-07
Fix compressed websocket fragment bug

MellowYarker:
We failed to allocate enough space for the message if it was fragmented because subsequent frames don't set the RSV1 (compression) bit.

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -2563,7 +2563,7 @@ public:
     kj::Maybe<size_t> originalMaxSize; // maxSize from first `receive()` call
     if (isFin) {
       size_t amountToAllocate;
-      if (recvHeader.isCompressed()) {
+      if (recvHeader.isCompressed() || fragmentCompressed) {
         // Add 4 since we append 0x00 0x00 0xFF 0xFF to the tail of the payload.
         // See: https://datatracker.ietf.org/doc/html/rfc7692#section-7.2.2
         amountToAllocate = payloadLen + 4;
@@ -2588,6 +2588,7 @@ public:
 
         fragments.clear();
         fragmentOpcode = 0;
+        fragmentCompressed = false;
       } else {
         // Single-frame message.
         message = kj::heapArray<byte>(amountToAllocate);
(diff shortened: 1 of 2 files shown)


c++/src/kj/compat/http.c++
```diff
@@ -2603,6 +2603,7 @@ public:
       if (fragments.empty()) {
         // This is the first fragment, so set the opcode.
         fragmentOpcode = opcode;
+        fragmentCompressed = recvHeader.isCompressed();
```
jclee, line 2606: Do you also need to reset `fragmentCompressed` here?:
```
        fragments.clear();
        fragmentOpcode = 0;
```

MellowYarker: Nice catch, yes it should reset. 

Practically speaking, I don't expect to see a situation where a connection uses compression for some messages and no compression for others. The chromium source code has a comment in it that suggests they could "predict" whether it would be worth it to compress some data, but IIRC it's been left as an unimplemented class. I haven't come across any libraries that let you intertwine (un)compressed messages either.

Very confused why the linux-musl test is failing, is there a way to get a stack trace?

jclee: Might be the exception shown earlier in the output? :
```
kj/compat/http.c++:2447: failed: expected compressionConfig == nullptr [(can't stringify) == nullptr]; WebSocket compression is only supported if KJ is compiled with Zlib.
```

Don't know much about the musl build, but perhaps it lacks zlib?  If so, might need to wrap the new test with `#if KJ_HAS_ZLIB` .

MellowYarker: Ahhhh... totally slipped past me, thanks Joe, that must be it.

jclee: Yeah, it doesn't help that the output appears interleaved with the earlier unrelated test results.  I assume that's an artifact of output buffering or github CI or something.

(FWIW, in reference to the earlier question, it's possible to dump stack trace addresses by adding a line like `KJ_DBG(kj::getStackTrace());` . But I don't know if that would necessarily have been useful here, since I'm not sure if there's an easy way set up to symbolize stack traces in github CI output.)

harrishancock (approved):

PR #1696, opened 2023-06-12
Revert "capnp rpc: do not report shutdown error after read exception

bcaimano:
There is suspicious behavior with tcp accept queues occurring since this was merged. Since this was only intended to deal with logging, let's back it out for now.

This reverts commit bb6c6ec477f6e7a63800e89dec559c63b5090574.


kentonv:
Hard to see how this could explain it but might as well try.

bcaimano:
> Hard to see how this could explain it but might as well try.

Yeah, agreed. But as said, might as well try.

PR #1698, opened 2023-06-18
Clarify that retroactive unionization is not forward-compatible.

### LINKED ISSUE #1694: putting a field into a new union is not forward compatible
As per https://github.com/capnproto/capnproto/issues/1424 I understand that https://capnproto.org/language.html#evolving-your-protocol states schema changes which are intended to be both backward and forward compatible.

I've found a problem with the following point:
* "A field can be moved into a group or a union, as long as the group/union and all other fields within it are new."

For example, if you evolve schema from
```
struct B1 {
  x @0: UInt64;
}
```
to
```
struct B2 {
  union {
    x @0: UInt64;
    y @1: UInt32;
  }
}
```
And you encode a `B2` message with `y` set, it will silently decode to `B1` with garbage `x` value.

kentonv:
Fixes #1694.

### DIFF
--- a/doc/language.md
+++ b/doc/language.md
@@ -738,7 +738,15 @@ without changing the [canonical](encoding.html#canonicalization) encoding of a m
 
 * A field can be moved into a group or a union, as long as the group/union and all other fields
   within it are new.  In other words, a field can be replaced with a group or union containing an
-  equivalent field and some new fields.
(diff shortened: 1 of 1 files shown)


kentonv:
Doc-only change, merging and deploying.

PR #1699, opened 2023-06-19
Fix incorrect schema construction when using a list-of-enum as a type parameter.

kentonv:
Fixes #1678


kentonv:
Note this is a bug in the RawSchema objects set up by the code generator to embed the schemas in the generated code. The encoded schemas are fine, and work correctly when loaded dynamically -- it's only the schema structures embedded in generated code that were broken.

harrishancock (approved):

PR #1700, opened 2023-06-19
Add RevocableServer, which makes it easy to RPC-wrap unowned objects.

kentonv:
If you have an object you don't own which has some well-defined lifetime, and you want to export it over RPC, until now you had to contend with the fact that the RPC peer could hold on to the capability until long past the underlying object's lifetime, and keep sending RPCs to it. This has been the source of a few UAF bugs and some extremely ugly work-arounds over the years.

RevocableServer is a simple and lightweight way to solve this problem.

As a demo, this PR uses RevocableContext to simplify ConnectClientRequestContextImpl in http-over-capnp. /cc @dom96

### DIFF
--- a/c++/src/capnp/capability.h
+++ b/c++/src/capnp/capability.h
@@ -101,6 +101,8 @@ class RequestHook;
 class ResponseHook;
 class PipelineHook;
 class ClientHook;
+template <typename T>
+class RevocableServer;
 
 template <typename Params, typename Results>
 class Request: public Params::Builder {
@@ -294,6 +296,9 @@ private:
   kj::Own<ClientHook> hook;
 
   static kj::Own<ClientHook> makeLocalClient(kj::Own<Capability::Server>&& server);
+  static kj::Own<ClientHook> makeRevocableLocalClient(Capability::Server& server);
+  static void revokeLocalClient(ClientHook& hook);
+  static void revokeLocalClient(ClientHook& hook, kj::Exception&& reason);
 
   template <typename, Kind>
   friend struct _::PointerHelpers;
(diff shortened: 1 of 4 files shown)


kentonv:
It appears the windows build has independently bitrotted. The failure is not caused by this change.

c++/src/capnp/capability.h
```diff
@@ -552,6 +560,39 @@ class Capability::Server {
...
+public:
+  RevocableServer(typename T::Server& server);
+  RevocableServer(RevocableServer&&) = default;
+  RevocableServer& operator=(RevocableServer&&) = default;
+  ~RevocableServer() noexcept(false);
+  KJ_DISALLOW_COPY(RevocableServer);
+
+  typename T::Client getClient();
+
+  void revoke();
```
bcaimano, line 587: I'm very excited for this work! That said, can I convince you to remove `revoke()` without a message? Default generated kj exceptions have proven difficult to diagnose, especially with remote stacks.

kentonv: I'm expecting that the common case will be to revoke implicitly when the RevocableServer goes out-of-scope. Unfortunately there isn't really an opportunity to force the caller to provide a message there.

bcaimano: I guess I can't argue after I made that change to kj::defer, huh? :sweat_smile: Perhaps that implies that there should be a default constructor as nullable type? Otherwise I'd expect people to just wrap it in a maybe. (And the movability does imply nullability)

kentonv: Sorry, I don't follow. How does that help improve the error message?

harrishancock (approved):

c++/src/capnp/capability.h
bcaimano: It does not. Though perhaps we could provide a default error to the constructor? I was just commenting that the type was in a weird spot where it's a movable nullable (if we expect implicit behavior as common) but it's missing semantics.

PR #1701, opened 2023-06-22
Improve WebSocket protocol error handling.

smerritt:
Currently, there's no good way to distinguish between WebSocket errors caused by the peer sending invalid data and WebSocket errors from other causes: they all result in FAILED-type exceptions.

This commit lets WebSocket users define a custom error handler for peer-caused errors. It only allows for customizing the thrown exception.

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -2542,7 +2542,7 @@ public:
     auto& recvHeader = *reinterpret_cast<Header*>(recvData.begin());
     if (recvHeader.hasRsv2or3()) {
       throw errorHandler.handleWebSocketProtocolError({
-        1002, "Protocol error", "Received frame had RSV bits 2 or 3 set", nullptr,
+        1002, "Received frame had RSV bits 2 or 3 set",
       });
     }
 
@@ -2551,8 +2551,7 @@ public:
     size_t payloadLen = recvHeader.getPayloadLen();
     if (payloadLen > maxSize) {
       throw errorHandler.handleWebSocketProtocolError({
-        1009, "WebSocket message is too large",
-        kj::str("Message is too large: ", payloadLen, " > ", maxSize), nullptr
+        1009, kj::str("Message is too large: ", payloadLen, " > ", maxSize)
       });
     }
 
@@ -2561,7 +2560,7 @@ public:
     if (opcode == OPCODE_CONTINUATION) {
       if (fragments.empty()) {
         throw errorHandler.handleWebSocketProtocolError({
-          1002, "Protocol error", "Unexpected continuation frame", nullptr
+          1002, "Unexpected continuation frame"
         });
       }
 
@@ -2569,7 +2568,7 @@ public:
     } else if (isData) {
       if (!fragments.empty()) {
         throw errorHandler.handleWebSocketProtocolError({
-          1002, "Protocol error", "Missing continuation frame", nullptr
+          1002, "Missing continuation frame"
         });
       }
     }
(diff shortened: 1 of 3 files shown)


c++/src/kj/compat/http.h
```diff
@@ -991,6 +991,19 @@ struct HttpClientSettings {
   // A reference to a TLS context that will be used when tlsStarter is invoked.
 };
 
+class WebSocketErrorHandler {
+public:
+  virtual kj::Exception handleWebSocketProtocolError(HttpHeaders::ProtocolError protocolError);
```
kentonv, line 996: Is HttpHeaders::ProtocolError the best f it for WebSocket errors? It looks like you're always passing the same values for all fields except `description`, so maybe just passing a string here would make sense?

kentonv (commented):
Generally looks great, just some bikeshedding comments...

c++/src/kj/compat/http.h
smerritt: It might not be. I only need statusCode (it's usually 1002, but is 1009 at one call site) and description, so I could define a WebSocket-specific one with just those two fields if you prefer.

c++/src/kj/compat/http.c++
```diff
@@ -3196,6 +3220,7 @@ private:
   kj::Own<kj::AsyncIoStream> stream;
   kj::Maybe<EntropySource&> maskKeyGenerator;
   kj::Maybe<CompressionParameters> compressionConfig;
+  kj::Maybe<WebSocketErrorHandler&> errorHandler;
```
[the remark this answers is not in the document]
smerritt: Sounds good to me. Done.

c++/src/kj/compat/http.h
kentonv: Yeah I think it would make sense.

smerritt: Done in commit d817243.

MellowYarker:
https://github.com/capnproto/capnproto/pull/1701/files#diff-09d28b46bd98866f20dd7241e02c70338a76fdb59a483ed02d4c16bbe6bcc3a1R3186-R3187 I suspect this will also need to be thrown in the same way as [the regular message](https://github.com/capnproto/capnproto/pull/1701/files#diff-09d28b46bd98866f20dd7241e02c70338a76fdb59a483ed02d4c16bbe6bcc3a1R2553-R2554), since we'll be enabling compression by default soon (AFAICT that implies all websocket connections from browsers to workers deployed after the compat date is enabled will be hitting this code).

c++/src/kj/compat/http.c++
```diff
@@ -5394,6 +5421,12 @@ HttpClient::WebSocketResponse HttpClientErrorHandler::handleWebSocketProtocolErr
   };
 }
 
+kj::Exception WebSocketErrorHandler::handleWebSocketProtocolError(
+      WebSocket::ProtocolError protocolError) {
+  return KJ_EXCEPTION(FAILED, kj::str("code=", protocolError.statusCode,
+                                        ": ", protocolError.description));
```
kentonv, line 5427: The explicit formatting of the error here actually confuses code which tries to parse KJ exception messages. It would be better to change this to:

```
  return KJ_EXCEPTION(FAILED, protocolError.statusCode, protocolError.description));
```

kentonv: Note that KJ will format this like:

> failed: WebSocket protocol error; protocolError.statusCode = 1009; protocolError.description = Message too large: 1000 > 900

kentonv: In Workers' sentry, this will mean all WebSocket protocol errors get grouped together but that's probably OK.

PR #1702, opened 2023-06-30
rpc.capnp: add a note clarifying the 4-way race

zenhack:
I had for some time been under the mistaken impression that this could only occur in a scenario involving >2 *distinct* vats, and thus assumed it did not concern level 1 implementations. This patch adds some more comments calling out the fact that this is not the case.


kentonv:
Indeed! This has been tricky to get right even at level 1.

PR #1705, opened 2023-07-05
Fix non-monotonic time on Intel Mac machines

fabiorossetto:
It has been observed that in rare circumstances, clock_gettime may return a non monotonic timestamp even when using CLOCK_MONOTONIC. This triggers the assertion in "advanceTo", breaking applications built on top of the KJ library.
This commit removes the assertion for those platforms on which the problem has been observed (only Mac with x86 processor).


fabiorossetto:
@tuxu @tonyp-zi FYI

PR #1706, opened 2023-07-06
Add kj::rawPtr()

jp4a50:
It'd be nice to have a convenient way of creating `kj::Own`s for resources allocated by C APIs like `getaddrinfo`.

This is more of a suggestion. If there is already a way to do this or a suggested alternative approach, I'm all ears.

### DIFF
--- a/c++/src/kj/memory.h
+++ b/c++/src/kj/memory.h
@@ -571,6 +571,19 @@ template <typename T>
 const HeapDisposer<T> HeapDisposer<T>::instance = HeapDisposer<T>();
 #endif
 
+template <typename T, void(*F)(T*)>
+class CustomDisposer: public Disposer {
+public:
+  static const CustomDisposer instance;
+
+  void disposeImpl(void* pointer) const override {
+    (*F)(reinterpret_cast<T*>(pointer));
+  }
+};
+
+template <typename T, void(*F)(T*)>
+const CustomDisposer<T, F> CustomDisposer<T, F>::instance = CustomDisposer<T, F>();
+
 }  // namespace _ (private)
 
 template <typename T, typename... Params>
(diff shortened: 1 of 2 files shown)


c++/src/kj/memory.h
```diff
@@ -594,6 +607,13 @@ Own<Decay<T>> heap(T&& orig) {
   return Own<T2>(new T2(kj::fwd<T>(orig)), _::HeapDisposer<T2>::instance);
 }
 
+template <typename T, void(*F)(T*)>
```
kentonv, line 610: Hmm, I guess using these template parameters implies that callers will always have to redundantly specify T, like:

    auto own = rawPtr<MyType, &freeMyType>(ptr);

But specifying `MyType` here is redundant, since it should be deduced from `ptr`. Ideally the syntax would be more like:

    auto own = rawPtr<&freeMyType>(ptr);

But I'm not sure how to declare a template that would accept this. Is it possible? I know that the following would be possible, but it's weird:

    // too weird
    auto own = rawPtr(ptr).withDisposer<&freeMyType>();

(I've run into this difficulty a few times in the past, and I swear I came across a solution once, but I've forgotten what it was...)

c++/src/kj/memory.h
```diff
@@ -594,6 +607,13 @@ Own<Decay<T>> heap(T&& orig) {
...
+Own<T> rawPtr(T* ptr) {
```
kentonv, line 611: Bikeshedding, but `rawPtr` doesn't quite feel like the right name to me. The important bit is that we're deriving a disposer from a function. Something like `disposedWith()`? I dunno.

kentonv (commented):
This pattern has come up a bunch when working with OpenSSL. I think it makes sense to add to the core library, yeah.

c++/src/kj/memory.h
jp4a50: > auto own = rawPtr<&freeMyType>(ptr);

Looks like we can achieve this (although `&` not necessary due to decay to function pointer) using:

```
template <auto F, typename T>
Own<T> rawPtr(T* ptr) {
...
```

and keeping the implementation the same, which lets the `CustomDisposer` enforce the constraints on F.

Unfortunately, that's C++17 only and it looks like capnproto is still on C++14?

kentonv: Oh nice, why didn't I know about that?

I'm happy to increase the requirement to C++17, or maybe even C++20 (for coroutines!), but probably not in this PR.

For now, you can wrap this in:

`#if __cplusplus > 201402L`

(We have some other features that are C++17-only like this, such as `kj::arr()` to construct a fixed-size array.)

c++/src/kj/memory.h
jp4a50: Yeah I intended to revisit after getting initial feedback. I don't think it's right either.

I can't think of anything better than `disposeWith` or `disposedWith`. I think I slightly prefer `disposeWith`.

c++/src/kj/memory.h
jp4a50: Done.

kentonv:
Thanks! Care to update the OpenSSL-using code in workerd to use this? :)

jp4a50:
Sure, will take a look!

PR #1707, opened 2023-07-08
Fix stack traces through coroutines

kentonv:
When an exception propagated through a coroutine, the trace simply skipped the coroutine altogether, rather than reporting the locations of the co_awaits.

I've tested this manually but it's hard to write a unit test since KJ currently cannot symbolize stacks without outside help.

### DIFF
--- a/c++/src/kj/async-inl.h
+++ b/c++/src/kj/async-inl.h
@@ -31,6 +31,10 @@
 #include "async.h"  // help IDE parse this file
 #endif
 
+#if _MSC_VER && KJ_HAS_COROUTINE
+#include <intrin.h>
+#endif
+
 KJ_BEGIN_HEADER
 
 #include <kj/list.h>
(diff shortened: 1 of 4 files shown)


c++/src/kj/async-inl.h
```diff
@@ -35,10 +35,10 @@
 #include <intrin.h>
 #endif
 
-KJ_BEGIN_HEADER
-
 #include <kj/list.h>
 
+KJ_BEGIN_HEADER
```
harrishancock, line 40: Since async-inl.h is only ever included from async.h, I wonder if all of async-inl.h's includes are covered by async.h's KJ_BEGIN_HEADER anyway...

harrishancock (approved):

c++/src/kj/async-inl.h
kentonv: Oh, I suppose the include of async-inl.h in async.h should really be after KJ_END_HEADER. Oh well.

PR #1708, opened 2023-07-09
Fix tls-test on Windows

kentonv:
For some reason our CI seems to declare (in the header) that it is v3, but the text it produces looks like v1. Maybe the header is a different version than the library? I don't actually care. Let's match both messages.


c++/src/kj/compat/tls-test.c++
```diff
@@ -702,25 +703,36 @@ void expectInvalidCert(kj::StringPtr hostname, TlsCertificate cert, kj::StringPt
...
-#else
-#define SSL_MESSAGE_DIFFERENT_IN_V3(v11, v30) v11
-#endif
+    KJ_IF_MAYBE(a, altMessage) {
+      if (kj::_::hasSubstring(e.getDescription(), *a)) {
+        return;
+      }
+    }
+
+    KJ_FAIL_EXPECT("exception didn't contain expected messag", message,
```
harrishancock, line 719:
```suggestion
    KJ_FAIL_EXPECT("exception didn't contain expected message", message,
```

harrishancock (approved):

PR #1711, opened 2023-07-10
Minor fixes for older Darwin and PowerPC

barracuda156:
1. Fix for a static assert failure for `arenaSpace`: Darwin ppc needs an additional padding too.
2. Older macOS also does not have `SOL_LOCAL`.


kentonv:
Thanks!

PR #1717, opened 2023-07-15
KJ_IF_MAYBE's variable can no longer be implicitly converted to a raw pointer

harrishancock:
This converts a very common mistake we have observed over the years into a compile-time error. See commit messages for further details.

### DIFF
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -1131,8 +1131,11 @@ public:
   inline const T&& operator*() const && { return kj::mv(value); }
   inline T* operator->() { return &value; }
   inline const T* operator->() const { return &value; }
-  inline operator T*() { return isSet ? &value : nullptr; }
-  inline operator const T*() const { return isSet ? &value : nullptr; }
+
+  inline explicit operator T*() { return isSet ? &value : nullptr; }
+  inline explicit operator const T*() const { return isSet ? &value : nullptr; }
+
+  inline explicit operator bool() const { return isSet; }
 
   template <typename... Params>
   inline T& emplace(Params&&... params) {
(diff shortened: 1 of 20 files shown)


c++/src/kj/common.h
```diff
@@ -1295,19 +1298,65 @@ class NullableValue {
...
+
+  T& operator*() { return *ref; }
+  T* operator->() { return ref; }
+  const T& operator*() const { return *ref; }
+  const T* operator->() const { return ref; }
+  // ExplicitPtr supports pointer-like dereference/indirection operations.
+
+  explicit operator T*() { return ref; }
+  explicit operator const T*() const { return ref; }
+  // An ExplicitPtr may be explicitly converted back to its raw pointer.
```
kentonv, line 1336: Is this actually useful, vs. `*&`?

c++/src/kj/common.h
```diff
@@ -1295,19 +1298,65 @@ class NullableValue {
...
+
+  explicit operator bool() const { return ref != nullptr; }
+  // An ExplicitPtr may be contextually converted to bool for use in `if` conditions.
```
kentonv, line 1339: I guess this is needed only for the KJ_IF_MAYBE macro itself to function correctly?

If we're switching to C++17 anyway, can we instead rely on init-statement syntax, like:

```
if (ExplicitPtr<T> foo = blah(); foo.isNotNull()) {
```

?

kentonv: Oh geez I finally figured out how to make the identifier a reference, if we can assume C++17:

```
#define KJ_IF_MAYBE2(name, exp) \
  if (bool b = true; auto _##name = ::kj::_::readMaybe(exp)) \
    for (auto& name = *_##name; b; b = false)
```

If the compiler optimizes this reasonably it might be worth it?

kentonv: Holy shit https://godbolt.org/z/6x9zvsxqT

c++/src/kj/common.h
harrishancock: It might not be -- I think I put it in for symmetry with the other two types that `readMaybe()` returns. But I suppose those other two types have other reasons for being convertible. I'll remove.

c++/src/kj/common.h
harrishancock: > I guess this is needed only for the KJ_IF_MAYBE macro itself to function correctly?

Yes.

As for the reference hack, that's awesome! :) It looks like that even works correctly with `KJ_IF_MAYBE() { ... } else { ... }` constructions.

Personally, I'd be willing to replace every `KJ_IF_MAYBE` with the reference-declaring version (after upgrading to C++17/20, of course), since it's obviously better.

harrishancock: Re: naming, how about `KJ_IF_NONNULL(var, maybe) { ... }`?

kentonv: I guess I could accept that though I also think that appending `2` to something, while aesthetically displeasing, has the nice effect of communicating that this is a new version of the thing, so people don't get confused about which one they should be using.

harrishancock:
Closed this PR, since we have a better path forward. I'll be opening a PR to upgrade capnproto to C++20 in the coming days, then will add the new and improved reference-declaring macro.

PR #1718, opened 2023-07-15
Fix kj-tls not working in half-duplex mode.

kentonv:
When EOF was received in the read direction, the write direction would also be treated as disconnected, since they shared the same `disconnected` flag.

I think this flag isn't actually needed at all. It looks like it was intended as an optimization of sorts, to avoid calling OpenSSL if we already knew we were at EOF. But the optimization is wrong, since it doesn't distinguish between read and write directions. So I just removed it.

### DIFF
--- a/c++/src/kj/compat/tls-test.c++
+++ b/c++/src/kj/compat/tls-test.c++
@@ -489,6 +489,38 @@ KJ_TEST("TLS basics") {
   }
 }
 
+KJ_TEST("TLS half-duplex") {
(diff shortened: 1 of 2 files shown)


kentonv:
I think Harris is on vacation. This is impacting SRS so added @a-robinson as a reviewer.

a-robinson (approved):

a-robinson:
Looks reasonable, nice find.

PR #1721, opened 2023-07-19
Fix release tests

kentonv:
It's time for some CI-driven-development, oh yeah...

### DIFF
--- a/c++/src/kj/async-inl.h
+++ b/c++/src/kj/async-inl.h
@@ -229,15 +229,14 @@ class PromiseArenaMember {
 
 public:
   virtual void destroy() = 0;
-  // Tell the node to run its destructor.
+  // Destroys and frees the node.
   //
-  // If the node was allocated using allocPromise(), then it should only run the destructor, it
-  // should not free the backing memory; PromiseDisposer will do that. If it was allocated some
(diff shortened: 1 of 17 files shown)


kentonv:
They passed!

c++/src/kj/async-inl.h
```diff
@@ -413,6 +418,31 @@ static kj::Own<T, PromiseDisposer> allocPromise(Params&&... params) {
...
+    // The object will have been allocated separately on the heap.
+    return delete ptr;
+  }
+};
+
+template <typename T>
+static void freePromise(T* ptr) {
+  // Free a PromiseNode originally allocated using `allocPromise<T>()`. The implementation of
+  // PromiseNode::destroy() must call this for any type that is allocated using allocPromise().
+  FreePromiseNode<T>::free(ptr);
```
harrishancock, line 443: Nit: I'd find `if constexpr` a bit more readable than the `FreePromiseNode` helper, but maybe not all our current compilers support it.

harrishancock (approved):

c++/src/kj/async-inl.h
kentonv: `if constexpr` is C++17, we're still on C++14 until we get this release out.

PR #1726, opened 2023-07-25
Expose Win32 handle in AsyncIoStream interface

fabiorossetto:
This commit adds a function to AsyncIoStream that helps in the interoperability of the class with other libraries. It allows to reclaim ownership of a file descriptor from an AsyncIoStream. This can allow for example to use KJ ConnectionReceiver, but then use the file descriptors natively, or pass them to another library.

### DIFF
--- a/c++/src/kj/async-io-win32.c++
+++ b/c++/src/kj/async-io-win32.c++
@@ -352,6 +352,10 @@ public:
     *length = socklen;
   }
 
+  Maybe<void*> getWin32Handle() const {
+    return reinterpret_cast<void*>(fd);
+  }
+
 private:
   Own<Win32EventPort::IoObserver> observer;
 
--- a/c++/src/kj/async-io.h
+++ b/c++/src/kj/async-io.h
@@ -169,6 +169,10 @@ public:
   virtual kj::Maybe<int> getFd() const { return nullptr; }
   // Get the underlying Unix file descriptor, if any. Returns nullptr if this object actually
   // isn't wrapping a file descriptor.
+
+  virtual Maybe<void*> getWin32Handle() const { return nullptr; }
+  // Get the underlying Win32 HANDLE, if any. Returns nullptr if this object actually isn't
+  // wrapping a handle.
 };
 
 Promise<uint64_t> unoptimizedPumpTo(


kentonv:
Looks like tests are failing because you need to tear down the FdObserver before releasing the FD.

That said, have you considered using `dup()` to make a duplicate of the FD instead? You could then destroy the `AsyncIoStream`, and use the duplicate. No need for changes to KJ.

fabiorossetto:
> That said, have you considered using dup() to make a duplicate of the FD instead? You could then destroy the AsyncIoStream, and use the duplicate. No need for changes to KJ.

Excellent suggestion! Yes that seems to work perfectly on Linux and Mac. However on Windows `getFd` is not implemented. So I cannot use dup right away... Is there a way to get the underlying SOCKET from AsyncIoStream in Windows?

kentonv:
On Windows, a SOCKET isn't really a file descriptor. It's a type of HANDLE. Windows implements file descriptors as a compatibility feature, but entirely as a shim in the C runtime; they aren't part of the underlying operating system API, which only uses HANDLEs. SOCKETs are not valid file descriptors in the eyes of the C runtime. In practice file descriptors on Windows are very limited and aren't typically used for networking.

I think the right approach here is to define a new method, `kj::Maybe<void*> getWin32Handle()`. (`HANDLE` is an alias for `void*`, so using `void*` here avoids the need to include the Windows headers.) This matches what we do in the KJ Filesystem APIs, which offer both `getFd()` and `getWin32Handle()` methods.

Unfortunately it's necessary for callers to handle Windows specially, but that's probably true regardless.

It looks like on Windows duplicating handles should be possible but is not quite as simple as on Unix... There's `DuplicateHandle()`, but the docs say not to use it on sockets. `WSADuplicateSocket()` looks like an alternative but the API is... weird.

fabiorossetto:
@kentonv I've changed the pull request to simply expose the Win32 handle. Would you be open to a PR that adds a `kj::Own<AsyncIoStream> duplicateSocketStream(const kj::Own<AsyncIoStream>&)` method to `LowLevelAsyncIoProvider`. I have to implement it anyway, might as well add it to the KJ library

kentonv:
Looks like this doesn't compile. You might have to `reinterpret_cast<HANDLE>(fd)` or something. My understanding is that `SOCKET` and `HANDLE` exist in the same namespace with the same numeric values, but `SOCKET` is defined as an integer type for improved compatibility with network code that uses file descriptors on other platforms.

> Would you be open to a PR that adds a kj::Own<AsyncIoStream> duplicateSocketStream(const kj::Own<AsyncIoStream>&) method to LowLevelAsyncIoProvider?

Hmm, does that actually solve your problem? You'd end up with a second stream that _also_ owns the underlying fd/handle. I thought what you wanted is to take ownership of the fd/handle. The use of dup() to accomplish this is a hack, but if we want to add something to KJ I think `releaseFd()` (or `releaseHandle()` on Windows) is actually what we want, it just has to actually fully tear down the stream before returning.

fabiorossetto:
@kentonv You are right. I will make a separate PR for a proposal for releaseFd. In the meantime, can we merge this PR?

PR #1727, opened 2023-07-25
Override the read-only HTTP headers only if specified

fabiorossetto:
This commit changes the logic around the "read-only" headers in the HttpClient. Instead of being always overwritten, they are overwritten only when HttpClient does have a value for them. If not, and if the user has provided a value, that value is used.

This change makes the HttpClient more customizable, without affecting the correctness of the HttpClient.

### DIFF
--- a/c++/src/kj/compat/http-test.c++
+++ b/c++/src/kj/compat/http-test.c++
@@ -184,6 +184,35 @@ KJ_TEST("HttpHeaders::parseRequest") {
       "\r\n");
 }
 
+KJ_TEST("HttpHeaders::serializeRequestNoOverride") {
+  HttpHeaderTable::Builder builder;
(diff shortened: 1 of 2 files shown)


fabiorossetto:
Hi Kenton,
no worries, and thank you for your detailed answer. I imagined that you would not approve of this MR. 

Just for context, what I'm trying to do is to make a protocol upgrade from HTTP to plain capnp. For this, I need to override the `Upgrade` header, which the kj library doesn't let me do.

kentonv:
@fabiorossetto I think it would make sense to extend KJ HTTP with a way to support arbitrary `Upgade`. Actually, it can probably be based on the same code added to support the `CONNECT` method -- except that instead of being signaled by a different HTTP method, it's signaled by the `Upgrade` header.

PR #1728, opened 2023-07-25
Improve error string for WebSocket protocol errors.

smerritt:
Most KJ errors are space-separated; "code=N" was atypical. This confused things that parse exception messages expect the usual format.


kentonv (approved):

kentonv:
To clarify, it's not the "code=N" that's the issue but rather using `kj::str()` to construct an exception message that is different every time (because it includes the message size, for message-to-large errors).

PR #1729, opened 2023-07-27
Add blog post announcing 1.0.

kentonv:


doc/_posts/2023-07-28-capnproto-1.0.md
```diff
@@ -0,0 +1,74 @@
...
+
+* **Breaking change:** Previously, servers could opt into allowing RPC cancellation by calling `context.allowCancellation()` after a call wa delivered. In 1.0, opting into cancellation is instead accomplished using an annotation on the schema (the `allowCancellation` annotation defined in `c++.capnp`). We made this change after observing that in practice, we almost always wanted to allow cancellation, but we almost always forgot to do so. The schema-level annotation can be set on a whole file at a time, which is easier not to forget. Moreover, the dynamic opt-in required a lot of bookkeeping that had a noticeable performance impact in practice; switching to the annotation provided a performance boost. For users that never used `context.allowCancellation()` in the first place, there's no need to change anything when upgrading to 1.0 – cancellation is still disallowed by default.
```
a-robinson, line 24:
```suggestion
* **Breaking change:** Previously, servers could opt into allowing RPC cancellation by calling `context.allowCancellation()` after a call was delivered. In 1.0, opting into cancellation is instead accomplished using an annotation on the schema (the `allowCancellation` annotation defined in `c++.capnp`). We made this change after observing that in practice, we almost always wanted to allow cancellation, but we almost always forgot to do so. The schema-level annotation can be set on a whole file at a time, which is easier not to forget. Moreover, the dynamic opt-in required a lot of bookkeeping that had a noticeable performance impact in practice; switching to the annotation provided a performance boost. For users that never used `context.allowCancellation()` in the first place, there's no need to change anything when upgrading to 1.0 – cancellation is still disallowed by default.
```

doc/_posts/2023-07-28-capnproto-1.0.md
```diff
@@ -0,0 +1,74 @@
...
+
+* KJ now uses `kqueue()` to handle asynchronous I/O on systems that have it (MacOS and BSD derivatives). KJ has historically always used `epoll` on Linux, but until now had used a slower `poll()`-based approach on other Unix-like platforms.
+
+* KJ's HTTP client and server implementations now supports the `CONNECT` method.
```
a-robinson, line 28:
```suggestion
* KJ's HTTP client and server implementations now support the `CONNECT` method.
```

a-robinson (approved):
Very nice :)

harrishancock (approved):

kentonv:
Github UI is lagging for some reason. Tweaks I force-pushed: https://github.com/capnproto/capnproto/compare/46811934831d16a8b3782abf478b9675a3885416..f3d45a85d61d7b0d70c54e8c6583f418a785edea

(Alex's suggestions, and also changing the v2 branch name to `v2`.)

PR #1730, opened 2023-07-28
Fix Clang 16 build

harrishancock:
This PR:
- fixes a couple compiler warnings
- disables a few `operator!=()` implementations in C++20 mode to get Clang 16 working
- and ports our Musl build to Clang 16 -- partly to prove that Clang 16 works, and partly because we will soon need to when we require coroutines (GCC ICEs on our coroutines).

### DIFF
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -1843,7 +1843,9 @@ public:
     }
     return true;
   }
+#if !__cpp_impl_three_way_comparison
   inline bool operator!=(const ArrayPtr& other) const { return !(*this == other); }
+#endif
 
   template <typename U>
   inline bool operator==(const ArrayPtr<U>& other) const {
@@ -1853,8 +1855,10 @@ public:
     }
     return true;
   }
+#if !__cpp_impl_three_way_comparison
   template <typename U>
   inline bool operator!=(const ArrayPtr<U>& other) const { return !(*this == other); }
+#endif
 
   template <typename... Attachments>
   Array<T> attach(Attachments&&... attachments) const KJ_WARN_UNUSED_RESULT;
--- a/c++/src/kj/string.h
+++ b/c++/src/kj/string.h
@@ -278,14 +278,18 @@ public:
   inline bool operator!=(decltype(nullptr)) const { return content.size() > 1; }
 
   inline bool operator==(const StringPtr& other) const { return StringPtr(*this) == other; }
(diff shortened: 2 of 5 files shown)


c++/src/kj/common.h
```diff
@@ -1843,7 +1843,9 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
     }
     return true;
   }
+#if !__cpp_impl_three_way_comparison
```
mikea, line 1846: TIL there are a lot of these macros: https://en.cppreference.com/w/cpp/feature_test

c++/src/kj/string.h
```diff
@@ -278,14 +278,18 @@ class String {
   inline bool operator!=(decltype(nullptr)) const { return content.size() > 1; }
 
   inline bool operator==(const StringPtr& other) const { return StringPtr(*this) == other; }
+#if !__cpp_impl_three_way_comparison
```
mikea, line 281: is the plan to drop these eventually? Maybe we need our own KJ_ alias so that it will be easier to do a cleanup?

c++/src/kj/string.h
```diff
@@ -278,14 +278,18 @@ class String {
...
   inline bool operator!=(const StringPtr& other) const { return StringPtr(*this) != other; }
+#endif
   inline bool operator< (const StringPtr& other) const { return StringPtr(*this) <  other; }
```
mikea, line 284: strictly speaking all these comparison have to be part of the previous ifdef?

mikea (approved):

c++/src/kj/string.h
harrishancock: Yep, I drop them in PR #1732. I considered trying to land that PR first, so I wouldn't add this stuff only to immediately remove it. But, I thought that might take longer to land.

c++/src/kj/string.h
harrishancock: Nope, we still need them -- the compiler will only synthesize "ordering" comparison operators if we define a spaceship operator -- which I think we should do, but figured I'd do the conservative thing for now.

I found this article helpful when learning about the new rules, by the way: https://brevzin.github.io/c++/2019/07/28/comparisons-cpp20/

harrishancock:
Switched to target the `v2` branch.

kentonv (approved):

kentonv:
Guess we'll also need to cherry-pick this into master.

PR #1732, opened 2023-07-28
Require C++20

harrishancock:
This PR updates capnproto to require C++20. Most notably, we now require coroutine support, forcing us to drop support for GCC due to a (hopefully temporary) compiler bug.

I did my best to update the autotools build to C++20, but I have no familiarity with m4, and I'm not confident I got it right.

I upgraded CMake to 3.16; we need at least 3.12 for the `cxx_std_20` feature, and 3.16 seems to be the oldest among Ubuntu and Debian LTS/stable editions.

Coroutine support was added to GCC 10, but it appears to have a bug (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=102051) blocking our use of them. Meanwhile, Clang 11 is the earliest version of Clang which can compile our code.

I tried to drop support for the Coroutines TS, to simplify our code a little, but this broke our MacOS build: I guess its version of XCode ships with LLVM 13, which is the version before the standard header `<coroutine>` was added to libc++.

I removed several hacks and features which were present only to support old versions of C++. In particular, I removed most of of the places where we check for `__cplusplus > someoldversion`. This had the interesting effect of exposing some code to MSVC for the very first time -- MSVC sets `__cplusplus` to sometime in 1997, a fact I always forget. This, combined with the requirement to build coroutines, necessitated some MSVC-related fiddling to get the Windows build working.

MSVC from VS2019 seems to contain a bug in C++20 mode that I didn't feel like chasing, so we now require VS2022.

I adjusted our Quick Tests CI runs as required by all of the foregoing points. Additionally, I increased the number of standard Ubuntu builds to 10: Clang 11-15, each with one run using `./super-test.sh quick` and one run using `bazel test //...`. We should probably reduce this, though.

Lastly, this PR introduces a single `co_await` keyword to kj-http, to show that it compiles in CI.


c++/src/kj/string.h
```diff
@@ -378,27 +358,18 @@ class ConstString {
...
   inline bool operator< (const String& other) const { return StringPtr(*this) <  StringPtr(other); }
   inline bool operator> (const String& other) const { return StringPtr(*this) >  StringPtr(other); }
   inline bool operator<=(const String& other) const { return StringPtr(*this) <= StringPtr(other); }
   inline bool operator>=(const String& other) const { return StringPtr(*this) >= StringPtr(other); }
 
   inline bool operator==(const ConstString& other) const { return StringPtr(*this) == StringPtr(other); }
-#if !__cpp_impl_three_way_comparison
-  inline bool operator!=(const ConstString& other) const { return StringPtr(*this) != StringPtr(other); }
-#endif
   inline bool operator< (const ConstString& other) const { return StringPtr(*this) <  StringPtr(other); }
```
mikea, line 373: should we have `<=>` operator for everything instead of these then?

c++/src/kj/compat/http.c++
```diff
@@ -7801,11 +7801,10 @@ kj::Promise<void> HttpServer::listenHttp(kj::ConnectionReceiver& port) {
 }
 
 kj::Promise<void> HttpServer::listenLoop(kj::ConnectionReceiver& port) {
-  return port.accept()
-      .then([this,&port](kj::Own<kj::AsyncIoStream>&& connection) -> kj::Promise<void> {
+  for (;;) {
```
mikea, line 7804: nice!

mikea (approved):
nice. can't comment much on chosen platforms, but the rest LG.

c++/src/kj/string.h
harrishancock: I definitely think so -- it'll reduce boilerplate across all of our codebases. I suspect there may be debate over it, so held off for now.

c++/src/kj/memory.h
```diff
@@ -574,15 +574,13 @@ const HeapDisposer<T> HeapDisposer<T>::instance = HeapDisposer<T>();
 template <typename T, void(*F)(T*)>
 class CustomDisposer: public Disposer {
 public:
-  static const CustomDisposer instance;
```
kentonv, line 577: Could this have been declared `constexpr` in-place, without un-nesting it? Not that I really care, more a curiosity.

.github/workflows/quick-test.yml
```diff
@@ -23,71 +25,73 @@ jobs:
...
-    steps:
-      - uses: actions/checkout@v2
-      - name: install dockcross
-        run: |
-            docker run ${{ matrix.docker-run-args }} --rm dockcross/${{ matrix.cross-compiler }} > ./dockcross
-            chmod +x ./dockcross
-      - name: super-test
-        run: |
-            ./dockcross ./super-test.sh quick g++
+  # ManyLinux:
```
kentonv, line 77: I'd be fine with deleting these commented builds, I suspect it'll be years before we try to support them again, if ever.

c++/src/kj/async-prelude.h
```diff
@@ -28,23 +28,17 @@
...
 #define KJ_COROUTINE_STD_NAMESPACE std
 #elif (__cpp_coroutines >= 201703L) && __has_include(<experimental/coroutine>)
 // Coroutines TS detected.
 #include <experimental/coroutine>
-#define KJ_HAS_COROUTINE 1
 #define KJ_COROUTINE_STD_NAMESPACE std::experimental
-#endif
-#endif
+#else
+#error "Cap'n Proto requires C++20 or Coroutines TS support"
```
kentonv, line 41: This reads a little weird, since we require C++20 either way. Is it meant to be like... "Cap'n Proto requires support for C++20 coroutines. Support for the coroutines TS will suffice."?

kentonv (approved):

c++/src/kj/memory.h
harrishancock: Sadly, no, because the type is still incomplete in the compiler's eyes at that point. I scratched my head a bit trying to think around it, then gave up.

.github/workflows/quick-test.yml
harrishancock: Agreed; I deleted the commented-out builds.

c++/src/kj/async-prelude.h
harrishancock: Fixed!

harrishancock:
There's also our Release Tests (.github/workflows/release-test.yml). I don't want to block this PR on it, but I guess we'll need to revise it for branch `v2`.

PR #1736, opened 2023-07-30
Support EV_TRIGGER as a fallback; disable KJ_USE_KQUEUE on old macOS

barracuda156:
@kentonv I ran tests, this works fine.

On 1.0.0 with this patch applied, 1041 test(s) passed, 4 test(s) failed.

### DIFF
--- a/c++/src/kj/async-unix.c++
+++ b/c++/src/kj/async-unix.c++
@@ -1065,7 +1065,11 @@ void UnixEventPort::captureChildExit() {
 void UnixEventPort::wake() const {
   // Trigger our user event.
   struct kevent event;
+#if defined(NOTE_TRIGGER)
   EV_SET(&event, 0, EVFILT_USER, 0, NOTE_TRIGGER, 0, nullptr);
+#elif defined(EV_TRIGGER)
+  EV_SET(&event, 0, EVFILT_USER, EV_TRIGGER, 0, 0, nullptr);
+#endif
   KJ_SYSCALL(kevent(kqueueFd, &event, 1, nullptr, 0, nullptr));
 }
 
--- a/c++/src/kj/async-unix.c++
+++ b/c++/src/kj/async-unix.c++
@@ -1069,6 +1069,8 @@ void UnixEventPort::wake() const {
   EV_SET(&event, 0, EVFILT_USER, 0, NOTE_TRIGGER, 0, nullptr);
 #elif defined(EV_TRIGGER)
   EV_SET(&event, 0, EVFILT_USER, EV_TRIGGER, 0, 0, nullptr);
+#else
+#error "neither NOTE_TRIGGER nor EV_TRIGGER is defined; we need at least one of these"
 #endif
   KJ_SYSCALL(kevent(kqueueFd, &event, 1, nullptr, 0, nullptr));
 }
--- a/c++/src/kj/async-unix.h
+++ b/c++/src/kj/async-unix.h
@@ -31,14 +31,20 @@
 #include <kj/io.h>
 #include <signal.h>
 
+#ifdef __APPLE__
+#include <AvailabilityMacros.h>
+#endif
+
 KJ_BEGIN_HEADER
 
 #if !defined(KJ_USE_EPOLL) && !defined(KJ_USE_KQUEUE)
 #if __linux__
 // Default to epoll on Linux.
 #define KJ_USE_EPOLL 1
-#elif __APPLE__ || __FreeBSD__ || __OpenBSD__ || __NetBSD__ || __DragonFly__
-// MacOS and BSDs prefer kqueue() for event notification.
+#elif (__APPLE__ && (MAC_OS_X_VERSION_MIN_REQUIRED > 1050)) \
+  || __FreeBSD__ || __OpenBSD__ || __NetBSD__ || __DragonFly__
+// MacOS and BSDs prefer kqueue() for event notification, however MacOS prior
+// to 10.6.x does not have NOTE_TRIGGER.
 #define KJ_USE_KQUEUE 1
 #endif
 #endif


kentonv:
Per the other thread, should we try again with EV_TRIGGER placed in the correct flags slot before giving up on it?

barracuda156:
@kentonv Rebased. Does the choice logic look good?

c++/src/kj/async-unix.c++
```diff
@@ -1065,7 +1065,11 @@ void UnixEventPort::captureChildExit() {
 void UnixEventPort::wake() const {
   // Trigger our user event.
   struct kevent event;
+#if defined(NOTE_TRIGGER)
   EV_SET(&event, 0, EVFILT_USER, 0, NOTE_TRIGGER, 0, nullptr);
+#elif defined(EV_TRIGGER)
+  EV_SET(&event, 0, EVFILT_USER, EV_TRIGGER, 0, 0, nullptr);
+#endif
```
kentonv, line 1072: If neither NOTE_TRIGGER nor EV_TRIGGER is defined, this code will still compile but produce undefined behavior.

```suggestion
#else
#error "neither NOTE_TRIGGER nor EV_TRIGGER is defined; we need at least one of these"
#endif
```

c++/src/kj/async-unix.h
```diff
@@ -31,14 +31,20 @@
...
+
 KJ_BEGIN_HEADER
 
 #if !defined(KJ_USE_EPOLL) && !defined(KJ_USE_KQUEUE)
 #if __linux__
 // Default to epoll on Linux.
 #define KJ_USE_EPOLL 1
-#elif __APPLE__ || __FreeBSD__ || __OpenBSD__ || __NetBSD__ || __DragonFly__
-// MacOS and BSDs prefer kqueue() for event notification.
+#elif (__APPLE__ && (MAC_OS_X_VERSION_MIN_REQUIRED > 1050)) \
```
kentonv, line 44: I am concerned about whether this macro actually gets defined to something sensible in most projects. From what I can find on the internet, `AvailabilityMacros.h` will default to assuming MacOS 10.4 on x64 (and 10.1 on PPC) if you don't explicitly specify otherwise. Does this mean people compiling modern programs with no actual intent to support such old versions might find kqueue disabled by accident? Does our _own_ build actually define this macro to something reasonable, or does this change accidentally make our own CI disable kqueue on mac?

I think I would prefer instead that people who are trying to build for such absurdly old versions should be responsible for using `-DKJ_USE_KQUEUE=0`.

barracuda156: > I am concerned about whether this macro actually gets defined to something sensible in most projects. From what I can find on the internet, AvailabilityMacros.h will default to assuming MacOS 10.4 on x64 (and 10.1 on PPC) if you don't explicitly specify otherwise.

No, this is wrong. I dunno who said that, but it is not how it works. Apple header, once included, sets correct values. It works robustly.

We already use `AvailabilityMacros.h` for `pthreads`. Countless port use those. I think this is a completely unwarranted concern.

P. S. There are cases where someone adds macros without thinking: that may not work, but this applies to any code. Perhaps that is the origin of an opinion you have referred to above. Yeah, some folks will set a comparison to a _verbal_ macro of a later OS, which does not exist on an earlier one )
But we know how to use these – and we used them hundreds of times (and not just locally in Macports, of course).

kentonv: When I google for availabilitymacros.h, I get this:

https://opensource.apple.com/source/CarbonHeaders/CarbonHeaders-8A428/AvailabilityMacros.h.auto.html

This clearly sets the default to 10.4, _unless_ `__ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__` is defined. What if the project forgets to define this? Do compiles define it automatically?

Or perhaps this is an old version of the header, and modern versions use a different default?

> I dunno who said that, but it is not how it works... I think this is a completely unwarranted concern.

I said it. I don't know anything about this. I'm asking you to show me how it works. Sorry, I have to actually understand before accepting code, I don't just accept PRs because someone assured me they are right.

> We already use AvailabilityMacros.h for pthreads.

I realize that, but in that case it is relatively benign if the work-around is used when not really needed, so I'm less worried about it.

barracuda156: @kentonv That one is the header from 10.4, but let me find what sets that macros. I will update on that soon. (As a matter of fact, something sets those, otherwise we would have had the code broken, but I need to look for specific documentation.)

barracuda156: UPD. I will update on the matter this weekend, please excuse me. Very little time, and it turned surprisingly hard to find documentation. I have found a description here https://todo.sr.ht/~breakfastquay/rubberband/1 saying that “If unspecified, the default (on Mac OS X 10.5 and later, which is probably as old as we care about) is for the deployment target to be equal to the major OS version on which the compile is taking place”, which matches actual behavior, but I understand this does not serve as a proof.

BTW, maybe we can change the condition in a way that is more satisfying? It would be nice to have it work on 10.5 without manual trickery, but of course we do not want to compromise any later systems, even potentially.

kentonv: Can you check what the header looks like on a modern MacOS machine, to verify it uses a more-recent version? I don't have a Mac handy or I'd check myself. As long as the header defaults to the version of the host then I'm happy.

barracuda156:
@kentonv Since we got 10.6.8 running on PowerPC natively now, this fix is redundant (10.6.8 has `NOTE_TRIGGER`, 10.5.8 has neither, so `EV_TRIGGER` won’t help there, it only fixed the build on a pre-release version of 10.6). Therefore I am closing this as not needed anymore.

I have confirmed now that v1.1.0 builds both with and without this patch on 10.6.8 ppc.

P. S. It won’t hurt having this merged, but we can also avoid extra code now.

PR #1737, opened 2023-07-31
Support coroutines on Windows as well

fabiorossetto:
The detection of the availability of coroutines relies on the __has_inclide macro, which is not available on Windows. Thus, coroutines cannot be used in Windows, despite KJ's implementation being standard compliant.

This commit checks whether the compiler is MSVC, and if the version is recent enough to support coroutines, it sets KJ_HAS_COROUTINE to 1.

### DIFF
--- a/.github/workflows/quick-test.yml
+++ b/.github/workflows/quick-test.yml
@@ -124,11 +124,11 @@ jobs:
             echo "Activate conda base environment"
             call activate base
             echo "Building Cap'n Proto with ${{ matrix.target }}"
-            cmake -Hc++ -Bbuild-output ${{ matrix.arch }} -G "${{ matrix.target }}" -DCMAKE_BUILD_TYPE=debug -DCMAKE_PREFIX_PATH="%CONDA_PREFIX%" -DCMAKE_INSTALL_PREFIX=%CD%\capnproto-c++-install
+            cmake -Hc++ -Bbuild-output ${{ matrix.arch }} -G "${{ matrix.target }}" -DCMAKE_BUILD_TYPE=debug -DCMAKE_CXX_STANDARD=20 -DCMAKE_PREFIX_PATH="%CONDA_PREFIX%" -DCMAKE_INSTALL_PREFIX=%CD%\capnproto-c++-install
             cmake --build build-output --config debug --target install
 
             echo "Building Cap'n Proto samples with ${{ matrix.target }}"
-            cmake -Hc++/samples -Bbuild-output-samples ${{ matrix.arch }} -G "${{ matrix.target }}" -DCMAKE_BUILD_TYPE=debug -DCMAKE_PREFIX_PATH=%CD%\capnproto-c++-install
+            cmake -Hc++/samples -Bbuild-output-samples ${{ matrix.arch }} -G "${{ matrix.target }}" -DCMAKE_BUILD_TYPE=debug -DCMAKE_CXX_STANDARD=20 -DCMAKE_PREFIX_PATH=%CD%\capnproto-c++-install
             cmake --build build-output-samples --config debug
 
             cd build-output\src
--- a/c++/CMakeLists.txt
+++ b/c++/CMakeLists.txt
@@ -1,6 +1,7 @@
 cmake_minimum_required(VERSION 3.6)
 project("Cap'n Proto" CXX)
 set(VERSION 1.1-dev)
+set(CMAKE_CXX_STANDARD_REQUIRED TRUE)
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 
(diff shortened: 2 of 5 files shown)


fabiorossetto:
FYI, the detection macro is the same used in Boost Asio

fabiorossetto:
Despite the change, coroutines are not picked up. I suspect that this is due to the flag `/Zc:__cplusplus` not being used, see
https://learn.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-170

I think the flag `/Zc:__cplusplus` should be used, but this could have some subtle backward compatibility implications, so it should be done carefully...

harrishancock:
Hi @fabiorossetto, we're actually about to require C++20 in PR #1732, which will have the effect of enabling coroutines under MSVC. It also has the unfortunate effect of requiring VS2022, dropping support for VS2019, though.

fabiorossetto:
@harrishancock Awesome! I guess this comes after release of 1.0.0. Will `mvCapture` be removed? I haven't seen it in #1732 

If C++20 becomes compulsory, then I guess we can close this MR...

kentonv:
@fabiorossetto To clarify, C++20 is now required on the `v2` branch, which is our new development branch for version 2.0.

You actually probably shouldn't use this branch! At least, not now. We're going to be making a lot of breaking changes; keeping up with it in the near term may be a pain. The `master` branch is now the "1.0 LTS" branch. I'm totally happy to accept changes like this on that branch.

kentonv:
> I think the flag /Zc:__cplusplus should be used, but this could have some subtle backward compatibility implications, so it should be done carefully...

I think this should be fine for KJ/Cap'n Proto, but I guess dependent projects will need to use it too, and it could affect them? Should we say that it's up to dependent projects to set this flag if they want to use coroutines?

Can you arrange to enable this in one of the CI builds so coroutines actually get tested?

fabiorossetto:
@harrishancock @kentonv I'm de-prioritizing this MR because me and my team decided to track v2 as we want to benefit from all improvements to coroutines and we are ok with breaking changes.
However, I would recommend we get this MR merged.

- I believe CI was building on MSVC with version C++14, when changing to C++20 the same issues appear that Harris fixed in v2. I've made a temporary commit that updates the C++ version for MSVC to C++20, but I guess it would be best to add an extra matrix entry that tests C++14 and C++20

- I have picked the coroutine fixes from Harris, but not the fixes for the other issues he found on MSVC, those should be cherry-picked as well.

- I've added /Z:cplusplus when using MSVC. There's perhaps a chance that this could cause issues, but IMO we shouldn't care. This MR forces you to specify the compiler version, and a standard compliant compiler should set __cplusplus to the right value. Capnproto already suffers from the lack of this flag. Me and my team had to reimplement the `arr` function because the version in KJ is protected by `#if __cplusplus > 201402L`, so it's not visible on MSVC at the moment

harrishancock:
I agree it'd be preferable to get capnproto 1.x's coroutines working under MSVC, as I was under the impression they were tested and working. Backporting the MSVC C++20 fixes while maintaining backwards-compatibility with C++14 is a bit more than I can pick up right now, though.

kentonv:
> my team decided to track v2

Please note that we don't intend to spend time supporting people outside our project using the v2 branch -- indeed, this is the whole reason for having a v2 branch. So, you are taking a risk here. We might remove features that you depend on without warning, if we don't happen to need them and they are slowing down development.

For example, MSVC support may bitrot in v2. For Windows builds `workerd` uses Clang in MSVC-compat mode, so while we expect to continue to support Windows, we have no need to chase down MSVC-specific bugs that Clang doesn't exhibit, and we might decide to stop doing so.

risa2000:
@fabiorossetto 
> * I believe CI was building on MSVC with version C++14, when changing to C++20 the same issues appear that Harris fixed in v2. I've made a temporary commit that updates the C++ version for MSVC to C++20, but I guess it would be best to add an extra matrix entry that tests C++14 and C++20

There has been a discussion about C++20 support in the PR here: https://github.com/capnproto/capnproto/pull/1819 and the one which precedes it.

> * I have picked the coroutine fixes from Harris, but not the fixes for the other issues he found on MSVC, those should be cherry-picked as well.

This is exactly what I did too following Kenton's recommendations :).

> * I've added /Z:cplusplus when using MSVC. There's perhaps a chance that this could cause issues, but IMO we shouldn't care. This MR forces you to specify the compiler version, and a standard compliant compiler should set __cplusplus to the right value. Capnproto already suffers from the lack of this flag. Me and my team had to reimplement the `arr` function because the version in KJ is protected by `#if __cplusplus > 201402L`, so it's not visible on MSVC at the moment

You may read the discussion about `__cplusplus` macro and the `/Zc:__cplusplus` compile option in the PR as well.
In short, for `kj`, being a library, it is not possible to guarantee that the library client will use this option too and if not you may get inconsistent C++ standard resolution in the library build and in the client code build. Forcing client code to turn this option for its build is the only way to go around it, but I felt it to be too intrusive.
Which eventually led to `KJ_CPP_STD` abstraction which should be safe and consistent in any scenario.

fabiorossetto:
@risa2000 thank you for your detailed answer. Should we close this PR in favour of #1819?

kentonv:
Makes sense to me.

PR #1738, opened 2023-07-31
Add HTTP `Range` header parser

mrbbot:
_(dependency of cloudflare/workerd#942)_

Hey! 👋 This PR adds a parser for HTTP `Range` headers to KJ, as per https://www.rfc-editor.org/rfc/rfc9110#section-14.1. It may make more sense for this code to be in `workerd`, but it seemed like it could be more generally useful, and the private `skipSpace` and `consumeNumber` functions in `http.c++` were very useful. 🙂

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -762,7 +762,7 @@ void HttpHeaders::takeOwnership(HttpHeaders&& otherHeaders) {
 
 // -----------------------------------------------------------------------------
 
-static inline char* skipSpace(char* p) {
+static inline const char* skipSpace(const char* p) {
   for (;;) {
     switch (*p) {
       case '\t':
@@ -773,6 +773,9 @@ static inline char* skipSpace(char* p) {
         return p;
     }
   }
+}
+static inline char* skipSpace(char* p) {
+  return const_cast<char*>(skipSpace(const_cast<const char*>(p)));
 }
 
 static kj::Maybe<kj::StringPtr> consumeWord(char*& ptr) {
@@ -805,14 +808,14 @@ static kj::Maybe<kj::StringPtr> consumeWord(char*& ptr) {
   }
 }
 
-static kj::Maybe<uint> consumeNumber(char*& ptr) {
-  char* start = skipSpace(ptr);
-  char* p = start;
+static kj::Maybe<uint> consumeNumber(const char*& ptr) {
+  const char* start = skipSpace(ptr);
+  const char* p = start;
 
   uint result = 0;
 
   for (;;) {
-    char c = *p;
+    const char c = *p;
     if ('0' <= c && c <= '9') {
       result = result * 10 + (c - '0');
       ++p;
@@ -822,6 +825,12 @@ static kj::Maybe<uint> consumeNumber(char*& ptr) {
       return result;
     }
   }
+}
+static kj::Maybe<uint> consumeNumber(char*& ptr) {
+  const char* constPtr = ptr;
+  auto result = consumeNumber(constPtr);
+  ptr = const_cast<char*>(constPtr);
+  return result;
 }
 
 static kj::StringPtr consumeLine(char*& ptr) {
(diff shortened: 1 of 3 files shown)


c++/src/kj/compat/http.c++
```diff
@@ -762,7 +762,8 @@ void HttpHeaders::takeOwnership(HttpHeaders&& otherHeaders) {
 
 // -----------------------------------------------------------------------------
 
-static inline char* skipSpace(char* p) {
+template<typename T>
+static inline T skipSpace(T p) {
```
mrbbot, line 766: For this function, and `consumeNumber`, I was looking to add support for `const char*` pointers as well. This seemed like a nice fix, but there might be something cleaner.

c++/src/kj/compat/http.h
```diff
@@ -161,7 +161,9 @@ class HttpHeaderId {
   MACRO(HOST, "Host") \
   MACRO(DATE, "Date") \
   MACRO(LOCATION, "Location") \
-  MACRO(CONTENT_TYPE, "Content-Type")
+  MACRO(CONTENT_TYPE, "Content-Type") \
+  MACRO(RANGE, "Range") \
+  MACRO(CONTENT_RANGE, "Content-Range")
```
mrbbot, line 166: I'm aware you can define these as needed in `HttpHeaderTable`s, but it seemed like a good idea to add support here if `Range` parsing was supported. The `TODO` comment below also seems to suggest the plan was to add more headers over time too.

c++/src/kj/compat/http.h
```diff
@@ -492,6 +494,27 @@ class HttpHeaders {
...
+
+  uint64_t start;
+  uint64_t end;
+
+  inline bool operator==(const HttpByteRange& other) const {
+    return start == other.start && end == other.end;
+  }
+  inline bool operator!=(const HttpByteRange& other) const {
+    return !(start == other.start && end == other.end);
+  }
```
mrbbot, line 508: Is there a nice way of declaring this as the inverse of `operator==`? I guess `return !(*this == other)` might work?

kentonv: That or `return !operator==(other);`

c++/src/kj/compat/http.c++
```diff
@@ -1110,6 +1112,150 @@ kj::String HttpHeaders::toString() const {
...
+
+namespace {
+
+// RFC9110 section 14.1: https://www.rfc-editor.org/rfc/rfc9110#section-14.1
+
+static bool consumeByteRangeUnit(const char*& ptr) {
+  const char* p = ptr;
+  p = skipSpace(p);
+
+  // Match case-insensitive "bytes"
```
kentonv, line 1126: Can you reuse `fastCaseCmp` here?

c++/src/kj/compat/http.c++
```diff
@@ -1110,6 +1112,150 @@ kj::String HttpHeaders::toString() const {
...
+  ++p;
+
+  p = skipSpace(p);
+  ptr = p;
+  return true;
+}
+
+enum HttpRangeParseError {
+  INVALID, // Parsing failed
+  UNSATISFIABLE // Able to parse and pointer advanced, incorrect for another reason
```
kentonv, line 1141: Below, it looks like this error code is used to mean "syntactically valid, but out-of-range", e.g. the range exceeds `Content-Length`.

Perhaps it would be clearer to separate the bounds-checking code from the parsing code? Parse the values as they appear first, then check the result against the content-length.

Then the functions could just return `Maybe` which is perhaps a bit clearer.

c++/src/kj/compat/http.c++
```diff
@@ -1110,6 +1112,150 @@ kj::String HttpHeaders::toString() const {
...
+  }
+}
+
+static HttpRangeParseResult consumeRangeSpec(const char*& ptr, uint64_t contentLength) {
+  KJ_SWITCH_ONEOF(consumeIntRange(ptr, contentLength)) {
+    KJ_CASE_ONEOF(range, HttpByteRange) {
+      return range;
+    }
+    KJ_CASE_ONEOF(error, HttpRangeParseError) {
+      if (error == UNSATISFIABLE) return error;
```
kentonv, line 1215: Do we really want to go on parsing if `consumeIntRange()` reported INVALID? Maybe add comments explaining why, if that's the case?

c++/src/kj/compat/http.c++
```diff
@@ -1110,6 +1112,150 @@ kj::String HttpHeaders::toString() const {
   return serialize(nullptr, nullptr, nullptr, nullptr);
 }
 
+// -----------------------------------------------------------------------------
+
+
+namespace {
+
+// RFC9110 section 14.1: https://www.rfc-editor.org/rfc/rfc9110#section-14.1
```
kentonv, line 1120: It'd help to add a few comments here summarizing what we're parsing so that people don't have to open the spec. Like, what does the overall text typically look like, and what's the difference between an int range and a suffix range? You don't have to go into full detail, people can refer to the spec for that.

c++/src/kj/compat/http.h
```diff
@@ -492,6 +494,27 @@ class HttpHeaders {
...
+};
+
+kj::String KJ_STRINGIFY(HttpByteRange range);
+
+kj::Maybe<kj::Array<HttpByteRange>> tryParseHttpRangeHeader(kj::ArrayPtr<const char> value, uint64_t contentLength);
+// Parses a HTTP Range header into an array of satisfiable inclusive ranges. Returns nullptr if the
+// range specifier was invalid/unsatisfiable per https://www.rfc-editor.org/rfc/rfc9110#section-14.
+// Returns an empty array if at least one of the specifier's range specs covers the full range.
```
kentonv, line 516: It's kind of confusing for an empty array to mean "match everything" while a non-empty array means "match only what's in the array". It would be easy for the code that calls this to be buggy, interpreting an empty array as "match nothing", and yet still seem to work most of the time.

Is there an important reason to treat "match everything" specially, rather than return an array with one range that covers the whole content?

c++/src/kj/compat/http.c++
kentonv: Hmm, I think I would define it as `const char* skipSpace(const char*)`, and then define an overload for `char*` which calls the const version and uses `const_cast` on the returned pointer. That way this doesn't accidentally get called with some other pointer type that doesn't make sense, and clangd can still understand the code well enough to flag errors and such.

That said it's a small function, it's not that big a deal.

kentonv (commented):
Mostly nitpicks here but I do worry returning an empty array for "match all" is a potential source of bugs.

c++/src/kj/compat/http.h
```diff
@@ -492,6 +494,27 @@ class HttpHeaders {
...
+    return start == other.start && end == other.end;
+  }
+  inline bool operator!=(const HttpByteRange& other) const {
+    return !(start == other.start && end == other.end);
+  }
+};
+
+kj::String KJ_STRINGIFY(HttpByteRange range);
+
+kj::Maybe<kj::Array<HttpByteRange>> tryParseHttpRangeHeader(kj::ArrayPtr<const char> value, uint64_t contentLength);
```
kentonv, line 513: (Non-actionable musings.)

It's a bit funny that we can't actually parse this header until we know the size of the resource. That implies you can't fully validate the headers until after looking up the file and querying it. Theoretically the API should be something like... parse into a representation of the header first, then validate that representation against the actual content length in a separate step, producing the final set of ranges.

But given this works fine for the present use case I guess it's fine. If we introduce a more complicated API later we can always implement this version in terms of it.

c++/src/kj/compat/http.c++
mrbbot: Not sure, since `bytes` isn't NUL terminated, and I'd like to avoid mutating the input to add a terminator.

c++/src/kj/compat/http.c++
mrbbot: Yep, since suffix ranges will be invalid int ranges. Added a comment. 👍

c++/src/kj/compat/http.h
mrbbot: You'd want to make sure you were returning a 200 response for match everything instead of 206, but you could equally detect the range matched everything in the caller. Will switch to returning a single range in this case. 👍

c++/src/kj/compat/http.h
mrbbot: That's fair, trying to think of cases where you'd want to split these parse and validation steps. I guess if you wanted to return a 416 response before trying to open a file? But even then, you might not know a range is unsatisfiable until you had the content length. 🤷

c++/src/kj/compat/http.h
kentonv: I see. Another alternative then would be something like `OneOf<Array, Everything, Unsatisfiable, Invalid>`.

mrbbot: Could we combine `Invalid` with `Unsatisfiable`? They'll both result in a 416 response, so I'm not sure we need to separate them.

kentonv: Oh, I suppose.

Weird, intuitively I would have expected an unparseable header should result in a 400.

mrbbot: Cool, changed the API to this 👍

PR #1739, opened 2023-07-31
RFC: CoExclusiveJoin

harrishancock:
This is a proof of concept, showing how a magic `co_await`able value could be implemented which exclusively-joins a promise with the current coroutine. The `co_await` is immediately ready, and results in an object which holds the passed promise and implements the Event necessary for the exclusive-join. This both provides cancellation -- the coroutine can let the returned object go out of scope to cancel the exclusively-joined background task -- and conveniently avoids any extra memory allocation.


harrishancock:
With some macro hackery, this could enable such syntax as:
```cpp
Promise<T> pump(AsyncInputStream& in, AsyncOutputStream& out) {
  {
    co_scoped_exclusive_join {
      co_await out.whenWriteDisconnected();
      throwFatalException(KJ_EXCEPTION(DISCONNECTED, "output disconnected"));
    };

    // Subject to cancellation by `out.whenWriteDisconnected()`.

    for (;;) {
      char buffer[1024];
      auto amount = co_await in.tryRead(buffer, sizeof(buffer));
      co_await out.write(buffer, amount);
    }
  }

  // No longer subject to cancellation.
}
```

kentonv:
I wonder if we can avoid the "exclusive join" terminology here, which I think most people won't understand. Can we call it a "background task"? Maybe even "failfast background task"?

I can also imagine several slightly different desired behaviors here. In all cases, if the background task fails, we want to cancel the coroutine and propagate the exception. But what do we want to do if the task finishes? I can imagine:

1. Await later: If the task completes, nothing happens immediately, but the coroutine can await the promise later to learn the result.
2. Exclusive join: The coroutine is canceled, returning the result of the promise. (Types must match.)
3. Never stop: If the task completes, this is considered an error, and an exception is thrown instead.

I suppose (3) can be implemented in terms of the others using syntax like in your example, so maybe we only really need (1) and (2).

harrishancock:
I've had similar thoughts. I agree there'll probably be a use case for the await-later semantics, but I'm not sure what naming makes sense for the two basic behaviors. Something like "background task" sounds fairly intuitive to me, though I've also considered something based on "race" -- I think the JavaScript Promise API uses that word for "exclusive join".

kentonv:
`co_race_with` and `co_failfast`?

Or maybe `KJ_CO_RACE_WITH` and `KJ_CO_FAILFAST`?

harrishancock:
I have been considering the pair `co_race` and `co_race_exceptions`, but `co_race_with` and `co_failfast` also work for me.

Regarding naming style, `KJ_MACRO`-style naming seems more conservative, so I kind of lean toward that. I don't have a strong preference, though. I'll use `KJ_CO_RACE_WITH` and `KJ_CO_FAILFAST` for the rest of this comment to see how it looks.

I have been noodling over whether we should support promises or just code. I of course think it'd be pretty amazing to be able to write

```cpp
auto task = KJ_CO_RACE_WITH {
  co_await other.stream->whenWriteDisconnected();
  kj::throwFatalException(blah);
}
```

But if we implemented that, then `auto task = KJ_CO_RACE_WITH promise` would become impossible to implement. We could support some sort of alternate syntax for promises, e.g. `auto task = co_yield kj::coRace(kj::mv(promise))`, or provide differently-named macros.

Another little detail is that `KJ_CO_FAILFAST` would actually need to be `KJ_CO_FAILFAST(T)`, because it needs to manually specify what the return type of the lambda will be (`-> kj::Promise<T>`). For `KJ_CO_RACE_WITH`, we can use magic to infer the enclosing coroutine's type.

kentonv:
Another possible syntax, which I raised in the thread about deferred proxying, might be:

```
co_magic KJ_RACE_WITH {
```

> Another little detail is that KJ_CO_FAILFAST would actually need to be KJ_CO_FAILFAST(T), because it needs to manually specify what the return type of the lambda will be (-> kj::Promise<T>).

Could also be `KJ_CO_FAILFAST -> kj::Promise<T> {`. But yeah, ugh.

PR #1741, opened 2023-08-01
coroutines: Refactor adapter for reusability

harrishancock:
I have a use case for implementing a custom coroutine adapter for a type derived from kj::Promise<T>, and would like to re-use the kj::Promise<T> coroutine implementation. It turns out that, with just a few modifications, the kj::Promise<T> coroutine adapter can be used as a component in a wrapper adapter.


kentonv (approved):

kentonv:
I'll be interested to see how this is actually used...

kentonv:
Oh I see: https://github.com/cloudflare/workerd/pull/953

PR #1743, opened 2023-08-02
Allow to release the fd from an AsyncIoStream

fabiorossetto:
This commit adds a function to AsyncIoStream that helps in the interoperability of the class with other libraries. It allows to
reclaim ownership of a file descriptor from an AsyncIoStream. This can allow for example to use KJ ConnectionReceiver, but then use the file descriptors natively, or pass them to another library.


c++/src/kj/async-io.h
```diff
@@ -169,6 +169,22 @@ class AsyncIoStream: public AsyncInputStream, public AsyncOutputStream {
   virtual kj::Maybe<int> getFd() const { return nullptr; }
   // Get the underlying Unix file descriptor, if any. Returns nullptr if this object actually
   // isn't wrapping a file descriptor.
+
+  virtual Maybe<void*> getWin32Handle() const { return nullptr; }
+  // Get the underlying Win32 HANDLE, if any. Returns nullptr if this object actually isn't
+  // wrapping a handle.
+
+  static Maybe<int> releaseFd(Own<AsyncIoStream>&& stream);
```
kentonv, line 177: It's a little unexpected that this is a static method with a private virtual method for implementation. I get that the intent is that it forces you to destroy the stream. But what if the function returns null? Then you've inadvertently lost the stream altogether.

I think possibly we need to make this an instance method with documentation saying "if this returns non-null, you must immediately destroy the stream object before using the FD".

c++/src/kj/async-io.c++
```diff
@@ -2629,6 +2629,22 @@ Promise<void> AsyncCapabilityStream::sendFd(int fd) {
   return promise.attach(kj::mv(fds));
 }
 
+Maybe<int> AsyncIoStream::releaseFd(Own<AsyncIoStream>&& stream) {
+  KJ_IF_MAYBE(fd, stream->getFd()) {
```
kentonv, line 2633: I think you intended to use `doReleaseFd()` here?

Why didn't the test catch this?

fabiorossetto:
@kentonv I reworked the tests according to your suggestion. However, the test in windows fail, When calling `CreateNewCompletionPort` with the fd released from the stream.

I'm not an expert when it comes to Windows, and I'm wondering whether calling `CreateNewCompletionPort` on a handle for which `CreateNewCompletionPort` was already called once works. I haven't found any call in KJ that actively drops the completion port (I was expecting the IoObserver to do something like that in the dtor)

c++/src/kj/async-io.h
```diff
@@ -173,6 +173,20 @@ class AsyncIoStream: public AsyncInputStream, public AsyncOutputStream {
   virtual Maybe<void*> getWin32Handle() const { return nullptr; }
   // Get the underlying Win32 HANDLE, if any. Returns nullptr if this object actually isn't
   // wrapping a handle.
+
+#if _WIN32
```
kentonv, line 177: I don't think we need this `#if`. Both methods can be defined on all platforms, but the one that isn't relevant to the platform will always return null, same as the `get` methods above.

fabiorossetto: As far as I can tell, `AutoCloseHandle` is defined only if _WIN32, that's why I've added if guard

fabiorossetto:
> Apparently, you can use [NtSetInformationFile](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/nf-ntifs-ntsetinformationfile) to remove the handle from an IOCP. This seems like a more obscure API that Microsoft doesn't recommend using for some reason, but it is documented at least.

You are right. I checked the implementation of `release` in the boost asio library, and they do indeed use NtSetInformationFile. 

> Perhaps `Win32IocpEventPort::IoObserverImpl`'s destructor should be using this API in the specific case that the handle is not owned?

I've added a `stopObserving` method to Win32IocpEventPort::IoObserver. Putting the code in the destructor would be more elegant, but we would still need a way to tell the IoObserver to release ownership...

c++/src/kj/async-win32.h
```diff
@@ -91,6 +91,9 @@ class Win32EventPort: public EventPort {
     virtual Own<IoOperation> newOperation(uint64_t offset) = 0;
     // Begin an I/O operation. For file operations, `offset` is the offset within the file at
     // which the operation will start. For stream operations, `offset` is ignored.
+
+    virtual void stopObserving() = 0;
```
kentonv, line 95: Hmm this doesn't feel right. Good RAII style would have the destructor do this automatically.

I think it'd be OK to have the destructor always remove the handle (even if it's about to be closed anyway). Then we end up with similar behavior to `UnixEventPort::FdObserver`.

We can then remove the `WARNING` in the doc comment for `observeIo()`, below.

c++/src/kj/async-win32.c++
```diff
@@ -133,13 +133,44 @@ private:
   Promise<IoResult> promise;
 };
 
+typedef LONG(NTAPI* NtSetInfoFuncPtr)(HANDLE, ULONG_PTR*, void*, ULONG, ULONG);
+
+NtSetInfoFuncPtr getSetInfoFileFunc()
+{
```
kentonv, line 139: Can you please make sure this code follows the prevailing style with regard to brace positioning and indentation width?

c++/src/kj/async-win32.c++
```diff
@@ -133,13 +133,44 @@ private:
...
+    static thread_local NtSetInfoFuncPtr setInfoFileFunc = []() -> NtSetInfoFuncPtr {
```
kentonv, line 140: Why `thread_local`? Should be the same in all threads right? Note that static local initialization is thread-safe since C++11.

c++/src/kj/async-win32.c++
```diff
@@ -133,13 +133,44 @@ private:
...
+        HMODULE ntDll = ::GetModuleHandleA("NTDLL.DLL");
+        if (!ntDll)
+        {
+            return nullptr;
```
kentonv, line 144: You could throw an exception here, rather than `KJ_REQUIRE` that it's non-null later. That'd give more information about what went wrong.

If an exception is throw during static local initialization, the next call to the function will retry the initialization.

c++/src/kj/async-win32.c++
fabiorossetto: Ah! Indeed, I did it because I feared it would not be thread-safe to use a non thread_local variable. Good to know that's not the case

c++/src/kj/async-win32.c++
fabiorossetto: Can you suggest which way of throwing an exception would be more appropriate in this case?
I'm thinking
```c++
KJ_ASSERT(ntDll != nullptr, "Could not load 'ntdll' dynamic library");
```

kentonv: `KJ_FAIL_WIN32` will also use `GetLastError()` to find out the error code and incorporate that into the exception.

PR #1745, opened 2023-08-02
http: Convert HttpEntityBodyReader to coroutines

harrishancock:


c++/src/kj/compat/http.c++
```diff
@@ -1823,15 +1822,18 @@ private:
         } else if (amount < minBytes) {
           // We requested a minimum 1 byte above, but our own caller actually set a larger minimum
           // which has not yet been reached. Keep trying until we reach it.
-          return tryReadInternal(reinterpret_cast<byte*>(buffer) + amount,
-                                 minBytes - amount, maxBytes - amount, alreadyRead + amount);
+          buffer = reinterpret_cast<byte*>(buffer) + amount;
+          minBytes -= amount;
```
mikea, line 1826: maybe rather than mutating these it is better to use `alreadyRead` variable everywhere?

c++/src/kj/compat/http.c++
```diff
@@ -1854,37 +1856,40 @@ private:
...
         chunkSize -= amount;
         if (amount == 0) {
           kj::throwRecoverableException(KJ_EXCEPTION(DISCONNECTED, "premature EOF in HTTP chunk"));
         } else if (amount < minBytes) {
           // We requested a minimum 1 byte above, but our own caller actually set a larger minimum
           // which has not yet been reached. Keep trying until we reach it.
-          return tryReadInternal(reinterpret_cast<byte*>(buffer) + amount,
-                                 minBytes - amount, maxBytes - amount, alreadyRead + amount);
+          buffer = reinterpret_cast<byte*>(buffer) + amount;
+          minBytes -= amount;
```
mikea, line 1885: same here?

mikea (changes requested):

c++/src/kj/compat/http.c++
```diff
@@ -1805,15 +1803,16 @@ private:
 
   Promise<size_t> tryReadInternal(void* buffer, size_t minBytes, size_t maxBytes,
```
kentonv, line 1804: Since the only caller of this is now `tryRead()` itself, can we inline it?

c++/src/kj/compat/http.c++
```diff
@@ -1854,37 +1856,40 @@ private:
...
```
kentonv, line 1857: Inline this too?

c++/src/kj/compat/http.c++
kentonv: (Preferably as a separate commit to be able to see changes more clearly.)

c++/src/kj/compat/http.c++
kentonv: Personally I think either way is fine, but changing things has some risk of introducing bugs.

harrishancock: @mikea I'm not sure I have a picture of what you mean. Do you mean, for example, that line 1814 would become this?
```cpp
      auto amount = co_await getInner().tryRead(
          reinterpret_cast<byte*>(buffer) + alreadyRead,
          1, kj::min(maxBytes - alreadyRead, length - alreadyRead));
```

c++/src/kj/compat/http.c++
harrishancock: Hmm, I screwed up the "separate commit" part, but I think the diff looks okay.

kentonv (approved):
LGTM pending decision on performance.

mikea (approved):

PR #1744, opened 2023-08-02
http: Convert HttpInputStreamImpl to coroutines

harrishancock:
This required a slight adjustment to a test, due to a change in the eagerness of a promise chain.

PR #1746, opened 2023-08-02
http: Convert HttpOutputStream to coroutines

harrishancock:


c++/src/kj/compat/http.c++
```diff
@@ -2177,10 +2177,14 @@ private:
     // is empty, then they make the write directly, using `writeInProgress` to detect and block
     // concurrent writes.
 
-    writeQueue = writeQueue.then([this,content=kj::mv(content)]() mutable {
-      auto promise = inner.write(content.begin(), content.size());
-      return promise.attach(kj::mv(content));
-    });
+    constexpr auto doWrite = [](Promise<void> writeQueue,
```
mikea, line 2180: maybe better name for writeQueue? `writesBefore`?

Also I wonder if there's a better way to express such a queue rather than using promises?

mikea (approved):

c++/src/kj/compat/http.c++
```diff
@@ -2073,49 +2073,49 @@ public:
...
 
-    return fork.addBranch().then([this,buffer,size]() {
-      return inner.write(buffer, size);
-    }).then([this]() {
-      writeInProgress = false;
-    });
+    co_await fork.addBranch();
+    co_await inner.write(buffer, size);
+
+    // TODO(cleanup): Use KJ_DEFER instead?
```
kentonv, line 2086: No, these intentionally don't use KJ_DEFER. Perhaps add a comment:

> We intentionally don't use KJ_DEFER to clean this up because if an exception is thrown, we want to block further writes.

c++/src/kj/compat/http.c++
```diff
@@ -2073,49 +2073,49 @@ public:
   }
 
   kj::Promise<void> writeBodyData(const void* buffer, size_t size) {
-    KJ_REQUIRE(!writeInProgress, "concurrent write()s not allowed") { return kj::READY_NOW; }
```
kentonv, line 2076: Interestingly these recovery blocks were leaking memory before (when exceptions are enabled!) due to the longstanding clang bug with throwing post-return.

Or at least, they were leaking memory up until the recent optimization making `kj::READY_NOW` into a singleton.

c++/src/kj/compat/http.c++
kentonv: Ideally we get rid of this write queue, it's a pattern that has been known to cause trouble. Once we have `AsyncOutputStream::end()` it should be easier, as we can write the chunk terminator as part of `end()` instead of deferring it.

I would keep the name `writeQueue` for now. It's a pattern that appears in a few places.

c++/src/kj/compat/http.c++
```diff
@@ -2177,10 +2177,14 @@ private:
...
-      return promise.attach(kj::mv(content));
-    });
+    constexpr auto doWrite = [](Promise<void> writeQueue,
+                                AsyncOutputStream& inner,
+                                String content) -> Promise<void> {
+      co_await writeQueue;
+      co_await inner.write(content.begin(), content.size());
+    };
+
+    writeQueue = doWrite(kj::mv(writeQueue), inner, kj::mv(content));
```
kentonv, line 2187: I feel like this function got more complicated TBH. Maybe it was better the way it was.

c++/src/kj/compat/http.c++
harrishancock: Ouch!

c++/src/kj/compat/http.c++
harrishancock: Works for me.

c++/src/kj/compat/http.c++
harrishancock: > Also I wonder if there's a better way to express such a queue rather than using promises?

I think we do need a general alternative, but I haven't been able to think one up in the past.

c++/src/kj/compat/http.c++
harrishancock: Removed.

harrishancock:
As noted in #1744, performance doesn't turn out to be a concern after all.

PR #1747, opened 2023-08-02
http: Convert HttpEntityBodyWriter to coroutines

harrishancock:


mikea (approved):

c++/src/kj/compat/http.c++
```diff
@@ -2339,10 +2336,13 @@ private:
   uint64_t length;
 
   kj::Promise<void> maybeFinishAfter(kj::Promise<void> promise) {
-    if (length == 0) {
-      return promise.then([this]() { doneWriting(); });
-    } else {
-      return kj::mv(promise);
+    // `length` shouldn't be mutated while we await `promise` ... but just in case, let's save it.
```
kentonv, line 2339: If the length could change during the promise, then it would actually be more correct in this case to check it after, not before. But in fact, it can't change, hence the old code checked it before out of convenience. In any case, storing a copy here definitely isn't needed.

c++/src/kj/compat/http.c++
```diff
@@ -2339,10 +2336,13 @@ private:
   uint64_t length;
 
   kj::Promise<void> maybeFinishAfter(kj::Promise<void> promise) {
-    if (length == 0) {
-      return promise.then([this]() { doneWriting(); });
-    } else {
-      return kj::mv(promise);
```
kentonv, line 2345: The old code here was a no-op in the common case that we're not at the end of the stream yet. Only in the very last call, once per stream, it would allocate a `.then()`.

The new code, unfortunately, allocates a coroutine for every write.

One way to approximate the old performance would be to factor out just `promise.then([this]() { doneWriting(); })` into a separate coroutine. `maybeFinishAfter()` itself would not need to be a coroutine, it just conditionally calls the other coroutine.

However, that would still be more verbose than the old code, defeating the purpose of the coroutine conversion. Maybe we could just leave this one as-is?

c++/src/kj/compat/http.c++
```diff
@@ -2362,23 +2362,22 @@ public:
   }
 
   Promise<void> write(const void* buffer, size_t size) override {
-    if (size == 0) return kj::READY_NOW;  // can't encode zero-size chunk since it indicates EOF.
+    if (size == 0) co_return;  // can't encode zero-size chunk since it indicates EOF.
 
     auto header = kj::str(kj::hex(size), "\r\n");
     auto parts = kj::heapArray<ArrayPtr<const byte>>(3);
```
kentonv, line 2368: This can be an on-stack array now.

kentonv: Also `header` could be serialized into an on-stack char array, since the size of the array cannot exceed 19 characters (including NUL terminator).

c++/src/kj/compat/http.c++
```diff
@@ -2362,23 +2362,22 @@ public:
...
 
   Promise<void> write(ArrayPtr<const ArrayPtr<const byte>> pieces) override {
     uint64_t size = 0;
     for (auto& piece: pieces) size += piece.size();
 
-    if (size == 0) return kj::READY_NOW;  // can't encode zero-size chunk since it indicates EOF.
+    if (size == 0) co_return;  // can't encode zero-size chunk since it indicates EOF.
 
     auto header = kj::str(kj::hex(size), "\r\n");
     auto partsBuilder = kj::heapArrayBuilder<ArrayPtr<const byte>>(pieces.size() + 2);
```
kentonv, line 2383: Maybe KJ_STACK_ARRAY?

harrishancock: That works, but only because our Clang implementation of KJ_STACK_ARRAY doesn't use VLAs. If we were to compile it on GCC, it would try to use a VLA, which aren't allowed in coroutines.

Perhaps we could have a KJ_STACK_ARRAY_NO_VLA?

c++/src/kj/compat/http.c++
harrishancock: Yeah, very good points. I've reverted this part.

c++/src/kj/compat/http.c++
harrishancock: I pushed a version using KJ_STACK_ARRAY in the meantime.

kentonv: Hmm that does seem like a problem. I suspect no compiler is likely ever to support VLAs in coroutines.

Maybe we should just change KJ_STACK_ARRAY to not use VLAs ever, given we only really care about clang which doesn't support them?

kentonv:
You force-pushed twice in a row so github will only show me the second one and I have no way to know what happened in the first. :(

Screenshot from 2023-08-07 13-00-44

c++/src/kj/compat/http.c++
```diff
@@ -2362,64 +2359,72 @@ public:
...
+    // The size of a hex-encoded uint64_t is 16 bytes. The CRLF and NUL terminator contribute an
+    // additional 3 bytes, for a total required space of 19.
+    kj::FixedArray<char, 19> headerSpace;
+    auto header = strPreallocated(headerSpace, kj::hex(size), "\r\n"_kj);
 
-    auto parts = partsBuilder.finish();
-    auto promise = getInner().writeBodyData(parts.asPtr());
-    return promise.attach(kj::mv(header), kj::mv(parts));
+    auto partCount = pieces.size() + 2;
+    constexpr auto MAX_IN_FRAME = 256;
```
kentonv, line 2392: TBH I'd probably make this more like... 8 or 16. 256 ArrayPtrs is 4kB...

c++/src/kj/compat/http.c++
```diff
@@ -2362,64 +2359,72 @@ public:
...
   Promise<void> write(const void* buffer, size_t size) override {
-    if (size == 0) return kj::READY_NOW;  // can't encode zero-size chunk since it indicates EOF.
+    if (size == 0) co_return;  // can't encode zero-size chunk since it indicates EOF.
+
+    // The size of a hex-encoded uint64_t is 16 bytes. The CRLF and NUL terminator contribute an
+    // additional 3 bytes, for a total required space of 19.
+    kj::FixedArray<char, 19> headerSpace;
+    auto header = strPreallocated(headerSpace, kj::hex(size), "\r\n"_kj);
+
+    auto partCount = 4;
```
kentonv, line 2369: You only fill in 3 parts below.

c++/src/kj/compat/http.c++
```diff
@@ -2362,64 +2359,72 @@ public:
...
+    constexpr auto MAX_IN_FRAME = 256;
```
kentonv, line 2370: There are only 3 parts, you don't need to allocate space for 256 of them. :)

c++/src/kj/compat/http.c++
```diff
@@ -2362,64 +2359,72 @@ public:
...
+    KJ_STACK_ARRAY(ArrayPtr<const byte>, parts, partCount, MAX_IN_FRAME, MAX_IN_FRAME);
```
kentonv, line 2371: Can just be `ArrayPtr<const byte> parts[3];`

c++/src/kj/compat/http.c++
harrishancock: Yeah, I think that change is pretty reasonable. It'd be nice to be able to safely use KJ_STACK_ARRAY guilt-free in coroutines. I'll see if I can make it.

c++/src/kj/compat/http.c++
harrishancock: Great point.

c++/src/kj/compat/http.c++
harrishancock: I wonder what the hell I was thinking here ...

c++/src/kj/compat/http.c++
harrishancock: Dropping VLA support from KJ_STACK_ARRAY in #1762.

harrishancock:
Rebased onto #1762 and replaced `KJ_STACK_ARRAY` with `kj::SboArrayBuilder`.

harrishancock:
Now that #1762 is in, I rebased on v2 and used SmallArrayBuilder. @kentonv, PTAL.

kentonv (approved):

PR #1748, opened 2023-08-03
Define KJ_MSVC_TRADITIONAL_CPP also in Release

fabiorossetto:
The code in common.h for the definition of KJ_MSVC_TRADITIONAL_CPP was
active only in debug. However, it is used in debug.h even when not in
debug. Thus, while the fix worked for debug builds, it did not work
for release builds.

This was probably undetected because CI builds only Debug for MSVC.

### DIFF
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -323,10 +323,11 @@ KJ_NORETURN(void unreachable());
 
 }  // namespace _ (private)
 
-#ifdef KJ_DEBUG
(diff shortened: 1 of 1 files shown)


kentonv:
Looks like #1342 introduced the macro but only actually used it immediately after, in debug-only code. Later changes used the macro elsewhere, without realizing it was debug-only. Doh.

kentonv:
I suspect we'll just merge everything that lands on the master branch into v2, but on our own schedule. This issue doesn't affect Workers since we don't use MSVC.

PR #1753, opened 2023-08-06
MSVC improvements, clean up constexpr declarations

fhanau:
- Define /Zc:__cplusplus, supported since MSVC 2017, which sets `__cplusplus` to the right value. By default, MSVC sets it to 199711L for compatibility with prehistoric codebases. This will allow us to use __cplusplus more easily in the future without needing special handling for MSVC. With this change MSVC versions before 2017 are no longer supported, but since we now require C++ 20 that should be acceptable.
- With MSVC 2017 and above warning 4503 is no longer being used, no need to disable it.
- With the upgrade to C++ 20, `CAPNP_NEED_REDUNDANT_CONSTEXPR_DECL` is no longer needed.


harrishancock:
Thanks!

Would you mind separating the .capnp.h/c++ file regeneration out into a separate commit?

harrishancock (approved):

fhanau:
Moved the regeneration to a new commit, also added a TODO message for converting regenerate-bootstraps.sh to a bazel command.

harrishancock (approved):

PR #1754, opened 2023-08-07
Coroutine fiddling: WaitScope parameter ban

harrishancock:
Three little changes:

- `WaitScope` parameters are now banned for coroutines, to prevent them from being simultaneously fibers and coroutines. I implemented this in `coroutine_traits` rather than `kj::_::Coroutine<T>`, because I was worried that if I passed `Args...` down to `kj::_::Coroutine<T, Args...>`, it might make coroutine-related error messages a bit too cluttered. That said, it means that custom coroutine adapters based on `kj::_::Coroutine<T>` can't name `kj::_::Coroutine<T>` themselves, but should always go through `coroutine_traits`, in order to benefit from the parameter type validation.

- ~~I made it so that `co_await blah` is now `[[nodiscard]]`. I could go either way on this. It'll lead to explicit `void` casts in a few places, but it also caught a real bug in https://github.com/cloudflare/workerd/pull/982.~~ Too many false positives, removed.

- ~~Defined `co_magic`, as discussed in #1739 and https://github.com/cloudflare/workerd/pull/953.~~ Moved to separate PR.

### DIFF
--- a/c++/src/kj/async-inl.h
+++ b/c++/src/kj/async-inl.h
@@ -2052,7 +2052,16 @@ PromiseCrossThreadFulfillerPair<T> newPromiseAndCrossThreadFulfiller() {
 // coroutine implementation type via a traits class parameterized by the coroutine return and
 // parameter types. We'll name our coroutine implementation `kj::_::Coroutine<T>`,
 
-namespace kj::_ { template <typename T> class Coroutine; }
+namespace kj::_ {
+
+template <typename T> class Coroutine;
+
+template <typename T>
+concept NoWaitScope = !isSameType<Decay<T>, WaitScope>();
+// Define a Concept to use in our `coroutine_traits` specialization to validate allowable coroutine
+// parameter types.
+
+}  // namespace kj::_
 
 // Specializing the appropriate traits class tells the compiler about `kj::_::Coroutine<T>`.
 
@@ -2061,6 +2070,17 @@ namespace KJ_COROUTINE_STD_NAMESPACE {
 template <class T, class... Args>
 struct coroutine_traits<kj::Promise<T>, Args...> {
   // `Args...` are the coroutine's parameter types.
+
+  static_assert((::kj::_::NoWaitScope<Args> && ...),
+      "Coroutines are not allowed to accept `WaitScope` parameters.");
+  // Coroutines should never have access to a WaitScope. If they could, coroutines could be both a
+  // coroutine and a fiber at the same time, and we'd like to ban that.
+  //
+  // Note: We could have just declared `Args` as `NoWaitScope...`, but the default compiler message
+  // about a missing `promise_type` in `coroutine_traits` is not the greatest.
+  //
+  // A second note: This has the reasonable side effect of making it impossible for us to write
+  // WaitScope member coroutines.
 
   using promise_type = kj::_::Coroutine<T>;
   // The C++ standard calls this the "promise type". This makes sense when thinking of coroutines


harrishancock:
I removed the nodiscard change, there were too many false positives.

c++/src/kj/async-inl.h
```diff
@@ -2335,6 +2355,29 @@ class Coroutine<T>::Awaiter: public AwaiterBase {
...
+// To invoke a coroutine's magic functionality `FOO`, use `co_magic FOO`. What `FOO` is, what the
+// expression evaluates to, and what side effects occur are all defined by the implementation of the
+// coroutine adapter.
+//
+// Coroutine magic is implemented in terms of the `co_yield` keyword, so this is just a define to
+// `co_yield`. The purpose of the macro is primarily as a visual signal that the code is not
+// actually yielding a value, but rather something different is going on.
+//
+// TODO(2.0): Unconditionally defining this is perhaps unwise. Should we have a KJ_DEFINE_CO_MAGIC
+//   macro to control this? Or we could `#define KJ_CO_MAGIC co_yield` and rely on projects to pass
```
kentonv, line 2378: Maybe we should just have the macro be called `KJ_CO_MAGIC`. That tells readers that this is a KJ feature and hints that they can jump-to-definition for more info, rather than have them googling for it.

c++/src/kj/async-inl.h
```diff
@@ -2052,7 +2052,16 @@ PromiseCrossThreadFulfillerPair<T> newPromiseAndCrossThreadFulfiller() {
 // coroutine implementation type via a traits class parameterized by the coroutine return and
 // parameter types. We'll name our coroutine implementation `kj::_::Coroutine<T>`,
 
-namespace kj::_ { template <typename T> class Coroutine; }
+namespace kj::_ {
+
+template <typename T> class Coroutine;
+
+template <typename T>
+concept NoWaitScope = !isSameType<Decay<T>, WaitScope>();
```
kentonv, line 2060: Hmm, is there any particular reason this is better-expressed as a concept rather than a constexpr function? I see you're just using it to evaluate to its boolean value in practice.

harrishancock: My initial version used
```cpp
template <typename T, NoWaitScope... Args>
struct coroutine_traits<T, Args...> {
  // ...
```
before I realized that a static_assert would provide a friendlier error message.

I don't have a strong preference among the four options (concept, constexpr variable, constexpr function, or just using isSameType() directly in the static_assert).

c++/src/kj/async-inl.h
harrishancock: Yeah, that works for me. The jump-to-definition is a great point.

c++/src/kj/async-inl.h
harrishancock: For reference, here is the error message from the concept-using static_assert:
```
In file included from src/kj/async-coroutine-test.c++:22:
In file included from bazel-out/k8-fastbuild/bin/src/kj/_virtual_includes/kj-async/kj/async.h:1330:
bazel-out/k8-fastbuild/bin/src/kj/_virtual_includes/kj-async/kj/async-inl.h:2074:3: error: static assertion failed: Coroutines are not allowed to accept `WaitScope` parameters.
  static_assert((::kj::_::NoWaitScope<Args> && ...),
  ^              ~~~~~~~~~~~~~~~~~~~~~~~~~~
src/kj/async-coroutine-test.c++:254:3: note: in instantiation of template class 'std::coroutine_traits<kj::Promise<void>, kj::WaitScope &>' requested here
  co_return;
  ^
bazel-out/k8-fastbuild/bin/src/kj/_virtual_includes/kj-async/kj/async-inl.h:2074:18: note: because 'kj::WaitScope &' does not satisfy 'NoWaitScope'
  static_assert((::kj::_::NoWaitScope<Args> && ...),
                 ^
bazel-out/k8-fastbuild/bin/src/kj/_virtual_includes/kj-async/kj/async-inl.h:2060:23: note: because '!isSameType<Decay<kj::WaitScope &>, kj::WaitScope>()' evaluated to false
concept NoWaitScope = !isSameType<Decay<T>, WaitScope>();
```

and here with using isSameType() directly:
```
In file included from src/kj/async-coroutine-test.c++:22:
In file included from bazel-out/k8-fastbuild/bin/src/kj/_virtual_includes/kj-async/kj/async.h:1330:
bazel-out/k8-fastbuild/bin/src/kj/_virtual_includes/kj-async/kj/async-inl.h:2069:3: error: static assertion failed due to requirement '!::kj::isSameType()': Coroutines are not allowed to accept `WaitScope` parameters.
  static_assert(!(::kj::isSameType<::kj::Decay<Args>, ::kj::WaitScope>() || ...),
  ^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/kj/async-coroutine-test.c++:254:3: note: in instantiation of template class 'std::coroutine_traits<kj::Promise<void>, kj::WaitScope &>' requested here
  co_return;
  ^
```

harrishancock: I think my logic goes like this:
1. If we want to define a constexpr nullary function template, e.g. `isSameType<T, U>()`, we should consider a constexpr variable template instead, because the `()` syntax doesn't really add much value. I suspect the reason we didn't use constexpr variables for this sort of thing in the past is mostly that constexpr functions arrived in the C++ standard first.
2. If we are defining a constexpr _boolean_ variable template, we should consider using a concept instead, because `template <typename T> concept ...` is the same thing as `template <typename T> constexpr bool ...`, with the added feature that the concept can be used to constrain template parameters. (`constexpr bool` can in fact be a drop-in replacement for `concept` in this particular code.)

I'll leave it as-is, because I'd like to start getting used to using them, though I admit this isn't the most compelling example.

> (I would approve this PR except it still has the co_magic define.)

D'oh, fixed.

PR #1756, opened 2023-08-08
kj/common: remove obsolete comments on minValue and maxValue.

balusch:
Type cast to char was supported since commit 98f9c8fca5c49f.

Signed-off-by: Jianyong Chen <baluschch@gmail.com>


kentonv:
Thanks! Hah, that one has been around for a while..

PR #1757, opened 2023-08-09
PROOF OF CONCEPT: This adds emscripten support in a commit to be able to at least compile and run tests via node(needs more testing)

germandiagogomez:
This PR adds emscripten support against version 0.10.3 of capnproto.

It has two commits, one with meson + emscripten support since it is what I have been working on and another with the C++-only changes.

Feel free to try it and especially to run the tests. Some tests still fail: some time-related tests and Capability/Basic (could be due to some missing emscripten flag when compiling since heap stuff is different and so on)  tests are failing. Many other tests pass.

I will try the library in the context of my game to see if failed tests are really relevant in my context. I expect to need more work to iron out a few details.

If a successful emscripten port can be added, I will port the build system to CMake. 

Meson support is not proposed, though you can read the commit to see what it does right.

## How to try this branch

If you do not have meson installed, assuming you have python, you can do:

```
python3 -m pip install meson
```

After meson ins installed, follow these instruction to test:

    1. use ./bootstrap-emscripten.sh. This will download the sdk and activate it and create a emscripten_constants.ini file in your meson/cross directory to point to your toolchain correctly.
   2. issue this command: meson setup -Dwith_openssl=disabled -Dwith_fibers=disabled -Dwith_zlib=disabled -Dbuild_testing=true --cross-file meson/cross/emscripten_constants.ini --cross-file meson/cross/emscripten.ini build-emscripten

You can run the tests by issuing from project root:

```
meson test -C build-emscripten --verbose
```

### DIFF
--- a/c++/src/capnp/compat/json-test.c++
+++ b/c++/src/capnp/compat/json-test.c++
@@ -22,7 +22,7 @@
 #include "json.h"
 #include <capnp/test-util.h>
 #include <capnp/compat/json.capnp.h>
-#include <capnp/compat/json-test.capnp.h>
+#include <capnp/json-test.capnp.h>
 #include <kj/debug.h>
 #include <kj/string.h>
 #include <kj/test.h>
--- a/c++/src/kj/async-unix.c++
+++ b/c++/src/kj/async-unix.c++
@@ -1051,7 +1051,7 @@ bool UnixEventPort::poll() {
         sigprocmask(SIG_SETMASK, &waitMask, &origMask);
         sigprocmask(SIG_SETMASK, &origMask, nullptr);
         break;
-#else
+#elif !defined(__EMSCRIPTEN__)
         sigsuspend(&waitMask);
         KJ_FAIL_ASSERT("sigsuspend() shouldn't return because the signal handler should "
                       "have siglongjmp()ed.");
--- a/c++/src/kj/exception.c++
+++ b/c++/src/kj/exception.c++
@@ -95,6 +95,10 @@ static void __lsan_ignore_object(const void* p) {}
 // TODO(cleanup): Remove the LSAN stuff per https://github.com/capnproto/capnproto/pull/1255
 // feedback.
 
+#if defined(__EMSCRIPTEN__)
+#include <exception>
+#endif
+
 namespace {
 template <typename T>
 inline T* lsanIgnoreObjectAndReturn(T* ptr) {
@@ -632,6 +636,7 @@ void printStackTraceOnCrash() {
   stack.ss_sp = reinterpret_cast<char*>(mmap(
       nullptr, stack.ss_size, PROT_READ | PROT_WRITE,
       MAP_ANONYMOUS | MAP_PRIVATE | MAP_GROWSDOWN, -1, 0));
+#if !defined(__EMSCRIPTEN__)
   KJ_SYSCALL(sigaltstack(&stack, nullptr));
 
   // Catch all relevant signals.
@@ -650,8 +655,9 @@ void printStackTraceOnCrash() {
 
   // Dump stack on unimplemented syscalls -- useful in seccomp sandboxes.
   KJ_SYSCALL(sigaction(SIGSYS, &action, nullptr));
+#endif
 
-#ifdef KJ_DEBUG
+#if defined(KJ_DEBUG) && !defined(__EMSCRIPTEN__)
   // Dump stack on keyboard interrupt -- useful for infinite loops. Only in debug mode, though,
   // because stack traces on ctrl+c can be obnoxious for, say, command-line tools.
   KJ_SYSCALL(sigaction(SIGINT, &action, nullptr));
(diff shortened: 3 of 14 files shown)


c++/src/kj/exception.c++
```diff
@@ -632,6 +636,7 @@ void printStackTraceOnCrash() {
   stack.ss_sp = reinterpret_cast<char*>(mmap(
       nullptr, stack.ss_size, PROT_READ | PROT_WRITE,
       MAP_ANONYMOUS | MAP_PRIVATE | MAP_GROWSDOWN, -1, 0));
+#if !defined(__EMSCRIPTEN__)
```
kentonv, line 639: Presumably this should be moved up to the top of this function -- there's no reason to allocate the stack if you're not going to use it.

kentonv: Or maybe `printStackTracesOnCrash()` should not be available at all in emscripten mode? The terminate handler is arguably relevant but mostly this function won't do what it's intended to do.

c++/src/capnp/compat/json-test.c++
```diff
@@ -22,7 +22,7 @@
 #include "json.h"
 #include <capnp/test-util.h>
 #include <capnp/compat/json.capnp.h>
-#include <capnp/compat/json-test.capnp.h>
+#include <capnp/json-test.capnp.h>
```
kentonv, line 25: This change presumably breaks the builds -- I guess your meson build is putting this file in the wrong place?

c++/src/kj/async-unix.c++
```diff
@@ -1051,7 +1051,7 @@ bool UnixEventPort::poll() {
         sigprocmask(SIG_SETMASK, &waitMask, &origMask);
         sigprocmask(SIG_SETMASK, &origMask, nullptr);
         break;
-#else
+#elif !defined(__EMSCRIPTEN__)
```
kentonv, line 1054: Seems like it'd be more appropriate to #if-out the entire block relating to signals, rather than just this inner part. The work to set up `sigsetjmp()` above is redundant.

c++/src/kj/exception.c++
```diff
@@ -95,6 +95,10 @@ static void __lsan_ignore_object(const void* p) {}
 // TODO(cleanup): Remove the LSAN stuff per https://github.com/capnproto/capnproto/pull/1255
 // feedback.
 
+#if defined(__EMSCRIPTEN__)
+#include <exception>
```
kentonv, line 99: This is already included unconditionally above.

c++/src/kj/exception.c++
```diff
@@ -1172,16 +1178,9 @@ struct FakeEhGlobals {
...
-// extern "C", the compiler wills get upset if we re-declare it even in a different namespace.
-#if _LIBCPPABI_VERSION
-extern "C" void* __cxa_get_globals();
-#else
-using abi::__cxa_get_globals;
-#endif
 
 uint uncaughtExceptionCount() {
-  return reinterpret_cast<FakeEhGlobals*>(__cxa_get_globals())->uncaughtExceptions;
+  return std::uncaught_exceptions();
```
kentonv, line 1183: Please use an `#if` here, otherwise this probably breaks some older environments.

c++/src/kj/filesystem-disk-unix.c++
```diff
@@ -388,18 +388,18 @@ public:
...
-      case EOPNOTSUPP:
-        // fall back to below
-        break;
-      default:
-        KJ_FAIL_SYSCALL("fallocate(FALLOC_FL_PUNCH_HOLE)", error) { return; }
-    } else {
-      return;
-    }
-#endif
+// #if defined(FALLOC_FL_PUNCH_HOLE) && !(__ANDROID__ && __BIONIC__ && __ANDROID_API__ < 21)
```
kentonv, line 391: Please use `#if` instead of commenting out.

c++/src/kj/exception.c++
```diff
@@ -1243,7 +1242,19 @@ static kj::String demangleTypeName(const char* name) {
 }
 
 kj::String getCaughtExceptionType() {
+  #if !defined (__EMSCRIPTEN__)
```
kentonv, line 1245: Double-negatives are hard to think about. When an `if` has an `else`, I prefer to make the `if` branch always be the positive case to make it easier to read.

c++/src/kj/exception.c++
kentonv, line 639: In KJ code we usually don't use `defined()`, but instead rely on the fact that undefined identifiers evaluate to zero for the purpose of preprocessor directives. So I'd prefer `#if !__EMSCRIPTEN__` rather than `#if !defined(__EMSCRIPTEN__)`.

kentonv (commented):
Hi @germandiagogomez,

Thanks for working on this, but I'm not sure what action you want me to take on this PR. It looks like it's not intended to be merged as-is.

If you'd like to have these `#if`s merged, can you please submit a PR that just contains the mergeable parts?

Regarding the tests that fail, if you could provide the test output I may be able to provide hints on how to fix them. However, I do not have bandwidth to try out the emscripten build myself, sorry.

I went ahead and reviewed the C++ changes commit.

PR #1758, opened 2023-08-10
Add `writeMessages()` overload

vickykont:


c++/src/capnp/serialize-async.c++
```diff
@@ -367,6 +367,15 @@ kj::Promise<void> writeMessages(
   return writeMessages(output, messages);
 }
 
+kj::Promise<void> writeMessages(
+    kj::AsyncOutputStream& output, kj::ArrayPtr<kj::Own<MessageBuilder>> builders) {
+  auto messages = kj::heapArray<kj::ArrayPtr<const kj::ArrayPtr<const word>>>(builders.size());
+  for (auto i : kj::indices(builders)) {
+    messages[i] = builders[i]->getSegmentsForOutput();
+  }
+  return writeMessages(output, messages);
```
kentonv, line 376: For a moment I thought that there was a potential use-after-free because `messages` isn't being attached to the promise so it would go away before the promise finishes (the function that you copied, above, has the same issue).

But then I noticed that it's not a problem because `writeMessageImpl` converts the whole segments array into a `pieces` array upfront, and then only uses that.

It might be worth a comment both here and in the function you copied from:

```
// Note that we don't have to `.attach(messages)` because `writeMessageImpl()` consumes the array upfront.
```

kentonv (approved):

c++/src/capnp/serialize-async.c++
vickykont: I added the comment.

PR #1760, opened 2023-08-17
Allow directory `openFile()` on Windows

mrbbot:
Hey! 👋 This PR adds support for opening directories using `openFile()`. This is used by `workerd` to support directory listing in disk directory services: https://github.com/cloudflare/workerd/blob/cae438f99d3e070e5530dbf42f338ee3392e68ae/src/workerd/server/server.c%2B%2B#L796-L802. Without this change, you'll get an error that looks like `kj/filesystem-disk-win32.c++:676: failed: CreateFile(path, OPEN_EXISTING): #5 Access is denied.`.

### DIFF
--- a/c++/src/kj/filesystem-disk-test.c++
+++ b/c++/src/kj/filesystem-disk-test.c++
@@ -449,6 +449,12 @@ KJ_TEST("DiskDirectory") {
     auto stats = dir->lstat(Path("bar"));
     KJ_EXPECT(stats.type == FsNode::Type::DIRECTORY);
   }
+  
+  {
+    auto subdir = dir->openFile(Path("bar"));
+    auto stats = subdir->stat();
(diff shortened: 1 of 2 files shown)


kentonv:
Hmmmmmmm arguably `workerd` is doing something that should never have worked, but works by accident on Unix.

I'm a little worried that `FILE_FLAG_BACKUP_SEMANTICS` may have side effects that we don't really want. The docs say:

> The file is being opened or created for a backup or restore operation. The system ensures that the calling process overrides file security checks when the process has SE_BACKUP_NAME and SE_RESTORE_NAME privileges. For more information, see Changing Privileges in a Token.

I don't really know the implications of that.

Maybe we need to have a `tryOpenAny()` method which returns `kj::Own<kj::FsNode>`, and then you can downcast it to the proper type after calling `stat()` on that.

PR #1761, opened 2023-08-18
await on ForkedPromise

mikea:
lightweight API to save on heap allocations.

### DIFF
--- a/c++/src/kj/async-inl.h
+++ b/c++/src/kj/async-inl.h
@@ -795,7 +795,7 @@ class ForkBranchBaseT: public ForkBranchBase {
 public:
   ForkBranchBaseT(OwnForkHubBase&& hub): ForkBranchBase(kj::mv(hub)) {}
 
- void get(ExceptionOrValue& output) noexcept override {
+  void get(ExceptionOrValue& output) noexcept override {
     ExceptionOr<T>& hubResult = getHubResultRef().template as<T>();
     KJ_IF_MAYBE(value, hubResult.value) {
       output.as<T>().value = copyOrAddRef(*value);
--- a/c++/src/kj/async-inl.h
+++ b/c++/src/kj/async-inl.h
@@ -791,7 +791,7 @@ template <typename T> Maybe<Own<T>> copyOrAddRef(Maybe<Own<T>>& t) {
 
 
 template <typename T>
-class  ForkBranchBaseT: public ForkBranchBase {
+class ForkBranchBaseT: public ForkBranchBase {
 public:
   ForkBranchBaseT(OwnForkHubBase&& hub): ForkBranchBase(kj::mv(hub)) {}
 
--- a/c++/src/kj/async-inl.h
+++ b/c++/src/kj/async-inl.h
@@ -824,7 +824,7 @@ class ForkedPromiseNode final: public ForkBranchBaseT<T> {
   // allocated on the stack.
 
 public:
-  ForkedPromiseNode(ForkedPromise<UnfixVoid<T>>&& promise)
+  ForkedPromiseNode(ForkedPromise<UnfixVoid<T>>& promise)
   : ForkBranchBaseT<T>(promise.hub->addRef()) {}
 
   void destroy() override { /* do nothing */ }
@@ -2277,8 +2277,8 @@ public:
 
   // called by co_awaiting on a forked promise.
   template <typename U>
-  ForkedPromiseAwaiter<U> await_transform(ForkedPromise<U>&& promise) { 
-    return ForkedPromiseAwaiter<U>(kj::mv(promise)); 
+  ForkedPromiseAwaiter<U> await_transform(ForkedPromise<U>& promise) { 
+    return ForkedPromiseAwaiter<U>(promise); 
   }
 
   void fulfill(FixVoid<T>&& value) {
@@ -2396,8 +2396,8 @@ template <typename T>
 template <typename U>
 class Coroutine<T>::ForkedPromiseAwaiter {
 public:
-  ForkedPromiseAwaiter(ForkedPromise<U>&& promise) 
-  : node(kj::mv(promise)), awaiter(OwnPromiseNode(&node)) { }
+  ForkedPromiseAwaiter(ForkedPromise<U>& promise) 
+  : node(promise), awaiter(OwnPromiseNode(&node)) { }
 
   template <typename V>
   inline bool await_suspend(stdcoro::coroutine_handle<V> coroutine) {
--- a/c++/src/kj/async-inl.h
+++ b/c++/src/kj/async-inl.h
@@ -789,16 +789,13 @@ template <typename T> Maybe<Own<T>> copyOrAddRef(Maybe<Own<T>>& t) {
   });
 }
 
+
 template <typename T>
-class ForkBranch final: public ForkBranchBase {
-  // A PromiseNode that implements one branch of a fork -- i.e. one of the branches that receives
-  // a const reference.
-
+class  ForkBranchBaseT: public ForkBranchBase {
 public:
-  ForkBranch(OwnForkHubBase&& hub): ForkBranchBase(kj::mv(hub)) {}
-  void destroy() override { freePromise(this); }
+  ForkBranchBaseT(OwnForkHubBase&& hub): ForkBranchBase(kj::mv(hub)) {}
 
-  void get(ExceptionOrValue& output) noexcept override {
+ void get(ExceptionOrValue& output) noexcept override {
     ExceptionOr<T>& hubResult = getHubResultRef().template as<T>();
     KJ_IF_MAYBE(value, hubResult.value) {
       output.as<T>().value = copyOrAddRef(*value);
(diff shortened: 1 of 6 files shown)


c++/src/kj/async-inl.h
```diff
@@ -2286,7 +2312,10 @@ class CoroutineBase::AwaiterBase {
 
 private:
   UnwindDetector unwindDetector;
-  OwnPromiseNode node;
+
+  // only one of these is set
```
kentonv, line 2316: You might be able to get away with using `OwnPromiseNode` for both. The trick is that `OwnPromiseNode` disposes the `PromiseNode` by calling its `destroy()` method. In this particular case, you can have the implementation of `destroy()` do nothing.

mikea:
Would be glad to hear alternative names suggestions.

c++/src/kj/async.h
```diff
@@ -388,6 +393,12 @@ class ForkedPromise {
   Promise<T> addBranch();
   // Add a new branch to the fork.  The branch is equivalent to the original promise.
 
+  _::BranchCoroPromise<T> addCoroBranch();
```
kentonv, line 396: Maybe `addBranchForCoAwait()`? The important thing being communicated here is: the caller is expected to immediately co_await the result.

c++/src/kj/async.h
```diff
@@ -388,6 +393,12 @@ class ForkedPromise {
...
+  // Add a new branch to the fork. 
+  // A variant of addBranch returning a coro-only promise-like value object.
+  // The return type is not supposed to be named directly: auto/co_await are the only way to handle
+  // the value.
```
kentonv, line 400:
```suggestion
  // Like addBranch(), but use this version when you will immediately `co_await` the result.
  // This version avoids allocation by leveraging the coroutine frame. You must not store the
  // result, the only thing you may do with it is immediately `co_await` it.
```

c++/src/kj/async-inl.h
```diff
@@ -913,6 +940,22 @@ inline ExceptionOrValue& ForkBranchBase::getHubResultRef() {
   return hub->getResultRef();
 }
 
+
+template <typename T>
+class BranchCoroPromise {
+  // coro-only promise-like value object for addCoroBranch() implementation
+
+public:
+  KJ_DISALLOW_COPY(BranchCoroPromise);
```
kentonv, line 949: I wish C++ had some feature where you could say "if this value is to be copied or moved, convert it to a different type", so that we could convert to `Promise<T>` in that case. Then we wouldn't need a separate method, we could auto-detect when the caller is going to co_await immediately.

But I don't think it does.

bcaimano: Can you give some example code on what you'd want the behavior to be? Move and copy semantics are surprisingly deep (but sometimes the edge cases hurt).

c++/src/kj/async-inl.h
```diff
@@ -2247,6 +2295,9 @@ class Coroutine final: public CoroutineBase,
   // a type-erased `await_suspend(stdcoro::coroutine_handle<void>)` override, and implement
   // suspension and resumption in terms of .then(). Yuck!
 
+  template <typename U>
+  Awaiter<U> await_transform(BranchForCoAwaitPromise<U>&& promise) { return Awaiter<U>(promise); }
```
harrishancock, line 2299: We might need an lvalue reference overload, too, unless we know we'll never store a branch as an lvalue before awaiting it.

Edit: I see that this is actually by design, but I'm not sure it's necessary -- see my comment below `addBranchForCoAwait()`.

c++/src/kj/async-inl.h
```diff
@@ -2326,7 +2378,8 @@ class Coroutine<T>::Awaiter: public AwaiterBase {
   // awaited promise result.
 
 public:
-  explicit Awaiter(Promise<U> promise): AwaiterBase(PromiseNode::from(kj::mv(promise))) {}
+  explicit Awaiter(Promise<U>&& promise): AwaiterBase(PromiseNode::from(kj::mv(promise))) {}
+  explicit Awaiter(BranchForCoAwaitPromise<U>& branch): AwaiterBase(branch.node()) {}
```
harrishancock, line 2382: It might be reasonable to move the calls to `PromiseNode::from()` and `branch.node()` into `await_transform()`, and have `Awaiter` accept only `OwnPromiseNode`, so that adding new awaitables like this doesn't require modifying both Coroutine and Coroutine::Adapter.

c++/src/kj/async.h
```diff
@@ -388,6 +393,12 @@ class ForkedPromise {
   Promise<T> addBranch();
   // Add a new branch to the fork.  The branch is equivalent to the original promise.
 
+  _::BranchForCoAwaitPromise<T> addBranchForCoAwait();
+  // Add a new branch to the fork. 
+  // Like addBranch(), but use this version when you will immediately `co_await` the result.
+  // This version avoids allocation by leveraging the coroutine frame. You must not store the
+  // result, the only thing you may do with it is immediately `co_await` it.
```
harrishancock, line 400: Isn't the requirement that we must not _move_ the result, rather than not store it? If so, could we make it `KJ_DISALLOW_COPY_AND_MOVE` and rely on guaranteed copy elision to make sure wherever it gets stored -- either as a temporary in `co_await fork.addBranch...()` or as an lvalue in `auto branch = fork.addBranch...()` -- never changes addresses?

harrishancock (commented):
Instead of adding a new `addBranchForCoAwait()` function, can we instead make `kj::ForkedPromise<T>` directly awaitable?

That is:
1. Add an `await_transform()` overload which accepts a `kj::ForkedPromise<U>&` and returns a `Coroutine<T>::ForkBranchAwaiter<U>`.
2. `Coroutine<T>::ForkBranchAwaiter<U>` would contain the existing `ForkBranchForCoAwait<U>` and a `Coroutine<T>::Awaiter<U>`, with the latter data member initialized from the former's PromiseNode.

Then we could write `co_await fork` instead of `co_await fork.addBranchForCoAwait()`, and it would be impossible for authors to accidentally store the result of that `addBranchForCoAwait()`.

mikea:
@harrishancock how's this?

c++/src/kj/async-inl.h
```diff
@@ -789,16 +790,13 @@ template <typename T> Maybe<Own<T>> copyOrAddRef(Maybe<Own<T>>& t) {
   });
 }
 
-template <typename T>
-class ForkBranch final: public ForkBranchBase {
-  // A PromiseNode that implements one branch of a fork -- i.e. one of the branches that receives
-  // a const reference.
 
+template <typename T>
+class  ForkBranchBaseT: public ForkBranchBase {
```
mikea, line 795: I decided to keep the ForBranchBase name since it is used quite a lot @harrishancock

c++/src/kj/async-inl.h
```diff
@@ -789,16 +789,13 @@ template <typename T> Maybe<Own<T>> copyOrAddRef(Maybe<Own<T>>& t) {
...
```
kentonv, line 794:
```suggestion
class ForkBranchBaseT: public ForkBranchBase {
```

c++/src/kj/async-inl.h
```diff
@@ -2355,6 +2390,29 @@ class Coroutine<T>::Awaiter: public AwaiterBase {
...
 };
 
+// Wait for forked promise.
+// Delegate all the work to usual awaiter on a special node.
+template <typename T>
+template <typename U>
+class Coroutine<T>::ForkedPromiseAwaiter {
+public:
+  ForkedPromiseAwaiter(ForkedPromise<U>&& promise) 
+  : node(kj::mv(promise)), awaiter(OwnPromiseNode(&node)) { }
```
kentonv, line 2400:
```suggestion
      : node(kj::mv(promise)), awaiter(OwnPromiseNode(&node)) { }
```

kentonv (commented):
Sorry, whitespace issues triggered my OCD here. (This batch of review comments is whitespace-only.)

c++/src/kj/async-inl.h
```diff
@@ -2247,6 +2272,15 @@ class Coroutine final: public CoroutineBase,
   // a type-erased `await_suspend(stdcoro::coroutine_handle<void>)` override, and implement
   // suspension and resumption in terms of .then(). Yuck!
 
+  template <typename U>
+  class ForkedPromiseAwaiter;
+
+  // called by co_awaiting on a forked promise.
+  template <typename U>
+  ForkedPromiseAwaiter<U> await_transform(ForkedPromise<U>&& promise) {
```
kentonv, line 2280: I think the idea is that this should take an lvalue reference, right?

Otherwise the ForkedPromise can only be `co_await`ed once, which defeats the purpose of `ForkedPromise`.

mikea: the idea was actually to have rvalue, and await on the promise once all the branches are already established:

```
auto fork = promise.fork();
auto branch = promise.branch();
...

co_await kj::mv(fork);
```

mikea: but I can easily redo this as an rvalue if you think it would be better.

kentonv: I believe harris's idea was in fact to accept an lvalue reference.

It doesn't seem like there's any benefit to requiring an rvalue here? We could accept an lvalue with exactly the same implementation, but it would then allow co_awating the forked promise multiple times.

mikea: done.

c++/src/kj/async-test.c++
```diff
@@ -603,6 +603,20 @@ TEST(Async, ForkMaybeRef) {
...
 
+KJ_TEST("addBranchForCoAwait") {
+  EventLoop loop;
+  WaitScope waitScope(loop);
+
+  Promise<int> promise = evalLater([&]() { return 123; });
+
+  auto coro = [&]() -> kj::Promise<int> {
+    auto fork = promise.fork();
+    fork.addBranch();
```
kentonv, line 614: This line has no effect?

PR #1762, opened 2023-08-21
Replace KJ_STACK_ARRAY with kj::SboArray and kj::SboArrayBuilder

harrishancock:
We'd like to be able to use KJ_STACK_ARRAY in coroutines, but KJ_STACK_ARRAY uses variable-length arrays on compilers which support them (that is: GCC). ~~So, we're dropping support for VLAs from KJ_STACK_ARRAY.~~

It's worth noting that at present, we don't actually support GCC, so we have no target which even uses VLAs anymore. But, we'd like to support GCC again in the future, once its coroutine support is fixed.

Update: Instead of dropping support for VLAs from KJ_STACK_ARRAY, we're replacing the macro entirely with `kj::SboArray`.

### DIFF
--- a/c++/src/kj/array.h
+++ b/c++/src/kj/array.h
@@ -273,14 +273,17 @@ public:
 
   static const HeapArrayDisposer instance;
 
-private:
   static void* allocateImpl(size_t elementSize, size_t elementCount, size_t capacity,
                             void (*constructElement)(void*), void (*destroyElement)(void*));
   // Allocates and constructs the array.  Both function pointers are null if the constructor is
   // trivial, otherwise destroyElement is null if the constructor doesn't throw.
+  //
+  // TODO(cleanup): Public so we can pass it as a pointer elsewhere.
 
-  virtual void disposeImpl(void* firstElement, size_t elementSize, size_t elementCount,
-                           size_t capacity, void (*destroyElement)(void*)) const override;
+private:
+  void disposeImpl(void* firstElement, size_t elementSize, size_t elementCount,
+                   size_t capacity, void (*destroyElement)(void*)) const override;
+};
 
   template <typename T, bool hasTrivialConstructor = KJ_HAS_TRIVIAL_CONSTRUCTOR(T),
                         bool hasNothrowConstructor = KJ_HAS_NOTHROW_CONSTRUCTOR(T)>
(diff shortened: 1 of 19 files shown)


harrishancock:
Bazel builds broke because a new version of zlib came out.

It's kind of nice that we're notified of new versions of zlib due to breaking builds, but also kind of not, so I updated the link to a stable URL.

kentonv:
As discussed, at this point we could make this a template type that wouldn't have to be stack-specific (indeed, that is the whole motivation -- coroutines aren't stack frames). Maybe `SmallArray<T, n>`?

harrishancock:
> As discussed, at this point we could make this a template type that wouldn't have to be stack-specific (indeed, that is the whole motivation -- coroutines aren't stack frames). Maybe SmallArray<T, n>?

I've updated this PR with an implementation of what I called `SboArray<T, n>`, and a related `SboArrayBuilder<T, n>`. (I'm happy to switch to SmallArray, I forgot about that suggestiohn.)

I implemented the new types in terms of a new `_::SboArrayDisposer` which is similar in shape, and shares some implementation with, `HeapArrayDisposer`. I did this because I wanted to make their behaviors as similar as possible to `heapArray()` and `heapArrayDisposer()`, which are the implementations we fall back to if the maximum inline size is exceeded. This may have ended up being overkill, though.

While testing, I noticed that `~Array()` is `noexcept`, while `~ArrayBuilder()` is `noexcept(false)`. Should they both instead be `      noexcept(noexcept(instance<T&>().~T()))`?

kentonv:
> I noticed that ~Array() is noexcept

Wha... that seems like a huge mistake, surprised I never noticed.

c++/src/kj/array.h
```diff
@@ -573,6 +601,81 @@ class CappedArray {
...
+// 2. SboArray/Builder are immovable. You must construct them in place wherever you want to use one.
+//    They cannot be "released", "finished", or assigned-to.
+//
+// 3. SboArray/Builder have no specific constructor functions like `heapArray<T>()` or
+//    `heapArrayBuilder<T>()`. Instead, use their constructors directly, passing a single `size`
+//    parameter.
+
+
+template <typename T, size_t smallSize>
+class SboArray: private _::SboArrayDisposer<T, smallSize>,
```
kentonv: Wow... this is... quite an implementation. I would not have thought of this, but it turns out pretty concise.

I guess inheriting from SboArrayDisposer forces the SboArray to contain an extra pointer for the virtual table for this disposer. I think if you placed `T space[smallSize];` inside `SboArray` itself, then you could reference a singleton disposer, eliminating this pointer?

harrishancock: Yeah, that's a better idea. I pushed 5e8076992236819efdd9aec1e6094cf619853803 with that change, plus a few more cleanup commits to simplify the allocation logic.

harrishancock:
Force-pushes were mostly rebases on `v2` and renaming `SboArray` to `SmallArray`. Additionally, I reverted my attempted modernization of the disposal/allocation implementation details -- my simplification caused `heapArray<const byte>()` to become a compile error in workerd, somehow.

PR #1763, opened 2023-08-21
Optimize JSON memory management

fhanau:
Implements some memory management optimizations for JSON parsing. While more extensive changes will be needed for further gains in performance, these changes should be relatively easy to review and perform better across use cases.
On a [simple benchmark](https://github.com/cloudflare/workerd/blob/main/src/workerd/tests/bench-json.c%2B%2B) (likely not very representative, but good enough to show impact), I'm seeing an improvement of ~16% for both encoding and parsing (json encode 3.64 => 3.04 us, json parse 4.91 => 4.12 us).

### DIFF
--- a/c++/src/capnp/compat/json.c++
+++ b/c++/src/capnp/compat/json.c++
@@ -132,6 +132,8 @@ struct JsonCodec::Impl {
     escaped.add('\0');
 
     return kj::String(escaped.releaseAsArray());
+    // optimization: If there are no escaped characters, the allocated buffer is exactly the right
+    // size and releaseAsArray does not need to allocate.
   }
 
   kj::StringTree encodeList(kj::Array<kj::StringTree> elements,
@@ -190,7 +192,10 @@ void JsonCodec::setHasMode(HasMode mode) { impl->hasMode = mode; }
 void JsonCodec::setRejectUnknownFields(bool enabled) { impl->rejectUnknownFields = enabled; }
 
 kj::String JsonCodec::encode(DynamicValue::Reader value, Type type) const {
-  MallocMessageBuilder message;
+  MallocMessageBuilder message(128);
+  // Use a smaller initial segment size, this significantly improves performance when encoding
+  // short strings.
+
   auto json = message.getRoot<JsonValue>();
   encode(value, type, json);
   return encodeRaw(json);
@@ -745,9 +750,19 @@ public:
 private:
   kj::String consumeQuotedString() {
     input.consume('"');
-    // TODO(perf): Avoid copy / alloc if no escapes encountered.
     // TODO(perf): Get statistics on string size and preallocate?
+
+    auto initalString = input.consumeWhile([](const char chr) {
+        return chr != '"' && chr != '\\';
+    });
+    if (input.nextChar() == '"') {
+      // optimization: Avoid allocating vector if no escapes encountered.
+      input.advance();
+      return kj::heapString(initalString.begin(), initalString.size());
+    }
+
     kj::Vector<char> decoded;
+    decoded.addAll(initalString);
 
     do {
       auto stringValue = input.consumeWhile([](const char chr) {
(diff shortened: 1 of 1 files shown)


c++/src/capnp/compat/json.c++
```diff
@@ -804,11 +814,7 @@ private:
 
     KJ_REQUIRE(numArrayPtr.size() > 0, "Expected number in JSON input.");
 
-    kj::Vector<char> number;
-    number.addAll(numArrayPtr);
-    number.add('\0');
-
-    return kj::String(number.releaseAsArray());
+    return kj::heapString(numArrayPtr.begin(), numArrayPtr.size());
```
kentonv, line 817: I wonder why this function is returning a string at all. Why not parse the number to `double` here, to avoid the need for a heap allocation?

kentonv:
Can you put the benchmark directly in `json-test.c++`, using KJ's existing benchmarking framework? (See `capnp/compat/http-over-capnp-perf-test.c++` for an example benchmark.)

c++/src/capnp/compat/json.c++
```diff
@@ -129,9 +129,8 @@ struct JsonCodec::Impl {
       }
     }
     escaped.add('"');
-    escaped.add('\0');
 
-    return kj::String(escaped.releaseAsArray());
+    return kj::heapString(escaped.begin(), escaped.size());
```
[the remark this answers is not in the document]
fhanau: I see – I wasn't fully familiar with kj's classes and assumed that avoiding another vector append would be useful, but I understand that here the original approach works better.

c++/src/capnp/compat/json.c++
```diff
@@ -745,9 +747,19 @@ public:
...
+    auto initalString = input.consumeWhile([](const char chr) {
+        return chr != '"' && chr != '\\';
+    });
+    if (input.nextChar() == '"') {
+      // optimization: Avoid allocating vector if no escapes encountered.
+      input.advance();
+      return kj::heapString(initalString.begin(), initalString.size());
+    }
+
     kj::Vector<char> decoded;
```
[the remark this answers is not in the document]
fhanau: Yes, that's pretty much what I understood the `Get statistics on string size and preallocate?` TODO to mean. It's certainly something worth exploring, for now I would like to keep this PR small and explore further optimizations later on.

c++/src/capnp/compat/json.c++
```diff
@@ -778,10 +790,8 @@ private:
     } while(input.nextChar() != '"');
 
     input.consume('"');
-    decoded.add('\0');
 
-    // TODO(perf): This copy can be eliminated, but I can't find the kj::wayToDoIt();
-    return kj::String(decoded.releaseAsArray());
+    return kj::heapString(decoded.begin(), decoded.size());
```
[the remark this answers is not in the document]
fhanau: Makes sense, reverted the change.

c++/src/capnp/compat/json.c++
fhanau: `parseDouble` takes a StringPtr, so this will require either an allocation or adding an API for ArrayPtr-based parsing. I did run a quick test and doing the latter improves the benchmark result by 1.5%, I added a TODO for now.

fhanau:
> Can you put the benchmark directly in `json-test.c++`, using KJ's existing benchmarking framework? (See `capnp/compat/http-over-capnp-perf-test.c++` for an example benchmark.)

I'll work on this soon – the JSON test case was mistakenly not included in the bazel build and fails to build with bazel. I have it working locally, but need to clean up the build system fixes.

c++/src/capnp/compat/json.c++
kentonv: Could use `strtod()` directly to parse from an array.

c++/src/capnp/compat/json.c++
```diff
@@ -804,11 +819,8 @@ private:
 
     KJ_REQUIRE(numArrayPtr.size() > 0, "Expected number in JSON input.");
 
-    kj::Vector<char> number;
-    number.addAll(numArrayPtr);
-    number.add('\0');
-
-    return kj::String(number.releaseAsArray());
+    return kj::heapString(numArrayPtr.begin(), numArrayPtr.size());
+    // TODO: parse double here directly to avoid memory allocation
```
kentonv, line 823: Nitpick

```suggestion
    // TODO(perf): parse double here directly to avoid memory allocation
```

kentonv (approved):

c++/src/capnp/compat/json.c++
fhanau: The current approach ends up calling `NoLocaleStrtod` in string.cpp, which provides locale handling, I assumed that's also needed for JSON.

PR #1764, opened 2023-08-21
Update zlib

fhanau:
The previous version is no longer available, causing CI build failures. This also fixes some of the issues with compiling on macOS, but we still need to disable a warning.


kentonv (approved):
LGTM but I think @harrishancock has his own fix, might want to check in with him.

Or... maybe it's fine to just merge and sort it out later, I dunno.

fhanau:
Harris' PR fixes the issue too, I prefer having this in as it updates a comment that would otherwise be misleading and narrows the scope of the macOS workaround.

PR #1767, opened 2023-08-24
Disallow new and delete

ohodson:


kentonv:
I guess you already withdrew this, but note that this approach unfortunately doesn't work: `requireOnStack()` will prevent a type from being used as a member variable of some other type which is itself allocated on the heap. Disabling new and delete won't prevent that, because the type isn't being newed/deleted directly in this case.

I do wish we had some way to check this statically but I haven't been able to find one.

PR #1769, opened 2023-08-28
mvsc: Un-bitrot coroutines

hornang:
If it is okay to backport this from the `v2` branch


kentonv:
I think we need to make sure CI tests actually cover this, otherwise it'll bitrot again.

hornang:
> I think we need to make sure CI tests actually cover this, otherwise it'll bitrot again.

All right. I'll close this one then.

PR #1770, opened 2023-08-29
Ensure TaskSet destructor unlinks Tasks, even when an exception occurs

jclee:
Should reduce the likelihood of stack overflow in cases involving very large TaskSets.

### DIFF
--- a/c++/src/kj/async-test.c++
+++ b/c++/src/kj/async-test.c++
@@ -881,22 +881,12 @@ TEST(Async, TaskSet) {
 }
 
 #if KJ_USE_FIBERS || !_WIN32
-// This test requires either fibers or pthreads in order to limit the stack size. Currently we
-// don't have a version that works on Windows without fibers, so skip the test there.
+// These tests require either fibers or pthreads in order to limit the stack size. Currently we
+// don't have a version that works on Windows without fibers, so skip the tests there.
 
-TEST(Async, LargeTaskSetDestruction) {
-  static constexpr size_t stackSize = 200 * 1024;
-
-  static auto testBody = [] {
-
-    ErrorHandlerImpl errorHandler;
-    TaskSet tasks(errorHandler);
-
-    for (int i = 0; i < stackSize / sizeof(void*); i++) {
-      tasks.add(kj::NEVER_DONE);
-    }
-  };
-
+template <typename Func>
+void runWithStackLimit(size_t stackSize, Func&& func) {
+  // Runs the given function in a context with a limited stack size.
 #if KJ_USE_FIBERS
   if (isLibcContextHandlingKnownBroken()) return;
 
@@ -904,10 +894,9 @@ TEST(Async, LargeTaskSetDestruction) {
   WaitScope waitScope(loop);
 
   startFiber(stackSize,
-      [](WaitScope&) mutable {
-    testBody();
+      [func = kj::mv(func)](WaitScope&) mutable {
+    func();
   }).wait(waitScope);
-
 #else
   pthread_attr_t attr;
   KJ_REQUIRE(0 == pthread_attr_init(&attr));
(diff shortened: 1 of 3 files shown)


c++/src/kj/async.c++
```diff
@@ -364,6 +364,16 @@ TaskSet::~TaskSet() noexcept(false) {
   // task to be able to schedule new tasks. So when we cancel our tasks... we might find new
   // tasks added! We'll have to repeatedly cancel. Additionally, we need to make sure that we destroy
   // the items in a loop to prevent any issues with stack overflow.
+
+  KJ_DEFER({
+    // If an exception occurs, ensure the remaining tasks still get unlinked.
+    while (tasks != nullptr) {
+      kj::runCatchingExceptions([&]() {
```
jclee, line 371: I suppose another approach would be to use `kj::UnwindDetector` here?  But I wasn't sure whether it would preferable, since the current approach avoids adding a field to `TaskSet`.

jclee: Ah, looks like I can just have `UnwindDetector` on the stack, which seems like a better approach... updated to use that instead.

jclee:
While looking for a better approach, I noticed that `TaskSet::clear()` suffers from similar problems as the destructor, so I updated the PR to cover it as well.

Note that the current PR changes the behavior of `clear()` a little; unlike the logic in `~TaskSet()`, `clear()` would just cancel the current set of tasks, but would allow cancellation logic to queue new tasks that would not be cancelled prior to calling the empty fulfiller.  Not sure if this might be considered desirable behavior or a bug.  In the current PR, both `~TaskSet()` and `clear()` use the same logic, waiting for all tasks to be cancelled before continuing.

c++/src/kj/async.c++
```diff
@@ -412,13 +406,34 @@ Promise<void> TaskSet::onEmpty() {
...
+  // You could argue it is dubious, but some applications would like for the destructor of a
+  // task to be able to schedule new tasks. So when we cancel our tasks... we might find new
+  // tasks added! We'll have to repeatedly cancel. Additionally, we need to make sure that we destroy
+  // the items in a loop to prevent any issues with stack overflow.
+
+  UnwindDetector unwindDetector;
+  KJ_DEFER({
+    // If an exception occurs, ensure the remaining tasks still get unlinked.
+    while (tasks != nullptr) {
+      unwindDetector.catchExceptionsIfUnwinding([&]() {
```
kentonv, line 426: I suppose we can confidently say that we are _always_ unwinding here? Because if there were no exceptions, then `tasks` would always be `nullptr` before the defer starts. Maybe we should just use runCatchingExceptions() directly?

kentonv (approved):

c++/src/kj/async.c++
jclee: Maybe?  I was originally calling `runCatchingExceptions()` directly, but wasn't sure if it was OK to arbitrarily drop exceptions on the floor.  I noticed that `UnwindDetector::catchExceptionsIfUnwinding()` calls `UnwindDetector::catchThrownExceptionAsSecondaryFault()`, which has a TODO to eventually capture secondary exceptions, and that it goes on to call `getCaughtExceptionAsKj()`, which has special logic rethrowing `CanceledException` and `TopLevelProcessContext::CleanShutdownException`, which are apparently not meant to be caught... although that does potentially conflict with this specific use case of large task set cleanup.

jclee: Oh, I guess `runCatchingExceptions()` also ends up calling `getCaughtExceptionAsKj()`, so maybe that's OK.

jclee: Per discussion, went ahead and just removed the secondary exception catching during the defer, since the defer should only do something during stack unwind, and we already assume that destructors will not throw during stack unwind.

c++/src/kj/async-test.c++
```diff
@@ -881,50 +881,114 @@ TEST(Async, TaskSet) {
...
-
 #else
   pthread_attr_t attr;
   KJ_REQUIRE(0 == pthread_attr_init(&attr));
   KJ_DEFER(KJ_REQUIRE(0 == pthread_attr_destroy(&attr)));
 
   KJ_REQUIRE(0 == pthread_attr_setstacksize(&attr, stackSize));
   pthread_t thread;
-  KJ_REQUIRE(0 == pthread_create(&thread, &attr, [](void*) -> void* {
+  KJ_REQUIRE(0 == pthread_create(&thread, &attr, [func = kj::mv(func)](void*) -> void* {
```
harrishancock, line 907: I suspect this might not compile if this `#else` branch is taken, because `pthread_create()` expects a function pointer, but we're now giving it a C++ lambda object.

jclee: Wow, good catch...  I put up a fix: https://github.com/capnproto/capnproto/pull/1786

Also went ahead and made it exercise both the fiber- and pthread-based stack limit testing methods where available, which should reduce the likelihood of future breakage.

PR #1772, opened 2023-08-30
Websockets: ignore older pings in a few more cases

jclee:
May reduce websocket overhead a bit.

### DIFF
--- a/c++/src/kj/compat/http-test.c++
+++ b/c++/src/kj/compat/http-test.c++
@@ -2190,6 +2190,51 @@ KJ_TEST("WebSocket double-ping mid-send") {
   serverTask.wait(waitScope);
 }
 
+KJ_TEST("WebSocket multiple ping outside of send") {
+  KJ_HTTP_TEST_SETUP_IO;
+
+  auto upPipe = newOneWayPipe();
+  auto downPipe = newOneWayPipe();
+  InputOutputPair client(kj::mv(downPipe.in), kj::mv(upPipe.out));
+  auto server = newWebSocket(kj::heap<InputOutputPair>(kj::mv(upPipe.in), kj::mv(downPipe.out)),
+                             nullptr);
+
+  byte DATA[] = {
+    0x89, 0x05, 'p', 'i', 'n', 'g', '1',
+    0x89, 0x05, 'p', 'i', 'n', 'g', '2',
+    0x89, 0x05, 'p', 'i', 'n', 'g', '3',
+    0x89, 0x05, 'p', 'i', 'n', 'g', '4',
+    0x81, 0x05, 'o', 't', 'h', 'e', 'r',
+  };
+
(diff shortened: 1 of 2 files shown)


c++/src/kj/compat/http-test.c++
```diff
@@ -2190,6 +2190,51 @@ KJ_TEST("WebSocket double-ping mid-send") {
...
+  }
+
+  auto bigString = kj::strArray(kj::repeat(kj::StringPtr("12345678"), 65536), "");
+  auto serverTask = server->send(bigString).eagerlyEvaluate(nullptr);
+
+  // We expect to receive pongs for only the first and last pings, because the server has the
+  // option of only sending pongs for the most recently processed ping, and the last three pings
+  // were processed while waiting for the write of the first pong to complete.
+  byte EXPECTED1[] = {
+    0x8A, 0x05, 'p', 'i', 'n', 'g', '1',
```
jclee, line 2225: I imagine it might be possible to eliminate the first pong, too, but that might require making the pong writing lazier, which might affect performance and the order things happen in the event loop?

kentonv: Yeah it seems fine as-is.

c++/src/kj/compat/http.c++
```diff
@@ -3378,8 +3378,9 @@ private:
   kj::ArrayPtr<const byte> sendParts[2];
 
   kj::Maybe<kj::Array<byte>> queuedPong;
-  // If a Ping is received while currentlySending is true, then queuedPong is set to the body of
-  // a pong message that should be sent once the current send is complete.
+  // queuedPong holds the body of the next pong to write, cleared when the pong is written.  If a
+  // more recent ping arrives before the pong is actually written, we can update this value to
+  // instead respond to the more recent ping.
```
smerritt, line 3383: I don't know that this is true. If we're writing the current pong, we may have sent some, but not all, of its bytes down the socket. If we have a pong AAAAAAAA and we overwrite it with BBBBBBBB, we might end up sending AAAABBBB and confusing the other end. Plus. there could be confusion if the pongs are of different length.

Would it help to have a "sending" pong and an "on-deck" pong so the on-deck pong could be freely overwritten?

jclee: I think in this case the current implementation is OK, since `writeQueuedPong()` moves the pong content from `queuedPong` to `payload` just prior to the write call, then attaches `payload` to the write promise.  So `queuedPong` is effectively the "on-deck" pong, while the promise and its attached payload are effectively the "sending" pong, and there shouldn't be a case where a client could receive fragmented content from multiple pongs.

But I agree the comment could be confusing.  Perhaps it would be less confusing with "...cleared at the time the next pong write is set up" rather than "...cleared when the pong is written"?  Wasn't quite sure of the best way to phrase it because we have multiple levels of "queuing" going on.

PR #1773, opened 2023-08-30
Coroutine conversion for kj-http WebSocket implementations

harrishancock:


c++/src/kj/compat/http.c++
```diff
@@ -3593,47 +3594,66 @@ kj::Own<WebSocket> newWebSocket(kj::Own<kj::AsyncIoStream> stream,
...
+      co_return;
+    }
+  }
 }
 
 kj::Promise<void> WebSocket::pumpTo(WebSocket& other) {
   KJ_IF_SOME(p, other.tryPumpFrom(*this)) {
     // Yay, optimized pump!
-    return kj::mv(p);
+    co_return co_await kj::mv(p);
```
MellowYarker, line 3644: For my own understanding, what difference does it make if we `co_return co_await p` without moving here? Is there some problem with `co_await`ing the reference? Move makes sense in the prior code but I'm not sure it's required now...

MellowYarker (approved):
Changes LGTM. Thanks for the quick cleanups as well, easier to follow now.

c++/src/kj/compat/http.c++
```diff
@@ -3593,47 +3594,66 @@ kj::Own<WebSocket> newWebSocket(kj::Own<kj::AsyncIoStream> stream,
...
     KJ_UNREACHABLE;
-  }, [&to](kj::Exception&& e) {
-    if (e.getType() == kj::Exception::Type::DISCONNECTED) {
-      return to.disconnect();
-    } else {
-      return to.close(1002, e.getDescription());
+  };
+
+  for (;;) {
+    auto message = co_await from.receive();
```
jclee, line 3613: Not sure if important, but I think that if the `from.receive()` promise threw an exception, the old code would have run one of the exception handlers, whereas the new code would just throw directly?

PR #1774, opened 2023-08-30
http: lazy output stream queue

mikea:
rather than branching on writing every chunk, use Maybe<Promise> to wait on previous writes.

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -2198,7 +2198,7 @@ public:
     return broken;
   }
 
-  void writeHeaders(String content) {
+  void queueHeaders(String content) {
     // Writes some header content and begins a new entity body.
 
     KJ_REQUIRE(!writeInProgress, "concurrent write()s not allowed") { return; }
@@ -2208,7 +2208,7 @@ public:
     queueWrite(kj::mv(content));
   }
 
-  void writeBodyData(kj::String content) {
+  void queueBodyData(kj::String content) {
     KJ_REQUIRE(!writeInProgress, "concurrent write()s not allowed") { return; }
     KJ_REQUIRE(inBody) { return; }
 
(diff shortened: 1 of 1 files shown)


c++/src/kj/compat/http.c++
```diff
@@ -2541,31 +2549,32 @@ public:
 
   Maybe<Promise<uint64_t>> tryPumpFrom(AsyncInputStream& input, uint64_t amount) override {
     KJ_IF_MAYBE(l, input.tryGetLength()) {
-      // Hey, we know exactly how large the input is, so we can write just one chunk.
-
-      uint64_t length = kj::min(amount, *l);
-      auto& inner = getInner();
-      inner.writeBodyData(kj::str(kj::hex(length), "\r\n"));
```
mikea, line 2548: these were actually calls to queue the write. Was it warranted? I replaced these by direct writes below.

c++/src/kj/compat/http.c++
```diff
@@ -2309,9 +2297,10 @@ public:
   }
 
   kj::Promise<void> flush() {
-    auto fork = writeQueue.fork();
-    writeQueue = fork.addBranch();
-    return fork.addBranch();
+    KJ_IF_MAYBE(promise, writeQueue) {
+      co_await *promise;
```
kentonv, line 2301: If awaiting `promise` throws, then `writeQueue` won't be nulled out, but will contain a moved-away promise, so the next use will crash.

Perhaps we should define `flush` as:

```
kj::Promise<void> flush() {
  KJ_IF_MAYBE(promise, writeQueue) {
    auto result = kj::mv(*promise);
    writeQueue = nullptr;
    return result;
  } else {
    return kj::READY_NOW;
  }
}
```

This approach also avoids allocating a coroutine frame for every call, so should be more efficient.

kentonv (approved):

PR #1775, opened 2023-08-31
convert HttpFixedLengthEntityWriter to coroutines

mikea:
I was hoping to avoid allocation in maybeFinishAfter but ended up converting it into coros.

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -2410,25 +2410,32 @@ public:
   }
 
   Promise<void> write(const void* buffer, size_t size) override {
-    if (size == 0) return kj::READY_NOW;
+    if (size == 0) co_return;
     KJ_REQUIRE(size <= length, "overwrote Content-Length");
     length -= size;
 
-    return maybeFinishAfter(getInner().writeBodyData(buffer, size));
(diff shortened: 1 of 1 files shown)


c++/src/kj/compat/http.c++
```diff
@@ -2446,35 +2453,28 @@ public:
...
       length += amount - actual;
       if (length == 0) doneWriting();
-      return actual;
-    });
+    }
 
     if (overshot) {
-      promise = promise.then([amount,&input](uint64_t actual) -> kj::Promise<uint64_t> {
         if (actual == amount) {
           // We read exactly the amount expected. In order to detect an overshoot, we have to
```
kentonv, line 2466: Not for this PR, but it appears there's a pre-existing missing optimization: Earlier in the function, if `tryGetLength()` returned non-null and we verified the length was correct there, then we don't actually have to try reading this extra byte here...

kentonv:
Didn't @harrishancock already convert this before?

kentonv (approved):

kentonv:
Yeah he did: #1747

kentonv:
Oddly, your PR and Harris's PR overlap but neither is a superset of the other.

PR #1777, opened 2023-08-31
Implement KJ_IF_SOME and kj::none

harrishancock:
The KJ_IF_MAYBE macro has a footgun in that it defines a pointer, rather than a reference, meaning it is easy to accidentally use pointers when you mean to use references. Now that we require C++20, we can use modern syntax (technically, from C++17) to implement a new macro which defines a reference instead of a pointer.

This PR:
- Defines a new KJ_IF_SOME macro to replace KJ_IF_MAYBE.
- Defines a new `kj::none` constant to be used instead of `nullptr` in Maybe comparisons/initializations.
- Deprecates all existing usage of KJ_IF_MAYBE and `nullptr`-meaning-empty-Maybe.
- Adds Bazel build flags to control the deprecation warnings, since otherwise they are very spammy.
- Updates the entire src/kj directory to use the new tools.

For now, I have converted libkj to prove out the approach. Conversion of the rest of the codebase (i.e., src/capnp) will come later.

### DIFF
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -1051,18 +1051,17 @@ inline void dtor(T& location) {
 // forces the caller to handle the null case in order to satisfy the compiler, thus reliably
 // preventing null pointer dereferences at runtime.
 //
-// Maybe<T> can be implicitly constructed from T and from nullptr.
+// Maybe<T> can be implicitly constructed from T and from kj::none.
 // To read the value of a Maybe<T>, do:
 //
-//    KJ_IF_MAYBE(value, someFuncReturningMaybe()) {
-//      doSomething(*value);
+//    KJ_IF_SOME(value, someFuncReturningMaybe()) {
+//      doSomething(value);
 //    } else {
-//      maybeWasNull();
+//      maybeWasNone();
 //    }
 //
-// KJ_IF_MAYBE's first parameter is a variable name which will be defined within the following
-// block.  The variable will behave like a (guaranteed non-null) pointer to the Maybe's value,
-// though it may or may not actually be a pointer.
+// KJ_IF_SOME's first parameter is a variable name which will be defined within the following
+// block.  The variable will be a reference to the Maybe's value.
 //
 // Note that Maybe<T&> actually just wraps a pointer, whereas Maybe<T> wraps a T and a boolean
 // indicating nullness.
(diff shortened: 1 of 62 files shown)


c++/src/kj/encoding-test.c++
```diff
@@ -241,14 +241,26 @@ KJ_TEST("round-trip invalid UTF-16") {
 }
 
 KJ_TEST("EncodingResult as a Maybe") {
-  KJ_IF_MAYBE(result, encodeUtf16("\x80")) {
-    KJ_FAIL_EXPECT("expected failure");
+  {
+    auto result = encodeUtf16("\x80");
+    KJ_EXPECT(result != nullptr);  // It has bytes ...
+    KJ_EXPECT(result == kj::none); // But an error.
```
harrishancock, line 247: I found this phenomenon to be interesting.

c++/src/kj/encoding-test.c++
```diff
@@ -241,14 +241,26 @@ KJ_TEST("round-trip invalid UTF-16") {
...
+    KJ_IF_SOME(unused, result) {
+      (void)unused;
```
kentonv, line 249: Hmm, does KJ_IF_SOME generate unused variable warnings where KJ_IF_MAYBE didn't? Can we inhibit this with KJ_UNUSED? Seems annoying.

c++/src/kj/common.h
```diff
@@ -1292,11 +1291,68 @@ inline T* readMaybe(const Maybe<T&>& maybe) { return maybe.ptr; }
...
+// Originally, we used `nullptr` to mean "an empty Maybe" in comparisons and initializations. For
+// example, `maybe == nullptr` would be true if `maybe` contained no value, and `Maybe<T>(nullptr)`
+// could be used to explicitly initialize an empty Maybe. This isn't the best design, because
+// `nullptr` could itself be a valid constructor parameter for the wrapped type T.
+//
+// Due to this flaw, we've deprecated using `nullptr` to mean "an empty Maybe" in favor of
+// `kj::none`. Please use `kj::none`.
+
+#if __GNUC__
+// Both clang and GCC define __GNUC__ and understand the GCC set of pragma directives.
```
kentonv, line 1328: I believe that clang does NOT define __GNUC__ when operating in MSVC-compatible mode, but I suspect it'll still accept the GCC-style pragmas in that case.

kentonv (approved):

c++/src/kj/encoding-test.c++
harrishancock: We can, but I see it as a benefit, myself. So, I'll merge as-is, but if it gets annoying in practice, I don't mind if we change it.

c++/src/kj/common.h
harrishancock: Thanks, switched to `__GNUC__ || __clang__`: https://github.com/capnproto/capnproto/compare/b8f0d95ca9fffa2dd374e746b7759957cb8ac1bd..2818ef1435057c62401ec6fbb7b50c65afee7957

PR #1778, opened 2023-08-31
coro-friendly kj::exclusiveJoin

mikea:


c++/src/kj/async-inl.h
```diff
@@ -2413,6 +2527,28 @@ class Coroutine<T>::ForkedPromiseAwaiter {
   Awaiter<U> awaiter;
 };
 
+template <typename T>
+template <typename U, size_t Size>
+class Coroutine<T>::ExclusiveJoinAwaiter {
+public:
+  ExclusiveJoinAwaiter(ExclusiveJoin<U, Size>&& promise) 
+      : node(kj::mv(promise.nodes), promise.location), awaiter(OwnPromiseNode(&node)) { }
```
harrishancock, line 2535: I suspect the `OwnPromiseNode(&node)` will cause `node`'s destructor to run twice, or double-freed: once when `awaiter`'s AwaiterBase base class destroys the OwnPromiseNode, and again when ExclusiveJoinAwaiter's destructor destroys `node`.

harrishancock: My first instinct would be to 

I see two options:
1. Make a non-owning version of Awaiter<U>/AwaiterBase.
2. Make two versions of ExclusiveJoinPromiseNode -- the current one would continue to be allocated by `appendPromise()` and would continue to call `freePromise()` in its `destroy()` override. The new one would be allocated on the stack/coroutine frame, and would instead explicitly destroy all of its dependency nodes in its `destroy()` override.

I think option 2 is the correct answer. :( Option 1 would probably work, but it worries me, because we need to uphold the guarantee that promises are destroyed after their `get()` method is called and before the resulting value is returned to user code. Right now, both of those things happen inside AwaiterBase. If we make a non-owning version of Awaiter<U>, then we decouple those two places, making things a bit harder to reason about.

mikea: I think `FreeOnDestroy` template parameter that I have just added should take care of this, no?

kentonv: Huh, it looks like `TransformPromiseNodeBase` actually drops its dependency _twice_ during `get()`: Once in getDepResult() and again in get(). That seems like a mistake, though probably has negligible impact.

For `TransformPromiseNode` it's pretty important that the dependency is dropped _before_ the transformation callback (continuation) is invoked, since the app may be expecting that. Dropping the dependency in `getDepResult()` achieves that.

For `ExclusiveJoinPromiseNode` I suppose it doesn't really matter since we aren't calling back into application code in the meantime, and the immediate caller will be sure to drop the node if needed.

`ForkBranchBase::releaseHub()` seems dubious to me: I suppose only the last branch to release the hub would observe any destruction exceptions.

PR #1780, opened 2023-09-04
Use KJ_IF_SOME and kj::none throughout libcapnp

harrishancock:
I previously converted libkj; this commit completes the job by converting the the remainder of the capnproto repository, namely libcapnp.

### DIFF
--- a/c++/.bazelrc
+++ b/c++/.bazelrc
@@ -36,7 +36,3 @@ build:windows --cxxopt='/Zc:__cplusplus' --host_cxxopt='/Zc:__cplusplus'
 build --//src/kj:openssl=True
 build --//src/kj:zlib=True
 build --//src/kj:brotli=True
-
-# Disable deprecation warnings until we have more of the codebase converted.
-build --//src/kj:deprecate_kj_if_maybe=False
-build --//src/kj:deprecate_empty_maybe_from_nullptr=False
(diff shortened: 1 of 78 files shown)


kentonv (approved):
I didn't actually read it all... let me know if there are any particularly interesting bits I should look at.

harrishancock:
> I didn't actually read it all... let me know if there are any particularly interesting bits I should look at.

Happily, no, it was all just mechanical changes.

I had left this in draft because I want to check to see if Windows or Mac report any warnings in code that I missed. But of course, they don't use Bazel, and Bazel is currently what enables the deprecation warning. So, I pushed a commit which enables the warnings by default in kj/common.h, if it discovers that the configuration macro is undefined.

That caused all of the autotools builds to fail, partly because they compile with `-Werror`, but more interestingly, I guess our Bazel builds don't produce warnings from header files whereas the autotools builds do, so I was missing a whole bunch of places I should have updated.

harrishancock:
I fixed the remaining warnings in one last commit, and CI is looking green, so I'm going to merge.

kentonv:
> I guess our Bazel builds don't produce warnings from header files whereas the autotools builds do

Perhaps we should fix that, presumably by defining `KJ_HEADER_WARNINGS` and `CAPNP_HEADER_WARNINGS` in that build.

PR #1783, opened 2023-09-06
converting http server loop to coroutines

mikea:
there are several more promises in the loop remaining, but the code and its decisions are more readable now imo


mikea:
btw it seems that github diff is confused by indentation change a lot.
Looking at this diff in vscode is much more pleasant.

kentonv:
> It seems that github diff is confused by indentation change a lot.

"Hide whitespace" fixes that, not sure why GitHub always turns it off by default.

Screenshot from 2023-09-06 11-53-31

harrishancock (approved):
Nice!

PR #1784, opened 2023-09-06
http: Call getInner() after every co_await in HttpChunkedEntityWriter

harrishancock:
The reference to the inner output stream provided by HttpChunkedEntityWriter's base class's getInner() can be invalidated before the HttpChunkedEntityWriter. To put it another way, body streams which we provide to user code (the HttpChunkedEntityWriter) can be prolonged by the user beyond the lifetime of the underlying connection. getInner() is written in such a way to catch the issue, by checking for validity before returning a reference, and I think the tryPumpFrom() implementation probably should call it after every resumption.

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -2555,20 +2555,18 @@ public:
 
   Promise<uint64_t> pumpImpl(AsyncInputStream& input, uint64_t length) {
     // Hey, we know exactly how large the input is, so we can write just one chunk.
-    auto& inner = getInner();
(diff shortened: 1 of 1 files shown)


harrishancock:
`getInner()` implementation, for reference: https://github.com/capnproto/capnproto/blob/v2/c%2B%2B/src/kj/compat/http.c%2B%2B#L2365-L2374

mikea (approved):

mikea:
this needs a rebase, sorry @harrishancock

PR #1786, opened 2023-09-06
Fix compilation of async stack limit tests

jclee:
...in configurations where we can't test stack limits with fibers but can test them with pthreads.  Also, exercise the pthreads-based code where possible, to reduce the likelihood of future breakage.

### DIFF
--- a/c++/src/kj/async-test.c++
+++ b/c++/src/kj/async-test.c++
@@ -25,7 +25,7 @@
 #include "mutex.h"
 #include "thread.h"
 
-#if !KJ_USE_FIBERS && !_WIN32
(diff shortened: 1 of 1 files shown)


jclee:
Huh... I guess all the CI passes based on `super-test` are expected to fail in the short-term, due to `KJ_IF_MAYBE` deprecation warnings-turned-errors.

kentonv:
@jclee No, I think the build is broken due to a bad merge. See #1785.

jclee:
Ah, thanks.  Yeah, just noticed was CI passing for that PR.  Will wait and rebase.

PR #1788, opened 2023-09-06
tls.c++ coroutine conversion

mikea:


mikea:
cc @dom96

c++/src/kj/compat/tls.c++
```diff
@@ -312,78 +317,57 @@ private:
...
-        return tryReadInternal(reinterpret_cast<byte*>(buffer) + n,
-                               minBytes - n, maxBytes - n, alreadyDone + n);
-      }
-    });
-  }
-
-  Promise<void> writeInternal(kj::ArrayPtr<const byte> first,
-                              kj::ArrayPtr<const kj::ArrayPtr<const byte>> rest) {
+  Promise<void> writeInternal(kj::ArrayPtr<const kj::ArrayPtr<const byte>> slices) {
     KJ_REQUIRE(shutdownTask == kj::none, "already called shutdownWrite()");
```
jclee, line 321: Looks like the original code was checking for shutdownTask every recursive iteration, but now just checks at the start?

c++/src/kj/compat/tls.c++
```diff
@@ -587,10 +567,8 @@ public:
     //   So, we make some copies here.
     auto& tlsRef = tls;
     auto hostnameCopy = kj::str(hostname);
-    return inner->connect().then(
-        [&tlsRef,hostname=kj::mv(hostnameCopy)](Own<AsyncIoStream>&& stream) {
-      return tlsRef.wrapClient(kj::mv(stream), hostname);
-    });
+    auto stream = co_await inner->connect();
+    co_return co_await tlsRef.wrapClient(kj::mv(stream), hostname);
```
jclee, line 571: Probably should be hostnameCopy instead of hostname here?

c++/src/kj/compat/tls.c++
```diff
@@ -599,10 +577,8 @@ public:
     //   So, we make some copies here.
     auto& tlsRef = tls;
     auto hostnameCopy = kj::str(hostname);
-    return inner->connectAuthenticated().then(
-        [&tlsRef, hostname = kj::mv(hostnameCopy)](kj::AuthenticatedStream stream) {
-      return tlsRef.wrapClient(kj::mv(stream), hostname);
-    });
+    auto stream = co_await inner->connectAuthenticated();
+    co_return co_await tlsRef.wrapClient(kj::mv(stream), hostname);
```
jclee, line 581: Similarly, probably should be hostnameCopy instead of hostname here?

c++/src/kj/compat/tls.c++
```diff
@@ -209,33 +209,28 @@ public:
...
-      auto result = SSL_get_verify_result(ssl);
-      if (result != X509_V_OK) {
-        const char* reason = X509_verify_cert_error_string(result);
-        KJ_FAIL_REQUIRE("TLS peer's certificate is not trusted", reason) { break; }
-      }
-    });
+    auto result = SSL_get_verify_result(ssl);
+    if (result != X509_V_OK) {
+      const char* reason = X509_verify_cert_error_string(result);
+      KJ_FAIL_REQUIRE("TLS peer's certificate is not trusted", reason) { break; }
```
dom96, line 221: I know this was here before, but what is that `{ break; }` doing here?

dom96: Actually, first time I'm seeing this kind of construct. The `co_return` above as well, how does that interact with `KJ_REQUIRE`?

c++/src/kj/compat/tls.c++
```diff
@@ -248,16 +243,26 @@ public:
...
+    size_t alreadyDone = 0;
+    while (true) {
+      auto n = co_await sslCall([this,buffer,maxBytes]() { return SSL_read(ssl, buffer, maxBytes); });
+      if (n >= minBytes || n == 0) {
+        co_return alreadyDone + n;
+      } else {
+        buffer = reinterpret_cast<byte*>(buffer) + n;
+        minBytes -= n;
+        maxBytes -= n;
+        alreadyDone += n;
```
dom96, line 255: Hope we have good test coverage for this

dom96: It may be safer to avoid optimising in the same PR and instead do what the previous code did (recursion).

c++/src/kj/compat/tls.c++
dom96: Yeah, I think it's safer to avoid optimising while converting to coroutines in the same PR. The optimisation should come in a separate PR to avoid us missing things like this.

c++/src/kj/compat/tls.c++
```diff
@@ -312,78 +317,57 @@ private:
...
       int error = SSL_get_error(ssl, result);
       switch (error) {
         case SSL_ERROR_ZERO_RETURN:
-          return constPromise<size_t, 0>();
+          co_return 0;
         case SSL_ERROR_WANT_READ:
-          return readBuffer.whenReady().then(
-              [this,func=kj::mv(func)]() mutable { return sslCall(kj::fwd<Func>(func)); });
+          co_await readBuffer.whenReady();
+          break;
```
dom96, line 353: It would be much clearer to call `sslCall` explicitly here, so much so that I think whatever slowdown this causes is worth it.

c++/src/kj/compat/tls.c++
mikea: > It may be safer to avoid optimising in the same PR and instead do what the previous code did (recursion).

we can't keep recursion with coroutines

c++/src/kj/compat/tls.c++
mikea: but this is a recursive call, and we can't have recursive calls with coro.

mikea:
@dom96 here's a link regarding the recursion: https://github.com/capnproto/capnproto/blob/v2/kjdoc/tour.md#coroutines

c++/src/kj/compat/tls.c++
mikea: > Looks like the original code was checking for shutdownTask every recursive iteration, but now just checks at the start?

My reasoning: since this code is single-threaded, shutdownWrite() can't happen while we write data out. So it should be enough calling it once?

I'm happy to move it inside the loop to if you think this is not correct.

c++/src/kj/compat/tls.c++
mikea: I don't think we need hostnameCopy anymore at all actually.

c++/src/kj/compat/tls.c++
mikea: not needed either

c++/src/kj/compat/tls.c++
mikea: fwiw I tried to keep as close to the original as possible eliminating recursion.

c++/src/kj/compat/tls.c++
```diff
@@ -312,78 +317,55 @@ private:
...
-
-    // SSL_write() with a zero-sized input returns 0, but a 0 return is documented as indicating
-    // an error. So, we need to avoid zero-sized writes entirely.
-    while (first.size() == 0) {
-      if (rest.size() == 0) {
-        return kj::READY_NOW;
+  Promise<void> writeInternal(kj::ArrayPtr<const kj::ArrayPtr<const byte>> slices) {
+    auto cork = writeBuffer.cork();
+    for (auto slice: slices) {
+      KJ_REQUIRE(shutdownTask == kj::none, "already called shutdownWrite()");
```
jclee, line 323: If I'm understanding correctly, this would still differ in behavior from the original code, since it just checks for the presence of shutdownTask at the beginning of each slice, while the original code would effectively check before every call to `SSL_write()`, potentially several times per slice.

If we wanted to approximate the original behavior, I think we need to do the check at least at the beginning of the inner `while` loop.  If we want to exactly replicate the original behavior, we might also need to call it elsewhere with more complex logic, since it looks like it will also perform the check in situations where the list is empty or contains empty slices.

c++/src/kj/compat/tls.c++
```diff
@@ -183,22 +183,22 @@ public:
...
     if (isIpAddress(expectedServerHostname)) {
       if (X509_VERIFY_PARAM_set1_ip_asc(verify, expectedServerHostname.cStr()) <= 0) {
-        return getOpensslError();
+        throw getOpensslError();
       }
     } else {
       if (X509_VERIFY_PARAM_set1_host(
           verify, expectedServerHostname.cStr(), expectedServerHostname.size()) <= 0) {
-        return getOpensslError();
+        throw getOpensslError();
```
jclee, line 201: I wonder if maybe these bare `throw`s should be `kj::throwFatalException(getOpensslError())` instead?  Or `throwOpensslError()`, which seems to be the same thing.

c++/src/kj/compat/tls.c++
```diff
@@ -248,16 +243,26 @@ public:
...
+    }
   }
 
   Promise<void> write(const void* buffer, size_t size) override {
-    return writeInternal(kj::arrayPtr(reinterpret_cast<const byte*>(buffer), size), nullptr);
+    return writeInternal({kj::arrayPtr(reinterpret_cast<const byte*>(buffer), size)});
   }
 
   Promise<void> write(ArrayPtr<const ArrayPtr<const byte>> pieces) override {
-    auto cork = writeBuffer.cork();
```
jclee, line 259: Not sure if important, but FWIW it looks like the original code would only call `writeBuffer.cork()` for the multi-piece `write()` overload, but the new code would also call it for the single-buffer `write()` overload.

c++/src/kj/compat/tls.c++
kentonv: Yes, we should not have bare `throw`s ever.

c++/src/kj/compat/tls.c++
```diff
@@ -248,16 +243,26 @@ public:
...
+        minBytes -= n;
+        maxBytes -= n;
+        alreadyDone += n;
+      }
+    }
   }
 
   Promise<void> write(const void* buffer, size_t size) override {
-    return writeInternal(kj::arrayPtr(reinterpret_cast<const byte*>(buffer), size), nullptr);
+    return writeInternal({kj::arrayPtr(reinterpret_cast<const byte*>(buffer), size)});
```
kentonv, line 261: This isn't safe -- you're passing a temporary array (the outer array) to a coroutine which expects the array to remain valid across co_awaits. This is why `writeInternal()` was structured the way it was before, with separate "first" and "rest" arrays. Can you keep the old approach?

mikea: good point. Wouldn't `co_return co_await` solve this problem rather than have first/rest arguments?

mikea: (I'm also wondering if we can setup asan build for capnproto?)

kentonv: `co_return co_await` would solve it at the expense of allocating an extra coroutine frame.

PR #1794, opened 2023-09-08
Revert "Merge pull request #1762 from capnproto/harris/kj-stack-array…

harrishancock:
…-no-longer-supports-vlas"

This reverts commit 15bcc0e55aec5b034e3736b098c5f2326e0a15a9, reversing changes made to 33991de6792fb9a96e46f632571106b2bbd23f8e.

### DIFF
--- a/c++/src/capnp/compat/json.c++
+++ b/c++/src/capnp/compat/json.c++
@@ -302,7 +302,7 @@ void JsonCodec::encode(DynamicValue::Reader input, Type type, JsonValue::Builder
       auto structValue = input.as<capnp::DynamicStruct>();
       auto nonUnionFields = structValue.getSchema().getNonUnionFields();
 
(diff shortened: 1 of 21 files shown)


harrishancock:
PR #1762 appears to have caused a regression downstream in the Workers runtime, so I'm reverting it.

Note that this also reverts the Array destructor noexcept change, which we should probably reapply later.

harrishancock:
I added a revert of https://github.com/capnproto/capnproto/pull/1774 due to downstream test failures it seems to have caused.

harrishancock:
Also added a revert of https://github.com/capnproto/capnproto/pull/1786 due to an arm64-only downstream test failure it seems to have caused.

ohodson (approved):

ohodson:
Approved assuming these are pure reverts.

PR #1796, opened 2023-09-11
Fix compilation of async stack limit tests

jclee:
...in configurations where we can't test stack limits with fibers but can test them with pthreads.  Also, exercise the pthread-based code where possible, to reduce the likelihood of future breakage.

Reapplying fcad1296 (PR #1786, reverted in 99115bdd) with changes to make pthread-based testing more tolerant of system-dependent minimum stack sizes; expected to fix tests on ARM, where the minimum stack size is apparently larger than 16k.

### DIFF
--- a/c++/src/kj/async-test.c++
+++ b/c++/src/kj/async-test.c++
@@ -25,7 +25,9 @@
 #include "mutex.h"
 #include "thread.h"
 
-#if !KJ_USE_FIBERS && !_WIN32
+#if !_WIN32
+#include <errno.h>
+#include <limits.h>
 #include <pthread.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


jclee:
Also needed to update the stack size value to be non-constexpr, since it is apparently not constexpr in some header configurations, which then required going back to passing around a lambda.  But presumably it's being passed and invoked safely now.

kentonv (approved):

c++/src/kj/async-test.c++
```diff
@@ -884,40 +886,68 @@ TEST(Async, TaskSet) {
 // These tests require either fibers or pthreads in order to limit the stack size. Currently we
 // don't have a version that works on Windows without fibers, so skip the tests there.
 
-template <typename Func>
+inline size_t getSmallStackSize() {
+#if !_WIN32
+  // pthread_attr_setstacksize() requires a stack size of at least PTHREAD_STACK_MIN, which can
+  // vary by platform.  We'll clamp that to a reasonable range for stack overflow tests, and skip
+  // the pthread-based tests if we can't get it.
+  return kj::max(16 * 1024, kj::min(256 * 1024, PTHREAD_STACK_MIN));
```
hebasto, line 894: Using `PTHREAD_STACK_MIN` results in a compiler failure on NetBSD.

Also see:
- https://gnats.netbsd.org/37169
- https://mail-index.netbsd.org/tech-userlevel/2014/08/14/msg008726.html

PR #1799, opened 2023-09-11
kj::Rc<T> smart reference

mikea:


c++/src/kj/refcount.h
```diff
@@ -115,6 +132,93 @@ Own<T> Refcounted::addRefInternal(T* object) {
   return Own<T>(object, *refcounted);
 }
 
+template <typename T>
+Rc<T> Refcounted::addRcRefInternal(T* object) {
+  Refcounted* refcounted = object;
+  ++refcounted->refcount;
+  return Rc<T>(object);
```
jasnell, line 139: Are there any challenges here if `this` wasn't actually created by Rc<T>::create? For instance, if T is allocated on the stack? Might be a good idea to add a check in here that the thing is actually refcounted?

jasnell:
Overall looking really good so far

c++/src/kj/refcount-test.c++
```diff
@@ -57,6 +57,45 @@ TEST(Refcount, Basic) {
...
+  EXPECT_FALSE(b);
+
+  ref3 = nullptr;
+  EXPECT_FALSE(ref2->isShared());
+  EXPECT_FALSE(b);
+
+  ref2 = nullptr;
+
+  // last reference dropped, SetTrueInDestructor destructor should execute
+  EXPECT_TRUE(b);
```
jasnell, line 96: TODO: Add a test for `EnabledAddRefToThis`?

c++/src/kj/refcount.h
```diff
@@ -115,6 +132,93 @@ Own<T> Refcounted::addRefInternal(T* object) {
...
+  static Rc<T> create(Params&&...params) {
+    // Main factory function for Rc<T>. 
+    // NOTE: This will become global refcounted() function once migration is completed.
+    return Refcounted::addRcRefInternal(new T(kj::fwd<Params>(params)...));
+  }
+
+  inline Rc(Rc&& other) noexcept: Own<T>(kj::mv(other)) { }
+
+  template <typename U, typename = EnableIf<canConvert<U*, T*>()>>
+  inline Rc(Rc<U>&& other) noexcept : Own<T>(kj::mv(other)) { }
```
jasnell, line 162: For later, I wonder if we could eliminate (or at least obsolete) `kj::Own<kj::RefcountedWrapper<T>>` with a constructor here that accepts `kj::Own<T>&&` and wraps it internally as a `Refcounted`?

```cpp
kj::Own<Foo> foo;
kj::Rc<Foo> rcFoo = kj::mv(foo);
```

Since this would be eliminating `kj::addRef(*foo)` and replacing with the `addRef` on `kj::Rc`, the fact that `Foo` does not extend from `Refcounted` could be ok?

PR #1809, opened 2023-09-18
Make Promises [[nodiscard]]

smerritt:
Bonus (malus?) update to zlib 1.3 to get the project to build again.

This emits a warning, not an error, when dropping a Promise on the floor. Since it's only a warning, I believe it's safe to apply to the 1.0 branch. If you'd rather have it on `v2` only, that's fine too.

### DIFF
--- a/c++/WORKSPACE
+++ b/c++/WORKSPACE
@@ -31,6 +31,7 @@ cc_library(
     name = "zlib",
     srcs = glob(["*.c"]),
     hdrs = glob(["*.h"]),
+    includes = ["."],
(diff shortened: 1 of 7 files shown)


smerritt:
The zlib WORKSPACE change is a little odd-looking. What I found was that Bazel was building kj with `-iquote external/zlib`, which works for `#include "zlib.h"` but not for `#include <zlib.h>`. However, where KJ included `zlib.h`, it was done with angle brackets, not quotes. Adding `includes = ["."]` makes Bazel invoke clang with `-Iexternal/zlib` so the includes work.

I have no idea why this used to work.

smerritt:
I've changed my mind; for anyone compiling their dependent project with `-Werror`, this will cause build failures. It belongs on the v2 branch.

PR #1811, opened 2023-09-19
Allow TLS users to configure preferred curves

Warfields:
Note, in TLS 1.3 curves are called groups. SSL libs have decided to use the legacy api that mentions curves.

### DIFF
--- a/c++/src/kj/compat/tls.c++
+++ b/c++/src/kj/compat/tls.c++
@@ -791,6 +791,12 @@ TlsContext::TlsContext(Options options) {
     throwOpensslError();
   }
 
+  KJ_IF_SOME(curves, options.curveList) {
+    if (!SSL_CTX_set1_curves_list(ctx, curves.cStr())){
+      throwOpensslError();
+    }
+  }
+
   // honor options.defaultKeypair
   KJ_IF_SOME(kp, options.defaultKeypair) {
     if (!SSL_CTX_use_PrivateKey(ctx, reinterpret_cast<EVP_PKEY*>(kp.privateKey.pkey))) {
--- a/c++/src/kj/compat/tls.h
+++ b/c++/src/kj/compat/tls.h
@@ -97,6 +97,11 @@ public:
     //   algorithms.
     // - You need quickly to disable an algorithm recently discovered to be broken.
 
+    kj::Maybe<kj::StringPtr> curveList;
+    // Sets the preferred curves (Groups in TLS 1.3), by default this is not set. Similar to the
+    // cipher list, this is a colon separated list of human readable names or NIDs.
+    // https://boringssl.googlesource.com/boringssl/+/refs/heads/master/include/openssl/nid.h
+
     kj::Maybe<const TlsKeypair&> defaultKeypair;
     // Default keypair to use for all connections. Required for servers; optional for clients.
 


c++/src/kj/compat/tls.c++
```diff
@@ -791,6 +791,13 @@ TlsContext::TlsContext(Options options) {
     throwOpensslError();
   }
 
+  // honor options.curveList
```
jp4a50, line 794: [nit] Comment doesn't add any value.

c++/src/kj/compat/tls.h
```diff
@@ -97,6 +97,9 @@ class TlsContext: public kj::SecureNetworkWrapper {
     //   algorithms.
     // - You need quickly to disable an algorithm recently discovered to be broken.
 
+    kj::Maybe<kj::StringPtr> curveList;
+    // Sets the preferred curves (Groups in TLS 1.3), by default this is not set.
```
jp4a50, line 101: I've not got the domain knowledge to know whether this is self-explanatory, but is it clear what format each "curve" should be in when specified as a string? The [docs](https://www.openssl.org/docs/man1.0.2/man3/SSL_CTX_set1_curves_list.html) talk about "curve NIDs".

jp4a50 (approved):

c++/src/kj/compat/tls.c++
Warfields: Keeping with the style around it.

c++/src/kj/compat/tls.h
Warfields: Yes, colon separated just like the cipher list.

> [Each EC_builtin_curve item has a unique integer id (nid), and a human readable comment string describing the curve.](https://www.openssl.org/docs/manmaster/man3/EC_GROUP_new.html)

jp4a50: What I'm saying is that this should be documented :P

c++/src/kj/compat/tls.c++
jp4a50: It's just a nit so it's not blocking but this is **very** widely considered to be bad style :)

Warfields: I agree that it's bad form, just would be a little weird in the context of the rest of the file.
https://github.com/capnproto/capnproto/pull/1811/files/85d4074ac5b5ec59527020b1c5b0bab1a6d7c6f0#diff-dac192e78e5c2b9147c63353b2f6f96e41d7101c9dbbe89698da0bf06fa49b47R789
https://github.com/capnproto/capnproto/pull/1811/files/85d4074ac5b5ec59527020b1c5b0bab1a6d7c6f0#diff-dac192e78e5c2b9147c63353b2f6f96e41d7101c9dbbe89698da0bf06fa49b47L794
https://github.com/capnproto/capnproto/pull/1811/files/85d4074ac5b5ec59527020b1c5b0bab1a6d7c6f0#diff-dac192e78e5c2b9147c63353b2f6f96e41d7101c9dbbe89698da0bf06fa49b47L766

c++/src/kj/compat/tls.h
Warfields: I linked to the source of the NID's and names in boring ssl.

PR #1816, opened 2023-09-22
Minor cleanup in rpc-twoparty.c++.

kentonv:
The code immediately before this alerady places a strong ref to the message into `queuedMessages`. There's no need to attach a second strong reference to the pump promise.

This change has no real impact, just makes the code clearer.

### DIFF
--- a/c++/src/capnp/rpc-twoparty.c++
+++ b/c++/src/capnp/rpc-twoparty.c++
@@ -179,8 +179,8 @@ public:
     }
 
     // On the other hand, if pendingMessages was empty, then we should set up the delayed write.
-    network.previousWrite = previousWrite.then([this, sendTime]() {
-      return kj::evalLast([this, sendTime]() -> kj::Promise<void> {
+    network.previousWrite = previousWrite.then([&network = network, sendTime]() {
+      return kj::evalLast([&network, sendTime]() -> kj::Promise<void> {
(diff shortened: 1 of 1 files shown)


c++/src/capnp/rpc-twoparty.c++
```diff
@@ -192,7 +192,7 @@ public:
           messages[i].fds = ownMessages[i]->fds;
         }
         return network.getStream().writeMessages(messages).attach(kj::mv(ownMessages), kj::mv(messages));
-      }).catch_([this](kj::Exception&& e) {
+      }).catch_([&network](kj::Exception&& e) {
```
jasnell, line 195: I see the `networker.queuedMessages.add(kj::addRef(*this))` on like 173 above... what I'm wondering is whether the `kj::mv(network.queuedMessages)` on line 186 could be problematic for this catch? I don't suspect so but is there any chance the captured `&network` on this catch handler could end up invalid after that move on 186? Just making sure I have the mental model correct.

kentonv: If the `network` had been destroyed, then its field `previousWrite` would have been destroyed, thereby cancelling this promise. Thus, we can conclude `network` is definitely still valid in this lambda.

jasnell: Got it!

jasnell (approved):

PR #1818, opened 2023-09-26
Add single param slice to arrays

Warfields:
Now array.slice(1234) will return a slice from 1234 to the end of the array, this is the same behaivor as kj::StringPtr's slicing.


kentonv:
Yeah I guess I should have done this earlier, having to do `x.slice(blah, x.size())` gets pretty tiring.

PR #1814, opened 2023-09-20
Fix C++20 compiler errors - MSVC (v193).

risa2000:
Fixes compiler errors when compiled with forced C++20 (option /std:c++20 in MSVC) in MSVC toolchain v193 aka Visual Studio 2022.

`u8` in C+20 creates `char8_t *` literals which are basically useless. Luckily for the tests they work with regular `char *`.
The only test which fails from `kj-heavy-tests.exe` is the stack trace parsing which I did not dare to look into.

PR #1819, opened 2023-10-01
Fix MSVC with C++20 on 1.x branch.

risa2000:
@kentonv 
This is continuation of the effort from #1814. Following your recommendations, I cherry-picked the few commits from `v2` branch which were addressing the issues I saw when building `release-1.0.1` with MSVC with C++20 standard.
I had to put one fix from @harrishancock (aa4d3699fc7c7a1524d377f55494fb2b7134592f) into a conditional compiled block as the fix only worked with higher C++ standard than C++14 which `v1` targets.

I tested this on MSVC v193 (VS2022) compiling with C++14, C++17, C++20 and it works - sort of.

The problem I found is that the `Release` build crashes in `kj-heavy-tests.exe` in particular in `async-test.c++` and `fiber pool limit`. What makes it a bit hard to debug is the fact that it only crashes in the release build (the `Debug` runs fine and even `RelWithDebInfo` runs fine). The trigger seems to be `/Ob2` compiler option which is only set in `Release` (as defined by `CMake`), while `RelWithDebInfo` is compiled for some reason with stepped down `/Ob1`.

This option controls inline expansion (https://learn.microsoft.com/en-us/cpp/build/reference/ob-inline-function-expansion?view=msvc-170).

It is possible to build the code with debug info even with `/O2` and `/Ob2` (which is the default for `Release`), but debugging is a bit limited and I, not being familiar with both the fibers capnproto implementation and the test, was not able to determine the reason. The behavior is same with `clang-cl`, so it kind of rules out that the culprit is the compiler alone. I suspect that it might be MSVC STL implementation (which `clang-cl` uses as well), or even a bug in the test.

This behavior is reproducible on any of `v1` commits, the one in this PR, the original `release-1.0.1` and even on `release-0.10.4`, regardless which C++ standard is targeted. So I am not sure what to do about it.

There is another thing I noticed during the tests. I use `clang-cl` as a tool to "cross check" the validity of the code and noticed a regular pattern in `capnproto` code which goes like this:
```c++
#if !_MSC_VER || defined(__clang__)
// do something here
#endif
```
which seems to be used to exclude a code for MSVC toolchain.
With `clang-cl` however this condition is met as the compiler defines both defines: `_MSC_VER` *and* `__clang__`. For example the version of `clang-cl` I am using defines it this way:
```
_MSC_VER=1937
__clang__=1
```
Now, depending on whether you are trying to target the particular compiler, or rather the STL behavior it may or may not work. For example this code https://github.com/capnproto/capnproto/blob/e7a7f2d6f9c4efc7d81cb2aac4fd13c561c47b63/c%2B%2B/src/kj/async-test.c%2B%2B#L40-L45 does not work on `clang-cl` either (and is thus correctly commented out) so my assumption would be this is rather the STL implementation than the toolchain problem.
On the other hand, this test gets compiled in (with `clang-cl`) and fails: https://github.com/capnproto/capnproto/blob/e7a7f2d6f9c4efc7d81cb2aac4fd13c561c47b63/c%2B%2B/src/kj/async-coroutine-test.c%2B%2B#L381-L437

### DIFF
--- a/c++/src/capnp/generated-header-support.h
+++ b/c++/src/capnp/generated-header-support.h
@@ -335,7 +335,7 @@ inline constexpr uint sizeInWords() {
 #endif
 
 // TODO(msvc): MSVC does not even expect constexprs to have definitions below C++17.
-#if (__cplusplus < 201703L) && !(defined(_MSC_VER) && !defined(__clang__))
+#if (KJ_CPP_STD < 201703L) && !(defined(_MSC_VER) && !defined(__clang__))
 #define CAPNP_NEED_REDUNDANT_CONSTEXPR_DECL 1
 #else
 #define CAPNP_NEED_REDUNDANT_CONSTEXPR_DECL 0
--- a/c++/src/kj/array-test.c++
+++ b/c++/src/kj/array-test.c++
@@ -378,7 +378,7 @@ TEST(Array, ReleaseAsBytesOrChars) {
   }
 }
 
-#if __cplusplus > 201402L
+#if __cplusplus > 201402L || _MSVC_LANG > 201402L
 KJ_TEST("kj::arr()") {
   kj::Array<kj::String> array = kj::arr(kj::str("foo"), kj::str(123));
   KJ_EXPECT(array == kj::ArrayPtr<const kj::StringPtr>({"foo", "123"}));
--- a/c++/src/kj/array-test.c++
+++ b/c++/src/kj/array-test.c++
@@ -378,7 +378,7 @@ TEST(Array, ReleaseAsBytesOrChars) {
   }
 }
 
-#if __cplusplus > 201402L || _MSVC_LANG > 201402L
+#if KJ_CPP_STD > 201402L
 KJ_TEST("kj::arr()") {
   kj::Array<kj::String> array = kj::arr(kj::str("foo"), kj::str(123));
   KJ_EXPECT(array == kj::ArrayPtr<const kj::StringPtr>({"foo", "123"}));
--- a/c++/src/kj/array.h
+++ b/c++/src/kj/array.h
@@ -848,7 +848,7 @@ inline Array<T> heapArray(std::initializer_list<T> init) {
   return heapArray<T>(init.begin(), init.end());
 }
 
-#if __cplusplus > 201402L || _MSVC_LANG > 201402L
+#if KJ_CPP_STD > 201402L
 template <typename T, typename... Params>
 inline Array<Decay<T>> arr(T&& param1, Params&&... params) {
   ArrayBuilder<Decay<T>> builder = heapArrayBuilder<Decay<T>>(sizeof...(params) + 1);
--- a/c++/src/kj/array.h
+++ b/c++/src/kj/array.h
@@ -848,7 +848,7 @@ inline Array<T> heapArray(std::initializer_list<T> init) {
   return heapArray<T>(init.begin(), init.end());
 }
 
-#if __cplusplus > 201402L
+#if __cplusplus > 201402L || _MSVC_LANG > 201402L
 template <typename T, typename... Params>
 inline Array<Decay<T>> arr(T&& param1, Params&&... params) {
   ArrayBuilder<Decay<T>> builder = heapArrayBuilder<Decay<T>>(sizeof...(params) + 1);
--- a/c++/src/kj/async-coroutine-test.c++
+++ b/c++/src/kj/async-coroutine-test.c++
@@ -378,7 +378,7 @@ KJ_TEST("co_await only sees coroutine destruction exceptions if promise was not 
       awaitPromise(kj::mv(rejectedThrowyDtorPromise)).wait(waitScope));
 }
 
-#if (!_MSC_VER || defined(__clang__)) && !__aarch64__
+#if !_MSC_VER  && !__aarch64__
 uint countLines(StringPtr s) {
   uint lines = 0;
   for (char c: s) {
--- a/c++/src/kj/async-inl.h
+++ b/c++/src/kj/async-inl.h
@@ -2094,7 +2094,13 @@ public:
   void destroy() override;
 
   auto initial_suspend() { return stdcoro::suspend_never(); }
-  auto final_suspend() noexcept { return stdcoro::suspend_always(); }
+  auto final_suspend() noexcept {
+#if _MSC_VER && !defined(__clang__)
+    // See comment at `finalSuspendCalled`'s definition.
+    finalSuspendCalled = true;
+#endif
+    return stdcoro::suspend_always();
+  }
   // These adjust the suspension behavior of coroutines immediately upon initiation, and immediately
   // after completion.
   //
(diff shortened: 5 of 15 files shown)


risa2000:
On the side note (which we should spin off into another topic if needed): `v2` does not compile in `Release` mode with MSVC (`Debug` compiles fine). Some macro expansion issue in logging macros. `clang-cl` compiles it fine, but the fact that `clang-cl` is introduced in some places as an exclusion from MSVC and elsewhere it passes as MSVC makes it harder to figure out what exactly were the authors intentions.

This is unrelated but shows the potential problem:
https://github.com/capnproto/capnproto/blob/e7a7f2d6f9c4efc7d81cb2aac4fd13c561c47b63/c%2B%2B/src/kj/exception.c%2B%2B#L257-L263

risa2000:
I was able to narrow down the culprit in the `Release` build crash in `fiber pool limit` test to `async.c++` file.
When this file is compiled with `/Ob2` the test crashes, when it is compiled with `/Ob1` the test passes. The rest (including the test source) can by compiled with `/Ob2` and does not have an impact.

kentonv:
The `#if !_MSC_VER || defined(__clang__)` checks are intentionally enabling clang-cl but disabling regular MSVC. We (Cloudflare  Workers team) actually use clang-cl to compile `workerd` on Windows, so this is a well-exercised target for us. In fact, clang-cl is probably better-tested than MSVC itself.

With that said, it does indeed appear that we don't have any CI set up to run these tests on Windows in release mode... only debug mode.

c++/CMakeLists.txt
```diff
@@ -134,13 +134,8 @@ if(MSVC)
...
   #   build with that warning level and clean out false positives.
 
-  add_compile_options(/wo4503)
-  # Only warn once on truncated decorated names. The maximum symbol length MSVC
-  # supports is 4k characters, which the parser framework regularly blows. The
-  # compiler likes to print out the entire type that went over the limit along
-  # with this warning, which gets unbearably spammy. That said, we don't want to
-  # just ignore it, so I'm letting it trigger once until we find some places to
-  # inject ParserRefs.
+  add_compile_options(/Zc:__cplusplus)
```
kentonv, line 137: Without this flag, does everything that worked before still work? I don't want to introduce a new requirement that downstream users use this flag, but I understand that we need the flag in order to enable some features that were disabled before.

risa2000: Not sure what you are asking about. The change (commit) comes from `v2`.

If you are asking about removing `/wo4503`, I thought it might have been an artefact from the past. MSVC generates quite a few different warnings and I did not see why this one should be treated differently. Leaving it there should be fine too, I just did not see a reason.

For adding `/Zc:__cplusplus` I just went with the `v2` change as well, just tweaked it that it works for C++14 `v1` target. If the code uses the `__cplusplus` define then this option should have been set right from the start so it is consistent with the standard.

risa2000:
> The `#if !_MSC_VER || defined(__clang__)` checks are intentionally enabling clang-cl but disabling regular MSVC. We (Cloudflare Workers team) actually use clang-cl to compile `workerd` on Windows, so this is a well-exercised target for us. In fact, clang-cl is probably better-tested than MSVC itself.

You are right. Once I thought it through I realized that the only case this condition actually makes difference is with `clang-cl`. I guess then that differentiating native `clang` and `clang-cl` (with Microsoft STL implementation) might need a different mechanism.

c++/CMakeLists.txt
kentonv: > Not sure what you are asking about. The change (commit) comes from v2.

Right, but v2 is intentionally breaking compat. The 1.x branch, OTOH, should stay stable enough that people can upgrade patch releases without any fear of breakage. I do understand that we probably _should_ have been setting this option all along, but at this point the priority is not to disrupt anyone, and it's not clear to me if this might do that.

I'd be more comfortable if we can leave these flags alone, or maybe if we could set the flag on one of the two MSVC CI targets but leave it off the other one so that we cover both configurations.

risa2000: Ok, I see what you mean. There are two ways (in MSVC) to detect the C++ standard compiler uses at the compilation:
 * `_MSVC_LANG` (https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170) supported from Visual Studio 2015 Update 3
 * `__cplusplus` macro (https://learn.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-170) which would give meaningful values from Visual Studio 2017 version 15.7, because for this macro to work the compiler option `/Zc:__cplusplus` must be set as well, Otherwise the macro expands to `199711L` which does not match any C++ standard.

It seems that `capnproto` does not use the former at all and instead uses some combination of detecting MSVC alone and using some guessing.

In few places `capnproto` uses `__cplusplus`, but since it does not define `/Zc:__cplusplus` the outcome on MSVC is probably not what was intended.

I have pulled the patch from `v2` and then conveniently used it for differentiating C++20 mode. Now you are right that if someone does not set this compiler option the code won't compile. The right way would be to review all `__cplusplus` uses and make it robust to MSVC by adding the test for `_MSVC_LANG` as well. This way the code should behave consistently with or without `/Zc: __cplusplus` being set.

After that it should not matter in `capnproto` too and the option could be omitted as well.

Otherwise we may face an opposite dilemma, when someone sets this option in his project, but `capnproto` was built without it so the corresponding tests for `__cplusplus` will resolve to different outcomes between the project and the library.

risa2000: I force pushed the corrected patch, which:
* removed `/Zc:__cplusplus` from compile options and instead mimicked it with `_MSVC_LANG` macro, so regardless whether the user sets the option or not, the detection will be governed by `_MSVC_LANG` (for MSVC). `clang-cl` seems to be setting both defines correctly so it should be fine too and any other toolchain will be treated as before. (This fix should not be needed in `v2` as the proper logic will be ensured by `/Zc:__cplusplus` option.)
* I trimmed `"Can trace through coroutines"` from the `async-coroutine test.c++` for `clang-cl` as well as this does not work and the comment at the test claims that the test depends on `GetFunctorStartAddress` function, which test is also skipped on `clang-cl` (and would not work either). So this is more to keep it on par with MSVC were both tests are skipped too.

There is still the crash in `fiber pool limit` test in `Release` build (both in `MSVC` and `clang-cl`), but I guess this is for another patch.

risa2000: Now, on the second thought, the fix for `/Zc:__cplusplus` might be a safe approach even on `v2` against cases where users would not set this option in their projects. So the header files which use `__cplusplus` logic will not get treated differently in the library build and the users' projects builds.

risa2000: After another thought I think it would make sense to define internal define to cover both `__cplusplus` and `_MSVC_LANG` for the purpose of `kj` lib. Something like:
```c++
#if defined(_MSVC_LANG)
#define KJ_CPP_STD _MSVC_LANG
#else
#define KJ_CPP_STD __cplusplus
#endif
```
possibly in `kj/common.h`?

I pushed a commit which applies the changes described above.

kentonv (approved):
Sorry for the long delay in reviewing this -- I recently moved and was out of the office for a while.

kentonv:
Weird, github did a squash-and-merge even though I explicitly chose merge commit... oh well.

kentonv:
Anyway, thanks for fixing this! I'll aim to do either a 1.0.2 or a 1.1 release in the next few weeks...

PR #1820, opened 2023-10-06
Add the expected and actual content length in http.c++ exception

Frederik-Baetens:
This will hopefully make debugging somewhat easier.

If the changes to the conditions on length are undesired, I can instead add an originalLength field.

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -1956,8 +1956,13 @@ public:
       if (length > 0) {
         // We haven't reached the end of the entity body yet.
         if (amount == 0) {
-          kj::throwRecoverableException(KJ_EXCEPTION(DISCONNECTED,
-              "premature EOF in HTTP entity body; did not reach Content-Length"));
+          size_t expectedLength = length + alreadyRead;
+          kj::throwRecoverableException(KJ_EXCEPTION(
+            DISCONNECTED,
+            "premature EOF in HTTP entity body; did not reach Content-Length",
+            expectedLength,
+            alreadyRead
+          ));
         } else if (amount < minBytes) {
           // We requested a minimum 1 byte above, but our own caller actually set a larger minimum
           // which has not yet been reached. Keep trying until we reach it.


c++/src/kj/compat/http.c++
```diff
@@ -1943,7 +1943,7 @@ public:
     size_t alreadyRead = 0;
 
     for (;;) {
-      if (length == 0) {
+      if (length == alreadyRead) {
```
jasnell, line 1946: The delta here could likely be minimized by just adding an additional size_t field for the full length as opposed to changing the way `length` is calculated/managed now. But consider that optional. LGTM so long as ci looks good

jasnell (approved):

c++/src/kj/compat/http.c++
```diff
@@ -1952,12 +1952,16 @@ public:
...
-      length -= amount;
-      if (length > 0) {
+      if (length > alreadyRead) {
         // We haven't reached the end of the entity body yet.
         if (amount == 0) {
-          kj::throwRecoverableException(KJ_EXCEPTION(DISCONNECTED,
-              "premature EOF in HTTP entity body; did not reach Content-Length"));
+          kj::throwRecoverableException(KJ_EXCEPTION(
+            DISCONNECTED,
+            "premature EOF in HTTP entity body; did not reach Content-Length. Expected length: ",
```
kentonv, line 1960: KJ exceptions do not simply concatenate the parameters, they produce a formatted error string that includes both the variable names and values. Your exception will end up looking like this:

`premature EOF in HTTP entity body; did not reach Content-Length. Expected length: ; length = 123; Received length: ; alreadyRead = 12`

I would suggest just doing: `KJ_EXCEPTION("premature EOF in HTTP entity body; did not reach Content-Length", length, alreadyRead)`

c++/src/kj/compat/http.c++
kentonv: No, please do not change the meaning of `length` like this. This logic is delicate and it's just too easy to introduce bugs. For instance, as of right now this PR introduces a severe HTTP desync bug because line 1953 has not been updated for the new meaning of `lengt`h.

kentonv (changes requested):
Please do not change the meaning of `length` -- this introduces bugs. Specifically, line 1953 is now buggy (allowing reading past the end of the body), and maybe other places as well. Let's leave the logic as it was.

c++/src/kj/compat/http.c++
jasnell: oh doh, I'd actually spotted that right when starting the review, managed to get distracted, then forgot about it by the time I actually wrote the review. Good catch.

kentonv:
> If the changes to the conditions on length are undesired, I can instead add an originalLength field.

FWIW you could compute this when needed (for the exception) as `originalLength = length + amountRead`. No need to store it as a field on the class.

Frederik-Baetens:
Thanks for catching that! Pushed a commit that doesn't alter the logic, but calculates the exptectedLength as suggested. Thanks for the thorough review!

c++/src/kj/compat/http.c++
Frederik-Baetens: Thanks! Switched to that way of formatting the exception!

c++/src/kj/compat/http.c++
```diff
@@ -1956,8 +1956,13 @@ public:
       if (length > 0) {
         // We haven't reached the end of the entity body yet.
         if (amount == 0) {
-          kj::throwRecoverableException(KJ_EXCEPTION(DISCONNECTED,
-              "premature EOF in HTTP entity body; did not reach Content-Length"));
+          size_t expectedLength = length + alreadyRead;
```
jasnell, line 1959: My only challenge with calculating it is that if any bugs sneak in calculating these it might not actually match the value passed in. Not a big deal tho given that such a bug would likely be noticeable elsewhere.

jasnell (approved):

kentonv (approved):

PR #1822, opened 2023-10-07
less ambigous STL string interop SFINAE

Sarcasm:
Hello,

Here is a fix I would be interested to have in the v1 branch (I guess it applies to v2 too) to fix some build issues.

The issue is on a "legacy" codebase with `using namespace boost::filesystem in header files and combined with a recent boost version (1.82).
I considered multiple fixes but I feel like the proper one would be if KJ was stricter in its std::string interop.

### DIFF
--- a/c++/WORKSPACE
+++ b/c++/WORKSPACE
@@ -31,12 +31,12 @@ cc_library(
     name = "zlib",
     srcs = glob(["*.c"]),
     hdrs = glob(["*.h"]),
-    # Temporary workaround for zlib warnings and mac compilation, should no longer be needed with next release https://github.com/madler/zlib/issues/633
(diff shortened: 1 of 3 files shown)


Sarcasm:
The build issue is already fixed on v2: https://github.com/capnproto/capnproto/pull/1764

I'm not sure what's the plan regarding backporting fixes from v2 but I can cherry-pick the commit if that's good with you.

kentonv:
Seems reasonable, so I merged. But I'm a little confused about the boost error. Are you saying it'd fail to compile if you simply #included the two headers with the `using namespace` decl, without trying to actually use them together? Or do you have to have some specific code which is trying to invoke KJ/std string iterop together with using these boost APIs?

PR #1824, opened 2023-10-11
Add -D_FILE_OFFSET_BITS flag

moyodiallo:
`stat` and `lstat` functions appear on many files. We could add the flag once.


kentonv:
Thanks for trying to clean things up, but this won't work. People build Cap'n Proto using many different build systems. This change will break all builds except autotools.

AFAICT the `stat` and `lstat` syscalls are only actually used in filesystem-disk-unix and in some tests. All the places where `stat` and `lstat` appear in other files are actually calling the KJ filesystem API, NOT the raw syscalls. Moreover, `st_size` is only used in `filesystem-disk-unix.c++` and in a test. So I think we are covered as-is.

moyodiallo:
I tried to fix this one from `capnp.c++` file : https://github.com/capnproto/capnproto/blob/c3b4b54f22f11b57e057e1aacd937d83fc9ba5c0/c%2B%2B/src/capnp/compiler/capnp.c%2B%2B#L444-L445

I was getting this error when I was building on a x86_32bit machine which the i-node number exceed 32bit : 
```
./capnp compile --src-prefix=./src -o./capnpc-c++:src -I./src $(for FILE in src/capnp/test.capnp src/capnp/test-import.capnp src/capnp/test-import2.capnp src/capnp/compat/json-test.capnp; do echo ./$FILE; done)
/src/capnproto/c++/.libs/capnp compile: -o ./capnpc-c++:src: output location is inaccessible or is not a directory
Try '/src/capnproto/c++/.libs/capnp compile --help' for more information.
make: *** [Makefile:4465: test_capnpc_middleman] Error 1
"/usr/bin/linux32" "/bin/sh" "-c" "cd capnproto/c++ && autoupdate && autoreconf -i && ./configure && make -j6 && sudo make install" failed with exit status 2
```
My bad, I was not precise on the PR. I think another PR with the same fix as it is in `filesystem-disk-unix.c++` would be Ok.

moyodiallo:
I opened this PR https://github.com/capnproto/capnproto/pull/1830 in which it is more precise.

PR #1827, opened 2023-10-20
Prevent attach(...) functions from taking functors

bcaimano:
This has been a source of bugs where someone intends to attach a _::Deferred but just attaches a functor.

### DIFF
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -99,9 +99,11 @@ KJ_BEGIN_HEADER
 #endif
 
 #include <stddef.h>
+#include <string.h>
+
 #include <cstring>
 #include <initializer_list>
-#include <string.h>
+#include <type_traits>
 
 #if _WIN32
 // Windows likes to define macros for min() and max(). We just can't deal with this.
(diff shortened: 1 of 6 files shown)


c++/src/kj/common.h
```diff
@@ -99,9 +99,11 @@ KJ_BEGIN_HEADER
 #endif
 
 #include <stddef.h>
+#include <string.h>
+
 #include <cstring>
 #include <initializer_list>
-#include <string.h>
+#include <type_traits>
```
kentonv, line 106: Sorry to do this but... Is there any way we can avoid bringing this fairly large and template-y std header into kj/common.h, which is included by everything?

bcaimano: I mean, it is possible of course. But type_traits is by definition for templates which will not generate any code until used *and* it's a somewhat reasonable 142K on my machine. So I guess my return question is to why do we care if it is large? Personally, my only concern would be a notable increase in compile time or binary size.

kentonv: Compile time is my concern. I wouldn't expect any impact on binary size from type_traits. But if every file has to parse it I wonder if that impacts build time...

My other concern is that this is a step away from KJ being independent of std but maybe that's an unrealistic dream I should stop thinking about.

PR #1828, opened 2023-10-20
gracefully handling http 431 error

mikea:
Replaces assertion in `readHeader` with protocol error propagation. I have stopped these errors with assertions in places where appropriate error handling isn't available. 

Returning the error takes special care to give client a chance to read it before dropping the connection.

### DIFF
--- a/c++/src/kj/compat/http-test.c++
+++ b/c++/src/kj/compat/http-test.c++
@@ -43,8 +43,8 @@
 #else
 // Run the test using in-process two-way pipes.
 #define KJ_HTTP_TEST_SETUP_IO \
-  kj::EventLoop eventLoop; \
-  kj::WaitScope waitScope(eventLoop)
+  auto io = kj::setupAsyncIo(); \
+  auto& waitScope KJ_UNUSED = io.waitScope
 #define KJ_HTTP_TEST_SETUP_LOOPBACK_LISTENER_AND_ADDR \
   auto capPipe = newCapabilityPipe(); \
   auto listener = kj::heap<kj::CapabilityStreamConnectionReceiver>(*capPipe.ends[0]); \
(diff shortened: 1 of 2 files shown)


mikea:
I have spent a tremendous amount of time and still can't figure out why in http socket test, the connection is abnormally closed when I return 431.

maybe @jasnell or @kentonv you have any idea? In pipes case the connection is closed gracefully.

mikea:
@kentonv suggests this is happening because write part of the socket is shutdown. The read part is not actually separate in socket case (but it is in pipe case), that's why closing the write closes the read. To increase chances of client seeing the error I'll make sure that logic similar to `canceledUploadGraceBytes` is executed when 431 happens.

c++/src/kj/compat/http.c++
```diff
@@ -2400,6 +2441,7 @@ class HttpDiscardingEntityWriter final: public kj::AsyncOutputStream {
   // Does not inherit HttpEntityBodyWriter because it doesn't actually write anything.
 public:
   Promise<void> write(const void* buffer, size_t size) override {
+    KJ_DBG(size);
```
kentonv, line 2444: KJ_DBG

c++/src/kj/compat/http.c++
```diff
@@ -1637,6 +1667,10 @@ public:
     return { headerBuffer.releaseAsBytes(), leftover.asBytes() };
   }
 
+  kj::Promise<uint64_t> pumpTo(AsyncOutputStream &output, size_t maxBytes) {
```
kentonv, line 1670: I'm concerned this method could easily be mistaken as being consistent with `tryRead()` above -- i.e. in any case where you want to use `tryRead()`, you might expect you can also use `pumpTo()` and get the same data out. But this isn't the case, they are not consistent. Technically since this class doesn't implement `AsyncInputStream`, there's no actual need for these methods to be consistent, but someone could easily accidentally assume that the interface here is meant to correspond to `AsyncInputStream`.

Since this method is _only_ used to discard bytes, could we change it to just do that, and call it `discard(uint64_t)`?

c++/src/kj/compat/http-test.c++
```diff
@@ -4003,60 +4003,103 @@ KJ_TEST("HttpServer threw exception") {
...
-  HttpServer server(timer, table, service);
-
-  auto listenTask = server.listenHttp(kj::mv(pipe.ends[0]));
-
-  static constexpr auto request = "GET / HTTP/1.1\r\nbad request\r\n\r\n"_kj;
-  auto writePromise = pipe.ends[1]->write(request.begin(), request.size());
-  auto response = pipe.ends[1]->readAllText().wait(waitScope);
-  KJ_EXPECT(writePromise.poll(waitScope));
-  writePromise.wait(waitScope);
+KJ_TEST("HttpServer request/response") {
```
kentonv, line 4006: Nitpick: The new name of this test isn't clear that it's testing error cases. I think the old name was better.

```suggestion
KJ_TEST("HttpServer bad request") {
```

jasnell (approved):
LGTM but let's also have @kentonv review and sign off.

Issue #868, opened 2019-08-14
Allow finding a substring in kj's string


xtuc:
Based on
https://github.com/capnproto/capnproto/blob/2ec719865adb64792b07c88ad8298dacf4567c72/c%2B%2B/src/kj/string.h#L129-L130

You can only find a char in a string, however, I would like to find a substring. It would be handy to extend the API.

jclee:
As far as implementation goes... would it be preferable to just overload the existing character functions to search for substrings? e.g.:
```
inline Maybe<size_t> findFirst(StringPtr s) const;
```
Or is it better to use a different name for substring searching?  FWIW, `std::string` overloads "find" and "rfind" to find substrings or individual characters ("substrings of length 1"); on the other hand, it overloads "find_first_of"/"find_last_of" to find sets of characters or individual characters ("sets of size 1").

kentonv:
The best algorithm to find a character is trivial, but the best algorithm to find a substring is somewhat more complicated. Ideally you want it to be O(len(haystack)) but the naive implementation is O(len(haystack)*len(needle)). I think the optimal-throughput algorithm involves a little bit of upfront setup, but that setup may not be worth it when the needle or haystack is short.

Every time I think about adding this I get stuck on being uncertain about how it should be done.

But I guess we should just call `memmem()`? I'd accept a PR that implements these based on `memmem()`.

xtuc:
I would overload. I don't really have a preference on the implementation, but if you are not comfortable having it in kj, we can leave it in our codebase.

jclee:
Sure, memmem() works.  I was just thinking strnstr(), since that's approximately what most kj substring searching is already doing, but yeah, would be nice to handle inline nulls.

kentonv:
Yes it should be an overload.

Yeah, `memmem()` is more correct that `strstr()` since KJ strings are technically allowed to contain NULs. Also, it seems likely that a good implementation of `memmem()` would take advantage of knowing the lengths upfront, which `strstr()` can only discover by scanning. I have no idea if glibc's `memmem()` is good though.

jclee:
[...]

PR #1836, opened 2023-11-07
Remove hasSubstring in favor of StringPtr::contains

justin-mp:
Now that we have StringPtr::contains, hasSubstring is redundant.

Inspired by issue #868.


kentonv:
Despite being in the "private" namespace, it looks like `hasSubstring` is used in a few places in `workerd` and in the internal Workers codebase. I guess we'll need to update those first. (More evidence that this was a good cleanup though.)

justin-mp:
I removed the uses of `kj::_::hasSubstring` in `workerd` and the internal Workers codebase.

kentonv (approved):

PR #1845, opened 2023-11-18
cleanup: Combine vat connection and canceler in to one struct

centromere:
Per the TODO item in `rpc.c++`, this PR is submitted for consideration.

### DIFF
--- a/c++/src/capnp/rpc.c++
+++ b/c++/src/capnp/rpc.c++
@@ -370,7 +370,7 @@ public:
     paf.promise = paf.promise.attach(kj::addRef(*questionRef));
 
     {
-      auto message = connection.get<Connected>()->newOutgoingMessage(
+      auto message = connection.get<Connected>().vatConn->newOutgoingMessage(
           objectId.targetSize().wordCount + messageSizeHint<rpc::Bootstrap>());
 
       auto builder = message->getBody().initAs<rpc::Message>().initBootstrap();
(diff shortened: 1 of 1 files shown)


kentonv:
Hi Alex,

Thanks so much for trying to help and clean up the code. I really appreciate your intent to help out and contribute.

That said, I unfortunately do not have the bandwidth to review changes like this.

That may sound strange, since it probably feels like you've already done all the work and therefore this is a free code improvement for me to merge. However, in practice I almost always find that guiding a new contributor through successfully landing their first few contributions takes more time than it would take me to make the changes myself directly. This comes about both from the need to carefully review changes for problems that the contributor may not be aware they are introducing, and to teach the contributor about our coding conventions. This is an investment I make regularly when it comes to new members of my team, whom I know will continue to contribute for years to come, but when it comes to external contributors I don't know anything about, it's hard to justify the time vs. everything else on my plate.

I'm generally willing to accept simple bugfixes where the fix is trivial to verify, and the real work was discovering and testing for the bug. Your PR #1846 to fix the FreeBSD build is an example; thanks for that. But as a rule I'm generally not up for reviewing refactorings of code that already works correctly, even if the refactoring is simple and prompted by a TODO comment.

Again, I know this is disapointing to hear and I deeply apologize that your time has gone to waste. I really should put this into a contributing guide so that people know what to expect before they start coding.

FWIW, a great way to contribute to Cap'n Proto that doesn't require my review is to work on implementations in languages other than C++, or on other tooling that can be maintained as an independent project.

centromere:
I appreciate the explanation, Kenton, and I empathize with you with respect to my own open-source projects. I was hoping to offer a few small contributions prior to offering a much more substantial contribution. Specifically, I have successfully added a transport based on the [Noise Protocol Framework](https://noiseprotocol.org/) to kj and I have been actively developing a Noise-based VatNetwork with level 3 protocol features.

However, after reading your message, it sounds to me like I should abandon all the work that I have done up to this point, and that my efforts should be directed to other projects instead. Is my understanding correct?

kentonv:
Hmm. A transport based on Noise is something I've always wanted to have, so I might be interested in seeing that. However, there are design details that are important to get right, like:
* How are keys managed and shared?
* Is zero-round-trip startup supported? If so, how?
* Does the design consider the future of three-party handoff (which is not implemented yet), or is it specifically for two-party connections?

I'd be interested in reviewing a design doc. I'd rather not jump into reviewing code without first getting on the same page with regards to design.

centromere:
Thank you Kenton. I've started a discussion [here](https://github.com/capnproto/capnproto/discussions/1850) for better visibility.

PR #1874, opened 2023-12-09
Add AF_VSOCK support to Unix async-io

nhynes:


kentonv:
Can you explain a bit about what this is and what kind of use case you have in mind?

Note that it's possible to import any kind of socket into KJ using `LowLevelAsyncIoProvider`, without having explicit support for parsing addresses of that family...

nhynes:
Hi Kenton, I'm currently using this code in my fork of workerd that supports running within [AWS Nitro Enclaves](https://docs.aws.amazon.com/enclaves/latest/user/nitro-enclave-concepts.html#term-socket), which have the vsock as its only external interface. This patch makes getting HTTP requests into workerd over the vsock trivial.

I did see that I could wrap the `fd` in `LowLevelAsyncIoProvider`, but threading it through alongside the `AsyncIoContext` seemed much more difficult than a central fix. Also I figure that since `AF_VSOCK` is already supported on Linux and Windows, it's widespread enough that other people could benefit from its support here. 

Getting this to run in CI might not be super fun, but QEMU seems to be supported and could do the trick.

kentonv:
Ah, wanting this for use in `workerd` makes sense I guess, and yeah I'd rather add it here than totally re-implement address parsing in workerd just to support this.

Is AF_VSOCK a linux-only feature? I see in your PR here some parts are protected by `#if __linux__`, others by `#if !__APPLE__`, and several diffs aren't guarded at all. Maybe we should guard everything vsock-related with `#ifdef AF_VSOCK`?

nhynes:
> Is AF_VSOCK a linux-only feature?

AF_VSOCK has a [Windows shim provided by AWS](https://docs.aws.amazon.com/enclaves/latest/user/vsock-win.html) but it's not installed anywhere by default, so it's probably safe to assume linux-only for now.

Unfortunately, `AF_VSOCK` is defined for macOS targets and Alpine despite neither distro having a kernel with support. It looks like it might better to add this as a build option. It's also not super easy to test this in CI since it's already virtualized and the vmm doesn't enable vsock. It works great on a bare-metal host, though.

Maybe the best approach here is to add the build option and then leave a closed PR for anyone else who needs this–admittedly uncommon–feature?

kentonv:
I don't think we need to test actually opening one of these sockets. I'm comfortable if we just test that the parsing works. That is, parse the address and then read back the sockaddr and make sure it's right -- without attempting to connect or listen.

c++/src/kj/async-io-test.c++
```diff
@@ -409,6 +452,10 @@ TEST(AsyncIo, AddressParsing) {
   EXPECT_EQ("unix-abstract:foo/bar/baz", tryParse(w, network, "unix-abstract:foo/bar/baz"));
 #endif
 
+#if __linux__
+  EXPECT_EQ("vsock:4294967295:1234", tryParse(w, network, "vsock:-1:1234"));
```
kentonv, line 456: Is this coercion of -1 to an unsigned value common expected behavior when dealing with vsocks? Otherwise to my unfamiliar eye it looks like a bug.

PR #1881, opened 2023-12-17
Add "lifetimebound" annotation to aggregate field accessors of Readers

rianhunter:
Hello,

Recently there was an [article posted to Lobsters](https://blog.dureuill.net/articles/recurring-lifetime/) by @dureuill that described a lifetime "footgun" that exists in capnproto: 

Essentially, when a response is returned from a send() promise, we must store the response in a variable for as long as we access the values that live inside it. If the response contains a list and we leave the response as a temporary while we retrieve the list handle, accesses to the list handle later on will be a UAF, e.g.: `auto list = request.send().wait(wait_scope).getList();`.

@davidchisnall in that Lobsters thread [suggested](https://lobste.rs/s/zzuirh/curiously_recurring_lifetime_issue#c_ymfkof) using `[[clang::lifetimebound]]` to fix this issue and indeed it would have caused the compiler to issue a warning here. All that needs to be done is to add the `[[clang::lifetimebound]]` attribute to all accessor methods in the `Reader` helper classes generated by `capnpc-c++`.

When that is done I get the following error when compiling his original code:

```
./test.cc:18:7: warning: temporary whose address is used as value of local variable 'list' will be destroyed at the end
      of the full-expression [-Wdangling]
      request.send().wait(wait_scope).getList();
      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

This simple change should improve the experience using capnproto and save lots of time debugging.

To get the ball rolling, I've created a minimal patch the fixes the specific problem in question. You may want to expand the patch to apply the same strategy in other cases. I'm not very familiar with capnproto internals so I didn't want to modify too many things.

Thanks,
Rian Hunter

### DIFF
--- a/c++/src/capnp/compiler/capnpc-c++.c++
+++ b/c++/src/capnp/compiler/capnpc-c++.c++
@@ -1674,7 +1674,7 @@ private:
             kj::mv(unionDiscrim.readerIsDecl),
             "  inline bool has", titleCase, "() const;\n",
             COND(shouldExcludeInLiteMode, "#if !CAPNP_LITE\n"),
-            "  inline ", readerType, " get", titleCase, "() const;\n",
+            "  inline ", readerType, " get", titleCase, "() const KJ_LIFETIMEBOUND;\n",
             COND(shouldExcludeInLiteMode, "#endif  // !CAPNP_LITE\n"),
             "\n"),
 
@@ -1723,7 +1723,7 @@ private:
             "}\n",
             COND(shouldExcludeInLiteMode, "#if !CAPNP_LITE\n"),
             templateContext.allDecls(),
-            "inline ", readerType, " ", scope, "Reader::get", titleCase, "() const {\n",
+            "inline ", readerType, " ", scope, "Reader::get", titleCase, "() const KJ_LIFETIMEBOUND {\n",
             unionDiscrim.check,
             "  return ::capnp::_::PointerHelpers<", type, ">::get(_reader.getPointerField(\n"
             "      ::capnp::bounded<", offset, ">() * ::capnp::POINTERS)", defaultParam, ");\n"


kentonv:
I love the concept of this but I suspect the semantics here are more complicated than lifetimebound can correctly represent, leading to false positives.

The problem is, when you do `fooReader.getBar()`, the returned Reader is not really bounded by `fooReader`'s lifetime, it is bounded by the lifetime of the message's backing buffer, i.e. the `MessageReader` instance that `fooReader` originated from. In the _particular_ case of `request.send().wait().getBar()`, these are actually the same object (the returned `Response` both inherits the Reader type of the top-level message, and owns the underlying `MessageReader`). But this isn't the common case.

kentonv:
For a very simple example, say I do:

````
auto response = request.send().wait();
auto list = response.getFoo().getList();
```

Here, the list is embedded in a nested struct. The code doesn't do anything wrong, but will be flagged as an error because `list` outlives `response.getFoo()`. This doesn't matter, because what actually matters is that `list` is outlived by `response`.

kentonv:
I think to accomplish this we'd need to generate another whole class like `T::OwningReader` with all the getter methods with the annotation on them, and then have `capnp::Response<T>` inherit that instead of `T::Reader`. And then, it would only catch those first-level cases. If you had `auto bar = send().wait().getFoo().getBar();`, there wouldn't be any warning, even though `bar` becomes a dangling pointer, because it wasn't a first-level member of the `capnp::Response`.

I guess you might argue it's worth it because the common case will be first-level, and people hopefully hit the warning early on and learn from it. OTOH people might get confused and think that they don't have a problem as long as they don't see the warning...

rianhunter:
> If you had `auto bar = send().wait().getFoo().getBar();`, there wouldn't be any warning

> OTOH people might get confused and think that they don't have a problem as long as they don't see the warning...

Admittedly I assumed the compiler would be able to catch the error in the example you provided. After thinking about this a bit more I think agree with the point you raise. The inconsistency in behavior between top-level accesses and second-level accesses may be confusing. One of those things where you may rarely see it but when you do it will not inspire joy.

I've been trying to think of another solution in the back of my mind. What if there were a way to pass a static object path to a method on the first level object and that method was marked with `[[lifetimebound]]`? It's been a while since I've done major C++ template hacking so an implementation in that vain is not immediately coming to mind. It may also be the case that newer C++20 features make implementing like this easier.

```C++
template<class ReturnType, int arg0, int ...argpath>
ReturnType retrieve() [[clang::lifetimebound]] {
     return static_cast<typename TypeFromArgIndex<arg0>::type>(pointer_retrieve(arg0))->retrieve<ReturnType, ...argpath)();
}
```

There is a lot of handwaving in that code but I think the general framework may work. When I have more free time I may try fleshing out more of a solution.

PR #1882, opened 2023-12-19
as<T>() syntax sugar methods for primitive types

mikea:
String, StringPtr, Array, ArrayPtr gain as<T>() method which is a syntax sugar for T::from(this).

This enables defining conversion domains (such as Std in tests) and use `.as<Std>()` chained calls for more fluid expressions.

### DIFF
--- a/c++/src/kj/array-test.c++
+++ b/c++/src/kj/array-test.c++
@@ -24,6 +24,7 @@
 #include <string>
 #include <list>
 #include <kj/compat/gtest.h>
+#include <span>
 
 namespace kj {
 namespace {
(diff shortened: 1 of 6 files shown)


c++/src/kj/array-test.c++
```diff
@@ -24,6 +24,7 @@
 #include <string>
 #include <list>
 #include <kj/compat/gtest.h>
+#include <span>
```
mikea, line 27: span is since c++ 20, which should be fine?

kentonv (approved):
Arguably in the absence of `from` it might make sense to invoke the constructor but I think we can punt on adding support for that.

PR #1885, opened 2023-12-22
Make getPreferredExtensions pure virtual func

MellowYarker:
Prior to this commit, `WebSocket::getPreferredExtensions()` had a default implementation, which meant derived classes would not be required to implement the method themselves.

Derived classes probably shouldn't refer to some default implementation, so we require it to be implemented going forward.

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -3901,6 +3901,19 @@ public:
     return transferredBytes;
   }
 
+  kj::Maybe<kj::String> getPreferredExtensions(ExtensionsContext ctx) override {
+    KJ_UNREACHABLE;
+  };
+
+  kj::Maybe<WebSocket&> destinationPumpingTo;
+  kj::Maybe<WebSocket&> destinationPumpingFrom;
+  // Tracks the outstanding pumpTo() and tryPumpFrom() calls currently running on the
+  // WebSocketPipeEnd, which is the destination side of this WebSocketPipeImpl. This is used by
+  // the source end to implement getPreferredExtensions().
+  //
+  // getPreferredExtensions() does not fit into the model used by all the other methods because it
+  // is not directional (not a read nor a write call).
+
 private:
   kj::Maybe<WebSocket&> state;
   // Object-oriented state! If any method call is blocked waiting on activity from the other end,
@@ -4019,6 +4032,10 @@ private:
     KJ_FAIL_ASSERT("Bytes are not counted for the individual states of WebSocketPipeImpl.");
    }
 
+  kj::Maybe<kj::String> getPreferredExtensions(ExtensionsContext ctx) override {
+    KJ_UNREACHABLE;
+  };
+
   private:
     kj::PromiseFulfiller<void>& fulfiller;
     WebSocketPipeImpl& pipe;
@@ -4104,6 +4121,10 @@ private:
       KJ_FAIL_ASSERT("Bytes are not counted for the individual states of WebSocketPipeImpl.");
     }
 
+    kj::Maybe<kj::String> getPreferredExtensions(ExtensionsContext ctx) override {
+      KJ_UNREACHABLE;
+    };
+
   private:
     kj::PromiseFulfiller<void>& fulfiller;
     WebSocketPipeImpl& pipe;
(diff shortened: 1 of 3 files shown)


c++/src/kj/compat/http.c++
```diff
@@ -4084,6 +4099,11 @@ private:
       KJ_FAIL_ASSERT("Bytes are not counted for the individual states of WebSocketPipeImpl.");
     }
 
+    kj::Maybe<kj::String> getPreferredExtensions(ExtensionsContext ctx) override {
+      KJ_FAIL_ASSERT(
+          "BlockedReceive WebSocket has no websocket to get a compression preference from.");
```
MellowYarker, line 4104: I wonder if in this case we should forward the call to the `pipe`, but I was under the impression we would have called _from_ the pipe into the `BlockedReceive` websocket, so calling back into the `pipe` would be circular.

c++/src/kj/compat/http.c++
```diff
@@ -4403,19 +4442,45 @@ public:
...
+    return in->pumpTo(other).attach(kj::mv(deferredUnregister));
   }
 
   uint64_t sentByteCount() override { return out->sentByteCount(); }
   uint64_t receivedByteCount() override { return in->sentByteCount(); }
 
+  kj::Maybe<kj::String> getPreferredExtensions(ExtensionsContext ctx) override {
+    // If our destinationPumpingTo doesn't give us preferred extensions, we should check our
+    // destinationPumpingFrom next.
+    KJ_IF_SOME(ws, out->destinationPumpingTo) {
```
MellowYarker, line 4471: It's worth a comment explaining that we read from `out` and set it on `in` when calling `pumpTo()/tryPumpFrom()` by convention to refer to the same pipe. Will add that in the relevant spots tomorrow.

c++/src/kj/compat/http.c++
```diff
@@ -4312,6 +4342,12 @@ public:
   uint64_t sentByteCount() override { return out->sentByteCount(); }
   uint64_t receivedByteCount() override { return in->sentByteCount(); }
 
+  kj::Maybe<kj::String> getPreferredExtensions(ExtensionsContext ctx) override {
+    // TODO(now): Do we want in or out?
```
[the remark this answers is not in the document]
MellowYarker: Thanks for writing this, it was very helpful. Also, sorry for not getting back to this until now. I had briefly skimmed over it during holidays and figured this would be a lot more work, so when other stuff came up I avoided the context switch... Should've just read this properly :man_facepalming:

c++/src/kj/compat/http.c++
```diff
@@ -3901,6 +3901,23 @@ public:
...
+  kj::Maybe<WebSocket&> destinationPumpingTo;
+  kj::Maybe<WebSocket&> destinationPumpingFrom;
+  // Tracks the outstanding pumpTo() and tryPumpFrom() calls currently running on the
+  // WebSocketPipeEnd, which is the destination side of this WebSocketPipeImpl. This is used by
+  // the source end to implement getPreferredExtensions().
+  //
+  // getPreferredExtensions() does not fit into the model used by all the other methods because it
+  // is not directional (not a read nor a write call).
+
+private:
```
kentonv, line 3920: Duplicated `private:`.

```suggestion
```

c++/src/kj/compat/http.c++
```diff
@@ -3901,6 +3901,23 @@ public:
     return transferredBytes;
   }
 
+  kj::Maybe<kj::String> getPreferredExtensions(ExtensionsContext ctx) override {
```
kentonv, line 3904: I believe this method will never be called, since the wrapper will always use the `destinationPumping{To,From}` directly. So the body of this can be `KJ_UNREACHABLE`. Same goes for all the specific state types I think.

c++/src/kj/compat/http.c++
```diff
@@ -4403,19 +4442,45 @@ public:
...
+    in->destinationPumpingTo = other;
+    auto deferredUnregister = kj::defer([this]() { in->destinationPumpingTo = kj::none; });
+    return in->pumpTo(other).attach(kj::mv(deferredUnregister));
   }
 
   uint64_t sentByteCount() override { return out->sentByteCount(); }
   uint64_t receivedByteCount() override { return in->sentByteCount(); }
 
+  kj::Maybe<kj::String> getPreferredExtensions(ExtensionsContext ctx) override {
+    // If our destinationPumpingTo doesn't give us preferred extensions, we should check our
```
kentonv, line 4469: I'd rephrase this comment like:

> We want to forward this call to whatever WebSocket the other end of the pipe is pumping to/from, if any. We'll check them in an arbitrary order and take the first one we see. But really, the hope is that both `destinationPumpingTo` and `destinationPumpingFrom` are in fact the same object. If they aren't the same, then it's not really clear whose extensions we should prefer; the choice here is arbitrary.

kentonv (approved):

PR #1890, opened 2023-12-27
Delete ez-rpc and KJ_THREADLOCAL_PTR

kentonv:
KJ_THREADLOCAL_PTR because it can safely be replaced by plain old `thread_local` these days.

ez-rpc because I've been wanting to delete it forever, and it contained a KJ_THREADLOCAL_PTR, so I figured let's just take the whole file.


kentonv:
Ughhhhhh CI actually builds the samples, I have to update them.

harrishancock (approved):

PR #1891, opened 2023-12-28
Extend UnixEventPort on Linux to allow external polling.

kentonv:
This makes it possible to use UnixEventPort inside a thread running some other event loop, by arranging for that other event loop to listen for the UnixEventPort's epoll FD becoming ready. Whenever it does, the application must pump the KJ event loop.

This also could allow multiple event loops to run in the same thread, or even across some thread pool, as long as only one event loop is active in a thread a time, and each event loop is active in no more than one thread at a time. A scheduler could schedule event loops to threads when their epoll FD becomes ready. 

**Reviewer:** Be sure to read the headers first, they explain a lot of what's going on.

### DIFF
--- a/c++/src/kj/async-unix-test.c++
+++ b/c++/src/kj/async-unix-test.c++
@@ -1244,6 +1244,62 @@ KJ_TEST("UnixEventPoll::getFd() for external waiting") {
     promise.wait(ws);
   }
 }
+
+KJ_TEST("m:n threads:EventLoops") {
+  // This test shows that it's possible for an EventLoop to switch threads, and for a thread to
+  // switch event loops.
+
+  UnixEventPort port1;
+  EventLoop loop1(port1);
+
+  UnixEventPort port2;
+  EventLoop loop2(port2);
+
+  kj::TimePoint startTime = kj::origin<TimePoint>();
+  kj::Promise<void> promise1 = nullptr;
+  PromiseCrossThreadFulfillerPair<void> xpaf { nullptr, {} };
+  const Executor* executor;
+
+  {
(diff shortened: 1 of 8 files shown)


c++/src/kj/async-unix-test.c++
```diff
@@ -1119,6 +1120,188 @@ KJ_TEST("UnixEventPort thread-specific signals") {
 }
 #endif
 
+#if KJ_USE_EPOLL
+KJ_TEST("UnixEventPoll::getFd() for external waiting") {
```
harrishancock, line 1124:
```suggestion
KJ_TEST("UnixEventPoll::getPollableFd() for external waiting") {
```

c++/src/kj/async-unix.c++
```diff
@@ -647,6 +650,19 @@ bool UnixEventPort::processEpollEvents(struct epoll_event events[], int n) {
...
+      int tfd = KJ_ASSERT_NONNULL(timerFd).get();
+      char buffer[16];
+      ssize_t n;
+      KJ_NONBLOCKING_SYSCALL(n = read(tfd, buffer, sizeof(buffer)));
+      KJ_ASSERT(n == 8 || n < 0);
+
+      timerfdIsArmed = false;
+
+      // The purpose of this event is just to wake up the event loop when needed. We'll check the
+      // timer queue separately, so we don't need to do anything special it response to this event
```
harrishancock, line 664:
```suggestion
      // timer queue separately, so we don't need to do anything special in response to this event
```

harrishancock (approved):
Very cool! Only had some comment nits.

PR #1894, opened 2024-01-03
Fix compilation issue for standards prior to C++ 20

uniflare:
Remove usage of designated member assignment in kj string header for const string attachments.
Related to discussion: https://github.com/capnproto/capnproto/commit/2e69f54eb8fcaa0d2c4c96a82005b5842e633392#r135989017

I am unsure of your branch integration direction, so I guessed it would be master to v2. So I PR'd on master.

### DIFF
--- a/c++/src/kj/string.h
+++ b/c++/src/kj/string.h
@@ -748,7 +748,7 @@ inline ConstString StringPtr::attach() const {
 
 template <typename... Attachments>
 inline ConstString StringPtr::attach(Attachments&&... attachments) const {
-  return ConstString { .content = content.attach(kj::fwd<Attachments>(attachments)...) };
+  return ConstString { content = content.attach(kj::fwd<Attachments>(attachments)...) };
 }
 
 inline String::operator ArrayPtr<char>() {


kentonv:
PR to `master` is correct, because the `v2` branch requires C++20, therefore this is not a problem for `v2`.

c++/src/kj/string.h
```diff
@@ -748,7 +748,7 @@ inline ConstString StringPtr::attach() const {
 
 template <typename... Attachments>
 inline ConstString StringPtr::attach(Attachments&&... attachments) const {
-  return ConstString { .content = content.attach(kj::fwd<Attachments>(attachments)...) };
+  return ConstString { content = content.attach(kj::fwd<Attachments>(attachments)...) };
```
kentonv, line 751: Does this actually compile? I'd expect this to be a compiler error, since the code now performs an assignment to `this->content`, but `this` is `const`.

I think what you want here is:

```suggestion
  return ConstString(content.attach(kj::fwd<Attachments>(attachments)...));
```

uniflare: Oops yeah good point.
Looks like I was getting the original error without calling this code, so this change wasn't being compiled anyway.
Cleaned it up a bit and moved to a new PR from a new branch: https://github.com/capnproto/capnproto/pull/1898

PR #1895, opened 2024-01-03
Move GlobFilter into a separate header/unit

jasnell:
Addresses a long standing todo in workerd where we duplicate GlobFilter there.


c++/src/kj/test-test.c++
```diff
@@ -21,6 +21,7 @@
 
 #include "common.h"
 #include "test.h"
+#include "glob-filter.h"
```
kentonv, line 24: Should the tests be moved into a `glob-filter-test.c++`?

kentonv (commented):
I think you need to update the cmake files too.

jasnell:
> I think you need to update the cmake files too.

Hmm... added but obviously there's still something wrong. Not quite sure what. cmake builds successfully locally but I'm obviously missing something

c++/Makefile.am
```diff
@@ -274,6 +275,7 @@ libkj_la_SOURCES=                                              \
   src/kj/string.c++                                            \
   src/kj/string-tree.c++                                       \
   src/kj/source-location.c++                                   \
+  src/kj/glob-filter.c++
```
kentonv, line 278: You're missing the trailing backslash on this line.

jasnell: doh! lol sigh.

kentonv (approved):

Issue #1903, opened 2024-01-12
No way to pass extra language inputs


sgammon:
When using the Capnproto tools for C++ from Bazel with other language plugins (for example, Cap'n'Proto for Java), there is no way to pass extra inputs to be included during the generation phase.

These inputs differ because they are not Cap'n'Proto sources, and so do not provide transitive input support. Effective targets that generate code into C++ libraries only need these inputs to complete the codegen phase, so the new inputs are not included transitively or exported to other targets.

Consider this definition:
```capnp
using Cxx = import "/capnp/c++.capnp";
using Java = import "/capnp/java.capnp";

$Java.package("...");
$Java.outerClassname("...");
$Cxx.namespace("...");

# definitions...
```

When generated with:
```starlark
load("@capnp-cpp//src/capnp:cc_capnp_library.bzl", "cc_capnp_library")

cc_capnp_library(
    name = "some-capnp",
    srcs = ["some.capnp"],
    visibility = ["//visibility:public"],
)

# bazel build //:some-capnp
```

The following error surfaces, even though we are not generating Java:
```
ERROR: /.../BUILD.bazel:96:20: GenCapnp some.capnp.h failed: (Exit 1): capnp_tool failed: error executing GenCapnp command (from target //:some-capnp) 
  (cd /private/var/tmp/_bazel_sam/.../_main && \
  exec env - \
  bazel-out/.../capnp_tool compile --verbose -obazel-out/.../src -Isrc -I external/capnp-cpp/src --src-prefix src some.capnp)
# Configuration: ...
# Execution platform: @@local_config_platform//:host

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
some.capnp:...: error: Import failed: /capnp/java.capnp
some.capnp:...: error: Not defined: Java
some.capnp:...: error: Not defined: Java
```

PR #1904, opened 2024-01-12
feat: extra language inputs for bazel codegen

sgammon:
## Summary

When using language plugins like Cap'n'proto for Java from Bazel, there needs to be a way to add language extras as inputs to the codegen phase.

This is only needed when generating capnp code in multiple languages. Even if the generator is only producing a C++ output in a given invocation, the imports for other languages will fail to be found due to Bazel's strict sandboxing. This new attribute fixes that condition.

Fixes and closes capnproto/capnproto#1903.

## Changelog

- add `language_extras` argument to `cc_capnp_library` macro
- add `language_extras` `label_list` attribute to rule definition
- process such inputs, if provided, into `inputs` and `args` (`-I <parent_dir_of_extra>`)

### DIFF
--- a/c++/src/capnp/cc_capnp_library.bzl
+++ b/c++/src/capnp/cc_capnp_library.bzl
@@ -49,8 +49,18 @@ def _capnp_gen_impl(ctx):
 
     args.add_all([s for s in ctx.files.srcs])
 
+    # process language extras, starting with files we should include
+    extra_inputs = ctx.files.language_extras or []
+
+    # then compute input flags, taking the parent of the first file in each extra as the include path
+    if len(ctx.attr.language_extras) > 0:
+        for extra in ctx.attr.language_extras:
+            files = extra.files.to_list()
+            if len(files) > 0:
+                args.add_all(["-I", files[0].dirname.removesuffix("/capnp")])
+
     ctx.actions.run(
-        inputs = inputs + ctx.files._capnpc_cxx + ctx.files._capnpc_capnp + ctx.files._capnp_system,
+        inputs = inputs + ctx.files._capnpc_cxx + ctx.files._capnpc_capnp + ctx.files._capnp_system + extra_inputs,
         outputs = ctx.outputs.outs,
         executable = ctx.executable._capnpc,
         arguments = [args],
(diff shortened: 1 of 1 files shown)


sgammon:
@kentonv / @mikea any objection to merging this one first? I can then rebase it out of my bigger PR for Bzlmod. This one should be pretty inert and sensible.

mikea:
Sorry for delay. Multi language support is something I wanted to add myself. I was hoping to do it similar to the way protobuf is structured, but quickly ran into limitations of my bazel knowledge. 


This change looks much simpler, however I don't understand how does it help with the task? Where are additional plug-in arguments specified?

Could you expand a little on the usage too? How do you use it to, generate another language? (Not sure what's your interest but I was looking into rust)

c++/src/capnp/cc_capnp_library.bzl
```diff
@@ -49,8 +49,18 @@ def _capnp_gen_impl(ctx):
...
+    # then compute input flags, taking the parent of the first file in each extra as the include path
+    if len(ctx.attr.language_extras) > 0:
+        for extra in ctx.attr.language_extras:
+            files = extra.files.to_list()
+            if len(files) > 0:
+                args.add_all(["-I", files[0].dirname.removesuffix("/capnp")])
+
     ctx.actions.run(
-        inputs = inputs + ctx.files._capnpc_cxx + ctx.files._capnpc_capnp + ctx.files._capnp_system,
+        inputs = inputs + ctx.files._capnpc_cxx + ctx.files._capnpc_capnp + ctx.files._capnp_system + extra_inputs,
```
sgammon, line 63: @mikea Here's the entire change. If the user passes additional targets via the `language_extras` attribute, they are included inputs in the run, and each directory _containing_ the language extras is included via `-I`.

The effect is that the user can pass arbitrary dependencies to be exposed in the sandbox, and can include them easily within the `capnp` sources.

We're using this change downstream in an internal Bazel project, and it's working great, specifically to generate Java capnp sources.

I actually know exactly what you're talking about re/`rules_proto`. It is a little brain-melting to get aspects all working together properly, certainly not for the faint of heart. I've died on that same hill before.

So, I think you're right, it would be _fantastic_ to get true first-class Bazel support for capnp in the same way Protocol Buffers is supported, including easy generation of different languages. But in the meantime, this gives capnp users an escape hatch in Bazel builds where there is no explicit support for their language yet.

Thus, these attributes could be kept to use in-house language implementations, or to facilitate easy testing, etc etc, alongside future attributes or rules or aspects which bring support to parity.

sgammon:
Usage example, which I'm happy to add to docs/README:

**`@capnp-java//:capnpj_system_library`**:
```starlark
filegroup(
   name = "capnpj_system_library",
   srcs = [
       "java.capnp",
   ],
)
```

**`BUILD.bazel`**:
```starlark
cc_capnp_library(
    src_prefix = "src",
    strip_include_prefix = "/src",
    language_extras = ["@capnp-java//:capnpj_system_library"],
    **kwargs
)
```

PR #1905, opened 2024-01-13
Add new KJ_DISALLOW_IN_COROUTINE mechanism

jasnell:
Allows a type to be explicitly marked in a way that prevent it from being passed into a kj promise coroutine as an arg.

```
class Foo {
private:
  KJ_DISALLOW_IN_COROUTINE();
};

kj::Promise<void> simpleCoroutine(Foo foo) { // Compile Error!
  co_return;
}

kj::Promise<void> simpleCoroutine2(Foo& foo) { // Compile Error!
  co_return;
}

kj::Promise<void> simpleCoroutine3(Foo* foo) { // Compile Error!
  co_return;
}
```

### DIFF
--- a/c++/src/kj/async-inl.h
+++ b/c++/src/kj/async-inl.h
@@ -2062,6 +2062,7 @@ template <typename T>
 concept NoWaitScope = !isSameType<Decay<T>, WaitScope>();
 // Define a Concept to use in our `coroutine_traits` specialization to validate allowable coroutine
 // parameter types.
+// TODO(cleanup): This can be removed by adding KJ_DISALLOW_AS_COROUTINE_PARAM to WaitScope.
 
 }  // namespace kj::_
 
@@ -2072,6 +2073,9 @@ namespace KJ_COROUTINE_STD_NAMESPACE {
 template <class T, class... Args>
 struct coroutine_traits<kj::Promise<T>, Args...> {
   // `Args...` are the coroutine's parameter types.
+
+  static_assert((!kj::_::isDisallowedInCoroutine<Args>() && ...),
+      "Disallowed type in coroutine");
 
   static_assert((::kj::_::NoWaitScope<Args> && ...),
       "Coroutines are not allowed to accept `WaitScope` parameters.");
--- a/c++/src/kj/common-test.c++
+++ b/c++/src/kj/common-test.c++
@@ -940,5 +940,33 @@ KJ_TEST("ArrayPtr::as<Std>") {
   KJ_EXPECT(stdPtr.size() == 5);
 }
 
+// Verifies the expected values of kj::isDisallowedInCoroutine<T>
+
+struct DisallowedInCoroutineStruct {
+  KJ_DISALLOW_AS_COROUTINE_PARAM;
+};
+class DisallowedInCoroutinePublic {
+public:
+  KJ_DISALLOW_AS_COROUTINE_PARAM;
+};
+class DisallowedInCoroutinePrivate {
+private:
+  KJ_DISALLOW_AS_COROUTINE_PARAM;
+};
+struct AllowedInCoroutine {};
+
+static_assert(_::isDisallowedInCoroutine<DisallowedInCoroutineStruct>());
+static_assert(_::isDisallowedInCoroutine<DisallowedInCoroutineStruct&>());
+static_assert(_::isDisallowedInCoroutine<DisallowedInCoroutineStruct*>());
+static_assert(_::isDisallowedInCoroutine<DisallowedInCoroutinePublic>());
+static_assert(_::isDisallowedInCoroutine<DisallowedInCoroutinePublic&>());
+static_assert(_::isDisallowedInCoroutine<DisallowedInCoroutinePublic*>());
+static_assert(_::isDisallowedInCoroutine<DisallowedInCoroutinePrivate>());
+static_assert(_::isDisallowedInCoroutine<DisallowedInCoroutinePrivate&>());
+static_assert(_::isDisallowedInCoroutine<DisallowedInCoroutinePrivate*>());
+static_assert(!_::isDisallowedInCoroutine<AllowedInCoroutine>());
+static_assert(!_::isDisallowedInCoroutine<AllowedInCoroutine&>());
+static_assert(!_::isDisallowedInCoroutine<AllowedInCoroutine*>());
+
 }  // namespace
 }  // namespace kj
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -2098,6 +2098,49 @@ _::Deferred<Func> defer(Func&& func) {
 #define KJ_DEFER(code) auto KJ_UNIQUE_NAME(_kjDefer) = ::kj::defer([&](){code;})
 // Run the given code when the function exits, whether by return or exception.
 
+// =======================================================================================
+// IsDisallowedInCoroutine
+
+namespace _ {
+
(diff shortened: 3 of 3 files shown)


c++/src/kj/async-inl.h
```diff
@@ -2063,8 +2063,39 @@ concept NoWaitScope = !isSameType<Decay<T>, WaitScope>();
...
+
+template <typename T>
+struct isDisallowedInCoroutine<T, decltype((void) Decay<T>::__kj_disallowed_in_coroutine, 0)> {
+  static constexpr bool value = true;
+};
+
 }  // namespace kj::_
 
+#define KJ_DISALLOW_IN_COROUTINE() \
+  void* __kj_disallowed_in_coroutine = nullptr; \
```
kentonv, line 2087: This declares a new instance member of every type it is used on, increasing the size by 8 bytes. I think it would be better to use a typedef.

```suggestion
  using _kj_DissalowedInCoroutine = void; \
```

I think this could be detected like:

```
template <typename T>
struct IsDisallowedInCoroutine<T, kj::Decay<T>::_kj_DissalowedInCoroutine> {
```

and for pointers:

```
template <typename T>
struct IsDisallowedInCoroutine<T*, T::_kj_DissalowedInCoroutine> {
```

Though I admit my mental model of SFINAE on types is incomplete.

c++/src/kj/async-inl.h
```diff
@@ -2063,8 +2063,39 @@ concept NoWaitScope = !isSameType<Decay<T>, WaitScope>();
...
+};
+
+template <typename T>
+struct isDisallowedInCoroutine<T, decltype((void) Decay<T>::__kj_disallowed_in_coroutine, 0)> {
+  static constexpr bool value = true;
+};
+
 }  // namespace kj::_
 
+#define KJ_DISALLOW_IN_COROUTINE() \
```
kentonv, line 2086: Bikeshedding:
* Maybe `KJ_DISALLOW_AS_COROUTINE_PARAM`? This more clearly communicates to the reader what is being disallowed -- it's still entirely possible to use the type _inside_ a coroutine, of course (and we can't prevent that unfortunately).
* I'd remove the `()` -- I don't think declare any other macros with empty args in KJ.

c++/src/kj/async-inl.h
```diff
@@ -2063,8 +2063,39 @@ concept NoWaitScope = !isSameType<Decay<T>, WaitScope>();
 // Define a Concept to use in our `coroutine_traits` specialization to validate allowable coroutine
 // parameter types.
 
+struct DisallowedWithinCoroutine {};
```
kentonv, line 2066: Unused?

c++/src/kj/async-inl.h
```diff
@@ -2063,8 +2063,39 @@ concept NoWaitScope = !isSameType<Decay<T>, WaitScope>();
...
+
+template <typename T, typename = int>
+struct isDisallowedInCoroutine {
```
kentonv, line 2069: Nitpick: Types should start with a capital letter.

Typically I will declare a sibling function like:

```
template <typename T>
inline constexpr bool isDisallowedInCoroutine<T>() {
  return IsDisallowedInCoroutine<T>::value;
}
```

This is just for syntax sugar, so callers don't have to add `::value`.

c++/src/kj/async-inl.h
```diff
@@ -2063,8 +2063,39 @@ concept NoWaitScope = !isSameType<Decay<T>, WaitScope>();
...
+
 }  // namespace kj::_
 
+#define KJ_DISALLOW_IN_COROUTINE() \
+  void* __kj_disallowed_in_coroutine = nullptr; \
+  template<typename T, typename> friend struct kj::_::isDisallowedInCoroutine;
+// Certain types are not safe to be passed into and used within a coroutine.
+// The KJ_DISALLOW_IN_COROUTINE() macro allows such types to be marked. It's
+// important to note that this does not prevent the type from being stack
+// allocated within the coroutine.
```
kentonv, line 2092: Maybe:

```suggestion
// Place in the body of a class or struct to indicate that an instance of or reference to this
// type cannot be passed as the parameter to a KJ coroutine. This makes sense, for example, for
// mutex locks, or other types which should never be held across a co_await.
//
// (Types annotated with this likely also should not be used as local variables inside coroutines,
// but there is no way for us to enforce that.)
```

c++/src/kj/async-inl.h
kentonv, line 1: `async-inl.h` is an internal implementation file, so not the right place to declare public interfaces.

Actually, it might make sense to declare this in the base KJ library, not kj-async, because it would make sense to apply to `kj::Locked` (in `kj/mutex.h`).

I'd be fine with putting this declaration in `kj/common.h`.

c++/src/kj/async-inl.h
```diff
@@ -2073,6 +2104,9 @@ template <class T, class... Args>
 struct coroutine_traits<kj::Promise<T>, Args...> {
   // `Args...` are the coroutine's parameter types.
 
+  static_assert((!kj::_::isDisallowedInCoroutine<Args>::value && ...),
+      "Disallowed type in coroutine");
+
   static_assert((::kj::_::NoWaitScope<Args> && ...),
```
kentonv, line 2110: Perhaps we should mark `WaitScope` with this annotation and then remove the separate `static_assert` for it?

jasnell: We certainly can. Let's make sure we're set on the basic implementation first.

jasnell:
Updated based on the feedback.

c++/src/kj/common-test.c++
```diff
@@ -940,5 +940,15 @@ KJ_TEST("ArrayPtr::as<Std>") {
...
 
+// Verifies the expected values of kj::isDisallowedInCoroutine<T>
+
+struct DisallowedInCoroutine {
+  KJ_DISALLOW_AS_COROUTINE_PARAM;
+};
+struct AllowedInCoroutine {};
+
+static_assert(_::isDisallowedInCoroutine<DisallowedInCoroutine>());
+static_assert(!_::isDisallowedInCoroutine<AllowedInCoroutine>());
```
kentonv, line 951: Maybe add tests for reference and pointer types?

c++/src/kj/common.h
```diff
@@ -2098,6 +2098,49 @@ _::Deferred<Func> defer(Func&& func) {
...
+template <typename T>
+constexpr bool isDisallowedInCoroutine() {
+  return IsDisallowedInCoroutine<T>::value;
+}
+
+}  // namespace _
+
+#define KJ_DISALLOW_AS_COROUTINE_PARAM \
+  using _kj_DissalowedInCoroutine = void; \
+  template<typename T> friend constexpr bool kj::_::isDisallowedInCoroutine(); \
```
kentonv, line 2130: Nitpick: I think these friend declarations aren't necessary unless we're expecting the macro to be used in the `private:` section, but I'd argue it should be under `public:` anyway.

kentonv (approved):

c++/src/kj/common-test.c++
jasnell: Ah right, good call.

c++/src/kj/common.h
jasnell: Yep, that's exactly why I added them. I'd prefer to keep them as I'd prefer folks to not have to think much about whether the macro should be public or private.

PR #1909, opened 2024-01-16
Feature: Bzlmod Support

### LINKED ISSUE #1906: Bazel BZLMOD Support
Hi all!
Just wondering if bzlmod support is planned or has been considered? It would be great to have this tool in the bazel central registry!
Cheers!

sgammon:
## Summary

This changeset adds support for Bzlmod, both in the sense that Cap'n'Proto now uses Bzlmod for development dependencies, and also in the sense that this codebase now defines a module which can be published for Cap'n'Proto C++.

Nearly all dependencies for capnp are available on Bzlmod; the sole exception being Brotli, which will still need to be added by downstream projects.

New macros have been added so that Cap'n'Proto can be used in projects downstream, either via `WORKSPACE.bazel` or Bzlmod.

- [x] Add and use macros for repositories and workspace init
- [x] Add initial `MODULE.bazel` definition
- [x] Add integration tests for `bzlmod` and `workspace` use
- [x] Add support for Bazel CI and BCR publishing on release
- [x] Encapsulate rule internals as module-private
- [x] Re-organize settings in `.bazelrc` to share with integration tests

Fixes and closes #1906.

### Follow-up from review

- [ ] Clean-up, reduce PR size as much as possible
- [ ] Add comments to places where dependency management needs to happen
- [ ] Look at `includes=...` issue reported in testing
- [ ] Release guide
- [ ] Update Bazel docs page to include `.bazelrc` note for `c++20` (maybe a copy-paste sample?)
- [ ] Replay changes on top of `master`, file peer PR to decide which should be merged

## Changelog

- feat: add `capnp_repositories` in `defs/repositories.bzl`
- feat: add `capnp_workspace` in `defs/workspace.bzl`
- feat: add `MODULE.bazel` with dev and API dependencies
- feat: add support for `.bazelci` via presubmit config
- test: add integration test for `workspace` via bazel
- test: add integration test for bzlmod support
- feat: add bcr publishing flow (occurs on release)
- docs: add doc page with instructions for use from bazel
- chore: generate initial bzlmod dependency lockfile
- chore: move root `WORKSPACE` logic to `defs` calls
- chore: split `.bazelrc` into `base.bazelrc` to reuse settings
- chore: move `cc_capnp_library` to `internal`


sgammon:
For now, this is on top of my previous PR at capnproto/capnproto#1904, but I could easily extricate it if needed. We are testing it on this revision in our own projects and so far it's working great.

sgammon:
cc / @kentonv

.bazelci/presubmit.yml
```diff
@@ -0,0 +1,29 @@
+---
+matrix:
+  bazel_version: ["7.0.0"]
+  platform: ["centos7", "debian10", "macos", "ubuntu2004"]
```
sgammon, line 4: This file adds support for Bazel CI, which is used for testing with BCR, I believe.

.bcr/metadata.template.json
```diff
@@ -0,0 +1,12 @@
+{
+  "homepage": "https://github.com/capnproto/capnproto",
+  "maintainers": [
+    {
+      "github": "kentonv",
+      "name": "Kenton Varda"
+    }
+  ],
```
sgammon, line 8: Module template; @kentonv, you are listed, and you may want to list others, I don't know. Anybody listed here will show up on the BCR module page. I stopped short of listing your email ;)

.bcr/presubmit.yml
```diff
@@ -0,0 +1,11 @@
...
+bcr_test_module:
+  module_path: "integration_tests/bzlmod"
+  matrix:
+    platform: ["debian10", "macos", "ubuntu2004"]
+  tasks:
+    build_bzlmod_test:
+      name: "Build test module"
+      platform: ${{ platform }}
+      build_targets:
+        - "//:sample-bin"
```
sgammon, line 11: BCR integration tests. Pretty straightforward.

.bcr/source.template.json
```diff
@@ -0,0 +1,5 @@
+{
+  "integrity": "",
+  "strip_prefix": "{REPO}-{VERSION}",
+  "url": "https://github.com/{OWNER}/{REPO}/releases/download/{TAG}/capnproto-cpp-{VERSION}.zip"
```
sgammon: To properly use this library downstream, the repo must use GitHub releases, and must archive each release at this name, as a download artifact on the release.

This ensures a stable hash even if GitHub [changes how their endpoints work](https://github.blog/2023-02-21-update-on-the-future-stability-of-source-code-archives-and-hashes/).

The `on.release.yml` flow will publish the archives at this URL automatically when a GitHub release is published.

.bcr/source.template.json
```diff
@@ -0,0 +1,5 @@
+{
+  "integrity": "",
+  "strip_prefix": "{REPO}-{VERSION}",
+  "url": "https://github.com/{OWNER}/{REPO}/releases/download/{TAG}/capnproto-cpp-{VERSION}.tar.gz"
```
sgammon, line 4: The project will need to start using the Github Releases mechanism, and listing archives there, so that [URLs remain stable](https://github.blog/2023-02-21-update-on-the-future-stability-of-source-code-archives-and-hashes/). This should be considered a requirement both for use from `WORKSPACE` and Bzlmod (BCR will reject the module otherwise).

.github/workflows/ci.bazelrc
```diff
@@ -0,0 +1,21 @@
+# This file contains Bazel settings to apply on CI only.
+# It is referenced with a --bazelrc option in the call to bazel in ci.yaml
```
sgammon, line 2: CI settings; this file is required as part of the BCR release script.

.github/workflows/module.bazelci.yml
```diff
@@ -0,0 +1,29 @@
+---
+name: Bazel CI
+
+# Controls when the action will run.
+"on":
+  workflow_dispatch: {}
```
sgammon, line 6: Dispatchable workflow for testing Bazel CI jobs

.github/workflows/on.release.yml
```diff
@@ -0,0 +1,32 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  release:
+    name: "Release: BCR"
+    uses: bazel-contrib/.github/.github/workflows/release_ruleset.yaml@v2
+    with:
+      release_files: capnp-cpp-*.tar.gz
```
sgammon, line 32: When a tag is pushed, `release-prep.sh` is run by this step, which then grabs the outputs described and publishes them on a Github Release as artifacts. This whole flow is provided by the `bazel-contrib` repository as a reusable workflow, paired with in-repo scripts/configs (i.e. `.bcr`, `.github/workflows/ci.bazelrc`, and so on).

.github/workflows/release_prep.sh
```diff
@@ -0,0 +1,37 @@
...
+
+set -o errexit -o nounset -o pipefail
+
+# Set by GH actions, see
+# https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables
+TAG=${GITHUB_REF_NAME}
+# The prefix is chosen to match what GitHub generates for source archives
+PREFIX="capnp-cpp-${TAG:1}"
+ARCHIVE="capnp-cpp-$TAG.tar.gz"
+git archive --format=tar --prefix=${PREFIX}/ ${TAG} | gzip > $ARCHIVE
```
sgammon, line 11: Effectively, this is the line that generates the `capnp-cpp-${VERSION}.tar.gz` archive

c++/.bazelrc
```diff
@@ -1,38 +1,6 @@
...
-build:windows --compiler=clang-cl
-
-build:windows --cxxopt='/std:c++20' --host_cxxopt='/std:c++20' --verbose_failures
-# The `/std:c++20` argument is unused during boringssl compilation and we don't
-# want a warning when compiling each file.
-build:windows --cxxopt='-Wno-unused-command-line-argument' --host_cxxopt='-Wno-unused-command-line-argument'
-# MSVC disappointingly sets __cplusplus to 199711L by default. Defining /Zc:__cplusplus makes it
-# set the correct value.
-build:windows --cxxopt='/Zc:__cplusplus' --host_cxxopt='/Zc:__cplusplus'
+import %workspace%/../tools/bazel/base.bazelrc
```
sgammon, line 3: I moved the bulk of these settings to `/tools/bazel/base.bazelrc`, where they can be reused from the `./c++` workspace and from the `./integration_tests`. Saves duplication of golden compiler settings.

c++/MODULE.bazel
```diff
@@ -0,0 +1,60 @@
+"Cap'n'Proto for C++"
+
+module(
+    name = "capnp-cpp",
+    version = "1.0.2",  # keep in sync with current version
+)
```
sgammon, line 6: The actual module definition.

c++/MODULE.bazel
```diff
@@ -0,0 +1,60 @@
...
+bazel_dep(
+    name = "zlib",
+    version = "1.3",
+)
+bazel_dep(
+    name = "boringssl",
+    version = "0.0.0-20230215-5c22014",
+    repo_name = "ssl",
+)
```
sgammon, line 29: Full set of API dependencies, passed on to downstream projects

c++/WORKSPACE
```diff
@@ -1,61 +1,17 @@
...
+#
+
+# Import repositories
+load("//defs:repositories.bzl", "capnp_repositories")
+
+# Only Brotli is actually imported here, because it isn't yet available on Bazel Registry.
+capnp_repositories(
+    omit_skylib = True,
+    omit_boringssl = True,
+    omit_zlib = True,
```
sgammon, line 16: The Bazel `WORKSPACE` file for `./c++` now just re-uses the macros that downstream users would be expected to use. This lets the Cap'n'Proto project maintainers focus on their own needs and avoid duplication of code (i.e. because the dependencies used in the project all come from one place).

c++/defs/capnp.bzl
```diff
@@ -0,0 +1,9 @@
+"Provides macro exports for Cap'n'Proto use from Bazel. This API surface is public."
+
+load(
+    "@capnp-cpp//internal:cc_capnp_library.bzl",
+    _cc_capnp_library = "cc_capnp_library"
+)
+
+## Exports.
+cc_capnp_library = _cc_capnp_library
```
sgammon, line 9: Public API surface, usage example:

```starlark
load("@capnp-cpp//defs:capnp.bzl", "cc_capnp_library")

cc_capnp_library(
  # ...
)
```

c++/defs/dependencies.bzl
```diff
@@ -0,0 +1,29 @@
...
+BROTLI = struct(
+    version = "1.1.0",
+    digest = "e720a6ca29428b803f4ad165371771f5398faba397edf6778837a18599ea13ff",
+)
+
+## Zlib version to depend on.
+ZLIB = struct(
+    version = "1.3",
+    digest = "8a9ba2898e1d0d774eca6ba5b4627a11e5588ba85c8851336eb38de4683050a7",
+)
```
sgammon, line 14: Dependencies for `WORKSPACE`-based builds are now defined here, and should be updated here, in sync with `MODULE.bazel`. Unfortunately there is no way to import from `MODULE.bazel` even for simple symbols, so this minimal duplication must exist.

c++/defs/repositories.bzl
```diff
@@ -0,0 +1,78 @@
...
+    # Brotli is not available yet on the Bazel Registry.
+    if not omit_brotli:
+        maybe(
+            http_archive,
+            name = "brotli",
+            sha256 = BROTLI.digest,
+            strip_prefix = "brotli-%s" % BROTLI.version,
+            type = "tgz",
+            urls = ["https://github.com/google/brotli/archive/refs/tags/v%s.tar.gz" % BROTLI.version],
+        )
```
sgammon, line 46: This is the `capnp_repositories` macro, which downstream users use in their `WORKSPACE` file. **This needs to be loaded and called for every downstream project,** including projects which use Bzlmod, because Brotli isn't available on BCR yet.

If users have some of these installed already, or would encounter conflicts if Cap'n'Proto installed them too early in their `WORKSPACE`, they can pass `omit_*` to prevent a dependency from being added (for example, `omit_brotli = True`). Dependencies are added with `maybe(...)` so they won't overwrite existing ones anyway.

For Bzlmod users who must call this function, `enable_bzlmod = True` is passed, which _omits all dependencies available on Bzlmod_, so only Brotli is installed in that case. (looking at you, Brotli).

Usage example:

**In a `WORKSPACE` file**:
```starlark
load("@capnp-cpp//defs:workspace.bzl", "capnp_repositories")

capnp_repositories()
```

**For Bzlmod users, in a `WORKSPACE.bzlmod` file**:
```starlark
load("@capnp-cpp//defs:workspace.bzl", "capnp_repositories")

capnp_repositories(enable_bzlmod = True)
```

c++/defs/workspace.bzl
```diff
@@ -0,0 +1,14 @@
...
+
+def capnp_workspace(omit_skylib = False):
+    """Call initialization macros for a Cap'n'Proto project via a Bazel workspace.
+
+    Args:
+        omit_skylib: Don't initialize skylib.
+    """
+
+    if not omit_skylib:
+        bazel_skylib_workspace()
```
sgammon, line 14: This macro, `capnp_workspace`, must be called **only in `WORKSPACE` projects**. Bzlmod projects do not have to load or call this because such initialization is handled automatically by the Bzlmod infrastructure.

Usage example:

**In a `WORKSPACE` file**:
```starlark
load("@capnp-cpp//defs:workspace.bzl", "capnp_workspace")

capnp_workspace()
```

c++/internal/cc_capnp_library.bzl
```diff
@@ -0,0 +1,150 @@
...
+})
+
+def _workspace_path(label, path):
+    if label.workspace_root == "":
+        return path
+    return label.workspace_root + "/" + path
+
+def _capnp_gen_impl(ctx):
+    label = ctx.label
+    src_prefix = _workspace_path(label, ctx.attr.src_prefix) if ctx.attr.src_prefix != "" else ""
```
sgammon, line 16: This stuff is now moved to `c++/internal`, so that it can easily be imported by both the public API layer (so it can be re-exported to users), and imported within the Cap'n'Proto library itself, but without asking users to import from `//src`, etc.

Also makes sure the public API surface is clean, obvious, and maintainable. Targets within the `internal` or `src` directories could therefore be made private to external parties.

Other than moving it (and the notes below), the logic for the `cc_capnp_library` rule is unmodified.

c++/internal/cc_capnp_library.bzl
```diff
@@ -0,0 +1,150 @@
...
+        name,
+        srcs = [],
+        data = [],
+        deps = [],
+        src_prefix = "",
+        language_extras = [],
+        visibility = None,
+        target_compatible_with = None,
+        repo = _DEFAULT_REPO_NAME,
+        rule = native.cc_library,
```
sgammon, line 109: Two very small additions to the macro are these `repo` and `rule` arguments, both with defaults. The `rule` argument will let downstream users customize this macro to use `@rules_cc`, or otherwise wrap the `cc_library` rule defined by `cc_capnp_library`.

The `repo` argument is necessary because Cap'n'Proto now may be installed at a different repo name (than `@capnp-cpp`) within a downstream project's codebase. Using this argument, they can set it to the name they are using so that the runtime dependencies can be resolved properly (`@capnp-cpp//src/capnp_runtime`, etc).

c++/internal/cc_capnp_library.bzl
```diff
@@ -0,0 +1,150 @@
...
+        "_capnpc_cxx": attr.label(executable = True, allow_single_file = True, cfg = "exec", default = "@capnp-cpp//src/capnp:capnpc-c++"),
+        "_capnpc_capnp": attr.label(executable = True, allow_single_file = True, cfg = "exec", default = "@capnp-cpp//src/capnp:capnpc-capnp"),
+        "_capnp_system": attr.label(default = "@capnp-cpp//src/capnp:capnp_system_library"),
+    },
+    output_to_genfiles = True,
+    implementation = _capnp_gen_impl,
+)
+
+_DEFAULT_REPO_NAME = "@capnp-cpp"
+_RUNTIME_REPO_TPL = "%s//src/capnp:capnp_runtime"
```
sgammon, line 96: Note to self: move this to a clearly-public target alias.

c++/src/capnp/cc_capnp_library.bzl
```diff
@@ -1,128 +1,11 @@
...
-        "_capnpc_capnp": attr.label(executable = True, allow_single_file = True, cfg = "exec", default = "@capnp-cpp//src/capnp:capnpc-capnp"),
-        "_capnp_system": attr.label(default = "@capnp-cpp//src/capnp:capnp_system_library"),
-    },
-    output_to_genfiles = True,
-    implementation = _capnp_gen_impl,
+load(
+    "//internal:cc_capnp_library.bzl",
+    _capnp_provider = "capnp_provider",
+    _cc_capnp_library = "cc_capnp_library",
 )
```
sgammon, line 7: Here in `src/capnp/cc_capnp_library.bzl`, we just import from the internals now. I decided to do this import-and-re-export so that I wouldn't need to touch every file in the codebase.

c++/third_party/zlib/zlib.bzl
```diff
@@ -0,0 +1,18 @@
+"Root build file for Zlib."
+
+cc_library(
+    name = "zlib",
+    srcs = glob(["*.c"]),
+    hdrs = glob(["*.h"]),
+    includes = ["."],
+    # Workaround for zlib warnings and mac compilation. Some issues were resolved in v1.3, but there are still implicit function declarations.
```
sgammon, line 8: I've pulled this file out into `c++/third_party/zlib`, rather than being listed inline in the `WORKSPACE` file. This keeps the root `WORKSPACE` clean; users often reference that file (misguided as it may be) for instructions to install a Bazel dependency in their own project.

Plus, `release-prep.sh` uses `awk` on the root `WORKSPACE`, and escaping these file contents may confuse it. Also enables syntax highlighting, which is nice.

doc/bazel.md
```diff
@@ -0,0 +1,85 @@
+---
+layout: page
+title: Usage from Bazel
+---
+
+# Using Cap'n'Proto from Bazel
+
+Cap'n'Proto is available for use directly from downstream Bazel projects, via the two supported mechanisms for Bazel dependencies: `WORKSPACE.bazel` and `MODULE.bazel` (also known as Bzlmod).
```
sgammon, line 8: Here is a doc page for using Cap'n'Proto from Bazel.

sgammon: Note to self: Add to navigation and publish preview

doc/bazel.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Installation via `WORKSPACE`
+
+In your `WORKSPACE.bazel` file:
```
sgammon, line 12: Instructions included both for `WORKSPACE`...

doc/bazel.md
```diff
@@ -0,0 +1,85 @@
...
+## Installation via Bzlmod
+
+In your `MODULE.bazel` file:
+
+```python
+bazel_dep(
+    name = "capnp-cpp",
+    version = "x.x.x",  # find latest version at registry.bazel.build
+)
+```
```
sgammon, line 42: ... and for `MODULE.bazel`.

integration_tests/bzlmod/WORKSPACE.bazel
```diff
@@ -0,0 +1,5 @@
+"Tests use of Cap'n'Proto via Bazel WORKSPACE projects."
+
+workspace(
+    name = "capnp-cpp-bzlmod-test"
+)
```
sgammon, line 5: Integration test for use from `WORKSPACE`...

integration_tests/bzlmod/MODULE.bazel
```diff
@@ -0,0 +1,12 @@
+"Tests use of Cap'n'Proto from a Bzlmod-based project."
+
+module(
+    name = "capnp-test-bzlmod",
+)
```
sgammon, line 5: ... and one for Bzlmod.

integration_tests/bzlmod/src/sample.c++
```diff
@@ -0,0 +1,34 @@
...
+
+int main() {
+  ::capnp::MallocMessageBuilder message;
+
+  AddressBook::Builder addressBook = message.initRoot<AddressBook>();
+  ::capnp::List<Person>::Builder people = addressBook.initPeople(2);
+
+  Person::Builder alice = people[0];
+  alice.setId(123);
+  alice.setName("Alice");
```
sgammon, line 14: Integration tests actually build and use a cap'n'proto from C++ to keep things honest

sgammon (commented):
Quick self-review / tour.

c++/lib/BUILD.bazel
```diff
@@ -0,0 +1,12 @@
...
+alias(
+    name = "runtime",
+    actual = "@capnp-cpp//src/capnp:capnp_runtime",
+    visibility = ["//visibility:public"],
+)
+alias(
+    name = "lib",
+    actual = ":runtime",
+    visibility = ["//visibility:public"],
+)
```
sgammon, line 12: There's now a public target for the Cap'n'Proto runtime library:

```
# Reference as a package
@capnp-cpp//lib

# Or a target
@capnp-cpp//lib:runtime
```

c++/src/capnp/BUILD.bazel
```diff
@@ -1,5 +1,9 @@
 load("@capnp-cpp//src/capnp:cc_capnp_library.bzl", "cc_capnp_library")
 
+package(default_visibility = [
+    "@//:__subpackages__",
+])
+
 cc_library(
     name = "capnp",
     srcs = [
```
sgammon, line 9: All project targets for Cap'n'Proto itself are now only visible within this codebase. This should prevent users from depending directly on targets within `//src/...`.

c++/defs/repositories.bzl
sgammon: Update: Nevermind? I guess Brotli is only used outside the main library, maybe? Because the integration tests didn't complain and didn't have this declaration. Open to advice from somebody who knows better how Brotli is used, maybe we can avoid this.

Attempt3035:
This is beyond fantastic. Thank you so much! I was able to get it up and running straight away with:

```
bazel_dep(name = "capnp-cpp")
git_override(
    module_name = "capnp-cpp",
    remote = "https://github.com/elide-dev/capnproto.git",
    commit = "e79814a9d977be16bbdde95b102de82eaebb404f",
    patch_cmds = ["mv 'c++'/* ."],
)
```

I'm already testing it in some projects, here's a couple of things I've noticed so far you might be interested in!

- ~~With a fairly default setup using bzlmod and bazel 6.4, it doesn't automatically set the c++ standard to 20~~ Silly me, of course you have that sorted. I can see it in the .bazelrc files, probably wasn't effective just because of the way I overrode it. Once it's in the BCR it should be perfect.
- It might be nice if there was a target for just the capnp libs, not just having them included via a compiled file. I was thinking of what my use case would be, but it doesn't really apply to the way I've got my projects structured. However, it's possible that someone wants to define a target (which relies on the generated files) and maybe does some extra functions. This target would have a public header which might include the generated files, but won't necessarily #include all the capnproto headers that they might want to use from other targets that depend on this one. Rather than using the generation macro twice in two different build files, they would just depend on the first target which inherintly provides them access to the generated headers, but might also want to be able to depend upon just capnp as it's own target so they can access other headers from capnp. Not sure about this, maybe that is just a sign of a badly structured build system or code.
- Building from a parent module screws up the macro. If I have a module built that works when cd'd into it, and then go to a parent project which has it's own module file, and try and build the target directly ie `bazel build //Backend/main_app:main_app`, the macro generates the header from the .capnp at a weird location, likely relative to the current parent module file which means it fails to compile the code due to not finding the file in the #include directory. I think that's because of the way some bazel commands like genrule work with the root being the module file, rather than the build file, I've had a similar issue passing in files to tests when trying to make a bazel build for xz (lzma). Not sure if it's really an issue, as it doesn't often make sense to try to build the target this way, (you can simply set it as a dep of the main MODULE and then refer to it with @ and it works perfectly), but thought it was interesting to note from my testing anyway!

Overall, thank you so much for all this work, it's already super helpful to me and working great! Can't wait to see it in the BCR soon!!

c++/defs/repositories.bzl
Attempt3035: Seemed to work fine for me in bzlmod 6.4 without this! 🤷‍♀️

mikea (commented):
I took a quick look mostly at the structure of things, but looks like a great start. 

The complexity here is not to break downstream dependencies (you can check cloudflare/workerd yourself to see if there are potential problems) and not to create much more work for maintainers of these repos and downstream. We might eventually adopt bazelmod for workerd if it is mature enough, but edgeworker will probably be stuck with `WORKSPACE` for some time.

To simplify further review I'd recommend to break this down in several PRs if possible. In particular CI changes and publishing could be separate. Maybe you could see some other opportunities too?

As for publishing PR it should probably include publishing guide, and of course you should discuss publishing strategy with @kentonv since we're mostly interested in rolling consumption.

kentonv:
Hi @sgammon,

Thanks for working on this, this looks like something people want. It's a big change, though, and to be honest I don't have a good grasp of everything that's going on. I'll probably lean on @mikea to review and decide what to accept here. But a couple high-level points:

1. You've made these changes on the `v2` branch, which is our development branch that is intended to be consumed only by Cloudflare Workers ([`workerd`](https://cloudflare/workerd)). While anyone is of course allowed to use this branch if they want, we expect to make (and have made) some big backwards-incompatible changes in this branch. As such, this branch is probably not appropriate to publish to any kind of repository. The `master` branch has been repurposed to be the v1.x release branch, which is what we recommend most people outside the Workers team use. Given this, is the `v2` branch the right place for this change? Should we perhaps focus on publishing the 1.x release?

2. The existing bazel files in this repository were actually written specifically to support `workerd`, and we haven't really put effort into making sure they are appropriate for external consumption. Upgrading these to a public API and publishing it all to a repository implies some commitment to maintenance which I'm not sure our team has bandwidth for. Would you be able to commit to maintaining this? Or is this something you're expecting us to take full ownership of upon merge?

tools/bazel/base.bazelrc
```diff
@@ -0,0 +1,37 @@
+# Baseline settings included in both Cap'n'Proto and in integration tests.
```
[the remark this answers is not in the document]
sgammon: This folder is nested two layers deep intentionally: `tools/bazel` has no `BUILD.bazel` so it is not considered a package. Thus, build resources can be placed here without being importable to downstream projects. Obviously a `.bazelrc` isn't importable this way anyway, so I suppose we could move it if we need to, so long as the nesting remains.

integration_tests/workspace/.bazelversion
```diff
@@ -0,0 +1 @@
+7.0.0
```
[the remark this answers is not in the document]
sgammon: I'm not entirely sure how Bazelisk would handle that, although I'm happy to try. Tests may be run within the `integration_tests/workspace` directory as cwd so it would be reaching backwards up the directory hierarchy. `7.0.0` is also technically the default now, so we could also just delete it.

sgammon:
Hey there @mikea / @kentonv, thanks for the review and the response :)

I'll respond to @kentonv's comments first and then your comments from review @mikea.

> Thanks for working on this, this looks like something people want. It's a big change, though[...]

I just wanted to say, I know the PR _looks_ quite big, but out of 4.5k lines "added", fully 3,952 of those are from lockfiles generated by Bazel:

Screenshot 2024-01-17 at 11 29 14 AM

In terms of code I've _actually added_, it's only 925 lines. Obviously it still touches quite a few files, but I just wanted to point that out so that you know this change is as minimally invasive as possible.

> You've made these changes on the v2 branch, which is our development branch that is intended to be consumed only by Cloudflare Workers

We are actually using Cap'n'Proto downstream within Workers, and WorkerD. So we've already tested against this branch from there (as a side note)

> The master branch has been repurposed to be the v1.x release branch, which is what we recommend most people outside the Workers team use. Given this, is the v2 branch the right place for this change? Should we perhaps focus on publishing the 1.x release?

Understood. I think these changes would replay fine over the `master` branch, I'm happy to amend to clean things up as I go. I will defer to yours/Mike's preference here since our consumption of the library is only for WorkerD. However, we may want to ask OP which version he's looking for in BCR?

> Upgrading these to a public API and publishing it all to a repository implies some commitment to maintenance which I'm not sure our team has bandwidth for. Would you be able to commit to maintaining this?

Yes, I'm happy to maintain the BCR publishing piece and I'm happy to contribute regularly to Cap'n'Proto. I'm a longtime lurker, first time contributor. I already maintain several Bazel modules ([`rules_graalvm`](https://github.com/sgammon/rules_graalvm) most notably), so I'm familiar with the process. We use capnp at work so I could fold it into my regular library maintenance activities, should you want me to.

> The existing bazel files in this repository were actually written specifically to support `workerd`

I think, if you published a serious API change to those Bazel files, you'd quickly hear about users who were depending on them anyway. Bazel users know they shouldn't do this but of course they work with what they can get, and Bazel provides many tools for overriding or using custom rules, so it's hard to prevent. We certainly were doing this in previous projects, and it looks like OP might be.

To the extent possible, this PR should actually make this story much simpler: targets within the Cap'n'Proto source tree (underneath `//src` within `c++`) are now _no longer visible at all_ from downstream projects. This should help the `capnproto` project move freely and feel unencumbered to organize sources as needed, without impacting downstream users.

Placing the rules in an `internal/` folder and redirecting them with an export is how many built-in or Google-official rules packages accomplish this. It also would allow shipping a separate public ruleset and private ruleset, should there be a need in the future for Cloudflare to change how these rules work (but wishes to retain backwards compatibility with public consumers).

Point is, I understand that this repo hasn't really been an _officially supported_ API in that manner, but I think people are already using it, and this change should make that usage safer in the future.

sgammon:
@mikea 

> I took a quick look mostly at the structure of things, but looks like a great start.

Thank you!

> The complexity here is not to break downstream dependencies (you can check cloudflare/workerd yourself to see if there are potential problems) and not to create much more work for maintainers of these repos and downstream.

Understood. We are already testing against `workerd` downstream, in our fork. I can't speak to internal repos, though. There shouldn't be much change in how capnp is developed day to day; the only change would be dependency management, which you pointed out deserves a comment (to keep Bzlmod in sync with `WORKSPACE`).

I'll make sure I add comments to clarify that in the code. Otherwise, it should be really easy. All of the BCR publishing stuff happens in CI when a release is published - automatically - and a PR is filed to make the module live.

I can document that release process, of course, no sweat there. But the entire thing is automated and should happen in the background without being a burden to capnp maintainers.

> We might eventually adopt bazelmod for workerd if it is mature enough, but edgeworker will probably be stuck with WORKSPACE for some time.

Fwiw, I think this could be added to a `MODULE.bazel` file for incremental migration, but I'm not sure how edgeworker differs in structure from the public workerd repository. [`WORKSPACE.bzlmod`](https://bazel.build/external/migration#workspace.bzlmod) provides an ability to easily switch between the two systems and covers toolchain registration as well. These changes also do not *at all* prevent or worsen the use experience from a `WORKSPACE`; on the contrary, it should make it much cleaner to just import these two symbols and call them now. That's definitely a supported route here in addition to Bzlmod and it shouldn't suffer.

sgammon:
Sorry, one more reply, this time addressing your comments @Attempt3035 

> This is beyond fantastic.

Hey, thank you!

> With a fairly default setup using bzlmod and bazel 6.4, it doesn't automatically set the c++ standard to 20

This should be added to the docs, or maybe I can find a way to propagate this flag, but my gut says it'll need to be in `.bazelrc` in downstream projects. In that case it should be added to the docs page I created as part of this PR. I'll add a note there.

> It might be nice if there was a target for just the capnp libs

You should be able to access these at a new target:
```
# As a package
@capnp-cpp//lib

# As a target
@capnp-cpp//lib:runtime
```

This target is cleaner for public consumption rather than reaching into `//src/...`.

>  the macro generates the header from the .capnp at a weird location, likely relative to the current parent module file which means it fails to compile the code due to not finding the file in the #include directory

This might actually be fixed by adding `includes=["."]` to the library target. I'll try some stuff out and see if I can't improve this one.

.bazelci/presubmit.yml
```diff
@@ -0,0 +1,29 @@
...
+bcr_test_module:
+  module_path: "integration_tests/bzlmod"
+  matrix:
+    platform: ["debian10", "macos", "ubuntu2004", "centos7"]
+  tasks:
+    build_bzlmod_test:
+      name: "Build test module"
+      platform: ${{ platform }}
+      build_targets:
+        - "//:sample-bin"
```
sgammon, line 29: Note to self: [this block is for BCR, not BCI](https://github.com/bazelbuild/continuous-integration/issues/1724)

Attempt3035:
Perfect, thank you for clarifying those targets! I'll give them a go and let you know if there's any troubles! For the c++ 20 standard, does .bazelrc get properly used when the module is in BCR? We'd want end users to not have to set the standards of their projects just to get the dep building, I'm guessing it should pass through the entire module from this level down which is great so you aren't specifying cops everywhere, just wanted to make sure that was indeed how it should function in the BCR as a module though. Also, as for which version of the lib in the BCR, I think it would be super important to have the sort of long term release V1 that also has the download link on the capnp website (also use that hosted link for the source maybe? Or just the tag in git, save having extra traffic on your server😝). That way, for the majority of users who will likely be using that version via the curl and make script, they are getting exactly the same version via Bazel. I know I'd prefer that for my projects as I don't need the cutting edge features of v2, stability is much more important. However, there's no reason we can't have both versions in BCR, I'd probably recommend a compatibility version value bump between the two major versions so nothing gets auto-resolved to v2 when it's meant to be v1, but yes, could definitely put v2 once you've got a release system sorted and are tagging them in git for stable urls

sgammon:
@Attempt3035 You are for sure on V1, though? Because the BCR module and default branch on this repo is V2, so you may already be on that. If you tested with the BCR module it must have at least built with V2.

With regard to the `.bazelrc`, there's actually no way that Bazel provides to propagate those. It's on a per-project basis, as part of the root `.bazelrc` (or the user's Bazel settings, or the system-level settings [as described here](https://bazel.build/run/bazelrc)).

So it may need to be added to downstream project `.bazelrc` files as part of installation, unfortunately. For example, look at how [`hermetic_cc_toolchain`](https://github.com/uber/hermetic_cc_toolchain?tab=readme-ov-file#usage) does it:

Screenshot 2024-01-17 at 1 47 45 PM

In any case, if there _is_ a way to propagate those and I don't know about it, I'm all ears, because it definitely complicates installation of Bazel modules and I'd use it in my other ones if I could.

Attempt3035:
Using v2 pulling from your branch for testing, but all previous projects have been with v1.0.2 from the website, https://capnproto.org/capnproto-c++-1.0.2.tar.gz

- I definitely think that version should be first release especially if it's not getting pushed all the beta features etc. For many of the users who want capnp from the BCR it would be best to have the most stable release with just security patches applied when needed. I do think v2 should ultimately end up there too, but it might be a bit more work to keep up with publishing changes and yanking versions if there are security issues or major bugs discovered etc, as it's under such active development. Also, it would be great to simply add BCR as one of the installation methods alongside the others on the website!

As for .bazelrc, I think I understand what you mean. Later today I'll put your fork in a local BCR and do some testing, see if I find anything noteworthy!

Attempt3035:
I've been playing around with setting this up in a local BCR and testing how it would be presumably once in BCR. I'm not sure that the way .bazelrc is currently set up can work with BCR for a couple of reasons:

Assuming the source.json looks something like (this has my testing fork though):
```Starlark
{
  "integrity": "sha256-EL2R/+pjgX665LGQ8O/E2cmFafm2tqu2FrpWwzbMwbU=",
  "strip_prefix": "capnproto-tempTag2/c++",
  "url": "https://github.com/dynacondev/capnproto/archive/refs/tags/tempTag2.tar.gz"
}
```

- The strip_prefix strips down to c++ as that's the root of the module (also not sure if the / works on windows?? If it doesn't, there might need to be a little restructure or something for BCR compatibility as all the relative targets relative to the module file will be wrong if MODULE.bazel is moved to root...), which means in `.bazelrc` the `import %workspace%/../tools/bazel/base.bazelrc` wouldn't work as those top level files and folders are lost in the source download and unpack, however...
- The `.bazelrc` of the module seems to get ignored anyway, however, if I do as you suggest and add these options to my project `.bazelrc`, some of them don't work due to relative paths (`--//src/kj:openssl=True` etc), and they all apply to the entire project which is also not good.

I can't say I'm inclined to follow the way [hermetic_cc_toolchain](https://github.com/uber/hermetic_cc_toolchain?tab=readme-ov-file#usage) does it either as that's modifying the sandbox and seems unnecessary for the options needed?

With my current moderate understanding and experience, I would've thought the best path forward would be to:
1. Define a constant in a .bzl somewhere meaningful which has all the copts you need per system (using the platforms selector). Optionally another for linkopts.
2. Load the .bzl and import the constants in any file that defines build targets that need them. Use those constants there

That would provide both the flags needed automatically and platform dependant, wouldn't require repeated code, and wouldn't require adding anything to a user's `.bazelrc`. It would keep the module self contained and wouldn't affect any parts of projects it's used in. Thoughts?

sgammon:
@Attempt3035 Thanks for this testing; I'll get back to you shortly as I return to this PR to clean it up and rebase.

Attempt3035:
Hey @sgammon! Curious if you're still interested / have time at the moment to finish up this PR, I'm using it in production 😝 on a small project at the moment, would be really good to have capnproto in the BCR and to also have the v1.0.2 with these same mods applied so we can switch to the more stable capnproto! Let me know your thoughts!

sgammon:
Hey, that's great testing! Ha ha. Yes, I'm happy to clean this up, but I can't promise the Cap'n'Proto team will merge it. @kentonv what do you think? Would you guys be willing to accept this change if I was willing to maintain BCR support? We are also using Cap'n'Proto from Bazel and we would be happy to tolerate breaking changes in the future if needed.

kentonv:
I defer to @mikea and @fhanau who tend to be the Bazel experts on our team.

kentonv:
But also want to reiterate: We should not be publishing the `v2` branch to any sort of a package manager, only the 1.x releases, since v2 may undergo breaking changes at any time. People can, of course, pull v2 from git if they really want to, and if these changes make life easier for people pulling from git, then they could make sense on the v2 branch as well.

Attempt3035:
Sounds like it might be best to copy these changes to V1 and add V1 to the Bazel Central Registry, but we can still add this code to V2 without adding to BCR as in Bazel it’s easy to do a git override on a dependency, so users who want V2 can still easily add it to a bazel project

dpemmons:
Now that Bazel 8 is the new LTS release and bzlmod is default for dependency management, are there any updated plans for this pull request?

sgammon:
There are no plans to update this pull request because the Capnproto team has signaled that it won't get merged. If at some later time this changes, please tag me and I'd be happy to help

PR #1910, opened 2024-01-16
Make splitParts toKeyAndVals public

garrettgu10:
These functions are used by Cloudflare Workers, so we're refactoring them to make it part of the official kj public API.


kentonv:
Making these public implies a certain commitment to maintain these APIs in their current form. I don't think we want to make such a commitment, as the APIs at present are pretty janky, and live in the wrong place (they have nothing specifically to do with HTTP, so they shouldn't be in http.h).

If we want to make these public, we need to come up with a better design -- perhaps they could be methods on `ArrayPtr` or `StringPtr`.

But I don't think there's any need to do this. It's fine for Workers to be calling these directly. It just implies that Workers takes responsibility for dealing with it if the APIs change or disappear. Especially since it's actually the same team maintaining both codebases, this isn't a big problem.

PR #1921, opened 2024-01-24
kj::Rc<T> reference counting smart pointer

mikea:


c++/src/kj/refcount.h
```diff
@@ -115,6 +142,102 @@ Own<T> Refcounted::addRefInternal(T* object) {
...
 
+template <typename T>
+Rc<T> Refcounted::addRcRefInternal(T* object) {
+  Refcounted* refcounted = object;
+  ++refcounted->refcount;
+  return Rc<T>(object);
+}
+
+template<typename T>
+class Rc: public Own<T> {
```
jasnell, line 153: I'm generally not a big fan of `kj::Rc<T>` extending from `kj::Own<T>`. Yes it makes migration easier but also makes it easy to miss and get things wrong.

c++/src/kj/refcount.h
```diff
@@ -115,6 +142,102 @@ Own<T> Refcounted::addRefInternal(T* object) {
...
+  // Smart pointer for reference counted objects. 
+  //
+  // There are only two ways to obtain new Rc instances:
+  // - use kj::rc<T>(...) function to create new T.
```
jasnell, line 157:
```suggestion
  // - use kj::Rc<T>(...) function to create new T.
```

mikea: It is actually `kj::rc`.

jasnell: oh! my apologies. I misread this was a function.

jasnell (approved):

mikea:
@kentonv should I wait for your review here? I'll be happy to address any issues later.

c++/src/kj/refcount-test.c++
```diff
@@ -57,6 +57,55 @@ TEST(Refcount, Basic) {
 #endif
 }
 
+KJ_TEST("Rc") {
+  bool b = false;
+
+  Rc<SetTrueInDestructor> ref1 = kj::rc<SetTrueInDestructor>(&b);
```
jasnell, line 63: Not blocking for now but it might be good to have a test that mixes use of refcounted `kj::Own<T>` and `kj::Rc<T>` to make sure the behavior is as expected.

c++/src/kj/refcount.h
```diff
@@ -115,6 +142,105 @@ Own<T> Refcounted::addRefInternal(T* object) {
...
+    return *this;
+  }
+
+  template <typename U>
+  Rc<U> downcast() {
+    auto own = Own<T>::template downcast<U>();
+    return Rc<U>(kj::mv(own));
+  }
+
+  inline bool operator==(decltype(nullptr)) { return this->get() != nullptr; }
```
jasnell, line 207: Non-blocking for now... since we're not exposing the `T*`, implementing `operator==(const Rc<T>&)` would be helpful.

c++/src/kj/refcount-test.c++
mikea: since we don't expose `Own` inheritance, there is no mixed use of `kj::Own<T>` and `kj::Rc<T>` possible.

mikea:
cmake build fails if several tests start on the same line... sigh...

image

kentonv:
> @kentonv should I wait for your review here? I'll be happy to address any issues later.

Yes, please wait.

c++/src/kj/refcount.h
```diff
@@ -115,6 +142,106 @@ Own<T> Refcounted::addRefInternal(T* object) {
...
+template<typename T>
+class Rc: private Own<T> {
+  // Smart pointer for reference counted objects. 
+  //
+  // There are only two ways to obtain new Rc instances:
+  // - use kj::rc<T>(...) function to create new T.
+  // - use kj::Rc::addRef() and the existing Rc instance.
+  // - use EnableAddRefToThis to allow T instance to add new references to itself.
+  //
+  // kj::Rc<T> can be used everywhere Own<T> is expected.
```
kentonv, line 161: Can it? Doesn't the private inheritance block that?

c++/src/kj/refcount.h
```diff
@@ -115,6 +142,106 @@ Own<T> Refcounted::addRefInternal(T* object) {
...
+    return Rc<U>(kj::mv(own));
+  }
+
+  inline bool operator==(const Rc<T>& other) const { return this->get() == other.get(); }
+  inline bool operator==(decltype(nullptr)) const { return this->get() != nullptr; }
+  inline bool operator!=(decltype(nullptr)) const { return this->get() != nullptr; }
+  Rc(decltype(nullptr)) { }
+
+  using Own<T>::operator->;
+  // do not expose * to avoid dangling references
```
kentonv, line 213: It seems unrealistic to me that we're going to avoid this.

c++/src/kj/refcount.h
```diff
@@ -115,6 +142,106 @@ Own<T> Refcounted::addRefInternal(T* object) {
...
 
+template <typename T>
+Rc<T> Refcounted::addRcRefInternal(T* object) {
+  Refcounted* refcounted = object;
+  ++refcounted->refcount;
+  return Rc<T>(object);
+}
+
+template<typename T>
+class Rc: private Own<T> {
```
kentonv, line 153: Why private inheritance instead of composition?

c++/src/kj/refcount.h
```diff
@@ -115,6 +142,106 @@ Own<T> Refcounted::addRefInternal(T* object) {
...
+  //     void setMyService(kj::Rc<MyService>&& service)
+  // - store kj::Rc as data member:
+  //     struct MyComputation { kj::Rc<MyService> service; };
+  //
+
+public:
+  KJ_DISALLOW_COPY(Rc);
+  Rc() { }
+
+  inline Rc(Rc&& other) noexcept: Own<T>(kj::mv(other)) { }
```
kentonv, line 176: Nitpick, I think your move constructor and assignment operator can just be defaulted rather than fully defined.

c++/src/kj/refcount.h
mikea: well, initially it was public, but James has convinced me to drop it. Let me cleanup.

c++/src/kj/refcount.h
mikea: why? let's try it out and see if we need it. Maybe to be on a safer side it will return something reference wrapper?

mikea:
all done, ptal.

mikea:
> I think the ability to convert `kj::Rc` to `kj::Own` is important -- it makes it possible to upgrade an existing object to refcounting without having to modify tons of use sites immediately. Disallowing this will probably discourage use of `kj::Rc`.

Why would you convert to Rc without changing clients? The idea of Rc/Arc is to make refcounting explicit rather than implicit. If you have an Own then there's no guarantee there's no addRef applied to it?

mikea:
I've added `Rc<T>::toOwn()` function.

c++/src/kj/refcount.h
```diff
@@ -115,6 +142,108 @@ Own<T> Refcounted::addRefInternal(T* object) {
...
+template<typename T>
+class Rc {
+  // Smart pointer for reference counted objects. 
+  //
+  // There are only two ways to obtain new Rc instances:
+  // - use kj::rc<T>(...) function to create new T.
+  // - use kj::Rc::addRef() and the existing Rc instance.
+  // - use EnableAddRefToThis to allow T instance to add new references to itself.
+  //
+  // kj::Rc<T> can be used everywhere Own<T> is expected.
```
kentonv, line 161: This line still seems incorrect -- you have to explicitly use `toOwn()` to convert it.

kentonv (approved):

c++/src/kj/refcount.h
mikea: argh, I deleted the wrong line (next sentence instead).
Fixed and rewrote into a separate `-` item. Sorry for that.

PR #1925, opened 2024-01-30
kj::Arc<T> smart poiner for atomic reference counted objects

mikea:
weak references coming later.

### DIFF
--- a/c++/src/kj/refcount.h
+++ b/c++/src/kj/refcount.h
@@ -144,6 +144,7 @@ Own<T> Refcounted::addRefInternal(T* object) {
 
 template <typename T>
 Rc<T> Refcounted::addRcRefInternal(T* object) {
+  static_assert(kj::canConvert<T&, Refcounted&>());
   Refcounted* refcounted = object;
   ++refcounted->refcount;
   return Rc<T>(object);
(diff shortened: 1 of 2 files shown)


c++/src/kj/refcount.h
```diff
@@ -143,7 +144,8 @@ Own<T> Refcounted::addRefInternal(T* object) {
 }
 
 template <typename T>
-Rc<T> Refcounted::addRcRefInternal(T* object) {
+Rc<T> Refcounted::addPtrRefInternal(T* object) {
+  static_assert(std::is_base_of_v<Refcounted, T>);
```
kentonv, line 148: You can use `kj::canConvert<T&, Refcounted&>()` to avoid bringing in `<type_traits>` here.

c++/src/kj/refcount.h
```diff
@@ -89,7 +90,7 @@ class Refcounted: private Disposer {
   static Own<T> addRefInternal(T* object);
 
   template <typename T>
-  static Rc<T> addRcRefInternal(T* object);
+  static Rc<T> addPtrRefInternal(T* object);
```
kentonv, line 93: Why the rename?

jasnell:
LGTM once Kenton's comments are addressed.

c++/src/kj/refcount.h
mikea: I needed the same name to make `EnableAddRefToThis` work with both `Refcounted` and `AtomicRefcounted`

kentonv:
It looks like you force-pushed an update and a rebase at the same time, which is very hard to incrementally review. I figured it out, but please try to avoid that.

kentonv (approved):

PR #1926, opened 2024-01-30
PROPOSAL: kj::Pin<T>, kj::Uniq<T> and kj::Ptr<T> smart reference wrappers

mikea:
`Pin<T>/Ptr<T>` and `Uniq<T>/Ptr<T>` are designed to replace T/&T usage with additional runtime checks for T lifetime.

### DIFF
--- a/c++/src/kj/memory-test.c++
+++ b/c++/src/kj/memory-test.c++
@@ -19,7 +19,12 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
+#include "kj/common.h"
+#include "kj/string.h"
+#include "kj/test.h"
 #include "memory.h"
+#include <csignal>
+#include <cstdint>
 #include <kj/compat/gtest.h>
 #include "debug.h"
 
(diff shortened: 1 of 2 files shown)


c++/src/kj/memory.h
```diff
@@ -656,6 +657,155 @@ class SpaceFor {
   };
 };
 
+#define KJ_ASSERT_PIN_PTRS
```
mikea, line 660: rather than have an assertion I can make:

```
template <typename T, bool ChecksEnabled = true>
class Pin
```
 to be able to use both.

mikea:
Updated to inclue `kj::Uniq<T>`

c++/src/kj/memory.h
smerritt: I like the template better than the macro since I can have some `Pin`s checked but not all.

On the other hand, I don't think I'd ever want a `Pin<T, false>`. If I want a bare pointer, I can just  make a `T` and take its address. If I'm going to the trouble of constructing a `Pin<T>`, then it's because I want safety.

c++/src/kj/memory.h
```diff
@@ -656,6 +657,275 @@ class SpaceFor {
...
+
+  template <typename X, typename... Params>
+  friend Pin<X> pin(Params&&... params);
+
+  template <typename X>
+  friend class Ptr;
+};
+
+template <typename T, typename... Params>
+inline Pin<T> pin(Params&&... params) {
```
kentonv, line 712: Why not give `Pin<T>` a public constructor that forwards args to the inner value?

Does this separate function actually work if `T` itself isn't movable?

c++/src/kj/memory.h
```diff
@@ -656,6 +657,275 @@ class SpaceFor {
...
+  return Pin<T>(T(kj::fwd<Params>(params)...));
+}
+
+template <typename T>
+class Pin<kj::Own<T>> {
+  // Pin<kj::Own<T>> represent fully owned (and also pinned) object on the heap.
```
kentonv, line 718: Could we perhaps just make it so all `Own<T>` can vend `Ptr<T>` directly? Since `Own<T>` already guarantees that its target doesn't move, this seems like it could avoid a lot of boilerplate.

c++/src/kj/memory.h
```diff
@@ -656,6 +657,275 @@ class SpaceFor {
...
+inline Pin<kj::Own<T>> heapPin(Params&&... params) {
+  return Pin<kj::Own<T>>(kj::heap<T>(kj::fwd<Params>(params)...));
+}
+
+// =======================================================================================
+// Uniq<T>
+
+template <typename T>
+class Uniq {
+  // Uniq<T> is a weaker version of Pin<T>: you can move it as long as
```
kentonv, line 764: Why not give this property to all `Pin<T>`? It may be inconsistent with the name "Pin", but ignoring the name, I don't feel like there's any downside, safety-wise?

c++/src/kj/memory.h
```diff
@@ -656,6 +657,275 @@ class SpaceFor {
...
+  inline const T* operator->() const { return get(); }
+
+  inline Ptr<T> operator*() { return Ptr<T>(this); }
+
+private:
+  KJ_DISALLOW_COPY_AND_MOVE(Pin);
+  
+  kj::Own<T> t;
+#ifdef KJ_ASSERT_PIN_PTRS  
+  uint64_t ptrs = 0;
```
kentonv, line 741: This could be `size_t` instead of `uint64_t`, and then you can avoid including `stdint.h`.

Of course, to overflow a 32-bit number here you'd need to have 64GB of memory space just made up of `kj::Ptr<T>` all pointing to the same object, which, if you do, you have bigger problems. So personally I think `uint` would be fine. `kj::Refcounted` uses `uint`. But as long as it's debug-only anyway I don't care that much.

c++/src/kj/memory.h
```diff
@@ -656,6 +657,275 @@ class SpaceFor {
...
+private:
+  KJ_DISALLOW_COPY(Ptr);
+
+  inline Ptr(Pin<T>* pin) : ptr(pin->get())
+#ifdef KJ_ASSERT_PIN_PTRS  
+  , ptrs(&pin->ptrs) 
+#endif
+  {
+#ifdef KJ_ASSERT_PIN_PTRS  
+    (*ptrs)++;
```
kentonv, line 881: What if multiple threads have `Ptr`s to the same thing? Does this need to be atomic?

I think at least if `T` is `const` then it needs to be atomic.

c++/src/kj/memory.h
```diff
@@ -656,6 +657,275 @@ class SpaceFor {
...
+  inline const T* operator->() const { return get(); }
+
+  inline bool operator==(const Pin<T>& other) const { return get() == other.get(); }
+  inline bool operator==(const Pin<kj::Own<T>>& other) const { return get() == other.get(); }
+  inline bool operator==(const Uniq<T>& other) const { return get() == other.get(); }
+  inline bool operator==(const Uniq<kj::Own<T>>& other) const { return get() == other.get(); }
+  inline bool operator==(const Ptr<T>& other) const { return get() == other.get(); }
+
+private:
+  KJ_DISALLOW_COPY(Ptr);
```
kentonv, line 873: Why disallow copying `Ptr`? I would expect to be able to copy pointers.

c++/src/kj/memory.h
```diff
@@ -656,6 +657,275 @@ class SpaceFor {
...
+template <typename T>
+class Ptr {
+public:
+  inline ~Ptr() {
+#ifdef KJ_ASSERT_PIN_PTRS  
+    (*ptrs)--;
+#endif
+  }
+
+  inline T* operator->() { return get(); }
```
kentonv, line 863: I think we need a way, given a `Ptr<T>`, to get a `Ptr` pointing at one of `T`'s members. It should share the same count as the parent, of course.

I am not sure what the syntax should look like. It could be a template where one of the template args is a pointer-to-member, but that'd be pretty verbose...

c++/src/kj/memory-test.c++
```diff
@@ -500,5 +505,214 @@ KJ_TEST("disposeWith") {
...
+    kj::Pin<Obj> localPin = kj::pin<Obj>("a");
+    holder = new PtrHolder { *localPin };
+    KJ_EXPECT(holder->ptr->name == "a"_kj);
+  });
+}
+#endif  
+
+KJ_TEST("heap kj::Pin<T>") {
+  // create heap object pin by surrendering kj::Own
+  kj::Pin<kj::Own<Obj>> pin = kj::heap<Obj>("b");
```
jasnell, line 556: Ergonomically this feels a bit odd. Could it be possible to have

```cpp
kj::Pin<Obj> = kj::heap<Obj>("b"); 
```

c++/src/kj/memory.h
jasnell: Agree here... if I'm reading Kenton's comment here correctly then I agree it would be easier...

```cpp
kj::Own<Obj> obj = kj::heap<Obj>();

kj::Pin<Obj> pinnedObj = kj::mv(obj);

// Moving the pin to a kj::Own<T> only works if it's not pinned
obj = kj::mv(pinnedObj);
```

kentonv: No, my suggestion is when you have `Own<T>`, you shouldn't need `Pin<T>` at all, you can directly create a `Ptr<T>` from an `Own<T>`.

jasnell: @mikea updated me on what was discussed in detail with regards to `kj::Pin<T>`/`kj::Own<T>`/`kj::Unique<T>`(or whatever it ends up being called) and I think we're on the same page now.

c++/src/kj/memory.h
jasnell: Given the choice of these two options, I'd agree that the define (and the ability to turn off the checks entirely in a release build) is likely the better choice. That said, we could do both.

```
template <typename T, bool AlwaysChecked = false>
class Pin;
```

When `AlwaysChecked` is `false`, we fallback to the define. If `true`, we always check this instance regardless of define.

It adds a bit more complexity so not sure if it's a more desirable approach, but it's an option. May be useful as an escape hatch for cases where we really want to check for extra safety.

c++/src/kj/memory.h
```diff
@@ -656,6 +657,275 @@ class SpaceFor {
...
+  inline Uniq(Uniq<T>&& other): t(kj::mv(other.t)) {
+#ifdef KJ_ASSERT_PIN_PTRS  
+    KJ_IREQUIRE(other.ptrs == 0, "Uniq with active references moved");
+#endif
+  }
+
+  inline ~Uniq() {
+#ifdef KJ_ASSERT_PIN_PTRS  
+    KJ_IREQUIRE(ptrs == 0, "Uniq with active references destroyed");
+#endif
```
jasnell, line 777: super minor nit: this check and the one above could be unified to remove duplication. definitely not blocking of course tho.

mikea:
Will reopen a new PR soon with new, almost-final implementation.

PR #1937, opened 2024-02-06
Implement Rc<const T> for copy-on-write

jasnell:
Initial version to request feedback.

### DIFF
--- a/c++/src/kj/refcount.h
+++ b/c++/src/kj/refcount.h
@@ -151,7 +151,7 @@ Rc<T> Refcounted::addRcRefInternal(T* object) {
 
 template<typename T>
 class Rc {
-  // Smart pointer for reference counted objects. 
+  // Smart pointer for reference counted objects.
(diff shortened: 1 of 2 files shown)


c++/src/kj/refcount.h
```diff
@@ -230,6 +232,76 @@ class Rc {
...
+  inline Rc(Rc&& other) noexcept = default;
+
+  kj::Own<const T> toOwn() {
+    // Convert Rc<const T> to Own<const T>.
+    // Nullifies the original Rc<const T>.
+    return kj::mv(own);
+  }
+
+  Rc<T> copy() {
+    return refcounted<T>(*own);
```
jasnell, line 252: This, of course, assumes `T` implements a copy constructor, which `Refcounted` by default implicitly deletes. Should we look for a `clone()` member method instead? Or have the option to pass in a static copy function as part of the template?

PR #1944, opened 2024-02-14
remove lock blocking tracking

mikea:
the feature was never really used downstream, and is removed from there.


kentonv (approved):

fhanau:
A few missing things – `c++/build/configure.bzl` still has a track_lock_blocking variable and the `Inherits from LockSourceLocation to perform an empty base class optimization when lock tracking...` comment in `c++/src/kj/mutex.h` is obsolete.

fhanau (approved):

PR #1945, opened 2024-02-15
[bazel] Avoid linker errors by aligning source/header dependency sets

fhanau:
`cc_library()` targets containing a header file without the corresponding source file could lead to puzzling linker errors due to missing symbols. The source files being moved to different targets have been carefully chosen to keep the main capnp target small while minimizing the need for dependency changes in upstream targets, in some cases dependencies on capnp-rpc and capnpc have been resolved.
This also enables some minor dependency cleanup and linting.

Note that this may still require small changes in upstream repos – PRs for that to follow.

### DIFF
--- a/c++/src/capnp/serialize-text.h
+++ b/c++/src/capnp/serialize-text.h
@@ -21,10 +21,10 @@
 
 #pragma once
 
+#include <capnp/dynamic.h>
+#include <capnp/orphan.h>
+#include <capnp/schema.h>
 #include <kj/string.h>
-#include "dynamic.h"
-#include "orphan.h"
-#include "schema.h"
 
 CAPNP_BEGIN_HEADER
 
--- a/c++/src/capnp/BUILD.bazel
+++ b/c++/src/capnp/BUILD.bazel
@@ -7,7 +7,9 @@ cc_library(
         "arena.c++",
         "blob.c++",
         "c++.capnp.c++",
+        "capability.c++",
         "dynamic.c++",
+        "dynamic-capability.c++",
         "layout.c++",
         "list.c++",
         "message.c++",
@@ -31,7 +33,6 @@ cc_library(
         "generated-header-support.h",
         "layout.h",
         "list.h",
-        "membrane.h",
         "message.h",
         "orphan.h",
         "pointer-helpers.h",
@@ -41,11 +42,8 @@ cc_library(
         "schema.h",
         "schema-lite.h",
         "schema-loader.h",
-        "schema-parser.h",
         "serialize.h",
-        "serialize-async.h",
         "serialize-packed.h",
-        "serialize-text.h",
         "stream.capnp.h",
     ],
     include_prefix = "capnp",
@@ -58,8 +56,6 @@ cc_library(
 cc_library(
     name = "capnp-rpc",
     srcs = [
-        "capability.c++",
-        "dynamic-capability.c++",
         "membrane.c++",
         "persistent.capnp.c++",
         "reconnect.c++",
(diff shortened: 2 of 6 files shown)


mikea (approved):

kentonv:
This is increasingly diverging from the canonical autotools build.

`libcapnp` is not supposed to depend on `libkj-async`, only on `libkj`. But I see in bazel land, it already has this dependency. `capability.c++` is part of the RPC system, and definitely depends on `libkj-async`, which is why in the autotools build it is not included in `libcapnp`. Unfortunately, it is necessary for the `capability.h` header to be included by some parts of `libcapnp`, but those parts don't actually use any symbols from `capability.c++`, so it works out, albeit being a bit weird.

It seems unfortunate that the bazel build has lost the ability to use Cap'n Proto's serialization layer without depending on kj-async. This is probably something we'd want to fix before publishing the bazel build files in any sort of repository.

c++/src/capnp/serialize-text.h
```diff
@@ -22,9 +22,9 @@
 #pragma once
 
 #include <kj/string.h>
-#include "dynamic.h"
-#include "orphan.h"
-#include "schema.h"
+#include "capnp/dynamic.h"
```
kentonv, line 25: Please don't use quoted syntax for absolute-path includes. These should use `<>` syntax. I know that Google feels otherwise but they are crazy. We have used `<>` for all such includes consistently in KJ and Workers code.

fhanau: Updated the includes

fhanau:
> This is increasingly diverging from the canonical autotools build.
> 
> `libcapnp` is not supposed to depend on `libkj-async`, only on `libkj`. But I see in bazel land, it already has this dependency. `capability.c++` is part of the RPC system, and definitely depends on `libkj-async`, which is why in the autotools build it is not included in `libcapnp`. Unfortunately, it is necessary for the `capability.h` header to be included by some parts of `libcapnp`, but those parts don't actually use any symbols from `capability.c++`, so it works out, albeit being a bit weird.
> 
> It seems unfortunate that the bazel build has lost the ability to use Cap'n Proto's serialization layer without depending on kj-async. This is probably something we'd want to fix before publishing the bazel build files in any sort of repository.

My concern here has been to avoid making headers available as part of a library target that does not include the corresponding source file. When building code that includes `capability.h` and depends on symbols from `capability.c++`, this will cause a linker error, which can be confusing for developers of downstream projects. If both files are kept together, the necessary symbols will always be available when the header is available to be included.

In this case there's upsides to separating the capability header and source files, but that introduces an inherent risk of the linker issues described above. I think of this less of an autotools vs. bazel question – the bazel capnp target can be adjusted to only depend on a target that includes the headers of kj-async instead of the library. It's more about having a choice of over-including source files in the dependencies or risking under-including when a binary depends on capability.c++ but doesn't include capnp-rpc. I prefer the over-including approach as having the source and header files together avoids the unexpected linker errors. Either way, there should be a comment in the BUILD file about this issue and perhaps a cleanup TODO to break the header dependency between libcapnp and kj-async.

kentonv:
Yes, I understand your concern. It is confusing that the header can be included when the source file isn't linked.

However, moving `capability.c++` to `libcapnp` defeats the purpose of separating `libcapnp` from `libcapnp-rpc` in the first place, as well as the purpose of separating `libkj` and `libkj-async`. We might as well just merge those if we're going to go this route. But a lot of people like that separation.

> the bazel capnp target can be adjusted to only depend on a target that includes the headers of kj-async instead of the library.

That won't work -- `capability.c++` depends on the implementation of `kj-async`.

PR #1950, opened 2024-02-28
check for empty AggregateConnectionReceiver

vaci:
Empty AggregateConnectionReceivers work fine, except those methods which access the first entry.

### DIFF
--- a/c++/src/kj/async-io-test.c++
+++ b/c++/src/kj/async-io-test.c++
@@ -3087,13 +3087,7 @@ KJ_TEST("AggregateConnectionReceiver empty") {
   int value;
   uint length = sizeof(value);
 
-  KJ_IF_SOME(exception, kj::runCatchingExceptions([&]() {
-    aggregate->getsockopt(0, 0, &value, &length);
-  })) {
-    (void)exception;
-  } else {
-    KJ_FAIL_EXPECT("Expected an exception");
-  }
+  KJ_EXPECT_THROW_MESSAGE("receivers.size() > 0", aggregate->getsockopt(0, 0, &value, &length));
 }
 
 // =======================================================================================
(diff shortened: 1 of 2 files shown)


kentonv:
Should we have a test for this?

Do you want this on the 1.x release branch (`master`), or do you actually use `v2`?

vaci:
I can add a bit of a test just to confirm that the calls no longer barf in this scenario.
I'm still targeting 1.x until the GCC issues are fixed, but can target patch at both branches.

c++/src/kj/async-io-test.c++
```diff
@@ -3080,6 +3080,22 @@ KJ_TEST("AggregateConnectionReceiver") {
...
+  uint length = sizeof(value);
+
+  KJ_IF_SOME(exception, kj::runCatchingExceptions([&]() {
+    aggregate->getsockopt(0, 0, &value, &length);
+  })) {
+    (void)exception;
+  } else {
+    KJ_FAIL_EXPECT("Expected an exception");
+  }
+}
```
kentonv, line 3097:
```suggestion
  KJ_EXPECT_THROW_MESSAGE("receivers.size() > 0", aggregate->getsockopt(0, 0, &value, &length));
```

c++/src/kj/async-io-test.c++
```diff
@@ -3080,6 +3080,22 @@ KJ_TEST("AggregateConnectionReceiver") {
...
+  int value;
+  uint length = sizeof(value);
+
+  KJ_IF_SOME(exception, kj::runCatchingExceptions([&]() {
+    aggregate->getsockopt(0, 0, &value, &length);
+  })) {
+    (void)exception;
+  } else {
+    KJ_FAIL_EXPECT("Expected an exception");
+  }
```
kentonv, line 3096:
```suggestion
  KJ_EXPECT_THROW_MESSAGE("receivers.size() > 0", aggregate->getsockopt(0, 0, &value, &length));
```

kentonv:
Thanks, will merge when CI passes.

PR #1953, opened 2024-03-04
kj/table: fix the initialization of BTreeImpl::MaybeUnit with uint.

balusch:
This constructor is currently unused, so the current issue has no immediate impact.

Signed-off-by: Jianyong Chen <baluschch@gmail.com>


kentonv:
Nice catch, thanks!

PR #1954, opened 2024-03-05
capnp-rpc: retain RpcConnectionState reference in error handling lambda

jclee:
Fixes potential read-after-free that can happen if the connection's shutdown() method throws an exception.

### DIFF
--- a/c++/src/capnp/rpc-test.c++
+++ b/c++/src/capnp/rpc-test.c++
@@ -342,6 +342,9 @@ public:
       }
     }
     kj::Promise<void> shutdown() override {
+      KJ_IF_SOME(e, network.shutdownExceptionToThrow) {
+        return kj::cp(e);
+      }
       KJ_IF_SOME(p, partner) {
         auto paf = kj::newPromiseAndFulfiller<void>();
         p.fulfillOnEnd = kj::mv(paf.fulfiller);
@@ -410,11 +413,16 @@ public:
     }
   }
 
+  void setShutdownExceptionToThrow(kj::Exception&& e) {
+    shutdownExceptionToThrow = kj::mv(e);
+  }
+
 private:
   TestNetwork& network;
   kj::StringPtr self;
   uint sent = 0;
   uint received = 0;
+  kj::Maybe<kj::Exception> shutdownExceptionToThrow = kj::none;
 
   std::map<const TestNetworkAdapter*, kj::Own<ConnectionImpl>> connections;
   std::queue<kj::Own<kj::PromiseFulfiller<kj::Own<Connection>>>> fulfillerQueue;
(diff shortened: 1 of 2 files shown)


c++/src/capnp/rpc-test.c++
```diff
@@ -1303,6 +1311,46 @@ TEST(Rpc, Abort) {
   EXPECT_TRUE(conn->receiveIncomingMessage().wait(context.waitScope) == nullptr);
 }
 
+KJ_TEST("handles exceptions thrown during disconnect") {
+  // This is similar to the earlier "abort" test, but throws an exception on
```
jclee, line 1315: Not sure if there might be a better way to test this condition... I just happened to find that the "abort" test case was one that exercised the exception handler I was interested in.

c++/src/capnp/rpc.c++
```diff
@@ -503,7 +503,8 @@ public:
     auto shutdownPromise = dyingConnection->shutdown()
         .attach(kj::mv(dyingConnection))
         .then([]() -> kj::Promise<void> { return kj::READY_NOW; },
-              [this, origException = kj::mv(exception)](kj::Exception&& shutdownException) -> kj::Promise<void> {
+              [self = kj::addRef(*this), origException = kj::mv(exception)](
```
jclee, line 506: Here I'm assuming it's OK to capture an additional strong reference to the RpcConnectionState in the error handler -- i.e. that doing so doesn't cause a reference cycle that would cause objects to leak... Not sure if there is existing machinery to detect such problems.

If we want something less likely to form cycles (but also less performant), we could move the receiveIncomingMessageError flag into its own refcounted object and refer to it from both RpcConnectionState and this lambda.

jclee:
This was causing some "heap-use-after-free" detections in asan.

The connection setup logic is a bit hard to follow, but what I think is happening is:

- When we set up a new connection (in `RpcSystemBase::Impl::getConnectionState()`), we create a new `RpcConnectionState` object and add it to the `connections` map.  At the same time, we pass it a `disconnectFulfiller` to allow the state object to eventually pass back a `DisconnectInfo` containing a `shutdownPromise` for when the connection shuts down, and using the fulfiller's promise, we add a task that removes the entry from the `connections` map, then adds the `shutdownPromise` to the tasks.

- When it comes time to disconnect (in `RpcConnectionState::disconnect()`), we build a `shutdownPromise` that calls the connection's `shutdown()` method, then suppresses some common exceptions via a catch handler.  One type of suppressed exception is detected via the `receiveIncomingMessageError` member flag, so the catch handler captures a `this` reference.

- In the (uncommon?) case that the `connections` map holds the only remaining reference to the connection state when `disconnect()` is called, the fulfillment of the `disconnectFulfiller` can cause the RPC system to remove the connection state entry and destroy it before the shutdown promise completes.  If the connection's `shutdown()` method throws after that point, the catch lambda may read the `receiveIncomingMessageError` member value from freed memory.

Adding a strong reference to the lambda prevents the invalid read.

kentonv (approved):

c++/src/capnp/rpc.c++
kentonv: This looks correct -- the promise being fulfilled is not itself held by the RpcConnectionState so there's no cycle.

Milek7:
I think this should be backported to master, I'm getting segfaults on this.

PR #1959, opened 2024-03-07
capnp support for kj::Rc

mikea:


mikea:
see internal PR 7405 with demo.

c++/src/capnp/capability.h
```diff
@@ -205,6 +205,17 @@ class Capability::Client {
   // Make a client capability that wraps the given server capability.  The server's methods will
   // only be executed in the given EventLoop, regardless of what thread calls the client's methods.
 
+  template <typename T, typename = kj::EnableIf<kj::canConvert<T*, Capability::Server*>()>>
+  Client(kj::Rc<T>&& server);
```
kentonv, line 209: It's actually sort of unsafe to have a capnp `Server` object be refcounted. In particular it means that `thisCap()` cannot be implemented correctly. I don't think we should encourage it.

Generally the right pattern is to always allocate server objects with `kj::heap`, and then, if you need to be able to refcount it, pass around the `Client` object and make clones of that.

kentonv (changes requested):
Hate to say it but the internal code which uses a refcounted `Server` object is actually wrong and needs to change to not do that. Cap'n Proto really needs the `Client` to manage all refcounting; `Server`s must be allocated with `kj::heap`.

mikea:
> Hate to say it but the internal code which uses a refcounted `Server` object is actually wrong and needs to change to not do that. Cap'n Proto really needs the `Client` to manage all refcounting; `Server`s must be allocated with `kj::heap`.

Do you think I should add a static assertion then that T is not refcounted?

mikea:
Now that #1965 is in it should be safe to pass `Rc<T>` to a client. Removing `Arc<T>` support since server is supposed to belong to a thread. PTAL @kentonv

c++/src/capnp/compiler/capnpc-c++.c++
```diff
@@ -2445,6 +2445,8 @@ private:
           "  explicit Client(::kj::Own< ::capnp::ClientHook>&& hook);\n"
           "  template <typename _t, typename = ::kj::EnableIf< ::kj::canConvert<_t*, Server*>()>>\n"
           "  Client(::kj::Own<_t>&& server);\n"
+          "  template <typename _t, typename = ::kj::EnableIf< ::kj::canConvert<_t*, Server*>()>>\n"
+          "  Client(::kj::Rc<_t>&& server);\n"
```
kentonv, line 2449: This code generator change should require regenerating bootstraps.

Looks like the test for stale bootstraps is an ekam-rule still, so we're missing it in CI, oops: `c++/src/capnp/bootstrap-test.ekam-rule`

kentonv (approved):
Looks good but please run `c++/regenerate-bootstraps.sh` to regenerate checked-in generated code.

kentonv:
Hrm, clearly someone (probably me) forgot to regenerate bootstraps after a previous change!

FWIW it looks like some non-generated files ended up in the regenerate-bootstrap commit.

c++/src/capnp/compiler/capnpc-c++.c++
mikea: I'll take a look at fixing it.

mikea:
> ~Hrm, clearly someone (probably me) forgot to regenerate bootstraps after a previous change!~
> 
> It looks like you regenerated bootstraps using an old version of capnp and/or capnpc-c++.
> 
> FWIW it looks like some non-generated files ended up in the regenerate-bootstrap commit.

fixed. Also fixed the script here: https://github.com/capnproto/capnproto/pull/1970

PR #1963, opened 2024-03-11
Add instructions to install on windows using scoop

GioPat:


GioPat:
FYI I've created the PR on scoop to add Cap'n Proto to the Main bucket: [https://github.com/ScoopInstaller/Main/pull/5473](https://github.com/ScoopInstaller/Main/pull/5473)

kentonv:
Sorry, I haven't heard of scoop before and I'm not comfortable endorsing it like this.

I would be OK with adding a "From Package Managers" section, similar to what appears in the Unix section (with the same disclaimer). We could list Scoop under that (and maybe also Conan and Conda?). This section should appear second, i.e. after "From Release Zip".

GioPat:
Definitely it's better to keep it under a section "From Package Managers"  after "From Release Zip"! I'll do it ASAP. Would you like to keep the link to scoop or not?

kentonv:
I'm fine with listing Scope alongside other package managers that have Cap'n Proto. I just don't want to list it exclusively, that would seem unfair.

GioPat:
You are definitely right, I've moved under the list of package managers!

kentonv:
This is not what I asked for. Scoop is not the only package manager on windows.

Sorry but I just don't have time to continue this thread, let's leave the site as-is.

GioPat:
That's unfortunate since I worked to include it in scoop which is one of the main command line installers.

kentonv:
I will make a note to update the site with a list of package managers supporting Windows in the future. There are several that have Cap'n Proto, including Conda and Conan as I mentioned.

GioPat:
Scoop is not a package manager, it's rather a command line installer. It's simply automating what you are mentioning in the section "From release zip".

kentonv:
For this purpose that's really no different from a package manager.

PR #1965, opened 2024-03-11
Allow Server objects to be refcounted, safely.

kentonv:
Previously, refcounting a Server object could allow you to create multiple ClientHooks pointing to the same server, which caused problems:
* `thisCap()` could return the wrong one.
* The usual guarantee that streaming calls are called one-at-a-time wasn't honored since the two ClientHooks were unaware of each other's stream queues.
* Probably other issues?

This change makes it so that when we try to create a Client for a Server that already has Clients, we reuse the existing ClientHook. This seems to solve all the problems, with the caveat that we can't have revocable and non-revocable Clients pointing to the same object (but I don't think anyone does that today).

### DIFF
--- a/c++/src/capnp/capability.c++
+++ b/c++/src/capnp/capability.c++
@@ -562,6 +562,10 @@ public:
     }
   }
 
+  bool isRevocable() {
+    return revoker != kj::none;
+  }
+
   void revoke(kj::Exception&& e) {
     KJ_IF_SOME(s, server) {
       KJ_ASSERT_NONNULL(revoker).cancel(e);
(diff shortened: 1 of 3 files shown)


mikea (approved):
thanks

c++/src/capnp/capability.c++
```diff
@@ -900,11 +904,31 @@ private:
...
+    KJ_REQUIRE(!kj::downcast<LocalClient>(hook).isRevocable(),
+        "can't create additional Clients for a Server whose existing Client is revocable");
+    return hook.addRef();
+  } else {
+    return kj::refcounted<LocalClient>(kj::mv(server));
+  }
 }
 
 kj::Own<ClientHook> Capability::Client::makeRevocableLocalClient(Capability::Server& server) {
+  KJ_REQUIRE(server.thisHook == nullptr,
```
mikea, line 929: `== kj::none`. CMake build is tripping on this.

c++/src/capnp/capability.c++
```diff
@@ -1148,6 +1172,9 @@ namespace _ {  // private
 
 Capability::Client CapabilityServerSetBase::addInternal(
     kj::Own<Capability::Server>&& server, void* ptr) {
+  KJ_REQUIRE(server->thisHook == kj::none,
```
mikea, line 1175: ditto

PR #1967, opened 2024-03-18
kj::Pin<T>, kj::Ptr<T> smart pointers

mikea:
`Pin<T>/Ptr<T>` pair is designed to replace `T/T&` usage with additional runtime checks for T's lifetime.

### DIFF
--- a/c++/src/kj/memory-test.c++
+++ b/c++/src/kj/memory-test.c++
@@ -19,7 +19,11 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
+#include "kj/common.h"
+#include "kj/string.h"
+#include "kj/test.h"
 #include "memory.h"
+#include <signal.h>
 #include <kj/compat/gtest.h>
 #include "debug.h"
 
@@ -500,5 +504,125 @@ KJ_TEST("disposeWith") {
 
 // TODO(test):  More tests.
 
-}  // namespace
+struct Obj {
+  Obj(kj::StringPtr name) : name(kj::str(name)) { }
+  Obj(Obj&&) = default;
+
+  kj::String name;
+
+  KJ_DISALLOW_COPY(Obj);
+};
+
+struct PtrHolder {
+  kj::Ptr<Obj> ptr;
+};
+
+KJ_TEST("kj::Pin<T> basic properties") {
+  // kj::Pin<T> guarantees that T won't move or disappear while there are active pointers.
+  
+  // pin constructor is a simple argument pass through
+  kj::Pin<Obj> pin("a");
+
+  // pin is a smart pointer and can be used so
+  KJ_EXPECT(pin->name == "a"_kj);
+
+  // pin can be auto converted to Ptr<T>
+  kj::Ptr<Obj> ptr1 = pin;
+  KJ_EXPECT(ptr1 == pin);
+  KJ_EXPECT(pin == ptr1);
+
+  // Ptr<T> is a smart pointer too
+  KJ_EXPECT(ptr1->name == "a"_kj);
+
+  // you can have more than one Ptr<T> pointing to the same object
+  kj::Ptr<Obj> ptr2 = pin;
+  KJ_EXPECT(ptr1 == ptr2);
+  KJ_EXPECT(ptr2->name == "a"_kj);
+
+  // when leaving the scope ptrs will be destroyed first,
+  // so pin will be destroyed without problems
+}
+
+KJ_TEST("moving kj::Pin<T>") {
+  kj::Pin<Obj> pin("a");
+
+  // you can move pin around as long as there are no pointers to it
+  kj::Pin<Obj> pin2(kj::mv(pin));
+  
+  // data belongs to a new pin now
+  KJ_EXPECT(pin2->name == "a"_kj);
+
+  // it is C++ and old pin still points to a valid object
+  KJ_EXPECT(pin->name == ""_kj);
+
+  // you can add new pointers to the pin with asPtr() method as well
+  kj::Ptr<Obj> ptr1 = pin2.asPtr();
+  KJ_EXPECT(ptr1 == pin2);
+  KJ_EXPECT(ptr1->name == "a"_kj);
+
+  {
+    // you can copy pointers
+    kj::Ptr<Obj> ptr2 = ptr1;
+    KJ_EXPECT(ptr2 == ptr1);
+    KJ_EXPECT(ptr2->name == "a"_kj);
+
+    // ptr2 will be auto-destroyed
+  }
+
+  // you can move the pin again if all pointers are destroyed
+  ptr1 = nullptr;
+  kj::Pin<Obj> pin3(kj::mv(pin2));
+  KJ_EXPECT(pin3->name == "a"_kj);
+}
+
+struct Obj2 : public Obj {
+  Obj2(kj::StringPtr name, int size) : Obj(name), size(size) {}
+  int size;
+};
+
+KJ_TEST("kj::Ptr<T> subtyping") {
+  // pin the child
+  kj::Pin<Obj2> pin("obj2", 42);
+
+  // pointer to the child
+  kj::Ptr<Obj2> ptr1 = pin;
+  KJ_EXPECT(ptr1->name == "obj2"_kj);
+  KJ_EXPECT(ptr1->size == 42);
+
+  // pointer to the base
+  kj::Ptr<Obj> ptr2 = pin;
+  KJ_EXPECT(ptr2->name == "obj2"_kj);
+  KJ_EXPECT(ptr2 == pin);
+  KJ_EXPECT(ptr1 == ptr2);
+
+  // pointers can be converted to the base type too
+  kj::Ptr<Obj> ptr3 = kj::mv(ptr1);
+  KJ_EXPECT(ptr3->name == "obj2"_kj);
+  KJ_EXPECT(ptr3 == pin);
+}
+
+#ifdef KJ_ASSERT_PTR_COUNTERS  
+KJ_TEST("kj::Pin<T> destroyed with active ptrs crashed") {
+  PtrHolder* holder = nullptr;
+  
+  KJ_EXPECT_SIGNAL(SIGABRT, {
+    kj::Pin<Obj> obj("b");
+    // create a pointer and leak it
+    holder = new PtrHolder { obj };
+    // destroying a pin when exiting scope crashes
+  });
+}
+
+KJ_TEST("kj::Pin<T> moved with active ptrs crashes") {
+  KJ_EXPECT_SIGNAL(SIGABRT, {
+    kj::Pin<Obj> obj("b");
+    auto ptr = obj.asPtr();
+    // moving a pin with active reference crashes
+    kj::Pin<Obj> obj2(kj::mv(obj));
+  });
+}
+#endif  
+
+} // namespace 
+
 }  // namespace kj
--- a/c++/src/kj/memory.h
+++ b/c++/src/kj/memory.h
@@ -22,6 +22,7 @@
 #pragma once
 
 #include "common.h"
+#include <atomic>
 
 KJ_BEGIN_HEADER
 
@@ -167,6 +168,53 @@ public:
 
   void disposeImpl(void* pointer) const override {}
 };
+
+
+// =======================================================================================
+// Ptr Counters
+
+#ifdef KJ_DEBUG
+#define KJ_ASSERT_PTR_COUNTERS
+// When defined, keeps track of active Ptr<T> instances and asserts validity of their ownership
+#endif
+
+namespace _ {
+#ifdef KJ_ASSERT_PTR_COUNTERS
+
+void atomicPtrCounterAssertionFailed(const char* const);
+
+class AtomicPtrCounter {
+  // AtomicPtrCounter uses atomic operations to keep track of active pointers.
+  // Since no other memory location is observed, memory_order_relaxed is used.
+
+public:
+  inline void dec() {
+    size_t prevCount = count.fetch_sub(1, std::memory_order_relaxed);
+    if (KJ_UNLIKELY(prevCount == 0)) {
+      atomicPtrCounterAssertionFailed("unbalanced inc/dec");
+    }
+  }
+
+  inline void inc() {
+    count.fetch_add(1, std::memory_order_relaxed);
+  }
+
+  inline void assertEmpty() {
+    size_t c = count.load(std::memory_order_relaxed);
+    if (KJ_UNLIKELY(c != 0)) {
+      atomicPtrCounterAssertionFailed("active pointers exist");
+    }
+  }
+
+private:
+  std::atomic<size_t> count = 0;
+};
+
+using PtrCounter = AtomicPtrCounter;
+// Default counter type to use
+
+#endif
+}
 
 // =======================================================================================
 // Own<T> -- An owned pointer.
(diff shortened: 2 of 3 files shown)


mikea:
Forthcoming PRs will address:

- `kj::Own<T>/kj::Ptr<T>` integration (`kj::Own` can be always moved, but can only be destroyed when there are no active pointers)
- Acquiring `kj::Ptr<T>` for another's pointer field.

Presenting this a separate PR since it is a self-contained work, simplifies review and is independently useful.

c++/src/kj/memory.h
```diff
@@ -656,6 +697,172 @@ class SpaceFor {
...
+
+template <typename T>
+class Ptr;
+
+template <typename T>
+class Pin {
+// Pin<T> is a smart, in-place storage for T. 
+//
+// Pin<T> should be created on the stack or used as a data member. It should not be 
+// allocated on the heap.
```
garrettgu10, line 711: Should we enforce this by overriding operator new and operator new[]?

c++/src/kj/memory-test.c++
```diff
@@ -500,5 +505,97 @@ KJ_TEST("disposeWith") {
...
+  // kj::Pin<T> guarantees that T won't move or disappear while there are active pointers
+  kj::Pin<Obj> pin("a");
+
+  // pin is a smart pointer and can be used so
+  KJ_EXPECT(pin->name == "a"_kj);
+
+  // pin can be auto converted to Ptr<T>
+  kj::Ptr<Obj> ptr1 = pin;
+  KJ_EXPECT(ptr1 == pin);
+  KJ_EXPECT(pin == ptr1);
```
jasnell, line 531: What about to/from super types / subtypes? e.g. 

```
// pseudocode

class A : public B {};

Pin<A> a(a);
Pin<B> b = a;
Pin<A> = downcast(b)
```

c++/src/kj/memory.h
jasnell: There's an existing kj utility for enforcing this rule.... `kj::requireOnStack(...)` I think... not sure if it's appropriate for this case tho. @kentonv could say tho

c++/src/kj/memory.h
```diff
@@ -656,6 +697,172 @@ class SpaceFor {
...
+
+  inline Pin(Pin<T>&& other): t(kj::mv(other.t)) {
+  // Move T's ownership. Illegal when live pointers exist.
+#ifdef KJ_ASSERT_PTR_COUNTERS
+    other.ptrCounter.assertEmpty();
+#endif
+  }
+
+  inline ~Pin() {
+  // Destroy a Pin with underlying object. Illegal when live pointers exist.
```
jasnell, line 731: non-blocking nit: for the documentation on this, it's worth stressing that the "Illegal when live pointers exist" is only actually checked when `KJ_ASSERT_PTR_COUNTERS` is defined.

jasnell (approved):
Very happy to see this progressing. LGTM but let's multiple reviews and sign offs.

c++/src/kj/memory.h
mikea: kj::requireOnStack wouldn't work, since `kj::Pin` could be techically on the heap as a data field (which is fine).

overriding operators is best here I think.

c++/src/kj/memory.h
mikea: technically it is still illegal. I made some changes to wording:

```
  // Undefined behavior when live pointers exist, asserted when KJ_ASSERT_PTR_COUNTERS is defined.
```

is this better?

c++/src/kj/memory.h
```diff
@@ -168,6 +170,46 @@ class NullDisposer: public Disposer {
...
+namespace _ {
+#ifdef KJ_ASSERT_PTR_COUNTERS
+class AtomicPtrCounter {
+// AtomicPtrCounter uses atomic operations to keep track of active pointers.
+
+public:
+  void dec() {
+    size_t c = count.load(std::memory_order_consume);
+    KJ_IREQUIRE(c > 0, "unbalanced dec");
+    count.fetch_sub(1, std::memory_order_release);
```
garrettgu10, line 191: I think there's a race condition here. if multiple threads call dec at once, there's a possibility of both observing c == 1 then both subtracting resulting in a missed unbalanced dec (c == -1).

I think it would be more correct to capture the return value of fetch_sub then assert it's nonnegative

garrettgu10 (changes requested):

c++/src/kj/memory.h
garrettgu10: Actually, maybe the memory ordering makes things okay 🫣

AdityaAtulTewari: I think it is better to change the count variable to `int64_t` and then this would be something like:
```cpp
int64_t c = count.fetch_sub(1, std::memory_order_consume);
KJ_IREQUIRE(c == 0, "unbalanced dec");
```

kentonv: > I think it would be more correct to capture the return value of fetch_sub then assert it's nonnegative

I agree with this.

> I think it is better to change the count variable to int64_t and then this would be something like:

Hello Aditya, I'm not sure if we've met.

I don't think we need to make the count signed. It shouldn't be possible for the count to go negative in the first place -- if it can, that already suggest there is undefined behavior. I think we should probably make this function abort the process (not just throw) if it sees the counter wrap around. Then it doesn't matter if we correctly handle the case that the counter was already below zero, because the process will crash immediately when the counter hits zero.

AdityaAtulTewari: Hello Kenton, I don't believe we have met either. 

I do see your point, if the goal is to crash the system immediately. 
The other concern I have is counting up to the size of an unsigned 32 bit integer and overflowing it is possible with single increments on multiple threads in about 30 minutes. I'm not sure how out of scope it is that this can occur, that is another reason to suggest using a 64 bit width integer for this, even if it should be a `uint64_t`. Unfortunately, I'm not super familiar with this system, so I'm not 100% sure what the context is, @garrettgu10 mentioned it in passing and I thought I'd look into the race.

kentonv: The counter is counting the number of pointers pointing at an object. As a size_t it cannot possibly overflow, since the pointers required to overflow it would not fit into memory.

c++/src/kj/memory.h
```diff
@@ -656,6 +698,205 @@ class SpaceFor {
...
+#ifdef KJ_ASSERT_PTR_COUNTERS
+      counter->dec();
+      counter = nullptr;
+#endif
+      ptr = nullptr;
+    }
+  }
+
+  inline T* operator->() { return get(); }
+  inline const T* operator->() const { return get(); }
```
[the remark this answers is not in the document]
jasnell: I think the goal, at least as I understand it, is to strongly discourage use of the bare references as opposed to completely prohibiting it. I'm happy with this, as is. If it proves to be an annoyance in various cases, then we can always add it later?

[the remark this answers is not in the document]
mikea: Here's my thinking here:

- `Ptr<T>` is _the_ replacement for `T&`, obtaining it is easy (you can always clone one too).
- the more restrictive interface, the easier it is to reason about the code. Without `operator*` the value can't escape the pointer.
- I have tried converting several references upstream and so far I didn't need `operator*` (I do need Own/Ptr compatibility since we sometimes have same `T&` pointing to stack and sometimes to heap).

So yeah, I agree with @jasnell : I'd like to give it a go as is.

If it does turn out to be an annoyance, I would rather added `T& unsafeGet<T>(Ptr<T> ptr)` or something similar that is greppable ('*' isn't).

c++/src/kj/memory.h
mikea: Done. I also wrote some notes regarding the use of release/consume memory ordering. I'm not the expert here, is my understanding correct?

I am also not sure if it is guaranteed that you will ever observe `0` previous value when dec/inc are unbalanced.
At this point I think I'd rather dropped the assert: it is not something that could be really triggered by the user (but Ptr implementer).

c++/src/kj/memory.h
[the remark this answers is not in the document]
kentonv: > the more restrictive interface, the easier it is to reason about the code. Without operator* the value can't escape the pointer.

Sure they can. I can always write `*ptr.operator->()`. And if I _need_ a reference, that's exactly what I'm going to do.

I think there are two cases:

1. I don't really need a reference. In this case I, as a developer working in good faith, will try to use `kj::Ptr`. It's not the absence of `operator*` that makes me do this, it's the fact that I want to write safe code.
2. I need a reference, for some reason. I this case I will see that `*ptr` doesn't work, roll my eyes, and write `*ptr.operator->()` instead. Or worse, if I don't know that `*ptr.operator->()` works, I'll be stuck and frustrated.

I don't really buy that hiding `operator*` in itself is going to meaningfully reduce the use of bare references, but I do think it will lead to frustration.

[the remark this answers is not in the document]
mikea: How about `ptr.unsafeGet()`? Could it be a reasonable compromise?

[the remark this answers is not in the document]
mikea: I've been think why would one _need_ a reference, and so far interaction with 3rd party code is the only thing that comes to mind. This is definitely reasonable, but rather rare, so imo anything more explicit than `*ptr` will be a gain.

c++/src/kj/memory-test.c++
```diff
@@ -19,7 +19,12 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
+#include "kj/common.h"
+#include "kj/string.h"
+#include "kj/test.h"
 #include "memory.h"
+#include <csignal>
```
kentonv, line 26:
```suggestion
#include <signal.h>
```

c++/src/kj/memory-test.c++
```diff
@@ -19,7 +19,12 @@
...
+#include <cstdint>
```
kentonv, line 27:
```suggestion
#include <stdint.h>
```

c++/src/kj/memory.h
```diff
@@ -22,6 +22,8 @@
 #pragma once
 
 #include "common.h"
+#include <atomic>
+#include <cstddef>
```
kentonv, line 26:
```suggestion
#include <stddef.h>
```

What makes this header needed, exactly?

c++/src/kj/memory.h
```diff
@@ -168,6 +170,53 @@ class NullDisposer: public Disposer {
...
+// - every dec() observes the effect of corresponding inc() because of release/consume order
+// - when assertEmpty() is called with balanced dec/inc calls, all releases were consumed, and
+//   additional load with consume memory order will see the correct last value.
+// - when assertEmpty() is called with unbalanced dec/inc calls int will observe the result of 
+//   one of the unbalanced inc(), which is always > 0.
+
+public:
+  void dec() {
+    size_t prevCount = count.fetch_sub(1, std::memory_order_consume);
+    KJ_IREQUIRE(prevCount > 0, "unbalanced inc/dec");
```
kentonv, line 197: Should this crash instead of throw on failure? Since it suggests we're going to hit UB shortly anyway.

c++/src/kj/memory.h
```diff
@@ -168,6 +170,53 @@ class NullDisposer: public Disposer {
...
+  }
+
+  void inc() {
+    size_t prevCount = count.fetch_add(1, std::memory_order_release);
+    KJ_IREQUIRE(prevCount >= 0, "unbalanced inc/dec");
```
kentonv, line 202: This always passes since `prevCount` is unsigned. You could fix by doing `add_fetch` instead and then checking that the result is `> 0`.

But I think this check is not needed. `dec()` would have had to fail previously before this check could fail.

c++/src/kj/memory.h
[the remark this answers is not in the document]
kentonv: I still don't believe that this kind of prohibition is going to be helpful but I'm willing to leave it until the first time it actually comes up.

c++/src/kj/memory.h
kentonv: > I am also not sure if it is guaranteed that you will ever observe 0 previous value when dec/inc are unbalanced.

It is guaranteed, there's no way you can skip a value if you're using the result of `fetch_sub()`.

> At this point I think I'd rather dropped the assert:

I'd be OK with that, but what you have currently looks right.

c++/src/kj/memory.h
```diff
@@ -656,6 +705,205 @@ class SpaceFor {
...
+// Zero-overhead replacement for T if KJ_ASSERT_PTR_COUNTERS is not defined.
+
+public:
+  template <typename... Params>
+  inline Pin(Params&&... params) : t(kj::fwd<Params>(params)...) {  }
+  // Create new Pin<T> using corresponding T constructor.
+
+  inline Pin(Pin<T>&& other): t(kj::mv(other.t)) {
+  // Move T's ownership.
+  // Undefined behavior when live pointers exist, asserted when KJ_ASSERT_PTR_COUNTERS is defined.
```
kentonv, line 733:
```suggestion
    // Move T's ownership.
    // Undefined behavior when live pointers exist, asserted when KJ_ASSERT_PTR_COUNTERS is defined.
```

c++/src/kj/memory.h
```diff
@@ -656,6 +705,205 @@ class SpaceFor {
...
+// =======================================================================================
+// Ptr<T>
+
+template <typename T>
+class Ptr {
+// Ptr<T> is a smart alternative to T&.
+// 
+// When used together with Pin<T> it keeps track of active pointers.
+// Asserts lifetime constraints when KJ_ASSERT_PTR_COUNTERS is defined.
+// Zero-overhead alternative for T& if KJ_ASSERT_PTR_COUNTERS is not defined.
```
kentonv, line 794:
```suggestion
  // Ptr<T> is a smart alternative to T&.
  // 
  // When used together with Pin<T> it keeps track of active pointers.
  // Asserts lifetime constraints when KJ_ASSERT_PTR_COUNTERS is defined.
  // Zero-overhead alternative for T& if KJ_ASSERT_PTR_COUNTERS is not defined.
```

c++/src/kj/memory.h
```diff
@@ -656,6 +705,205 @@ class SpaceFor {
...
+class Pin {
+// Pin<T> is a smart, in-place storage for T. 
+//
+// Pin<T> should be created on the stack or used as a data member. It should not be 
+// allocated on the heap.
+// Pin<T> is integrated with Ptr<T>, and is legal to move/destroy only when there are no active
+// pointers. 
+// When KJ_ASSERT_PTR_COUNTERS is defined, pointers are tracked and validity of these 
+// operations are asserted.
+// Zero-overhead replacement for T if KJ_ASSERT_PTR_COUNTERS is not defined.
```
kentonv, line 724:
```suggestion
  // Pin<T> is a smart, in-place storage for T. 
  //
  // Pin<T> should be created on the stack or used as a data member. It should not be 
  // allocated on the heap.
  // Pin<T> is integrated with Ptr<T>, and is legal to move/destroy only when there are no active
  // pointers. 
  // When KJ_ASSERT_PTR_COUNTERS is defined, pointers are tracked and validity of these 
  // operations are asserted.
  // Zero-overhead replacement for T if KJ_ASSERT_PTR_COUNTERS is not defined.
```

c++/src/kj/memory.h
```diff
@@ -168,6 +170,53 @@ class NullDisposer: public Disposer {
...
+#ifdef KJ_ASSERT_PTR_COUNTERS
+class AtomicPtrCounter {
+// AtomicPtrCounter uses atomic operations to keep track of active pointers.
+// 
+// Thread safety notes:
+// - every dec() observes the effect of corresponding inc() because of release/consume order
+// - when assertEmpty() is called with balanced dec/inc calls, all releases were consumed, and
+//   additional load with consume memory order will see the correct last value.
+// - when assertEmpty() is called with unbalanced dec/inc calls int will observe the result of 
+//   one of the unbalanced inc(), which is always > 0.
```
kentonv, line 192:
```suggestion
  // AtomicPtrCounter uses atomic operations to keep track of active pointers.
  // 
  // Thread safety notes:
  // - every dec() observes the effect of corresponding inc() because of release/consume order
  // - when assertEmpty() is called with balanced dec/inc calls, all releases were consumed, and
  //   additional load with consume memory order will see the correct last value.
  // - when assertEmpty() is called with unbalanced dec/inc calls int will observe the result of 
  //   one of the unbalanced inc(), which is always > 0.
```

c++/src/kj/memory.h
```diff
@@ -656,6 +705,205 @@ class SpaceFor {
...
+  // Move T's ownership.
+  // Undefined behavior when live pointers exist, asserted when KJ_ASSERT_PTR_COUNTERS is defined.
+#ifdef KJ_ASSERT_PTR_COUNTERS
+    other.ptrCounter.assertEmpty();
+#endif
+  }
+
+  inline ~Pin() {
+  // Destroy a Pin with underlying object. 
+  // Undefined behavior when live pointers exist, asserted when KJ_ASSERT_PTR_COUNTERS is defined.
```
kentonv, line 741:
```suggestion
    // Destroy a Pin with underlying object. 
    // Undefined behavior when live pointers exist, asserted when KJ_ASSERT_PTR_COUNTERS is defined.
```

c++/src/kj/memory.h
```diff
@@ -656,6 +705,205 @@ class SpaceFor {
...
+
+  template <typename U, typename = EnableIf<canConvert<U*, T*>()>>
+  inline bool operator==(const Pin<U>& other) const { return get() == other.get(); }
+
+  template <typename U, typename = EnableIf<canConvert<U*, T*>()>>
+  inline bool operator==(const Ptr<U>& other) const { return get() == other.get(); }
+
+private:
+  inline Ptr(Pin<T>* pin) : ptr(pin->get())
+#ifdef KJ_ASSERT_PTR_COUNTERS
```
kentonv, line 851: Non-blocking nitpick but I kind of think it'd be easier to read if you just had two copies of each constructor definition, one with counters enabled and one without. The partial-line ifdefs are pretty rough.

c++/src/kj/memory.h
```diff
@@ -656,6 +705,205 @@ class SpaceFor {
...
+  inline const T* operator->() const { return get(); }
+
+  inline bool operator==(const Pin<T>& other) const { return get() == other.get(); }
+  inline bool operator==(const Ptr<T>& other) const { return get() == other.get(); }
+  inline bool operator==(const T* const other) const { return get() == other; }
+
+  template <typename U, typename = EnableIf<canConvert<U*, T*>()>>
+  inline bool operator==(const Pin<U>& other) const { return get() == other.get(); }
+
+  template <typename U, typename = EnableIf<canConvert<U*, T*>()>>
```
kentonv, line 846: I think the `EnableIf` is not quite right here, you should be allowed to compare the pointers if converting in either direction is allowed. I think you can just remove the `EnableIf` entirely and rely on the compiler to decide whether the pointers can be compared when the method is instantiated.

c++/src/kj/memory.h
```diff
@@ -656,6 +705,205 @@ class SpaceFor {
...
+    if (ptr == nullptr) {
+      // the value was moved out
+      return;
+    }
+#ifdef KJ_ASSERT_PTR_COUNTERS
+    counter->dec();
+#endif
+  }
+
+  Ptr(Ptr&& other) : ptr(other.ptr)
```
kentonv, line 807: Should there be constructors from `Ptr<U>`, where `U` is a subclass of `T`?

c++/src/kj/memory.h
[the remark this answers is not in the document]
mikea: I've added this:

```
  inline T& asRef() { return *get(); }
  // Obtain a `T&` reference.
  // This is an unsafe operation and should be avoided unless absolutely necessary.
  // It is undefined behavior to use the reference after the object managed by this Ptr<T>
  // ceased to exist.
```

c++/src/kj/memory.h
mikea: this is not needed anymore, you're right (uint64_t)

c++/src/kj/memory.h
mikea: yes, sure, but I wasn't able to find KJ_CRASH or something. Should I use abort() (and print to stderr)?

mikea:
all comments addressed (except the crash, I don't know what's the preferred way). ptal @kentonv

garrettgu10 (approved):
LGTM, I look forward to using this :-)

c++/src/kj/memory.h
kentonv: You need to define some out-of-line function that crashes, then call it.

Probably the function should be `noexcept`, and then it should `KJ_FAIL_REQUIRE`. This produces a much better error report than `abort()`.

I suppose this pattern has come up enough to warrant adding macros for it, but that seems like a separate change...

kentonv (approved):

PR #1970, opened 2024-03-19
fix regenerate-bootstrap.sh to use bazel artifacts

mikea:


kentonv:
Hmm, external contributors sometimes need to run this, and they often aren't using bazel, since it's not the official build system. Can we make the script detect the build system or use a flag or something?

PR #1971, opened 2024-03-19
KJ_ENABLE_UNTYPED_REFCOUNT flag to enable/disable untyped refcount ops

mikea:
Enables kj::Own refcount operations when not defined or defined to 1. Keeps only Rc/Arc operations when defined to 0.

It is easy to use with bazel's local_defines for gradual translation.


mikea:
I can do `KJ_DISABLE_UNTYPED_REFCOUNT` if negative is easier to use.

PR #1975, opened 2024-03-21
kj::Exception get/set/hasDetail<T>

jasnell:
Allows attaching arbitrary, serializable application detail to an exception.

```
struct Detail {
  static constexpr uint64_t EXCEPTION_DETAIL_TYPE_ID = 1;
  static kj::Array<kj::byte> serializeForException(const Detail& detail) {
    return kj::heapArray<kj::byte>(...);
  }
  static Detail deserializeForException(kj::ArrayPtr<kj::byte> data) {
    return {};
  }

  // additional fields
};

auto exception = KJ_EXCEPTION(FAILED, "foo");
exception.setDetail(Detail{...});

KJ_ASSERT(exception.hasDetail<Detail>());

KJ_IF_SOME(detail, exception.getDetail<Detail>()) {
  // use detail
}
```


c++/src/kj/exception.h
```diff
@@ -133,6 +135,48 @@ class Exception {
   KJ_NOINLINE void addTraceHere();
   // Adds the location that called this method to the stack trace.
 
+  template <typename T>
+  void setDetail(const T& detail) {
+    // TODO(reivew): A more sophisticated implementation could avoid the serialization here
+    // for when the exception is being thrown/handled locally without serialization.
+    // It adds a bit of additional complexity, however so we're trading some performance
+    // for simpler code. But... should we?
```
jasnell, line 143: The current implementation is pretty simple but does incur the cost of serializaing/deserializing to/from `T` on every call to `setDetail/getDetail`, even if the exception is created and handled locally without RPC serialization. Per the comment, we could make this more sophisticated at the cost of making the implementation a bit more complicated. I wasn't sure if it was worth the additional effort.

jasnell: The more I dig on use cases, the more I think the more complex implementation is warranted.

Specifically, the underlying detail map should be `kj::OneOf<Detail, kj::Array<byte>>`... where `Detail` is effectively a type erased holder for `T`. If the `kj::Exception` is never serialized, then `getDetail(...)` can just pass the `T&` back to the caller.

On the deserialized `kj::Exception`, the map would initially contain `kj::Array<kj::byte>` until `getDetail(...)` is called. This would cause the value to be deserialized and replaced with `T`, then the `T&` would be returned. 

A `Detail` value shouild be able to opt-out of serialization by simply returning `kj::none` when asked to serialize. This would omit that field from the serialized result.

jasnell:
Had marked this ready but then remember I still had a few outstanding todos... specifically around thread-safety.... so back into draft it goes :-)

c++/src/kj/exception.h
kentonv: > I think the more complex implementation is warranted.

Why is that? Your comment here seems to describe the implementation, but not why it is warranted.

kentonv: I'm not a fan of the new complexified design.

- It appears that when an exception is copied, the copy holds a reference to the same detail map as the original. So if you copy an exception and modify the copy, the original is also modified.
- This design seems to add new allocation and atomic refcounting to all exceptions even when the new feature is not used. I worry this could have a meaningful impact on performance of existing systems.
- I suspect that the memoized parsing isn't as useful as it appears. I suspect in most use cases, the "details" of an exception will be consumed only once anyway. So if it's done every time you call the getter, that's fine, since it'll only be called once.
- I am very skeptical of non-serializable details. I don't like the idea that inserting an RPC boundary could inadvertently break things because the exceptions being thrown have non-serializable details.
- Moreover, exceptions have always been a "value type", I really don't want to see people attaching stuff to exceptions that are not plain data, especially as lifetimes will be difficult to reason about.
    - In particular, I do not think we can realistically tunnel a JavaScript error "by identity" through an Exception. We can't attach a V8 handle to an exception because we have no idea if the exception will be accessed under isolate lock, or even outlive the isolate itself, etc.
- As we discussed, I think a HashMap is overkill as it's highly unlikely that an exception will have more than one or two details.

PR #1978, opened 2024-03-27
Add exception detail API.

kentonv:
Compared to #1975, this version maintains a very simple map of type IDs to bytes.

I actually decided that the fancy templating and expectation of serialize/deserialize methods doesn't work. In our specific case, serialization and deserialization wolud require a V8 isolate lock to operate, and there may be use cases where we'll want to detach the bytes without deserializing them. Better to let the caller get access to the raw bytes and use them as they see fit.

### DIFF
--- a/c++/src/kj/exception.h
+++ b/c++/src/kj/exception.h
@@ -24,6 +24,7 @@
 #include "memory.h"
 #include "array.h"
 #include "string.h"
+#include "vector.h"
 #include "windows-sanity.h"  // work-around macro conflict with `ERROR`
 
 KJ_BEGIN_HEADER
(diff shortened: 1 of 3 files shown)


c++/src/kj/exception.h
```diff
@@ -133,6 +134,25 @@ class Exception {
   KJ_NOINLINE void addTraceHere();
   // Adds the location that called this method to the stack trace.
 
+  using DetailTypeId = unsigned long long;
```
jclee, line 137: Looking at other kj headers, I guess `unsigned long long` is preferred to `uint_64t`, maybe because it doesn't need to pull in additional headers and is probably equivalent on any existing platform.

jclee (approved):
I assume that since the identity and content of details are application-specific, there's nothing particularly useful we could add to an exception's stringification to denote the presence of details.  And similarly, that we could later add a way for applications to customize the Exception stringifier, if needed.

jclee:
Looks like the msvc build is unhappy about something, but maybe that's currently expected?  Doesn't seem to be specific to this commit.

  - unable to compile coroutine test?:
    https://github.com/capnproto/capnproto/actions/runs/8455149621/job/23161931057#step:5:659
  - unable to find "heavy" tests exe:
    https://github.com/capnproto/capnproto/actions/runs/8455149621/job/23161931057#step:5:1028

c++/src/kj/exception.h
kentonv: Yeah, I didn't want to pull in stdint.h here, given this header is very widely included.

c++/src/kj/exception.h
```diff
@@ -133,6 +134,25 @@ class Exception {
...
+  //
+  // Every type of detail must have a unique ID, which is a 64-bit integer. It's suggested that
+  // you use `capnp id` to generate these.
+  //
+  // It is expected that exceptions will rarely have more than one or two details, so the
+  // implementation uses a flat array with O(n) lookup.
+  //
+  // The main use case for details is to be able to tunnel exceptions of a different type through
+  // KJ / Cap'n Proto. In particular, Cloudflare Workers commonly has to convert a JavaScript
+  // exception to KJ and back. The exception is serialized using V8 serialization.
```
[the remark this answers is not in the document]
kentonv: Right. For the moment, I didn't need this, and I kind of like that applications can only query for the types they actually know about, rather than trying to explore unrecognized types.

That said, I suppose this may have to change when we extend this to transit RPC correctly. At present the details will be lost when crossing RPC -- but for the present use case, that doesn't matter.

kentonv:
The MSVC build has been broken for a while. I haven't gotten around to looking into it.

PR #1987, opened 2024-04-12
Use kj::none in the poll(2) implementation of UnixEventPort.

balusch:
No functional change.


kentonv:
Thanks! I guess we failed to pay attention to deprecation warnings in the mac build...

balusch:
:) That's probably because the `poll`(2) implementation is not enabled by default on most modern platforms, where `epoll`(2) or `kqueue`(2) is available and thus preferred.

kentonv:
@balusch Oh, I misread since your other PR was addressing kqueue, I thought this one was also addressing kqueue (which is used on mac). Indeed, the poll implementation is used almost nowhere at this point.

PR #1990, opened 2024-04-16
make Membrane refcounted

mikea:
All real membrane implementations already are. Explicit refcounting makes reasoning about lifetime much easier.

You could argue that this style makes memory leaks more possible, but arguably this is the right tradeoff to make instead of lifetime issue.

I'm seeking for initial feedback (yay/nay) before I spend too much time cleaning this up.

Corresponding downstream pull requests are:

https://github.com/cloudflare/workerd/pull/2029 and EW/7744

WDYT @kentonv  ?

### DIFF
--- a/c++/src/capnp/membrane.c++
+++ b/c++/src/capnp/membrane.c++
@@ -75,8 +75,8 @@ private:
 
 class MembraneCapTableBuilder final: public _::CapTableBuilder {
 public:
-  MembraneCapTableBuilder(kj::Rc<MembranePolicy> policy, bool reverse)
-      : policy(kj::mv(policy)), reverse(reverse) {}
+  MembraneCapTableBuilder(kj::Rc<MembranePolicy>& policy, bool reverse)
+      : policy(policy), reverse(reverse) {}
 
   AnyPointer::Builder imbue(AnyPointer::Builder builder) {
     KJ_REQUIRE(inner == nullptr, "can only call this once");
@@ -111,7 +111,7 @@ public:
 
 private:
   _::CapTableBuilder* inner = nullptr;
-  kj::Rc<MembranePolicy> policy;
+  kj::Rc<MembranePolicy>& policy;
   bool reverse;
 };
 
@@ -158,7 +158,7 @@ public:
   MembraneRequestHook(kj::Own<RequestHook>&& inner, kj::Rc<MembranePolicy> policy, bool reverse)
       : RequestHook(&MEMBRANE_BRAND),
         inner(kj::mv(inner)), policy(kj::mv(policy)),
-        reverse(reverse), capTable(this->policy.addRef(), reverse) {}
+        reverse(reverse), capTable(this->policy, reverse) {}
 
   static Request<AnyPointer, AnyPointer> wrap(
       Request<AnyPointer, AnyPointer>&& inner, kj::Rc<MembranePolicy> policy, bool reverse) {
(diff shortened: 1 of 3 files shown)


kentonv:
I guess as written this will increase the number of refcount operations a bunch since a lot of places where we previously passed by reference, now we're passing by refcount. Maybe it's not a big deal in practice since they are non-atomic refcounts.

Most users of MembranePolicy don't ever see the refcounting, so I'm not sure converting this class will have much impact on applications. (Applications typically allocate a MembranePolicy and just immediately pass it to capnp::membrane(); they don't normally keep or pass references anywhere else.)

These are minor issues though. Feel free to change this to non-draft status if you want to proceed.

c++/src/capnp/membrane.c++
```diff
@@ -111,62 +111,62 @@ public:
...
-      : inner(kj::mv(inner)), policy(kj::mv(policy)), capTable(*this->policy, reverse) {}
+      kj::Own<ResponseHook>&& inner, kj::Rc<MembranePolicy> policy, bool reverse)
+      : inner(kj::mv(inner)), policy(kj::mv(policy)), capTable(this->policy.addRef(), reverse) {}
 
   AnyPointer::Reader imbue(AnyPointer::Reader reader) { return capTable.imbue(reader); }
 
 private:
   kj::Own<ResponseHook> inner;
-  kj::Own<MembranePolicy> policy;
+  kj::Rc<MembranePolicy> policy;
```
kentonv, line 152: This member is no longer needed, since you're passing a strong ref directly into `capTable`, and this member isn't otherwise used.

c++/src/capnp/membrane.c++
```diff
@@ -234,23 +234,23 @@ public:
...
   AnyPointer::Pipeline sendForPipeline() override {
     return AnyPointer::Pipeline(kj::refcounted<MembranePipelineHook>(
-        PipelineHook::from(inner->sendForPipeline()), policy->addRef(), reverse));
+        PipelineHook::from(inner->sendForPipeline()), policy.addRef(), reverse));
   }
 
 private:
   kj::Own<RequestHook> inner;
-  kj::Own<MembranePolicy> policy;
+  kj::Rc<MembranePolicy> policy;
```
kentonv, line 242: ugh... It bugs me that we increment the refcount twice when MembraneRequestHook is created and decrement it twice on destruction, once here and once for the `capTable` member below.

c++/src/capnp/membrane.c++
```diff
@@ -311,7 +311,7 @@ public:
 
 private:
   kj::Own<CallContextHook> inner;
-  kj::Own<MembranePolicy> policy;
+  kj::Rc<MembranePolicy> policy;
```
kentonv, line 314: In this case we increment the refcount THREE times for every MembraneCallContext, only to be decremented all together in the destructor.

kentonv:
Can the redundant references be kj::Ptr intsead?

mikea:
> Can the redundant references be kj::Ptr intsead?

not really, e.g. MembraneCapTableReader does addRef, MembraneCapTableBuilder as well

c++/src/capnp/membrane.c++
mikea: Can confirm, here's ghidra decompile in opt mode: https://gist.github.com/mikea/40d74e2e849e9a14b83209a364f90fc7

I can change `MembraneCapTableReader` to accept `Rc<MembranePolicy>&` instead, wdyt?

But also I want to see if I can make `kj::Rc` have Own with static disposer to inline all those virtual calls.

mikea:
I have optimized extra addRef out (2 additional commits). Here's the destructor now: https://gist.github.com/mikea/bfa1139ac56468becc80fa7830930f92

@kentonv

PR #1996, opened 2024-04-22
kj::memzero<T>() - zero-initialize memory region

mikea:
This PR introduces `kj::memzero<T>()` utility to 0-initialize memory. It is intended to be used as a type-safe alternative to some usages of `memset` (other additions are coming).

The provided `kj::memzero<T>(T&)` version works for normal storage as well as for arrays.

An alternative design is to define `kj::memzero<T>(T*)` but that wouldn't let us provide array version and is
less type-safe (because you can pass array to it).

### DIFF
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -2158,6 +2158,12 @@ constexpr bool isDisallowedInCoroutine() {
 // }
 //
 
+template<typename T, typename = EnableIf<KJ_HAS_TRIVIAL_CONSTRUCTOR(T)>>
+inline void memzero(T& t) {
+  // Zero-initialize memory region belonging to t. Type-safe wrapper around `memset`.
+  memset(&t, 0, sizeof(T));
+}
+
 }  // namespace kj
 
 constexpr kj::ArrayPtr<const kj::byte> operator "" _kjb(const char* str, size_t n) {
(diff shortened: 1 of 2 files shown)


mikea:
I also wonder if I should make `kj::zero` work with `ArrayPtr<T>` too?

jasnell (approved):

c++/src/kj/common.h
```diff
@@ -2141,6 +2141,12 @@ constexpr bool isDisallowedInCoroutine() {
 // }
 //
 
+template<typename T>
+inline void zero(T& t) {
```
kentonv, line 2145: Should we limit this so you can only call it on types with trivial constructors and destructors? Otherwise it is ironically unsafe, but not as obviously so as memset. E.g. I could imagine someone trying to zero a `kj::Array`, thinking that it'll zero the contents of the array.

Could this be, like, `memzero` instead of `zero`? Might be clearer what it does that way.

kentonv:
> I also wonder if I should make kj::zero work with ArrayPtr<T> too?

I was thinking that but I think it'd be better to have a method on `ArrayPtr`.

c++/src/kj/common.h
mikea: added `KJ_HAS_TRIVIAL_CONSTRUCTOR`. Also it is easy to add `kj:zero` override for Array which I did. Should I remove it and rename to `memzero`?

mikea: renamed to memzero and removed Array version. PTAL.

kentonv (approved):

kentonv:
Oops clicked merge myself out of habit.

PR #1997, opened 2024-04-22
kj::ArrrayPtr<T>::fill()

mikea:
This is another memset alternative: `kj::ArrayPtr::fill()` function.

The implementation uses memset for small T and normal for-loop otherwise which should be easy to vectorize.

### DIFF
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -1935,6 +1935,14 @@ public:
   // Syntax sugar for invoking U::from.
   // Used to chain conversion calls rather than wrap with function.
 
+  inline void fill(T t) { 
+    // Fill the area by copying t over every element.
+
+    for (size_t i = 0; i < size_; i++) { ptr[i] = t; } 
+    // All modern compilers are smart enough to optimize this loop for simple T's.
+    // libc++ std::fill doesn't have memset specialization either.
+  }
+
 private:
   T* ptr;
   size_t size_;
--- a/c++/src/kj/common-test.c++
+++ b/c++/src/kj/common-test.c++
@@ -926,6 +926,33 @@ KJ_TEST("kj::ArrayPtr startsWith / endsWith / findFirst / findLast") {
   KJ_EXPECT(arr.findLast(78).orDefault(100) == 100);
 }
 
+KJ_TEST("kj::ArrayPtr fill") {
+  int64_t int64Array[] = {12, 34, 56, 34, 12};
+  arrayPtr(int64Array).fill(42);
+  for (auto i: int64Array) {
+    KJ_EXPECT(i == 42);
+  }
+
+  // test small sizes separately, since compilers do a memset optimization
+  byte byteArray[256];
+  arrayPtr(byteArray).fill(42);
+  for (auto b: byteArray) {
+    KJ_EXPECT(b == 42);
+  }
+
+  // test an object
+  struct SomeObject {
+    int64_t i;
+    double d;
+  };
+  SomeObject objs[256];
+  arrayPtr(objs).fill(SomeObject{42, 3.1415926});
+  for (auto& o: objs) {
+    KJ_EXPECT(o.i == 42);
+    KJ_EXPECT(o.d == 3.1415926);
+  }
+}
+
 struct Std {
   template<typename T>
   static std::span<T> from(ArrayPtr<T>* arr) {


c++/src/kj/common.h
```diff
@@ -1932,6 +1932,17 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
...
   // Used to chain conversion calls rather than wrap with function.
 
+  inline void fill(T t) { 
+    // Fill the area by copying t over every element.
+
+    if constexpr (sizeof(T) == 1) {
+      // use memset for small T types, needs to be convertible to kj::byte
+      memset(begin(), (byte)t, size());
+    } else {
+      for (size_t i = 0; i < size_; i++) { ptr[i] = t; }
```
jasnell, line 1942: super minor nit: `std::fill(...)` would work here too, right? Any reason not to use it?

mikea: @kentonv is usually against using `std::`, happy to change if I'm wrong.

c++/src/kj/common.h
```diff
@@ -1932,6 +1932,17 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
   // Syntax sugar for invoking U::from.
   // Used to chain conversion calls rather than wrap with function.
 
+  inline void fill(T t) {
```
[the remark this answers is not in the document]
mikea: > Could this perhaps take an optional offset position to begin filling from?

why not `arr.slice(offset).fill()`?

jasnell (approved):
LGTM with a question and a nit.

c++/src/kj/common.h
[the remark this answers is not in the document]
jasnell: That works too, I just tend to prefer the more concise `arr.fill(0, offset)` ... but either is fine.

c++/src/kj/common.h
```diff
@@ -1932,6 +1932,17 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
   // Syntax sugar for invoking U::from.
   // Used to chain conversion calls rather than wrap with function.
 
+  inline void fill(T t) { 
+    // Fill the area by copying t over every element.
+
+    if constexpr (sizeof(T) == 1) {
```
kentonv, line 1938: This is not necessarily safe, since someone could have a non-trivial class of one-byte size.

IMO we should change this method to strictly zero instead of fill. And then optimize as follows:
- For types that are trivially constructable/destructable, use memset.
- For other types, use a loop assigning to zero.

c++/src/kj/common.h
mikea: > How often do we actually fill an array with anything other that zeros?

There are quite a few usages in tests (which deserve this method too imo?). Outside of tests there's usage with ' ' in stringify.c++.

c++/src/kj/common.h
mikea: The other approach is to use `std::fill` under the hood. It should include this optimization?

mikea:
I find this in libc++:

```
// fill isn't specialized for std::memset, because the compiler already optimizes the loop to a call to std::memset.
```

https://github.com/llvm/llvm-project/blob/main/libcxx/include/__algorithm/fill.h#L22

I suggest we drop all smartness and keep the loop? WDYT @kentonv @jasnell 

PS here's the example that verifies this: https://godbolt.org/z/azErf1ME8 both GCC and Clang generate identical code for `memset` and `for` loop.

jasnell:
Still LGTM with a nit about expanding tests a bit more

kentonv:
Is there any loss of optimization due to the fill value being a variable? If we made this a template instead, i.e. `fill<0>()` instead of `fill(0)`, does this end up helping the compiler at all? (Presumably they are equivalent if fill() is inlined, but is it always inlined?)

mikea:
> Is there any loss of optimization due to the fill value being a variable? If we made this a template instead, i.e. `fill<0>()` instead of `fill(0)`, does this end up helping the compiler at all? (Presumably they are equivalent if fill() is inlined, but is it always inlined?)

I don't really see any missing optimization: https://godbolt.org/z/YE5fj3Ye3

kentonv:
> I don't really see any missing optimization: https://godbolt.org/z/YE5fj3Ye3

This seems convincing.

I notice that explicitly declaring `fill()` as `inline` matters in this example -- but you have it as `inline` in your PR, so great.

kentonv (approved):

PR #1998, opened 2024-04-23
Add Array::slice(start) methods

justin-mp:
We have the single-parameter slice methods in ArrayPtr and they’re
really convenient.  While it’s always possible to use Array::asPtr()
to convert an Array to an ArrayPtr, that’s extra noise when we could
add these methods on Array directly.

### DIFF
--- a/c++/src/kj/array.h
+++ b/c++/src/kj/array.h
@@ -189,6 +189,14 @@ public:
     KJ_IREQUIRE(start <= end && end <= size_, "Out-of-bounds Array::slice().");
     return ArrayPtr<const T>(ptr + start, end - start);
   }
+  inline ArrayPtr<T> slice(size_t start) KJ_LIFETIMEBOUND {
+    KJ_IREQUIRE(start <= size_, "Out-of-bounds ArrayPtr::slice().");
+    return ArrayPtr<T>(ptr + start, size_ - start);
+  }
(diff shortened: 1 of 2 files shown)


c++/src/kj/array.h
```diff
@@ -189,6 +189,14 @@ class Array {
     KJ_IREQUIRE(start <= end && end <= size_, "Out-of-bounds Array::slice().");
     return ArrayPtr<const T>(ptr + start, end - start);
   }
+  inline ArrayPtr<T> slice(size_t start) {
```
mikea, line 192: you'll need to add KJ_LIFETIMEBOUND like above.

justin-mp: Fixed in https://github.com/capnproto/capnproto/compare/d4db83d435411a7e34262e1e21c180f8b386e969..e8bb906f6c2b58491f62d3ef439141cdb91a2e3d.

mikea:
see also https://github.com/capnproto/capnproto/pull/2000 @justin-mp that helps me a lot.

mikea (approved):

jasnell (approved):

kentonv (approved):

kentonv:
Ah yes... `KJ_IREQUIRE` only throws in debug builds.

justin-mp:
I added #2008 to run the throw tests in debug mode only.

PR #1999, opened 2024-04-24
""_kjb byte array pointer literal 

mikea:
It is very common to create literal byte sequences (even more so after IO api change), _kjb utility
removes some noise.

### DIFF
--- a/c++/src/capnp/compat/http-over-capnp-test.c++
+++ b/c++/src/capnp/compat/http-over-capnp-test.c++
@@ -528,7 +528,7 @@ void runWebSocketBasicTestCase(
   }
 
   {
-    auto promise = serverWs.send("bar"_kj.asBytes());
+    auto promise = serverWs.send("bar"_kjb);
     auto message = clientWs.receive().wait(waitScope);
     promise.wait(waitScope);
     KJ_ASSERT(message.is<kj::Array<kj::byte>>());
--- a/c++/src/capnp/compat/json-test.c++
+++ b/c++/src/capnp/compat/json-test.c++
@@ -981,8 +981,8 @@ KJ_TEST("rename fields") {
 
     root.setCustomFieldHandler("waldo");
 
-    root.setTestBase64("fred"_kj.asBytes());
-    root.setTestHex("plugh"_kj.asBytes());
+    root.setTestBase64("fred"_kjb);
+    root.setTestHex("plugh"_kjb);
 
     root.getBUnion().setBar(678);
 
--- a/c++/src/kj/async-io-test.c++
+++ b/c++/src/kj/async-io-test.c++
@@ -616,8 +616,8 @@ TEST(AsyncIo, CapabilityPipeMultiStreamMessage) {
   streams.add(kj::mv(pipe2.ends[0]));
   streams.add(kj::mv(pipe3.ends[0]));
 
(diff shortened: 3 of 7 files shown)


jasnell (commented):
hmm, `kjb` might be an unfortunately overloaded acronym to use ;-) ...

I'd actually just prefer `_bytes`, but don't feel strongly about it.

jasnell (approved):

kentonv:
You can do `""_kj.asBytes()`. Do we really need an abbreviation?

mikea:
> You can do `""_kj.asBytes()`. Do we really need an abbreviation?

with slice-based API I have to convert even more `write("abc", 3)` to `write("abc"_kj.asBytes())` which gets really tiresome and annoying. `"abc"_kjb` is so much shorter.

mikea:
> hmm, `kjb` might be an unfortunately overloaded acronym to use ;-) ... (too close to kgb)

can't say this has escaped my mind  by I decided to be consistent :)

> I'd actually just prefer `_bytes`, but don't feel strongly about it.

My personal vote is simple `_b`. Why `_kj`?

kentonv:
> Why `_kj`?

Unfortunately, these suffixes aren't namespaced in any way. Some other toolkit library could also decide to define `_bytes` and then we'd have compile errors.

mikea:
> Some other toolkit library could also decide to define `_bytes` and then we'd have compile errors.

Do you mean "_somebody_" will have a compile error? Not sure how that will result in us having compile errors?

Also presumably this will happen only if include _both_ that toolking and kj in the same file. Which, I am sure, can be made carefully? Imo this is very theoretical, but readability benefits of `_bytes` or even `_b` or huge.

kentonv:
By "we" here I meant the Workers Runtime could end up with compile errors if we happen to include some other library that also tries to define these suffixes. For instance, we end up with a dependency on Abseil, which is also a C++ toolkit library that has just as much right to define `_bytes` as KJ does. This is why all our other literal suffixes start with `_kj`.

mikea:
> By "we" here I meant the Workers Runtime could end up with compile errors if we happen to include some other library that also tries to define these suffixes. For instance, we end up with a dependency on Abseil, which is also a C++ toolkit library that has just as much right to define `_bytes` as KJ does. This is why all our other literal suffixes start with `_kj`.

I agree, but in reality our list of 3rd party dependencies is very small and we're very careful to keep it that way. If we're talking absl in particular, then we include it only once and tcmalloc once or twice. So I'm sure we'll be able to overcome them definining anything like that?

To make things easier in case it does happen, I can wrap `_b` in  `#ifndef KJ_NO_B_STR_SUFFIX`? `KJ_NAMESPACE_STR_SUFFIXES` that will trigger between `_b` and `_kjb` in case we'll need to coexist in that one file?

kentonv:
I'd prefer to stick with `_kjb` or even just leave things as `_kj.asBytes()`.

* I don't think it's reasonable to pollute the global scope this way even if we doubt it will affect ourselves. This is still code we intend for third parties to use eventually (when we make a 2.0 release) and this could cause frustrating errors for someone.
* More practically for us: If a new engineer just sees `"foo"_b` somewhere, and they aren't already familiar with these suffixes, where do they even start looking to figure out what this means?
* While these may be relatively weak arguments, saving two keystrokes, or even 11, also seems like a weak argument.

kentonv (approved):

PR #2000, opened 2024-04-24
first(size_t count) sequences helper method

mikea:
I found I have to use it a lot while writing IO slice-based code.

Doing `.slice(0, count)` is very common. `std::span` has `first` method, why shouldn't we?

`Path` and `PathPtr` is the only place I didn't add `.first` method. Let me know if you want it there too.

### DIFF
--- a/c++/src/kj/string.h
+++ b/c++/src/kj/string.h
@@ -145,6 +145,9 @@ public:
   // A string slice is only NUL-terminated if it is a suffix, so slice() has a one-parameter
   // version that assumes end = size().
 
+  inline ArrayPtr<const char> first(size_t count) const;
+  // Return string prefix.
+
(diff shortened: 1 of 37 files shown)


c++/src/kj/string.h
```diff
@@ -734,6 +740,9 @@ inline StringPtr StringPtr::slice(size_t start) const {
 inline ArrayPtr<const char> StringPtr::slice(size_t start, size_t end) const {
   return content.slice(start, end);
 }
+inline ArrayPtr<const char> StringPtr::first(size_t count) const {
+  return content.slice(0, count);
```
justin-mp: This could be `content.first(count);`, right?

mikea: I actually prefer it to be `this.slice(0, count)` everywhere. fixed this place too.

PR #2007, opened 2024-04-25
arrayPtr() utility for fixed-sized arrays

mikea:
This is very useful when working with local buffers.

### DIFF
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -1995,6 +1995,12 @@ inline constexpr ArrayPtr<T> arrayPtr(T* begin KJ_LIFETIMEBOUND, T* end KJ_LIFET
   return ArrayPtr<T>(begin, end);
 }
 
+template <typename T, size_t Size>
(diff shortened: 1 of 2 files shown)


mikea:
extracted this from 1997 since I reach out for it quite a lot.

jasnell (approved):

c++/src/kj/common.h
```diff
@@ -1995,6 +1995,12 @@ inline constexpr ArrayPtr<T> arrayPtr(T* begin KJ_LIFETIMEBOUND, T* end KJ_LIFET
   return ArrayPtr<T>(begin, end);
 }
 
+template <typename T, size_t Size>
```
kentonv, line 1998: Not worth a PR to fix, but FWIW, I've always gone by the rule that type names start with a capital letter, values start with a lower-case letter -- including for template parameters. So `Size` here should actually be `size`.

mikea: no worries: https://github.com/capnproto/capnproto/pull/2010

PR #2009, opened 2024-04-26
Initialize TestExceptionCallback in new threads

justin-mp:
In a test that I wrote using threads, I noticed my test program still succeeded (by exiting with a 0 return code) when the program logged an ERROR-severity log message in the non-main thread.  This happens because TestExceptionCallback does not implement getThreadInitializer, so we get a default RootExceptionCallback.  The RootExceptionCallback will write errors to stderr, but doesn’t inform the ProcessContext that an error has occured and the process should exit with a non-zero return code.

The way to fix this is to implement getThreadInitializer in TestExceptionCallback.  My implementation is simple: in a new thread, I construct a new TestExceptionCallback and then run the thread’s main function.

Tehnically, such a construction is not thread-safe because the ProcessContext we use isn’t const.  Tests use the
TopLevelProcessContext whose error method is:

```
void TopLevelProcessContext::error(StringPtr message) {
  hadErrors = true;
  writeLineToFd(STDERR_FILENO, message);
}
```

`writeLineToFd` looks to be thread-safe, but I’m less sure that writing a boolean is safe to do on all architectures.  (It should be fine on x86_64.)

### DIFF
--- a/c++/src/kj/main.c++
+++ b/c++/src/kj/main.c++
@@ -177,11 +177,11 @@ static void writeLineToFd(int fd, StringPtr message) {
 #endif
 }
 
-void TopLevelProcessContext::warning(StringPtr message) {
+void TopLevelProcessContext::warning(StringPtr message) const {
   writeLineToFd(STDERR_FILENO, message);
 }
 
-void TopLevelProcessContext::error(StringPtr message) {
+void TopLevelProcessContext::error(StringPtr message) const {
   hadErrors.store(true);
   writeLineToFd(STDERR_FILENO, message);
 }
--- a/c++/src/kj/main.c++
+++ b/c++/src/kj/main.c++
@@ -59,7 +59,7 @@ StringPtr TopLevelProcessContext::getProgramName() {
 }
 
 void TopLevelProcessContext::exit() {
-  int exitCode = hadErrors ? 1 : 0;
+  int exitCode = hadErrors.load() ? 1 : 0;
   if (cleanShutdown) {
     throw CleanShutdownException { exitCode };
   }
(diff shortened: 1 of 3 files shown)


justin-mp:
I'd like help on the best way to make this thread-safe.

I could update TopLevelProcessContext::hadErrors to an std::atomic_bool to make sure that's fine.  But it doesn't change the fact that by KJ's coding convention, I can't call `ProcessContext::error` from a thread.  To fix this problem, I could make const versions of the ProcessContext methods and then implement them in TopLevelProcessContext.

justin-mp:
https://github.com/capnproto/capnproto/compare/e5127f7fe0d5598947ee11bcd9a10f52c9e79d0b..d8750283613fa699ed59f76e9370b497d28379a2 is a rebase on v2.

justin-mp:
https://github.com/capnproto/capnproto/compare/a0fc807b9dba8ab70f56591da09bc696316398b8..46bb1d1bc8b73ab147a3ba564fcf8a24e046d72a and https://github.com/capnproto/capnproto/compare/46bb1d1bc8b73ab147a3ba564fcf8a24e046d72a..5da277db4185d372cb6e8d20e6e78d1016bdbdff rename topLevelCallback to mainThreadCallback, which describes our my actual intent.

kentonv:
Looks like we've uncovered a flake in mutex-test? :grimacing: 

I suppose we'll need to make sure this passes all tests in the workers runtime too, before merging.

justin-mp:
> Looks like we've uncovered a flake in mutex-test? 😬

See some fixes for the test in #2016.

> I suppose we'll need to make sure this passes all tests in the workers runtime too, before merging.

See 7816 internally.

kentonv (approved):

PR #2011, opened 2024-04-26
Use std::atomic_flag to set executorDone

justin-mp:
On arm64, I noticed that this test spews the following:

```
[ TEST ] async-unix-test.c++:1248: m:n threads:EventLoops
kj/async-unix-test.c++:1283: error: failed: expected executorDone
[ PASS ] async-unix-test.c++:1248: m:n threads:EventLoops
```

.  On x86_64, everything is fine, so it could be due to different memory semantics on arm64.  Given that we’re working across threads, we might as well use atomics.

Note that the default constructor of std::atomic_flag and std::atomic_flag::test_and_set() are new in C++20, but I think that’s the C++ version Cap’n Proto uses nowadays.

### DIFF
--- a/c++/src/kj/async-unix-test.c++
+++ b/c++/src/kj/async-unix-test.c++
@@ -1272,7 +1272,7 @@ KJ_TEST("m:n threads:EventLoops") {
   static thread_local uint threadId = 0;
 
   threadId = 1;
-  bool executorDone = false;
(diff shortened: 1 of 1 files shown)


justin-mp:
I still get the `error: failed: expected executorDone` on my ARM target, so this didn't work.

PR #2014, opened 2024-04-29
Fix Maybe nullptr assignment deprecation notice loophole

harrishancock:
While reading some code in async.c++, I noticed a place where we nullify a Maybe with a nullptr, which surprised me, because I thought we had deprecated all such code. It turns out that while our default Maybe<T> implementation bans pointer assignment,  our Maybe<T&> specialization does not. This commit adds a deprecated nullptr assignment operator overload to Maybe<T&> to flush out any last remaining nullptr nullifications.

Note that this doesn't fix any bug in our code or anything, it just makes the nullification call sites a bit clearer.

### DIFF
--- a/c++/src/capnp/capability.c++
+++ b/c++/src/capnp/capability.c++
@@ -558,7 +558,7 @@ public:
 
   ~LocalClient() noexcept(false) {
     KJ_IF_SOME(s, server) {
-      s->thisHook = nullptr;
+      s->thisHook = kj::none;
     }
   }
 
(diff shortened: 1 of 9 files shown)


kentonv:
Looks like you missed one or more spots.

kentonv (approved):
Re-ran the MacOS build which failed but didn't look related, if it fails again we'll have to investigate.

MSVC build has been broken at master for a while so I guess you can ignore that one... :/

PR #2015, opened 2024-04-29
new sync IO api

mikea:


c++/src/kj/io.h
```diff
@@ -78,13 +78,22 @@ class OutputStream {
...
   virtual ~OutputStream() noexcept(false);
 
-  virtual void write(const void* buffer, size_t size) = 0;
-  // Always writes the full size.  Throws exception on error.
-
-  virtual void write(ArrayPtr<const ArrayPtr<const byte>> pieces);
-  // Equivalent to write()ing each byte array in sequence, which is what the default implementation
-  // does.  Override if you can do something better, e.g. use writev() to do the write in a single
-  // syscall.
+  virtual void write(ArrayPtr<const byte> data,
```
kentonv, line 81: I think this signature for write() didn't work out well after all. Sorry for sending you down the wrong path.

In many places, you've ended up implementing this in terms of a `writeImpl` that writes a single buffer. Meanwhile, most callers only pass a single buffer. So it actually turns out to just be a lot of wasted checks. Also in a lot of these cases you've declared `write()` inline in the header file, but this is inappropriate since a virtual call cannot be inlined, and the `forEachPiece()` invocation is too complex to online anyway, so these method bodies need to be moved to the source files to avoid bloating build times.

I think we should go back to having two methods, one that takes `ArrayPtr<const byte>` and one that takes `ArrayPtr<const ArrayPtr<const byte>>`, and give the latter a default implementation that loops and invokes the former.

c++/src/kj/io.h
```diff
@@ -55,10 +55,10 @@ class InputStream {
   // If the InputStream can't produce minBytes, it MUST throw an exception, as the caller is not
   // expected to understand how to deal with partial reads.
 
-  virtual size_t tryRead(void* buffer, size_t minBytes, size_t maxBytes) = 0;
+  virtual size_t tryRead(ArrayPtr<byte> buffer, size_t minBytes) = 0;
   // Like read(), but may return fewer than minBytes on EOF.
 
-  inline void read(void* buffer, size_t bytes) { read(buffer, bytes, bytes); }
+  inline void read(ArrayPtr<byte> buffer) { read(buffer, buffer.size()); }
```
jasnell, line 61: Since this is just falling back to `read(buffer, minBytes)` I'm really not sure it's all that useful? We could just specify a default value for `minBytes` that is set to max(size_t)` with the interpretation that any `minBytes` larger than `buffer` can be handled as `minBytes = buffer.size()`

c++/src/kj/io.h
```diff
@@ -38,9 +38,9 @@ class InputStream {
...
   virtual ~InputStream() noexcept(false);
 
-  size_t read(void* buffer, size_t minBytes, size_t maxBytes);
-  // Reads at least minBytes and at most maxBytes, copying them into the given buffer.  Returns
-  // the size read.  Throws an exception on errors.  Implemented in terms of tryRead().
+  size_t read(ArrayPtr<byte> buffer, size_t minBytes);
+  // Reads at least minBytes and at most maxBytes = buffer.size(), copying them into the given buffer.
+  // Returns the size read.  Throws an exception on errors.  Implemented in terms of tryRead().
   //
   // maxBytes is the number of bytes the caller really wants, but minBytes is the minimum amount
```
jasnell, line 45: nit: comment here mentioning `maxBytes` needs to be updated.

c++/src/kj/io.h
kentonv: I think that would impose additional complexity on all implementations just to eliminate this one inline method. Doesn't seem worth it.

kentonv:
Looks like the MSVC tests are failing. (#2024 fixed them at master, so presumably it's a new breakage.)

Otherwise looks good.

jasnell:
LGTM once CI is green. I'll leave it to someone else to actually hit approve tho.

mikea:
> Looks like the MSVC tests are failing. (#2024 fixed them at master, so presumably it's a new breakage.)

Windows build is fixed. I've opened downstream PRs that I will update now to the latest hash:

- https://github.com/cloudflare/workerd/pull/2107
- EW 7902

kentonv (approved):

mikea:
just a rebase

PR #2016, opened 2024-04-29
De-flake and clean up mutex-test

justin-mp:


c++/src/kj/mutex-test.c++
```diff
@@ -73,15 +73,19 @@ TEST(Mutex, MutexGuarded) {
 #if KJ_USE_FUTEX
     auto timeout = MILLISECONDS * 50;
 
-    auto startTime = systemPreciseMonotonicClock().now();
-    EXPECT_TRUE(value.lockExclusiveWithTimeout(timeout) == kj::none);
-    auto duration = startTime - systemPreciseMonotonicClock().now();
-    EXPECT_TRUE(duration < timeout);
```
kentonv, line 79: I think the intent here was actually to verify that we waited for at least the expected amount of time before giving up on the lock. So, with a correctly-calculated `duration`, we actually want this to be:

```
EXPECT_TRUE(duration >= timeout);
```

This expectation should be safe against flakes, since if the test runs overly slowly, it will only make `duration` increase further.

kentonv (approved):
Somehow I suspect this flake will not end here but I guess we'll see.

c++/src/kj/mutex-test.c++
kentonv: (That is to say, I think we should bring the expectation back, corrected.)

justin-mp: Fixed in https://github.com/capnproto/capnproto/compare/e54256c7aa760d638f43ac612fde44027b5dfea2..3f037ce9b166f180a789bbc8cfda52a1ff268408

justin-mp:
https://github.com/capnproto/capnproto/compare/3f037ce9b166f180a789bbc8cfda52a1ff268408..c4d2bc48db846817beba794189d7b621cbbd0f76 is a rebase on `v2`.

PR #2017, opened 2024-04-30
Attempt to fix an obscure tracePromise() UaF

harrishancock:
I've observed null pointer dereferences with stack traces showing CoroutineBase::tracePromise() calling into ChainPromiseNode::tracePromise(). The only explanation I could think of is that perhaps the coroutine's tracePromise() function is being called after a currently-awaited ChainPromiseNode has transitioned to STEP2, which could invalidate the promiseNodeForTrace reference stored in AwaiterBase. To fix this, I'm storing a reference to the currently-awaited OwnPromiseNode, rather than to PromiseNode directly.

I tried to modify async-coroutine-test.c++'s trace test case to validate this change (by modifying the awaited .then() promise to return NEVER_DONE, and tracing after fulfilling the base promise on which .then() was called), but was unable to reproduce the buggy behavior. I'm open to ideas.

### DIFF
--- a/c++/src/kj/async-inl.h
+++ b/c++/src/kj/async-inl.h
@@ -2232,9 +2232,11 @@ private:
   // preserve that assertion.
 #endif
 
-  Maybe<PromiseNode&> promiseNodeForTrace;
+  Maybe<OwnPromiseNode&> promiseNodeForTrace;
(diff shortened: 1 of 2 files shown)


kentonv (approved):

harrishancock:
Months later, I still haven't found time to develop a regression test, but this is still biting us. I'm going to go ahead and merge.

PR #2021, opened 2024-05-02
Fix undefined behavior violations

fhanau:
Fixes an instance of undefined behavior when calling `releaseAs*()` on a kj::Array that has been allocated using a nullptr to indicate an empty array and thus has a nullptr disposer.
There might be more places where releaseAs is used in this way and a better way to fix this, open to suggestions here but also want to ship this soon.

Added a second commit to fix a remaining issue. Also see the downstream branch.

### DIFF
--- a/c++/src/capnp/rpc.c++
+++ b/c++/src/capnp/rpc.c++
@@ -2340,7 +2340,7 @@ private:
     RpcServerResponseImpl& response;
     kj::Own<RpcCallContext> context;  // owns `response`
 
-    kj::Own<ClientHook> getResolutionAtReturnTime(
+    static kj::Own<ClientHook> getResolutionAtReturnTime(
         kj::Own<ClientHook> original, RpcServerResponseImpl::Resolution resolution) {
       // Wait for `original` to resolve to `resolution.returnedCap`, then return
       // `resolution.unwrapped`.
@@ -2358,7 +2358,7 @@ private:
 
       KJ_IF_SOME(p, ptr->whenMoreResolved()) {
         return newLocalPromiseClient(p.then(
-            [this, original = kj::mv(original), resolution = kj::mv(resolution)]
+            [original = kj::mv(original), resolution = kj::mv(resolution)]
             (kj::Own<ClientHook> r) mutable {
           return getResolutionAtReturnTime(kj::mv(r), kj::mv(resolution));
         }));
--- a/c++/src/kj/array.h
+++ b/c++/src/kj/array.h
@@ -210,6 +210,7 @@ public:
     // Like asBytes() but transfers ownership.
     static_assert(sizeof(T) == sizeof(byte),
         "releaseAsBytes() only possible on arrays with byte-size elements (e.g. chars).");
+    if (disposer == nullptr) return nullptr;
     Array<PropagateConst<T, byte>> result(
         reinterpret_cast<PropagateConst<T, byte>*>(ptr), size_, *disposer);
     ptr = nullptr;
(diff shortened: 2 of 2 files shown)


c++/src/capnp/rpc.c++
```diff
@@ -2358,7 +2358,7 @@ private:
 
       KJ_IF_SOME(p, ptr->whenMoreResolved()) {
         return newLocalPromiseClient(p.then(
-            [this, original = kj::mv(original), resolution = kj::mv(resolution)]
+            [original = kj::mv(original), resolution = kj::mv(resolution)]
             (kj::Own<ClientHook> r) mutable {
           return getResolutionAtReturnTime(kj::mv(r), kj::mv(resolution));
```
kentonv, line 2363: I don't quite understand the commit message. Is the point here that this call previously could be a use-after-free of `this`?

c++/src/capnp/schema-parser.c++
```diff
@@ -102,7 +102,13 @@ public:
...
   kj::Maybe<kj::Array<const byte>> embedRelative(kj::StringPtr embedPath) override {
     KJ_IF_SOME(importedFile, file->import(embedPath)) {
-      return importedFile->readContent().releaseAsBytes();
+      auto content = importedFile->readContent();
+      // Do not try to read an empty file, but return an empty array (kj::none would indicate a
+      // read error).
+      if (!content.size()) {
+        return kj::Array<const byte>(nullptr);
+      }
+      return content.releaseAsBytes();
```
kentonv, line 111: I think we should fix `releaseAsBytes()` and `releaseAsChars()` directly, rather than make all the call sites think about this.

The problem appears to be that these functions allocate a new `Array`, calling the constructor where they pass a custom disposer. The constructor takes the disposer by reference, so if the disposer is null, this is UB, since references can't be null.

Probably, `releaseAsBytes()` needs to check for `disposer == nullptr` like:

```
  inline Array<PropagateConst<T, byte>> releaseAsBytes() {
    // Like asBytes() but transfers ownership.
    static_assert(sizeof(T) == sizeof(byte),
        "releaseAsBytes() only possible on arrays with byte-size elements (e.g. chars).");
    if (disposer == nullptr) return nullptr;
    Array<PropagateConst<T, byte>> result(
        reinterpret_cast<PropagateConst<T, byte>*>(ptr), size_, *disposer);
    ptr = nullptr;
    size_ = 0;
    return result;
  }
```

And similarly for `releaseAsChars()`.

c++/src/capnp/rpc.c++
fhanau: Not entirely sure TBH, the vptr check is confusing to me. I think there would be UAF if `this` was actually being used here, but `getResolutionAtReturnTime` doesn't actually access any class variables. The vptr check still detects that `this` is no longer valid, resulting in error messages like this:

`external/capnp-cpp/src/capnp/rpc.c++:2363:18: runtime error: member call on address 0x7f20b4080000 which does not point to an object of type 'capnp::_::(anonymous namespace)::RpcConnectionState::PostReturnRpcPipeline'|0x7f20b4080000: note: object has invalid vptr| 00 00 00 00  60 fb [...] || SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior external/capnp-cpp/src/capnp/rpc.c++:2363:18`

c++/src/capnp/schema-parser.c++
fhanau: 👍  that approach should be more comprehensive, updated the commit to use that instead

c++/src/capnp/rpc.c++
kentonv: Yes, I think this is simply saying that the object has been destroyed before this point.

ASAN doesn't catch it because we don't actually access the object, but it looks like UBSAN adds checks to verify the object is valid on any method call.

kentonv (approved):

PR #2022, opened 2024-05-02
Add clang::musttail attributes to a couple recursive calls

jclee:
Should reduce noise in stack traces, and also ensures that clang will tell us if future code changes inadvertently prevent tail call optimization.

### DIFF
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -259,6 +259,14 @@ typedef unsigned char byte;
 // by allowing a syntax like `[[clang::lifetimebound(*this)]]`.
 // https://clang.llvm.org/docs/AttributeReference.html#lifetimebound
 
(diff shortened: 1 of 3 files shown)


kentonv:
Does this actually cause clang to use TCO when it might not have otherwise? Or does it just cause an error in cases where Clang otherwise would not have used TCO?

fhanau:
> Does this actually cause clang to use TCO when it might not have otherwise? Or does it just cause an error in cases where Clang otherwise would not have used TCO?

Sounds like it will use TCO when it didn't otherwise when optimization is disabled:
https://clang.llvm.org/docs/AttributeReference.html#musttail
```If a return statement is marked musttail, this indicates that the compiler must generate a tail call for the program to be correct, even when optimizations are disabled. This guarantees that the call will not cause unbounded stack growth if it is part of a recursive cycle in the call graph.```

kentonv (approved):

PR #2030, opened 2024-05-09
Add Exception Details to RPC Exceptions

Warfields:
Propagates details from exceptions over RPC.


c++/src/kj/exception.h
```diff
@@ -135,8 +135,14 @@ class Exception {
...
 
   using DetailTypeId = unsigned long long;
+  struct Detail {
+    DetailTypeId id;
+    kj::Array<byte> value;
+  };
 
   kj::Maybe<kj::ArrayPtr<const byte>> getDetail(DetailTypeId typeId) const;
+  kj::ArrayPtr<const Detail> getDetails() const;
+  bool hasDetails() const;
```
jasnell, line 145: `hasDetails()` isn't strictly necessary here. Someone could just check `getDetails().size() != 0`, or have `getDetail()` return a `kj::Maybe<kj::ArrayPtr<const Detail>>` (but I prefer just the size check).

Warfields: Done, I was over complicating that.

Warfields: Helps if I push up the change :D

jasnell (commented):
The changes to `kj::Exception` LGTM. I'll leave it to someone else to review the `src/capnp/rpc*` changes

c++/src/capnp/test-util.c++
```diff
@@ -1192,6 +1192,13 @@ kj::Promise<void> TestMoreStuffImpl::throwRemoteException(ThrowRemoteExceptionCo
   return KJ_EXCEPTION(FAILED, "remote exception: test exception");
 }
 
+kj::Promise<void> TestMoreStuffImpl::throwExceptionWithDetail(
+    ThrowExceptionWithDetailContext context) {
+  auto e = KJ_EXCEPTION(FAILED, "test exception");
+  e.setDetail(1, kj::heapArray<byte>(0));
```
kentonv, line 1198: We should probably test a non-empty detail to make sure the bytes are in fact encoded correctly?

PR #2031, opened 2024-05-09
enable CI on v2 branch

mikea:
I look at https://github.com/capnproto/capnproto/commits/v2/ and it doesn't seem that we have builds enabled.


fhanau (approved):
Nice catch!

PR #2033, opened 2024-05-10
Delete c++/ekam-provider directory

mikea:
I assume we don't need it anymore?


kentonv:
The workers runtime doesn't use ekam anymore, but I'm not sure I want to delete Ekam support from capnp. Frankly I'd be more excited about deleting autotools and cmake.

Anyway, these files don't seem like they're getting in the way of anything?

PR #2034, opened 2024-05-10
heapArray filled version

mikea:
For trivial types, heapArray returns unitialized memory, which could be a security issue.
While the implementation is possible for non-trivial types, it is less of an issue and will probably be seldom used.

As an alternative I could suggest `heapZeroArray()` and provide only zero-initialized heap arrays.

### DIFF
--- a/c++/src/kj/array.h
+++ b/c++/src/kj/array.h
@@ -322,6 +322,15 @@ template <typename T, typename Iterator> Array<T> heapArray(Iterator begin, Iter
 template <typename T> Array<T> heapArray(std::initializer_list<T> init);
 // Allocate a heap array containing a copy of the given content.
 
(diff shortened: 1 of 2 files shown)


c++/src/kj/array.h
```diff
@@ -322,6 +322,15 @@ template <typename T, typename Iterator> Array<T> heapArray(Iterator begin, Iter
 template <typename T> Array<T> heapArray(std::initializer_list<T> init);
 // Allocate a heap array containing a copy of the given content.
 
+template <typename T, typename = EnableIf<KJ_HAS_TRIVIAL_CONSTRUCTOR(T)>>
+inline Array<T> heapArray(size_t size, T t) {
+  // Allocate array pre-filled with t.
+  // TODO: implement for complex T types without creating `size` instances first.
```
kentonv, line 328: I think you can easily implement this using ArrayBuilder FWIW.

kentonv (approved):

PR #2035, opened 2024-05-10
kj::ArrayPtr<T>::copyFrom method

mikea:
Type-safe alternative to memcpy. As with the fill case compilers are pretty good at optimizing it too:

https://godbolt.org/z/dfMxeT1M5


c++/src/kj/common.h
```diff
@@ -1954,6 +1954,13 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
     // libc++ std::fill doesn't have memset specialization either.
   }
 
+  inline void copyFrom(kj::ArrayPtr<T> other) {
+    // Copy data from the other array pointer.
+    // Arrays have to be of the same size and memory area MUST NOT overlap.
```
mikea, line 1959: `memcpy` has non-overlapping requirement too. I could implement overlapping copy if you think it is important.

jasnell: I don't think we have any cases currently where we have an overlapping copy so I think this is fine. I do wonder if it wouldn't be worthwhile to at least have a `KJ_DASSERT` in here ensuring `other` is not overlapping?

c++/src/kj/common.h
```diff
@@ -1954,6 +1954,13 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
...
+    KJ_IREQUIRE(size_ == other.size(), "copy requires arrays of the same size");
```
jasnell, line 1960: I do also wonder here why these need to be the same size. It should still be safe if `other.size() <= size()`, and just have the method return the actual number of elements copied.

mikea:
> There is probably a significant performance difference between memcpy and a sequential loop, since memcpy can use vector instructions and such.

Compiler happily generates vector instructions too when you mark function as noinline. I do not think memcpy has any leg here.

> In your godbolt, if you mark `copy()` as noinline, then it no longer optimizes to memcpy. If you add `__restrict` to one of the input pointers, then it is able to use memcpy again. The problem is that the compiler can only optimized to memcpy if it knows the arrays do not overlap.

But I _do_ have method as inline.

> In real-world applications, it will probably often be the case that the compiler does not know if the two ArrayPtrs point at overlapping memory, so it won't be able to optimize to memcpy.

Would you rather I used manual memcpy or used restrict pointers to help compiler figure it out?

c++/src/kj/common.h
mikea: I was thinking about it but imo it is much more error-prone this way. It is very simple to size-align slices. But spotting the fact that you don't have enough in the target and are loosing data might be hard. 

I'd rather force clients to check lengths.

kentonv:
> But I do have method as inline.

Yes but the point is not the inlining itself, but that in your particular test case, the compiler is able to see at the inline site that the arrays are not overlapping. If it couldn't see that, then even with inlining it would still have the same problem.

> Compiler happily generates vector instructions too when you mark function as noinline.

Normally, a loop like this without restrict pointers cannot use vector instructions because the loop cannot be unrolled because the behavior would differ in the case of overlap.

However, you are right that in the godbolt example, marking the copy() function noinline does produce an implementation with vector ops.

However, this appears to be because it produces a 100-line long output that actually dynamically tests for overlap and chooses an optimized or unoptimized approach depending on what it finds. We really don't want to be generating this quantity of code whenever we do a copy. We really need to let the compiler know that it's safe to assume no overlap.

> Would you rather I used manual memcpy or used restrict pointers to help compiler figure it out?

Since it looks like the compiler will optimized a restrict pointer copy into memcpy(), I guess either way is equivalent. I don't have a strong opinion.

c++/src/kj/common.h
jasnell: Fair enough!

jasnell:
LGTM from an API perspective. I'll leave it to someone else to hit the approve button (I'm assuming after the compiler optimization discussion is resolved)

mikea:
updated the loop to be more optimization-friendly and added intersect check. ptal.

jasnell:
Still LGTM but will let someone else hit the approve button

c++/src/kj/common.h
```diff
@@ -1954,9 +1954,28 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
...
+
 private:
   T* ptr;
   size_t size_;
+
+  inline bool intersects(kj::ArrayPtr<const T> other) const {
+    // Checks if memory area intersects with another pointer.
+
+    // Memory _does not_ intersect if one of the arrays is completely on one side of the other:
+    //    begin() >= other.end || other.begin() <= end()
```
kentonv, line 1975:
```suggestion
    //    begin() >= other.end || other.begin() >= end()
```

kentonv (approved):

PR #2036, opened 2024-05-13
kj::arrayPtr<T>(T &t) creates a single-element array pointer

mikea:
This is a common operation when dealing with IO.


jasnell (approved):

mikea:
> `arrayPtr(x)` could be fine as long as there is no other single-arg overload... if there are other single-arg overloads then I worry about ambiguity.

There is a single-arg overload that takes `T[size]`, so maybe `T&` overload is ok and clear enough?
I've updated the pr, how does it look? @kentonv @jasnell

kentonv (approved):

PR #2037, opened 2024-05-14
replacing few memset usages with safer alternatives

mikea:


c++/src/capnp/compiler/capnp.c++
```diff
@@ -1704,8 +1704,7 @@ public:
 
     // Evaluate this schema to a DynamicValue.
     DynamicValue::Reader value;
-    word zeroWord[1];
-    memset(&zeroWord, 0, sizeof(zeroWord));
+    word zeroWord[1]{};
```
kentonv, line 1707: Does this actually zero-initialize given the unusual way `class word` is defined?

kentonv: Even if it does... I wonder if this ends up violating aliasing rules. Word arrays are never actually accessed using the type `word`, but rather by reinterpret-casting to some other pointer type and reading that. Technically it is UB to write type `word` and then read as some other non-byte-sized type, or vice versa. There's an exception for byte-oriented access, which is allowed to alias everything; memset() operates on bytes, and so doesn't break aliasing rules.

Unsure to what extent any of this matters in practice.

mikea: >Does this actually zero-initialize given the unusual way class word is defined?

yes, I was curious about this as well: https://godbolt.org/z/Wcsn99WKc

>  Technically it is UB to write type word and then read as some other non-byte-sized type, or vice versa.
Not sure if this matters, but we don't write to word here, but default-initialize it?

Is there a special compiler mode that could check this for us?

> Unsure to what extent any of this matters in practice.

I could use `kj::memzero()` for word arrays instead.

mikea: Compilers claim to have some aliasing analysis but I wasn't able to trigger it, did I break the rules here: https://godbolt.org/z/WrrG9jres ?

kentonv: That does appear to break the rules, not sure why it isn't detected. My understanding is that the compiler could legally re-order the assignment to `buf2->content` vs. `f->d`.

PR #2039, opened 2024-05-14
[nfc] Clean up recursive header dependencies

fhanau:
This should slightly improve compile speeds, especially under MSVC. To keep this change small this primarily affects headers, which should already allow us to eliminate many superfluous recursive includes.

- Limit usage of intrin.h for MSVC, this is generally heavy
- Drop some support for MSVC versions older than VS 2017 Update 9 (MSVC_VER 1916). We likely don't support this version anyway based on coroutines/C++20 support, but dropping more _MSC_VER checks can be done in another PR.

Further potential to reduce recursive includes exists by refactoring async code so that coroutines are not always included, as well as isolating usage of large libc++ headers like \<set\> and \<atomic\>. I experimented with those and based on looking at preprocessed source code sizes in a more rigorous way this leads to larger improvements in include sizes, but the advantage in compile time is harder to evaluate.

This will likely require some include changes downstream which are still Work-In-Progress, but this should be final and ready for review.

### DIFF
--- a/c++/src/capnp/compat/json.capnp.h
+++ b/c++/src/capnp/compat/json.capnp.h
@@ -5,9 +5,6 @@
 
 #include <capnp/generated-header-support.h>
 #include <kj/windows-sanity.h>
-#if !CAPNP_LITE
-#include <capnp/capability.h>
-#endif  // !CAPNP_LITE
 
 #ifndef CAPNP_VERSION
 #error "CAPNP_VERSION is not defined, is capnp/generated-header-support.h missing?"
--- a/c++/src/capnp/any.h
+++ b/c++/src/capnp/any.h
@@ -22,7 +22,6 @@
 #pragma once
 
 #include "layout.h"
-#include "pointer-helpers.h"
 #include "orphan.h"
 #include "list.h"
 #include <kj/windows-sanity.h>  // work-around macro conflict with `VOID`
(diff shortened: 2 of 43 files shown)


c++/src/capnp/compat/json.capnp.h
```diff
@@ -5,9 +5,6 @@
 
 #include <capnp/generated-header-support.h>
 #include <kj/windows-sanity.h>
-#if !CAPNP_LITE
```
fhanau, line 8: Regenerated the header.

mikea (approved):

c++/src/capnp/compat/json.capnp.h
kentonv: How did you generate this header? When I run `regenerate-bootstraps.sh`, it brings this include back. It turns out this is due to a bug in the code generator which I'll fix: if a previous file listed on the same command line contained interfaces, then latter files have this include even if they don't need it.

But, are you generating this header using something other than the `regenerate-bootstraps.sh` script?

fhanau: TBH I used that script before but forgot that it existed. Based on zsh history, I just ran capnp_tool manually (`bazel-bin/src/capnp/capnp_tool compile --verbose -Isrc --no-standard-import --src-prefix=src -obazel-bin/src/capnp/capnpc-c++:src src/capnp/compat/json.capnp`) after noticing that the header had includes it didn't need.

PR #2040, opened 2024-05-15
[build] Increase clang minimum version to 14, use newly added Ubuntu 24.04 CI image to test clang 14 - 18

fhanau:
This updates the clang version window we support. A few points to consider here:
- There is no short-term need to merge this, but doing so would help us avoid breakage with modern clang versions. The Ubuntu 24.04 image is still in 'beta' state, but in my experience the 'release' CI runners aren't all that reliable either unfortunately.
- clang 15 or 16 are now supported on Ubuntu oldstable and Debian oldoldstable, so the new version window should be widely supported. The Ubuntu 22.04 image has clang 15 as its highest version and the 24.04 image has clang 14 as its lowest version, so rolling forward by fewer than 3 versions would be non-trivial.


c++/src/kj/common.h
```diff
@@ -275,9 +275,8 @@ typedef unsigned char byte;
 #define KJ_UNUSED_MEMBER
 #endif
 
-#if __cplusplus > 201703L || (__clang__  && __clang_major__ >= 9 && __cplusplus >= 201103L)
-// Technically this was only added to C++20 but Clang allows it for >= C++11 and spelunking the
-// attributes manual indicates it first came in with Clang 9.
+#if __cplusplus > 201703L || (__clang__  && __cplusplus >= 201103L)
+// Technically this was only added to C++20 but Clang allows it for >= C++11.
```
kentonv, line 279: At this point we require C++20 anyway so we can probably remove this conditional?

kentonv (approved):

PR #2042, opened 2024-05-17
enable -Wsuggest-override

mikea:
Having override annotations in place greatly helps with API
refactorings. I also believe this is a commonly accepted coding
practice.

### DIFF
--- a/c++/.bazelrc
+++ b/c++/.bazelrc
@@ -2,8 +2,9 @@ common --noenable_bzlmod
 common --enable_platform_specific_config
 
 build:unix --cxxopt='-std=c++20' --host_cxxopt='-std=c++20' --force_pic --verbose_failures
-build:unix --cxxopt='-Wall' --host_cxxopt='-Wall'
-build:unix --cxxopt='-Wextra' --host_cxxopt='-Wextra'
+build:unix --cxxopt='-Wall'
+build:unix --cxxopt='-Wextra'
+build:unix --cxxopt='-Wsuggest-override'
 build:unix --cxxopt='-Wno-strict-aliasing' --host_cxxopt='-Wno-strict-aliasing'
 build:unix --cxxopt='-Wno-sign-compare' --host_cxxopt='-Wno-sign-compare'
 build:unix --cxxopt='-Wno-unused-parameter' --host_cxxopt='-Wno-unused-parameter'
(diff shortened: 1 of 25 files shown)


jasnell (approved):

c++/.bazelrc
```diff
@@ -2,12 +2,14 @@ common --noenable_bzlmod
...
-build:unix --cxxopt='-Wall' --host_cxxopt='-Wall'
-build:unix --cxxopt='-Wextra' --host_cxxopt='-Wextra'
+build:unix --cxxopt='-Wall'
+build:unix --cxxopt='-Wextra'
+build:unix --cxxopt='-Wsuggest-override'
 build:unix --cxxopt='-Wno-strict-aliasing' --host_cxxopt='-Wno-strict-aliasing'
 build:unix --cxxopt='-Wno-sign-compare' --host_cxxopt='-Wno-sign-compare'
 build:unix --cxxopt='-Wno-unused-parameter' --host_cxxopt='-Wno-unused-parameter'
 build:unix --cxxopt='-Wno-deprecated-this-capture' --host_cxxopt='-Wno-deprecated-this-capture'
+build:unix --cxxopt='-Wno-deprecated-this-capture' --host_cxxopt='-Wno-deprecated-this-capture'
```
fhanau, line 12: Nit: duplicate line

fhanau:
Figured out why the super-test build fails with a suggest-override warning turned into an error, while the bazel build doesn't even emit the warning: The bazel build fails to enable `-DCAPNP_HEADER_WARNINGS`, causing the warning to not be emitted with bazel. Since the warning is in a generated capnp header we might need changes to the capnp code generator to fully address this.

mikea:
> Figured out why the super-test build fails with a suggest-override warning turned into an error, while the bazel build doesn't even emit the warning: The bazel build fails to enable `-DCAPNP_HEADER_WARNINGS`, causing the warning to not be emitted with bazel. Since the warning is in a generated capnp header we might need changes to the capnp code generator to fully address this.

awesome, this indeed helps. I have added `CAPNP_HEADER_WARNINGS` to `.bazelrc`

mikea:
I tried to bring bazel behavior closer to super-test here: https://github.com/capnproto/capnproto/pull/2045 @fhanau

fhanau (approved):

kentonv:
Oh cool, if I'd known about this flag I would have enabled it a long time ago!

PR #2043, opened 2024-05-17
supporting constexpr _kjb suffix

mikea:
The original implementation can't be used because of cast limitations: https://godbolt.org/z/fWb7ncrh5

This version casts each character separately to achieve the goal (by @harrishancock)

### DIFF
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -2194,12 +2194,30 @@ inline void memzero(T& t) {
   memset(&t, 0, sizeof(T));
 }
 
+namespace _ {
+template <size_t N>
+struct ByteLiteral {
+  // Helper class to implement _kjb suffix: constexpr is not allowed to cast `char*`
+  // to `unsigned char*`. To overcome this limitation every char needs to be cast individually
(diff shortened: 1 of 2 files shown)


mikea:
This will also require us dropping clang < 14, idk how do we feel about that?

jasnell (approved):
LGTM but let's have @kentonv review also

c++/src/kj/common.h
```diff
@@ -2194,12 +2194,32 @@ inline void memzero(T& t) {
...
+  // This class may not contain any private members, or else you get a
+  // misleading compiler error.
+
+  constexpr ByteLiteral(const char (&init)[N]) {
+    for (auto i = 0; i < N; ++i) {      
+      data[i] = (kj::byte)(init[i]);
+    }
+  }
+  constexpr size_t size() const { return N - 1; }
+  kj::byte data[N];
```
kentonv, line 2212: As long as we're here, how about making the size `N - 1` so that we don't store the NUL terminator?

kentonv (approved):

PR #2044, opened 2024-05-17
fix implicit this capture warning

mikea:


c++/src/kj/async-io-unix.c++
```diff
@@ -199,7 +199,7 @@ public:
 
     if (n < 0) {
       // EAGAIN -- need to wait for writability and try again.
-      return observer.whenBecomesWritable().then([=]() {
+      return observer.whenBecomesWritable().then([=,this]() {
```
jasnell, line 202: nit: should we avoid the `=` capture and grab each capture individually?

jasnell (approved):

PR #2045, opened 2024-05-17
run bazel CI build with -Werror

mikea:
this matches super-test.sh behavior

### DIFF
--- a/.github/workflows/quick-test.yml
+++ b/.github/workflows/quick-test.yml
@@ -35,7 +35,7 @@ jobs:
         driver: [bazel, super-test]
         include:
           - driver: bazel
-            run-test: cd c++ && bazel test --verbose_failures --test_output=errors -k //...
+            run-test: cd c++ && bazel test --verbose_failures --test_output=errors -k --cxxopt='-Werror' //...
           - driver: super-test
             run-test: ./super-test.sh quick
           - clang: 12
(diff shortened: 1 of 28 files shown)


c++/.bazelrc
```diff
@@ -8,6 +8,7 @@ build:unix --cxxopt='-Wno-strict-aliasing' --host_cxxopt='-Wno-strict-aliasing'
 build:unix --cxxopt='-Wno-sign-compare' --host_cxxopt='-Wno-sign-compare'
 build:unix --cxxopt='-Wno-unused-parameter' --host_cxxopt='-Wno-unused-parameter'
 build:unix --cxxopt='-Wno-deprecated-this-capture' --host_cxxopt='-Wno-deprecated-this-capture'
+build:unix --cxxopt='-DCAPNP_HEADER_WARNINGS'
```
fhanau, line 12: Also add `KJ_HEADER_WARNINGS` (same thing for KJ) and `CAPNP_DEBUG_TYPES` (enabled by default in non-bazel build, code size overhead is acceptable for debug builds) as well, the latter surfaces another warning.

.github/workflows/quick-test.yml
```diff
@@ -35,7 +35,7 @@ jobs:
         driver: [bazel, super-test]
         include:
           - driver: bazel
-            run-test: cd c++ && bazel test --verbose_failures --test_output=errors -k //...
+            run-test: cd c++ && bazel test --verbose_failures --test_output=errors -k --cxxopt='-Werror' //...
```
fhanau, line 38: I was concerned about Werror slowing down local development but this way it is only applied on CI, that works well for me.

fhanau (approved):

PR #2047, opened 2024-05-17
Log peerId in kj::Exception when tls cert is expired

jqmmes:
When a client cert is expired, we will get an SSL error informing of that in the server. But we don't log the details of the client that originated this error.

This PR changes `wrapServer` to catch the exception thrown and attach the `stream.peerIdentity` to the `kj::Exception` description.

### DIFF
--- a/c++/src/kj/compat/tls.c++
+++ b/c++/src/kj/compat/tls.c++
@@ -926,9 +926,13 @@ kj::Promise<kj::AuthenticatedStream> TlsContext::wrapServer(kj::AuthenticatedStr
       return KJ_EXCEPTION(DISCONNECTED, "timed out waiting for client during TLS handshake");
     }).exclusiveJoin(kj::mv(promise));
   }
+  auto peerId = stream.peerIdentity->toString();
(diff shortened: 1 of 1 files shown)


c++/src/kj/compat/tls.c++
```diff
@@ -926,9 +926,13 @@ kj::Promise<kj::AuthenticatedStream> TlsContext::wrapServer(kj::AuthenticatedStr
       return KJ_EXCEPTION(DISCONNECTED, "timed out waiting for client during TLS handshake");
     }).exclusiveJoin(kj::mv(promise));
   }
+  auto peerId = kj::str(stream.peerIdentity->toString());
```
harrishancock, line 929: It looks like `toString()` returns a `kj::String` already, so the `kj::str()` is unnecessary.

Separately, I think we could save at least one allocation on the happy path if we converted this to a coroutine. Then we could let `stream` live on the "stack" (coroutine frame) and only call its `toString()` in the catch block. If you'd rather not convert this to a coroutine just to report this, that's fine with me, though.

kentonv (approved):

PR #2048, opened 2024-05-18
[nfc] Clean up remaining lock blocking tracking bazel configuration

fhanau:
These are obsolete after #1944 and break `bazel cquery 'deps(//...)'` since the corresponding bool_flags are no longer available.


mikea (approved):
Thanks

PR #2049, opened 2024-05-19
[nfc] Clean up bazel defines, apply buildifier

fhanau:
Moving configuration to `build/BUILD` instead of including it in `src/kj/BUILD.bazel` is better in terms of style and a bit more concise.

Note: This will require slight downstream changes, PRs for that will follow.


c++/build/BUILD
```diff
@@ -0,0 +1,88 @@
+load("@bazel_skylib//rules:common_settings.bzl", "bool_flag")
```
mikea, line 1: I think @kentonv wanted flags to be in src/kj since they are kj flags and not capnp flags.

PR #2050, opened 2024-05-20
kj::ArrayPtr::split()

mikea:


mikea:
I am not sure what's the best way to handle Vector dependency here?

c++/src/kj/common-test.c++
```diff
@@ -1039,6 +1040,12 @@ KJ_TEST("single item arrayPtr()") {
   KJ_EXPECT(obj.d == 0);
 }
 
+KJ_TEST("ArrayPtr:split") {
+  auto buffer = "abc def ghij"_kjb;
+  auto buffers = buffer.split(' ');
+  KJ_EXPECT(buffers.size() == 3);
+}
```
jasnell, line 1047: In some of the practical use cases we have (for instance, see my new workerd PR  implementing the `data:` URL support ... we might actually need to split on the last instance of delim, or perhaps split *only* around the first instance. My fear here is that, `split(...)` as defined here will only be useful for a limited number of cases

mikea: I see people reaching out to std:: streams in our code base to do splits using `\n` and converting things back. So it is definitely usefull.

Happy to rename. `splitAll()`?

c++/src/kj/common.h
```diff
@@ -1981,7 +1985,6 @@ inline Maybe<size_t> ArrayPtr<const char>::findFirst(const char& c) const {
     return pos - ptr;
   }
 }
```
jasnell, line 1987: non-blocking nit
```suggestion
}

```

c++/src/kj/common-test.c++
jasnell: Oh, no doubt it'll be useful, I just think we need to be considering the other use cases here too.

Another api alternative here could be to have `split(...)` return a bidirectional iterator rather than a `kj::Array` so that if I want to limit the number of splits, or reverse the iterator to start from the back, it would be trivial to do so

mikea: You just use `kj::Array` as bi-di iterator, can you? It would be a bit vasteful, but at the very least the code pattern would already be correct and future-proof.

> Obviously, tho, that's a more complicated implementation so consider it a non-blocking suggestion

Sure, I'm game for a follow up iterator implementation.

jasnell (approved):

kentonv:
I guess this implementation does memory allocations to construct an array of results. I wonder if it would make sense to return an iterable that doesn't need to allocate at all?

Or, perhaps something like `splitFirst()`, which returns the first piece and also updates the array to contain only the remaining data... so you can call `splitFirst()` repeatedly without allocating. I think this is the pattern I usually end up with in practice. (An iterable seems cleaner than this, but a bit more work.)

PR #2053, opened 2024-05-22
fix PromiseArena memory leak

mikea:
was caught by async-coroutine-test

### DIFF
--- a/c++/src/kj/async-inl.h
+++ b/c++/src/kj/async-inl.h
@@ -351,8 +351,10 @@ public:
 
   static void dispose(PromiseArenaMember* node) {
     PromiseArena* arena = node->arena;
+    // Defer the `delete` to protect against exception in `destroy()`.
+    // Reminder: `delete` automatically ignores null pointers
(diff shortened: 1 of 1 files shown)


c++/src/kj/async-inl.h
```diff
@@ -351,8 +351,10 @@ class PromiseDisposer {
 
   static void dispose(PromiseArenaMember* node) {
     PromiseArena* arena = node->arena;
+    // Defer the `delete` to protect against exception in `destroy()`.
+    // Reminder: `delete` automatically ignores null pointers
+    KJ_DEFER(delete arena);
     node->destroy();
```
jasnell, line 357: Any way of having a test for this?

jasnell (approved):

c++/src/kj/async-inl.h
mikea: as I wrote in comment - there is already a test for this. It is just that capnproto doesn't have an asan CI configuration.

kentonv (approved):

PR #2056, opened 2024-05-23
[nfc] asan CI build

mikea:
I plan to add ubsan to the matrix next

### DIFF
--- a/.github/workflows/quick-test.yml
+++ b/.github/workflows/quick-test.yml
@@ -12,6 +12,13 @@ on:
     - v2
     - 'release-*'
 
+env:
+  INSTALL_DEPS: |
+    install_deps() {
+      local clang=$1
+      export DEBIAN_FRONTEND=noninteractive && sudo apt-get update && sudo apt-get install --no-upgrade -y build-essential git zlib1g-dev cmake libssl-dev clang-$clang libc++-$clang-dev libc++abi-$clang-dev
+    }
+
 jobs:
   Linux-musl:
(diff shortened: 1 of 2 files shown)


.github/workflows/quick-test.yml
```diff
@@ -46,12 +46,32 @@ jobs:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        clang: [18]
+        config: [asan]
+    steps:
+      - uses: actions/checkout@v4
+      - name: install dependencies
+        # I observed 404s for some packages and added an `apt-get update`.
+        run: |
```
mikea, line 66: not sure if there's a way to dedup this step here? @fhanau

fhanau: You could add `driver: bazel-asan` `run-test:  cd c++ && bazel test --verbose_failures --test_output=errors -k --config=asan //...` to `Linux-clang` above and then filter out clang versions other than 18. Might not be much shorter though

mikea: So, I wanted to try yaml anchors, but apparently they are not supported on github actions: https://github.com/orgs/community/discussions/4501

mikea: OK, I did some bash hackery. How about now? @fhanau

fhanau (approved):

fhanau:
LGTM, thought I had already approved yesterday. `libc++abi-$clang-dev` looks a bit wonky, but looks like it is working correctly and not trying to resolve `$clang-dev`.

PR #2062, opened 2024-05-29
Support comparators for ArrayPtr and refactor StringPtr to use them internally.

AdityaAtulTewari:
ArrayPtr comparators  are implemented by implementing `operator<` and `operator==`.

This was chosen over `operator<=>` or spaceship as that generally requires the `compare` header which was deemed to include too much of stdlib and thus make compilation times less under our control.

### DIFF
--- a/c++/src/kj/common-test.c++
+++ b/c++/src/kj/common-test.c++
@@ -793,30 +793,178 @@ TEST(Common, ArrayAsBytes) {
   }
 }
 
-KJ_TEST("ArrayPtr operator ==") {
-  KJ_EXPECT(ArrayPtr<const int>({123, 456}) == ArrayPtr<const int>({123, 456}));
-  KJ_EXPECT(!(ArrayPtr<const int>({123, 456}) != ArrayPtr<const int>({123, 456})));
-  KJ_EXPECT(ArrayPtr<const int>({123, 456}) != ArrayPtr<const int>({123, 321}));
-  KJ_EXPECT(ArrayPtr<const int>({123, 456}) != ArrayPtr<const int>({123}));
+enum TestOrdering {
+  UNORDERED,
+  EQUAL,
+  LESS,
+  GREATER,
+  NOTEQUAL,
+};
 
-  KJ_EXPECT(ArrayPtr<const int>({123, 456}) == ArrayPtr<const short>({123, 456}));
-  KJ_EXPECT(!(ArrayPtr<const int>({123, 456}) != ArrayPtr<const short>({123, 456})));
-  KJ_EXPECT(ArrayPtr<const int>({123, 456}) != ArrayPtr<const short>({123, 321}));
-  KJ_EXPECT(ArrayPtr<const int>({123, 456}) != ArrayPtr<const short>({123}));
-
-  KJ_EXPECT((ArrayPtr<const StringPtr>({"foo", "bar"}) ==
-             ArrayPtr<const char* const>({"foo", "bar"})));
-  KJ_EXPECT(!(ArrayPtr<const StringPtr>({"foo", "bar"}) !=
-             ArrayPtr<const char* const>({"foo", "bar"})));
-  KJ_EXPECT((ArrayPtr<const StringPtr>({"foo", "bar"}) !=
-             ArrayPtr<const char* const>({"foo", "baz"})));
-  KJ_EXPECT((ArrayPtr<const StringPtr>({"foo", "bar"}) !=
-             ArrayPtr<const char* const>({"foo"})));
-
-  // operator== should not use memcmp for double elements.
-  double d[1] = { nan() };
-  KJ_EXPECT(ArrayPtr<double>(d, 1) != ArrayPtr<double>(d, 1));
+template<typename A, typename B>
+void verifyEqualityComparisons(A a, B b, TestOrdering ord) {
+  const bool expectedEq = ord == EQUAL;
+  KJ_EXPECT((a == b) == expectedEq);
+  KJ_EXPECT((b == a) == expectedEq);
+  KJ_EXPECT((a != b) == !expectedEq);
+  KJ_EXPECT((b != a) == !expectedEq);
 }
+
+template<typename T>
+void strongComparisonsTests(T a, T b, TestOrdering ord) {
+  const bool expectedEq = ord == EQUAL;
+  const bool expectedLT = ord == LESS;
+  verifyEqualityComparisons(a, b, ord);
+  KJ_EXPECT((a <= b) == (expectedEq || expectedLT));
+  KJ_EXPECT((b <= a) == !expectedLT);
+  KJ_EXPECT((a >= b) == !expectedLT);
+  KJ_EXPECT((b >= a) == (expectedEq || expectedLT));
+  KJ_EXPECT((a < b) == expectedLT);
+  KJ_EXPECT((b < a) == !(expectedEq || expectedLT));
+  KJ_EXPECT((a > b) == !(expectedEq || expectedLT));
+  KJ_EXPECT((b > a) == expectedLT);
+}
+
+template<typename A, typename B>
+struct ArrayComparisonTest {
+  Array<A> left;
+  Array<B> right;
+  TestOrdering expectedResult;
+  ArrayComparisonTest(std::initializer_list<A> left, std::initializer_list<B> right, TestOrdering expectedResult) :
+    left(heapArray(left)), right(heapArray(right)), expectedResult(expectedResult) {}
+
+  template<size_t N, size_t M>
+  ArrayComparisonTest(A (&left) [N], B(&right) [M], TestOrdering expectedResult) :
+    left(heapArray(left, N)), right(heapArray(right, M)), expectedResult(expectedResult) {}
+
+};
+
+KJ_TEST("ArrayPtr comparators for nullptr type") {
+  verifyEqualityComparisons(ArrayPtr<const int>({}), nullptr, EQUAL);
+  verifyEqualityComparisons(ArrayPtr<const int>({123}), nullptr, GREATER);
+}
+
+KJ_TEST("ArrayPtr comparators for same int type") {
+  using Test = ArrayComparisonTest<const int, const int>;
+  Test testCases[] = {
+    {{1,2}, {1,2}, EQUAL},
+    {{1,2}, {1,3}, LESS},
+    {{1,3}, {1,2}, GREATER},
+    {{1}  , {1,2}, LESS},
+    {{2}  , {1,2}, GREATER},
+    {{257,258}, {257,258}, EQUAL},
+    {{0xFF,0xFF}, {0x101,0xFF}, LESS},
+    {{0xFF,0x101}, {0xFF,0xFF}, GREATER},
+    {{0xFF}  , {0xFF,0x101}, LESS},
+    {{0x101}  , {0xFF,0x101}, GREATER},
+    {{-1,-2}, {-1,-2}, EQUAL},
+    {{-1,-3}, {-1,-2}, LESS},
+    {{-1,-2}, {-1,-3}, GREATER},
+    {{-1}  , {-1,-2}, LESS},
+    {{-1}  , {-2,-3}, GREATER},
+    {{-1,1}, {-1,1}, EQUAL},
+    {{-1,-1}, {-1,1}, LESS},
+    {{-1,1}, {-1,-1}, GREATER},
+    {{-1}  , {1,-2}, LESS},
+    {{1}  , {-1,2}, GREATER},
+  };
+
+  for (auto const& testCase : testCases) {
+    strongComparisonsTests(testCase.left.asPtr(), testCase.right.asPtr(), testCase.expectedResult);
+  }
+}
+KJ_TEST("ArrayPtr comparators for same int type") {
+  using Test = ArrayComparisonTest<const unsigned int, const unsigned int>;
+  Test testCases[] = {
+    {{1,2}, {1,2}, EQUAL},
+    {{1,2}, {1,3}, LESS},
+    {{1,3}, {1,2}, GREATER},
+    {{1}  , {1,2}, LESS},
+    {{2}  , {1,2}, GREATER},
+    {{257,258}, {257,258}, EQUAL},
+    {{0xFF,0xFF}, {0x101,0xFF}, LESS},
+    {{0xFF,0x101}, {0xFF,0xFF}, GREATER},
+    {{0xFF}  , {0xFF,0x101}, LESS},
+    {{0x101}  , {0xFF,0x101}, GREATER},
+    {{0x101}  , {0xFF}, GREATER},
+  };
+
+  for (auto const& testCase : testCases) {
+    strongComparisonsTests(testCase.left.asPtr(), testCase.right.asPtr(), testCase.expectedResult);
+  }
+}
+
+KJ_TEST("ArrayPtr equality comparisons for different int type") {
+  using Test = ArrayComparisonTest<const int, const short>;
+  Test testCases[] = {
+    {{1,2}, {1,2}, EQUAL},
+    {{1,2}, {1,3}, LESS},
+    {{1,3}, {1,2}, GREATER},
+    {{1}  , {1,2}, LESS},
+    {{2}  , {1,2}, GREATER},
+  };
+
+  for (auto const& testCase : testCases) {
+    verifyEqualityComparisons(testCase.left.asPtr(), testCase.right.asPtr(), testCase.expectedResult);
+  }
+}
+
+KJ_TEST("ArrayPtr comparators for doubles (testing partial orderings)") {
+  using Test = ArrayComparisonTest<const double, const double>;
+  const double d = nan();
+  Test testCases[] = {
+    {{0.0}, {0.0}, EQUAL},
+    {{1.0}, {0.0}, NOTEQUAL},
+    {{0.0}, {1.0}, NOTEQUAL},
+    {{0,0, 0.0}, {0.0}, NOTEQUAL},
+    {{0.0, 0.0}, {1.0}, NOTEQUAL},
+    {{d}, {d}, UNORDERED},
+  };
+
+  for (auto const& testCase : testCases) {
+    verifyEqualityComparisons(testCase.left.asPtr(), testCase.right.asPtr(), testCase.expectedResult);
+  }
+}
+
+KJ_TEST("ArrayPtr comparator for arrays of the same string type") {
+  using TestCase = ArrayComparisonTest<const StringPtr, const StringPtr>;
+  TestCase testCases[] = {
+    {{"foo", "bar"}, {"foo", "bar"}, EQUAL},
+    {{"foo", "bar"}, {"foo", "baz"}, LESS},
+    {{"foo", "bar"}, {"foo"       }, GREATER},
+  };
+
+  for (auto const& testCase : testCases) {
+    strongComparisonsTests(testCase.left.asPtr(), testCase.right.asPtr(), testCase.expectedResult);
+  }
+}
+
+KJ_TEST("ArrayPtr equality comparisons for UTF-8") {
+  using TestCase = ArrayComparisonTest<const char, const char>;
+
+  TestCase testCases[] = {
+    {"hello", "żółć", LESS},
+  };
+
+  for (auto const& testCase : testCases) {
+    strongComparisonsTests(testCase.left.asPtr(), testCase.right.asPtr(), testCase.expectedResult);
+    strongComparisonsTests(testCase.left.asBytes(), testCase.right.asBytes(), testCase.expectedResult);
+  }
+}
+
+KJ_TEST("ArrayPtr equality for arrays of different string types") {
+  using Test = ArrayComparisonTest<const StringPtr, const char* const>;
+  Test testCases[] = {
+    {{"foo", "bar"}, {"foo", "bar"}, EQUAL},
+    {{"foo", "bar"}, {"foo", "baz"}, LESS},
+    {{"foo", "bar"}, {"foo"       }, GREATER},
+  };
+
+  for (auto const& testCase : testCases) {
+    verifyEqualityComparisons(testCase.left.asPtr(), testCase.right.asPtr(), testCase.expectedResult);
+  }
+}
+
 
 KJ_TEST("kj::range()") {
   uint expected = 5;
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -1928,6 +1928,30 @@ public:
     return true;
   }
 
+  inline bool operator<(const ArrayPtr& other) const {
+    size_t comparisonSize = kj::min(size_, other.size_);
+    if constexpr (isSameType<RemoveConst<T>, char>() || isSameType<RemoveConst<T>, unsigned char>()) {
+      int ret = memcmp(ptr, other.ptr, comparisonSize * sizeof(T));
+      if (ret != 0) {
+        return ret < 0;
+      }
+    } else {
+      for (size_t i = 0; i < comparisonSize; i++) {
+        bool ret = ptr[i] == other.ptr[i];
+        if (!ret) {
+          return ptr[i] < other.ptr[i];
+        }
+      }
+    }
+    // arrays are equal up to comparisonSize
+    return size_ <  other.size_;
+  }
+
+  inline bool operator<=(const ArrayPtr& other) const { return !(other < *this); }
+  inline bool operator>=(const ArrayPtr& other) const { return other <= *this; }
+  // Note that only strongly ordered types are currently supported
+  inline bool operator> (const ArrayPtr& other) const { return other < *this; }
+
   template <typename... Attachments>
   Array<T> attach(Attachments&&... attachments) const KJ_WARN_UNUSED_RESULT;
   // Like Array<T>::attach(), but also promotes an ArrayPtr to an Array. Generally the attachment
@@ -1940,10 +1964,10 @@ public:
   // Syntax sugar for invoking U::from.
   // Used to chain conversion calls rather than wrap with function.
 
-  inline void fill(T t) { 
+  inline void fill(T t) {
     // Fill the area by copying t over every element.
 
-    for (size_t i = 0; i < size_; i++) { ptr[i] = t; } 
+    for (size_t i = 0; i < size_; i++) { ptr[i] = t; }
     // All modern compilers are smart enough to optimize this loop for simple T's.
     // libc++ std::fill doesn't have memset specialization either.
   }
@@ -1955,7 +1979,7 @@ public:
     KJ_IREQUIRE(!intersects(other), "copy memory area must not overlap");
     T* __restrict__ dst = begin();
     const T* __restrict__ src = other.begin();
-    for (size_t s = size_, i = 0; i < s; i++) { dst[i] = src[i]; } 
+    for (size_t s = size_, i = 0; i < s; i++) { dst[i] = src[i]; }
   }
 
 private:
@@ -2209,7 +2233,7 @@ struct ByteLiteral {
   constexpr size_t size() const { return N - 1; }
   constexpr const kj::byte* begin() const { return data; }
   kj::byte data[N-1]; // do not store 0-terminator
-}; 
+};
 
 template<>
 struct ByteLiteral<1ul> {
--- a/c++/src/kj/string.h
+++ b/c++/src/kj/string.h
@@ -721,15 +721,11 @@ inline constexpr ArrayPtr<const char> StringPtr::asArray() const {
 }
 
 inline bool StringPtr::operator==(const StringPtr& other) const {
-  return content.size() == other.content.size() &&
-      memcmp(content.begin(), other.content.begin(), content.size() - 1) == 0;
+  return content == other.content;
 }
 
-inline bool StringPtr::operator<(const StringPtr& other) const {
-  bool shorter = content.size() < other.content.size();
-  int cmp = memcmp(content.begin(), other.content.begin(),
-                   shorter ? content.size() : other.content.size());
-  return cmp < 0 || (cmp == 0 && shorter);
+inline bool StringPtr::operator< (const StringPtr& other) const {
+  return content < other.content;
 }
 
 inline StringPtr StringPtr::slice(size_t start) const {


c++/src/kj/common-test.c++
```diff
@@ -818,6 +818,188 @@ KJ_TEST("ArrayPtr operator ==") {
   KJ_EXPECT(ArrayPtr<double>(d, 1) != ArrayPtr<double>(d, 1));
 }
 
+KJ_TEST("ArrayPtr operator <=>") {
+  KJ_EXPECT((ArrayPtr<const int>({}) <=> nullptr) == std::strong_ordering::equal);
```
mikea, line 822: wow, this is wall of text. Can we do something better here?

Couple things come to mind:
- local typedefs
- data-driven tests

c++/src/kj/common-test.c++
```diff
@@ -818,6 +818,188 @@ KJ_TEST("ArrayPtr operator ==") {
...
+  double d[1] = { nan() };
+  KJ_EXPECT((ArrayPtr<double>(d, 1) <=> ArrayPtr<double>(d, 1)) == std::partial_ordering::unordered);
+  KJ_EXPECT(!(ArrayPtr<double>(d, 1) == ArrayPtr<double>(d, 1)));
+  KJ_EXPECT((ArrayPtr<double>(d, 1) != ArrayPtr<double>(d, 1)));
+  KJ_EXPECT(!(ArrayPtr<double>(d, 1) <= ArrayPtr<double>(d, 1)));
+  KJ_EXPECT(!(ArrayPtr<double>(d, 1) >= ArrayPtr<double>(d, 1)));
+  KJ_EXPECT(!(ArrayPtr<double>(d, 1) < ArrayPtr<double>(d, 1)));
+  KJ_EXPECT(!(ArrayPtr<double>(d, 1) > ArrayPtr<double>(d, 1)));
+  // operator== should not use memcmp for double elements.
+}
```
mikea, line 1001: let's add test for a structs with custom operators. I think we need to check cases like: struct with <=> only, struct with == only, maybe couple other examples (if any) from our code

c++/src/kj/common.h
```diff
@@ -1928,6 +1929,51 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
     return true;
   }
 
+  inline auto operator<=>(decltype(nullptr)) const {
```
mikea, line 1932: why is it auto? can't it be `std::strong_ordering`?

c++/src/kj/common.h
```diff
@@ -1928,6 +1929,51 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
     return true;
   }
 
+  inline auto operator<=>(decltype(nullptr)) const { 
+    using ReturnType = decltype(ptr[0] <=> ptr[0]);
+    return size_ == 0 ? (ReturnType) std::strong_ordering::equal
+                      : (ReturnType) std::strong_ordering::greater;
+  }
+
+  inline auto operator<=>(const ArrayPtr& other) const {
```
mikea, line 1938: same

c++/src/kj/common.h
```diff
@@ -1928,6 +1929,51 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
...
+    //Lexographic 3-way comparitor function
```
mikea, line 1939: nit: space after //

mikea, line 1939: comparitor => comparator, but I think official name is "three-way comparison operator"

c++/src/kj/common.h
```diff
@@ -1928,6 +1929,51 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
...
 
+  inline auto operator<=>(decltype(nullptr)) const { 
+    using ReturnType = decltype(ptr[0] <=> ptr[0]);
+    return size_ == 0 ? (ReturnType) std::strong_ordering::equal
+                      : (ReturnType) std::strong_ordering::greater;
+  }
+
+  inline auto operator<=>(const ArrayPtr& other) const {
+    //Lexographic 3-way comparitor function 
+    //NOTE: Should not be used to erase equality, as does not short circuit on size
```
mikea, line 1940: who is this note for?

c++/src/kj/common.h
```diff
@@ -1928,6 +1929,51 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
...
+        }
+      }
+    }
+    return size_ == other.size_ ? (ReturnType) std::strong_ordering::equal 
+          : size_ <  other.size_ ? (ReturnType) std::strong_ordering::less 
+          : (ReturnType) std::strong_ordering::greater;
+  }
+
+  template <typename U>
+  inline auto operator<=>(const ArrayPtr<U>& other) const {
```
mikea, line 1963: do we actually use cross-type comparisons?

c++/src/kj/common.h
AdityaAtulTewari: if the internal type is double then it needs to be `partial_ordering`.

c++/src/kj/common.h
AdityaAtulTewari: It's important to specify it is lexographic, otherwise it is ambiguous if the size is checked first.

c++/src/kj/common.h
AdityaAtulTewari: Anyone wanting to naively replace `a == b` with the default from `<=>`.

c++/src/kj/common.h
AdityaAtulTewari: The tests certainly do, I'm not sure if we put them to use anywhere I'll look into it and respond here.

c++/src/kj/common-test.c++
AdityaAtulTewari: is the goal to figure out the minimum generators for operations?

c++/src/kj/common.h
```diff
@@ -102,6 +102,7 @@ KJ_BEGIN_HEADER
 #include <cstring>
 #include <initializer_list>
 #include <string.h>
+#include <compare>
```
kentonv, line 105: Do we really have to include this? This header includes dozens of other headers transitively, and including it here means adding all those file to every single translation unit in our entire codebase.

If this is an absolute requirement to implement the spaceship operator, I would honestly rather implement `operator<` and `operator==` instead (which I believe the compiler can leverage to implement all the other operators).

c++/src/kj/common-test.c++
AdityaAtulTewari: I think the latest updates have taken care of this, I'd love to see if there are ways to make it even more clean.

c++/src/kj/common.h
AdityaAtulTewari: There should now be comments that properly address this.

c++/src/kj/common.h
AdityaAtulTewari: Comments should address this.

c++/src/kj/common.h
AdityaAtulTewari: Comments should explain this better now.

c++/src/kj/common.h
AdityaAtulTewari: Because we are using such a small subset of the compare header, we could write our own version and this would turn each of the comparator functions into clean one liners instead of default. however using explicit implementations for `operator<` and `operator==` to implement `operator<=` would call `memcmp` twice on the minimum size of the arrays being compared which I think is too expensive.

AdityaAtulTewari: Also from exploring the implementation I've seen it only includes about 10 extra headers of which only 4 have not been included anywhere else is that really too many? Most of those headers are just macros for alignment sizes or assignments for template meta-programming objects.

The specific headers I saw were:
- `ordering.h`: the original header included mostly everything is constexpr
- `enable_if.h`: the enable_if meta-programming header everything is constexpr
- `is_same.h`: the is_same meta-programming header everything is constexpr or templated
- `integral_constant.h`: template structs for integral constants.
All of the below are already in `<initializer_list>`
- `__config`: this is a bunch of c++ and specific defines macros
- `__config_site`: A bunch of defines for the LIBCPP abi namespace.
- `features.h`: Some more macros
- `features-time64.h: Macros for time bits
- `timesize.h`: Macros for time sizes
- `wordsize.h`: Macros for word sizes

kentonv: > however using explicit implementations for operator< and operator== to implement operator<= would call memcmp twice on the minimum size of the arrays being compared which I think is too expensive.

No it wouldn't: `x <= y` would be rewritten to `!(y < x)`.

AdityaAtulTewari: This isn't true for doubles right?

AdityaAtulTewari: `nan() < nan()` is false, and `nan() <= `nan()` is also false.

AdityaAtulTewari: What would be a way to measure the impact that that the previous commit versus the current commit:
> Do we really have to include this? This header includes dozens of other headers transitively, and including it here means adding all those file to every single translation unit in our entire codebase.

Is it something like compile time of a large binary?
Is there some kind of test we can add?

c++/src/kj/common.h
```diff
@@ -1928,6 +1928,83 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
...
+  inline bool operator<(const ArrayPtr<U>& other) const {
+    size_t comparisonSize = kj::min(size_, other.size());
+    for(size_t i = 0; i < comparisonSize; i++) {
+      bool ret = ptr[i] == other.begin()[i];
+      if(!ret) {
+        return ptr[i] < other.begin()[i];
+      }
+    }
+    return size_ < other.size();
+  }
```
AdityaAtulTewari, line 1991: Remove this.

c++/src/kj/common-test.c++
```diff
@@ -793,29 +794,163 @@ TEST(Common, ArrayAsBytes) {
...
-  KJ_EXPECT(ArrayPtr<double>(d, 1) != ArrayPtr<double>(d, 1));
+constexpr auto strong_equal = std::strong_ordering::equal;
+constexpr auto strong_less = std::strong_ordering::less;
+constexpr auto strong_greater = std::strong_ordering::greater;
+constexpr auto unordered = std::partial_ordering::unordered;
+constexpr auto partial_equal = std::partial_ordering::equivalent;
+constexpr auto partial_less = std::partial_ordering::less;
+constexpr auto partial_greater = std::partial_ordering::greater;
+
+template<typename A, typename B>
```
mikea, line 805: only one typename needed, right? (here and below)

c++/src/kj/common-test.c++
```diff
@@ -793,29 +794,163 @@ TEST(Common, ArrayAsBytes) {
...
+void allComparisons(A a, B b, bool ans_eq, bool ans_lte, bool ans_gte, bool ans_lt, bool ans_gt) {
```
mikea, line 806: let's call all these `verify*` or `test*`

mikea, line 806: camelCaseNames.

should probably be `expectedEq` etc.

c++/src/kj/common-test.c++
```diff
@@ -793,29 +794,163 @@ TEST(Common, ArrayAsBytes) {
...
+    ans3way == partial_greater);
+}
+
+template<typename A, typename B>
+void partialEqualityTests(A a, B b, std::partial_ordering ans3way){  
+  equalityComparisons<A, B>(a, b, ans3way == partial_equal);
+}
+
+template<typename A, typename B, typename C>
+struct GenericArrayTestCase {
```
mikea, line 860: ArrayComparisonTest

c++/src/kj/common-test.c++
```diff
@@ -793,29 +794,163 @@ TEST(Common, ArrayAsBytes) {
...
+    std::strong_ordering result;
+    TestCase(std::initializer_list<const int> left, nullptr_t right, std::strong_ordering result) :
+      left(heapArray(left)), right(right), result(result) {}
+  };
+  TestCase testCases[] = {
+    {{}, nullptr, strong_equal},
+    {{123}, nullptr, strong_greater},
+  };
+  for(auto const& testCase : testCases) {
+    strongComparisonsTests(ArrayPtr<const int>(testCase.left), testCase.right, testCase.result);
```
mikea, line 881: if right is null, why do you need it? just hardcode `nullptr` here

c++/src/kj/common-test.c++
```diff
@@ -793,29 +794,163 @@ TEST(Common, ArrayAsBytes) {
...
+struct GenericArrayTestCase {
+  Array<A> left;
+  Array<B> right;
+  C result;
+  GenericArrayTestCase(std::initializer_list<A> left, std::initializer_list<B> right, C result) :
+    left(heapArray(left)), right(heapArray(right)), result(result) {}
+};
+
+KJ_TEST("ArrayPtr comparators for nullptr type") {
+  struct TestCase {
```
mikea, line 869: TestCase -> Test?

c++/src/kj/common-test.c++
mikea, line 881: test.left.asPtr()?

c++/src/kj/common.h
```diff
@@ -1928,6 +1928,55 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
     return true;
   }
 
+private:
+  template <bool equals>
```
mikea, line 1932: strictEquals

c++/src/kj/common.h
```diff
@@ -1928,6 +1928,55 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
     return true;
   }
 
+private:
+  template <bool equals> 
+  inline bool lessImpl(const ArrayPtr& other) const {
```
mikea, line 1933: need comment, especially explaining strictEquals.

c++/src/kj/common.h
```diff
@@ -1928,6 +1928,55 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
...
+  inline bool operator< (decltype(nullptr)) const { return false; }
+  inline bool operator> (decltype(nullptr)) const { return size_ > 0; }
+  friend inline bool operator<=(decltype(nullptr), const ArrayPtr& other) { return other >= nullptr; }
+  friend inline bool operator>=(decltype(nullptr), const ArrayPtr& other) { return other <= nullptr; }
+  friend inline bool operator< (decltype(nullptr), const ArrayPtr& other) { return other >  nullptr; }
+  friend inline bool operator> (decltype(nullptr), const ArrayPtr& other) { return other <  nullptr; }
+  inline bool operator>=(const ArrayPtr& other) const { return other <= *this; }
+  inline bool operator> (const ArrayPtr& other) const { return other < *this; }
+  template<typename U>
+  inline bool operator>=(const ArrayPtr<U>& other) const { return other <= *this; }
```
mikea, line 1976: remove U?

c++/src/kj/common-test.c++
AdityaAtulTewari: No multiple are needed to do equality comparisons.

c++/src/kj/common-test.c++
```diff
@@ -793,29 +794,162 @@ TEST(Common, ArrayAsBytes) {
...
-  KJ_EXPECT((ArrayPtr<const StringPtr>({"foo", "bar"}) !=
-             ArrayPtr<const char* const>({"foo"})));
-
-  // operator== should not use memcmp for double elements.
-  double d[1] = { nan() };
-  KJ_EXPECT(ArrayPtr<double>(d, 1) != ArrayPtr<double>(d, 1));
+constexpr auto strongEqual = std::strong_ordering::equal;
+constexpr auto strongLess = std::strong_ordering::less;
+constexpr auto strongGreater = std::strong_ordering::greater;
+constexpr auto unordered = std::partial_ordering::unordered;
```
mikea, line 800: let's clean these up since we don't care about partial comparators?

c++/src/kj/common-test.c++
```diff
@@ -793,29 +794,162 @@ TEST(Common, ArrayAsBytes) {
...
+constexpr auto partialEqual = std::partial_ordering::equivalent;
+constexpr auto partialLess = std::partial_ordering::less;
+constexpr auto partialGreater = std::partial_ordering::greater;
+
+template<typename A, typename B>
+void verifyAllComparisons(A a, B b, bool expectedEq, bool expectedLTE, bool expectedGTE, bool expectedLT, bool expectedGT) {
```
mikea, line 806: I wonder if we need this many parameters? Surely well written strong order comparison has less dimensions?

c++/src/kj/common-test.c++
```diff
@@ -793,29 +794,162 @@ TEST(Common, ArrayAsBytes) {
...
+  KJ_EXPECT((b >=  a) == expectedLTE);  
+  KJ_EXPECT((a <   b) == expectedLT);  
+  KJ_EXPECT((b <   a) == expectedGT);  
+  KJ_EXPECT((a >   b) == expectedGT);  
+  KJ_EXPECT((b >   a) == expectedLT);  
+}
+
+template<typename A, typename B>
+void verifyEqualityComparisons(A a, B b, bool expectedEq){
+  KJ_EXPECT((a ==  b) == expectedEq);
```
mikea, line 823: I think your new lines have two spaces after operators for some reason?

c++/src/kj/common-test.c++
```diff
@@ -793,29 +794,162 @@ TEST(Common, ArrayAsBytes) {
...
-             ArrayPtr<const char* const>({"foo", "bar"})));
-  KJ_EXPECT((ArrayPtr<const StringPtr>({"foo", "bar"}) !=
-             ArrayPtr<const char* const>({"foo", "baz"})));
-  KJ_EXPECT((ArrayPtr<const StringPtr>({"foo", "bar"}) !=
-             ArrayPtr<const char* const>({"foo"})));
-
-  // operator== should not use memcmp for double elements.
-  double d[1] = { nan() };
-  KJ_EXPECT(ArrayPtr<double>(d, 1) != ArrayPtr<double>(d, 1));
+constexpr auto strongEqual = std::strong_ordering::equal;
```
mikea, line 797: since we need these only as enums for tests just declare an enum yourself and drop <compare>

c++/src/kj/common-test.c++
```diff
@@ -793,29 +794,162 @@ TEST(Common, ArrayAsBytes) {
...
+template<typename A, typename B>
+void partialEqualityTests(A a, B b, std::partial_ordering ans3way){  
+  verifyEqualityComparisons<A, B>(a, b, ans3way == partialEqual);
+}
+
+template<typename A, typename B, typename C>
+struct ArrayComparisonTest {
+  Array<A> left;
+  Array<B> right;
+  C result;
```
mikea, line 863: expected_result

c++/src/kj/common-test.c++
```diff
@@ -793,29 +794,162 @@ TEST(Common, ArrayAsBytes) {
...
+  ArrayComparisonTest(std::initializer_list<A> left, std::initializer_list<B> right, C result) :
+    left(heapArray(left)), right(heapArray(right)), result(result) {}
+};
+
+KJ_TEST("ArrayPtr comparators for nullptr type") {
+  struct Test {
+    Array<const int> left;
+    std::strong_ordering result;
```
mikea, line 871: expected_result

c++/src/kj/common-test.c++
```diff
@@ -793,29 +793,152 @@ TEST(Common, ArrayAsBytes) {
...
+  KJ_EXPECT((a !=  b) == !expectedEq);
+  KJ_EXPECT((b !=  a) == !expectedEq);
+}
+
+template<typename A, typename B>
+void strongComparisonsTests(A a, B b, strong_ordering ans3way){
+  verifyStrongComparisons<A, B>(a, b,
+    ans3way == strongEqual,
+    ans3way == strongLess);
+}
```
AdityaAtulTewari, line 838: should include the real implementation, we have too many levels of abstraction here.

c++/src/kj/common-test.c++
```diff
@@ -793,29 +793,152 @@ TEST(Common, ArrayAsBytes) {
...
+  Test testCases[] = {
+    {{"foo", "bar"}, {"foo", "bar"}, strongEqual},
+    {{"foo", "bar"}, {"foo", "baz"}, strongLess},
+    {{"foo", "bar"}, {"foo"       }, strongGreater},
+  };
+
+  runComparisonTests(strongEqualityTests, testCases);
+}
+
+KJ_TEST("ArrayPtr comparator for different string types") {
```
AdityaAtulTewari, line 934: Description needs to be changed.

c++/src/kj/common-test.c++
```diff
@@ -793,30 +793,149 @@ TEST(Common, ArrayAsBytes) {
...
-  // operator== should not use memcmp for double elements.
-  double d[1] = { nan() };
-  KJ_EXPECT(ArrayPtr<double>(d, 1) != ArrayPtr<double>(d, 1));
+enum strong_ordering {
+  strongEqual,
+  strongLess,
+  strongGreater,
+};
+
+enum partial_ordering {
```
mikea, line 802: let's remove partial

c++/src/kj/common-test.c++
```diff
@@ -793,30 +793,149 @@ TEST(Common, ArrayAsBytes) {
...
-             ArrayPtr<const char* const>({"foo", "bar"})));
-  KJ_EXPECT((ArrayPtr<const StringPtr>({"foo", "bar"}) !=
-             ArrayPtr<const char* const>({"foo", "baz"})));
-  KJ_EXPECT((ArrayPtr<const StringPtr>({"foo", "bar"}) !=
-             ArrayPtr<const char* const>({"foo"})));
-
-  // operator== should not use memcmp for double elements.
-  double d[1] = { nan() };
-  KJ_EXPECT(ArrayPtr<double>(d, 1) != ArrayPtr<double>(d, 1));
+enum strong_ordering {
```
mikea, line 796: nit: CamelCaseNaming

c++/src/kj/common-test.c++
```diff
@@ -793,30 +793,149 @@ TEST(Common, ArrayAsBytes) {
...
+  partialEqual,
+  partialLess,
+  partialGreater,
+};
+
+template<typename A, typename B>
+void strongComparisonsTests(A a, B b, strong_ordering ans3way) {
+  const bool expectedEq = ans3way == strongEqual;
+  const bool expectedLT = ans3way == strongLess;
+  KJ_EXPECT((a ==  b) == expectedEq);
```
mikea, line 813: this should reuse verifyEqualityComparisons

c++/src/kj/common-test.c++
```diff
@@ -793,30 +793,149 @@ TEST(Common, ArrayAsBytes) {
...
+};
+
+enum partial_ordering {
+  unordered,
+  partialEqual,
+  partialLess,
+  partialGreater,
+};
+
+template<typename A, typename B>
```
mikea, line 809: A is always equal B, so let's have X everywhere.

c++/src/kj/common-test.c++
```diff
@@ -793,30 +793,149 @@ TEST(Common, ArrayAsBytes) {
...
+struct ArrayComparisonTest {
+  Array<A> left;
+  Array<B> right;
+  C expectedResult;
+  ArrayComparisonTest(std::initializer_list<A> left, std::initializer_list<B> right, C expectedResult) :
+    left(heapArray(left)), right(heapArray(right)), expectedResult(expectedResult) {}
+
+};
+
+#define runComparisonTests(f, tests) \
```
mikea, line 855: let's inline this instead.

c++/src/kj/common-test.c++
```diff
@@ -793,30 +793,149 @@ TEST(Common, ArrayAsBytes) {
...
+void strongEqualityTests(A a, B b, strong_ordering ans3way){
+  verifyEqualityComparisons<A, B>(a, b, ans3way == strongEqual);
+}
+
+template<typename A, typename B>
+void partialEqualityTests(A a, B b, partial_ordering ans3way){
+  verifyEqualityComparisons<A, B>(a, b, ans3way == partialEqual);
+}
+
+template<typename A, typename B, typename C>
```
mikea, line 845: typename C is not needed either, right?

c++/src/kj/common-test.c++
AdityaAtulTewari: That's not the case for nullptr comparisons, so I'll leave it as is.

c++/src/kj/common-test.c++
```diff
@@ -793,31 +793,144 @@ TEST(Common, ArrayAsBytes) {
...
-             ArrayPtr<const char* const>({"foo", "bar"})));
-  KJ_EXPECT((ArrayPtr<const StringPtr>({"foo", "bar"}) !=
-             ArrayPtr<const char* const>({"foo", "baz"})));
-  KJ_EXPECT((ArrayPtr<const StringPtr>({"foo", "bar"}) !=
-             ArrayPtr<const char* const>({"foo"})));
-
-  // operator== should not use memcmp for double elements.
-  double d[1] = { nan() };
-  KJ_EXPECT(ArrayPtr<double>(d, 1) != ArrayPtr<double>(d, 1));
+enum testOrdering {
```
mikea, line 796: TestOrdering

c++/src/kj/common-test.c++
```diff
@@ -793,31 +793,144 @@ TEST(Common, ArrayAsBytes) {
...
+  unordered,
+  equal,
```
mikea, line 798: the style for enum values is ALL_CAPS

c++/src/kj/string.h
```diff
@@ -133,11 +133,15 @@ class StringPtr {
...
 
   inline constexpr bool operator==(decltype(nullptr)) const { return content.size() <= 1; }
+  inline constexpr bool operator<=(decltype(nullptr)) const { return *this == nullptr; }
+  inline constexpr bool operator>=(decltype(nullptr)) const { return true; }
+  inline constexpr bool operator< (decltype(nullptr)) const { return false; }
+  inline constexpr bool operator> (decltype(nullptr)) const { return content.size() > 1; }
 
-  inline bool operator==(const StringPtr& other) const;
-  inline bool operator< (const StringPtr& other) const;
+  inline bool operator==(const StringPtr& other) const { return content.first(content.size() - 1) == other.content.first(other.content.size() - 1); }
```
mikea, line 141: why -1 here but not below? If this is expected, then it needs careful commenting.

mikea, line 141: you're delegating comparison to ArrayPtr, right? Can you use asBytes() or something?

c++/src/kj/string.h
```diff
@@ -133,11 +133,15 @@ class StringPtr {
...
+  inline bool operator< (const StringPtr& other) const { return content.first(content.size()) < other.content.first(other.content.size()); }
   inline bool operator> (const StringPtr& other) const { return other < *this; }
-  inline bool operator<=(const StringPtr& other) const { return !(other < *this); }
+  inline bool operator<=(const StringPtr& other) const { return !(*this > other); }
```
mikea, line 144: why the flip?

c++/src/kj/common.h
```diff
@@ -1928,6 +1928,38 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
...
+        bool ret = ptr[i] == other.ptr[i];
+        if(!ret) {
+          return ptr[i] < other.ptr[i];
+        }
+      }
+    }
+    return size_ <  other.size_;
+  }
+
+  inline bool operator<=(decltype(nullptr)) const { return size_ == 0; }
```
mikea, line 1950: I wonder if all these additional `nullptr` declarations are really needed? nullptr is convertible to ArrayPtr, no?

c++/src/kj/common.h
```diff
@@ -1928,6 +1928,38 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
...
+  inline bool operator>=(decltype(nullptr)) const { return true; }
+  inline bool operator< (decltype(nullptr)) const { return false; }
+  inline bool operator> (decltype(nullptr)) const { return size_ > 0; }
+  friend inline bool operator<=(decltype(nullptr), const ArrayPtr& other) { return other >= nullptr; }
+  friend inline bool operator>=(decltype(nullptr), const ArrayPtr& other) { return other <= nullptr; }
+  friend inline bool operator< (decltype(nullptr), const ArrayPtr& other) { return other >  nullptr; }
+  friend inline bool operator> (decltype(nullptr), const ArrayPtr& other) { return other <  nullptr; }
+  inline bool operator<=(const ArrayPtr& other) const { return !(other < *this); }
```
mikea, line 1958: let's separate null/not-null case with new line

c++/src/kj/common-test.c++
mikea: Why? Because of nullptr? Isn't nullptr convertible to arrayptr? I also wonder if we really need comparison to nullptr rather than another ArrayPtr?

c++/src/kj/common-test.c++
```diff
@@ -793,31 +793,144 @@ TEST(Common, ArrayAsBytes) {
...
+  less,
+  greater,
+  notEqual,
+};
+
+template<typename A, typename B>
+void verifyEqualityComparisons(A a, B b, bool expectedEq) {
```
mikea, line 805: looks like this function is used only once in a one-liner below? let's inline

c++/src/kj/common-test.c++
AdityaAtulTewari: no because of things like const StringPtr vs const char* const.

c++/src/kj/string.h
AdityaAtulTewari: Why? either way it invokes the same memcmp? is there some benefit for using asBytes?

AdityaAtulTewari: We decided to only prserve legacy behavior in this pr, even though we both agreed that this behavior is wrong.

c++/src/kj/common-test.c++
```diff
@@ -793,31 +793,140 @@ TEST(Common, ArrayAsBytes) {
...
+  EQUAL,
+  LESS,
+  GREATER,
+  NOTEQUAL,
+};
+
+template<typename A, typename B>
+void verifyEqualityComparisons(A a, B b, TestOrdering ord) {
+  const bool expectedEq = ord == EQUAL;
+  KJ_EXPECT((a ==  b) == expectedEq);
```
mikea, line 807: nit: double space

c++/src/kj/common-test.c++
```diff
@@ -793,31 +793,140 @@ TEST(Common, ArrayAsBytes) {
...
+  KJ_EXPECT((a !=  b) == !expectedEq);
+  KJ_EXPECT((b !=  a) == !expectedEq);
+}
+
+template<typename T>
+void strongComparisonsTests(T a, T b, TestOrdering ans3way) {
+  const bool expectedEq = ans3way == EQUAL;
+  const bool expectedLT = ans3way == LESS;
+  verifyEqualityComparisons(a, b, ans3way);
+  KJ_EXPECT((a <=  b) == (expectedEq || expectedLT));
```
mikea, line 818: nit: more spaces than needed

c++/src/kj/common-test.c++
```diff
@@ -793,31 +793,140 @@ TEST(Common, ArrayAsBytes) {
...
+void verifyEqualityComparisons(A a, B b, TestOrdering ord) {
+  const bool expectedEq = ord == EQUAL;
+  KJ_EXPECT((a ==  b) == expectedEq);
+  KJ_EXPECT((b ==  a) == expectedEq);
+  KJ_EXPECT((a !=  b) == !expectedEq);
+  KJ_EXPECT((b !=  a) == !expectedEq);
+}
+
+template<typename T>
+void strongComparisonsTests(T a, T b, TestOrdering ans3way) {
```
mikea, line 814: ans3way => ordering or ord like above.

c++/src/kj/common-test.c++
```diff
@@ -793,31 +793,140 @@ TEST(Common, ArrayAsBytes) {
...
+    TestOrdering expectedResult;
+    Test(std::initializer_list<const int> left, TestOrdering expectedResult) :
+      left(heapArray(left)), expectedResult(expectedResult) {}
+  };
+  Test testCases[] = {
+    {{}, EQUAL},
+    {{123}, GREATER},
+  };
+
+  for(auto const& testCase : testCases) {
```
mikea, line 850: well, you have only tests here, so I wonder if you just eliminate test struct here and write couple verifies manually.

c++/src/kj/string.h
mikea: Let's open a seprate PR with a test and a fix. Looks to me current behavior is very broken.

mikea (commented):
Let's cleanup PR subject and description

c++/src/kj/string.h
AdityaAtulTewari: I added a comment, I think the behavior is actually right. Basically, `StringPtr` is a `NUL` terminated slice. Thus, the last character in the `content` array does not need to be compared.
However, the same length should not be used for operator< because it enables a faster short circuit.

c++/src/kj/string.h
```diff
@@ -134,8 +134,10 @@ class StringPtr {
 
   inline constexpr bool operator==(decltype(nullptr)) const { return content.size() <= 1; }
 
-  inline bool operator==(const StringPtr& other) const;
-  inline bool operator< (const StringPtr& other) const;
+  inline bool operator==(const StringPtr& other) const { return content.first(content.size() - 1) == other.content.first(other.content.size() - 1); }
+  //The -1 is to ignore the null character
```
kentonv, line 138: This is an implementation comment, but you've placed it where a doc comment belongs. In theory, if we were auto-generating docs or something, the `operator==()` would end up documented as "The -1 is to ignore the null character", which doesn't make sense.

If you want to keep this comment, I think you should change `operator==` back to being defined out-of-line later in the file, as it was before this PR. You can then put this implementation comment inside the definition, where it won't be interpreted as documentation.

Same goes for `operator<` below.

c++/src/kj/string.h
kentonv: > Looks to me current behavior is very broken.

How so? In both the cases of `operator==` and `operator<`, including or not including the `-1` makes no difference to the final outcome, because the last character is always NUL.

> enables a faster short circuit.

TBH I doubt this matters in practice.

c++/src/kj/string.h
AdityaAtulTewari: changed, thank you for the advice.

c++/src/kj/string.h
AdityaAtulTewari: Originally I thought the StringPtr could be sliced without being NUL terminated and that's where my confusion stemmed from.

c++/src/kj/string.h
```diff
@@ -721,15 +721,12 @@ inline constexpr ArrayPtr<const char> StringPtr::asArray() const {
...
+  return content.first(content.size() - 1) == other.content.first(other.content.size() - 1);
+  //The -1 is to ignore the NUL character
 }
-
-inline bool StringPtr::operator<(const StringPtr& other) const {
-  bool shorter = content.size() < other.content.size();
-  int cmp = memcmp(content.begin(), other.content.begin(),
-                   shorter ? content.size() : other.content.size());
-  return cmp < 0 || (cmp == 0 && shorter);
+inline bool StringPtr::operator< (const StringPtr& other) const {
```
mikea, line 727: nit: empty line between function

c++/src/kj/string.h
```diff
@@ -721,15 +721,12 @@ inline constexpr ArrayPtr<const char> StringPtr::asArray() const {
...
+  return content < other.content;
+  //This does not include the minus one as operator== so that it short circuts on the '\0' char in the memcmp
```
mikea, line 729: nit: space after //

c++/src/kj/string.h
```diff
@@ -721,15 +721,12 @@ inline constexpr ArrayPtr<const char> StringPtr::asArray() const {
 }
 
 inline bool StringPtr::operator==(const StringPtr& other) const {
-  return content.size() == other.content.size() &&
-      memcmp(content.begin(), other.content.begin(), content.size() - 1) == 0;
+  return content.first(content.size() - 1) == other.content.first(other.content.size() - 1);
+  //The -1 is to ignore the NUL character
```
mikea, line 725: nit: when you're commenting the line, comment goes _before_ it (I know...)

mikea (approved):
LG. @kentonv do you mind taking a final look?

c++/src/kj/string.h
```diff
@@ -721,15 +721,13 @@ inline constexpr ArrayPtr<const char> StringPtr::asArray() const {
 }
 
 inline bool StringPtr::operator==(const StringPtr& other) const {
-  return content.size() == other.content.size() &&
-      memcmp(content.begin(), other.content.begin(), content.size() - 1) == 0;
+  // The -1 is to ignore the NUL character
+  return content.first(content.size() - 1) == other.content.first(other.content.size() - 1);
```
kentonv, line 725: Kind of wonder if it's better to just do `content == other.content` here, the -1s might cost more time than they save?

c++/src/kj/common.h
```diff
@@ -1928,6 +1928,30 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
...
 
+  inline bool operator<(const ArrayPtr& other) const {
+    size_t comparisonSize = kj::min(size_, other.size_);
+    if constexpr (isIntegral<RemoveConst<T>>()) {
+      int ret = memcmp(ptr, other.ptr, comparisonSize * sizeof(T));
+      if(ret != 0){
+        return ret < 0;
+      }
+    }
+    else {
```
kentonv, line 1939: Please use consistent brace style. Should be `} else {`.

c++/src/kj/common.h
```diff
@@ -1928,6 +1928,30 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
     return true;
   }
 
+  inline bool operator<(const ArrayPtr& other) const {
+    size_t comparisonSize = kj::min(size_, other.size_);
+    if constexpr (isIntegral<RemoveConst<T>>()) {
```
kentonv, line 1933: For multi-byte integers on little-endian architectures, `memcmp()` will not produce the same result the for loop, so we have to check that the size is 1:

```suggestion
    if constexpr (isIntegral<RemoveConst<T>>() && sizeof(T) == 1) {
```

But actually, even that might be wrong if the elements are signed, as memcmp() will treat them as unsigned.

But `char` (with no qualifiers) is signed on some systems and not on others, but presumably "correct" sorting of strings should always treat it as unsigned?

So maybe we actually want to use the `memcmp()` approach specifically for `char` and `unsigned char`, but not any other type?

kentonv: Please also add tests comparing arrays of `int`, covering negative values and values larger than 256.

kentonv: `strcmp()` is explicitly documented (on the man page) as presuming unsigned characters, which seems desirable.

It's sort of weird if we say comparing ArrayPtr<char> will not produce the same results as a loop comparing individual values, on platforms where char is signed. However, I think for all real-world purposes, this is what people will want. If they actually want to compare signed chars they can use `ArrayPtr<signed char>`.

kentonv: (That is to say, for `ArrayPtr<char>` or `ArrayPtr<unsigned char>`, using `memcmp()` seems like the right thing. But for `signed char` or multi-byte integers it is still wrong.)

AdityaAtulTewari: - Is our goal then to support little endian architectures?
  - If so, I think it makes sense to add your change.
Should we specialize to `strcmp` for `signed char`?
  - This is something that would be known at compile time, so it makes sense to me to do this if we truly want to specialize.
[ ] Add test comparing arrays of ints including negative values and values larger than 256.

c++/src/kj/string.h
AdityaAtulTewari: Should I try to measure this?

c++/src/kj/common.h
kentonv: > Is our goal then to support little endian architectures?

All modern architectures are little-endian.

> If so, I think we should detect the target's endianness at compile-time right?

Not in this case. While technically memcmp() would work correctly for unsigned integers on big-endian architectures, there are no big-endian architectures anymore, so writing such code would be pointless. We should just not use memcmp() for multi-byte integers.

AdityaAtulTewari: > All modern architectures are little-endian.

This was the missing link for me.

AdityaAtulTewari: Should be fixed now.

c++/src/kj/common.h
```diff
@@ -1928,6 +1936,30 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
     return true;
   }
 
+  inline bool operator<(const ArrayPtr& other) const {
+    size_t comparisonSize = kj::min(size_, other.size_);
+    if constexpr (isIntegral<RemoveConst<T>>() && isUnsigned<RemoveConst<T>>() && sizeof(T) == 1) {
```
kentonv, line 1941: I think this won't actually match `char` as written. It'll only match `unsigned char`. But as I said before we should treat `char` as unsigned for comparison purposes. (Only `signed char` is actually signed.)

I think the best answer is probably just:

```suggestion
    if constexpr (isSameType<T, char>() || isSameType<T, unsigned char>()) {
```

Please be sure to add a test of comparisons on `Array<char>` with UTF-8 text containing non-ASCII characters. If `char` is incorrectly being treated as signed then you'll see non-ASCII characters sorting before ASCII, which is wrong, they should sort after.

AdityaAtulTewari: This should be fixed now, I also had to use `RemoveCOnst` and `byte`.

c++/src/kj/string.h
kentonv: I think you should just change it and not worry about measuring. The speed difference is likely negligible but it's nice to reduce code complexity.

kentonv: Especially given that `operator<` doesn't use the -1. They might as well be consistent.

c++/src/kj/string.h
```diff
@@ -721,15 +721,13 @@ inline constexpr ArrayPtr<const char> StringPtr::asArray() const {
...
 }
 
-inline bool StringPtr::operator<(const StringPtr& other) const {
-  bool shorter = content.size() < other.content.size();
-  int cmp = memcmp(content.begin(), other.content.begin(),
-                   shorter ? content.size() : other.content.size());
-  return cmp < 0 || (cmp == 0 && shorter);
+inline bool StringPtr::operator< (const StringPtr& other) const {
+  // This does not include the minus one as operator== so that it short circuts on the '\0' char in the memcmp
```
kentonv, line 729: I doubt this makes a difference, and in any case it wasn't my intent in the original code, so I'd remove the comment. Otherwise it implies that someone has actually measured this and found this to be an important optimization which is misleading.

AdityaAtulTewari: Comment removed.

kentonv: There's some very simple benchmarking built into kj-test, see `TestCase::doBenchmark()`.

kentonv (approved):
I think this is OK now but will leave to @mikea to do final review & merge.

c++/src/kj/common.h
```diff
@@ -1928,6 +1928,30 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
     return true;
   }
 
+  inline bool operator<(const ArrayPtr& other) const {
+    size_t comparisonSize = kj::min(size_, other.size_);
+    if constexpr (isSameType<RemoveConst<T>, char>() || isSameType<RemoveConst<T>, unsigned char>()) {
+      // Something different has to be done for big endian architectures in this if constexpr
```
mikea, line 1934: nit: comment is not relevant?

c++/src/kj/common.h
```diff
@@ -1928,6 +1928,30 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
...
+      int ret = memcmp(ptr, other.ptr, comparisonSize * sizeof(T));
+      if(ret != 0){
```
mikea, line 1936: nit: spaces

c++/src/kj/common.h
```diff
@@ -1928,6 +1928,30 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
...
+      }
+    } else {
+      for(size_t i = 0; i < comparisonSize; i++) {
+        bool ret = ptr[i] == other.ptr[i];
+        if(!ret) {
+          return ptr[i] < other.ptr[i];
+        }
+      }
+    }
+    return size_ <  other.size_;
```
mikea, line 1947: add a comment `// arrays are equal up to comparisonSize`

c++/src/kj/common-test.c++
```diff
@@ -793,31 +793,179 @@ TEST(Common, ArrayAsBytes) {
...
+    {"hello", "żółć", LESS},
+  };
+
+  for(auto const& testCase : testCases) {
+    strongComparisonsTests(testCase.left.asPtr(), testCase.right.asPtr(), testCase.expectedResult);
+    strongComparisonsTests(testCase.left.asBytes(), testCase.right.asBytes(), testCase.expectedResult);
+  }
+}
+
+KJ_TEST("ArrayPtr equality for different string types") {
```
mikea, line 955: nit:StringPtr?

c++/src/kj/common-test.c++
```diff
@@ -793,31 +793,179 @@ TEST(Common, ArrayAsBytes) {
...
+    {{0.0, 0.0}, {1.0}, NOTEQUAL},
+    {{d}, {d}, UNORDERED},
+  };
+
+  for(auto const& testCase : testCases) {
+    verifyEqualityComparisons(testCase.left.asPtr(), testCase.right.asPtr(), testCase.expectedResult);
+  }
 }
 
+KJ_TEST("ArrayPtr comparator for same string types") {
```
mikea, line 929: nit:StringPtr

c++/src/kj/common.h
```diff
@@ -1928,6 +1928,30 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
...
+  inline bool operator<(const ArrayPtr& other) const {
+    size_t comparisonSize = kj::min(size_, other.size_);
+    if constexpr (isSameType<RemoveConst<T>, char>() || isSameType<RemoveConst<T>, unsigned char>()) {
+      // Something different has to be done for big endian architectures in this if constexpr
+      int ret = memcmp(ptr, other.ptr, comparisonSize * sizeof(T));
+      if(ret != 0){
+        return ret < 0;
+      }
+    } else {
+      for(size_t i = 0; i < comparisonSize; i++) {
```
AdityaAtulTewari, line 1940: nit: space.

c++/src/kj/common.h
```diff
@@ -1928,6 +1928,30 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
...
+        bool ret = ptr[i] == other.ptr[i];
+        if(!ret) {
```
AdityaAtulTewari, line 1942: nit: space.

PR #2064, opened 2024-05-29
[nfc] ubsan CI configuration

mikea:


fhanau (approved):
Change LGTM. Not sure why there's builds being cancelled – looks like there's some sort of ASAN failure before it happens though

PR #2066, opened 2024-06-04
Allow reserved address ranges to be explicitly allowlisted in network filters.

kentonv:
Some people use these address ranges -- especially 240.0.0.0/4 -- as private networks. Is this wise? I dunno. Here's a blog post with extended discussion:

https://blog.benjojo.co.uk/post/class-e-addresses-in-the-real-world

KJ previously totally disallowed use of these ranges. With this change, they are now allowed, but not considered part of any of the named groups (public/private/network/local) -- you have to allow them explicitly.

Note that this also means that these addresses will now be allowed by the default network when no filter has been applied. I think this is OK because if you haven't appplied a filter at all you presumably trust your addresses.

### DIFF
--- a/c++/src/kj/async-io-test.c++
+++ b/c++/src/kj/async-io-test.c++
@@ -1154,7 +1154,7 @@ KJ_TEST("NetworkFilter") {
   _::NetworkFilter base;
 
   KJ_EXPECT(allowed4(base, "8.8.8.8"));
-  KJ_EXPECT(!allowed4(base, "240.1.2.3"));
+  KJ_EXPECT(allowed4(base, "240.1.2.3"));
 
   {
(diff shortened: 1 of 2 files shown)


kentonv:
Internal ticket: EW-8489

irvinebroque:
🙌

mikea (approved):
sorry, missed this

irvinebroque:
We still need to bring this into workerd, yes? (so that it can go out in a workerd release and then in Wrangler)

kentonv:
Sorry, thought that someone was regularly updating workerd deps.

Made a PR on workerd: https://github.com/cloudflare/workerd/pull/2280

PR #2071, opened 2024-06-06
cidr.c++ is used and tested a lot

AdityaAtulTewari:
Line 114 is tested in async-io-test.c++ with different choices of `CidrRange`.
Line 140 is tested in async-io-test.c++ with different choices of `CidrRange` to match.

Do we need to test these more?


mikea (commented):
LG but let's fix the title

mikea (approved):

kentonv:
Please update the commit description too, not just the PR title?

PR #2079, opened 2024-06-16
Rework promise continuations to reduce amount of templated functions, code size

fhanau:
**Rationale**
This PR introduces a series of patches related to promises that are intended to reduce code size and compile times.
Investigating the slow compile time of `kj/compat/http.c++` led me to look at the generated object file – it turns out that it defines ~15,000 symbols in ~8,000 lines on Darwin when optimization/inlining is disabled (< 5,000 symbols with -O1 enabled). This is believed to be a major driver of its slow compile time and code size. Looking at the generated object symbol names with [bloaty](https://github.com/google/bloaty) showed that many symbols are functions of the templated `TransformPromiseNode` class in `kj/promise.h`. This class is used in promise continuations (among other things), causing dozens of templated functions to be defined with every `then()` call.

**Implementation**
This PR includes several changes to reduce the number of templated functions being generated. This is divided into several self-contained patches to make it easier to understand, and in case we decide to only land a subset of changes. I plan to squash the individual commits before merging Note that I have no background in implementing async library code – this may include trivial mistakes. Open to suggestions on how to make this easier to maintain considering the duplication of `TransformPromiseNode` and `then()`.

**Results**
Compile time and code size are measured on the bazel kj-http library target since it uses promise continuations extensively. The number of symbols is for http.c++ alone.
Number of symbols (O0): 15243 => 14580
Code size (O0): 6.16MB  => 5.78MB
Code size (O1): 1.68MB  => 1.57MB
Code size (O1 + debug): 16.71MB  => 14.93MB

Compile time is a bit harder to quantify, but I observed an improvement from 5.69s => 5.25s user time on http.c++ (best of 3 runs, measured on M1 Mac).

**Discussion**
Compile time and code size are improved for kj-http, even if inlining is enabled and especially when debug information is present. Smaller gains are expected in downstream projects too based on promise continuation usage.
Similar improvements are expected to be possible by rewriting `kj::Own`, which also leads to many templated functions being generated. However, this is unlikely to be worth prioritizing; for now this represents a proof-of-concept for reducing the amount of templated functions.

### DIFF
--- a/c++/src/kj/async-inl.h
+++ b/c++/src/kj/async-inl.h
@@ -1482,12 +1482,12 @@ inline Promise<T> constPromise() {
 
 template <typename Func>
 inline PromiseForResult<Func, void> evalLater(Func&& func) {
-  return _::yield().then(kj::fwd<Func>(func), _::PropagateException());
+  return _::yield().then(kj::fwd<Func>(func));
 }
 
 template <typename Func>
 inline PromiseForResult<Func, void> evalLast(Func&& func) {
-  return _::yieldHarder().then(kj::fwd<Func>(func), _::PropagateException());
+  return _::yieldHarder().then(kj::fwd<Func>(func));
 }
 
 template <typename Func>
--- a/c++/src/kj/async-inl.h
+++ b/c++/src/kj/async-inl.h
@@ -367,9 +367,9 @@ public:
     } else {
       // Start a new arena.
       //
-      // NOTE: As in append() (below), we don't implement exception-safety because it causes code
-      //   bloat and these constructors probably don't throw. Instead this function is noexcept, so
-      //   if a constructor does throw, it'll crash rather than leak memory.
+      // NOTE: As in appendPromise() (below), we don't implement exception-safety because it causes
+      //   code bloat and these constructors probably don't throw. Instead this function is
+      //   noexcept, so if a constructor does throw, it'll crash rather than leak memory.
       auto* arena = new PromiseArena;
       ptr = reinterpret_cast<T*>(arena + 1) - 1;
       ctor(*ptr, kj::fwd<Params>(params)...);
@@ -382,9 +382,15 @@ public:
   }
 
   template <typename T, typename D = PromiseDisposer, typename... Params>
-  static kj::Own<T, D> append(
+  static kj::Own<T, D> appendPromise(
       OwnPromiseNode&& next, Params&&... params) noexcept {
-    // Implements appendPromise().
+    // Append a promise to the arena that currently ends with `next`. `next` is also still passed as
+    // the first parameter to the new object's constructor.
+    //
+    // This is semantically the same as `allocPromise()` except that it may avoid the underlying
+    // memory allocation. `next` must end up being destroyed before the new object (i.e. the new
+    // object must never transfer away ownership of `next`).
+
 
     PromiseArena* arena = next->arena;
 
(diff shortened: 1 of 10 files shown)


c++/src/kj/async-inl.h
```diff
@@ -719,34 +722,52 @@ class TransformPromiseNode final: public TransformPromiseNodeBase {
...
-    // by the dependency.
-    dropDependency();
-  }
-
 private:
   Func func;
   ErrorFunc errorHandler;
 
   void getImpl(ExceptionOrValue& output) override {
+    typedef _::FixVoid<_::ReturnType<Func, _DepT>> T;
```
fhanau, line 730: Deriving T instead of providing it as a template parameter feels clunky now that I look at it again, open to dropping this.

c++/src/kj/async-inl.h
```diff
@@ -384,7 +384,13 @@ class PromiseDisposer {
   template <typename T, typename D = PromiseDisposer, typename... Params>
   static kj::Own<T, D> append(
```
fhanau, line 385: - [ ] Rename this to appendPromise().

c++/src/kj/async-inl.h
```diff
@@ -1476,6 +1476,10 @@ inline PromiseForResult<Func, void> evalLater(Func&& func) {
   return _::yield().then(kj::fwd<Func>(func));
 }
 
+inline Promise<void> evalLaterSimple() {
```
kentonv, line 1486: I think `evalLater` no longer makes sense as a name if we're not passing anything to actually evaluate, and the meaning of "simple" isn't very clear either.

How about we actually name this `yield()`? So we're just making the existing function public?

We should also give `evalLast()` the same treatment I think. Its private implementation is currently called `yieldHarder()`, but that's obviously not a good public name. Maybe `yieldUntilQueueEmpty()`.

kentonv:
LGTM except for name bikeshedding.

c++/src/kj/async-inl.h
```diff
@@ -1434,14 +1480,17 @@ inline Promise<T> constPromise() {
   return _::PromiseNode::to<Promise<T>>(_::OwnPromiseNode(&NODE));
 }
 
+Promise<void> yield();
+Promise<void> yieldHarder();
```
kentonv, line 1484: Per my other comment I don't think this is the right name for a public API. Proposed `yieldUntilQueueEmpty()` as an alternative.

Also, these declarations should go in async.h and have doc comments, now that they are public.

fhanau: Sorry, I forgot about the renaming. Fixed that, moved the functions and added some brief doc comments (open to suggestions on those).
Turns out there already was a TODO to make yield() public, resolved that too.

c++/src/kj/async-inl.h
fhanau: Added comments explaining this for now, getting rid of the template parameter appears to contribute to the file size improvement significantly.

kentonv (approved):
There are quite a lot of `evalLater([]() {})` and `evalLast([]() {})` in this repo and our other repos which we should clean up with this.

fhanau:
> There are quite a lot of `evalLater([]() {})` and `evalLast([]() {})` in this repo and our other repos which we should clean up with this.

The affected calls should already be updated with the last commit in here. I'll make sure to do the same with the PRs for updating the downstream repos.

kentonv:
> The affected calls should already be updated with the last commit in here. I'll make sure to do the same with the PRs for updating the downstream repos.

I still see lots of matches in this repo:

    git grep 'eval\(Last\|Later\)(\[\]() *{})'

fhanau:
> > The affected calls should already be updated with the last commit in here. I'll make sure to do the same with the PRs for updating the downstream repos.
> 
> I still see lots of matches in this repo:
> 
> ```
> git grep 'eval\(Last\|Later\)(\[\]() *{})'
> ```

You're right. I was searching with ripgrep which excludes ignored files by default and somehow interpreted the gitignore file to not search the c++/capnp folder at all. I'll make a cleanup PR tomorrow (and think about how to grep better 😅).

PR #2081, opened 2024-06-18
 expose segments table building to support different serialization mechanisms

mikea:


c++/src/capnp/serialize.h
```diff
@@ -119,6 +119,10 @@ size_t expectedSizeInWordsFromPrefix(kj::ArrayPtr<const word> messagePrefix);
 // using the full KJ async infrastructure would be too difficult. Each time bytes are received,
 // use this function to determine if an entire message is ready to be parsed.
 
+kj::Array<kj::byte> segmentsTable(kj::ArrayPtr<const kj::ArrayPtr<const word>> segments);
```
kentonv, line 122: I think this should return `Array<word>`. Caller can always use `.asBytes()` on it if desired, but returning `Array<word>` promises that it'll be properly aligned, which in helpful.

c++/src/capnp/serialize.h
```diff
@@ -119,6 +119,10 @@ size_t expectedSizeInWordsFromPrefix(kj::ArrayPtr<const word> messagePrefix);
...
+// Returns the segments table for given message segments.
+// Fully serialized message consists of the table and segments written consequently.
```
kentonv, line 124:
```suggestion
// Fully serialized message consists of the table and segments written consecutively. 
```

c++/src/capnp/serialize.h
kentonv, line 122: Maybe `serializeSegmentTable()`?

c++/src/capnp/serialize-test.c++
```diff
@@ -476,6 +476,15 @@ TEST(Serialize, RejectHugeMessage) {
 }
 #endif  // !__MINGW32__
 
+TEST(Serialize, SegmentsTable) {
+  TestMessageBuilder builder(1);
+  initTestMessage(builder.initRoot<TestAllTypes>());
+
+  kj::Array<byte> table = segmentsTable(builder.getSegmentsForOutput());
+  byte expected[] = { 0, 0, 0, 0, 95, 1, 0, 0 };
```
kentonv, line 484: These bytes could change if `TestAllTypes` grows new fields.

Perhaps a more robust approach would be to serialize the message the normal way and verify that it leads to the same bytes as serializing the segment table and concatenating with the segments.

c++/src/kj/array.h
```diff
@@ -208,11 +208,11 @@ class Array {
 
   inline Array<PropagateConst<T, byte>> releaseAsBytes() {
     // Like asBytes() but transfers ownership.
-    static_assert(sizeof(T) == sizeof(byte),
-        "releaseAsBytes() only possible on arrays with byte-size elements (e.g. chars).");
+    static_assert(KJ_HAS_TRIVIAL_CONSTRUCTOR(T),
```
kentonv, line 211: The `ArrayDisposer` expects to have the element size and count passed back to it, but if the original array had different-sized elements, then these values will end up inconsistent when passed to the disposer. That's why the check here was checking for exact size match.

Your version works correctly with `HeapArrayDisposer` because `HeapArrayDisposer` doesn't actually care about the element size, it just passes the pointer to `operator delete[]`. But in general this violates the contract of `ArrayDisposer`.

Let's just have the new function return `Array<word>` and avoid making this change.

c++/src/capnp/serialize.h
mikea: Done. I think I have the math right?

kentonv (approved):

c++/src/capnp/serialize.c++
```diff
@@ -272,6 +272,26 @@ void readMessageCopy(kj::InputStream& input, MessageBuilder& target,
...
+  // We write the segment count - 1 because this makes the first word zero for single-segment
+  // messages, improving compression.  We don't bother doing this with segment sizes because
+  // one-word segments are rare anyway.
+  table[0].set(segments.size() - 1);
+  for (uint i = 0; i < segments.size(); i++) {
+    table[i + 1].set(segments[i].size());
+  }
+  if (segments.size() % 2 == 0) {
+    // Set padding byte.
+    table[segments.size() + 1].set(0);
```
ObsidianMinor, line 290: I don't think this is necessary. The `heapArray` will make all the words null right? So these 4 bytes of padding will be null already, right? If it doesn't, you can make it branchless by setting the last element in the word table 0 at the start.

jclee: I think it depends on the behavior of the allocator?  It looks like:
- `kj::heapArray<word>()`
- calls `_::HeapArrayDisposer::allocate<word>(size)`
- calls `HeapArrayDisposer::Allocate_<word, true, true>::allocate(count, count)`
- calls `_::HeapArrayDisposer::allocateImpl(sizeof(word), elementCount, capacity, nullptr, nullptr)`
- which basically just returns `operator new(elementSize * capacity)`

I don't think `operator new()` is guaranteed to zero-initialize memory, although I wouldn't be surprised if some allocators have a mode that does, for the security/reproducibility benefits?  I notice that there's a separate `_::HeapArrayDisposer::allocateUninitialized()` that makes an almost identical call to `allocateImpl()`, except with a zero `elementCount`.

kentonv: Right, kj::heap does not guarantee memory is zero'd, and quite often it isn't.

PR #2082, opened 2024-06-18
http-over-capnp: Discard disconnect exceptions on final pump.

kentonv:
See comment for explanation.


kentonv:
Merging based on internal approval.

PR #2084, opened 2024-06-20
ArrayPtr-based async read api.

mikea:


c++/src/kj/compat/http.c++
```diff
@@ -1514,7 +1514,7 @@ public:
         co_return true;
       }
 
-      auto amount = co_await inner.tryRead(headerBuffer.begin(), 1, headerBuffer.size());
+      auto amount = co_await inner.tryRead(headerBuffer.asBytes(), 1);
```
kentonv, line 1517: redundant `asBytes()`?

kentonv (commented):
Are there any parts of this PR that aren't just mechanical updates which we should pay particular attention to?

mikea:
> Are there any parts of this PR that aren't just mechanical updates which we should pay particular attention to?

Not really. I don't think I have changed anything in "real" code. Tests had some cleanup that was mostly related to string/bytes simplifications where possible. Otherwise I tried to stay as close as possible to the original.

mikea:
I also have downstream changes read and I obviously plan to make sure all builds pass everywhere.

jasnell (approved):

kentonv (approved):

PR #2086, opened 2024-06-21
Improve hashCode().

kentonv:
Previously, `hashCode()` merely needed to return a value that was likely to be unique. When the input was already a 32-bit integer, it was acceptable to just return exactly the input value. However, this meant that the distribution of hashes wasn't necessarily uniform. To compensate for this, the hashtable implementation would use prime-number bucket counts. This, however, made computing the correct bucket expensive, as it required an integer division operation.

After this change, `hashCode()` is expected to return a uniform distribution. Thus, hashtables can have power-of-two size, avoiding the integer division. This improves performance in our string-keyed hashtable benchmark by 10%. The integer-keyed benchmark is slower, due to the need to spend more cycles computing hashes, but integer keys are already very fast anyway and most real-world hash tables tend to be string-keyed.

This also improves the behavior of open addressing with linear probing (which KJ's hashtable uses). Previously, if you inserted many sequential keys into a hash table, they would occupy consecutive buckets, creating a large range of buckets that were full. If you then tried to insert a non-sequential key that happened to hash somewhere into the range of occupied buckets, the hashtable wolud have to scan all the way to the end of the range to find an appropriate slot for the new key. Cap'n Proto's RPC tables tended to have this problem in particular.

### DIFF
--- a/c++/src/kj/hash.h
+++ b/c++/src/kj/hash.h
@@ -22,11 +22,15 @@
 #pragma once
 
 #include "string.h"
+#include <stdint.h>
 
 KJ_BEGIN_HEADER
 
 namespace kj {
 namespace _ {  // private
+
+inline uint intHash32(uint32_t i);
+inline uint intHash64(uint64_t i);
 
 struct HashCoder {
   // This is a dummy type with only one instance: HASHCODER (below).  To make an arbitrary type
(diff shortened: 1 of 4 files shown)


mikea (approved):
LG but I wonder why not to do crc32 for old architectures too for consistency? You wouldn't need a full table for that but one or two multiplications?

jasnell (approved):

jasnell:
Basic code change generally LGTM but there are a fair number of CI failures to address.

kentonv:
> LG but I wonder why not to do crc32 for old architectures too for consistency? You wouldn't need a full table for that but one or two multiplications?

There is no need for consistency, in fact, we actively don't want anyone depending on hashCode() being consistent, since we want to be able to change it. (HashMap, though, provides consistent iteration order that doesn't depend on the details of hashCode().)

The CRC hardware instruction, when available, is faster than the Thomas Wang hash. That's really the only reason I used it. If a software implementation of CRC can be similarly fast and similarly small code, then great, but I'm guessing it would be slower and more complicated.

fhanau:
> LG but I wonder why not to do crc32 for old architectures too for consistency? You wouldn't need a full table for that but one or two multiplications?

If we're using a software crc32 implementation using a half-byte implementation should be a good compromise between code size and performance (implementation at https://github.com/stbrumme/crc32/blob/master/Crc32.cpp#L133 is 20 lines of code, zlib-licensed). For an overview of software implementations and performance see https://create.stephan-brumme.com/crc32/.

I kind of agree with Kenton though, hashing is not central to capnproto so we don't need to have a great fallback option here TBH.

c++/src/kj/table.h
```diff
@@ -893,7 +893,10 @@ inline size_t probeHash(const kj::Array<HashBucket>& buckets, size_t i) {
 
 kj::Array<HashBucket> rehash(kj::ArrayPtr<const HashBucket> oldBuckets, size_t targetSize);
 
-uint chooseBucket(uint hash, uint count);
+inline uint chooseBucket(uint hash, uint count) {
+  KJ_IASSERT(__builtin_popcount(count) == 1, "hash bucket count must be power of two!");
```
fhanau, line 897: MSVC wants __popcount() here – thankfully we have already handled that in `popCount()` defined in kj/common.h

c++/src/kj/hash.h
```diff
@@ -195,6 +211,60 @@ inline uint HashCoder::operator*(T e) const {
...
+
+inline uint intHash64(uint64_t i) {
+  // Basic 64-bit integer hash function.
+  //
+  // Like intHash32(), but where the input is 64 bits.
+
+#if __CRC32__
+  return __builtin_ia32_crc32di(0, i);
+#elif __ARM_FEATURE_CRC32
+  return __crc32d(0, i);
```
fhanau, line 253: Apparently Apple Clang expects `#include <arm_acle.h>` to be present for this to be included 🙄
Certainly harder than expected to get this right

fhanau: Alternatively, clang also has `__builtin_arm_crc32w()` which shouldn't need an extra header.

kentonv:
Fixups to this point: https://github.com/capnproto/capnproto/compare/837075da2fa6639058f474737ff058b4858e4595..3a63ccbbe4b58def33e2301f4e49b5b689cf0fed

fhanau (approved):
Looks reasonable to me. It would be helpful to add a comment in intHash32() explicitly stating that the hash implementation is not stable and may return different results in the future (in particular, the arm64 and x86 CRC32 implementations use different polynomials and thus return different results).

kentonv:
`kj::hashCode()` itself has the comment:

```
// Hash codes may differ between different processes, even running exactly the same code.
```

fhanau:
> `kj::hashCode()` itself has the comment:
> 
> ```
> // Hash codes may differ between different processes, even running exactly the same code.
> ```

Ok, that works for me.

PR #2088, opened 2024-06-23
Add ArrayPtr fill(ArrayPtr<T>) variant

jasnell:
Adds a new overload of ArrayPtr::fill that takes another ArrayPtr as an argument. This is useful for cases where we want to fill the target with a sequence of values rather than a single repeated value.

### DIFF
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -1940,12 +1940,25 @@ public:
   // Syntax sugar for invoking U::from.
   // Used to chain conversion calls rather than wrap with function.
 
-  inline void fill(T t) { 
+  inline void fill(T t) {
     // Fill the area by copying t over every element.
 
-    for (size_t i = 0; i < size_; i++) { ptr[i] = t; } 
+    for (size_t i = 0; i < size_; i++) { ptr[i] = t; }
     // All modern compilers are smart enough to optimize this loop for simple T's.
     // libc++ std::fill doesn't have memset specialization either.
+  }
+
+  inline void fill(ArrayPtr<const T> other) {
+    // Fill the area by copying each element of other, in sequence, over every element.
+    const size_t otherSize = other.size();
+    KJ_IREQUIRE(otherSize > 0, "fill requires non-empty source array");
+    T* __restrict__ dst = begin();
+    const T* __restrict__ src = other.begin();
+    size_t counter = 0;
+    for (size_t s = size_, i = 0; i < s; i++) {
+      dst[i] = src[counter];
+      if (++counter == otherSize) counter = 0;
+    }
   }
 
   inline void copyFrom(kj::ArrayPtr<const T> other) {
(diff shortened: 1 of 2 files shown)


c++/src/kj/common.h
```diff
@@ -1940,22 +1940,31 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
...
+    for (size_t i = 0; i < size_; i++) { ptr[i] = t; }
     // All modern compilers are smart enough to optimize this loop for simple T's.
     // libc++ std::fill doesn't have memset specialization either.
   }
 
+  inline void fill(ArrayPtr<const T> other) {
+    // Fill the area by copying each element of other, in sequence, over every element.
+    if (other.size() == 0) return;
+    T* __restrict__ dst = begin();
+    for (size_t s = size_, i = 0; i < s; i++) {
```
mikea, line 1955: I'm not sure if it matters, but I'd define src as __restrict__ as well (like in copyFrom) to enable max optimizations.

c++/src/kj/common.h
```diff
@@ -1940,22 +1940,31 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
...
+      dst[i] = other[i % other.size()];
```
mikea, line 1956: also paranoid hat on: can we trust compiler to hoist the size() or should we extract a immutable local?

c++/src/kj/common.h
```diff
@@ -1940,22 +1940,31 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
...
 
-    for (size_t i = 0; i < size_; i++) { ptr[i] = t; } 
+    for (size_t i = 0; i < size_; i++) { ptr[i] = t; }
     // All modern compilers are smart enough to optimize this loop for simple T's.
     // libc++ std::fill doesn't have memset specialization either.
   }
 
+  inline void fill(ArrayPtr<const T> other) {
+    // Fill the area by copying each element of other, in sequence, over every element.
+    if (other.size() == 0) return;
```
mikea, line 1953: I think that this has to be an error. I think `fill` contract should include initializing every element.

jasnell:
> I'm curious where did you need this?

The Node.js `Buffer` implementation needs this. There's an overly complicated implementation of it there now that I would like to simplify and eliminate

c++/src/kj/common.h
jasnell: We could make it an error.

c++/src/kj/common.h
jasnell: hmm... I think so but safer to extract I think

jclee, line 1956: I suppose `fill()` will only overwrite whatever T's assignment operator writes to, so it might leave some old content in elements, particularly for structures that have padding bytes.  Not sure if that actually matters for anything other than security applications, though, or things that compute equality or hashes from raw byte content?

kentonv: For that matter the div operation is quite expensive, keeping a separate counter and resetting it to zero each time it reaches size would be a lot faster -- but entirely possible the compiler is smart enough to do that already.

jasnell: Done

c++/src/kj/common.h
jasnell: Done

kentonv (approved):

jclee (approved):

PR #2090, opened 2024-06-25
[nfc] Convert remaining evalLater() calls to yield where possible

fhanau:
That should actually cover the remaining evalLater()/evalLast() calls that can be simplified, please correct me if I'm wrong.

### DIFF
--- a/c++/src/kj/async.c++
+++ b/c++/src/kj/async.c++
@@ -196,10 +196,6 @@ private:
   void* traceAddr;
 };
 
-struct DummyFunctor {
-  void operator()() {};
-};
-
 }  // namespace
 
 // =======================================================================================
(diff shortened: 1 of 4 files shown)


c++/src/kj/async.c++
```diff
@@ -2854,7 +2850,7 @@ Promise<void> yield() {
       output.as<_::Void>() = _::Void();
     }
     void tracePromise(_::TraceBuilder& builder, bool stopAtNextEvent) override {
-      builder.add(reinterpret_cast<void*>(&kj::evalLater<DummyFunctor>));
+      builder.add(reinterpret_cast<void*>(&kj::yield));
```
fhanau, line 2853: Not entirely sure about this and the one below – I think this is safe but referencing yield within yield feels slightly too clever.

kentonv: Yes this is correct, the goal is only to get an address which can be put into a stack trace such that when it's later symbolicated it tells people what this promise node means.

You can also delete the definition of `DummyFunctor` now.

kentonv: Oh you already deleted it, great!

kentonv (approved):

PR #2091, opened 2024-06-25
Make sure hashCode(uint64) and hashCode(uint32) are consistent.

kentonv:


justin-mp (approved):

c++/src/kj/hash.h
```diff
@@ -247,6 +247,13 @@ inline uint intHash64(uint64_t i) {
   //
   // Like intHash32(), but where the input is 64 bits.
 
+  // It's important that if the actual value is in the 32-bit range, the hash code is consistent
+  // with a 32-bit integer. Otherwise, if you have, say, a `HashMap<uint64_t, T>` and you write
+  // `map.find(1)` it won't work, because `1` is type `int` and will be hashed as a 32-bit integer.
+  if (i <= ~(uint32_t)0) {
```
justin-mp, line 253: nit: if we already have included `<stdint.h>` somewhere to get `uint32_t`, we might as well also use UINT32_MAX here for readability.

fhanau (approved):

Issue #2089, opened 2024-06-24
Adding a disown() method to kj::Own<T>?


jasnell:
One of the current capabilities of `std::unique_ptr<T>` that is not replicated in `kj::Own<T>` is the ability to surrender ownership of the underlying thing. There are obviously some concerns with that if the `kj::Own<T>` is using a custom destroyer but I think in general having a `own.disown()` method that releases the underlying pointer would be helpful. There's at least one case in the workerd codebase where this would be useful. (https://github.com/cloudflare/workerd/pull/2325#discussion_r1651678832)

```
template <typename T>
class Own {
 // ...

  T* disown();
}
```

/cc @mikea

kentonv:
Maybe it should require that you pass in the disposer, to prove you know how this thing is meant to be disposed? If you pass in the wrong disposer, you'd get an exception (rather than a memory leak or UB).

jasnell:
> Maybe it should require that you pass in the disposer...

Good idea.

kentonv:
Implemented in #2092.

PR #2092, opened 2024-06-25
Add disown() method to kj::Own

garrettgu10:
Addresses #2089

### DIFF
--- a/c++/src/kj/memory-test.c++
+++ b/c++/src/kj/memory-test.c++
@@ -376,7 +376,7 @@ TEST(Memory, OwnConstVoid) {
     Own<SingularDerivedDynamic> ptr = heap<SingularDerivedDynamic>(123, destructorCalled);
     SingularDerivedDynamic* addr = ptr.get();
 
-    ptr.disown(&_::HeapDisposer<SingularDerivedDynamic>::instance);
+    KJ_EXPECT(ptr.disown(&_::HeapDisposer<SingularDerivedDynamic>::instance) == addr);
     KJ_EXPECT(!destructorCalled);
     ptr = nullptr;
     KJ_EXPECT(!destructorCalled);
--- a/c++/src/kj/memory-test.c++
+++ b/c++/src/kj/memory-test.c++
@@ -370,6 +370,20 @@ TEST(Memory, OwnConstVoid) {
     maybe = kj::none;
     KJ_EXPECT(maybe == kj::none);
   }
+
+  {
+    bool destructorCalled = false;
+    Own<SingularDerivedDynamic> ptr = heap<SingularDerivedDynamic>(123, destructorCalled);
+    SingularDerivedDynamic* addr = ptr.get();
+
+    ptr.disown(&_::HeapDisposer<SingularDerivedDynamic>::instance);
+    KJ_EXPECT(!destructorCalled);
+    ptr = nullptr;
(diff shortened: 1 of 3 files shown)


c++/src/kj/memory.h
```diff
@@ -313,6 +313,14 @@ class Own<T, decltype(nullptr)> {
   inline operator T*() { return ptr; }
   inline operator const T*() const { return ptr; }
 
+  // Surrenders ownership of the underlying object to the caller. The caller must pass in the 
+  // correct disposer to prove that they know how the object is meant to be disposed of. 
+  inline T* disown(const Disposer* d) {
+    KJ_REQUIRE(d == disposer, "When disowning an object, disposer must be equal to Own's disposer");
```
kentonv, line 319: KJ_REQUIRE isn't defined here, and you probably can't #include it without creating a cyclic dependency. You will need to do something like:

`if (d != disposer) throwWrongDisposerError();`

and then implement `throwWrongDisposerError()` out-of-line (in `memory.c++`).

c++/src/kj/memory.h
```diff
@@ -224,6 +224,8 @@ using PtrCounter = AtomicPtrCounter;
 template <typename T, typename StaticDisposer = decltype(nullptr)>
 class Own;
 
+void throwWrongDisposerError();
```
kentonv, line 227: Nitpick: Probably should be in `_` namespace, where we put everything that's not intended to be a public API...

```
namespace _ {  // (private)
void throwWrongDisposerError();
}  // namespace _ (private)
```

jasnell:
The change generally LGTM but this needs tests

kentonv:
Sorry for the back-and-forth but I agree with James, there should be a unit test.

c++/src/kj/memory-test.c++
```diff
@@ -370,6 +370,20 @@ TEST(Memory, OwnConstVoid) {
     maybe = kj::none;
     KJ_EXPECT(maybe == kj::none);
   }
+
+  {
+    bool destructorCalled = false;
+    Own<SingularDerivedDynamic> ptr = heap<SingularDerivedDynamic>(123, destructorCalled);
+    SingularDerivedDynamic* addr = ptr.get();
+
+    ptr.disown(&_::HeapDisposer<SingularDerivedDynamic>::instance);
```
kentonv, line 379:
```suggestion
    KJ_EXPECT(ptr.disown(&_::HeapDisposer<SingularDerivedDynamic>::instance) == addr);
```

kentonv (approved):
Will merge when CI finishes.

(If someone else merges, please be sure to squash-merge...)

c++/src/kj/memory.h
```diff
@@ -429,6 +440,16 @@ class Own {
   inline operator T*() { return ptr; }
   inline operator const T*() const { return ptr; }
 
+  // Surrenders ownership of the underlying object to the caller. The caller must pass in the 
+  // correct disposer to prove that they know how the object is meant to be disposed of. 
+  template<typename SD>
+  inline T* disown() {
+    static_assert(std::is_same<StaticDisposer, SD>::value, "disposer must be the same as Own's disposer");
```
kentonv, line 447:
```suggestion
    static_assert(kj::isSameType<StaticDisposer, SD>, "disposer must be the same as Own's disposer");
```

PR #2093, opened 2024-06-28
RPC: Implement idle hints and other refactoring

kentonv:
The last (main) commit of this PR adds an API by which the RpcSystem will inform the VatNetwork when a connection is "idle", meaning it has no active calls or capabilities on it in either direction.

This gives us a reasonable way to shut down connections that are "done", which will enable us to switch over to a connection-per-session approach rather than our current connection-per-destination approach. This will in turn be important to make three-party handoff work well.


mikea (approved):
This looks good but I can't claim I understand rpc code enogh for a very thorough review.

The one thing that bugs me a little about this are new KJ_IREQUIRE related to implementation consistency. What if the logic is slightly wrong and we're missing a transition? Will our servers start to crash?

Issue #2067, opened 2024-06-05
DynamicStructBuilder::set hits KJ_UNREACHABLE when assigning a group


alexrobomind:
https://github.com/capnproto/capnproto/blob/5faf3144b5e1b0fbbe80992f7e3a890a472c82a3/c%2B%2B/src/capnp/dynamic.c%2B%2B#L663

I guess this case block should be terminated by a return after this line?

kentonv:
Indeed, and I guess this means we don't have any unit test coverage for this, so it may be broken in other ways too... :/

fhanau:
Closing this – https://github.com/capnproto/capnproto/pull/2094 has been merged to the master and v2 branches.

PR #2094, opened 2024-06-28
Fixes for group assignment in dynamic API and BufferInputStreamWrapper::tryRead

### LINKED ISSUE #1980: BufferedInputStreamWrapper::tryRead fails on EOF
Currently, the implementation of `BufferedInputStreamWrapper::tryRead` will throw on premature EOF due to forwarding to `InputStream::read` if the buffer is insufficient to fill the request:

https://github.com/capnproto/capnproto/blob/737dba8da94fcde061ebd8b9b3e7efff727b9877/c%2B%2B/src/kj/io.c%2B%2B#L156
https://github.com/capnproto/capnproto/blob/737dba8da94fcde061ebd8b9b3e7efff727b9877/c%2B%2B/src/kj/io.c%2B%2B#L164

Wouldn't it make more sense to forward the read requests to `InputStream::tryRead` instead?

alexrobomind:
This fixes #1980 and #2067 and adds a test case that would have fired on #2067. Based on master because I still work on v1 and didn't want to port it forward and back again.

### DIFF
--- a/c++/src/capnp/dynamic.c++
+++ b/c++/src/capnp/dynamic.c++
@@ -662,7 +662,7 @@ void DynamicStruct::Builder::set(StructSchema::Field field, const DynamicValue::
         }
       }
 	  
-	  return;
+      return;
     }
   }
 
--- a/c++/src/capnp/dynamic.c++
+++ b/c++/src/capnp/dynamic.c++
@@ -661,6 +661,8 @@ void DynamicStruct::Builder::set(StructSchema::Field field, const DynamicValue::
           dst.set(field, src.get(field));
         }
       }
+	  
+	  return;
     }
   }
 
--- a/c++/src/capnp/dynamic-test.c++
+++ b/c++/src/capnp/dynamic-test.c++
@@ -492,6 +492,25 @@ TEST(DynamicApi, BuilderAssign) {
   listValue.set(0, 123);
 }
 
+TEST(DynamicApi, SetGroup) {
+  MallocMessageBuilder srcBuilder;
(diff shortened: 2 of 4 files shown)


c++/src/capnp/dynamic.c++
```diff
@@ -661,6 +661,8 @@ void DynamicStruct::Builder::set(StructSchema::Field field, const DynamicValue::
           dst.set(field, src.get(field));
         }
       }
+	  
+	  return;
```
kentonv:
```suggestion
      return;
```

alexrobomind: Yeah, noticed it too in the editor. Looked the same offline, sorry.

kentonv: Oh, you did it right as I was commenting.

kentonv (approved):
Thanks. Looks good, will merge after CI passes.

alexrobomind:
Thanks a lot. Looking in the log, the new test seems to be there. Also, man you're fast ...

kentonv:
> Also, man you're fast ...

Not normally! You just happened to send this right as I was reaching inbox zero before heading out on vacation. :)

alexrobomind:
Lucky me :-P. Enjoy your holidays.

fhanau:
@kentonv this is also needed on the v2 branch right? I can make a PR.

kentonv:
@fhanau Yeah. Actually there's probably a lot of patches on the master branch that should be picked into v2.

PR #2095, opened 2024-07-03
Make WebSocket::disconnect() synchronous and fix some obscure promise ordering bugs

kentonv:
I was just sort of messing around with promise ordering, trying to use depth-first scheduling in more cases, and it revealed some bugs.

In a big of a yak shave, I ended up changing `WebSocket::disconnect()` to be synchronous, which I've wanted to do for a while.

### DIFF
--- a/c++/src/capnp/compat/http-over-capnp.c++
+++ b/c++/src/capnp/compat/http-over-capnp.c++
@@ -78,9 +78,8 @@ public:
     KJ_IF_SOME(e, error) {
       onEnd->reject(kj::mv(*e));
     } else KJ_IF_SOME(ws, webSocket) {
-      onEnd->fulfill(kj::evalNow([&]() {
-        return ws.disconnect().attach(kj::mv(ownWebSocket));
-      }));
+      ws.disconnect();
+      onEnd->fulfill(kj::READY_NOW);
     } else {
       // cancel() was called -- we assume no one is waiting on the fulfiller
     }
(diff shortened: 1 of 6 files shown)


c++/src/capnp/compat/http-over-capnp.c++
```diff
@@ -693,12 +690,30 @@ public:
...
+    // the reply to this RPC, because in many cases we will call this, write a response body,
+    // and then immediately return from the HttpService::requset() handler. At that point, we
+    // would like CapnpToKjHttpServiceAdapter::request() to be able to propagate this return
+    // immediately *without* waiting for a round trip to the client, so without waiting for
+    // `startResponse()` to finish. However, we also do not want to inadvertently cancel
+    // `startResponse()`, so we have to save the promise somewhere. Since this is an RPC that
+    // doesn't depend on the lifetime of any other object... we will just detach() it.
+
+    promise.detach([](kj::Exception&& e) {
+      if (e.getType() == kj::Exception::Type::FAILED) {
```
harrishancock, line 734: I'm used to seeing `!= DISCONNECTED` to inhibit logging, so this stuck out to me. I guess OVERLOADED isn't likely to occur, and we wouldn't want to log REMOTE exceptions?

harrishancock (approved):

c++/src/capnp/compat/http-over-capnp.c++
kentonv: There's no REMOTE type. I figured OVERLOADED is probably not relevant to log, and UNIMPLEMENTED isn't going to be an issue here. Really I could have skipped logging any exceptions here since presumably the problem is going to be noticed elsewhere but figured logging FAILED is pretty safe since it probably represents a real problem.

PR #2097, opened 2024-07-05
[build] Update bazel configuration, add clang-based Windows CI job

fhanau:
- Roll Bazel 7.1.2 : 7.2.1.
- Refactor how libc++ is being enabled to avoid modifying action_env
- Enable libc++ config option, helpful mostly to avoid breakage of downstream projects following include changes.
- Improve link performance under MSVC
- Add a CI job for building with bazel on Windows (using clang-cl compiler). This was broken so far, so we should test it to make sure that configuration continues to be supported.

On macOS, the `_LIBCPP_REMOVE_TRANSITIVE_INCLUDES` change reduces the total include size of source files in src/kj from 129.6MB to 86.1MB. More complex source files will likely see a smaller improvement, but this is certainly nice to have.


fhanau:
Updated the PR to give this a more descriptive name. These changes are not essential for capnproto, but having the Windows runner and libc++ define will help avoid changes that break downstream projects. See https://vitaut.net/posts/2024/faster-cpp-compile-times/ for some background on the define, it gives us some 'free lunch' in terms of header parse time.

c++/src/kj/async-coroutine-test.c++
```diff
@@ -381,9 +381,9 @@ KJ_TEST("co_await only sees coroutine destruction exceptions if promise was not
       awaitPromise(kj::mv(rejectedThrowyDtorPromise)).wait(waitScope));
 }
 
-#if (!_MSC_VER || defined(__clang__)) && !__aarch64__
+#if !_MSC_VER && !__aarch64__
 // TODO(msvc): This test relies on GetFunctorStartAddress, which is not supported on MSVC currently,
-//   so skip the test.
+//   so skip the test. It also appears to fail on clang-cl, which uses the MSVC backend.
```
kentonv, line 386: It looks like the relevant code in `async-inl.h` (in `PtmfHelper`) is guarded by `#if __GNUG__`, which evaluates false on clang-cl, which would indeed break this test.

The code in question depends on the details of the Itanium C++ ABI, which I'm guessing means it wouldn't work on clang-cl, which uses the MSVC ABI. So it seems the check is correct and skipping the test is the right thing to do.

Comment might be better as:

```suggestion
//   so skip the test. Note this is an ABI issue, so clang-cl is also not supported.
```

kentonv (approved):

PR #2100, opened 2024-07-07
Include <atomic> more often for kj::Pin

smerritt:
When KJ_ASSERT_PTR_COUNTERS was defined but KJ_DEBUG was not, then memory.h would fail to include <atomic>, resulting in compilation errors.


c++/src/kj/memory.h
```diff
@@ -22,7 +22,13 @@
 #pragma once
 
 #include "common.h"
-#ifdef KJ_DEBUG
+
+#if defined(KJ_DEBUG) && !defined(KJ_ASSERT_PTR_COUNTERS)
+#define KJ_ASSERT_PTR_COUNTERS
```
kentonv, line 27: This is not new in your change and I'm not asking you to fix it in your PR, but:

This really ought to be defined with the value `1`, and the `#ifdef KJ_ASSERT_PTR_COUNTERS` should be `#if KJ_ASSERT_PTR_COUNTERS`, so that someone can explicitly define it to `0` if they want to disable this while still building in debug mode. Most KJ feature macros work this way (although `KJ_DEBUG` in an exception, for no good reason).

kentonv (approved):

c++/src/kj/memory.h
kentonv: fyi @mikea

smerritt:
Fixup commit 328f49c3 makes sure `kj::_::atomicPtrCounterAssertionFailed` is always defined. Otherwise, if you compile your library's translation unit with `-DKJ_ASSERT_PTR_COUNTERS=1` but don't compile KJ with it, you get a linker error. If it's unused, LTO should strip that function out.

kentonv:
> Otherwise, if you compile your library's translation unit with -DKJ_ASSERT_PTR_COUNTERS=1 but don't compile KJ with it, you get a linker error.

If you do that, many horrible things will happen as different transaction units will expect various data types to have different layouts. TBH it may actually be better to keep this as a linker error to avoid disaster at runtime.

c++/src/kj/memory.h
mikea: https://github.com/capnproto/capnproto/pull/2262 should adress this.

PR #2109, opened 2024-07-19
Added Pragma and Cache-Control Headers to HttpHeaderId

AdityaAtulTewari:
Adds support for standard headers used to control caching in kj.


kentonv (changes requested):
We only make headers part of the base enum if they need to be used by the KJ HTTP implementation itself. Anything else can be defined by applications using the HttpHeaderTable API.

AdityaAtulTewari:
Ah got it thanks!

PR #2112, opened 2024-07-23
kj::Rc and kj::Arc escape hatches

mikea:
T* get() 
People have asked for it, and it makes migrating easier. The only danger is using addRef(T&) again, but that is on the implementer to carefully analyze the usage.

### DIFF
--- a/c++/src/kj/refcount.h
+++ b/c++/src/kj/refcount.h
@@ -210,7 +210,8 @@ public:
   inline T* operator->() { return own.get(); }
   inline const T* operator->() const { return own.get(); }
 
-  // do not expose * to avoid dangling references
+  inline T* get() { return own.get(); }
+  inline const T* get() const { return own.get(); }
 
 private:
   Rc(T* t) : own(t, *t) { }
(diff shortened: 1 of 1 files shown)


kentonv:
> The only danger is using addRef(T&) again, but that is on the implementer.

Hmm, seems like it'll be too easy to miss updating some `addRef()`s when migrating from Own to Rc. Is there some way we could avoid that, or would it make sense to leave out `operator*` (but keep `get()`)?

I'm also happy with the answer: "Making migration easier by supporting `operator*` is more important."

mikea:
> > The only danger is using addRef(T&) again, but that is on the implementer.
> 
> Hmm, seems like it'll be too easy to miss updating some `addRef()`s when migrating from Own to Rc. Is there some way we could avoid that, or would it make sense to leave out `operator*` (but keep `get()`)?
> 
> I'm also happy with the answer: "Making migration easier by supporting `operator*` is more important."

Not sure what I was thinking yesterday... We could either ask clients to use `*object.get()` or add `T& ref()` in Rust manner.
Do you have a preference?

kentonv:
Either of those sound fine to me.

mikea:
> Either of those sound fine to me.

I went back and forth on this one and in the end decided that `.get()` and `.ref()` need to be treated the same (as in carefully analyzed). So there's no real reason to differentiate between the two.

Keeping `get()` and expecting users to do `*rc.get()` to obtain references.

kentonv (approved):

PR #2114, opened 2024-07-25
3PH part 1: Random drive-bys

kentonv:
(Split off from #2111)

Just some random fixes and cleanups before implementing starts in earnest.

### DIFF
--- a/c++/src/capnp/rpc.c++
+++ b/c++/src/capnp/rpc.c++
@@ -1764,7 +1764,7 @@ private:
         return import(descriptor.getThirdPartyHosted().getVineId(), false, kj::mv(fd));
 
       default:
-        KJ_FAIL_REQUIRE("unknown CapDescriptor type") { break; }
+        KJ_FAIL_REQUIRE("unknown CapDescriptor type");
         return newBrokenCap("unknown CapDescriptor type");
     }
   }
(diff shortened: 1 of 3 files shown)


c++/src/capnp/rpc.c++
```diff
@@ -1180,6 +1180,16 @@ private:
     Request<AnyPointer, AnyPointer> newCall(
         uint64_t interfaceId, uint16_t methodId, kj::Maybe<MessageSize> sizeHint,
         CallHints hints) override {
+      if (isResolved()) {
+        // Go directly to the resolved capability.
+        //
+        // If we don't do this now, then well return an RpcRequest which, as soon as send() is
```
ObsidianMinor, line 1186:
```suggestion
      // If we don't do this now, then we'll return an RpcRequest which, as soon as send() is
```

c++/src/capnp/rpc.c++
```diff
@@ -1190,6 +1200,11 @@ private:
 
     VoidPromiseAndPipeline call(uint64_t interfaceId, uint16_t methodId,
                                 kj::Own<CallContextHook>&& context, CallHints hints) override {
+      if (isResolved()) {
+        // Same as in newCall().
+        return cap->call(interfaceId, methodId, kj::mv(context), hints);
+      }
+
       receivedCall = true;
       return cap->call(interfaceId, methodId, kj::mv(context), hints);
```
ObsidianMinor, line 1209: Not sure how this is different? Just avoiding the `receivedCall` field setting.

kentonv: Yeah you're right, it has no actual effect here, I was just trying to be consistent I guess.

kentonv: (The setting of `receivedCall` doesn't matter post-resolve, since it only impacts what happens at resolve time.)

ObsidianMinor (approved):

PR #2115, opened 2024-07-25
3PH part2: Protocol and basic interface changes

kentonv:
(Split off from https://github.com/capnproto/capnproto/pull/2111)

No implementation yet, but this:
- Updates the (not-yet-implemented) 3PH protocol spec and comments for modern thinking.
- Updates the VatNetwork interface with new virtual functions which a VatNetwork must implement to support three-party handoff.

Again, this is basically all design work -- there's no implementation code here.


kentonv:
@dwrensha, @lthibault: This PR contains most of the protocol changes for 3PH FWIW.

c++/src/capnp/rpc.capnp
```diff
@@ -821,10 +821,16 @@ struct Provide {
...
-  # returned once the third party has successfully received the capability.  The sender must at some
-  # point send a `Finish` message as with any other call, and that message can be used to cancel the
-  # whole operation.
+  # Question ID to be held open until the recipient has received the capability.  No `Return`
+  # message will be sent, similarly to a `Call` that has `onlyPromisePipeline = true`.  The sender
+  # must at some point send a `Finish` message as with any other call, which unregisters the
+  # provision. Typically, the sender would send such a `Finish` message once it has received
+  # confirmation from the third party that the capability has been successfully accepted.
+  #
+  # More than one party can accept the provision. The recipient will offer the capability to
```
ObsidianMinor, line 841: Does this mean provisions must **always** be available for any client that could possibly accept them? In the old model, you could clean up provisions if the target couldn't be contacted or got disconnected before they could accept. But now anyone could accept the provision. At any time. You can't clean up if the original target is lost because they could've forwarded it to someone downstream.

ObsidianMinor: Ah, wait, so provisions *can* be cleaned up, when the original connection has sent a finish or when the connection that created the provision is severed. So you just have to do a long chain of finish messages between all parties before the provision is closed I guess. But let's think about the disconnection case. A provisions Z to B, B forwards the provision to C, but A loses connection to B. The provision is lost, but C will still try to accept the provision from A.

What happens if the connection is reestablished and the same thing is provisioned again with the same setup. So everything matches, we just have to assume the user is making some sort of randomness as part of the provision data right?

In the event of a disconnect, can a forwarding vat inform the vat they forwarded to that the provision is gone?

kentonv: This is part of what the "vine" is for. The introducer passes a "vine" capability to the recipient. The recipient only drops the vine once they've successfully accepted the provision over a direct connection. Once the vine is dropped, then the provision can be cleaned up. If the same provision is sent to multiple parties, they all get the same vine capability. If any connections are disconnected, the vine is broken.

> What happens if the connection is reestablished and the same thing is provisioned again with the same setup. So everything matches,

I expect most VatNetworks will generate a new random token for each provision, so there's no ambiguity.

kentonv: I added commentary here: 05f81ef6

kentonv:
Ping @ObsidianMinor

c++/src/capnp/rpc.capnp
```diff
@@ -766,6 +766,10 @@ struct Disembargo {
   using EmbargoId = UInt32;
   # Used in `senderLoopback` and `receiverLoopback`, below.
 
+  using ThirdPartyEmbargoId = Data;
```
ObsidianMinor, line 776: Why not `AnyPointer` that's a network type parameter?

kentonv: Because the RPC system needs to create a HashMap keyed by this. Technically it could use an AnyPointer and canonicalize it for this purpose but that's a lot of overhead and I can't think of a reason why anyone would use anything other than random bytes here.

ObsidianMinor (approved):

PR #2116, opened 2024-07-31
Add `asBytes` functions to easily convert to ArrayPtr<byte>

AdityaAtulTewari:
This should force comparisons with `memcmp` much easier to manage because it is very easy to force now.

### DIFF
--- a/c++/src/kj/common-test.c++
+++ b/c++/src/kj/common-test.c++
@@ -965,6 +965,45 @@ KJ_TEST("ArrayPtr equality for arrays of different string types") {
   }
 }
 
+KJ_TEST("asBytes Tests") {
+  const char helloMessage[] = "helloThere";
+
+  // Use size to specify
+  {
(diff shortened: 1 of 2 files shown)


kentonv:
This just abbreviates `kj::arrayPtr(...).asBytes()` to `kj::asBytes(...)`? Is this pattern actually common enough to warrant the shortcuts? It doesn't really feel like it to me.

mikea:
> This just abbreviates `kj::arrayPtr(...).asBytes()` to `kj::asBytes(...)`? Is this pattern actually common enough to warrant the shortcuts? It doesn't really feel like it to me.

my not-very-prices grep shows 48 usages across 28 files. Given that IO now takes bytes, this will be more prevalent I think.

kentonv:
How about a simpler definition:

```
template <typename... Params>
auto asBytes(Params&&... params) {
  return kj::arrayPtr(kj::fwd<Params>(params)...).asBytes();
}
```

PR #2118, opened 2024-08-05
Fix websocket handling of oversized compressed messages.

danlapid:
Fixes: #2117 
The previous implementation had 2 issues:
1. Compressed messages of exactly maxsize were incorrectly rejected.
2. Compressed messages above maxsize threw exception outside of receive instead of returning status code 1009. I added a test case to ensure that the new implementation is correct.

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -2853,9 +2853,18 @@ public:
             }
             bool addNullTerminator = true;
             // We want to add the null terminator when receiving a TEXT message.
-            auto decompressed = decompressor.processMessage(message, originalMaxSize,
+            auto decompressedOrError = decompressor.processMessage(message, originalMaxSize, 
                 addNullTerminator);
-            return Message(kj::String(decompressed.releaseAsChars()));
+            KJ_SWITCH_ONEOF(decompressedOrError) {
+              KJ_CASE_ONEOF(protocolError, ProtocolError) {
+                KJ_DBG("Returning error with", protocolError.description);
+                return sendCloseDueToError(protocolError.statusCode, protocolError.description)
+                    .attach(kj::mv(decompressedOrError));
+              }
+              KJ_CASE_ONEOF(decompressed, kj::Array<byte>) {
+                return Message(kj::String(decompressed.releaseAsChars()));
+              }
+            }
           }
 #endif // KJ_HAS_ZLIB
           message.back() = '\0';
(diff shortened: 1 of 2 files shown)


c++/src/kj/compat/http.c++
```diff
@@ -3373,8 +3392,9 @@ private:
           KJ_IF_SOME(m, maxSize) {
             // This is only non-null for `receive` calls, so we must be decompressing. We don't want
             // the decompressed message to OOM us, so let's make sure it's not too big.
-            KJ_REQUIRE(totalBytesProcessed < m,
-                "Decompressed WebSocket message is too large");
+            if (totalBytesProcessed > m) {
+              kj::throwRecoverableException(KJ_EXCEPTION(FAILED, kj::str("Decompressed WebSocket message is too large: ", totalBytesProcessed, " > ", m)));
+            }
```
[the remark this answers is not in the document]
danlapid: Yeah makes sense, it felt a bit hacky but I wasn't sure what's the appropriate way to go about it.
It might be reasonable to return OneOf<..., HttpHeaders::ProtocolError> just to be a little more forward compatible, what do you think?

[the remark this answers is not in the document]
danlapid: Yeah the manual formatting was because I wanted to return the description but I agree it feels hacky,.
I switched ProcesMessage to return OneOf<..., ProtocolError> which seemed appropriate.
What do you think about how it looks now?

c++/src/kj/compat/http.c++
```diff
@@ -2853,9 +2853,16 @@ public:
...
             bool addNullTerminator = true;
             // We want to add the null terminator when receiving a TEXT message.
-            auto decompressed = decompressor.processMessage(message, originalMaxSize,
-                addNullTerminator);
-            return Message(kj::String(decompressed.releaseAsChars()));
+            auto decompressedOrError = decompressor.processMessage(message, originalMaxSize, addNullTerminator);
+            KJ_SWITCH_ONEOF(decompressedOrError) {
+              KJ_CASE_ONEOF(protocolError, ProtocolError) {
+                KJ_DBG("Returning error with", protocolError.description);
+                return sendCloseDueToError(protocolError.statusCode, protocolError.description).attach(kj::mv(decompressedOrError));
```
kentonv, line 2860: Nitpick: [KJ style guide](https://github.com/capnproto/capnproto/blob/v2/style-guide.md) requires wrapping at 100 columns.

```suggestion
                return sendCloseDueToError(protocolError.statusCode, protocolError.description)
                    .attach(kj::mv(decompressedOrError));
```

Can you fix the other places too? I suggest configuring your editor to show a line at 100 columns.

c++/src/kj/compat/http.c++
```diff
@@ -3373,8 +3396,11 @@ private:
           KJ_IF_SOME(m, maxSize) {
             // This is only non-null for `receive` calls, so we must be decompressing. We don't want
             // the decompressed message to OOM us, so let's make sure it's not too big.
-            KJ_REQUIRE(totalBytesProcessed < m,
-                "Decompressed WebSocket message is too large");
+            if (totalBytesProcessed > m) {
+              return ProtocolError {
```
kentonv, line 3400: Can you use `WebSocket::ProtocolError`, instead of `ProtocolError` which is meant for HTTP errors?

danlapid: Yea sure, it's using WebSocket::ProtocolError because the zlibcontext is defined inside of the websocket class but I can change it so that it's clearer we're using WebSocket's one

kentonv: Oh! In that case it's fine. I misread.

kentonv (approved):

PR #2119, opened 2024-08-05
Fix minor memory safety issue in http-over-capnp

fhanau:
See EW-8586. The affected code has several levels of indirection so make sure to convince yourself that this handles the underlying issue correctly, suggestions on how to best phrase the comment are welcome.

### DIFF
--- a/c++/src/capnp/compat/http-over-capnp.c++
+++ b/c++/src/capnp/compat/http-over-capnp.c++
@@ -151,6 +151,7 @@ public:
   }
 
 private:
+  friend HttpOverCapnpFactory::ClientRequestContextImpl;
   kj::Maybe<kj::WebSocket&> webSocket;  // becomes none when canceled
   kj::Own<kj::WebSocket> ownWebSocket;
   kj::Promise<Capability::Client> shorteningPromise;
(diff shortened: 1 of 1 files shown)


c++/src/capnp/compat/http-over-capnp.c++
```diff
@@ -311,6 +312,10 @@ public:
   ~ClientRequestContextImpl() noexcept(false) {
     KJ_IF_SOME(ws, maybeWebSocket) {
       ws.cancel();
+      // Clear out selfRef here – the CapnpToKjWebSocketAdapter destructor tries to clear it too to
+      // remove the reference to itself, but the reference to that will no longer be valid once this
+      // destructor completes.
+      ws.selfRef = kj::none;
```
kentonv, line 318: `ws.cancel()` already done this -- see line 114.

kentonv:
The error referenced in EW-8586 looks very much like what I saw before #2102 fixed some things. Is it possible it came from an instance that didn't have #2102?

kentonv:
Ah yes, indeed, the specific ASAN error was reported exactly once on a staging machine before #2102 landed. This bug has already been fixed.

fhanau:
> The error referenced in EW-8586 looks very much like what I saw before #2102 fixed some things. Is it possible it came from an instance that didn't have #2102?

You're right - turns out that this only happened before #2102, it's plausible that this has already been fixed then.

PR #2121, opened 2024-08-09
3PH part 3: Basic three-party handoff of capabilities

kentonv:
This does not implement embargoes, so e-order may be violated.

As a reminder, none of this is enabled unless the VatNetwork implements new methods, so merging this won't impact existing users.

PR #2122, opened 2024-08-09
3PH part4: Embargoes for capabilities

kentonv:
This builds on #2121 by doing some refactoring and then implementing embargoes, so e-order is preserved through a handoff.

(This PR's merge target is set to #2121.)

### DIFF
--- a/c++/src/capnp/rpc-test.c++
+++ b/c++/src/capnp/rpc-test.c++
@@ -312,6 +312,26 @@ public:
       }
     }
 
+    void block() {
+      // Block all messages sent on this connection until unblock() is called.
+      auto paf = kj::newPromiseAndFulfiller<void>();
+      currentBlock = paf.promise.fork();
+      currentBlockFulfiller = kj::mv(paf.fulfiller);
+    }
+
+    void blockAfter(kj::Function<bool(rpc::Message::Reader)> predicate) {
+      // Automatically call block() immediately after a message matching `predicate` is sent
+      // on this connection.
+      blockAfterPredicate = kj::mv(predicate);
+    }
+
+    void unblock() {
+      // Allow blocked messages to proceed.
+      KJ_REQUIRE_NONNULL(currentBlockFulfiller)->fulfill();
+      currentBlock = kj::none;
+      currentBlockFulfiller = kj::none;
+    }
+
     void disconnect(kj::Exception&& exception) {
       messageQueue.rejectAll(kj::cp(exception));
       networkException = kj::mv(exception);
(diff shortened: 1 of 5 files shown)


c++/src/capnp/rpc.c++
```diff
@@ -1672,43 +1663,46 @@ private:
...
+              KJ_CASE_ONEOF(redirect, kj::Own<ClientHook>) {
+                // writeThirdPartyDescriptor() ended up deciding that it's not pointing at a
+                // third party after all, oops. This is unusual, but DeferredThirdPartyClient has
+                // some edge cases where this happens.
+
+                // Unfortunately, this will leave a hole in the message. This should be rare.
+                // TODO(perf): Since we're removing the last object allocated, the segment ought
+                //   to be able to reclaim the memory, but this doesn't appear to be implemented
+                //   at present. Fix that?
+                descriptor.disownThirdPartyHosted();
```
ObsidianMinor, line 1739: This whole thing feels odd. Why doesn't the client hook call back into the connection when it knows what it wants to do, instead of the connection assuming what's going to happen?

c++/src/capnp/rpc-test.c++
```diff
@@ -312,6 +312,26 @@ public:
       }
     }
 
+    void block() {
+      // Block all messages sent on this connection until unblock() is called.
```
ObsidianMinor, line 316: All!? Why all, and not the specific capabilities we need to embargo?

kentonv: The RPC protocol isn't really designed to handle messages out-of-order. Perhaps with careful analysis you could decide which messages are safe to reorder, but that would be hard. This is just a test, and blocking everything accomplishes what we want for the test.

c++/src/capnp/rpc.c++
kentonv: It could probably be factored a different way such that we don't pre-allocate `ThirdPartyHosted` prematurely. But it seemed like it would make the code more complex, and this edge case is rare, so I decided I'd rather just accept that a hole is left in the message in this case.

kentonv:
Rebased on #2121.

kentonv:
> I'm not fully understanding how a race condition where the Disembargo arrives before the Accept. Would the Disembargo be blocked until it finds a matching Accept?

`Disembargo` is a one-way message, so it's not "blocked" per se, but the third party does have to record that it received a disembargo with the given token, so that when the `Accept` comes later, it can see that it is already disembargoed, yes.

> I also have some concerns about DoS prevention for public services: with 3PH, if the above was true, would it not be possible to create a lot of pending Disembargo messages that never resolved? (the same could apply to Provide as well, and other parts of the RPC system)

To be honest, the Cap'n Proto protocol already has quite a few easier ways that you could cause someone to allocate a lot of memory, unless the peer's implementation explicitly implements some form of limits. Here, you might want to enforce a limit on the number of disembargo tokens that can be registered. Or you might want to implement a more general "memory limit", and add accounting of memory usage throughout the RPC system.

ObsidianMinor (approved):

PR #2123, opened 2024-08-11
Update docs on capnp tool to use convert subcommand

joshthoward:
See the screenshot to check how this is rendered by Jekyll Screen Shot 2024-08-06 at 7 28 32 PM


joshthoward:
FYI I'm not sure if we are building the site based on the v2 or master branch. I guessed the former, but let me know if I need to cherry pick this commit somewhere else.

PR #2126, opened 2024-08-15
Add declaration positions to SourceInfo and identifiers to RequestedFile for definition jumping

trickstar0301:
This Pull Request addresses the challenge of obtaining the declaration position (file, line, column) from parsedSchema for definition jumping.

I primarily added:
- Start/end byte position to SourceInfo
- An identifier list to RequestedFile

Jump-to-declaration currently skips using aliases, going directly to their definitions.

See also the discussion: 
https://github.com/capnproto/capnproto/discussions/2124


c++/src/capnp/schema.capnp
```diff
@@ -170,6 +170,9 @@ struct Node {
     }
   }
 
+  startByte @34 :UInt32;
```
kentonv, line 173: I'd argue these belong in `SourceInfo` -- in fact, it has a TODO for this.

trickstar0301: @kentonv 

I'm sorry.
I moved byte position definition to ``SourceInfo``.

trickstar0301:
@kentonv 

I intend to define byte position and referenceNodeId(ID of the referenced type) in ``Member struct`` of ``SourceInfo``.
```
    struct Member {
      docComment @0 :Text;
      # Doc comment on the member.
      startByte @1 :UInt32;
      endByte @2 :UInt32;
      referenceNodeId @3 :Id;
    }
```

I could set byte position of member in ``schema::Field::Builder getSchema()`` but I have no idea for referenceNodeId.
Could you give me some advice?

kentonv:
I don't think you need `referenceNodeId` since the type of a declaration is already in the schema (e.g. in `struct Field`).

Your extension will need to cross-reference both the `Node` and the corresponding `SourceInfo` to find all the information it needs. The reason these are two separate types is so that we can clearly separate stuff that is needed at runtime vs. stuff that is just for development and debugging.

trickstar0301:
@kentonv 
Thank you for your advice.
Initially, I added the byte position to Declaration because adding it to Expression is a bit more complicated.

c++/src/capnp/schema.capnp
```diff
@@ -195,6 +195,8 @@ struct Node {
     }
 
     # TODO(someday): Record location of the declaration in the original source code.
```
kentonv, line 197: I guess this comment can be removed?

c++/src/capnp/compiler/node-translator.c++
```diff
@@ -698,6 +698,10 @@ void NodeTranslator::compileNode(Declaration::Reader decl, schema::Node::Builder
   if (decl.hasDocComment()) {
     di.setDocComment(decl.getDocComment());
   }
+  if (decl.which() != Declaration::ANNOTATION) {
```
kentonv, line 701: Why not for annotations?

kentonv (commented):
So is this really all that's needed?

I guess this tells you where a type is defined. But you also need to figure out when the cursor is over a type name, in order to be able to offer to jump to it. How will that work?

c++/src/capnp/compiler/node-translator.c++
trickstar0301: This is just my mistake. I misunderstood, so I removed this unnecessary check.

c++/src/capnp/schema.capnp
trickstar0301: I removed comment.

trickstar0301:
@kentonv 

> So is this really all that's needed?
> 
> I guess this tells you where a type is defined. But you also need to figure out when the cursor is over a type name, in order to be able to offer to jump to it. How will that work?

OK, I was initially planning to split the Pull Request for declaration position and expression position, but if you prefer, I include the changes for adding expression position in this Pull Request as well. Thank you for pointing that out.

trickstar0301:
vscode-capnproto-jump

c++/src/capnp/schema.capnp
```diff
@@ -192,9 +192,16 @@ struct Node {
     struct Member {
       docComment @0 :Text;
       # Doc comment on the member.
+      bytePosition @1 :BytePosition;
+      brandBytePositions @2 :List(BytePosition);
```
trickstar0301, line 196: @kentonv 

I added the byte position of brand parameters. Could you tell me how to set parameters?

c++/src/capnp/schema.capnp
```diff
@@ -192,9 +192,16 @@ struct Node {
     struct Member {
       docComment @0 :Text;
       # Doc comment on the member.
+      bytePosition @1 :BytePosition;
```
kentonv, line 195: I'm not sure I understand how this would be used. It looks like you're filling this in with the position of the entire member declaration -- the whole line. So like if you have a field declaration:

```
struct Foo {
  bar @0 :Bar;
}
```

The `bytePosition` for `Foo.bar` would specify the entire text `bar @0 :Bar`?

But, when would you use this? I think the vast majority of the time, jump-to-definition wants to jump to a type name, not a member?

Or are you actually implementing jump-to-definition for things like:

`const foo :Foo = (bar = (i = 2));`

Here, I suppose someone might want to put the cursor on `bar` and jump-to-definition?

But this feels like a fairly unusual thing. More often I think people want to jump to type definitions.

That is, when we have the field declaration `bar @0 :Bar;`, you might put your cursor specifically on the type, `Bar`, and jump-to-definition from there. But how will the language server know that `Bar` is a type declaration here?

Perhaps instead of `bytePosition` as you've defined it, you really want something like `typeBytePosition`, which is the byte position of the _type_?

Of course, the type itself might be an expression, like `SomeModule.ParentType.Bar` or `SomeGeneric(Foo, Bar)`. So there might be multiple symbols to jump _from_. How should we represent this?

I kind of wonder if this is the right data structure for this. Maybe we really need to have the compiler output an additional list of all symbols that can be jumped-to, like:

```
struct Identifier {
  startByte @0 :UInt32;
  endByte @1 :UInt32;

  union {
    typeId @2 :UInt64;
    # Identifier refers to a type. This is the type ID.

    member :group {
      # Identifier refers to a member of a type.
      parentTypeId @3 :UInt64;
      ordinal @4 :UInt16;
    }
  }
}

identifiers :List(Identifier);
# Positions of all identifiers in the file, and what they refer to.
```

kentonv (commented):
Sorry for the long delay, I have been swamped with work leading up to Cloudflare's Birthday Week (last week) when we launched a bunch of stuff. It's hard to get to OSS reviews when my inbox is full of urgent stuff. :/

Hoping to have a little less pressure now that Birthday Week is over. I'm really interested in using this extension so do want to get this figured out!

c++/src/capnp/schema.capnp
trickstar0301: I might have misunderstand.  
I was considering supporting the following case.
```
struct Map(Key, Value) {
  entries @0 :List(Entry);
  struct Entry {
    key @0 :Key;
    value @1 :Value;
  }
}

struct People {
  byName @0 :Map(Text, Person);
  # Maps names to Person instances.
}
```

Anyway, with your schema proposal, this can also be resolved, so there is no problem.

c++/src/capnp/schema.capnp
trickstar0301: @kentonv 

Thank you so much! I fully agree with your proposal.
I will replace the schema definition with your suggestion.  
Could you give me a hint on which files I should modify to fill in this information? In my understanding, modifying the node translator alone doesn't seem sufficient.

c++/src/capnp/compiler/node-translator.c++
```diff
@@ -698,6 +698,8 @@ void NodeTranslator::compileNode(Declaration::Reader decl, schema::Node::Builder
   if (decl.hasDocComment()) {
     di.setDocComment(decl.getDocComment());
   }
+  di.getBytePosition().setStartByte(decl.getStartByte());
```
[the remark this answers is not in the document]
trickstar0301: Thank you for the advice.  I should avoid redundant pointer re-validation.

c++/src/capnp/schema.capnp
kentonv: Ah yes, but what if you have:

```
someField @0 :SomeType(Foo, Bar(Baz(Qux), Corge(Grault, Garply)))
```

As you can see, it's not just a simple list of brand parameters. The type can be a whole expression tree. So I think the separate "identifiers list" approach is what we need here.

c++/src/capnp/schema.capnp
kentonv: So the trick is, everywhere where we are "resolving" an identifier, we need to make note of the identifier location and what it resolved to.

I think the vast majority of cases boil down to `BrandScope::compileDeclExpression()`. Here, we call `resolver.resolve()` or similar functions to handle RELATIVE_NAME, ABSOLUTE_NAME, and IMPORT expression types. In each case, we have a `LocatedText` that we're using as the input. You can see where we use `errorReporter.addErrorOn()` to report errors -- and the error of course reports exactly where the identifier was, so this proves we have the info we need.

The only question is: Where do we report the resolution result, so that it ends up in the identifier table in the source info?

I think the best way would be do add a method to the `ErrorReporter` interface, like:

```
virtual void reportResolution(uint32_t startByte, uint32_t endByte, uint64_t typeId) {}
```

(Later on we might need to expand the third parameter to support resolving to non-type members and to generic type parameters, but let's not worry about those for now (don't report anything in those cases)... let's just get basic type lookup working first.)

`ModuleLoader::ModuleImpl` will need to implement `reportResolution()` by appending to some list of all resolutions seen in the module.

The last question is, where do we actually put the identifier list in the schemas? `Node.SourceInfo` is associated with a specific node, but our identifier list will be for the whole file. I think there are two options:
1. We could say that the `Node.SourceInfo` corresponding to the file node gets the identifier list. All other nodes' `SourceInfo`s contain empty lists.
2. Define a new type `struct FileSourceInfo`, which contains file-level optional source information, and then place this in `CodeGeneratorRequest.RequestedFile`.

I think I slightly lean towards option 2 but I don't have a strong opinion.

trickstar0301: Thank you.  I also like option 2.
At first, I updated the `schema.capnp`. 
Does it align with your intentions?

trickstar0301: I added `reportResolution` for `ErrorReporter` interface and implement on `ModuleLoader::ModuleImpl`.
cf50a67c85ddd20230204c93cdf57c17f15ff5b8

trickstar0301: @kentonv 

I have a few comments:

1. **Resolutions Member in `ModuleLoader::ModuleImpl`**: I believe we need to introduce a `resolutions` member variable in `ModuleLoader::ModuleImpl` to store the list of resolutions.

2. **Getter for Resolutions**: We’ll also need a getter function for `ModuleLoader::ModuleImpl`. The signature could be something like:

   ```cpp
   kj::Vector<schema::CodeGeneratorRequest::RequestedFile::FileSourceInfo::Identifier::Reader> getResolutions();
   ```

   This method could be added as a virtual method to the `Module` class, allowing it to be accessed from the `CompilerMain` class.

3. **Schema's `FileSourceInfo`**: Should the schema’s `FileSourceInfo` be set in the `kj::MainBuilder::Validity generateOutput()` method in the `CompilerMain` class? Is this the correct approach?

kentonv: Hmm, I think the `resolutions` member of `ModuleImpl` should contain some C++ representation the resolutions, not Cap'n Proto representation, since allocating lots of little capnp messages would be inefficient. So we need a type like:

```
struct Resolution {
  uint32_t startByte;
  uint32_t endByte;

  struct Type { uint64_t typeId; };
  struct Member {
    uint64_t parentTypeId;
    uint16_t ordinal;
  };

  kj::OneOf<Type, Member>;
};
```

In fact this type should probably be the input type for `reportResolution()`, just to make things easier.

And then I think `getResolutions()` should return an array of that type, instead of trying to return struct readers:

```
kj::ArrayPtr<const Resolution> getResolutions();
```

Otherwise, everything you say makes sense, yes.

c++/src/capnp/compiler/module-loader.c++
```diff
@@ -272,6 +275,14 @@ kj::Maybe<kj::Array<const byte>> ModuleLoader::Impl::readEmbedFromSearchPath(kj:
   return nullptr;
 }
 
+void ModuleLoader::Impl::reportResolution(uint32_t startByte, uint32_t endByte, uint64_t typeId) {
+  auto identifier = kj::heap<capnp::MallocMessageBuilder>().get()->initRoot<schema::CodeGeneratorRequest::RequestedFile::FileSourceInfo::Identifier>();
```
kentonv, line 279: A `Reader` is like a pointer. If the message it is pointing into is destroyed, the `Reader` is no longer valid. Here, the `MassogMessageBuilder` will be destroyed at the end of the statement, since it's not being stored anywhere, so `identifier` is left dangling.

Per my other comment we should probably just store C++ structs instead of capnp structs.

c++/src/capnp/compiler/module-loader.h
```diff
@@ -53,6 +53,8 @@ class ModuleLoader {
   // Same as SchemaParser::setFileIdsRequired(). If set false, files will not be required to have
   // a top-level file ID; if missing a random one will be assigned.
 
+  void reportResolution(uint32_t startByte, uint32_t endByte, uint64_t typeId);
```
kentonv, line 56: I think this method actually has to be on `ModuleImpl`, not `ModuleLoader`.

c++/src/capnp/schema.capnp
trickstar0301: @kentonv 
I modified the source code as you suggested.  
Currently, as a trial, I have implemented `errorReporter.reportResolution()` only in the `Expression::RELATIVE_NAME` case. Which other cases should I implement it for?

kentonv: Seems like it would be useful to report ABSOLUTE_NAME and IMPORT as well.

ABSOLUTE_NAME is when a name is prefixed by a `.` to indicate it should be looked up starting from the global scope.

IMPORT is, of course, `import` directives -- these will resolve to a file.

I think people will expect to be able to jump-to-definition on both of these.

c++/src/capnp/compiler/error-reporter.h
```diff
@@ -54,6 +54,8 @@ class ErrorReporter {
   // is to inhibit the reporting of errors which may have been caused by previous errors, or to
   // allow the compiler to bail out entirely if it gets confused and thinks this could be because
   // of previous errors.
+
+  virtual void reportResolution(uint32_t startByte, uint32_t endByte, uint64_t typeId){};
```
kentonv, line 58: Perhaps this should take `Resolution` as its parameter?

c++/src/capnp/compiler/module-loader.c++
```diff
@@ -207,6 +207,18 @@ public:
     return loader.getErrorReporter().hadErrors();
   }
 
+  void reportResolution(uint32_t startByte, uint32_t endByte, uint64_t typeId) override{
+    Resolution resolution;
+    resolution.startByte = startByte;
+    resolution.endByte = endByte;
+    resolution.target = Resolution::Type{typeId};
+    resolutions.add(kj::mv(resolution));
```
kentonv, line 215:
```suggestion
    resolutions.add(Resolution {
      .startByte = startByte,
      .endByte = endByte,
      .target = Resolution::Type { typeId },
    });
```

c++/src/capnp/compiler/error-reporter.h
kentonv, line 58:
```suggestion
  virtual void reportResolution(uint32_t startByte, uint32_t endByte, uint64_t typeId) {};
  // Report that an identifier parsed from the input resolved to a particular declaration. This may
  // be useful, for example, to implement a jump-to-definition in a code editor.
  //
  // This is not an error, but is reported through ErrorReporter because ErrorReporter tends to be
  // passed to all the right places to collect this information.
```

c++/src/capnp/compiler/compiler.h
```diff
@@ -48,6 +61,9 @@ class Module: public ErrorReporter {
 
   virtual kj::Maybe<kj::Array<const byte>> embedRelative(kj::StringPtr embedPath) = 0;
   // Read and return the content of a file specified using `embed`.
+  virtual kj::ArrayPtr<const Resolution> getResolutions(){
+    return kj::ArrayPtr<const Resolution>();
+  };
```
kentonv, line 66:
```suggestion

  virtual kj::ArrayPtr<const Resolution> getResolutions() { return nullptr; }
  // Get the list of all resolutions reported using ErrorHandler::reportResolution().
```

c++/src/capnp/compiler/error-reporter.h
trickstar0301: I agree that. I will replace with `Resolution`.   
In addition, I will move declaration of `Resolution` to `error-reporter.h`.

c++/src/capnp/compiler/module-loader.c++
trickstar0301: Thank you for your suggestion.
I will change reportResolution can take `Resolution` as a parameter , so I will refer to it in the implementation on the caller side.

c++/src/capnp/compiler/module-loader.c++
trickstar0301: Thank you for your suggestion. I apologize for my lack of knowledge. It has been corrected.

c++/src/capnp/compiler/generics.c++
```diff
@@ -584,6 +600,11 @@ kj::Maybe<BrandedDecl> BrandScope::compileDeclExpression(
     case Expression::IMPORT: {
       auto filename = source.getImport();
       KJ_IF_MAYBE(decl, resolver.resolveImport(filename.getValue())) {
+        errorReporter.reportResolution(Resolution {
```
trickstar0301, line 603: Note: 
For example, in the following schema:
```
using Common = import "common.capnp";
```
The resolution is:
```
import "common.capnp"
```

c++/src/capnp/compiler/generics.c++
```diff
@@ -574,7 +582,15 @@ kj::Maybe<BrandedDecl> BrandScope::compileDeclExpression(
     case Expression::ABSOLUTE_NAME: {
       auto name = source.getAbsoluteName();
       KJ_IF_MAYBE(r, resolver.getTopScope().resolver->resolveMember(name.getValue())) {
-        return interpretResolve(resolver, *r, source);
+        auto result = interpretResolve(resolver, *r, source);
+        if (r->is<Resolver::ResolvedDecl>()){
+          errorReporter.reportResolution(Resolution {
```
trickstar0301, line 587: I'm not confident since it's not handling recursion.

trickstar0301: @kentonv 
Could you review it again when you have time?

c++/src/capnp/compiler/compiler.h
```diff
@@ -32,6 +32,8 @@ CAPNP_BEGIN_HEADER
 namespace capnp {
 namespace compiler {
 
+
```
kentonv, line 35: Remove spurious whitespace.

c++/src/capnp/compiler/capnp.c++
```diff
@@ -502,6 +502,23 @@ public:
...
           *sourceFiles[i].module, Orphanage::getForMessageContaining(requestedFile)));
+      // Add FileSourceInfo
+      auto fileSourceInfo = requestedFile.initFileSourceInfo();
+      const auto resolutions = sourceFiles[i].module->getResolutions();
+      auto identifiers = fileSourceInfo.initIdentifiers(resolutions.size());
+      for (size_t j = 0; j < resolutions.size(); j++) {
+        auto identifier = identifiers[j];
+        identifier.setStartByte(resolutions[j].startByte);
+        identifier.setEndByte(resolutions[j].endByte);
+        if (resolutions[j].target.is<Resolution::Type>()) {
```
kentonv, line 513: Can you use a KJ_SWITCH_ONEOF here?

```
KJ_SWITCH_ONEOF(resolutions[j].target) {
  KJ_CASE_ONEOF(type, Resolution::Type) {
    identifier.setTypeId(type.typeId);
  }
  KJ_CASE_ONEOF(member, Resolution::Member) {
    ...
  }
}
```

c++/src/capnp/compiler/generics.c++
kentonv: You just mean, the `import` keyword is considered part of the identifier, so jump-to-declaration will work even when the cursor is on `import` and not actually on `"common.capnp"`? I think that's fine.

c++/src/capnp/compiler/generics.c++
kentonv: This will correctly handle the `Foo` in `foo @0 :.Foo`, and `RELATIVE_NAME` will correctly handle `Bar` in `bar @1 :Bar`.

But if you have `baz :2 :Baz.Qux.Corge`, I think `Qux` and `Corge` aren't currently handled. To handle them you need to add code to the `Expression::MEMBER` case, below.

kentonv (commented):
Sorry for the long delay, I'm always swamped. :( Was also on vacation for a while.

We are getting close here. I have a couple minor comments, but also we need to handle `Expression::MEMBER`, as I think you noticed.

Do you have this code working with your plugin already?

c++/src/capnp/compiler/compiler.h
trickstar0301: I fixed it.

c++/src/capnp/compiler/capnp.c++
trickstar0301: Thank you for your advice. I modified it as you mentioned.

c++/src/capnp/compiler/generics.c++
trickstar0301: @kentonv 

That makes sense. My plugin is working as you described.
Could you tell me how to extract Type ID from `memberDecl`?
I don't know the way to access to the `ResolvedDecl` of `MEMBER`. Maybe I need to call `asResolveResult()`.

trickstar0301:
> Sorry for the long delay, I'm always swamped. :( Was also on vacation for a while.
> 
> We are getting close here. I have a couple minor comments, but also we need to handle `Expression::MEMBER`, as I think you noticed.
> 
> Do you have this code working with your plugin already?

No worries at all! I hope you had a great vacation!
The plugin is working well, except for handling Expression::MEMBER.  I also hope the definition jump plugin helps make your work a little easier!

c++/src/capnp/compiler/generics.c++
kentonv: Hmm, looks like we should add a method `BrandedDecl::getGenericTypeId()`, with the documentation:

> If this declaration points to a type, gets its type ID. Keep in mind that this drops the brand, that is, if the type is the result of applying a generic to some type parameters, this returns only the ID of the underlying generic type, which loses information about the type parameters.

That said, it occurs to me that we've designed this in a way where jump-to-declaration will skip aliases. That is, if the symbol points to an alias defined with `using`, we will jump straight through it and to the thing that alias points at. That's probably OK most of the time, and doing better would require deeper refactoring I think. So, let's just stick with it, but I wanted to point it out as an issue.

trickstar0301: I added the `kj::Maybe<uint64_t> BrandedDecl::getGenericTypeId()` with the documentation.
Thank you for pointing out the behavior regarding aliases.

trickstar0301:
@kentonv 

I believe the implementation is complete, so I would greatly appreciate it if you could review the entire code again.

trickstar0301:
demo

kentonv:
Hey @trickstar0301, it looks like you had a bad rebase, this PR now has 43 commits and most of them don't look like they are yours. Can you clean that up?

trickstar0301:
> Hey @trickstar0301, it looks like you had a bad rebase, this PR now has 43 commits and most of them don't look like they are yours. Can you clean that up?

@kentonv 
Apologies for the messy history. 
I've squashed my commits and cleaned up the PR to include only my intended changes.
Let me know if further adjustments are needed. Thank you.

trickstar0301:
@kentonv 

Hi, I’ve implemented a basic Language Server for Cap'n Proto in my repository. If you’d like to see how it works, feel free to check it out here:
[capnp-ls: feature/initial-setup](https://github.com/trickstar0301/capnp-ls/tree/feature/initial-setup).

It includes initial LSP functionalities such as definition jumping and symbol analysis. I'd appreciate any feedback or suggestions!

c++/src/capnp/compiler/generics.c++
```diff
@@ -564,7 +564,15 @@ kj::Maybe<BrandedDecl> BrandScope::compileDeclExpression(
...
 
       KJ_IF_MAYBE(r, resolver.resolve(nameValue)) {
-        return interpretResolve(resolver, *r, source);
+        auto result = interpretResolve(resolver, *r, source);
+        if (r->is<Resolver::ResolvedDecl>()){
+          errorReporter.reportResolution(Resolution {
+            .startByte = source.getStartByte(),
+            .endByte = source.getEndByte(),
+            .target = Resolution::Type { r->get<Resolver::ResolvedDecl>().id },
+          });
```
kentonv, line 573: Ugh, I forgot that the master branch is still on C++17 and it doesn't support designated initializers (although only MSVC actually complains)...

```suggestion
          errorReporter.reportResolution(Resolution {
            /* .startByte = */ source.getStartByte(),
            /* .endByte = */ source.getEndByte(),
            /* .target = */ Resolution::Type { r->get<Resolver::ResolvedDecl>().id },
          });
```

c++/src/capnp/compiler/generics.c++
```diff
@@ -574,7 +582,15 @@ kj::Maybe<BrandedDecl> BrandScope::compileDeclExpression(
...
       auto name = source.getAbsoluteName();
       KJ_IF_MAYBE(r, resolver.getTopScope().resolver->resolveMember(name.getValue())) {
-        return interpretResolve(resolver, *r, source);
+        auto result = interpretResolve(resolver, *r, source);
+        if (r->is<Resolver::ResolvedDecl>()){
+          errorReporter.reportResolution(Resolution {
+            .startByte = source.getStartByte(),
+            .endByte = source.getEndByte(),
+            .target = Resolution::Type { r->get<Resolver::ResolvedDecl>().id },
+          });
```
kentonv, line 591:
```suggestion
          errorReporter.reportResolution(Resolution {
            /* .startByte = */ source.getStartByte(),
            /* .endByte = */ source.getEndByte(),
            /* .target = */ Resolution::Type { r->get<Resolver::ResolvedDecl>().id },
          });
```

c++/src/capnp/compiler/generics.c++
```diff
@@ -584,6 +600,11 @@ kj::Maybe<BrandedDecl> BrandScope::compileDeclExpression(
     case Expression::IMPORT: {
       auto filename = source.getImport();
       KJ_IF_MAYBE(decl, resolver.resolveImport(filename.getValue())) {
+        errorReporter.reportResolution(Resolution {
+          .startByte = source.getStartByte(),
+          .endByte = source.getEndByte(),
+          .target = Resolution::Type { decl->id },
+        });
```
kentonv, line 607:
```suggestion
        errorReporter.reportResolution(Resolution {
          /* .startByte = */ source.getStartByte(),
          /* .endByte = */ source.getEndByte(),
          /* .target = */ Resolution::Type { decl->id },
        });
```

c++/src/capnp/compiler/generics.c++
```diff
@@ -635,6 +656,13 @@ kj::Maybe<BrandedDecl> BrandScope::compileDeclExpression(
       KJ_IF_MAYBE(decl, compileDeclExpression(member.getParent(), resolver, implicitMethodParams)) {
         auto name = member.getName();
         KJ_IF_MAYBE(memberDecl, decl->getMember(name.getValue(), source)) {
+          KJ_IF_MAYBE(id, memberDecl->getGenericTypeId()) {
+            errorReporter.reportResolution(Resolution {
+              .startByte = source.getStartByte(),
+              .endByte = source.getEndByte(),
+              .target = Resolution::Type { *id },
+            });
```
kentonv, line 664:
```suggestion
            errorReporter.reportResolution(Resolution {
              /* .startByte = */ source.getStartByte(),
              /* .endByte = */ source.getEndByte(),
              /* .target = */ Resolution::Type { *id },
            });
```

c++/src/capnp/compiler/generics.h
```diff
@@ -141,6 +141,17 @@ class BrandedDecl {
   kj::String toString();
   kj::String toDebugString();
 
+  kj::Maybe<uint64_t> getGenericTypeId() {
+  // If this declaration points to a type, gets its type ID. Keep in mind that this
+  // drops the brand, that is, if the type is the result of applying a generic to
+  // some type parameters, this returns only the ID of the underlying generic type,
+  // which loses information about the type parameters.
```
kentonv, line 148:
```suggestion
    // If this declaration points to a type, gets its type ID. Keep in mind that this
    // drops the brand, that is, if the type is the result of applying a generic to
    // some type parameters, this returns only the ID of the underlying generic type,
    // which loses information about the type parameters.
```

kentonv:
OK if CI passes this time I will merge...

kentonv:
Ugh I failed to commit all the suggestions, trying again.

c++/src/capnp/compiler/node-translator.c++
71: Sorry for resurrecting an old PR, but there is a similar check above which made it through:

https://github.com/capnproto/capnproto/blob/ff191c66c2bbff428c10923b5d4032087b592fe0/c%2B%2B/src/capnp/compiler/node-translator.c%2B%2B#L694-L697

@trickstar0301 Was it intentional?

PR #2129, opened 2024-08-20
exposing http-over-capnp headers capnpToKj

mikea:


kentonv (approved):
Seems fine, curious what it's for.

mikea:
> Seems fine, curious what it's for.

we're using http-over-capnp definitions in new api and I want to expose them as familiar api/http.h objects. Headers is the most complicated to convert in my case.

PR #2139, opened 2024-09-11
add more constexpr stringptr functions

anonrig:
Converts more operators&functions to constexpr

### DIFF
--- a/c++/src/kj/array.h
+++ b/c++/src/kj/array.h
@@ -169,14 +169,14 @@ public:
     return ptr[index];
   }
 
-  inline const T* begin() const KJ_LIFETIMEBOUND { return ptr; }
-  inline const T* end() const KJ_LIFETIMEBOUND { return ptr + size_; }
-  inline const T& front() const KJ_LIFETIMEBOUND { return *ptr; }
-  inline const T& back() const KJ_LIFETIMEBOUND { return *(ptr + size_ - 1); }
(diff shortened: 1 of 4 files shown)


jasnell (approved):

kentonv:
MSVC errors seem legit although it's complaining about missing C++20 which I thought we already enabled by default. Maybe there's a problem with the MSVC CI config?

anonrig:
> Maybe there's a problem with the MSVC CI config?

This is kind of what I had in mind, but other pull-requests doesn't seem to have this problem. Maybe the changes on `common.h` is causing this error? Did we enable C++20 only on specific files, or omitted it for common.h?

kentonv:
Actually the C++20 errors appear to be a red herring, I don't think those are actually breaking the build. (It looks like the samples are built with the wrong options, but then errors from building the samples end up being ignored?)

What's actually breaking is: `[ FAIL ] common-test.c++:942: ArrayPtr equality comparisons for UTF-8`

I think this is actually a problem introduced in this change. Previously, we used memcmp() for char arrays, but now MSVC compares the chars with a for loop. The problem is that chars are signed (on this platform, at least), but for comparison purposes we need to treat them as unsigned, which memcmp does.

kentonv:
I think you'll find the same test fails on Linux if you disable the `#if __has_feature(cxx_constexpr_string_builtins)` part.

jasnell (approved):

kentonv (approved):

PR #2143, opened 2024-09-20
[nfc] extracting bazel capnp_gen rule

mikea:
Rust and js rules basically copy it and are not compatible with each other. By having reusable implementation not only language rules are simpler, but they will also support language-independent deps because they will share provider.

No functional change at all, just a copy-paste + attr rename


fhanau (approved):
LGTM

PR #2145, opened 2024-09-27
[nfc] rust_capnp_library rule

mikea:
No-op when gen_rust is set to false. Needs `--//src/capnp:gen_rust=true`


c++/src/capnp/compat/BUILD.bazel
```diff
@@ -49,6 +50,30 @@ cc_capnp_library(
...
+    }),
+    visibility = ["//visibility:public"],
+)
+
+rust_capnp_library(
+    name = "byte-stream_capnp_rust",
+    srcs = ["byte-stream.capnp"],
+    crate_name = "byte_stream_capnp",
+    src_prefix = "src",
+    target_compatible_with = select({
```
fhanau, line 70: Nit: Only realized this recently – using syntax like `target_compatible_with = ["//src/capnp:gen_rust_true"],` can be used to express this more concisely.

c++/src/capnp/rust_capnp_library.bzl
```diff
@@ -0,0 +1,82 @@
...
+def rust_capnp_library(
+        name,
+        srcs,
+        crate_name,
+        data = [],
+        deps = [],
+        src_prefix = "",
+        tags = [],
+        visibility = None,
+        target_compatible_with = None,
```
fhanau, line 18: Nit: target_compatible_with and visibility can be omitted if you also add `**kwargs` to the `capnp_gen()` invocation.

fhanau (approved):

c++/src/capnp/rust_capnp_library.bzl
mikea: This is not exactly what I want: rust_library has lots of additional arguments that don't need to be passed to other targets here.

c++/src/capnp/compat/BUILD.bazel
mikea: I don't know why, but it works differently: with `select` statement it doesn't require rust toolchain, but if I put what you suggest it does:

```
diff --git a/c++/src/capnp/compat/BUILD.bazel b/c++/src/capnp/compat/BUILD.bazel
index b47abbe0..4e5fd134 100644
--- a/c++/src/capnp/compat/BUILD.bazel
+++ b/c++/src/capnp/compat/BUILD.bazel
@@ -55,10 +55,7 @@ rust_capnp_library(
     srcs = ["json.capnp"],
     crate_name = "json_capnp",
     src_prefix = "src",
-    target_compatible_with = select({
-        "//src/capnp:gen_rust_true": [],
-        "//conditions:default": ["@platforms//:incompatible"],
-    }),
+    target_compatible_with = ["//src/capnp:gen_rust_true"],
```

```
ERROR: /home/maizatskyi/src/github.com/capnproto/capnproto/c++/src/capnp/compat/BUILD.bazel:53:19: While resolving toolchains for target //src/capnp/compat:json_capnp_rust (c671c1d): No matching toolchains found for types @@rules_rust//rust:toolchain_type.
To debug, rerun with --toolchain_resolution_debug='@@rules_rust//rust:toolchain_type'
If platforms or toolchains are a new concept for you, we'd encourage reading https://bazel.build/concepts/platforms-intro.
ERROR: no such package '@@crates_vendor//': The repository '@@crates_vendor' could not be resolved: Repository '@@crates_vendor' is not defined
ERROR: /home/maizatskyi/src/github.com/capnproto/capnproto/c++/src/capnp/compat/BUILD.bazel:53:19: no such package '@@crates_vendor//': The repository '@@crates_vendor' could not be resolved: Repository '@@crates_vendor' is not defined and referenced by '//src/capnp/compat:json_capnp_rust_gen'
```

I'll keep the `select()` form since it doesn't require setting up rust toolchain.

PR #2146, opened 2024-09-29
Check if atomic intrinsics are implemented by libatomic

MingcongBai:
By testing whether a simple atomic-enabled C++ source (taken from CMake: /Source/Checks/cm_cxx_atomic.cxx) builds, determine whether an extra `-latomic' linkage is needed (if not implemented by built-in compiler intrinsics, something like GCC would use what they call a "library fallback", i.e., libatomic). [^1]

On an i486-target system, building the aforementioned source would return the following error (with g++ 13.2.0):

ld: /tmp/cccsCQSO.o: in function `main':
test.cxx:(.text+0xaa): undefined reference to `__atomic_load_8'

Whereas on platforms with atomics implemented by built-in compiler intrinsics, the source would build successfully. Use this mechanism to determine if the said extra linkage is needed.

[^1]: https://gcc.gnu.org/onlinedocs/libstdc++/manual/ext_concurrency_impl.html

### DIFF
--- a/c++/src/capnp/CMakeLists.txt
+++ b/c++/src/capnp/CMakeLists.txt
@@ -111,6 +111,21 @@ if(NOT CAPNP_LITE)
   install(FILES ${capnp-rpc_headers} ${capnp-rpc_schemas} DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/capnp")
 endif()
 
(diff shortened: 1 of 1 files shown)


kentonv:
Seems reasonable but I don't know cmake. @harrishancock what do you think?

kentonv:
Also do we need a similar check for the autotools build?

PR #2148, opened 2024-10-01
Fix non-monotonic time on all Mac machines

fabiorossetto:
The issue was originally observed only on Macs with intel processor, and it was believed to affect only those. However the issue has been observed on a Mac with M2 processor, thus the need to relax the preprocessor directive.


kentonv:
Thanks!

PR #2152, opened 2024-10-04
Added generic compare function for uses of memcmp that returns an int

AdityaAtulTewari:


kentonv:
Where do we need this?

AdityaAtulTewari:
We need this in edgeworker for fake-server.

kentonv:
That's test code, and kind of unusual. I would just leave it as it is. Or it could use separate < and == operations since perf isn't that critical there anyway.

I'm hesitant to introduce `cmp()` to our public API without a strong use case. I think if we must do it we need to re-open the discussion about whether we should really be implementing `operator<=>` instead, but I'd rather just avoid it.

AdityaAtulTewari:
When I tried replacing it in fake-server with < and ==, earlier it caused a lot of timeout failures. I prefer `operator<=>` although this does mean we need to involve another header. I do think not having cmp leaves the api somewhat incomplete in-case someone truly wants different behavior for each of the three directions on the comparator.

PR #2153, opened 2024-10-05
Delete CharArrayCompare because ArrayPtr comparators exist.

AdityaAtulTewari:


c++/src/kj/main.c++
```diff
@@ -245,12 +245,7 @@ struct MainBuilder::Impl {
 
   struct CharArrayCompare {
```
kentonv, line 246: You can just delete CharArrayCompare altogether, and remove the type parameter from `std::map` below, since comparing using `operator<` is the default.

PR #2154, opened 2024-10-05
Make kj::encodeHex() go brrrrr

fhanau:
Optimize `kj::encodeHex()` by allocating the string up front to avoid allocating memory for every input byte. This drastically improves performance, even when optimization is enabled. A number of downstream tests that include large text files benefit from this.
It is trivial to get another performance improvement by looking up a byte at a time in a larger lookup table but this is "good enough" for our needs.

### DIFF
--- a/c++/src/kj/encoding.c++
+++ b/c++/src/kj/encoding.c++
@@ -371,9 +371,12 @@ static Maybe<uint> tryFromOctDigit(char c) {
 }  // namespace
 
 String encodeHex(ArrayPtr<const byte> input) {
-  return strArray(KJ_MAP(b, input) {
-    return heapArray<char>({HEX_DIGITS[b/16], HEX_DIGITS[b%16]});
-  }, "");
+  auto result = heapString(input.size() * 2);
+  for (auto i: kj::indices(input)) {
+    result[i*2] = HEX_DIGITS[input[i] / 16];
+    result[i*2+1] = HEX_DIGITS[input[i] % 16];
+  }
+  return result;
 }
 
 EncodingResult<Array<byte>> decodeHex(ArrayPtr<const char> text) {


c++/src/kj/encoding.c++
```diff
@@ -410,7 +414,7 @@ String encodeUriComponent(ArrayPtr<const byte> bytes) {
     } else {
       result.add('%');
       result.add(HEX_DIGITS_URI[b/16]);
-      result.add(HEX_DIGITS_URI[b%16]);
```
fhanau, line 413: @kentonv I doubt it matters much when using modern compilers and modern hardware but I prefer `b & 15` over `b % 16` to get the bottom four bits of a uint8_t/kj::byte. I changed this in encodeHex and elsewhere in the file, but ready to revert it if you disagree.

kentonv: The compiler will certainly produce the same code either way. I prefer `%16` because it's easier for a human to think about, and it corresponds to the `/16` on the previous line (which you might argue should be `>>4` instead, but again, `/16` is easier for a human to read).

c++/src/kj/encoding.c++
```diff
@@ -371,9 +371,13 @@ static Maybe<uint> tryFromOctDigit(char c) {
 }  // namespace
 
 String encodeHex(ArrayPtr<const byte> input) {
-  return strArray(KJ_MAP(b, input) {
-    return heapArray<char>({HEX_DIGITS[b/16], HEX_DIGITS[b%16]});
-  }, "");
+  // Do not use KJ_MAP here, we want to do this with a single memory allocation.
```
kentonv, line 374: Wow I'm really surprised to see this code was written to do O(n) allocations! It looks like I wrote it, I don't know what I was thinking.

I think we can delete this comment -- there's no reason someone would expect to see KJ_MAP here, so I think the comment is more distracting than useful.

kentonv (approved):

jclee:
Nice catch!  I took a scan through other capnp/workerd code, and it looks like we have a bunch of different code doing hex encoding/decoding, but looks like most places preallocate the string.

It does look like `warnIfBadHeaderString()` in the workerd source might have similar inefficiencies -- although it's in a code path that would uncommonly get hit (only in inspected workers handling weird header content).

fhanau:
> Nice catch! I took a scan through other capnp/workerd code, and it looks like we have a bunch of different code doing hex encoding/decoding, but looks like most places preallocate the string.
> 
> It does look like `warnIfBadHeaderString()` in the workerd source might have similar inefficiencies -- although it's in a code path that would uncommonly get hit (only in inspected workers handling weird header content).

Yup, that looks pretty inefficient too. Thankfully examples like this are rare. I initially found this while looking at reducing the total amount of memory allocations – there's a number of spots where hot path memory allocations can be avoided but no O(n) behavior as far as I can tell.

PR #2156, opened 2024-10-06
Refactor async-io to use ArrayPtr Comparator

AdityaAtulTewari:


c++/src/kj/async-io-unix.c++
```diff
@@ -902,7 +902,7 @@ public:
     if (addrlen < other.addrlen) return true;
     if (addrlen > other.addrlen) return false;
 
-    return memcmp(&addr.generic, &other.addr.generic, addrlen) < 0;
+    return kj::asBytes(addr).first(addrlen) < kj::asBytes(other.addr).first(addrlen);
```
mikea, line 905: you can add a comment:

```
// addrlen == other.addrlen at this point
```

c++/src/kj/async-io-win32.c++
```diff
@@ -474,7 +474,7 @@ public:
     if (addrlen < other.addrlen) return true;
     if (addrlen > other.addrlen) return false;
 
-    return memcmp(&addr.generic, &other.addr.generic, addrlen) < 0;
+    return kj::asBytes(addr).first(addrlen) < kj::asBytes(other.addr).first(addrlen);
```
mikea, line 477: same

PR #2157, opened 2024-10-11
convert DelayedCloseWebSocket to coros

mikea:
I am debugging websocket issue and stack trace doesn't make much sense. Maybe after getting rid of lambdas it will become clearer.

Not much but I have to start somewhere.

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -2853,7 +2853,7 @@ public:
             }
             bool addNullTerminator = true;
             // We want to add the null terminator when receiving a TEXT message.
-            auto decompressedOrError = decompressor.processMessage(message, originalMaxSize, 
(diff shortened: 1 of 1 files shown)


danlapid (approved):

c++/src/kj/compat/http.c++
```diff
@@ -2885,7 +2885,6 @@ public:
               // We must reset context on each message.
               decompressor.reset();
             }
-
```
kentonv, line 2888: Why the randomly-deleted blank lines?

kentonv (approved):
This also increases allocations FWIW.

But I'm mostly indifferent.

c++/src/kj/compat/http.c++
mikea: these lines consisted of spaces, so that probably threw something off. Not sure myself, but fixed.

PR #2158, opened 2024-10-11
convert BlockedSend and BlockedPump to coros

mikea:


danlapid (approved):

kentonv:
IIRC @harrishancock did some conversion like this once upon a time and found that it actually negatively impacted performance? Essentially because the `.then()`s use PromiseArena but coroutine frames can't.

You could run the "regular HTTP" benchmark in http-over-capnp-perf-test to find out. Note that if you compile it with the right flags it'll even give you a count of mallocs which is a lot more reliable than timing.

kentonv:
Oh wait, these are all WebSocket methods, so http-over-capnp-perf-test won't cover them.

But also I don't care so much about perf of WebSocket messages anyawy, so maybe this is fine.

kentonv (approved):
As mentioned I think this will increase memory allocations, but I'm not too concerned about it in the WebSocket path.

PR #2159, opened 2024-10-11
convert pumpWebSocketLoop into coros

mikea:
Opens doors for improved error handling according to the comment.

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -3759,34 +3759,33 @@ kj::Own<WebSocket> newWebSocket(kj::Own<kj::AsyncIoStream> stream,
 }
 
 static kj::Promise<void> pumpWebSocketLoop(WebSocket& from, WebSocket& to) {
-  return from.receive().then([&from,&to](WebSocket::Message&& message) {
-    KJ_SWITCH_ONEOF(message) {
-      KJ_CASE_ONEOF(text, kj::String) {
-        return to.send(text)
-            .attach(kj::mv(text))
-            .then([&from,&to]() { return pumpWebSocketLoop(from, to); });
+  try {
+    while (true) {
+      auto message = co_await from.receive();
+      KJ_SWITCH_ONEOF(message) {
+        KJ_CASE_ONEOF(text, kj::String) {
+          co_await to.send(text);
+        }
+        KJ_CASE_ONEOF(data, kj::Array<byte>) {
+          co_await to.send(data);
+        }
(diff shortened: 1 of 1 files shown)


c++/src/kj/compat/http.c++
```diff
@@ -3759,34 +3759,33 @@ kj::Own<WebSocket> newWebSocket(kj::Own<kj::AsyncIoStream> stream,
...
       }
-    }
-    KJ_UNREACHABLE;
-  }, [&to](kj::Exception&& e) -> kj::Promise<void> {
-    // We don't know if it was a read or a write that threw. If it was a read that threw, we need
-    // to send a disconnect on the destination. If it was the destination that threw, it
-    // shouldn't hurt to disconnect() it again, but we'll catch and squelch any exceptions.
-    kj::runCatchingExceptions([&to]() { to.disconnect(); });
+      // continue the loop
+    } catch (kj::Exception e) {
```
danlapid, line 3779: I believe this should be 
```
catch (...) {
    auto e = kj::getCaughtExceptionAsKj();
    ...
```
But if you're only getting the variable to rethrow you can just
```
catch(...) {
    ...
    throw;
}
```

jclee: I guess another possibility would be to put the disconnect code in a `KJ_ON_SCOPE_FAILURE()` block at the top of the scope.

https://github.com/capnproto/capnproto/blob/9ba75a7c/c%2B%2B/src/kj/exception.h#L386-L394

danlapid: Honestly I prefer idiomatic c++ try catch clauses.

c++/src/kj/compat/http.c++
```diff
@@ -3759,34 +3759,33 @@ kj::Own<WebSocket> newWebSocket(kj::Own<kj::AsyncIoStream> stream,
...
-        return to.send(data)
-            .attach(kj::mv(data))
-            .then([&from,&to]() { return pumpWebSocketLoop(from, to); });
-      }
-      KJ_CASE_ONEOF(close, WebSocket::Close) {
-        // Once a close has passed through, the pump is complete.
-        return to.close(close.code, close.reason)
-            .attach(kj::mv(close));
+  while (true) {
+    try {
```
kentonv, line 3763: The `try` could be outside the `while` loop. Not sure if it's any more efficient that way... but it's certainly not less efficient.

danlapid (approved):

PR #2161, opened 2024-10-15
Fix client->server WebSocket pong encoding.

kentonv:
We weren't applying the mask correctly. (Server->client doesn't have a mask, so wasn't impacted.)

cc @threepointone

### DIFF
--- a/c++/src/kj/compat/http-test.c++
+++ b/c++/src/kj/compat/http-test.c++
@@ -2077,12 +2077,12 @@ KJ_TEST("WebSocket unsolicited pong") {
   clientTask.wait(waitScope);
 }
 
-KJ_TEST("WebSocket ping") {
+void doWebSocketPingTest(kj::Maybe<EntropySource&> maskGenerator) {
   KJ_HTTP_TEST_SETUP_IO;
   auto pipe = KJ_HTTP_TEST_CREATE_2PIPE;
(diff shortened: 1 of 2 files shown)


c++/src/kj/compat/http-test.c++
```diff
@@ -5166,16 +5186,16 @@ KJ_TEST("HttpClient WebSocket: client can have a custom WebSocket error handler"
...
-    'R', 'e', 'c', 'e', 'i', 'v', 'e', 'd', ' ',
-    'f', 'r', 'a', 'm', 'e', ' ',
-    'h', 'a', 'd', ' ',
-    'R', 'S', 'V', ' ',
-    'b', 'i', 't', 's', ' ',
-    '2', ' ',
-    'o', 'r', ' ',
-    '3', ' ',
-    's', 'e', 't',
+    0x88, 0xa8, 12, 34, 56, 78, 0x3^12, 0xea^34, // FIN, opcode=Close, code=1009
```
kentonv, line 5189: Seems this test didn't realize that `0xC, 0x22, 0x38, 0x4e` was actually the fake mask value of `12, 34, 56, 78`. If we noticed we'd probably also have noticed that the payload wasn't masked...

MellowYarker (approved):

PR #2163, opened 2024-10-16
[nfc] couple exclusiveJoin() tests

mikea:
Wrote them to double check contract for myself.


kentonv (commented):
Maybe comment on what these are testing?

The first test looks to me like it's relying on unspecified execution order. Since both sides of the join resolve immediately, the join could legitimately resolve to either one.

mikea:
> Maybe comment on what these are testing?
> 
> The first test looks to me like it's relying on unspecified execution order. Since both sides of the join resolve immediately, the join could legitimately resolve to either one.

So I got the contract wrong in the end: I was hoping that it would resolve into first successfull promise.

kentonv:
It resolves to the first promise to settle (either resolve or reject), and then cancels the other one.

PR #2164, opened 2024-10-16
Promise<T> kj::raceSuccessful(kj::Array<T> promises)

mikea:
Executes multiple promises and returns the first successuful result, canceling the rest.
In case of every promise failing fails with the last (in time) error.

### DIFF
--- a/c++/src/kj/async-inl.h
+++ b/c++/src/kj/async-inl.h
@@ -1141,6 +1141,81 @@ private:
   Array<ExceptionOr<_::Void>> resultParts;
 };
 
+class RaceSuccessfulPromiseNodeBase : public PromiseNode {
+public:
+  RaceSuccessfulPromiseNodeBase(Array<OwnPromiseNode> promises,
+                                ExceptionOrValue &output,
+                                SourceLocation location);
+  ~RaceSuccessfulPromiseNodeBase() noexcept(false);
+
+  void onReady(Event *event) noexcept override final;
+  void get(ExceptionOrValue &output) noexcept override final;
+  void tracePromise(TraceBuilder &builder, bool stopAtNextEvent) override final;
+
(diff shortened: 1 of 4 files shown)


mikea:
I don't think I missed the already existing method, have I?

kentonv:
I guess this is the same as `exclusiveJoin()` except it takes an arbitrary-sized array?

Technically you could implement it by chaining a bunch of `exclusiveJoin()`s together?

kentonv:
Oh I see, it does not immediately propagate rejected promises, unless they _all_ reject?

That seems unusual requirement to me. Every use case I've ever encountered, I wanted to propagate the first failure. What is the use case for this exactly?

I think this is not the obvious expected behavior for a function called `racePromises`, so I'd probably call it something like `racePromisesIgnoreFailures` or `racePromisesFailSlow` (it's the opposite of "fail-fast").

Also while I agree the word "race" is better than "exclusiveJoin", it seems confusing to introduce a new term to mean the same thing. Maybe we should actually rename `exclusiveJoin` to `raceWith`?

mikea:
> That seems unusual requirement to me. Every use case I've ever encountered, I wanted to propagate the first failure. What is the use case for this exactly?

Reading from multiple storage buckets and taking the fastest. One obviously doesn't want to fail if first request to resolve was unsuccessful. 

Happy to change the name.

mikea:
> Maybe we should actually rename exclusiveJoin to raceWith?

I am all for this disregarding our decision in this PR: current name is too confusing. 

I'll create a PR.

kentonv (approved):
Vaguely feels like if all the promises fail, it should be the first error that gets returned -- but I can see how that is more annoying to implement, and it probably doesn't really matter in practice, so this seems fine.

PR #2168, opened 2024-10-23
Throw an exception on corrupt gzip data

erikcorry:


c++/src/kj/compat/gzip-test.c++
```diff
@@ -193,6 +193,19 @@ KJ_TEST("async gzip decompression") {
...
+  {
+    Vector<byte> bytes;
+    bytes.addAll(ArrayPtr<const byte>(FOOBAR_GZIP));
+    bytes[20] ^= 0x55;  // Corrupt gzip stream.
+    MockAsyncInputStream rawInput(bytes, kj::maxValue);
+    GzipAsyncInputStream gzip(rawInput);
+
+    char text[16]{};
+    KJ_EXPECT_THROW_MESSAGE("gzip decompression failed; ctx.msg = incorrect data check",
+        gzip.tryRead(text, 1, sizeof(text)).wait(io.waitScope));
```
harrishancock, line 206: Does this test fail without the patch? I would expect the call to `tryRead()` to throw before the patch, and the `.wait()` to throw after the patch, but I'd expect the test to pass either way.

c++/src/kj/compat/gzip.c++
```diff
@@ -232,9 +232,9 @@ Promise<size_t> GzipAsyncInputStream::readImpl(
     }
   } else {
     if (ctx.msg == nullptr) {
-      KJ_FAIL_REQUIRE("gzip decompression failed", inflateResult);
+      return KJ_EXCEPTION(FAILED, "gzip decompression failed", inflateResult);
```
kentonv, line 235: The usual convention in KJ is that an async function is free to throw exception either synchronously or asynchronously. If the caller wants to force all exceptions to be async, it is their responsibility to wrap in `kj::evalNow()` or something like that.

So the change here doesn't really make a difference unless the caller is buggy.

What motivated this?

PR #2169, opened 2024-10-23
breaking up http-over-capnp bazel dep

mikea:


anonrig (approved):

fhanau:
Glad we're doing this – I was about to complain that `bazel cquery "deps(//src/...)"` is broken due to the missing Rust target, but this fixes it.

mikea:
> Glad we're doing this – I was about to complain that `bazel cquery "deps(//src/...)"` is broken due to the missing Rust target, but this fixes it.

this was exactly the reason I have fixed this

PR #2184, opened 2024-11-20
Fix warning: cmake_minimum_required() should be called prior to this top-level project()

balusch:
As per the CMake document, we should call the cmake_minimum_required() command at the beginning of the top-level CMakeLists.txt file even before calling the project() command.

### DIFF
--- a/c++/samples/CMakeLists.txt
+++ b/c++/samples/CMakeLists.txt
@@ -16,8 +16,8 @@
 #      cmake ../c++/samples
 #      cmake --build .
 
-project("Cap'n Proto Samples" CXX)
 cmake_minimum_required(VERSION 3.1)
+project("Cap'n Proto Samples" CXX)
 
 find_package(CapnProto CONFIG REQUIRED)
 


kentonv:
I don't know cmake but maybe @harrishancock can review.

You've created this PR against the `v2` branch, but is this something you intend to make its way into a 1.x release? If so, please rebase it over to the `master` branch and we can merge it there.

balusch:
Hi @kentonv. The change should be applied to both the master and v2 branches. In this regard, should I also rebase it and adjust the target branch to master? I'm not very familiar with multi-release development. Once it is merged into master, will you cherry-pick it to v2?

kentonv:
I would recommend targeting master. Everything that lands on master, I will probably cherry-pick into v2 at some point.

balusch:
Rebased and re-targeted. Note that the two branches use different versions of CMake, so the commit may not be able to be cherry-picked directly.

harrishancock (approved):

PR #2188, opened 2024-11-30
[build][perf][capnpc] Move function definition in generated capnp files out of headers

fhanau:
**Motivation**
For downstream projects, parsing capnproto headers is a major factor in slow compile times. This is mediated by capnproto definitions being compiled to classes with many sub-classes and functions, including templated ones; defining sub-classes which can't be forward declared; and relying on recursively including generated headers. Based on compiler profiling, parsing generated capnp headers is disproportionately more expensive than parsing other headers; I assume having many small functions plays a role here. On many downstream source files that include capnp headers, the compiler spends more time parsing capnp headers than doing anything else, including codegen.

**Proposed change**
capnpc-c++ translates capnp files to a C++ source and header file. Most class methods are first declared as inline methods in the header and then defined later in the header. Defining methods in the header allows the compiler to inline them, but contributes to compile time overhead. With this change, non-templated functions will be declared as non-inline and moved to the generated source file instead of the header. This results in massive improvements in include size and compile times.
I have regenerated the bootstrap capnp files here, which serve as an example of how the changes look like.

**Results**
To get a somewhat representative picture of the build performance impact I tested this on both small tests and larger source files. On a small downstream test file including large capnp headers as part of the test framework, this change reduces the include size (i.e. size of the preprocessed source file with all includes resolved) by **24%** (4.95MB => 3.66MB) and compile time by **46% (2.85s => 1.54s)**. There are many tests where this applies. On a large-ish downstream file including even more capnp code, include size shrinks **27%** (10.23MB => 7.46MB) and compile time **43% (6.58s => 3.77s)**. 
On workerd api/crypto/impl.c++, smaller improvements of 8% for include size (6.72MB => 6.16MB) and 22% for compile time (3.29s => 2.58s) can be observed. Similarly io/io-context.c++ sees improvements of 9% (6.07MB => 5.52MB) and 23% (3.87s => 2.98s), respectively. The smaller improvement for workerd reflects that it relies less on multiple inheritance of capnproto structures, and has some more header parsing overhead from JSG/V8.
These numbers are for -O0 -g0, the impact is inevitably smaller when optimization is enabled. I'm happy to elaborate on the build performance impact out-of-band.

**Limitations**
- There is a **binary size regression** since the compiler will need to generate code for every capnp function moved to the source file instead of only those that are included elsewhere. On an internal binary used to run tests, the debug binary size increases by 7%. However, this effect all but vanishes when enabling optimization, using ffunction-sections/gc-sections, and/or reducing the level of debug information, suggesting that this is caused by code that is not actually being used and the debug information being generated for it. Release binaries for the full runtime see a much smaller impact (~0.2%).
- It is conceivable that this makes the performance of binaries using capnp worse as the compiler will have fewer opportunities for inlining. I'd imagine that this overhead will generally be low as "hot" code is unlikely to repeatedly call generated code for capnp structures. Additionally, the compiler should still be able to perform inlining when thinLTO is enabled.
- This PR is not aiming to be exhaustive in moving code outside of generated headers – to keep the size of this change reasonable, I'm not attempting to move functions generated in makeInterfaceText and makeMethodText. Based on limited testing this change set already achieves most of the potential speedup here.
- As with any larger code generator change getting things wrong here can easily cause compile / linker errors or simply subtle changes to formatting, so this will require a close review. I have created a downstream branch under the same name to try this, all builds are passing there so this looks relatively stable.
- If this change is considered to have negative side effects, making it configurable as an opt-in option would still speed up local development and boost developer productivity.

### DIFF
--- a/c++/src/capnp/BUILD.bazel
+++ b/c++/src/capnp/BUILD.bazel
@@ -141,6 +141,10 @@ cc_binary(
     srcs = [
         "compiler/capnpc-c++.c++",
     ],
+    defines = select({
+        "//src/capnp:capnp_no_inline_accessors_true": ["CAPNP_NO_INLINE_ACCESSORS"],
+        "//conditions:default": [],
+    }),
     visibility = ["//visibility:public"],
     deps = [
         ":capnpc",
--- a/c++/src/capnp/compiler/capnpc-c++.c++
+++ b/c++/src/capnp/compiler/capnpc-c++.c++
@@ -1133,11 +1133,13 @@ private:
   DiscriminantChecks makeDiscriminantChecks(kj::StringPtr scope,
                                             kj::StringPtr memberName,
                                             StructSchema containingStruct,
-                                            const TemplateContext& templateContext) {
+                                            const TemplateContext& templateContext,
+                                            bool inplace = false) {
     auto discrimOffset = containingStruct.getProto().getStruct().getDiscriminantOffset();
 
     kj::String titleCase = toTitleCase(memberName);
     kj::String upperCase = toUpperCase(memberName);
+    kj::StringPtr maybeInline = inplace ? ""_kjc : "inline "_kjc;
 
     return DiscriminantChecks {
         kj::str(
@@ -1152,15 +1154,15 @@ private:
             "  _builder.setDataField<", scope, "Which>(\n"
             "      ::capnp::bounded<", discrimOffset, ">() * ::capnp::ELEMENTS, ",
                       scope, upperCase, ");\n"),
-        kj::strTree("  inline bool is", titleCase, "() const;\n"),
-        kj::strTree("  inline bool is", titleCase, "();\n"),
+        kj::strTree("  ", maybeInline, "bool is", titleCase, "() const;\n"),
+        kj::strTree("  ", maybeInline, "bool is", titleCase, "();\n"),
         kj::strTree(
             templateContext.allDecls(),
-            "inline bool ", scope, "Reader::is", titleCase, "() const {\n"
+            maybeInline, "bool ", scope, "Reader::is", titleCase, "() const {\n"
             "  return which() == ", scope, upperCase, ";\n"
             "}\n",
             templateContext.allDecls(),
-            "inline bool ", scope, "Builder::is", titleCase, "() {\n"
+            maybeInline, "bool ", scope, "Builder::is", titleCase, "() {\n"
             "  return which() == ", scope, upperCase, ";\n"
             "}\n")
     };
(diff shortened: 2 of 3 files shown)


kentonv:
Cap'n Proto's primitive field accessors are intended to be "as fast as accessing struct fields" -- just a load-at-offset. With this change, though, the accessors can no longer be inlined, which will make them significantly slower than a simple load-at-offset and will also disrupt other compiler optimizations (generally, the optimizer has to assume that any out-of-line call could change the whole state of the world before returning).

> I'd imagine that this overhead will generally be low as "hot" code is unlikely to repeatedly call generated code for capnp structures.

I don't think this is a valid assumption. Message parsing and serialization often occupies a large portion of CPU time in real-world distributed services.

I agree that the compile time improvements are surprisingly large and maybe worth it for debug builds. Perhaps we simply need a "debug" flag / environment variable which causes capnp to generate out-of-line declarations for debug builds specifically. The default should be inline, though, so that people with unsophisticated builds default to fast runtime rather than fast build times.

fhanau:
I don't necessarily agree, but concede that I can't rule out large performance regressions for some use cases without extensive benchmarking.
For conditionally enabling this, would a compile-time switch be acceptable? This would be the easiest to integrate it with a bazel config flag and would not require adding this to the CLI and propagating it across bazel,capnp_tool and capnpc-c++ itself. For Bazel, capnpc-c++ gets built automatically and .capnp.{c++,h} build files get generated automatically, so needing to recompile capnpc-c++ when e.g. switching from debug to opt builds is not a big concern.

kentonv:
Sure, a compile-time flag sounds fine.

fhanau:
Updated to only enable this when CAPNPC_GENERATE_STUBBED_INCLUDES is defined (open to other define names too). Tested this by diffing the bootstrap capnp files – when this is defined the same files as with 38673e2cb1197c012701b923ef9864e42e9063ac are generated, otherwise we get the same files we had before.
In some cases the order of templated types in the generated header may change based on how shouldTemplatizeInit is handled, seems like this does not happen for any of the bootstrap files though. If this is objectionable I can try to change that, it that would result in a larger diff.
For upstream, this can easily be toggled using `--@capnp-cpp//src/capnp:capnpc_generate_stubbed_includes=True`.

fhanau:
I'd love to see this change succeed – what can I do to make this easier to review? The change as-is was mostly written to limit the diff size, but if this will be an opt-in other approaches might make more sense. Instead of the maybeInline trick, I can leave the existing definition blocks as-is and just duplicate them behind the compile time flag with the inline keyword removed, that should make it easier to confirm that behavior is identical if the define is not enabled.

So far I've been using this on a local branch and it has improved my productivity working on upstream projects. I'm sure other team members will also benefit from having it as an option to speed up debug builds in local development.

kentonv:
Sorry, this fell off my radar.

This seems fine but the name "stubbed includes" isn't intuitive to me. Maybe `CAPNP_NO_INLINE_ACCESSORS`?

fhanau:
Updated to use `CAPNP_NO_INLINE_ACCESSORS`.

kentonv (approved):

PR #2190, opened 2024-12-03
Fix useSystemTrustStore behaviour in Windows

danlapid:
Under windows SSL_CTX_set_default_verify_paths doesn't populate the trusted ca store with all the trusted CAs. This commit should fix the error.

### DIFF
--- a/c++/src/kj/compat/tls.c++
+++ b/c++/src/kj/compat/tls.c++
@@ -38,6 +38,16 @@
 #include <kj/debug.h>
 #include <kj/vector.h>
 
+#ifdef _WIN32
+#include <kj/win32-api-version.h>
+#include <windows.h>
+#include <cryptuiapi.h>
+#include <wincrypt.h>
+#undef CONST
+#undef X509_NAME
+#include <kj/windows-sanity.h>
+#endif
+
 #if OPENSSL_VERSION_NUMBER < 0x10100000L
 #define BIO_set_init(x,v)          (x->init=v)
 #define BIO_get_data(x)            (x->ptr)
@@ -79,6 +89,31 @@ void throwOpensslError() {
 
   kj::throwFatalException(getOpensslError());
 }
+
+#ifdef _WIN32
+// Adds windows CA store certificates to OpenSSL's ca store
+// Taken from https://stackoverflow.com/questions/9507184/can-openssl-on-windows-use-the-system-certificate-store
+void updateOpenSSLCAStoreWithWindowsCertificates(SSL_CTX* ctx) {
+  X509_STORE* store = SSL_CTX_get_cert_store(ctx);
+  if (store == nullptr) {
+    throwOpensslError();
+  }
+  HCERTSTORE hStore;
+  KJ_WIN32(hStore = CertOpenSystemStoreA(NULL, "ROOT"));
+  KJ_DEFER(KJ_WIN32(CertCloseStore(hStore, 0)));
+  PCCERT_CONTEXT pContext = nullptr;
+  KJ_DEFER(CertFreeCertificateContext(pContext));
+  while ((pContext = CertEnumCertificatesInStore(hStore, pContext))) {
+    X509* x509 = d2i_X509(nullptr, (const unsigned char**)&pContext->pbCertEncoded, pContext->cbCertEncoded);
+    if (x509) {
+      KJ_DEFER(X509_free(x509));
+      if (!X509_STORE_add_cert(store, x509)) {
+        throwOpensslError();
+      }
+    }
+  }
+}
+#endif
 
 #if OPENSSL_VERSION_NUMBER < 0x10100000L && !defined(OPENSSL_IS_BORINGSSL)
 // Older versions of OpenSSL don't define _up_ref() functions.
@@ -744,6 +779,9 @@ TlsContext::TlsContext(Options options) {
     if (!SSL_CTX_set_default_verify_paths(ctx)) {
       throwOpensslError();
     }
+#ifdef _WIN32
+    updateOpenSSLCAStoreWithWindowsCertificates(ctx);
+#endif
   }
 
   // honor options.trustedCertificates
--- a/c++/src/kj/CMakeLists.txt
+++ b/c++/src/kj/CMakeLists.txt
@@ -204,6 +204,9 @@ if(WITH_OPENSSL)
 
     target_compile_definitions(kj-tls PRIVATE KJ_HAS_OPENSSL)
     target_link_libraries(kj-tls PRIVATE OpenSSL::SSL OpenSSL::Crypto)
+    if(WIN32)
+      target_link_libraries(kj-tls PUBLIC crypt32)
+    endif()
 
     # Ensure the library has a version set to match autotools build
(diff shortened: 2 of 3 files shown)


c++/src/kj/compat/tls-win32.h
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2016 Sandstorm Development Group, Inc. and contributors
```
kentonv, line 1:
```suggestion
// Copyright (c) 2024 Cloudflare, Inc. and contributors
```

c++/src/kj/compat/tls-win32.c++
```diff
@@ -0,0 +1,67 @@
...
+void throwOpensslError() {
+  // Call when an OpenSSL function returns an error code to convert that into an exception and
+  // throw it.
+
+  kj::throwFatalException(getOpensslError());
+}
+
+// Adds windows CA store certificates to OpenSSL's ca store
+// Taken from https://stackoverflow.com/questions/9507184/can-openssl-on-windows-use-the-system-certificate-store
+void updateOpenSSLCAStoreWithWindowsCertificates(void* ctx) {
```
kentonv, line 45: I guess this function is the only new code in this file -- the other functions above are copy/pasted from `tls.c++`.

Can we just put this in `tls.c++`? Having a whole separate file for this seems unnecessary.

danlapid: I will deduplicate soon, it will be cleaner.
I tried to put it in tls.c++ and been fighting with the compiler for a day but there seems to be something from the windows headers that's ruining compilation and I can't figure out why
This is the error I'm getting
image

kentonv: It looks like `wincrypt.h` defines `X509_NAME` as a macro constant. You can probably just `#undef X509_NAME` after including `wincrypt.h`.

https://stackoverflow.com/questions/49504648/x509-name-macro-in-c-wont-compile

danlapid: Thanks! that's really helpful sad I couldn't figure that out

danlapid:
@kentonv Ok this is now ready for review.
I think the code looks sane and I've checked and it solves the errors we had in workerd so it is working as expected. (https://github.com/cloudflare/workerd/actions/runs/12164190382/job/33925279974?pr=3203).
It would also allow us to delete the following hack from miniflare:
https://github.com/cloudflare/workers-sdk/blob/main/packages/miniflare/src/plugins/core/index.ts#L76-L92

c++/src/kj/compat/tls.c++
```diff
@@ -80,6 +90,33 @@ void throwOpensslError() {
...
+  }
+  HCERTSTORE hStore = CertOpenSystemStoreW(NULL, L"ROOT");
+  KJ_ASSERT(hStore != nullptr, "Failed to open system certificate store");
+  
+  PCCERT_CONTEXT pContext = nullptr;
+  while ((pContext = CertEnumCertificatesInStore(hStore, pContext))) {
+    X509* x509 = d2i_X509(nullptr, (const unsigned char**)&pContext->pbCertEncoded, pContext->cbCertEncoded);
+    if (x509) {
+      if (!X509_STORE_add_cert(store, x509)) {
+        throwOpensslError();
```
kentonv, line 109: Looks like if this throws we'll leak some memory as various cleanup below will be skipped. Maybe use KJ_DEFER() to register all the cleanup steps so they happen either way?

danlapid: Done

c++/src/kj/compat/tls.c++
```diff
@@ -80,6 +90,31 @@ void throwOpensslError() {
...
+void updateOpenSSLCAStoreWithWindowsCertificates(SSL_CTX* ctx) {
+  X509_STORE* store = SSL_CTX_get_cert_store(ctx);
+  if (store == nullptr) {
+    throwOpensslError();
+  }
+  HCERTSTORE hStore = CertOpenSystemStoreW(NULL, L"ROOT");
+  KJ_ASSERT(hStore != nullptr, "Failed to open system certificate store");
+  KJ_DEFER(CertCloseStore(hStore, 0));
+  PCCERT_CONTEXT pContext = nullptr;
+  KJ_DEFER(CertFreeCertificateContext(pContext));
```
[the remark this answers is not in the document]
danlapid: If you're willing to believe this function hasn't changed much from Windows XP then yes.
Defined in newstor.cpp (https://github.com/tongzx/nt5src/blob/master/Source/XPSP1/NT/ds/security/cryptoapi/pki/certstor/newstor.cpp#L4770)
It calls `ReleaseContextElement`
Which at first does:
```
    if (pEle == NULL)
        return;
```

c++/src/kj/compat/tls.c++
```diff
@@ -80,6 +90,31 @@ void throwOpensslError() {
...
+// Adds windows CA store certificates to OpenSSL's ca store
+// Taken from https://stackoverflow.com/questions/9507184/can-openssl-on-windows-use-the-system-certificate-store
+void updateOpenSSLCAStoreWithWindowsCertificates(SSL_CTX* ctx) {
+  X509_STORE* store = SSL_CTX_get_cert_store(ctx);
+  if (store == nullptr) {
+    throwOpensslError();
+  }
+  HCERTSTORE hStore = CertOpenSystemStoreW(NULL, L"ROOT");
+  KJ_ASSERT(hStore != nullptr, "Failed to open system certificate store");
+  KJ_DEFER(CertCloseStore(hStore, 0));
```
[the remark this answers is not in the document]
danlapid: > Oh, also `CertOpenSystemStoreW` should probably check errors:
> 
> ```
>   HCERTSTORE hStore;
>   KJ_WIN32(hStore = CertOpenSystemStoreW(NULL, L"ROOT"));
> ```

I checked errors with `KJ_ASSERT(hStore != nullptr, "Failed to open system certificate store");` but the `KJ_WIN32` approach is indeed much nicer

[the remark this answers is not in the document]
danlapid: > `CertCloseStore` can fail, so we should wrap it in `KJ_WIN32`, which is the Windows equivalent of `KJ_SYSCALL`:
> 
> Meanwhile, according to docs, `CertFreeCertificateContext` always returns true... so I guess it doesn't really need it?
> 
> `CertEnumCertificatesInStore`, for its part, is a bit ambiguous... we expect it to return NULL at the end of the list, but it might also return NULL to indicate an error. Annoyingly, there are in fact two different error codes that might indicate end? Ugh I guess we can just skip checking the code and always assume null means end.

Yeah honestly I considered making this entire function just return early on failure without exceptions altogether just because if windows for some odd reason fails any of these calls we should probably just ignore and move on.
Eventually decided to do a half-assed attempt at error checking rather than none but I appreciate your comments trying to actually make this correct.

c++/src/kj/compat/tls.c++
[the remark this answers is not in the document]
danlapid: Btw CertFreeCertificateContext really does always return TRUE 😂

Issue #2187, opened 2024-11-29
capnp::writeMessageToFd fails for large messages on macOS


JanosGit:
In our application, messages that should be written to files can become quite large, depending on user input. The Messages are created using `capnp::MallocMessageBuilder`, passing an externally allocated (potentially very big) first segment to it.

Under some circumstances I found `capnp::writeMessageToFd` throwing an exception which is caused by the call to `::writev` here: https://github.com/capnproto/capnproto/blob/d4ba9809ddd89b85d401353b9dfb5b59cf705db7/c%2B%2B/src/kj/io.c%2B%2B#L405

I looked at the issue under the debugger and found out that the `iov` array passed in that scenario contains three elements of the sizes 16, 2132300288 and 76629368 bytes. `::writev` then returns -1 and `errno` is set to `EINVAL`. I read the online man pages for `writev` multiple times finding no clues why this should fail until I eventually ended up looking at the man pages on my actual system in the terminal. That felt so old school but revealed that it indeed contains quite a bit of extra information neither found in the archived macOS man pages or linux man pages that can be found online. To my surprise I did not find any only version of the man pages for the macOS 14.1 SDK, so I'll paste the content of it below.

The relevant parts are:
> write() and pwrite() will fail if the parameter nbyte exceeds INT_MAX, and they do not attempt a partial write.

and further below under ERRORS specific for `writev`

> [EINVAL]		The sum of the iov_len values in the iov array overflows a 32-bit integer.

Looking at the values of my test case again this indeed is the case here. So I think there either needs to be some logic that splits the data into multiple write calls. Do you agree?

```
WRITE(2)		      System Calls Manual		      WRITE(2)

NAME
     pwrite, write, pwritev, writev – write output

LIBRARY
     Standard C Library (libc, -lc)

SYNOPSIS
     #include <unistd.h>

     ssize_t
     pwrite(int fildes, const void *buf, size_t nbyte, off_t offset);

     ssize_t
     write(int fildes, const void *buf, size_t nbyte);

     #include <sys/uio.h>

     ssize_t
     writev(int fildes, const struct iovec *iov, int iovcnt);

     ssize_t
     pwritev(int fildes, const struct iovec *iov, int iovcnt, off_t offset);

DESCRIPTION
     write() attempts to write nbyte of data to the object referenced by the
     descriptor fildes from the buffer pointed to by buf.  writev() performs
     the same action, but gathers the output data from the iovcnt buffers
[...]

PR #2191, opened 2024-12-03
kj::FdOutputStream::write: Split writes for macOS in case byte count exceeds INT_MAX

JanosGit:
Fix for the issue discussed in https://github.com/capnproto/capnproto/issues/2187

I decided to allocate a temporary Vector in case the size exceeded the limit as I don't consider it a performance issue in case that much data is written out. I tried to match the style of the existing implementation but feel free to remark if you feel that the code doesn't match your preferred style.

### DIFF
--- a/c++/src/kj/io.c++
+++ b/c++/src/kj/io.c++
@@ -31,6 +31,7 @@
 #include "debug.h"
 #include "miniposix.h"
 #include <algorithm>
+#include <numeric>
 #include <errno.h>
 #include "vector.h"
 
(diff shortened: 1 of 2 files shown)


kentonv:
Hi @JanosGit,

Thanks for submitting a PR. But:
* It looks like you didn't make any attempt to match the coding style of the project.
* I think we can solve this without memory allocations.

Rather than try to go back-and-forth a lot on how to fix this code, I just went ahead and wrote #2198. However, I don't have a Mac to test it on. Can you try it and make sure it works?

PR #2105, opened 2024-07-13
[RFC] Backport v1.x fixes to v2

fhanau:
I was curious if the v2 branch was missing any important changes that were contributed to the master branch and reviewed all code changes present on master but not on v2 – turns out that this largely consists of smaller changes that are not very relevant to downstream projects, so there's **no need to prioritize** porting them. I still wanted to track them here so we don't lose them. I'm not sure if cherry-picking the changes individually is even the right approach for this, but this should include all relevant code changes as of writing. 2a9ba4eca3253a3e53ac91b6aefaa7e0f9475474 which fixes compilation errors for recent compilers is not included as we already resolved that differently.

Cherry-picked commits:
9da0b118049cbf017935ac20a04a9aa5bbc54514
2d41340e102607d4478b7b79d0dcfe5fdd7f67be
2ed5cdcb6c67fd885401b204fb7b6cace8aadf74
12b49cca010fde10d1ee8109c813a9eabe1cb43e
2754d74e3eea68413936c4ab546d9a0923c6bb35
58e509db5473fab6acba3c74a2be648f05ce934e
0eea6f1a5095c16225a6f22e3cd534d017a52957
5303d8776ab8b1dd883957dffe3a304dc44d6e01
7ad12031f05ff7d36bcdd41b8baeec0b33bfcf4a
3dc31dbca86e561249e7c2a0ca366514523d9a91

PR #2203, opened 2024-12-21
Cherry-pick changes from master branch as of 1.1.0 release into v2

kentonv:
Similar to #2105, this pulls in various bugfixes (and a feature) that landed directly on the master branch, as of the 1.1.0 release (which I just pushed today).

I produced this using a plain git rebase from the master branch onto v2, then carefully removing commits that either weren't appropriate to carry over or which were already present on v2. (In some cases git automatically recognizes already-applied cherry-picks and drops them from the rebase list, but in some cases it doesn't and I had to find them manually.)

So, next time we do this, we can start from the point of 1.1.0 release.


fhanau (approved):
LGTM, that should cover all commits that were not on v2 yet as far as I can see.

PR #2205, opened 2024-12-24
Improve FD handling: Add KJ_SYSCALL_FD and rename AutoCloseFd

kentonv:
This pattern is very common in KJ code:

```
int fd_;
KJ_SYSCALL(fd_ = open(...));
kj::AutoCloseFd fd(fd_);
```

Now it can be written as:

```
auto fd = KJ_SYSCALL_FD(open(...));
```

I've wanted this for like ten years but never got around to writing it.

Additionally, I've renamed `AutoCloseFd` to `OwnFd`, with a backwards-compatibility alias. This is another thing I've wanted to do for ten years; the old name was terrible. I think I didn't realize how widely it would be used at the time I first named the class.

I've verified this PR doesn't break the Workers Runtime builds, although I also plan to update them.


fhanau (approved):
LGTM

PR #2207, opened 2024-12-28
[bazel] Add support for header parsing

fhanau:
This allows us to use the bazel header parsing check. capnproto appears to already be compliant, just need to disable it for async-inl.h.

### DIFF
--- a/c++/src/kj/BUILD.bazel
+++ b/c++/src/kj/BUILD.bazel
@@ -116,6 +116,8 @@ cc_library(
         "//conditions:default": ["async-unix.h"],
     }),
     include_prefix = "kj",
+    # Disable header parsing based on async-inl.h
+    features = ["-parse_headers"],
     linkopts = select({
         "@platforms//os:windows": [
             "Ws2_32.lib",


c++/src/kj/BUILD.bazel
```diff
@@ -116,6 +116,8 @@ cc_library(
         "//conditions:default": ["async-unix.h"],
     }),
     include_prefix = "kj",
+    # Disable header parsing based on async-inl.h
+    features = ["-parse_headers"],
```
fhanau, line 120: Disabling the feature here means that Bazel will not parse the headers of kj-async. This is the cleanest way to get header parsing to work given the `#error "Do not include this directly; include kj/async.h."` directive in `async-inl.h`. Alternatively we could add a define to allow disabling that check while still including async.h but this is easier.

kentonv (approved):
Cool but... what _is_ header parsing?

kentonv:
Ah, I see: https://bazel.build/docs/bazel-and-cpp

Just parsing the headers to make sure they are self-contained...

PR #2212, opened 2025-01-14
filesystem-disk-unix.c++: add a missing header for Apple

barracuda156:
This fixes a build error due to undeclared `renameat`.

### DIFF
--- a/c++/src/kj/filesystem-disk-unix.c++
+++ b/c++/src/kj/filesystem-disk-unix.c++
@@ -53,6 +53,10 @@
 #include <sys/sendfile.h>
 #endif
 
+#if __APPLE__
+#include <sys/stdio.h> // renameat
+#endif
+
 namespace kj {
 namespace {
 


kentonv:
Please target the master branch, not a release branch.

I'm a bit confused though: We run CI tests on mac and there has been no issue with this, and the man page documents `stdio.h` as the correct header (which we already include). Where did `sys/stdio.h` come from and why is it required on your system but not in our CI?

barracuda156:
@kentonv Thank you for responding!
It might have moved across the headers, but it was introduced in 10.10 in `sys/stdio.h`: https://github.com/alexey-lysiuk/macos-sdk/blob/69ddec41e6224afebd5f19bb24764305ac3e828a/MacOSX10.10.sdk/usr/include/sys/stdio.h#L39

barracuda156:
I have rebased now to master, took a few attempts )

kentonv:
But `sys/stdio.h` is included from `stdio.h`, even in the version you reference:

https://github.com/alexey-lysiuk/macos-sdk/blob/69ddec41e6224afebd5f19bb24764305ac3e828a/MacOSX10.10.sdk/usr/include/stdio.h#L75

So an explicit include shouldn't be necessary.

barracuda156:
@kentonv Good point. Allow me until tomorrow, I just left home from the machine. I will check why that fails (it is not just with `capnproto`, I have encountered this error elsewhere, and including `sys/stdio.h` fixed it). As I blind guess, `stdio.h` of < 10.10 would not include the second header, obviously, since it does not exist (at least in a few SDKs I checked), so if `legacy-support` does not make up for that, indirect inclusion gonna fail.

barracuda156:
@kentonv I tend to think this is a bug in `legacy-support`. I have opened a ticket for it: https://trac.macports.org/ticket/71842

I am not immediately closing this PR, just making it a draft, just in case someone on MacPorts side provides a reason why `legacy-support` is correct and `sys/stdio.h` should be included directly, but most likely it should be fixed there.

kentonv:
Cool. If this PR ends up being needed then I'd just ask that it include a comment explaining exactly under what circumstances it is needed, otherwise someone is liable to come along later and say "Wait, current MacOS doesn't need this" and delete it.

barracuda156:
@kentonv This will be fixed in `legacy-support` (see reply in the Trac ticket), so I am closing this PR.

PR #2213, opened 2025-01-14
Attempt to add equality comparison to kj::OneOf

Frederik-Baetens:
Using concepts to avoid compiler errors for oneof types that don't implement equality as an alternative to sfinae.

### DIFF
--- a/c++/src/kj/one-of-test.c++
+++ b/c++/src/kj/one-of-test.c++
@@ -210,6 +210,44 @@ KJ_TEST("OneOf copy/move from alternative variants") {
   }
 }
 
+KJ_TEST("OneOf equality") {
+  {
+    OneOf<int, bool> a = 5;
+    OneOf<int, bool> b = 5;
+    KJ_EXPECT(a == b);
+  }
+  {
+    OneOf<int, bool> a = false;
+    OneOf<int, bool> b = false;
+    KJ_EXPECT(a == b);
+  }
+  {
+    OneOf<int, bool> a = true;
+    OneOf<int, bool> b = false;
+    KJ_EXPECT(a != b);
+  }
+  {
+    OneOf<int, bool> a = 0;
+    OneOf<int, bool> b = false;
+    KJ_EXPECT(a != b);
+  }
+  {
+    OneOf<int, bool> a = 1;
+    OneOf<int, bool> b = true;
+    KJ_EXPECT(a != b);
+  }
+  {
+    OneOf<int, bool> a = 5;
+    OneOf<int, bool> b = 6;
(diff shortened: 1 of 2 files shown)


Frederik-Baetens:
I also tried doing something similar to make str() work on oneof objects if all variants supported str(), through the operator* implementation, but that caused compiler errors in unrelated files. I think those might have been related to some macros forcibly using str() on oneofs that didn't support it, and thus causing compiler errors? Should I just get around that by printing a default value for non-str supporting variants?

Frederik-Baetens:
> Concepts are neat, but are they really necessary for this?
> 
> The `operator==` template will only be instantiated if you try to use it, so if we just define it without any gate, it won't cause a problem unless someone actually tries to compare two OneOfs that have non-comparable members. And then, they will get a compiler error which tells them exactly which member is non-comparable, which actually seems more useful than getting an error saying that the entire OneOf is non-comparable.

:+1: Dropped the use of concepts.

c++/src/kj/one-of-test.c++
```diff
@@ -210,6 +210,24 @@ KJ_TEST("OneOf copy/move from alternative variants") {
...
+  {
+    OneOf<int, bool> a = 5;
+    OneOf<int, bool> b = 6;
+    KJ_EXPECT(a != b);
+  }
+  {
+    OneOf<int, bool> a = 5;
+    OneOf<int, bool> b = true;
+    KJ_EXPECT(a != b);
+  }
```
kentonv, line 228: Can you add test cases for:
* Both are bool, and equal (verifies that equality is judged correctly for more than just the first type in the list).
* Both are bool, and non-equal.
* Compare int zero to bool false (should not be equal, even though C++ considers these values equal if compared directly).

c++/src/kj/one-of.h
```diff
@@ -652,6 +652,32 @@ void OneOf<Variants...>::allHandled() {
 // - _kj_switch_done is really used as a boolean flag to prevent the for() loop from actually
 //   looping, but it's defined as a pointer since that's all we can define in this context.
 
+// Helper that tries comparing a and b as type T, but only if a.is<T>().
+template <typename T, typename ...Variants>
+bool compareIfIs(const OneOf<Variants...>& a, const OneOf<Variants...>& b) {
```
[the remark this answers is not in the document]
Frederik-Baetens: moved it into namespace _

c++/src/kj/one-of-test.c++
Frederik-Baetens: Added these test cases

PR #2215, opened 2025-01-17
Make mutex-test less flaky

### LINKED ISSUE #2208: mutex-test is flaky
I observed an expectation failure on this line under a UBSAN build: https://github.com/capnproto/capnproto/blob/4c2071bf467cb81622c1d86e083e957027dee5bd/c%2B%2B/src/kj/mutex-test.c%2B%2B#L94.

```
    kj/mutex-test.c++:94: failed: expected maybeLock == kj::none [(can't stringify) == (can't stringify)]; duration = 20.285787ms
```

I think this can come about if the thread started on line 89 does not actually start until after the `delay()` calls on lines 105 and 106 have at least partially completed. This seems to be supported by the reported timeout thread's `duration` of over 20ms, which is greater than the two `delay()` calls combined, and is much greater than the timeout thread's timeout of 8ms. Presumably there was about a 12ms+ delay between the `kj::Thread` constructor being called and the thread actually beginning to run. On a UBSAN build, this is perhaps not too surprising.

Perhaps we could use a second mutex to make sure the timeout thread is about to start its lock attempt before the main thread begins its `delay()` calls? I'm not sure if that would truly fix the flake, or if it would just make it less likely. If it would just make it less likely, perhaps an easier thing to do is add a few more `delay()` calls.

harrishancock:
Fixes #2208.

Also includes some unrelated drive-by typo fixes.

### DIFF
--- a/c++/src/kj/async-inl.h
+++ b/c++/src/kj/async-inl.h
@@ -197,7 +197,7 @@ public:
   virtual void traceEvent(TraceBuilder& builder) = 0;
   // Build a trace of the callers leading up to this event. `builder` will be populated with
   // "return addresses" of the promise chain waiting on this event. The return addresses may
-  // actually the addresses of lambdas passed to .then(), but in any case, feeding them into
+  // actually be the addresses of lambdas passed to .then(), but in any case, feeding them into
   // addr2line should produce useful source code locations.
   //
   // `traceEvent()` may be called from an async signal handler while `fire()` is executing. It
@@ -298,7 +298,7 @@ public:
   // when it is ready, and then TransformPromiseNode applies the .then() transformation during the
   // call to .get().
   //
-  // So, when we trace the chain of Events backwards, we end up hoping over segments of
+  // So, when we trace the chain of Events backwards, we end up hopping over segments of
   // TransformPromiseNodes (and other similar types). In order to get those added to the trace,
   // each Event must call back down the PromiseNode chain in the opposite direction, using this
   // method.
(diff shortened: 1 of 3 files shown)


harrishancock:
macOS failed with
```
[ TEST ] kj/async-io-test.c++:669: legacy test: AsyncIo/CapabilityPipeBlockedSendStream
kj/async-io-test.c++:707: failed: expected endpoint2->readAllText().wait(io.waitScope) == "foo" [ == foo]
[ FAIL ] kj/async-io-test.c++:669: legacy test: AsyncIo/CapabilityPipeBlockedSendStream (203878 μs)
stack: 10281b137 10281a9e3 10346a47b 103469cf7 103ad11db 103acf67b 103acf3e7 103469237 198c6b153
    kj::(anonymous namespace)::TestCase669::run() (in capnp-test) + 1155: returning here
    kj::TestRunner::run() (in libkj-test-2.0-dev.dylib) + 551: returning here
    kj::MainBuilder::MainImpl::operator()(kj::StringPtr, kj::ArrayPtr<kj::StringPtr const>) (in libkj-2.0-dev.dylib) + 1787: returning here
    kj::runMainAndExit(kj::ProcessContext&, kj::Function<void (kj::StringPtr, kj::ArrayPtr<kj::StringPtr const>)>&&, int, char**) (in libkj-2.0-dev.dylib) + 415: returning here
    main (in libkj-test-2.0-dev.dylib) + 119: returning here
    start (in dyld) + 2475: returning here
```
Re-running.

kentonv (approved):

PR #2216, opened 2025-01-17
Add Executor::newPromiseAndCrossThreadFulfiller()

harrishancock:
Currently, `newPromiseAndCrossThreadFulfiller()` can only create Promises associated with the calling thread's event loop. This is the overwhelmingly common use case for CrossThreadPromiseFulfillers.

Recently, while working on the ability to `co_await` Rust Futures across an FFI boundary, I encountered a case where it would be helpful to create Promises associated with a different thread's event loop.

To elaborate on the use case:
- Rust Futures have a `poll()` function which drives the Future to make some forward progress.
- The caller of `poll()` must pass a reference to a Waker object to `poll()`. Let's call this Waker the "initial Waker". It lives on the stack of the thread which calls `poll()`.
- Waker is an abstract interface, with any number of actual implementations.
- If the Future is not yet ready, it stores a clone of the initial Waker somewhere, and arranges to call `Waker::wake()` asynchronously on the clone when it thinks it can next make forward progress. Let's call this second Waker the "cloned Waker".
- Rust may call `Waker::wake()` on the cloned Waker from any thread.
- So, when awaiting Futures from KJ, CrossThreadPromiseFulfiller serves as a good implementation for the cloned Waker. The associated Promise can be arranged to fire an Event which calls `Future::poll()`.
- But creating a CrossThreadPromiseFulfiller has some overhead -- memory allocations, and a slow-path toward fulfilling the Promise. So ideally, we won't create the CrossThreadPromiseFulfiller unless we know we actually need it -- that is, when Rust tries to clone the initial Waker.
- The only problem: a Rust Future may clone the Waker from a different thread, as long as it does so before the call to `poll()` on the main thread returns. For this reason, it would be helpful to be able to call `newPromiseAndCrossThreadFulfiller()` from any thread.

This commit adds an `Executor::newPromiseAndCrossThreadFulfiller()` member function, which allows for the above use case. It's a very narrow use case, and unlikely to be commonly useful. But, the Rust API allows it, so I need to handle it.

~~One final note: the new function is a bit more dangerous than the existing `newPromiseAndCrossThreadFulfiller()` free function. This is because, if your code fails to transfer the created Promise back to its "home" thread before that thread exits, your process _will_ crash when the fulfiller tries to fulfill/reject the Promise (and if you destroy the fulfiller, it implicitly rejects the Promise). In the Rust use case delineated above, that synchronization naturally occurs because the Waker can only be cloned during the main thread's call to `poll()`.~~

This change exposed a footgun: callers needed to arrange to move the promise back to its native event loop before that event loop exited, or else they risk UB. To make this situation safer, the XThreadPaf now holds a strong reference to the Executor, and fulfilling such a Promise after its event loop has exited is now a no-op.

### DIFF
--- a/c++/src/kj/async-inl.h
+++ b/c++/src/kj/async-inl.h
@@ -2014,7 +2014,7 @@ class XThreadFulfiller;
 
 class XThreadPaf: public PromiseNode {
 public:
-  XThreadPaf();
+  XThreadPaf(Own<const Executor> executor);
   virtual ~XThreadPaf() noexcept(false);
   void destroy() override;
 
@@ -2057,9 +2057,9 @@ private:
     // object.
   } state;
 
-  const Executor& executor;
-  // Executor of the waiting thread. Only guaranteed to be valid when state is `WAITING` or
-  // `FULFILLING`. After any other state has been reached, this reference may be invalidated.
+  Own<const Executor> executor;
+  // Executor of the waiting thread. We hold a strong reference to it so that we have no risk of UB
+  // if the waiting thread exits before the promise is fulfilled.
 
   ListLink<XThreadPaf> link;
   // In the FULFILLING/FULFILLED states, the object is placed in a linked list within the waiting
@@ -2080,6 +2080,8 @@ private:
 template <typename T>
 class XThreadPafImpl final: public XThreadPaf {
 public:
+  using XThreadPaf::XThreadPaf;
+
   // implements PromiseNode ----------------------------------------------------
   void get(ExceptionOrValue& output) noexcept override {
     output.as<FixVoid<T>>() = kj::mv(result);
@@ -2101,7 +2103,7 @@ public:
   FulfillScope(XThreadPaf** pointer);
   // Atomically nulls out *pointer and takes ownership of the pointer.
 
-  ~FulfillScope() noexcept;
+  ~FulfillScope() noexcept(false);
 
   KJ_DISALLOW_COPY_AND_MOVE(FulfillScope);
 
@@ -2167,7 +2169,12 @@ public:
 
 template <typename T>
 PromiseCrossThreadFulfillerPair<T> newPromiseAndCrossThreadFulfiller() {
-  kj::Own<_::XThreadPafImpl<T>, _::PromiseDisposer> node(new _::XThreadPafImpl<T>);
+  return getCurrentThreadExecutor().newPromiseAndCrossThreadFulfiller<T>();
+}
+
+template <typename T>
+PromiseCrossThreadFulfillerPair<T> Executor::newPromiseAndCrossThreadFulfiller() const {
+  kj::Own<_::XThreadPafImpl<T>, _::PromiseDisposer> node(new _::XThreadPafImpl<T>(addRef()));
   auto fulfiller = kj::heap<_::XThreadFulfiller<T>>(node);
   return { _::PromiseNode::to<_::ReducePromises<T>>(kj::mv(node)), kj::mv(fulfiller) };
 }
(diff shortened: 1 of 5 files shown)


c++/src/kj/async-inl.h
```diff
@@ -1985,36 +1985,14 @@ class XThreadEventImpl<Func, Promise<T>> final: public XThreadEvent {
   friend Executor;
 };
 
-}  // namespace _ (private)
-
-template <typename Func>
```
harrishancock, line 1990: I moved these down below so that all the Executor member functions are all defined together. I'm happy to leave them up here, too, though.

c++/src/kj/async-xthread-test.c++
```diff
@@ -1040,5 +1040,94 @@ KJ_TEST("cross-thread fulfiller multiple fulfills") {
...
+    // Get our Promise from the main thread.
+    Promise<int>* promisePtr;
+    {
+      auto lock = promise.lockExclusive();
+      lock.wait([&](const Maybe<Promise<int>>& value) { return value != kj::none; });
+      promisePtr = &KJ_ASSERT_NONNULL(*lock);
+    }
+
+    // Wait for the main thread to send us the MAGIC_VALUE.
+    KJ_EXPECT(promisePtr->poll(waitScope));
```
harrishancock, line 1064: Will remove this line, it's obviously flaky in retrospect.

c++/src/kj/async-xthread-test.c++
```diff
@@ -1040,5 +1040,94 @@ KJ_TEST("cross-thread fulfiller multiple fulfills") {
...
+    })();
+  }
+
+  ([&]() noexcept {
+    auto& exec = KJ_ASSERT_NONNULL(ownExecutor);
+
+    // The other thread has now exited.
+    KJ_EXPECT(!exec->isLive());
+
+    KJ_EXPECT_THROW_MESSAGE("Executor is not live", exec->newPromiseAndCrossThreadFulfiller<void>());
```
harrishancock, line 1128: This crashes in NDEBUG mode, because the KJ_IREQUIRE gets compiled out, and instead an exception hits a `noexcept` boundary later on.

c++/src/kj/async-inl.h
harrishancock: I think I'll move them back, since Executor's function implementations are already a bit spread out, I now see.

c++/src/kj/async.h
```diff
@@ -1093,6 +1098,21 @@ class Executor {
...
   // executor thread has signaled completion. The return value is transferred between threads.
 
+  template <typename T>
+  PromiseCrossThreadFulfillerPair<T> newPromiseAndCrossThreadFulfiller() const;
+  // Like `newPromiseAndCrossThreadFulfiller()`, but the Promise is tied to the event loop
+  // associated with this Executor. This allows the caller to construct Promises associated with
+  // other threads. The caller will have to arrange to move the constructed Promise to its home
+  // thread somehow in order to await or cancel (destroy) it.
+  //
+  // Note that, unless your threads live forever, this function all but _requires_ some cross-thread
```
kentonv, line 1108: Hmm, this seems pretty hard to get right. Can we make it be the case that if the promise is destroyed without ever making it to the target EventLoop, or the target EventLoop is destroyed without waiting on the promise, then fulfilling the fulfiller just does nothing? That seems like the natural answer, but I think it means that `XThreadPaf` needs to hold a strong reference to `Executor`.

harrishancock: Updated with the change. Indeed I ended up just holding a strong reference in XThreadPaf.

harrishancock: https://github.com/capnproto/capnproto/compare/fa48cdbbd5fd9594f6d54ba7df6cc19dd39ad54d..c0bc4f269c1eafc0ad6c5c800e75c7b4dc974eec

kentonv (approved):

PR #2219, opened 2025-01-20
ci: Fix "Linux (clang-14)" job

hebasto:
On the master branch @ 0c56102fe67cbbc43ff05ea54a481c7a676d90bc, the "Linux (clang-14)" job fails with the following error:
```
checking whether clang++-14 supports C++11 library features with -stdlib=libc++... no
configure: error: *** A C++ library with support for C++11 features is required.
```

This PR fixes this issue.


kentonv:
Thanks!

PR #2221, opened 2025-01-21
Fix source of "Promise callback destroyed itself." in RPC system.

kentonv:
In `~RpcClient()`, if the capability was used for streaming, we arrange to wait for all streaming calls to complete in the background, by calling `waitAllAcked()` on the flow controller and then adding the task -- with the flow controller attached -- to the connection's `TaskSet`.

This is problematic, though, if the connection was already disconnected. Normally, we clear the `TaskSet` on disconnect, but this is now adding new tasks into it. Moreover, these tasks can cyclically hold the last remaining references to the connection state itself, since it is no longer connected. When the tasks complete, this can lead to a promise canceling itself, because it drops the last reference to the connection, which destroys the TaskSet.

This change refactors all the `tasks.add(waitAllAcked())` calls into a helper `waitAllAckedInBackground()`. The helper differs from the previous behavior in that if the connection is already disconnected, it throws away the flow controller instead of adding the task.

### DIFF
--- a/c++/src/capnp/rpc.c++
+++ b/c++/src/capnp/rpc.c++
@@ -858,6 +858,18 @@ private:
     }
   }
 
+  void waitAllAckedInBackground(kj::Own<RpcFlowController> flowController) {
+    // Calls `waitAllAcked()` on `flowController` as a background task. This is needed when an
+    // `RpcFlowController` is being abandoned but we don't actually intend to cancel the calls
+    // that were made on it.
+
+    // If we're disconnected, there's no need to wait for anything, so we'll just drop the
+    // flowController in that case.
+    if (connection.is<Connected>()) {
+      tasks.add(flowController->waitAllAcked().attach(kj::mv(flowController)));
+    }
+  }
+
   // =====================================================================================
   // ClientHook implementations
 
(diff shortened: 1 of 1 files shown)


c++/src/capnp/rpc.c++
```diff
@@ -858,6 +858,18 @@ private:
     }
   }
 
+  void waitAllAckedInBackground(kj::Own<RpcFlowController> flowConroller) {
```
jclee, line 861: Nit: `flowConroller` -> `flowController`

c++/src/capnp/rpc.c++
```diff
@@ -858,6 +858,18 @@ private:
...
   }
 
+  void waitAllAckedInBackground(kj::Own<RpcFlowController> flowController) {
+    // Calls `waitAllAcked()` on `flowController` as a background task. This is needed when an
+    // `RpcFlowController` is being abandoned but we don't actually intend to cancel the calls
+    // that were made on it.
+
+    // If we're disconnected, there's no need to wait for anything, so we'll just drop the
+    // flowController in that case.
+    if (connection.is<Connected>()) {
```
jclee, line 868: Seems like the main change is the addition of this "connected" check, which makes sense given the description of the problem in the PR summary, although I must admit I don't yet have a firm enough grasp of the ownership graph to have tracked the problem down myself.

But, yeah, it looks like this check plus the `tasks.clear()` in `disconnect()` would maintain an invariant that ensures that `RpcConnectionState.tasks` will only contain flowController tasks for connections in the Connected state.

jclee (approved):

PR #2231, opened 2025-01-27
Implement newWebSocketPipe with newTwoWayPipe.

danlapid:
Blocked on validating upstream branch tests pass


c++/src/capnp/compat/websocket-rpc.c++
```diff
@@ -118,10 +118,7 @@ kj::Promise<void> WebSocketMessageStream::end() {
           //
           // * https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/close
           // * https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent
-
```
danlapid, line 121: http.c++:3603 disallows reason on 1005, I can either allow it there or remove the reason here. Seems like we were just inconsistent before.

c++/src/capnp/compat/websocket-rpc-test.c++
```diff
@@ -101,8 +101,8 @@ KJ_TEST("WebSocketMessageStreamByteCount") {
   msgStreamB.tryReadMessage(nullptr).wait(waitScope);
   pumpTask.wait(waitScope);
   endPromise.wait(waitScope);
-  KJ_EXPECT(pipe1.ends[0]->sentByteCount() == 2585);
-  KJ_EXPECT(pipe1.ends[1]->receivedByteCount() == 2585);
-  KJ_EXPECT(pipe2.ends[0]->sentByteCount() == 2585);
-  KJ_EXPECT(pipe2.ends[1]->receivedByteCount() == 2585);
+  KJ_EXPECT(pipe1.ends[0]->sentByteCount() == 2566);
```
danlapid, line 104: Size change due to reason removal.

kentonv (changes requested):
This looks like a loss to me.

danlapid:
> This looks like a loss to me.

Do you mind expanding on why?

kentonv:
Thanks for explaining your motivation. Please include such explanations in the PR description in the future.

I'm against this because:

1. This change makes WebSocketPipe much less efficient as it now has to serialize and parse messages.

2. `WebSocketPipe` is a tool that facilitates optimized pumping, and this change actually breaks that!

`WebSocketPipe` is designed to shorten the paths of pumps. If you pump into one end, and out the other end, this gets reduced to a single end-to-end pump. So if you have a pump chain like `someSocket > WebSocketPipe > WebSocketPipe > WebSocketPipe > otherSocket`, this actually reduces to `someSocket > otherSocket`, as in literally `someSocket.pumpTo(otherSocket)` ends up being invoked, and if the two sockets implement an optimized pump, that can then kick in.

With your change, `someSocket.pumpTo(otherSocket)` doesn't end up happening.

Now, it's true that, if `someSocket` and `otherSocket` are both _native_ sockets, then you will perhaps end up with an end-to-end native pump just because each WebSocket > WebSocket pump in the chain will be optimized to a direct stream-to-stream pump. Then you fall back to the KJ stream pumping optimizations which work in much the same way. However:
* This will only work if you carefully set the compression settings on each WebSocket to be compatible, as optimized pumping between two native WebSockets is only possible when they use the same settings.
* You are still breaking any form of optimized pump where the ends are not native sockets. For example, http-over-capnp implements an optimized pump when an RPC-based WebSocket is pumped to another RPC-based WebSocket; in this case, one capnp capability ends up redirecting to the other, which allows path shortening to kick in, which can entirely remove the RPC node from the path. But if you have WebSocketPipes in between that are implemented by serializing and reparsing messages then this optimization ends up blocked.

danlapid:
Closing in favor of an alternative approach.

PR #2235, opened 2025-02-06
Avoid gcc/clang ABI incompatibility caused by PlacementNew

ryanofsky:
GCC and clang do not use same calling convention for passing empty struct parameters. There is more information about this in https://itanium-cxx-abi.github.io/cxx-abi/cxx-abi-dev/archives/2015-December/002869.html

Unfortunately this can create an issue in capnproto if it is built without optimizations in GCC, and the resulting static libraries are used in a clang program, or vice versa.

Depending on what order libraries are specified on the linker command line, and whether code compiled with the other compiler is calling any header functions that cause weak a `operator new(unsigned int, kj::_::PlacementNew, void*)` symbol to be defined in its own objects, this can cause the linker to link a GCC-generated `kj::ctor` with a clang-generated `operator new`, and the resulting program to crash due to the compilers using different calling conventions for `operator new`.

This problem is difficult to avoid in general, but pretty easy to avoid here by changing `operator new` parameter order so the empty struct parameter is last.

This change should be beneficial for capnproto users that may be compiling it without optimizations, and not necessarily using a single compiler to build all their dependencies.

The problem does not occur if any optimizations are enabled because `operator new` calls are inlined in that case.

This problem was encountered in https://github.com/bitcoin/bitcoin/issues/31772

### DIFF
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -1034,24 +1034,27 @@ private:
 
 // We want placement new, but we don't want to #include <new>.  operator new cannot be defined in
 // a namespace, and defining it globally conflicts with the definition in <new>.  So we have to
-// define a dummy type and an operator new that uses it.
+// define a dummy type and an operator new that uses it.  The dummy type is intentionally passed
+// as the last parameter so clang and GCC ABI calling conventions for empty struct struct parameters
+// are compatible, and there are not segfaults trying to call clang operator new/delete from GCC or
+// vice versa.
 
 namespace _ {  // private
 struct PlacementNew {};
 }  // namespace _ (private)
 } // namespace kj
 
-inline void* operator new(size_t, kj::_::PlacementNew, void* __p) noexcept {
+inline void* operator new(size_t, void* __p, kj::_::PlacementNew) noexcept {
   return __p;
 }
 
-inline void operator delete(void*, kj::_::PlacementNew, void* __p) noexcept {}
+inline void operator delete(void*, void* __p, kj::_::PlacementNew) noexcept {}
 
 namespace kj {
 
 template <typename T, typename... Params>
 inline void ctor(T& location, Params&&... params) {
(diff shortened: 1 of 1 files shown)


kentonv:
Yikes!

I guess this hasn't come up earlier because the most likely time when you'd be compiling an application with a different compiler from libkj is when libkj is installed globally on the system, and in that case it would probably be an optimized build so the globally-installed library would not include the incompatible inline instantiations?

Or is it also the case that this only applies to static libraries and not dynamic? That would also explain why it hasn't been an issue for installed libs.

On another note, you should probably target this at the `master` branch, since that's the v1.x release branch, and I suppose we might want this to be in the next point release? The `v2` branch is an unstable development branch.

ryanofsky:
Thanks I rebased this on master, but I don't seem to have ability to change target branch in GitHub. This might have to do with github permissions. It does seem possible to change the target branch according to https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request

And yes I think the main reason this issue would not come up normally is that a globally-installed library would be probably be optimized and calls to these functions would be inlined. But also the issue depends on what link order and what function calls the client program is making, and what weak symbols will be included in its object files. So a combination of things have to happen for there to be a problem. I do think the same problem could happen for dynamic libraries, since weak symbols are handled similarly by default in static and dynamic libraries. Also I forgot to mention that this issue came when compiling for i386, so I'm not actually sure if it affects x86_64.

kentonv:
Hmm, I seem to be able to change the base branch, so I did. I thought PR authors had been able to do that in the past, weird.

kentonv:
Huh, the parent commit of your PR seems to be from two years ago for some reason. But I suppose it's perfectly mergeable as this code hasn't changed in a very long time.

kentonv:
Thanks!

PR #2236, opened 2025-02-10
Reject unfulfilled promises on WaiterQueue destruction.

AnthonyVH:
When a `WaiterQueue` is destroyed with unfulfilled promises, these promises become "dangling". I.e. waiting on them will hang indefinitely. This PR changes that by rejecting all unfulfilled promises remaining in the queue.

### DIFF
--- a/c++/src/kj/async-queue.h
+++ b/c++/src/kj/async-queue.h
@@ -41,6 +41,14 @@ public:
   WaiterQueue() = default;
   KJ_DISALLOW_COPY_AND_MOVE(WaiterQueue);
 
+  ~WaiterQueue() noexcept(false) {
+    while (!empty()) {
+      reject(KJ_EXCEPTION(FAILED, "WaiterQueue destroyed while some promises returned by wait()"
+        " are still outstanding. This is undefined behavior. You must cancel all promises before"
+        " destroying the WaiterQueue."));
+    }
+  }
+
   Promise<T> wait() {
     return newAdaptedPromise<T, Node>(queue);
   }


AnthonyVH:
Alternatively, this could also be implemented at the `WaiterQueue::BaseNode/Node` level, or `PromiseFulfiller`.

Seeing as to how `PromiseFulfiller` doesn't reject the associated `Promise` on destruction if it's unfulfilled makes me think this "leave the `Promise` unfulfilled" might be an intentional choice though. Is it?

kentonv:
Sorry for the slow reply.

Looking closely at this, I think the reason the existing code didn't cover this is because it's not supposed to happen: [The KJ style guide says](https://github.com/capnproto/capnproto/blob/v2/style-guide.md#ownership) that, unless otherwise indicated, when a method returns a promise, it's the caller's responsibility to ensure that the object outlives the promise. Hence, if a promise returned by `WaiterQueue::wait()` is still outstanding, then it is an error (and potentially undefined behavior) for `WaiterQueue` to be destroyed at all. The caller _must_ await or cancel (destroy) all such promises before destroying the `WaiterQeueue`. Note that `kj::Canceler` is a useful tool to accomplish this, and can be used to achieve the same behavior as this change.

With that said, I'm willing to be convinced that this change is a good idea because it changes undefined behavior into merely an exception. Yes, a program that triggers it is wrong, but bugs are unavoidable and if we can make them not memory errors that's probably a good thing. I would say, though, that the exception error messages should be clear that the caller is buggy here.

c++/src/kj/async-queue.h
```diff
@@ -41,6 +41,12 @@ class WaiterQueue {
   WaiterQueue() = default;
   KJ_DISALLOW_COPY_AND_MOVE(WaiterQueue);
 
+  ~WaiterQueue() {
+    while (!empty()) {
+      reject(KJ_EXCEPTION(FAILED, "WaiterQueue destroyed"));
```
kentonv, line 46: Per my comment above, I'd make the text here like:

> WaiterQueue destroyed while some promises returned by `wait()` are still outstanding. This is undefined behavior. You must cancel all promises before destroying the `WaiterQueue`.

c++/src/kj/async-queue.h
```diff
@@ -41,6 +41,12 @@ class WaiterQueue {
   WaiterQueue() = default;
   KJ_DISALLOW_COPY_AND_MOVE(WaiterQueue);
 
+  ~WaiterQueue() {
```
kentonv, line 44: By KJ convention this should be:

```suggestion
  ~WaiterQueue() noexcept(false) {
```

AnthonyVH:
Thanks for the review! I've updated the code as requested.

I think changing this from undefined behavior to an explicit crash is much preferable. I took me a while to figure out why my code was hanging without this change, so if this can help others then all the better.

kentonv:
Thanks! Sorry for slow responses... swamped as always.

PR #2240, opened 2025-02-17
Add Thread::getCpuTime() to read another thread's CPU usage.

kentonv:
(Blind-coded the Windows version, let's see if it works in CI...)


kentonv:
OMG I blind-coded the Win32 version successfully but it turns out MacOS just straight-up doesn't implement the POSIX standard function so it fails.

mikea (approved):

Issue #1768, opened 2023-08-28
Permission denied to "/" 


chaudhryfaisal:
trying to compile capnproto in temux on android 11 which does not give access to "/" 
```bash
# download and build
 wget -c --tries=inf https://capnproto.org/capnproto-c++-${capnproto_c_ver}.tar.gz -O - | tar -xz
  cd capnproto-c*
  CXXFLAGS="-fPIC -O2" ./configure --prefix=$PREFIX
  make -j $(nproc)

....
./capnp compile --src-prefix=./src -o./capnpc-c++:src -I./src $(for FILE in src/capnp/test.capnp src/capnp/test-import.capnp src/capnp/test-import2.capnp src/capnp/compat/json-test.capnp; do echo ./$FILE; done)
*** Fatal uncaught kj::Exception: kj/filesystem-disk-unix.c++:1661: failed: open(dir, O_RDONLY | MAYBE_O_CLOEXEC | MAYBE_O_DIRECTORY): Permission denied
make: *** [Makefile:4232: test_capnpc_middleman] Error 1

#simple help command fails
./capnp --help
*** Fatal uncaught kj::Exception: kj/filesystem-disk-unix.c++:1661: failed: open(dir, O_RDONLY | MAYBE_O_CLOEXEC | MAYBE_O_DIRECTORY): Permission denied

#strace ./capnp --help
strace ./capnp --help
openat(AT_FDCWD, "/dev/__properties__/u:object_r:vendor_socket_hook_prop:s0", O_RDONLY|O_NOFOLLOW|O_CLOEXEC) = 3
openat(AT_FDCWD, "/", O_RDONLY|O_CLOEXEC|O_DIRECTORY) = -1 EACCES (Permission denied)
*** Fatal uncaught kj::Exception: kj/filesystem-disk-unix.c++:1661: failed: open(dir, O_RDONLY | MAYBE_O_CLOEXEC | MAYBE_O_DIRECTORY): Permission denied

#ls to / root also fails
ls /
ls: cannot open directory '/': Permission denied

#strace ls
strace ls /
...
openat(AT_FDCWD, "/", O_RDONLY|O_CLOEXEC|O_DIRECTORY) = -1 EACCES (Permission denied)
```

I see "/" is used here 
```cpp
1635 class DiskFilesystem final: public Filesystem {
1636 public:
1637   DiskFilesystem()
1638       : root(openDir("/")),
1639         current(openDir(".")),
[...]

PR #2246, opened 2025-03-05
Platform-dependent fix for root permissions error

EmmaJaneBonestell:
At least Linux (including Android) and FreeBSD provide an extension for `open`, [O_PATH](https://man7.org/linux/man-pages/man2/open.2.html#:~:text=O_PATH%20(since%20Linux%202.6.39)), that is essentially an inode-only reference; it avoids potential permission issues. When available, replacing O_RDONLY with it in DiskFilesystem's openDir appears to be compatible with capnproto's use case.

When compiling for Android API >= 30 (for memfd_create), testing on an Android device passes all tests except the DiskDirectory hard linking test (https://github.com/capnproto/capnproto/blob/20af9fe2de10a2e7f621a212351d20860cc32a0d/c%2B%2B/src/kj/filesystem-disk-test.c%2B%2B#L605) , which is not allowed on Android. (Could only test on Android without fibers, as some parts of libucontext will fail.)

Fixes https://github.com/capnproto/capnproto/issues/1768 .

### DIFF
--- a/c++/src/kj/filesystem-disk-unix.c++
+++ b/c++/src/kj/filesystem-disk-unix.c++
@@ -1675,7 +1675,25 @@ private:
   Path currentPath;
 
   static OwnFd openDir(const char* dir) {
-    auto result = KJ_SYSCALL_FD(open(dir, O_RDONLY | MAYBE_O_CLOEXEC | MAYBE_O_DIRECTORY));
+    int fd;
+    KJ_SYSCALL_HANDLE_ERRORS(fd = open(dir, O_RDONLY | MAYBE_O_CLOEXEC | MAYBE_O_DIRECTORY)) {
+      case EACCES:
+        // If we don't have read permission, fall back to O_PATH if available
+#ifdef O_PATH
+        {
+          auto result = KJ_SYSCALL_FD(open(dir, O_PATH | MAYBE_O_CLOEXEC | MAYBE_O_DIRECTORY));
+#ifndef O_CLOEXEC
+          setCloexec(result);
+#endif
+          return result;
+        }
+	KJ_FALLTHROUGH;
+#endif
+      default:
+        KJ_FAIL_SYSCALL("open(dir, O_RDONLY)", error, dir);
+    }
+
+    auto result = OwnFd(fd);
 #ifndef O_CLOEXEC
     setCloexec(result);
 #endif


kentonv:
Hmm, I think if someone does `fs.getRoot().list()` or `fs.getCurrent().list()`, this will break them, since the root and current directories are no longer open for reading?

What is the problem you're trying to solve this change, exactly?

c++/src/kj/async-io-test.c++
```diff
@@ -176,8 +176,10 @@ TEST(AsyncIo, SimpleNetworkAuthentication) {
 TEST(AsyncIo, UnixSocket) {
   auto ioContext = setupAsyncIo();
   auto& network = ioContext.provider->getNetwork();
-
-  auto path = kj::str(TMPDIR "/kj-async-io-test.", getpid());
+ 
+  const char* tmpDir_env = getenv("TEST_TMPDIR");
+  const auto tmpDir = tmpDir_env != nullptr ? tmpDir_env : TMPDIR;
+  auto path = kj::str(tmpDir, "/kj-async-io-test.", getpid());
```
kentonv, line 182: Unfortunately this change will likely break bazel users as bazel's TEST_TMPDIR is often a very long path, and unix sockets have a length limit of 104 characters.

This change isn't described in the PR description, was this intended to be part of the PR?

EmmaJaneBonestell:
This should fix issues with it breaking any actua read/write filesystem operations. Some quick local tests for listing, reading, writing,  deleting, etc., work now, ll instead of terminating from a bad fd.

You could place this behind a cmake flag for constrained environments rather than always enabling when O_PATH is available.

kentonv:
> It fixes the issue that was linked, https://github.com/capnproto/capnproto/issues/1768

Oh sorry, somehow I missed that.

I think a simpler fix might be:
- Try open() with O_RDONLY first.
- If that fails, try again with O_PATH. In this case, trying to call `list()` will fail naturally since the FD doesn't have permissions -- but I think that's just fine.

Sorry you already did all this work on a different solution, but I think this simple solution is probably best.

EmmaJaneBonestell:
> Sorry you already did all this work on a different solution, but I think this simple solution is probably best.

It's no problem. I developed this solution to fix a capnproto failure where it was used as a transitive dependency in another project, and it still accomplished that.

I'd be happy to modify the PR to try O_RDONLY first, then fallback to O_PATH on systems where it's available. That's a better approach than just starting out with O_PATH as I did.

However, if I understand correctly, it seems like you only want to have a workaround for EBADACCES, and not address the list() (read access) issue that you initially brought up? To clarify, eliminating the `getReadableFd()` mechanism I implemented would not cause just list() to fail; all operations besides path navigation would fail with an unhandled EBADF/EBADFD.

If you'd rather take a different approach than what I've implemented here, I completely understand. You're welcome to close this PR and incorporate any parts of this solution that you find useful for your preferred implementation.

kentonv:
> However, if I understand correctly, it seems like you only want to have a workaround for EBADACCES, and not address the list() (read access) issue that you initially brought up?

`list()` will work correctly as long as the initial open attempt with `O_RDONLY` worked. If that initial open attempt failed (and we fell back to `O_PATH`), then clearly we don't have the ability to list the directory anyway, so it's expected that `list()` should fail in that case.

> If you'd rather take a different approach than what I've implemented here, I completely understand. You're welcome to close this PR and incorporate any parts of this solution that you find useful for your preferred implementation.

Sorry but I don't have bandwidth to work on this myself. If you change the PR like I suggested I'm happy to merge it. If you don't need this anymore and would rather abandon it, that's fine with me.

EmmaJaneBonestell:
>If that initial open attempt failed (and we fell back to `O_PATH`), then clearly we don't have the ability to list the directory anyway, so it's expected that `list()` should fail in that case.

I feel like there's a disconnect here, because this is wrong.

This specific issue is not regarding permissions for an arbitrary directory, but for the root directory "/"; because the DiskFilesystem class is designed around an anchored file descriptor to "/",  if you do not have read permissions for "/", the program entirely fails.

Either I did a very poor job explaining, or you seem to be under the misapprehension that not having read permissions for "/" means that you cannot have read permissions for subdirectories and files beneath "/". That is objectively not how Linux permissions work.  It is not relevant to this issue, but that scenario woulx only happen if your user had no execute permissions for a directory, and  then O_PATH would fail, too.

Instead of trying to discuss this in terms of the lower-level code in capnproto, you can see an exact replication of this behavior with a short shell script:
```sh
#!/bin/sh

root="${PWD}/root"

mkdir --parents "${root}"/sub_dir && \
chmod a-rw "${root}" && \
ls "${root}"
# This will FAIL with a permissions error
touch "${root}"/sub_dir/file1 "${root}"/sub_dir/file2
# This will successfully create root/sub_dir/file1 and root/sub_dir/file2 
# despite having neither read or write ermissions for "root/".
ls --numeric-uid-gid "${root}"/sub_dir
# Same thing -- ls will successfully list the contents of "root/sub_dir/",
# the files we just made, even without "root/" permissions.
```

kentonv:
Yes, I understand that permissions are not transitive to subdirectories.

If someone wants to read _a subdirectory_, they need to open it first:

```
auto& root = fs.getRoot();
auto dir = root.openSubdir(somePath);
dir->list();
```

Under my proposal, if the root directory is not readable, then it would be opened as O_PATH. But when you call `root.openSubdir(somePath)`, assuming that `somePath` _is_ readable, then the subdirectory will be opened O_RDONLY, not O_PATH. Therefore, `dir->list()` will work fine. The only thing that doesn't work is calling `root.list()`, which makes sense because the root is not readable.

To be clear, `list()` only returns a shallow listing. It does not recurse into subdirectories.

Does this clear up the confusion?

EmmaJaneBonestell:
Thanks for the clarification. The amended file still fixes the error, and I believe addresses your requested changes.

kentonv:
Thanks, this looks good.

I suppose we probably want to backport this to the 1.x release branch (`master`)? I guess it's probably time for me to do a release, too.

PR #2247, opened 2025-03-05
bazelmod upgrade for v1 branch

sebastianbergt:
I am trying to make capnproto available in the [Bazel Central Registry](https://registry.bazel.build/) [with this PR](https://github.com/bazelbuild/bazel-central-registry/pull/3947).
So contributing it here, since I have the changes anyway and it would be great to get them merged :)


kentonv:
This appears to be a large and breaking change to the bazel build. I don't think we can merge this as people may be depending on this branch being stable.

sebastianbergt:
Yes, you are right about that. Would you accept it for V2?

kentonv:
We don't want v2 published on any sort of registry right now as the API is frequently undergoing breaking changes. We use bazel to build the Workers Runtime, which depends on the v2 branch (we update it at the same time as any breaking change). So any change to the bazel build files would need to be coordinated with the workers runtime. Since we're not ready to put this in a registry anyway, it seems like it's probably not worth it to change anything now? But @fhanau and @mikea are the bazel experts here so I'll let them decide if it makes sense.

fhanau:
It's encouraging to see that adopting bzlmod looks pretty straightforward here. A transition to bzlmod should be coordinated with doing the same for the Workers Runtime, where it will be a significantly larger effort. IMHO this should wait until we can enable bzlmod for downstream repos.

sebastianbergt:
My project has no dependency on the workers runtime and I have no desire to migrate it at this time.
However thank you for explaining your point.

PR #2248, opened 2025-03-06
Remove KJ_LIFETIMEBOUND from StringPtr constructor.

erikcorry:
KJ_LIFETIMEBOUND causes false alarms when compiling with clang 21.

For example in workerd:

```
bazel-out/k8-dbg/bin/external/workerd/src/workerd/io/_virtual_includes/io/workerd/api/streams/../basics.h:36:14: error: initializing pointer member 'type' with the stack address of parameter 'ownType' [-Werror,-Wdangling-field]
   36 |       : type(ownType),
      |              ^~~~~~~
bazel-out/k8-dbg/bin/external/workerd/src/workerd/io/_virtual_includes/io/workerd/api/streams/../basics.h:198:17: note: pointer member declared here
  198 |   kj::StringPtr type;
      |                 ^
```

### DIFF
--- a/c++/src/kj/string.h
+++ b/c++/src/kj/string.h
@@ -80,7 +80,7 @@ public:
   }
   inline constexpr StringPtr(const char* begin KJ_LIFETIMEBOUND, const char* end KJ_LIFETIMEBOUND): StringPtr(begin, end - begin) {}
   inline constexpr StringPtr(String&& value KJ_LIFETIMEBOUND) : StringPtr(value) {}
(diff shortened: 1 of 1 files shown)


jclee:
Oh... I think that would actually be flagging a potential bug, if workerd wasn't later moving the String into a member, which clang probably has difficulty seeing -- in which case, the lifetimebound annotation is probably useful to keep.  Looks like it might have already been addressed in workerd here:

https://github.com/cloudflare/workerd/commit/415d733dcd

kentonv:
Fixed in workerd, no problem here.

PR #2249, opened 2025-03-06
Support large ws messages over capnp

danlapid:
* Still in draft stage, gotta make sure all internal tests are passing before actual review.
* This should be mostly close to the final implementation.
* For now this explicitly does not handle WS with extension headers.

### DIFF
--- a/c++/src/capnp/compat/http-over-capnp-test.c++
+++ b/c++/src/capnp/compat/http-over-capnp-test.c++
@@ -1049,9 +1049,9 @@ public:
     ByteStreamFactory streamFactory2;
     kj::HttpHeaderTable::Builder tableBuilder;
     HttpOverCapnpFactory clientFactory(streamFactory1, tableBuilder,
-                                       TEST_PEER_OPTIMIZATION_LEVEL);
+                                       HttpOverCapnpFactory::LEVEL_2);
     HttpOverCapnpFactory serverFactory(streamFactory2, tableBuilder,
-                                       TEST_PEER_OPTIMIZATION_LEVEL);
+                                       HttpOverCapnpFactory::LEVEL_2);
     auto headerTable = tableBuilder.build();
     auto donePaf = kj::newPromiseAndFulfiller<void>();
     auto back = serverFactory.kjToCapnp(kj::heap<DummyWebSocketAccepter>(
--- a/c++/src/capnp/compat/http-over-capnp.c++
+++ b/c++/src/capnp/compat/http-over-capnp.c++
@@ -21,6 +21,7 @@
 
 #include "http-over-capnp.h"
 #include <kj/debug.h>
+#include <kj/async-io.h>
 #include <capnp/schema.h>
 #include <capnp/message.h>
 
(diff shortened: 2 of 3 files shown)


c++/src/capnp/compat/http-over-capnp-test.c++
```diff
@@ -960,14 +990,14 @@ KJ_TEST("HTTP-over-Cap'n-Proto Connect with startTls") {
   auto request = frontCapnpHttpClient->connect(
       "https://example.org"_kj, kj::HttpHeaders(*table), settings);
 
-  KJ_ASSERT_NONNULL(*tlsStarter);
+  auto& tlsStarterCallback = KJ_ASSERT_NONNULL(*tlsStarter);
```
danlapid, line 993: This change is just to appease my editor's syntax highlighter.
The below line "`KJ_ASSERT_NONNULL(*tlsStarter)("example.com")`" caused the rest of the file to be treated as a string..

c++/src/capnp/compat/http-over-capnp.c++
```diff
@@ -731,26 +789,64 @@ public:
     KJ_REQUIRE(!responseSent, "already called send() or acceptWebSocket()");
     responseSent = true;
 
-    auto req = clientContext.startWebSocketRequest();
+    if (headers.isWebSocket() && headers.get(kj::HttpHeaderId::SEC_WEBSOCKET_EXTENSIONS) == kj::none) {
```
danlapid, line 792: The diff ended up being really annoying, not sure if there's a clean way to fix it.
The old logic should remain as-is in the `else` clause of this statement.

c++/src/capnp/compat/http-over-capnp.capnp
```diff
@@ -84,6 +84,10 @@ interface HttpService {
     #
     # Client -> Server WebSocket frames will be sent via method calls on `upSocket`, while
     # Server -> Client will be sent as calls to `downSocket`.
+
+    startOptimizedWebSocket @2 (headers :List(HttpHeader), upStream :ByteStream)
+                            -> (downStream :ByteStream);
+    # TODO(now): comment
```
danlapid, line 90: TODO

c++/src/capnp/compat/http-over-capnp.capnp
```diff
@@ -84,6 +84,12 @@ interface HttpService {
     #
     # Client -> Server WebSocket frames will be sent via method calls on `upSocket`, while
     # Server -> Client will be sent as calls to `downSocket`.
+
+    startOptimizedWebSocket @2 (headers :List(HttpHeader), upStream :ByteStream)
```
mikea, line 88: is this pumping websocket-specific? The signature looks generic enough: could this be generic "pump" call?

danlapid: This is intended to extend the optimizedPumpTo support on KJ-HTTP websockets to http-over-capnp.
Generic http doesn't really have the concept of pump as it's a request-response model.

This specific call is very WS specific as implemented.

c++/src/capnp/compat/http-over-capnp-test.c++
kentonv: Weird, what editor?

c++/src/capnp/compat/http-over-capnp.c++
kentonv: It looks fine if I tell github to hide whitespace changes.

c++/src/capnp/compat/http-over-capnp-test.c++
danlapid: vscode 😞 
I can revert it back if you dislike it, just assumed you wouldn't care much

kentonv: No it's fine, I was just curious.

I also use vscode and... wow, you're right, the highlight is all messed up after `("example.com")`

PR #2250, opened 2025-03-06
kj::Arc::addRef() const version

mikea:
const in KJ means here "thread-safe", expose the const version of addRef.

### DIFF
--- a/c++/src/kj/refcount-test.c++
+++ b/c++/src/kj/refcount-test.c++
@@ -321,4 +321,12 @@ KJ_TEST("Arc disown / reown") {
   KJ_EXPECT(b == true);
 }
 
+
+KJ_TEST("Arc<const T>") {
+  bool b = false;
+
+  const kj::Arc<AtomicSetTrueInDestructor> ref1 = kj::arc<AtomicSetTrueInDestructor>(&b);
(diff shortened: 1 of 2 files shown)


c++/src/kj/refcount-test.c++
```diff
@@ -321,4 +321,12 @@ KJ_TEST("Arc disown / reown") {
   KJ_EXPECT(b == true);
 }
 
+
+KJ_TEST("Arc<const T>") {
```
kentonv, line 325: I think what you're testing here is not actually `Arc<const T>`, but rather, `const Arc<T>`. The problem was when the `Arc` itself was const.

Incidentally, I think we should be requiring that `Arc<T>`'s type parameter is _always_ const, because the use of `Arc` implies that multiple threads may hold references, and so any access had better be thread-safe.

kentonv (approved):

c++/src/kj/refcount-test.c++
mikea: You're right, this doesn't make any sense. I have opened #2251 which seems better to me? 

As a result of this I'll upload a PR to workerd later today that introduces `Arc<T>` to `IoOwn<T>` and correctly forces me to declare arc-accessed methods as `const` so it works rather well I think.

PR #2257, opened 2025-03-17
async-io: upgrade read(buffer, size) usage

mikea:


c++/src/kj/async-io.h
```diff
@@ -68,7 +68,7 @@ class AsyncInputStream: private AsyncObject {
     return tryRead(buffer.begin(), minBytes, buffer.size()); 
   }
   Promise<void> read(ArrayPtr<byte> buffer) {
-    return read(buffer.begin(), buffer.size());
+    co_await read(buffer, buffer.size());
```
harrishancock, line 71: Nit: I don't think this needs to be a `co_await`, and introducing it adds a memory allocation. I don't know how much difference it makes, though.

harrishancock (approved):

PR #2260, opened 2025-03-19
nit: use _kj and _kjb where possible

mikea:
Automatic regexp-based translation. Helpful while updating async calls (char => byte) in tests

### DIFF
--- a/c++/src/capnp/compat/json.c++
+++ b/c++/src/capnp/compat/json.c++
@@ -110,16 +110,16 @@ struct JsonCodec::Impl {
     escaped.add('"');
     for (char c: chars) {
       switch (c) {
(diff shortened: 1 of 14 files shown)


mikea:
> Changes LGTM. For string-test.c++ there's a case to be made for keeping StringPtr() in some cases so that it gets test coverage, but I guess the _kj operator is simple enough that we expect them to behave identically.

these tests seem to test various utilitiy methods. I have added a separate StringPtr constructor test.

fhanau:
> > Changes LGTM. For string-test.c++ there's a case to be made for keeping StringPtr() in some cases so that it gets test coverage, but I guess the _kj operator is simple enough that we expect them to behave identically.
> 
> these tests seem to test various utilitiy methods. I have added a separate StringPtr constructor test.

That test looks reasonable, LGTM.

PR #2262, opened 2025-03-24
KJ_DEBUG_MEMORY flag

mikea:
Will enable additional checks in follow up prs too.

Redoes the way KJ_ASSERT_PTR_COUNTERS is defined, which adresses the comment in #2100.

### DIFF
--- a/c++/.bazelrc
+++ b/c++/.bazelrc
@@ -6,6 +6,12 @@ common --enable_platform_specific_config
 common --enable_workspace
 common --incompatible_autoload_externally=
 
+# Debug build configuration with maximum runtime checks
+build:dbg -c dbg --//src/kj:debug_memory=True
+
+# Opt build
+build:opt -c opt
+
 build:unix --cxxopt='-std=c++20' --host_cxxopt='-std=c++20' --force_pic --verbose_failures
 build:unix --cxxopt='-Wall'
 build:unix --cxxopt='-Wextra'
--- a/c++/src/kj/memory.h
+++ b/c++/src/kj/memory.h
@@ -22,9 +22,24 @@
 #pragma once
 
 #include "common.h"
-#ifdef KJ_DEBUG
-#include <atomic>  // std::atomic for KJ_ASSERT_PTR_COUNTERS
+
+// KJ_DEBUG_MEMORY == 1 enables variety of checks designed to catch memory usage errors.
+// KJ_DEBUG_MEMORY undefined or KJ_DEBUG_MEMORY == 0 disables all such checks.
+#if !defined(KJ_DEBUG_MEMORY)
+#define KJ_DEBUG_MEMORY 0
 #endif
+
+// KJ_ASSERT_PTR_COUNTERS == 1 keeps track of active Ptr<T> instances and asserts validity
+// of their ownership.
+// Matches KJ_DEBUG_MEMORY by default.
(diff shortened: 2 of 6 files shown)


c++/src/kj/memory.h
```diff
@@ -22,12 +22,31 @@
...
+
+// KJ_ASSERT_PTR_COUNTERS == 1 keeps track of active Ptr<T> instances and asserts validity
+// of their ownership.
+// Enabled by KJ_DEBUG_MEMORY by default.
+#if !defined(KJ_ASSERT_PTR_COUNTERS)
+#define KJ_ASSERT_PTR_COUNTERS KJ_DEBUG_MEMORY
+#endif
+
+#if KJ_ASSERT_PTR_COUNTERS
+#include <atomic>  // std::atomic for KJ_ASSERT_PTR_COUNTERS
```
mikea, line 47: @kentonv I am not sure what role does KJ_BEGIN_HEADER play? Can I have `#include <atomic>`  here?

mikea: Is it ok to put these defines before KJ_BEGIN_HEADER then?

kentonv: (You can place your `#if` machinery above `KJ_BEGIN_HEADER` as well, just try to write it so it won't generate any warnings even with pedantic settingcs.)

c++/src/kj/memory.h
```diff
@@ -22,9 +22,29 @@
 #pragma once
 
 #include "common.h"
-#ifdef KJ_DEBUG
-#include <atomic>  // std::atomic for KJ_ASSERT_PTR_COUNTERS
-#endif
+
+// KJ_DEBUG_MEMORY == 1 enables variety of checks designed to catch memory usage errors.
+// KJ_DEBUG_MEMORY == 0 disables all such checks.
+// When not set it is initialized to 1 in KJ_DEBUG mode, 0 otherwise.
```
kentonv, line 28: We probably don't want KJ_DEBUG to change the ABI as people commonly build (and install) release libs, then build debug binaries against them. I think KJ_DEBUG_MEMORY does affect the ABI, so we'll have to have it off by default even under KJ_DEBUG I think.

c++/src/kj/memory.h
mikea: done. 🤞

c++/.bazelrc
```diff
@@ -6,6 +6,12 @@ common --enable_platform_specific_config
 common --enable_workspace
 common --incompatible_autoload_externally=
 
+# Debug build configuration with maximum runtime checks
+build:dbg -c dbg --//src/kj:debug_memory=True
```
mikea, line 10: since this is a library these matter only for us.

c++/src/kj/memory.h
mikea: Redid not to enable `KJ_DEBUG_MEMORY` by default. Also updated the CI to exercise debug_memory.

c++/.bazelrc
kentonv: It's actually pretty reasonable to assume that bazel users compile their whole project with consistent flags so this actually seems fine for everyone. It's more the autotool users who will probably build opt builds and then link against them from debug builds.

kentonv (approved):

PR #2263, opened 2025-03-24
run CI in opt mode too

mikea:
I think it makes sense to run tests without KJ_DEBUG too. Added `opt` runs to bazel builds


kentonv:
what? Why?

kentonv:
Sorry, misread.

kentonv:
Specifically I missed the word "too" means "we're running the tests twice, once in debug mode, then again in opt mode".

PR #2264, opened 2025-03-24
kj::ThreadId

mikea:
I intend to use it from `refcount.h`. Let me know if `common.h` is not a best place for this. @kentonv

### DIFF
--- a/c++/src/kj/common-test.c++
+++ b/c++/src/kj/common-test.c++
@@ -24,6 +24,7 @@
 #include <inttypes.h>
 #include <kj/compat/gtest.h>
 #include <span>
+#include "thread.h"
 
 namespace kj {
 namespace {
@@ -1273,6 +1274,23 @@ KJ_TEST("memzero<T>()") {
   for (auto& t: arr) {
     KJ_EXPECT(t.pi == 0);
   }
+}
+
+KJ_TEST("ThreadId") {
+  auto id1 = ThreadId::current();
+  id1.assertCurrentThread();
+
+  auto id2 = ThreadId::current();
+  KJ_ASSERT(id2 == id1);
+  
+  Thread thread([&]() {
+    auto id3 = ThreadId::current();
+    id3.assertCurrentThread();
+    KJ_ASSERT(id1 != id3);
+    KJ_ASSERT(id2 != id3);
+
+    KJ_EXPECT_THROW_MESSAGE("expected id == &currentThreadId", id1.assertCurrentThread());
+  });
 }
 
 }  // namespace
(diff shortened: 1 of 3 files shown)


c++/src/kj/common.c++
```diff
@@ -44,6 +44,15 @@ void unreachable() {
   KJ_KNOWN_UNREACHABLE(abort());
 }
 
+thread_local bool currentThreadId;
```
kentonv, line 47: Declare this `static` and/or put it in an anonymous namespace to make sure the compiler gives it internal linkage.

(I think `static` is assumed for variables unless you explicitly declare them `extern` but doesn't hurt to be explicit.)

c++/src/kj/common.h
```diff
@@ -2278,6 +2278,23 @@ struct ByteLiteral<1ul> {
 
 }
 
+class ThreadId {
+// Unique thread identifier.
+// Implemented as thread_local address, so could be efficient even for production
+// environments especially with LTO.
```
kentonv, line 2284:
```suggestion
  // Unique thread identifier.
  // Implemented as thread_local address, so could be efficient even for production
  // environments especially with LTO.
```

c++/src/kj/common.h
```diff
@@ -2278,6 +2278,23 @@ struct ByteLiteral<1ul> {
...
+
+public:  
+  ThreadId();
+  // Initializes to current thread id
```
kentonv, line 2288: IMO default constructor should probably initialize to null, or should be deleted.

We should have a static `ThreadId::current()` to get the current thread.

c++/src/kj/common-test.c++
```diff
@@ -1275,5 +1276,20 @@ KJ_TEST("memzero<T>()") {
   }
 }
 
+KJ_TEST("ThreadId") {
+  ThreadId id1;
+  id1.assertCurrentThread();
```
kentonv, line 1281: Be sure to check a failure case, you can use `KJ_ASSERT_THROW_MESSAGE()`.

c++/src/kj/common.h
kentonv, line 1: I guess the argument for adding yet another thing to common.h is that we expect to use this directly in a lot of low-level types for debugging?

mikea: I'm going to need it first in `refcount.h` but it doesn't look right for this kind utility. `memory.h` is better but still not ideal? I was thinking `thread.h` but that is too heavy and there will probably be mess if I include it from `refcount.h`. So `common.h` seems the most logical place for such thing unless you want me to create new header.

PR #2266, opened 2025-03-25
add clang 19 to CI

mikea:
unfortunately clang 20 is still not available


fhanau:
Positively surprised that 19 is available, I thought Ubuntu didn't backport LLVM releases to LTS releases (Debian is doing better in that regard). Per longstanding practice, the 5 latest clang/LLVM versions should be supported, so we can drop 14 here.

fhanau (approved):

PR #2268, opened 2025-03-26
Allow calling Reader::asGeneric() on const readers

MasterLogick:
Pretty much the title. In my case I want to change generic's parameter type, and reader is passed as const reference. Maybe this function should be also marked as `const` in Builders and Clients.  
```
struct Msg(T) {
    val @0 :T;
    metadata @1 :Text;
}

struct SimpleMessage {
   val @1 :UInt32;
}
```

```c++
capnp::FlatArrayMessageReader reader{...};
processSimpleMessage(reader.getRoot<Msg<SimpleMessage>>());

void processSimpleMessage(const Msg<SimpleMessage>::Reader &message) {
    debugDump(message.asGeneric<capnp::AnyStruct>()); // <-- now it fails because Reader::asGeneric is not const
    // ...
}

void debugDump(const Msg<capnp::AnyStruct>::Reader &message) {
    // dump message
}
```

### DIFF
--- a/c++/src/capnp/compiler/capnpc-c++.c++
+++ b/c++/src/capnp/compiler/capnpc-c++.c++
@@ -1834,7 +1834,7 @@ private:
         case AsGenericRole::READER:
           self = kj::strTree(
               "  ", templateContext.decl(true, "2"),
-              "  typename ", returnType(), "::Reader ", kj::mv(asGeneric), " {\n"
+              "  typename ", returnType(), "::Reader ", kj::mv(asGeneric), " const {\n"
               "    return typename ", returnType(), "::Reader(_reader);\n"
               "  }\n"
               "\n");


kentonv:
Yep, that looks like an oversight.

Do you want this included in the next 1.x release? If so please re-target the PR at the `master` branch (which is the 1.x release branch). The `v2` branch is our unstable development branch that most people shouldn't be using directly.

MasterLogick:
Done. Should I add `const` qualifier for other generic roles?

kentonv:
Hmm, for some reason CI isn't offering to run on this PR... I'd prefer that it run before merging so I'm going to have to figure that out.

> Done. Should I add const qualifier for other generic roles?

Essentially all methods of `Reader` ought to be const, so if you've noticed others that aren't I'm happy to accept PRs fixing them.

kentonv:
@MasterLogick Can you amend your patch with a trivial change (like add a character to the commit description) and force-push it? Hoping that makes the CI pick it up.

MasterLogick:
> Essentially all methods of `Reader` ought to be const, so if you've noticed others that aren't I'm happy to accept PRs fixing them.

No. I meant to add const qualifiers for `Builder::asGeneric` and `Pipeline::asGeneric`.

kentonv:
> No. I meant to add const qualifiers for Builder::asGeneric and Pipeline::asGeneric.

Ah. No, those should not be const, because Builders and Pipelines fundamentally are not const. If Builder::asGeneric() were marked const, then you could convert a const Builder to a non-const Builder by calling asGeneric(), but that shouldn't be allowed.

Anyway CI is running now so I'll merge this when it's done. Thanks!

PR #2271, opened 2025-03-28
Support building with `KJ_STD_COMPAT` and `bazel`

### LINKED ISSUE #1565: Cannot use capnp bazel library with KJ_STD_COMPAT
`list.h` [conditionally includes `compat/std-iterator.h`](https://github.com/capnproto/capnproto/blob/3b2e368cecc4b1419b40c5970d74a7a342224fac/c%2B%2B/src/capnp/list.h#L549), but this is not included in the [bazel library](https://github.com/capnproto/capnproto/blob/3b2e368cecc4b1419b40c5970d74a7a342224fac/c%2B%2B/src/capnp/BUILD.bazel#L22-L49). This gives an error when `KJ_STD_COMPAT` is defined, since the file is not in the sandbox.

lukasoyen:
fixes #1565

### DIFF
--- a/c++/src/capnp/BUILD.bazel
+++ b/c++/src/capnp/BUILD.bazel
@@ -47,6 +47,11 @@ cc_library(
         "serialize-packed.h",
         "serialize-text.h",
         "stream.capnp.h",
+        # This is just to make it avaible if `KJ_STD_COMPAT` is defined.
+        # This is an unfortunate hack, but if we would make `std-iterator.h`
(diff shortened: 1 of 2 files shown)


fhanau (approved):

fhanau:
> @fhanau does this seem reasonable?

Yes, I think this addresses that use case pretty cleanly.

kentonv:
Thanks @lukasoyen!

PR #2272, opened 2025-04-01
Update minimum CMake version to 3.10.

teo-tsirpanis:
Fixes build errors when using CMake 4.0. Technically, just 3.5 would be sufficient to avoid the errors, but targeting 3.10 avoids deprecation warnings as well.

### DIFF
--- a/c++/src/kj/CMakeLists.txt
+++ b/c++/src/kj/CMakeLists.txt
@@ -78,8 +78,7 @@ set(kj-std_headers
 )
 add_library(kj ${kj_sources})
 add_library(CapnProto::kj ALIAS kj)
-# TODO(cleanup): Use cxx_std_14 once it's safe to require cmake 3.8.
-target_compile_features(kj PUBLIC cxx_generic_lambdas)
+target_compile_features(kj PUBLIC cxx_std_14)
 
 if(UNIX AND NOT ANDROID)
   target_link_libraries(kj PUBLIC pthread)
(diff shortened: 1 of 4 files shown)


c++/src/kj/CMakeLists.txt
```diff
@@ -78,8 +78,7 @@ set(kj-std_headers
 )
 add_library(kj ${kj_sources})
 add_library(CapnProto::kj ALIAS kj)
-# TODO(cleanup): Use cxx_std_14 once it's safe to require cmake 3.8.
-target_compile_features(kj PUBLIC cxx_generic_lambdas)
+target_compile_features(kj PUBLIC cxx_std_14)
```
kentonv, line 81: We actually require C++17 these days (on the master branch; even newer on v2).

kentonv:
@harrishancock does this seem like the right thing to do?

(I don't know cmake.)

c++/src/kj/CMakeLists.txt
teo-tsirpanis, line 81:
```suggestion
target_compile_features(kj PUBLIC cxx_std_17)
```

teo-tsirpanis: Done.

harrishancock (approved):

harrishancock (approved):

harrishancock:
Yep, LGTM!

PR #2279, opened 2025-04-15
Ask Claude Code to implement kj::PromisesUnordered

harrishancock:
I've wanted a PromisesUnordered primitive for a while now, and I need to learn Claude Code, so I had Claude roll it up for me.

WARNING: While I have read all of the code in this PR, I would not say I have reviewed it in depth, and decided I am comfortable with all of the details. Nevertheless, I wanted to share what Claude was able to come up with in about 2 days of coaxing.


kentonv:
Hmm, I'm surprised by the amount of code here.

Off the top of my head, I'd expect an implementation would be a TaskSet plus a ProducerConsumerQueue. Each task is wrapped to push its result to the ProducerConsumerQueue, then placed in the TaskSet.

PR #2285, opened 2025-04-23
Update 1.x release branch CI to Ubuntu 22.04.

kentonv:


kentonv:
uuuuuuuuugghhhh this is going to be a fight, isn't it?

fhanau:
I think this can be fixed:
- For the clang test I assume clang more recent than 11 is required. This seems acceptable with clang-20 being out.
- For the Bazel build `setup-bazelisk` causes trouble, but it should not be needed since bazelisk is pre-installed on the image.

kentonv:
> For the clang test I assume clang more recent than 11 is required. This seems acceptable with clang-20 being out.

This is the 1.x release branch. The supported compilers ideally shouldn't change between 1.x point releases as it may be a surprise to people who are stuck on old compilers for various reasons. So I'm trying to use the oldest compiler versions that are readily available.

Clang 11 is available on Ubuntu 22.04 but seems to have some sort of misconfiguration where it can't find the C++ libs. Trying clang 12 now.

> For the Bazel build setup-bazelisk causes trouble, but it should not be needed since bazelisk is pre-installed on the image.

Thanks, that seems to have helped.

PR #2286, opened 2025-04-23
Reduce promise continuation templating for Promise<void> where possible

fhanau:
if constexpr allows us to easily define a special case for void promise continuations where we don't need to include the continuation function in the promise template, which reduces the number of duplications of TransformPromiseNode variants. This significantly improves code size and likely also compile times for code using promises.

Rationale: Recursive templating for promise continuations is a big contributor to file size bloat (see #2079). By reducing such templating for Promise<void> -> Promise<void> continuations, this reduces opt binary sizes by **6.7%** for http-test (which uses promises a lot); compile times should improve too. ~~I expect smaller but still significant gains for downstream projects~~ Looks like downstream gains are not even close, the downstream release binary should shrink by ~0.3%.

Learned more C++17/C++20 techniques than I ever wanted to while working on this. Still need to test this with downstream. Also do we have clang-format set up for capnproto in some capacity? It works so well in workerd that I'm no longer used to manually formatting code 😅

### DIFF
--- a/c++/src/kj/async-inl.h
+++ b/c++/src/kj/async-inl.h
@@ -26,6 +26,8 @@
 
 #pragma once
 
+#include <type_traits>  // For std::is_same_v
+
 #ifndef KJ_ASYNC_H_INCLUDED
 #error "Do not include this directly; include kj/async.h."
 #include "async.h"  // help IDE parse this file
@@ -35,6 +37,7 @@
 #include <intrin.h>  // _ReturnAddress
 #endif
 
+#include <kj/function.h>
 #include <kj/list.h>
 
 KJ_BEGIN_HEADER
@@ -704,6 +707,8 @@ private:
   friend class TransformPromiseNode;
   template <typename, typename>
   friend class SimpleTransformPromiseNode;
+  template <typename>
+  friend class VoidTransformPromiseNode;
 };
 
 template <typename _DepT, typename Func, typename ErrorFunc>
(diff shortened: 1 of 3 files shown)


c++/src/kj/async-inl.h
```diff
@@ -26,6 +26,8 @@
 
 #pragma once
 
+#include <type_traits>  // For std::is_same_v
```
kentonv, line 29: Use kj::isSameType instead.

c++/src/kj/async.h
```diff
@@ -202,6 +202,11 @@ class [[nodiscard]] Promise: protected _::PromiseBase {
   template <typename Func>
   PromiseForResult<Func, T> then(Func&& func) KJ_WARN_UNUSED_RESULT;
 
+  // Specialized variants for Promise<void> continuations. This results in substantial code size
+  // improvements.
+  Promise<void> thenV(kj::Function<void()>&& func) requires std::is_void_v<T> KJ_WARN_UNUSED_RESULT;
```
kentonv, line 207: kj::Function introduces significant overhead -- an extra heap allocation on construction and virtual function dispatch when called. Have you measured the performance impact?

fhanau: That is unfortunate. I agree that allocation overhead is not desirable here and assumed kj::Function was just a thin wrapper around C-style function pointers. C++ doesn't allow using function pointers for non-capturing lambdas and almost all of the impact comes from capturing lambdas, so I guess eliminating Func as a parameter to SimpleTransformPromiseNode just won't be feasible here.

kentonv: Yeah function pointers fundamentally can't work with captures, because where do the captures go? And the captures are also why we need either templates or dynamic allocation + virtual dispatch.

PR #2288, opened 2025-04-24
[nfc] if constexpr and send().ignoreResult() optimizations

fhanau:
This simplifies a bunch of code and reduces the debug binary size by 0.9% downstream if the sendIgnore() trick is used where applicable. Let's hope this optimization won't turn out to have pesky side effects 😅

### DIFF
--- a/c++/src/capnp/capability.h
+++ b/c++/src/capnp/capability.h
@@ -121,6 +121,9 @@ public:
   RemotePromise<Results> send() KJ_WARN_UNUSED_RESULT;
   // Send the call and return a promise for the results.
 
+  kj::Promise<void> sendIgnoringResult();
+  // Equivalent to send().ignoreResult(), but is a bit more efficient.
+
   typename Results::Pipeline sendForPipeline();
   // Send the call in pipeline-only mode. The returned object can be used to make pipelined calls,
   // but there is no way to wait for the completion of the original call. This allows some
@@ -1090,6 +1093,13 @@ RemotePromise<Results> Request<Params, Results>::send() {
       kj::mv(kj::implicitCast<AnyPointer::Pipeline&>(typelessPromise)));
 
   return RemotePromise<Results>(kj::mv(typedPromise), kj::mv(typedPipeline));
+}
+
+template <typename Params, typename Results>
+kj::Promise<void> Request<Params, Results>::sendIgnoringResult() {
+  auto typelessPromise = hook->send();
+  hook = nullptr;  // prevent reuse
+  return kj::mv(typelessPromise).ignoreResult();
 }
 
 template <typename Params, typename Results>
--- a/c++/src/capnp/capability-test.c++
+++ b/c++/src/capnp/capability-test.c++
@@ -1430,7 +1430,7 @@ KJ_TEST("RevocableServer") {
 
   KJ_EXPECT_THROW_RECOVERABLE_MESSAGE(
       "capability was revoked",
-      revocable.getClient().waitForeverRequest().send().ignoreResult().wait(waitScope));
+      revocable.getClient().waitForeverRequest().sendIgnoringResult().wait(waitScope));
 
   KJ_EXPECT(!revocable.isInUse());
 }
--- a/c++/src/capnp/compat/byte-stream.c++
+++ b/c++/src/capnp/compat/byte-stream.c++
@@ -141,7 +141,7 @@ public:
     KJ_SWITCH_ONEOF(state) {
       KJ_CASE_ONEOF(redirected, Redirected) {
         // Ugh I guess we need to send a real end() request here.
-        return redirected.replacement.endRequest(MessageSize {2, 0}).send().ignoreResult();
+        return redirected.replacement.endRequest(MessageSize {2, 0}).sendIgnoringResult();
       }
       KJ_CASE_ONEOF(e, Ended) {
         // whatever
@@ -154,7 +154,7 @@ public:
       KJ_CASE_ONEOF(streaming, Streaming) {
         auto req = streaming.callback.endedRequest(MessageSize {4, 0});
         req.setByteCount(completed);
-        auto promise = req.send().ignoreResult();
+        auto promise = req.sendIgnoringResult();
         streaming.parent.returnStream(completed);
         state = Ended();
         return promise;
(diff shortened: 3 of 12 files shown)


c++/src/capnp/capability.h
```diff
@@ -1092,6 +1096,17 @@ RemotePromise<Results> Request<Params, Results>::send() {
   return RemotePromise<Results>(kj::mv(typedPromise), kj::mv(typedPipeline));
 }
 
+template <typename Params, typename Results>
+kj::Promise<void> Request<Params, Results>::sendIgnore() {
+  auto typelessPromise = hook->send();
+  hook = nullptr;  // prevent reuse
+
+  // Wrap the typeless pipeline in a pipeline wrapper.
+  typename AnyPointer::Pipeline typedPipeline(
```
fhanau, line 1105: - I'm unfamiliar with the underlying code – lmk if there's correctness issues here or more simplification potential.
- Open to a different name for the function, not sure if sendIgnore() could lead developers to end up ignoring promises when they shouldn't?

kentonv: `typelessPromise` is already of type `RemotePromise<AnyPointer>`, so it looks like the code here is redundantly taking apart and reconstructing a `RemotePromise<AnyPointer>`? You could just return `typelessPromise.ignoreResult()`.

That said, it looks to me like this whole strategy would not save anything in release builds -- the compiler should be able to optimize away everything and produce the same code as `send().ignoreResult()`. I guess this only benefits debug builds? I don't really feel like we should be optimizing for debug build size, that feels like it leads to perverse incentives (making code less elegant to work around lack of inlining).

On the other hand, there could be an opportunity here for a deeper optimization, if we added a `sendIgnoringResult()` method to `RequestHook` which returns `Promise<void>`. That would actually save some allocations (even in release builds) by avoiding the need to construct a `PipelineHook`, and maybe even avoiding the creation of a `ResponseHook`. But that's a deeper change.

It could also be worth considering _automatically_ returning `Promise<void>` for any RPC methods that have empty results, since there's really no benefit to returning a `capnp::Response` that is empty anyway. That would be a breaking change, though -- we'd have to update all callers.

A `sendNoPipeline()` method could also be worthwhile, for cases where the caller only wants a `Promise<capnp::Response<T>>` and doesn't need pipelining. We actually already have the opposite, `sendForPipeline()`, which _only_ returns a pipeline with no Promise. (If only we could automatically detect when the caller only uses the promise, or only uses the pipeline... this should be statically checkable but I don't think there's a way to make the C++ compiler check for us, we'd need to write a custom linter.)

c++/src/capnp/capability.h
```diff
@@ -121,6 +121,10 @@ class Request: public Params::Builder {
   RemotePromise<Results> send() KJ_WARN_UNUSED_RESULT;
   // Send the call and return a promise for the results.
 
+  kj::Promise<void> sendIgnore();
```
kentonv, line 124: I would name this `sendIgnoringResult()`.

c++/src/kj/async-io.h
```diff
@@ -59,15 +59,9 @@ class AsyncInputStream: private AsyncObject {
...
-  Promise<size_t> read(ArrayPtr<byte> buffer, size_t minBytes) { 
-    return read(buffer.begin(), minBytes, buffer.size()); 
-  }
-  Promise<size_t> tryRead(ArrayPtr<byte> buffer, size_t minBytes) {
-    return tryRead(buffer.begin(), minBytes, buffer.size()); 
-  }
-  Promise<void> read(ArrayPtr<byte> buffer) {
-    return read(buffer, buffer.size()).ignoreResult();
-  }
+  Promise<size_t> read(ArrayPtr<byte> buffer, size_t minBytes);
```
kentonv, line 62: Why de-inline these? The inner code of these methods is about the same size as the function call itself so this seems like a strict loss, adding a layer of function call overhead for no improvement in code size in release mode.

fhanau: I was concerned about the ignoreResult() call, but since this function and class is non-templated it's acceptable to have it in a header. Didn't notice much difference in code size either way, so I dropped this part in the new push.

c++/src/capnp/capability.h
fhanau: 👍

c++/src/capnp/capability.h
fhanau: - Simplified the function as suggested and renamed it.
- The optimization here is not about inlining the ignoreResult() call, but about getting rid of the promise type conversion from Anypointer to Response – this requires a then() continuation for each instance of the Request<Params, Results> class where send() is called. We can avoid that overhead in the sendIgnoringResult() case since we don't care about getting the type right, we cast to Promise<void> anyway.
We still benefit from this in release mode – the internal downstream binary shrinks by 0.24% from this change alone when sendIgnoringResult() is applied consistently (including downstream). Not much by itself, but worth adding a small function. Presumably this means the then() call is not optimized out, so we also avoid allocating memory for that.

kentonv: Oh, I totally missed the `.then()` for the type conversion. OK, I agree now this is a win. But the things I suggested could be even bigger (for a future PR).

c++/src/capnp/capability.h
```diff
@@ -121,6 +121,10 @@ class Request: public Params::Builder {
   RemotePromise<Results> send() KJ_WARN_UNUSED_RESULT;
   // Send the call and return a promise for the results.
 
+  kj::Promise<void> sendIgnoringResult();
+// equivalent of calling send().ignoreResult(), defining a consolidated function for this common
+// pattern avoids the overhead of an additional promise continuation.
```
kentonv, line 126:
```suggestion
  // Equivalent of calling send().ignoreResult(), defining a consolidated function for this common
  // pattern avoids the overhead of an additional promise continuation.
```

c++/src/kj/async-io.h
kentonv: Oh sorry, you're right that the third function here, which has `.ignoreResult()`, is significantly more complicated than a simple function call, and so might not be appropriate to inline. But this function is rarely used anyway.

The other two functions, though, are just trivial forwarding functions so really should stay inline.

Anyway, let's just leave them alone (as you have now).

kentonv (approved):
Looks good (after comment formatting fix above).

PR #2290, opened 2025-04-29
[bazel] cidr.h uses inet_ntop from Ws2_32.lib

mikea:
I think this is the root cause of win32 linkage problems. kj-async also requires linkstatic


fhanau (approved):

kentonv:
I think this suggests `cidr.{h,c++]` should be in `libkj-async` instead of `libkj`.

PR #2291, opened 2025-04-30
Add static ExplicitEndOutputStream::wrap method

joshthoward:
`ExplicitEndOutputStream::wrap` is a helper method for wrapping a "regular" `AsyncOutputStream` with an `ExplicitEndOutputStream` allowing an EOF to be signaled without destroying the object.


c++/src/capnp/compat/byte-stream.h
```diff
@@ -50,6 +50,12 @@ class ExplicitEndOutputStream: public kj::AsyncOutputStream {
   //   is built with RTTI enabled, then its `end()` method will be used when appropriate.
 public:
   virtual kj::Promise<void> end() = 0;
+
+  // Wraps a regular stream with an `end()` method that drops the inner stream.
```
mikea, line 54: nit: the comment goes after the function.

c++/src/capnp/compat/byte-stream.c++
```diff
@@ -23,8 +23,62 @@
...
+  }
+
+  private:
+  struct State {
+    kj::Own<kj::AsyncOutputStream> inner;
+    kj::Function<void()> uncleanEnd;
+  };
+  kj::Maybe<State> state;
+
+  kj::AsyncOutputStream& getInner() {
```
mikea, line 68: nit: function can be const?

mikea (approved):

c++/src/capnp/compat/byte-stream.c++
joshthoward: The usage is `getInner().whenWriteDisconnected()` and AsyncInputStream::whenWriteDisconnected cannot be const.

kentonv:
This PR confused me for a while because I could have sworn I wrote this code.

Turns out I did, this is actually moved from the SRS codebase.

Issue #2261, opened 2025-03-21
kj is not tolerant of monotonic clock going backwards on Linux


lf-:
This unfortunately happens sometimes: https://git.lix.systems/lix-project/lix/issues/754#issuecomment-9727

```
vm-test-run-remoteBuilds_remote_2_18> client # [   35.994285] nix-build[909]: Exception: kj::ExceptionImpl: kj/timer.c++:120: failed: expected newTime >= time [35.994285455s >= 35.994442458s]; can't advance backwards in time
vm-test-run-remoteBuilds_remote_2_18> client # stack: 7f5f1c380930 7f5f1c329596 7f5f1c32931c 7f5f1c3145f2 7f5f1c314d66 561861966bc1 561861abe172 561861ab7532 561861ac63b7 561861a2e379 561861a32adc 7f5f1c3e28ef 561861a325aa 7f5f1bcaa27d 7f5f1bcaa338 561861962524
vm-test-run-remoteBuilds_remote_2_18> client # [   35.994285] nix-build[909]: Stack trace:
vm-test-run-remoteBuilds_remote_2_18> client # [   35.994285] nix-build[909]:  0# nix::getStackTrace[abi:cxx11]() in /nix/store/dy807d7l4s7pzmmxazzq6wg7p02fvy7a-lix-2.93.0-devpre20250321_5339ffb/lib/liblixutil.so
vm-test-run-remoteBuilds_remote_2_18> client #  1# 0x00007F5F1C3D3537 in /nix/store/dy807d7l4s7pzmmxazzq6wg7p02fvy7a-lix-2.93.0-devpre20250321_5339ffb/lib/liblixmain.so
vm-test-run-remoteBuilds_remote_2_18> client #  2# 0x00007F5F1BF5A20A in /nix/store/ybjcla5bhj8g1y84998pn4a2drfxybkv-gcc-13.3.0-lib/lib/libstdc++.so.6
vm-test-run-remoteBuilds_remote_2_18> client #  3# 0x00007F5F1BF5A275 in /nix/store/ybjcla5bhj8g1y84998pn4a2drfxybkv-gcc-13.3.0-lib/lib/libstdc++.so.6
vm-test-run-remoteBuilds_remote_2_18> client #  4# 0x00007F5F1BF5A4C7 in /nix/store/ybjcla5bhj8g1y84998pn4a2drfxybkv-gcc-13.3.0-lib/lib/libstdc++.so.6
[...]

PR #2296, opened 2025-05-05
Remove check for monotonic time

fabiorossetto:
The check was disabled on Mac, but in #2261 it was reported also on Linux. At this point, it makes more sense to remove the KJ_REQUIRE for monotonicity alltogether.


kentonv:
Thanks!

PR #2297, opened 2025-05-08
Add Timer::repeatedly utility

vaci:
A repeating timer


kentonv:
Hmm, it seems like this is pretty easy to write this in application code like:

```c++
for (;;) {
  co_await timer.afterDelay(t);
  doSomething();
}
```

Is it really worth extending the Timer API?

vaci:
Oh, that is more elegant... I don't have coroutines available yet, so I will await them :)

Issue #538, opened 2017-08-19
Backwards compatibility verification tool


Vorlent:
When I modify a schema file I want to see if it's backwards compatible with the previous version as a part of the build process. A semantic versioning annotation could be used to distinguish between non breaking and intentionally breaking changes.

I have version 1.0.0 of my schema file and I upgrade to 1.1.0 which means the tool should check if it's compatible. If I accidentially make a mistake the tool should tell me that the file is no longer compatible. If I have version 1.1.0 and want to make a breaking change I will set the version to 2.0.0 and the tool will skip the verification step.

Does such a tool already exist and if yes can it be easily integrated into the build process?

deansher:
I added a $100 bounty to this on [BountySource](https://www.bountysource.com/issues/48433936-backwards-compatibility-verification-tool). After doing that, I realized this functionality is nearly present in SchemaLoader -- see [this discussion group thread](https://groups.google.com/forum/#!searchin/capnproto/compatibility%7Csort:date/capnproto/dmnD1zWo7Ug/F-6dduXyAwAJ). I also see Kenton's caution in that thread: "Note that just because two schemas are "compatible" as checked by CompatibilityChecker does not prove that they are compatible at the application level." 

All this said, I'd be delighted if someone claimed my bounty by adding a public API call exposing the "backward compatibility check" functionality in a nice way.

kentonv:
How about something like:

    capnp lint-update [<option>...] <old> <new>

> Checks for compatibility violations when upgrading the file `<old>` to `<new>`. Either `<old>` or `<new>` (but not both) may be `-` to read from standard input (the input will be treated as if it had the same filename as the other file, e.g. for resolving imports). If any incompatibilities are discovered, the command writes a description to standard error and exits with a non-zero status. If no incompatibilities are detected, the tool outputs nothing and exits with status code zero (success).
>
> This tool is useful to use in e.g. a git commit hook, to catch the introduction of backwards-incompatible changes. Keep in mind, though, that no tool can catch all possible incompatibilities. For example, if the application changes how it interprets a field without changing the way it is declared, obviously the tool cannot detect this.

kentonv:
(Disclaimer: Not promising to implement this soon, but wanted to check if this suits both of your needs.)

deansher:
This sounds great, Kenton! My use case is ideally library API rather than
command-line, but I imagine the normal pattern would be to implement both?

And by the way, I think Cap’N Proto rocks!

[...]

PR #2299, opened 2025-05-13
Backwards compatibility check (lint-update)

sbaudlr:
This change implements a lint-update command as requested and described in https://github.com/capnproto/capnproto/issues/538

We have had this running in our CI workflows for a little over a month now. The changes we have been using have been based on v1 rather than v2, but the changes I am contributing here are a direct cherry-pick of what we have running; I have verified that the changes appear to behave the same. I'm happy to be pointed to how to write tests for this.

Please let me know if a backport to v1 would be desired! I can follow up with a PR for that after this is merged.

### DIFF
--- a/c++/src/capnp/compiler/capnp.c++
+++ b/c++/src/capnp/compiler/capnp.c++
@@ -52,6 +52,7 @@
 #include <capnp/serialize.h>
 #include <capnp/serialize-packed.h>
 #include <capnp/serialize-text.h>
+#include <capnp/schema-parser.h>
 #include <capnp/compat/json.h>
 #include <errno.h>
 #include <stdlib.h>
@@ -107,6 +108,8 @@ public:
                             "DEPRECATED (use `convert`)")
              .addSubCommand("encode", KJ_BIND_METHOD(*this, getEncodeMain),
                             "DEPRECATED (use `convert`)")
+             .addSubCommand("lint-update", KJ_BIND_METHOD(*this, getLintUpdateMain),
+                            "Check that a schema is backwards compatible with another schema")
              .addSubCommand("eval", KJ_BIND_METHOD(*this, getEvalMain),
                             "Evaluate a const from a schema file.");
       addGlobalOptions(builder);
@@ -244,6 +247,24 @@ public:
     return builder.build();
   }
 
+  // Handler for the `lint-update` command.
+  kj::MainFunc getLintUpdateMain() {
+    kj::MainBuilder builder(context, VERSION_STRING,
+          "Compares <old-schema-file> and <new-schema-file> to check whether <new-schema-file> "
+          "is backwards comparible with <old-schema-file>. If so then this command will exit with "
+          "code 0, otherwise a non-zero exit code will be returned. Note that only the data types "
+          "used within these files are checked, so if the schemas import data types from schemas "
+          "that export other data types that are not used in either <new-schema-file> or <old-schema-file> "
+          "then these additional data types will not be evaluated. As such, if it is desired to check the "
+          "whole contents of a set of schemas within a directory, this command should be called for all of the old "
+          "and new file pairs from said directory to ensure full coverage.");
+    addGlobalOptions(builder);
+    builder.expectArg("<old-schema-file>", KJ_BIND_METHOD(*this, addSchema))
+          .expectArg("<new-schema-file>", KJ_BIND_METHOD(*this, addSchema))
+          .callAfterParsing(KJ_BIND_METHOD(*this, confirmSchemaIsCompatible));
+    return builder.build();
+  }
+
   kj::MainFunc getEvalMain() {
     // Only parse the schemas we actually need for decoding.
     compileEagerness = Compiler::NODE;
@@ -373,6 +394,46 @@ public:
       sourceFiles.add(SourceFile { compiled.getId(), compiled, module.getSourceName(), &module });
     } else {
       return "no such file";
+    }
+
+    return true;
+  }
+
+  // Adds a schema file to our `schemaLoader`.
+  // Will resolve all references within the schema.
+  // This is primarily useful for cases such as checking backwards compatibility
+  // where the additional safety checks provided by the compiler are undesirable
+  // (e.g. we want to load two _different_ schema files that happen to have the same
+  // Id in order to compare two versions of the schema.)
+  //
+  // If a schema is loaded that is incompatible with an already loaded schema, this
+  // function will fail, which is itself enough for a surface-level check of backwards
+  // compatibility, where the most recently loaded schema is not compatible with all
+  // previously loaded schemas.
+  //
+  // Note that only types in the top-level schema and types relied on by the top-level
+  // schema will be validated. For example, if we add `a.capnp` which depends on `Blorp`
+  // form `b.capnp` then the compatibility of `Blorp` will be checked when a newer version
+  // of `a.capnp` is loaded (and imports a newer version of `b.capnp`). However, if
+  // `b.capnp` also declares `Blarp` but `a.capnp` does not use `Blarp` then `Blarp`
+  // will not be evaluated for compatibility.
+  kj::MainBuilder::Validity addSchema(kj::StringPtr file) {
+    // schemaLoader is lazy-loaded as it's not needed for all commands.
+    // We _could_ use the compiler here and reach into it for its schema loader
+    // but we are then dependent on the implementation of the compiler when
+    // we don't need to be.
+    if (!schemaLoaderConstructed) {
+      schemaLoader = schemaLoaderSpace.construct();
+      schemaLoaderConstructed = true;
+    }
+
+    // Now we start to actually load the schema files.
+    capnp::SchemaParser parser;
+    DirPathPair dirPathPair = interpretSourceFile(file);
+    parser.parseFromDirectory(dirPathPair.dir, dirPathPair.path.clone(), nullptr);
+
+    for (capnp::Schema loadedSchema: parser.getAllLoaded()) {
+      schemaLoader->load(loadedSchema.getProto());
     }
 
     return true;
@@ -1668,6 +1729,18 @@ public:
     return convert();
   }
 
+  // Should be called after validating that a set of schemas are compatible
+  // with each other. Primarily exists to satisfy the requirement of building
+  // commands for the CLI.
+  kj::MainBuilder::Validity confirmSchemaIsCompatible() {
+    // If we've made it to this point then the schemas have been
+    // successfully loaded and thus the new schema is backwards
+    // compatible with the old. If there are compatibility errors
+    // then we will not make it this far.
+    context.exitInfo("Schemas are compatible!");
+    return true;
+  }
+
   kj::MainBuilder::Validity setEvalOutputFormat(kj::StringPtr format) {
     KJ_IF_SOME(f, parseFormatName(format)) {
       convertTo = f;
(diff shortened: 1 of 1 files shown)


c++/src/capnp/compiler/capnp.c++
```diff
@@ -244,6 +247,24 @@ public:
     return builder.build();
   }
 
+  // Handler for the `lint-update` command.
+  kj::MainFunc getLintUpdateMain() {
+    kj::MainBuilder builder(context, VERSION_STRING,
+          "Compares <old-schema-file> and <new-schema-file> to check whether <new-schema-file> "
+          "is backwards comparible with <old-schema-file>. If so then this command will exit with "
```
kentonv, line 254:
```suggestion
          "is backwards compatible with <old-schema-file>. If so then this command will exit with "
```

c++/src/capnp/compiler/capnp.c++
```diff
@@ -1860,6 +1933,9 @@ private:
   kj::SpaceFor<Compiler> compilerSpace;
   bool compilerConstructed = false;
   kj::Own<Compiler> compiler;
+  kj::SpaceFor<SchemaLoader> schemaLoaderSpace;
```
kentonv, line 1936: These three members could probably be replaced with `kj::Maybe<SchemaLoader>`.

c++/src/capnp/compiler/capnp.c++
```diff
@@ -378,6 +399,46 @@ public:
...
+      schemaLoaderConstructed = true;
+    }
+
+    // Now we start to actually load the schema files.
+    capnp::SchemaParser parser;
+    DirPathPair dirPathPair = interpretSourceFile(file);
+    parser.parseFromDirectory(dirPathPair.dir, dirPathPair.path.clone(), nullptr);
+
+    for (capnp::Schema loadedSchema: parser.getAllLoaded()) {
+      schemaLoader->load(loadedSchema.getProto());
```
kentonv, line 436: I guess if there's a failure this throws an exception, which will result in a crash dump with a stack trace? Maybe we should wrap this in a try/catch and report the exception more gracefully?

c++/src/capnp/compiler/capnp.c++
```diff
@@ -378,6 +399,46 @@ public:
...
+    // but we are then dependent on the implementation of the compiler when
+    // we don't need to be.
+    if (!schemaLoaderConstructed) {
+      schemaLoader = schemaLoaderSpace.construct();
+      schemaLoaderConstructed = true;
+    }
+
+    // Now we start to actually load the schema files.
+    capnp::SchemaParser parser;
+    DirPathPair dirPathPair = interpretSourceFile(file);
```
kentonv, line 432: My original proposal suggested one of the files could be specified as `-` to read the schema from standard input, which I figured would be convenient since you don't typically have both the old and new file sitting side-by-side... they are the same file. So you'd probably be using `git` to regenerate the original file for comparison.

How does your CI handle this exactly?

c++/src/capnp/compiler/capnp.c++
```diff
@@ -378,6 +399,46 @@ public:
...
+    parser.parseFromDirectory(dirPathPair.dir, dirPathPair.path.clone(), nullptr);
```
kentonv, line 433: I'm a little surprised the parser doesn't error out when you tell it to parse two different files that contain the same type IDs. Isn't that an error?

Also the `SchemaParser` itself contains an internal `SchemaLoader`, so isn't it redundant to use a second `SchemaLoader` below?

Entirely possible that I don't remember correctly how this stuff works, I did write it over 10 years ago...

c++/src/capnp/compiler/capnp.c++
```diff
@@ -244,6 +247,24 @@ public:
     return builder.build();
   }
 
+  // Handler for the `lint-update` command.
+  kj::MainFunc getLintUpdateMain() {
+    kj::MainBuilder builder(context, VERSION_STRING,
+          "Compares <old-schema-file> and <new-schema-file> to check whether <new-schema-file> "
```
kentonv, line 253: I think as written, if you reverse the order of the files, you'll get the same results, since the SchemaLoader is designed to allow loading the schemas in either order. But in practice you really want the comparison to be directional: it is _not_ backwards-compatible to _remove_ a field, only to add one. Do we need to extend SchemaLoader somehow to enforce the directionality?

kentonv (commented):
Thanks for working on this!

c++/src/capnp/compiler/capnp.c++
sbaudlr: Addressed by https://github.com/capnproto/capnproto/pull/2299/commits/3b74ffe8a4be3d3a04c2a7173c7edc8e84982ae4

c++/src/capnp/compiler/capnp.c++
sbaudlr: I've wrapped it in a try/catch https://github.com/capnproto/capnproto/pull/2299/commits/07e72811a2f6a423e0bf058acbfdb04c23760dcb

lmk if you were thinking of some other way of handling the errors :)

c++/src/capnp/compiler/capnp.c++
sbaudlr: I've added handling for removals in https://github.com/capnproto/capnproto/pull/2299/commits/d842a522756cc4db642fb95ccf3f310a8d5cb293

I hope this is along the right lines?

c++/src/capnp/compiler/capnp.c++
sbaudlr: If my understanding is correct, `SchemaParser` contains an internal `Compiler` which gets called down the call stack of `parseFromDirectory` which _does_ throw if we provide two schemas with the same Id. `SchemaLoader` on the other hand is perfectly happy to accept schemas with duplicate Ids (and seems to be designed to do so from what I've seen). This is why a `SchemaParser` is created each time, so that the `Compiler` state is thrown away after loading each schema.

c++/src/capnp/compiler/capnp.c++
sbaudlr: I made an attempt at loading files from stdin but ran into an issue: How do we load imported schemas? We can't load them from disk because we need to assume that:
- They also could have changed.
- Given that we're being provided the schema on stdin, the file on disk is likely to not be in the "modified state".

As for how this works in our CI:
- We do a shallow clone of just our schemas to compare against.
- We then iterate over all of the schemas that were cloned and run this `lint-update` command on them.
- We reject if any of the checks returned a non-zero status code.

We use the same script to do this in CI as we use for local development, so we don't assume anything about the state of git as a developers local environment may not be in a state suitable for us to run git commands without disrupting their work.

sbaudlr:
Apologies for the bump, I want to make sure this doesn't get forgotten :)

kentonv:
This is still in my inbox, it's just tough to prioritize when I am currently not successfully getting to all the things my team needs from me. :/ But I should be under a lot less pressure after Cloudflare's Birthday week and Connect conference are done (i.e. mid-next-month)...

c++/src/capnp/schema-loader.h
```diff
@@ -94,7 +94,7 @@ class SchemaLoader {
   // Convenience method which interprets a schema::Type to produce a Type object. Implemented in
   // terms of get().
 
-  Schema load(const schema::Node::Reader& reader);
+  Schema load(const schema::Node::Reader& reader, bool replacementExpectedToBeNewer = false);
```
kentonv, line 97: Bikeshedding: Boolean arguments are sort of hard to read. How about we add a separate method, `loadNewer()`, which is equivalent to `load()` except that it fails if the schema is older than the existing one?

c++/src/capnp/schema-loader.c++
```diff
@@ -1263,7 +1331,10 @@ _::RawSchema* SchemaLoader::Impl::load(const schema::Node::Reader& reader, bool
     // Prefer to replace the existing schema if the existing schema is a placeholder.  Otherwise,
     // prefer to keep the existing schema.
     shouldReplace = checker.shouldReplace(
-        existing, validatedReader, schema->lazyInitializer != nullptr);
+        existing, validatedReader, schema->lazyInitializer != nullptr, replacementExpectedToBeNewer);
```
kentonv, line 1334: Hmm, it seems to me that there should be no need to pass a new flag down to `CompatibilityChecker`. The class already decides which schema is "newer", so all we need to do is then verify that the replacement was deemed newer or equivalent.

Why do we need any deeper changes to the checker? Is it just to improve the error message in the case of removed fields?

c++/src/capnp/compiler/capnp.c++
```diff
@@ -378,6 +399,53 @@ public:
...
+    // we don't need to be.
+    if (schemaLoader == kj::none) {
+      schemaLoader.emplace();
+    }
+
+    KJ_IF_SOME(loader, schemaLoader) {
+      // Now we start to actually load the schema files.
+      capnp::SchemaParser parser;
+      DirPathPair dirPathPair = interpretSourceFile(file);
+      parser.parseFromDirectory(dirPathPair.dir, dirPathPair.path.clone(), nullptr);
```
kentonv, line 433: Hmm, so does passing `nullptr` as the import path mean this tool doesn't work with schemas that have imports?

c++/src/capnp/compiler/capnp.c++
```diff
@@ -378,6 +399,53 @@ public:
...
+  kj::MainBuilder::Validity addSchema(kj::StringPtr file) {
+    // schemaLoader is lazy-loaded as it's not needed for all commands.
+    // We _could_ use the compiler here and reach into it for its schema loader
+    // but we are then dependent on the implementation of the compiler when
+    // we don't need to be.
+    if (schemaLoader == kj::none) {
+      schemaLoader.emplace();
+    }
+
+    KJ_IF_SOME(loader, schemaLoader) {
```
kentonv, line 429:
```suggestion
    auto& loader = KJ_ASSERT_NONNULL(schemaLoader);
```

And then remove the `else` block below.

c++/src/capnp/compiler/capnp.c++
```diff
@@ -244,6 +247,24 @@ public:
...
+          "Compares <old-schema-file> and <new-schema-file> to check whether <new-schema-file> "
+          "is backwards compatible with <old-schema-file>. If so then this command will exit with "
+          "code 0, otherwise a non-zero exit code will be returned. Note that only the data types "
+          "used within these files are checked, so if the schemas import data types from schemas "
+          "that export other data types that are not used in either <new-schema-file> or <old-schema-file> "
+          "then these additional data types will not be evaluated. As such, if it is desired to check the "
+          "whole contents of a set of schemas within a directory, this command should be called for all of the old "
+          "and new file pairs from said directory to ensure full coverage.");
+    addGlobalOptions(builder);
+    builder.expectArg("<old-schema-file>", KJ_BIND_METHOD(*this, addSchema))
```
kentonv, line 262: Kinda feel like the code would be easier to understand if you had `setOldSchema()` and `setNewSchema()`, then did all the actual logic in `confirmSchemaIsCompatible()`.

Then e.g. the `SchemaLoader` can just be declared as a local.

kentonv: Or alternatively you could allow this to take an arbitrary number of schema files and assert that they are a linear progression, though that's probably overdoing it!

kentonv (commented):
Sorry for the long delay, I finally got a chance to look at capnproto stuff...

Mostly nitpicks and clarifying questions below, but I'm worried about imports in particular. That feels important to support? Or do they actually work and I'm just not understanding how?

PR #2300, opened 2025-05-15
Reset auto reconnect clients

jp4a50:
Add support for "resetting" auto-reconnect clients.
    
The motivating use case for this is to be able to free memory-hungry
buffers associated with capnp connections while they are not in use
without having to replace the Capability::Client itself with a new
instance.


c++/src/capnp/capability.h
```diff
@@ -867,12 +867,6 @@ class ClientHook {
   const void* brand;
 };
 
-class RevocableClientHook: public ClientHook {
```
jp4a50, line 870: Unrelated drive-by.

c++/src/capnp/reconnect.h
```diff
@@ -28,51 +28,77 @@ CAPNP_BEGIN_HEADER
 
 namespace capnp {
 
+class AutoReconnectController : public kj::Refcounted {
```
kentonv, line 31: I think we can keep the inheriting of `Refcounted` on `ReconnectHook` only. This "controller" object can be presented publicly as not being refcounted.

c++/src/capnp/reconnect.h
```diff
@@ -28,51 +28,77 @@ CAPNP_BEGIN_HEADER
...
+
+template <typename T>
+requires std::is_convertible_v<T, Capability::Client>
+struct AutoReconnectClient {
+  T cap;
+  kj::Rc<AutoReconnectController> controller;
+};
+
 template <typename ConnectFunc>
 auto autoReconnect(ConnectFunc&& connect);
```
kentonv, line 46: Perhaps we should not change the signature of the existing functions, but just add a new function like `lazyAutoReconnectWithController`?

(Do we even need a non-lazy version of the new function? It seems like the controller is not particularly useful in the non-lazy case.)

jp4a50: > Do we even need a non-lazy version of the new function? It seems like the controller is not particularly useful in the non-lazy case.

There's not any immediate use case, no. I think it's plausible there'd be one but I'm happy to break it out into a separate function.

- [ ] Do it.

c++/src/capnp/reconnect.h
jp4a50: > This "controller" object can be presented publicly as not being refcounted.

Just to clarify, are you suggesting I just move the inheritance of `kj::Refcounted` to `ReconnectHook` or are you suggesting I expose a bare reference to the `AutoReconnectController` in the result?

kentonv: Move the inheritance of `kj::Refcounted` to `ReconenctHook` (where it was before), and change `controller` below to be type `kj::Own<AutoReconnectController>` instead of `kj::Rc`.

The prevailing style in Cap'n Proto is that `kj::Refcounted` is never inherited publicly. I know this style has been broken in the workers runtime codebase and maybe it's a weird style but let's maintain it here.

c++/src/capnp/reconnect.h
```diff
@@ -28,35 +28,55 @@ CAPNP_BEGIN_HEADER
...
+public:
+  // Reset the reconnect client to its initial state.
+  virtual void reset() = 0;
+};
+
+template <typename T>
+requires std::is_convertible_v<T, Capability::Client>
+struct AutoReconnectClient {
+  T cap;
+  kj::Rc<AutoReconnectController> controller;
```
kentonv, line 41: I don't think you can actually use `kj::Rc` on something that doesn't (publicly) inherit `Refcounted`? Presumably `.addRef()` wouldn't work at least.

```suggestion
  kj::Own<AutoReconnectController> controller;
```

c++/src/capnp/reconnect.h
```diff
@@ -28,35 +28,55 @@ CAPNP_BEGIN_HEADER
...
-// invoked again each time the capability is found to be disconnected. `connect()` may return
-// any capability `Client` type.
+// The `connect()` function is initially invoked synchronously to construct the capability and then
+// invoked again each time the capability is found to be disconnected. `connect()` may return any
+// capability `Client` type.
 //
 // Example usage might look like:
 //
-//     Foo::Client foo = autoReconnect([&rpcSystem, vatId]() {
+//     auto client = autoReconnect([&rpcSystem, vatId]() {
```
kentonv, line 55: Shall we revert these comment changes and/or move the new comments to `lazyAutoReconnectWithController`?

c++/src/capnp/reconnect.h
```diff
@@ -28,35 +28,55 @@ CAPNP_BEGIN_HEADER
...
 // to avoid an infinite loop in the case that the DISCONNECTED exception actually represents a
 // permanent problem. Consider using `kj::retryOnDisconnect()` to implement this behavior.
 
 template <typename ConnectFunc>
 auto lazyAutoReconnect(ConnectFunc&& connect);
-// The same as autoReconnect, but doesn't call the provided connect function until the first
-// time the capability is used. Note that only the initial connection is lazy -- upon
-// disconnected errors this will still reconnect eagerly.
+// The same as autoReconnect, but doesn't call the provided connect function until the capability
+// is first used (this applies to both initial construction and after a reset).
```
jp4a50, line 69: - [ ] Woops. Need to remove the reference to `reset` here.

c++/src/capnp/reconnect.h
jp4a50: - [ ] Yeah I'll just revert all the changes to existing comments.

c++/src/capnp/reconnect.h
jp4a50: > kj::Own<AutoReconnectController> instead of kj::Rc

Ah, yeah, that makes sense now.

c++/src/capnp/reconnect.h
jp4a50: Yeah that's why I wasn't sure exactly what you were asking. You can still use the `kj::Rc` but, as you say, `addRef()` (for example) doesn't work.

kentonv (approved):

PR #2315, opened 2025-06-10
fix missing #else

mikea:
gives warnings otherwise. I think else is implied?

### DIFF
--- a/c++/src/kj/filesystem-disk-unix.c++
+++ b/c++/src/kj/filesystem-disk-unix.c++
@@ -1687,6 +1687,7 @@ private:
 #endif
           return result;
         }
+#else
 	KJ_FALLTHROUGH;
 #endif
       default:


kentonv:
Ugh yeah this broke my brain a bit when I reviewed it, perhaps I shouldn't have accepted it as-is.

Perhaps a cleaner solution would be to move the `#ifdef O_PATH` up a couple lines so that the whole EACCESS case is removed without it, then the fallthrough isn't needed at all.

mikea:
> Ugh yeah this broke my brain a bit when I reviewed it, perhaps I shouldn't have accepted it as-is.
> 
> Perhaps a cleaner solution would be to move the `#ifdef O_PATH` up a couple lines so that the whole EACCESS case is removed without it, then the fallthrough isn't needed at all.

not sure how would it look like? `KJ_SYSCALL_HANDLE_ERRORS` has a switch so I can't not have a `case EACCESS`?

The easiest way to eliminate fallthrough is to copy `KJ_FAIL_SYSCALL("open(dir, O_RDONLY)", error, dir);` statement imo.

PR #2316, opened 2025-06-11
Warn against use of kj::Own<T>::attach() for classes deriving from kj::Refcounted

jclee:
...since calling `.attach()` on a `kj::Refcounted` instance is probably a bug.  Just warning via `[[deprecated]]` for now, but can be cleaned up to a `static_assert()` once existing occurrences are fixed.

### DIFF
--- a/c++/src/kj/memory-test.c++
+++ b/c++/src/kj/memory-test.c++
@@ -21,6 +21,7 @@
 
 #include "kj/common.h"
 #include "kj/string.h"
+#include "kj/refcount.h"
 #include "kj/test.h"
 #include "function.h"
 #include "memory.h"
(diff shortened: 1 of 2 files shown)


kentonv:
Hmm, I suspect there are places where we do this intentionally -- put an attachment on the Own which we intend to be released when the specific Own is released, independent of the underlying refcounted object. Maybe we need an alternative name like `attachToThisReference()` or something that people can use when they know it's safe?

jclee:
> Hmm, I suspect there are places where we do this intentionally -- put an attachment on the Own which we intend to be released when the specific Own is released, independent of the underlying refcounted object. Maybe we need an alternative name like `attachToThisReference()` or something that people can use when they know it's safe?

Ah, that's a good idea... I wasn't sure if there were cases where we want an attachment to track the `kj::Own<T>` lifetime instead of the `T` lifetime -- but if there are, that seems like a good way to opt into it.

jclee:
- Added `kj::Own<T>::attachToThisReference()` for refcounted objects, and some testing
- Made `IsRefcounted` concept also cover `kj::AtomicRefcounted`

kentonv (approved):

jclee:
Heh... looks like the code that I added to memory-test to test the deprecation warnings breaks CI because the build disallows deprecation warnings -- probably a good thing.  I'm disabling just that part via `#ifdef 0` for now.

PR #2318, opened 2025-06-11
Use kj::max instead of std::max in kj/timer.c++ (was: Include <algorithm> in kj/timer.c++ to satisfy clang-20)

lewinb-corp:
I noticed that compiling `libkj` fails with clang-20.1.5 for me:

```
/home/lewinb/.conan2/p/b/capnp25c9fdddbfd6b/b/src/c++/src/kj/timer.c++:118:10: error: no member named 'max' in namespace 'std'; did you mean simply 'max'?
  118 |   time = std::max(time, newTime);
      |          ^~~~~~~~
      |          max
```

This PR fixes the error by including `<algorithm>` for `std::max()` on `master`.

### DIFF
--- a/c++/src/kj/timer.c++
+++ b/c++/src/kj/timer.c++
@@ -112,10 +112,10 @@ Maybe<uint64_t> TimerImpl::timeoutToNextEvent(TimePoint start, Duration unit, ui
 void TimerImpl::advanceTo(TimePoint newTime) {
   // It has been observed that clock_gettime 
   // may return non monotonic time, even when CLOCK_MONOTONIC is used.
-  // We use std::max to guard against this rare issue.
+  // We use kj::max to guard against this rare issue.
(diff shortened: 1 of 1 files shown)


kentonv:
This may have broken with 6cf179bd799b66cb56185dc834b54b3355163c08 -- before that the line was only compiled on Mac.

The right answer here, though, is probably to use `kj::max` instead of `std::max`. I'm a little surprised to see this using `std::max`, actually. (`kj::max` is generally better as it can deal with the argument types not being the same, not that that matters here.)

lewinb-corp:
Sounds good, I've updated the commit accordingly.

kentonv:
Oh man, I just pushed a release yesterday, forgot this was still pending. Guess I'll have to push another one.

Oh well, thanks for the fix in any case!

PR #2319, opened 2025-06-11
ArrayPtr-based AsyncInputStream::read()

mikea:


c++/src/kj/async-io.h
```diff
@@ -56,16 +56,16 @@ class AsyncInputStream: private AsyncObject {
   // Asynchronous equivalent of InputStream (from io.h).
 
 public:
-  virtual Promise<size_t> read(void* buffer, size_t minBytes, size_t maxBytes);
+  virtual Promise<size_t> read(ArrayPtr<byte> buffer, size_t minBytes);
```
kentonv: Maybe not related to this specific change but I've long thought this never should have been virtual, it should just be based on tryRead... just never got around to changing it.

c++/src/kj/async-coroutine-test.c++
```diff
@@ -444,9 +444,8 @@ Promise<void> sendData(Promise<Own<NetworkAddress>> addressPromise) {
 Promise<String> receiveDataCoroutine(Own<ConnectionReceiver> listener) {
   auto server = co_await listener->accept();
   char buffer[4]{};
-  auto n = co_await server->read(buffer, 3, 4);
-  KJ_EXPECT(3u == n);
-  co_return heapString(buffer, n);
+  co_await server->read(kj::arrayPtr(buffer).first(3).asBytes());
```
kentonv: Seems like you're changing semantics here? Previously the test would verify that no more than 3 bytes were available for reading, now it'll ignore extra bytes. (I don't remember if this is important, though.)

c++/src/kj/async-io.h
mikea: sure, can do. I'm touching this anyway.

c++/src/kj/compat/http.c++
```diff
@@ -1791,7 +1791,7 @@ private:
           maxBytes = kj::min(maxBytes, MAX_CHUNK_HEADER_SIZE);
         }
 
-        readPromise = inner.read(headerBuffer.begin() + bufferEnd, 1, maxBytes);
+        readPromise = inner.read(headerBuffer.slice(bufferEnd).first(maxBytes).asBytes(), 1);
```
mikea, line 1794: I think this is the only non-trivial change.

PR #2320, opened 2025-06-12
fix canMemcpy

mikea:
currently canMemcpy<char> is false (because of the second branch)

### DIFF
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -642,7 +642,7 @@ constexpr bool canMemcpy() {
   // Returns true if T can be copied using memcpy instead of using the copy constructor or
   // assignment operator.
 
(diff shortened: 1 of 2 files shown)


mikea:
Looks like build failures come from broken master?

c++/src/kj/common.h
```diff
@@ -642,7 +642,7 @@ constexpr bool canMemcpy() {
   // Returns true if T can be copied using memcpy instead of using the copy constructor or
   // assignment operator.
 
-  return __is_trivially_constructible(T, const T&) && __is_trivially_assignable(T, const T&);
```
kentonv, line 645: Apparently the problem is that this should have been `__is_trivially_assignable(T&, const T&)` all along. I can't believe this wasn't tested somewhere.

I think there was some reason I opted against `__it_trivially_copyable`.

PR #2321, opened 2025-06-12
Put kj::Own<T>::attach() refcounted warning behind a flag

jclee:
Put `kj::Own<T>::attach()` refcounted warning behind a `KJ_WARN_REFCOUNTED_ATTACH` flag, to allow deprecation-sensitive builds of capnproto code current using refcounted attach to pass.

### DIFF
--- a/c++/src/kj/filesystem-disk-unix.c++
+++ b/c++/src/kj/filesystem-disk-unix.c++
@@ -1687,6 +1687,7 @@ private:
 #endif
           return result;
         }
+#else
 	KJ_FALLTHROUGH;
 #endif
       default:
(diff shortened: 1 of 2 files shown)


jclee:
For context:

It looks like the two failures reported in the CI log were in rpc.c++, added in 46720cc9.

In these two cases, it looks like it's due to a false positive in the "DerivedFrom" concept that I hand-rolled to avoid pulling in additional dependencies, like `type_traits`...  Both cases are calling `.attach()` on a `kj::Own<void>`, which should be fine, but gets flagged as being derived from `kj::Refcounted` due to `static_cast<const Refcounted*>(const void*)` being valid.  I guess I either need to come up with a better implementation of the concept, or just pull in `type_traits`.

This PR puts the deprecation warning behind a preprocessor flag.

mikea (approved):

PR #2323, opened 2025-06-13
enable canMemcpy<capnp::word>

mikea:


kentonv:
Well OK, to be fair, the comment is vague about the matter.

The problem is capnp::word is used to point at data that may ultimately be accessed as a variety of types, depending on the exact schema of the message. E.g. the word could contain a uint64_t or two uint32_ts. If it contains two uint32_ts, then it would be _incorrect_ to assign it as a uint64_t as this breaks aliasing rules. But aliasing rules say you can always use byte-wise access (including memcpy).

PR #2324, opened 2025-06-13
fix memcpy warnings

mikea:


anonrig (approved):

fhanau:
The warning is new in clang-20, IIRC we weren't fully compliant with it anyway and disabled it in workerd and the downstream build. Either way, glad it doesn't appear here anymore.

PR #2325, opened 2025-06-13
replace deprecated attach() usage

mikea:


jclee:
Sorry, Mike, these are actually false positives.  I have a PR up to temporarily disable the deprecation here: https://github.com/capnproto/capnproto/pull/2321

mikea:
> Sorry, Mike, these are actually false positives.

I saw your change, but I didn't realy these are false positive: I thought they need to be fixed.

PR #2326, opened 2025-06-13
restrict asBytes and asChars to canMemcpy types

mikea:
Depends on #2323 
Downstreams do not have any problem with this.


c++/src/capnp/common-test.c++
```diff
@@ -25,13 +25,17 @@
 #include <kj/debug.h>
 #include <capnp/test.capnp.h>
 
+#include <type_traits>
```
kentonv: Not used?

kentonv (approved):

PR #2338, opened 2025-06-24
Add OHOS macro to C++ file for Cross-Compile method of OpenHarmony

SC404:
we are using capnproto in OpenHarmony, we'd like to contribute the build guide, hope it can help developers


kentonv:
Why doesn't Harmony define `__ANDROID__`? Seems like it should since it's based on AOSP?

PR #2339, opened 2025-06-24
fix compilation error when compiling with -Werror and C++20

doj:
these changes fix compilation errors when compiling with -Werror and C++20. This was seen with the following settings in the cmake configuration and the gcc 11.4.0 compiler.

```
--- a/c++/CMakeLists.txt
+++ b/c++/CMakeLists.txt
@@ -16,6 +16,9 @@ if(NOT HAS_CXX14)
   message(SEND_ERROR "Requires a C++14 compiler and standard library.")
 endif()
 
+set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_CXX_STANDARD_REQUIRED True)
+
 # these arguments are passed to all install(TARGETS) calls
 set(INSTALL_TARGETS_DEFAULT_ARGS
   EXPORT CapnProtoTargets
@@ -152,7 +155,7 @@ else()
   #   recognize as safe.
   # * sign-compare: Low S/N ratio.
   # * unused-parameter: Low S/N ratio.
-  add_compile_options(-Wall -Wextra -Wno-strict-aliasing -Wno-sign-compare -Wno-unused-parameter)
+  add_compile_options(-Wall -Wpedantic -Wextra -Werror -Wno-vla -Wno-strict-aliasing -Wno-sign-compare -Wno-unused-parameter)
 
   if(DEFINED CMAKE_CXX_EXTENSIONS AND NOT CMAKE_CXX_EXTENSIONS)
     message(SEND_ERROR "Cap'n Proto requires compiler-specific extensions (e.g., -std=gnu++14). Please leave CMAKE_CXX_EXTENSIONS undefined or ON.")
```

### DIFF
--- a/c++/src/capnp/encoding-test.c++
+++ b/c++/src/capnp/encoding-test.c++
@@ -1828,7 +1828,8 @@ TEST(Encoding, NameAnnotation) {
   EXPECT_EQ(true, root.getGoodFieldName());
   EXPECT_TRUE(root.isGoodFieldName());
 
-  root.setBar(0xff);
+  // should the following set 255 (unsigned char) or -1 (signed char)?
+  root.setBar(static_cast<signed char>(0xff));
   EXPECT_FALSE(root.isGoodFieldName());
 
   root.setAnotherGoodFieldName(test::RenamedStruct::RenamedEnum::QUX);
--- a/c++/src/kj/async-io-unix.c++
+++ b/c++/src/kj/async-io-unix.c++
@@ -199,7 +199,7 @@ public:
 
     if (n < 0) {
       // EAGAIN -- need to wait for writability and try again.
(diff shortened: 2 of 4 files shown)


kentonv:
Hi Dirk,

Thanks for contributing, but our policy on warnings is that we're only interested in fixing the ones that we've enabled in our own CI (which does run with `-Werror`). I've received a huge number of requests to enable and fix various warnings over the years, and if I accepted them all, I'd end up having to run with every possible warning enabled. Some of them are just way noisier than they are worth. So we have to draw a line.

All of Cap'n Proto's headers are explicitly marked as system headers when they are included by dependents, which should prevent the compiler from reporting any warnings in them. Please configure your build system to ignore warnings in Cap'n Proto itself.

PR #2340, opened 2025-06-27
Improve building with bazel on Windows (using cl.exe)

lukasoyen:
- build(bazel): use `includes` instead of `include_prefix`
    
    With this `bazel` does not need to create virtual include folders to
    make headers available under a specific prefix. Virtual include folders
    contain symlinks to the actual folders. This:
    - confuses IDEs and debuggers
    - confuses `cl.exe` on Windows

- build(bazel): use `/TP` on Windows
    
    To enable compiling `.c++` files as C++ with `cl.exe`

This fixes workarounds like: https://github.com/cloudflare/workerd/blob/a018eee88da5245a50c9340cf1bb07747cda6846/.bazelrc#L296-L302

### DIFF
--- a/c++/src/capnp/BUILD.bazel
+++ b/c++/src/capnp/BUILD.bazel
@@ -53,6 +53,10 @@ cc_library(
         # `//src/capnp` and `//src/capnp/compat:std-iterator``.
         "//src/capnp/compat:std-iterator.h",
     ],
+    copts = select({
+        "@platforms//os:windows": ["/TP"],  # compile .c++ files as .cpp files
+        "//conditions:default": [],
+    }),
     include_prefix = "capnp",
     visibility = ["//visibility:public"],
     deps = [
@@ -85,6 +89,10 @@ cc_library(
         "rpc-twoparty.capnp.h",
         "rpc-twoparty.h",
     ],
+    copts = select({
+        "@platforms//os:windows": ["/TP"],  # compile .c++ files as .cpp files
+        "//conditions:default": [],
+    }),
     include_prefix = "capnp",
     visibility = ["//visibility:public"],
     deps = [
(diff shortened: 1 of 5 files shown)


lukasoyen:
cc @fhanau

fhanau:
MSVC is not currently supported in the downstream projects for which we added Bazel support, we only support clang-cl for them on Windows. There are no short-term plans to change that (which would be a large effort since Bazel does not support Windows all that well in the first place).
If one wanted to add the /TP flag to improve MSVC support, that can be done much more easily by passing `--cxxopt=/TP` to Bazel, no need to add a select() statement to every target.
I often get confused by how Bazel handles includes myself, not sure if the current approach is ideal but it seems to work without using the "includes" parameter, which is discouraged by Bazel. I've never had problems during development with virtual includes being used – this seems to work well with clang-cl, clang on Unix and IDEs.

lukasoyen:
>  If one wanted to add the /TP flag to improve MSVC support, that can be done much more easily by passing --cxxopt=/TP to Bazel, no need to add a select() statement to every target.

You are right with using `--cxxopt`. But that would then apply to all code in the repository that pulls in `capnproto`. An alternative would be to use `--per_file_copt` and `--host_per_file_copt`, but that complicates the `.bazelrc`, the maintenance of that file and every consuming project would need to include those.

> using the "includes" parameter, which is discouraged by Bazel

I disagree that this is discouraged. https://bazel.build/reference/be/c-cpp#cc_library says:
> This should only be used for third-party libraries that do not conform to the Google style of writing #include statements. 

I think that only applies to google internal code, it is a internal comment that made it into the open-source version. And in fact `capnproto` does not conform to the google style. If it would, the project would use `#include "c++/src/capnp/common.h"` instead of the currently used `#include "common.h"`

PR #2342, opened 2025-07-03
Set connection state to Disconnected during idle shutdown

ObsidianMinor:
If the connection's shutdown task throws an exception during an idle shutdown for any reason, the connection gets torn down again, causing shutdown() to get called twice. This can cause assertions to be thrown in code expecting shutdown() to only be called once which leads to other failures that can cause aborts.

### DIFF
--- a/c++/src/capnp/rpc.c++
+++ b/c++/src/capnp/rpc.c++
@@ -3483,8 +3483,6 @@ private:
             co_return;
           }
 
-          Connected& state = connection.get<Connected>();
-
           // At this point, the last reference to this connection state *should* be the one in
           // the RpcSystem's map. The refcount should therefore be 1, and `isShared()`.
           if (isShared()) {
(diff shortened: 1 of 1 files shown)


c++/src/capnp/rpc.c++
```diff
@@ -3502,6 +3500,12 @@ private:
             co_return;
           }
 
+          // Make sure to mark ourselves as Disconnected so if the shutdown task fails it doesn't
+          // cause us to call shutdown() again.
+          Connected state = kj::mv(connection.get<Connected>());
+          connection.init<Disconnected>(KJ_EXCEPTION(DISCONNECTED, "Peer disconnected."));
```
kentonv: Perhaps the error message should be "RPC was disconnected due to idleness" or something?

kentonv: Or "Peer disconnected idle session."?

kentonv (approved):

PR #2343, opened 2025-07-03
add contains() method to kj::HashMap

anonrig:
Adds a simple contains method that avoids the extra `.map()` when `.find() != kj::none` is used to check if key the is included.


kentonv:
Can you add a test please?

> avoids the extra .map()

FWIW that .map() probably gets optimized away. But I agree a `contains()` helper is nice to have for readability.

c++/src/kj/map.h
```diff
@@ -77,6 +77,8 @@ class HashMap {
   kj::Maybe<Value&> find(KeyLike&& key);
   template <typename KeyLike>
   kj::Maybe<const Value&> find(KeyLike&& key) const;
+  template <typename KeyLike>
+  bool contains(KeyLike&& key) const;
```
jasnell, line 81: Should this really be inserted here before the code comment below? The code comment applies to the `find(...)` methods. Does it also apply to this new `contains(...)` method? If not, then `contains(...)` should be moved to after the comment.

c++/src/kj/map.h
```diff
@@ -177,6 +179,9 @@ class TreeMap {
   // then update(Value& existingValue, Value&& newValue) is called to modify the existing value.
   // If no function is provided, the default is to simply replace the value (but not the key).
 
+  template<typename KeyLike>
+  bool contains(KeyLike&& key) const;
```
jasnell, line 183: In the above, the `contains` is added grouped under the `find` with the same code comment, here it is added separately from the `find` methods. Should likely be consistent.

PR #2346, opened 2025-07-07
Add Favicon to Capnp Docs

Warfields:
This change adds a favicon, so that tabs on https://capnproto.org/ should have a unique red icon, making it easy to find the docs.


kentonv:
I could be convinced to add a favicon, but I think 288x288 and 30kb is overly large, and the lettering would be too small, when rendered as an icon.

I could imagine an icon that is just "C'P" or "C'n'P" on a red background. 96x96 seems like the right size, unless we want to have multiple sizes, but that's a lot more work, probably not worth it.

PR #2351, opened 2025-07-11
add shrink() to kj::String and kj::ConstString

anonrig:
Includes some formatter fixes as well.


anonrig:
I'm thinking about adding a shrink() method to kj::Array as well so we can get away from copying while shrinking. What do you think @mikea?

jasnell:
Can you include a description of the use case in the PR text?

c++/src/kj/string.h
```diff
@@ -276,6 +276,8 @@ class String {
   inline constexpr size_t size() const;
   // Result does not include NUL terminator.
 
+  inline constexpr void shrink(size_t size);
```
jasnell, line 279: Some code comments here explaining the semantics here would be good. e.g. does this cause a reallocation, etc (I know it does I'm just saying it should be included in the docs)

c++/src/kj/array.h
```diff
@@ -201,21 +201,21 @@ class Array {
   inline ArrayPtr<T> first(size_t count) KJ_LIFETIMEBOUND { return slice(0, count); }
   inline ArrayPtr<const T> first(size_t count) const KJ_LIFETIMEBOUND { return slice(0, count); }
 
-  inline ArrayPtr<const byte> asBytes() const KJ_LIFETIMEBOUND { 
+  inline ArrayPtr<const byte> asBytes() const KJ_LIFETIMEBOUND {
```
kentonv, line 204: Please put unrelated cleanups in their own commit.

c++/src/kj/string.h
```diff
@@ -796,6 +800,19 @@ inline constexpr const char* ConstString::cStr() const { return content == nullp
 inline constexpr size_t String::size() const { return content == nullptr ? 0 : content.size() - 1; }
 inline constexpr size_t ConstString::size() const { return content == nullptr ? 0 : content.size() - 1; }
 
+inline constexpr void String::shrink(size_t size) {
+  if (content != nullptr) {
```
kentonv, line 804: This seems to be skipping the IREQUIRE in the case that the string is empty, but the check still seems relevant in that case.

c++/src/kj/string.h
```diff
@@ -796,6 +800,19 @@ inline constexpr const char* ConstString::cStr() const { return content == nullp
...
+    KJ_IREQUIRE(content.size() >= size, "New size must be less than or equal to the current size");
+    content = kj::heapArray(content.first(size));
```
kentonv, line 806: This doesn't work: This loses the required NUL terminator, making the string invalid. Your test does not seem to verify the content of the string.

kentonv, line 806: Reassigning the content in-place is an unusual semantic for this class. I would expect this to return the new string instead. This would also make it more obvious that the implementation re-allocates the string.

kentonv (changes requested):
I'm not convinced this is a common enough need to warrant its own helper method. You can do `s = kj::str(s.first(size))` easily enough. Uncommon helpers can actually make code harder to read as people won't be familiar with what they do -- the meaning of `shrink()` as implemented in this PR is not all that obvious to me, especially the in-place modification semantics.

PR #2352, opened 2025-07-11
add KJ_ASSERT_SOME and KJ_REQUIRE_SOME

anonrig:
This was a recommendation by @kentonv from couple of months ago, when I recommended using KJ_REQUIRE_NONNULL for checking for `nullptr`.

### DIFF
--- a/c++/src/kj/debug.h
+++ b/c++/src/kj/debug.h
@@ -395,7 +395,7 @@ namespace kj {
 // Use "ASSERT" in place of "REQUIRE" when the problem is local to the immediate surrounding code.
 // That is, if the assert ever fails, it indicates that the immediate surrounding code is broken.
 
-#define KJ_ASSERT_SOME KJ_REQUIRE_NONNULL
+#define KJ_ASSERT_SOME KJ_ASSERT_NONNULL
 #define KJ_REQUIRE_SOME KJ_REQUIRE_NONNULL
 // SOME is a better wording since KJ_REQUIRE_NONNULL checks for kj::Maybe.
 


c++/src/kj/debug.h
```diff
@@ -395,6 +395,10 @@ namespace kj {
 // Use "ASSERT" in place of "REQUIRE" when the problem is local to the immediate surrounding code.
 // That is, if the assert ever fails, it indicates that the immediate surrounding code is broken.
 
+#define KJ_ASSERT_SOME KJ_REQUIRE_NONNULL
```
jasnell:
```suggestion
#define KJ_ASSERT_SOME KJ_ASSERT_NONNULL
```

jasnell (approved):

kentonv:
Will you be updating all the use sites in our codebases and then deprecating or removing the old macro? I don't really want to add permanent aliases as this can create confusion about what the difference is. But I'd be in favor of migrating to the new macro.

We should also make sure the new macro does not work on raw pointers. I think the old version does by mistake.

anonrig:
Yes that's my intention @kentonv

jasnell:
If that's the case, it would likely be easiest to rename the existing macros and define the current names as aliases. Then the follow on PR after the updates are made is just to remove the old aliases.

kentonv:
IMO the KJ_ASSERT_SOME macros should probably be defined to work only on `Maybe` from the start -- not on pointers.

Then we can update all the call sites of KJ_ASSERT_NONNULL that take a Maybe to use KJ_ASSERT_SOME instead.

We can then either change KJ_ASSERT_NONNULL to work _only_ on pointers, or we can get rid of the macro and convert the remaining call sites to do a separate `!= nullptr` check.

jasnell:
> IMO the KJ_ASSERT_SOME macros should probably be defined to work only on Maybe from the start -- not on pointers.

Ah yeah, that's a good point. We've had confusion about that before and there are a number of places where I still think we have the `KJ_ASSERT_NONNULL` on pointers which is just... weird.

Issue #2353, opened 2025-07-18
cmake config incompatible with FetchContent


brawer:
CapnProto seems incompatible with modern cmake (specifically, adding a subdirectory found by FetchContent). This looks like a problem with the CapnProto CMake configuration, but sadly I'm not proficient enough with CMake to fix your codebase.

For example, with the following `CMakeLists.txt` in a custom project:

```cmake
cmake_minimum_required (VERSION 3.31)

project(Foo)

include(FetchContent)

FetchContent_Declare(
    CapnProto
    URL https://github.com/capnproto/capnproto/archive/refs/tags/v1.2.0.tar.gz
    EXCLUDE_FROM_ALL
    FIND_PACKAGE_ARGS)

FetchContent_MakeAvailable(CapnProto)

if(NOT CapnProto_FOUND)
    add_subdirectory(${capnproto_SOURCE_DIR} EXCLUDE_FROM_ALL)
endif()
```

cmake produces the following series of errors and warnings:

```
$ cmake -B build -S .
-- The C compiler identification is GNU 14.2.0
-- The CXX compiler identification is Clang 20.1.8
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/clang++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Looking for C++ include initializer_list
-- Looking for C++ include initializer_list - found
-- Found OpenSSL: /usr/lib/libcrypto.so (found version "3.5.1") found components: Crypto SSL
-- Found ZLIB: /usr/lib/libz.so (found version "1.3.1")
-- Looking for makecontext in c
-- Looking for makecontext in c - not found
-- Found PkgConfig: /usr/bin/pkg-config (found version "2.4.3")
-- Checking for module 'libucontext'
--   Found libucontext, version 1.3.2
CMake Error at build/_deps/capnproto-src/c++/src/CMakeLists.txt:41 (add_custom_target):
  add_custom_target cannot create target "check" because another target with
  the same name already exists.  The existing target is a custom target
  created in source directory "/build/build/_deps/capnproto-src/c++/src".
  See documentation for policy CMP0002 for more details.


[...]

PR #2354, opened 2025-07-18
Guard against multiple cmake inclusions

brawer:
https://github.com/capnproto/capnproto/issues/2353


brawer:
> Does FetchContent use include() in its implementation or something like that?

As far as I understand, `add_subdirectory()` (which user call typically calls on the fetched content) searches the added subdirectory for cmake configs, and includes it so the exported targets become available to the user project.

> Also note this should probably be backported to master to get into a 1.x release.

Sure, no problem. Sent out https://github.com/capnproto/capnproto/pull/2355 for the backport.

PR #2356, opened 2025-07-18
Wrap shutdown() call in disconnect() to catch synchronously thrown exceptions

ObsidianMinor:
If this happens when we're inside taskFailed(), it's Bad.

### DIFF
--- a/c++/src/capnp/rpc.c++
+++ b/c++/src/capnp/rpc.c++
@@ -566,8 +566,9 @@ public:
 
     // Indicate disconnect.
     auto& dyingConnectionRef = *dyingConnection;
-    auto shutdownPromise = dyingConnection->shutdown()
-        .attach(kj::mv(dyingConnection))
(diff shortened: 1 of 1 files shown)


c++/src/capnp/rpc.c++
```diff
@@ -566,7 +566,9 @@ public:
 
     // Indicate disconnect.
     auto& dyingConnectionRef = *dyingConnection;
-    auto shutdownPromise = dyingConnection->shutdown()
+    auto shutdownPromise = kj::evalNow([&dyingConnectionRef]() {
+          return dyingConnectionRef.shutdown();
+        })
         .attach(kj::mv(dyingConnection))
```
kentonv: Nitpick...

```suggestion
    auto shutdownPromise = kj::evalNow([&dyingConnectionRef]() {
      return dyingConnectionRef.shutdown();
    }).attach(kj::mv(dyingConnection))
```

kentonv (approved):

Issue #2175, opened 2024-11-07
Expose kj::SourceLocation in timer and eval*


fabiorossetto:
I'm chasing a very hard to debug "promise destroyed itself" bug. The source location in the `Event` and in the `TaskSet` are very useful to track the source of the event causing the issue.

While investigating this I noticed some cases where the sourceLocation is hidden but, in my opinion, it should be exposed. Namely:
- `evalNow`, `evalLater` and `evalLast` should have a source location as argument
- `Timer::timeoutAt` and `Timer::timeoutAfter`

So for example, I'd change the implementation of `evalLater` to:

```c++
template <typename Func>
inline PromiseForResult<Func, void> evalLater(Func&& func, kj::SourceLocation sourceLocation = {});

template <typename Func>
inline PromiseForResult<Func, void> evalLater(Func&& func, kj::SourceLocation sourceLocation) {
  return _::yield().then(kj::fwd<Func>(func), _::PropagateException(), sourceLocation);
}
```

This way the event has as source location the line number where `evalLater` was called, instead of the line number of the `evalLater` implementation (which I don't think it's very useful).

I can do the change myself, I just wanted to make sure first before investing time in a PR

kentonv:
Hmm. I'm fine with this, but for the case of "Promise callback destroyed itself", you might be able to do better. `Event` has a method `traceEvent()` which returns a full async stack trace for the event, whereas `location` will only give you the top of said stack.

`location` was added to debug cases where an `Event` was used after being freed, at which point we can't call `traceEvent()` because the object doesn't exist anymore (technically `location` doesn't exist anymore either, but we can hope that it hasn't been overwritten yet).

But "Promise callback destroyed itself" is generated in `Event`'s destructor, while the object still exists. Technically it's _slightly_ too late because `traceEvent(TraceBuilder&)` is a virtual method, and the subclass has already been destroyed, so you'd crash with "pure virtual function called". But if you could figure out where the `Event` is being destroyed, you could potentially add a line of code immediately before that to log the event's stack trace.

fabiorossetto:
Thank you for your suggestion Kenton. I found the root cause of the issue (more or less).
In my code I often use TaskSets. In a class I use a `TaskSet::ErrorHandler` like the one in `ez-rpc.c++`, that is:

```c++
// See ez-rpc.c++:317
  void taskFailed(kj::Exception&& exception) override {
    kj::throwFatalException(kj::mv(exception));
  }
```

[...]

PR #2358, opened 2025-07-22
Wrap `taskFailed()` call in noexcept IIFE

ObsidianMinor:


c++/src/kj/async.c++
```diff
@@ -342,7 +342,10 @@ protected:
 
     // Call the error handler if there was an exception.
     KJ_IF_SOME(e, result.exception) {
-      taskSet.errorHandler.taskFailed(kj::mv(e));
+      // `taskFailed()` cannot throw an exception as doing so will cause an assertion failure later.
```
kentonv: More descriptive:

> If we throw an exception here, `self` will be dropped, which means we're cancelling ourselves, which will crash. Even if that were not the case, throwing here will unwind straight out of the event loop. Hence, `taskFailed()` callbacks really shouldn't throw. But if one does, it's better that we just crash here rather than try to unwind. So, wrap the whole thing in a noexcept IIFE.

kentonv (approved):
Related issue, for reference: #2175

PR #2361, opened 2025-07-24
Retain exception details for exceptions that cause RPC disconnection

sohpeach:
I'd like to be able to set details on the exceptions that cause the underlying connection to disconnect and later read them out, but currently the details get stripped.


kentonv:
Makes sense!

PR #2362, opened 2025-07-26
KJ: Move cidr to kj-async

HazardyKnusperkeks:
This is where it belongs.

### DIFF
--- a/c++/Makefile.am
+++ b/c++/Makefile.am
@@ -266,7 +266,6 @@ endif
 libkj_la_LIBADD = $(PTHREAD_LIBS)
 libkj_la_LDFLAGS = -release $(SO_VERSION) -no-undefined
 libkj_la_SOURCES=                                              \
-  src/kj/cidr.c++                                              \
   src/kj/common.c++                                            \
   src/kj/units.c++                                             \
   src/kj/memory.c++                                            \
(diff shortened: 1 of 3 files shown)


kentonv:
The KJ part shouldn't depend on winsock. Whatever code is depending on winsock probably belongs in kj-async instead.

But if the build is broken, why hasn't our CI complained about it? Probably the first thing we should do is fix CI such that it exposes the problem.

HazardyKnusperkeks:
I've put it into KJ, since the relevant file is `cidr.c++` which is listed in `kj_sources_lite`.
It only happens if KJ is build as shared library (`-DBUILD_SHARED_LIBS:BOOL=ON`), as it is done in the msys2 package and as far as I've seen your CI does not build the shared library.

kentonv:
Yeah, `cidr.c++` is only used by kj-async. I think it was added to libkj by mistake, it should have been in kj-async all along.

HazardyKnusperkeks:
Then do this. :) There is ws2_32 already linked.

kentonv:
Thanks! Do you mind updating Makefile.am and kj/BUILD.bazel to match?

HazardyKnusperkeks:
> Thanks! Do you mind updating Makefile.am and kj/BUILD.bazel to match?

Done. (To the best of my knowledge.)

kentonv:
Thanks!

PR #2364, opened 2025-07-26
Capnp: Set reasonable default include path on Windows

### LINKED ISSUE #1062: How to add search directories for capnp?

HazardyKnusperkeks:
At least for the msys Package this is the right thing to do.

Fixes #1062

### DIFF
--- a/c++/src/capnp/compiler/capnp.c++
+++ b/c++/src/capnp/compiler/capnp.c++
@@ -355,13 +355,28 @@ public:
         KJ_CONCAT(CAPNP_INCLUDE_DIR, _kj),
 #endif
       };
-      for (auto path: STANDARD_IMPORT_PATHS) {
+      auto add = [this](kj::StringPtr path) {
         KJ_IF_MAYBE(dir, getSourceDirectory(path, false)) {
           loader.addImportPath(*dir);
         } else {
           // ignore standard path that doesn't exist
         }
+      };
+      for (auto path : STANDARD_IMPORT_PATHS) {
+          add(path);
       }
+#if _WIN32
+      char exePath[MAX_PATH + 1];
+      KJ_SYSCALL(GetModuleFileNameA(nullptr, exePath, kj::size(exePath)));
+      kj::ArrayPtr path = exePath;
+      KJ_IF_MAYBE(slashPos, path.findLast('\\')) {
+        path = path.slice(0, *slashPos);
+        KJ_IF_MAYBE(slashPos, path.findLast('\\')) {
+          path = path.slice(0, *slashPos + 1);
+          add(kj::str(path, "include\\"));
+        }
+      }
+#endif
 
       addStandardImportPaths = false;
     }


HazardyKnusperkeks:
I don't know if the lambda is to your liking. And one could argue, that the `STANDARD_IMPORT_PATHS` could be completely dropped on Windows, since only `CAPNP_INCLUDE_DIR` ever can be a meaningful path on Windows.

kentonv:
This is a major change in behavior, and I want to be cautious not to introduce the wrong thing, as it'll be hard to revert later without breaking people.

At present, I don't quite understand why this makes sense for Windows, but not for other platforms.

In fact, it seems like the opposite to me: this code is assuming the binary lives in a `bin` directory which is a sibling of the `include` directory. That's assuming a Unix-style directory tree.

On Windows I would think the more typical thing would be for the binary to be installed in a folder under "Program Files". That folder might contain an "include" directory containing capnp imports, but typically the exe itself would not be under "bin" (although it could be! everyone does things differently on Windows!).

The bug you linked talks about MSYS2. MSYS indeed provides a unix-like directory tree, and so in that case, this approach does indeed make sense. But it seems wrong to assume all Windows users are using MSYS.

* Maybe there should be a compile-time flag to say we're targeting MSYS?
* Or maybe we can check specifically if the parent directory is called "bin", and only remove it in that case, before adding "include"?
* Even then, should we still have a compile-time flag to enable this new behavior at all, due to the risk of breaking people who might not be expecting it? I guess it doesn't seem incredibly likely that people would be broken, but it's possible.

It would be nice to know exactly what other tools with similar problems do, e.g. GCC. It's better to match existing well-understood behavior than to invent something new.

HazardyKnusperkeks:
Yes this assumes the msys2 paths. I should've added a check for `bin`, I see that too now.

I don't know about other ways to install capnproto under Windows. My colleague uses it installed by NuGet, and if I read the MSBuild code correctly he added the path of the binary (so without an `include`) to the import paths to get the c++ schema file.

I'm totally fine with a flag to activate that behavior, otherwise I would just propose this patch to the msys packages, but getting it in the official code is always preferred.

HazardyKnusperkeks:
With this version I added a flag, I oriented it on the CAPNP_LITE flag, as such I've only added to the CMake build.

And I decided to remove the `STANDARD_IMPORT_PATHS` for Windows (with the exception of `CAPNP_INCLUDE_DIR`), since they can never truly be found on a Windows machine.

HazardyKnusperkeks:
A little ping.

HazardyKnusperkeks:
FWIW it's added as a patch in Msys.

kentonv:
Sorry for the lack of replies, I'm constantly drowning in other priorities and just haven't had the bandwidth to really think about what's the right thing to do here.

But this PR as it stands doesn't seem like the right general solution. The problem exists across all flavors of Unix, but the PR is targeting only Windows, which ironically isn't a flavor of Unix, except specifically when using MSYS.

PR #2365, opened 2025-07-27
fix(kj): gracefully handle aborted sockets in accept on macOS

AaronO:
This fixes https://github.com/cloudflare/workerd/issues/4623

Adding two extra tests, reproducing the 2 crashes macOS users might observe
(`EINVAL` on `setsockopt()` and `addrlen == 0` on `accept()`, throwing in `NetworkFilter::shouldAllow()`)
```
❯ make -j6 capnp-test && ./capnp-test --filter 'src/kj/async-io*'
...
[ TEST ] kj/async-io-test.c++:3585: accept() skips aborted IPv4 connection
kj/async-io-unix.c++:1365: failed: setsocketopt(IPPROTO_TCP, TCP_NODELAY): Invalid argument
stack: 103ad1e73 103ad1667 102e03cfb 103521987
[ FAIL ] kj/async-io-test.c++:3585: accept() skips aborted IPv4 connection (2.366ms)
[ TEST ] kj/async-io-test.c++:3658: accept() skips aborted dual-stack connection
kj/async-io.c++:3120: failed: expected addrlen >= sizeof(addr->sa_family) [0 >= 1]
stack: 103a9d2e7 103ad1a8f 103ad1667 102e04e3b 103521987
[ FAIL ] kj/async-io-test.c++:3658: accept() skips aborted dual-stack connection (2.28ms)
95 test(s) passed
2 test(s) failed
```

Pushing a first commit with tests, so we can see the failures in CI, followed by a fix.

##

### DIFF
--- a/c++/src/kj/async-io-test.c++
+++ b/c++/src/kj/async-io-test.c++
@@ -3575,14 +3575,17 @@ KJ_TEST("pump file to socket") {
 }
 
 // ---------------------------------------------------------------------------------------------
-// accept() skips aborted connection – IPv4-only listener
+// accept() with aborted connection - IPv4 listener
 // ---------------------------------------------------------------------------------------------
-// An IPv4 client (AF_INET) connects and immediately resets (RST).  On BSD / macOS the kernel
-// removes the half-open connection from the backlog *during* the three-way handshake, so the
-// server observes the failure as an ECONNABORTED coming directly out of accept().  The test
-// verifies that KJ's accept loop eats the error and returns the next healthy connection.
+// Test accept() behavior when connections are aborted (send RST) before being accepted.
+// Creates an aborted IPv4 connection followed by a valid one, then verifies proper handling.
+//
+// On Unix platforms, accept() typically returns the aborted connection, which fails on first
+// read (throws "Connection reset by peer" or returns 0 bytes). The test then calls accept()
+// again to get the valid connection. Some platforms may filter out aborted connections.
 
-KJ_TEST("accept() skips aborted IPv4 connection") {
+#if !_WIN32
+KJ_TEST("accept() with aborted connection - IPv4") {
   // -- async-io boilerplate ---------------------------------------------------
   auto io = kj::setupAsyncIo();
   auto& net = io.provider->getNetwork();
@@ -3591,13 +3594,12 @@ KJ_TEST("accept() skips aborted IPv4 connection") {
   auto listener = listenerAddr->listen();
   uint16_t port = listener->getPort();
 
-  // -- Thread that opens Conn A (RST) ----------------------------------------
+  // Create a connection that will be aborted (sends RST packet)
   kj::Thread abortThread([&] {
     int s = ::socket(AF_INET, SOCK_STREAM, 0);
     KJ_ASSERT(s >= 0);
-    // KJ_DEFER(::close(s));
 
-    // Force RST on close.
+    // Configure socket to send RST on close instead of FIN
     struct linger lg = {1, 0};
     KJ_SYSCALL(setsockopt(s, SOL_SOCKET, SO_LINGER, &lg, sizeof(lg)));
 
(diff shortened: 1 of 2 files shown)


c++/src/kj/async-io-unix.c++
```diff
@@ -1355,6 +1362,10 @@ public:
         int one = 1;
         KJ_SYSCALL_HANDLE_ERRORS(::setsockopt(
               ownFd.get(), IPPROTO_TCP, TCP_NODELAY, (char*)&one, sizeof(one))) {
+#if __APPLE__
+          case EINVAL:  // macOS: fd already dead (RST race)
+            return acceptImpl(authenticated);   // retry accept()
```
kentonv: Hmm, I don't think we should throw away the socket just because trying to set TCP_NODELAY threw EINVAL. This is a pretty vague error. If you look a few lines below, we see FreeBSD apparently returns EINVAL in the case that the socket is a Unix socket -- but that doesn't mean the socket is dead, just that TCP_NODELAY isn't a valid option. I worry that there may be cases (now or in the future) where MacOS also returns EINVAL, but it doesn't actually mean there's anything wrong with the socket, only that TCP_NODELAY doesn't make sense for that socket.

I think instead we should match the FreeBSD behavior here and simply ignore the error -- pretend we never tried to set TCP_NODELAY at all, since it's not a required step anyway.

That would mean we end up returning a socket to the application which is already reset. I think that is fine -- it should result in exactly the same behavior as if the connection had been reset immediately after we returned it to the app. The app will just try to read and immediately get EOF or an error.

For the `addrlen = 0` case above, it's a bit more complicated. Are we _sure_ that a zero-length address _definitely_ always means the socket is already dead? Or is it possible there are network stacks that, like, don't have an address? It would be nice to see this documented somewhere. TBH the behavior seems like a kernel bug -- why would accept() bother to return a socket that it knows is already dead?

AaronO: I agree, even if the `EINVAL` here for tcp sockets on macOS is symptomatic of a dead connection, we can have it fail on first read/write.

I think we should then simply widen EINVAL handling to all platforms, instead of just FreeBSD, since setting NODELAY is an optimization and should have no effect on correctness.

`addrlen = 0` definitely seems to be a macOS quirk/bug (possibly other BSDs) and surprising.

I linked the relevant XNU `accept` syscall code in the parent workerd issue: https://github.com/apple-oss-distributions/xnu/blob/e3723e1f17661b24996789d8afc084c0c3303b26/bsd/kern/uipc_syscalls.c#L663-L691 we're hitting the `gotnoname` path.

`go` also handles this edge-case:
- Original fix: https://github.com/golang/go/commit/5197fa80405a470eef7cd2540380172aea6d60af
- Current code: https://github.com/golang/go/blob/6fbad4be75e7746512bbe55794694ed788ea5c5b/src/syscall/syscall_bsd.go#L282-L289

AaronO: The XNU bug is almost certainly a result of it [ignoring `soacceptlock()`'s return error](https://github.com/apple-oss-distributions/xnu/blob/e3723e1f17661b24996789d8afc084c0c3303b26/bsd/kern/uipc_syscalls.c#L663):
```c
	(void) soacceptlock(so, &sa, 0);
```

`soacceptlock` and the stack specific handlers, like [tcp_usr_accept](https://github.com/apple-oss-distributions/xnu/blob/e3723e1f17661b24996789d8afc084c0c3303b26/bsd/netinet/tcp_usrreq.c#L902-L913) do return `ECONNABORTED` but it's getting ignored there and so ultimately not passed through to the syscaller.

We should handle this edge case ourselves, since it seems to have been around for over a decade, possibly two or more. If they do fix it, then it'll take the happy path like other OSes. We can have the `addrlen == 0` check be `__APPLE__` specific if you prefer.

kentonv: Wow, I have to wonder if this has security implications. The OS is returning an uninitialized socket, I guess?

It looks like the code in this file will not handle zero-length addresses well (e.g. it assumes that `sa_family` is always present). Hopefully no other OSes ever return zero-length addresses, but it make sense to leave in a check for this just in case. Perhaps, though, on everything other than MacOS, we should `KJ_LOG(ERROR, "accept() returned zero-size address?");` about it. Otherwise it could be very confusing if some OS returns sockets that are actually valid, but missing addresses, and we just throw them away silently. On MacOS, though, we'll skip the log with a comment pointing at the bug in XNU.

kentonv: And as for the setsockopt error handling, I'd still prefer to keep the EINVAL check restricted to Apple and FreeBSD. EINVAL is such a generic error that it could easily mean something else in other contexts, and I don't want to miss it...

AaronO: Reflected these changes in the code.

Also updated tests, they're a bit verbose but exercise the failures we care about in the parent issue, they should also run on other platforms (except windows, don't have machine/vm at hand, so excluding it for now)

Could you rerun CI ?

AaronO:
The tests respectively exercise the `setsockopt` and the `addrlen == 0` failures macOS users encountered.
They test that these aborted connections are returned by our `accept()` but fail immediately on IO, except for IPv4/IPv6 case where we hit the `addrlen == 0` bug, can't pass a sockaddr through the `NetworkFilter` so have to silently discard and retry.

AaronO:
@kentonv Anything blocking landing this ?

c++/src/kj/async-io-test.c++
```diff
@@ -3574,5 +3574,203 @@ KJ_TEST("pump file to socket") {
...
+    a.sin_port = htons(port);
+    a.sin_addr.s_addr = htonl(0x7f000001);
+    KJ_SYSCALL(::connect(s, reinterpret_cast<sockaddr*>(&a), sizeof(a)));
+
+    // Close immediately to send RST packet
+    KJ_SYSCALL(::close(s));
+  });
+
+  // Allow aborted connection to reach accept queue first
+  io.provider->getTimer().afterDelay(2 * kj::MILLISECONDS)
```
kentonv, line 3617: Using delays like this commonly leads to flaky tests as CI can easily be too overloaded to schedule threads promptly. Could we instead wait for an explicit signal from the other thread, e.g. using conditional waits on a `kj::MutexGuarded<bool>`?

For that matter, is it actually necessary to use a separate thread to create the connection? It seems to me that this wait here will usually wait until `abortThread` is done, which makes it seem like you could just inline the body of `abortThread` into the main thread. But maybe that doesn't reproduce the bug?

c++/src/kj/async-io-test.c++
```diff
@@ -3574,5 +3574,203 @@ KJ_TEST("pump file to socket") {
...
+                        .wait(io.waitScope);
+
+  // Create a valid connection that should work
+  auto clientP = net.parseAddress("127.0.0.1", port)
+      .then([](Own<NetworkAddress> addr) { return addr->connect(); });
+  // Accept first connection - this should be the aborted one
+  auto firstConnection = listener->accept().wait(io.waitScope);
+  KJ_ASSERT(firstConnection);
```
kentonv, line 3625: This is redundant, `Own<T>` should never be null (otherwise it should be `Maybe<Own<T>>`). I'm actually slightly surprised this compiles.

```suggestion
```

c++/src/kj/async-io-test.c++
```diff
@@ -3574,5 +3574,203 @@ KJ_TEST("pump file to socket") {
...
+    // Read failed with exception (connection was aborted)
+    firstConnectionAborted = true;
+  }
+
+  kj::Own<AsyncIoStream> serverCon;
+  if (firstConnectionAborted) {
+    // First connection was aborted as expected - accept the second (valid) connection
+    serverCon = listener->accept().wait(io.waitScope);
+  } else {
+    // Unexpected: first connection was valid (kernel filtered out aborted one)
```
kentonv, line 3647: Is this really so unexpected? The Linux man page for accept() explicitly says that accept() itself may fail with ECONNABORTED (or other errors) rather than return a socket that is already dead.

c++/src/kj/async-io-test.c++
```diff
@@ -3574,5 +3574,203 @@ KJ_TEST("pump file to socket") {
...
+      .then([](Own<NetworkAddress> addr) { return addr->connect(); });
+  // Accept first connection - this should be the aborted one
+  auto firstConnection = listener->accept().wait(io.waitScope);
+  KJ_ASSERT(firstConnection);
+
+  char buffer[16];
+  bool firstConnectionAborted = false;
+  
+  // Test if first connection is aborted by attempting to read
+  if (kj::runCatchingExceptions([&]() {
```
kentonv, line 3631: Might be good to check that the exception caught is type `DISCONNECTED`, otherwise this code inadvertently passes if it triggers some unrelated exception.

c++/src/kj/async-io-test.c++
```diff
@@ -3574,5 +3574,203 @@ KJ_TEST("pump file to socket") {
...
+
+  // Create valid IPv6 connection
+  auto clientP = net.parseAddress("::1", port)
+      .then([](Own<NetworkAddress> addr) { return addr->connect(); });
+  // Accept connection - behavior differs by platform
+  auto serverCon = listener->accept().wait(io.waitScope);
+  KJ_ASSERT(serverCon);
+
+#if __APPLE__
+  // On Apple platforms: IPv4->IPv6 aborted connections return addrlen=0,
```
kentonv, line 3730: This is effectively testing for the XNU bug. If they fix the bug this test may start failing.

I think if you just delete the __APPLE__ side of the branch here, and use the other side for all platforms, it should work? Because that code is OK if the platform filters out the aborted connection. But it would still fail without your fix, right?

c++/src/kj/async-io-test.c++
```diff
@@ -3574,5 +3574,203 @@ KJ_TEST("pump file to socket") {
...
+  // Test data transfer on the valid connection
+  auto writePromise = clientCon->write(kj::StringPtr("hello").asBytes());
+  auto readPromise2 = serverCon->read(kj::arrayPtr(reinterpret_cast<kj::byte*>(buffer), sizeof(buffer)), 5);
+
+  writePromise.wait(io.waitScope);
+  auto amount = readPromise2.wait(io.waitScope);
+  KJ_ASSERT(amount == 5);
+  KJ_ASSERT(memcmp(buffer, "hello", 5) == 0);
+
+  abortThread.detach();
```
kentonv, line 3664: There's no need to detach the thread, as it has already completed. The destructor will join it which completes immediately.

Detaching the thread bad form since it contains references back to our stack frame which is going out-of-scope, so would then be dangling.

kentonv (commented):
Sorry for the delay, this got lost in my inbox.

The implementation looks good but I have a few concerns about the test.

AaronO:
> Sorry for the delay, this got lost in my inbox.
> 
> The implementation looks good but I have a few concerns about the test.

Yeah I agree, the tests are a bit rough, they reliably reproduce/test the parent issues but can be cleaned up.

kentonv:
Let me know if I should have someone on the workers team take over the test. I appreciate you're donating a fix for a workerd issue here which you have no obligation to spend time on.

AaronO:
> Let me know if I should have someone on the workers team take over the test. I appreciate you're donating a fix for a workerd issue here which you have no obligation to spend time on.

Not a problem, been busy, I'll cleanup the tests later today when I get some time.

c++/src/kj/async-io-test.c++
AaronO: We don't need a separate thread. We do however "need" the timeout, it's inherently somewhat racy.
We need that `RST` packet to be processed by kernel threads and moved through the net stack.

If we remove the sleep or yield (`sleep/usleep(0)`) that doesn't drive enough progress and we don't trigger this failure, we'll instead see an `ECONNRESET` on the `read`, this bug precisely happens when an `RST` is received before the conn is `accept`'d.

Practically speaking, very little work needs to happen, so even if CI is "overloaded" the kernel shouldn't be completely CPU starved, interrupts should allow the kernel to drive enough progress to make this non flaky on the order of a few ms of wallclock time.

It's not ideal I agree, but should be stable and non-flaky in practice (we can raise to ~10 or 100ms to reduce P(flaky) further, but I've yet to see it on many runs with 2ms, even `usleep(100)` seems sufficient).

AaronO:
Landed in #2389

PR #2368, opened 2025-08-01
Implement SmallArray

harrishancock:
These utilities originally landed in https://github.com/capnproto/capnproto/pull/1762. Unfortunately, they landed at the same time as a [subtle bug in our fiber implementation](https://github.com/capnproto/capnproto/pull/1842) began causing production crashes in the Workers runtime, and we reverted this `SmallArray` feature out of an abundance of caution.

I've been meaning to re-submit `SmallArray`. It is a more ergonomic interface than KJ_STACK_ARRAY, and the primary compiler we target nowadays (clang) does not support VLAs, which is KJ_STACK_ARRAY's competitive advantage.

In the intervening years, some HeapArrayDisposer was refactored to use `if constexpr` instead of a dispatch struct template in: https://github.com/capnproto/capnproto/pull/2288/files#diff-a50b82e04212952a2c02eee72504e7cca446a9b6bc0c89348bdf3f34fa5b4fa0. This meant that my old PR did not cleanly apply, but ultimately it made the diff here easier to follow, obviating a preliminary refactoring commit.

In my original PR where this work landed, I replaced KJ_STACK_ARRAY wholesale in a synchronized multiple-repo. That was, in retrospect, dumb and unnecessary. I would like to go through and replace instances of KJ_STACK_ARRAY with `SmallArray` and `SmallArrayBuilder`, but that can come at our leisure.

### DIFF
--- a/c++/src/kj/array.h
+++ b/c++/src/kj/array.h
@@ -144,7 +144,7 @@ public:
       : ptr(firstElement), size_(size), disposer(&disposer) {}
 
   KJ_DISALLOW_COPY(Array);
-  inline ~Array() noexcept { dispose(); }
+  inline ~Array() noexcept(false) { dispose(); }
 
   inline operator ArrayPtr<T>() KJ_LIFETIMEBOUND {
     return ArrayPtr<T>(ptr, size_);
--- a/c++/src/kj/array.h
+++ b/c++/src/kj/array.h
@@ -201,21 +201,21 @@ public:
   inline ArrayPtr<T> first(size_t count) KJ_LIFETIMEBOUND { return slice(0, count); }
   inline ArrayPtr<const T> first(size_t count) const KJ_LIFETIMEBOUND { return slice(0, count); }
 
-  inline ArrayPtr<const byte> asBytes() const KJ_LIFETIMEBOUND { 
+  inline ArrayPtr<const byte> asBytes() const KJ_LIFETIMEBOUND {
     KJ_ASSERT_CAN_MEMCPY(RemoveConst<T>);
-    return asPtr().asBytes(); 
+    return asPtr().asBytes();
   }
-  inline ArrayPtr<PropagateConst<T, byte>> asBytes() KJ_LIFETIMEBOUND { 
+  inline ArrayPtr<PropagateConst<T, byte>> asBytes() KJ_LIFETIMEBOUND {
     KJ_ASSERT_CAN_MEMCPY(RemoveConst<T>);
-    return asPtr().asBytes(); 
+    return asPtr().asBytes();
   }
-  inline ArrayPtr<const char> asChars() const KJ_LIFETIMEBOUND { 
+  inline ArrayPtr<const char> asChars() const KJ_LIFETIMEBOUND {
     KJ_ASSERT_CAN_MEMCPY(RemoveConst<T>);
-    return asPtr().asChars(); 
+    return asPtr().asChars();
   }
-  inline ArrayPtr<PropagateConst<T, char>> asChars() KJ_LIFETIMEBOUND { 
+  inline ArrayPtr<PropagateConst<T, char>> asChars() KJ_LIFETIMEBOUND {
     KJ_ASSERT_CAN_MEMCPY(RemoveConst<T>);
-    return asPtr().asChars(); 
+    return asPtr().asChars();
   }
 
   inline Array<PropagateConst<T, byte>> releaseAsBytes() {
--- a/c++/src/kj/array.h
+++ b/c++/src/kj/array.h
@@ -627,6 +627,63 @@ private:
 };
 
 // =======================================================================================
+// Small-buffer-optimized SmallArray
+//
+// SmallArray is useful when you need a temporary buffer, whose size you cannot know until runtime
+// but is likely to be small, and whose lifetime can be bounded by either the stack or some
+// immovable parent object.
+//
+// SmallArray is not an Array. In particular, it has the following differences:
+//
+// 1. SmallArray has an inline buffer of `smallSize` elements, where `smallSize` is a size_t
+//    template parameter. If one is constructed with a size less than or equal to `smallSize`, the
+//    inline space is used, and no heap allocation is performed. Otherwise, a regular heap Array is
+//    allocated.
+//
+// 2. SmallArray is immovable. You must construct one in place wherever you want to use one. They
+//    cannot be "released", "finished", or assigned-to.
+//
+// 3. SmallArray has no specific constructor functions like `heapArray<T>()`. Instead, use its
+//    constructor directly, passing a single `size` parameter.
+//
+// SmallArray requires its element type T to have a default constuctor. This is because SmallArray
+// always constructs and destructs the objects in its inline space, even if it ends up falling back
+// to a heap Array. This is done for implementation simplicity, and notably matches the behavior of
+// the `KJ_STACK_ARRAY` macro, which has the same use case as SmallArray.
+//
+// TODO(someday): Implement SmallArrayBuilder to support types which have no default constructor.
+
+template <typename T, size_t smallSize>
+class SmallArray final: private Array<T> {
+public:
+  explicit SmallArray(size_t size);
+
+  // We support the full Array<T> API except `releaseAsBytes()`, `releaseAsChars()`, `attach()`,
+  // `operator=()`, and move-construction.
+
+  KJ_DISALLOW_COPY_AND_MOVE(SmallArray);
+
+  using Array<T>::operator ArrayPtr<T>;
+  using Array<T>::operator ArrayPtr<const T>;
+  using Array<T>::asPtr;
+  using Array<T>::size;
+  using Array<T>::operator[];
+  using Array<T>::begin;
+  using Array<T>::end;
+  using Array<T>::front;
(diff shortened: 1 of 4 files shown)


c++/src/kj/array.h
```diff
@@ -144,7 +144,7 @@ class Array {
       : ptr(firstElement), size_(size), disposer(&disposer) {}
 
   KJ_DISALLOW_COPY(Array);
-  inline ~Array() noexcept { dispose(); }
+  inline ~Array() noexcept(false) { dispose(); }
```
harrishancock, line 147: This is, technically, a substantive change. I can't imagine that this change will cause problems, but it's worth calling out.

c++/src/kj/array.h
```diff
@@ -694,10 +807,28 @@ void ArrayDisposer::dispose(T* firstElement, size_t elementCount, size_t capacit
...
+            _::SmallArrayDisposer::allocateUninitialized<RemoveConst<T>>(space, size),
+            size,
+            _::SmallArrayDisposer::instance)
+        : heapArrayBuilder<T>(size)) {}
+
 namespace _ {  // private
 
 template <typename T>
-struct HeapArrayDisposer::Allocate_ {
+struct Allocate_ {
```
harrishancock: I had to un-nest this helper.

c++/src/kj/array.h
```diff
@@ -722,6 +853,24 @@ T* HeapArrayDisposer::allocateUninitialized(size_t count) {
 
 }
 
+template <typename T>
+T* SmallArrayDisposer::allocate(void* firstElement, size_t count) {
```
harrishancock: These two functions are now largely just copies of their HeapArrayDisposer counterparts. In my earlier PR, I had them all share code -- but the code they shared was the template metaprogramming version of `if constexpr`, and it's much clearer to just duplicate that, now.

c++/src/kj/array.h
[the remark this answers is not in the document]
harrishancock: Copilot is right that this is technically a breaking change -- processes which used to terminate when destroying an Array with an element which threw from its destructor will now see an exception propagate out of Array's destructor. But KJ objects are generally presumed to be able to throw from their destructors, thus KJ containers also need to be able to. This is a bugfix in that regard.

fhanau:
> the primary compiler we target nowadays (clang) does not support VLAs

That's the first time I'm hearing this. As I understand VLAs are supported in clang, a `-Wvla-cxx-extension` warning about VLAs being a C++ extension was only added in a recent release. We disallow VLAs in the internal build though, so I'm not saying they should be used here.

harrishancock:
Interesting, I stand corrected, then. I think I must have assumed clang didn't support them because our KJ_STACK_ARRAY implementation only used them on GCC, but perhaps that's just because GCC was the primary compiler in capnp's early years.

c++/src/kj/array.h
```diff
@@ -316,9 +316,26 @@ class HeapArrayDisposer final: public ArrayDisposer {
...
+public:
+  template <typename T>
+  static T* allocate(void* firstElement, size_t count);
   template <typename T>
-  struct Allocate_;
+  static T* allocateUninitialized(void* firstElement, size_t count);
+
+  static const SmallArrayDisposer instance;
+
+  static void* allocateImpl(void* firstElement, size_t elementSize, size_t elementCount,
```
fhanau: Nit: This can be private, as with HeapArrayDisposer

fhanau:
LGTM otherwise, but someone else should probably take a look too

c++/src/kj/array.h
harrishancock: Fixed

c++/src/kj/array.h
kentonv: Yeah this seems like a (very longstanding) bug.

c++/src/kj/array.h
kentonv: Should it have a less-vague name?

c++/src/kj/array.c++
```diff
@@ -68,26 +68,34 @@ struct AutoDeleter {
   inline ~AutoDeleter() { operator delete(ptr); }
 };
 
-void* HeapArrayDisposer::allocateImpl(size_t elementSize, size_t elementCount, size_t capacity,
-                                      void (*constructElement)(void*),
-                                      void (*destroyElement)(void*)) {
-  AutoDeleter result(operator new(elementSize * capacity));
+namespace {
 
+void allocateImplImpl(void* firstElement, size_t elementSize, size_t elementCount,
```
kentonv: Are you trying to get me back for the FactoryFactory?

c++/src/kj/array.h
```diff
@@ -316,9 +316,26 @@ class HeapArrayDisposer final: public ArrayDisposer {
 
   virtual void disposeImpl(void* firstElement, size_t elementSize, size_t elementCount,
                            size_t capacity, void (*destroyElement)(void*)) const override;
+};
 
+class SmallArrayDisposer: public ArrayDisposer {
```
kentonv: This could use some comments. Something like:

> ArrayDisposer used by SmallArray in the case that the array fits into the inlined buffer and so does not require a heap allocation.

c++/src/kj/array.h
```diff
@@ -316,9 +316,26 @@ class HeapArrayDisposer final: public ArrayDisposer {
...
+  static T* allocate(void* firstElement, size_t count);
   template <typename T>
-  struct Allocate_;
+  static T* allocateUninitialized(void* firstElement, size_t count);
+
+  static const SmallArrayDisposer instance;
+
+private:
+  static void* allocateImpl(void* firstElement, size_t elementSize, size_t elementCount,
+                     size_t capacity, void (*constructElement)(void*),
```
kentonv: Hmm, I wonder if having to virtualize the element constructor/destructor even in the small-buffer case will prove to be a significant performance hit vs. KJ_STACK_ARRAY.

With KJ_STACK_ARRAY we are assuming the type always has a default constructor and that it's OK for us to construct more elements than we actually use. Construction is probably pretty cheap in most if not all cases (probably just zeroing), though I suppose destruction could be doing a lot of wasteful branches checking if things are still null...

c++/src/kj/array.c++
```diff
@@ -105,5 +113,24 @@ void HeapArrayDisposer::disposeImpl(
 
 const HeapArrayDisposer HeapArrayDisposer::instance = HeapArrayDisposer();
 
+void* SmallArrayDisposer::allocateImpl(
```
mikea: I think SmallArrayDisposer should have <T> parameter to implement these better. Is it possible with current disposer design?

c++/src/kj/array.h
harrishancock: To solve this properly, I think we'd need to duplicate the HeapArrayDisposer logic (and related stuff like ExceptionSafeArrayUtil): SmallArray / SmallArrayBuilder would have a templatized disposer implementation, and regular (heap) Array / ArrayBuilder would continue having a type-erased disposer implementation. Or we could templatize the regular HeapArrayDisposer, too, but I presume that will have compile-time disadvantages that we don't want.

We could benchmark and do one of those courses of action.

Another option I thought of:
- `static_assert` that `SmallArray` is only used with types with trivial constructors and destructors.
- `static_assert` that `SmallArrayBuilder` is only used with types with trivial destructors.

But that would unfortunately prohibit using a `SmallArrayBuilder<Maybe<T>>`, which is my immediate use case, so I think I'd prefer to just duplicate the HeapArrayDisposer logic.

c++/src/kj/array.c++
harrishancock: Any specific ideas? I'm not sure I follow.

c++/src/kj/array.c++
harrishancock: I hate it, too. Maybe `exceptionSafeAllocate()`?

c++/src/kj/array.h
harrishancock: Thanks, I like that.

c++/src/kj/array.c++
harrishancock: Renamed to `exceptionSafeArrayAllocate()`.

c++/src/kj/array.h
harrishancock: Renamed to `AllocateConstructorHelper`.

c++/src/kj/array.h
kentonv: I'm not sure we actually need the full logic of `HeapArrayDisposer`. If we simply require that the type has a default constructor, then we can allocate the small array as a static-sized array with all elements constructed upfront (which is what KJ_STACK_ARRAY does). In this case construction and destruction are just handled by the compiler and we don't need a disposer at all.

harrishancock: Done, and dropped SmallArrayBuilder.

harrishancock:
Updated with a simpler implementation which always default-constructs the inline array, like KJ_STACK_ARRAY does. I dropped SmallArrayBuilder, meaning HeapArrayDisposer is now undisturbed. Additionally, I added `first()` and `as()` support to SmallArray, which were added to Array since my previous PR. Lastly, I fleshed out the commit message a bit more.

kentonv (approved):

PR #2369, opened 2025-08-03
fix #2359

### LINKED ISSUE #2359: fd passing breaks if the message is received in multiple chunks
fd passing breaks if capnp receives an rpc message in multiple chunks, e.g. because the sender is slow enough or the message exceeds the capnp receive buffer size. it seems that capnp ignores fds passed by all fragments of a message *except* the final one. this can easily happen when transferring large list arguments.

take this rpc definition and sample code:
```capnp
# rpc.capnp
@0xcc51fc1c24f3a383;

interface A {
}

interface B {
  run @0 (fill :List(UInt8), a :A);
}
```
```c++
// rpc.cc
#include "rpc.capnp.h"
#include <capnp/rpc-twoparty.h>
#include <iostream>

struct ASrv final : A::Server
{
    kj::Maybe<int> getFd() override { return 1; }
};

struct BSrv final : B::Server
{
    kj::Promise<void> run(RunContext rc) override
    {
        std::cerr << (co_await rc.getParams().getA().getFd()).orDefault(-1);
    }
};

int main()
{
    auto aio = kj::setupAsyncIo();
    auto pipe = aio.provider->newCapabilityPipe();
    capnp::TwoPartyServer srv{kj::heap<BSrv>()};
    auto srvp = srv.accept(*pipe.ends[0], 16);
    capnp::TwoPartyClient c{*pipe.ends[1], 16};
    auto foo = c.bootstrap().castAs<B>();
    auto runReq = foo.runRequest();
    runReq.initFill(65536);
    runReq.setA(kj::heap<ASrv>());
    runReq.send().wait(aio.waitScope);
}
```

compiling and running this produces and rpc message exceeding 65536 bytes, so it must be read in multiple fragments. stracing the process shows that the `A` fd is definitely sent, received, and ultimately closed, but `BSrv` receives an empty fd.

originally found in 1.0.2, but persists in 1.2.0 and the v2 branch.

pennae:
fixes #2359. should probably be backported to the next v1 release as well since this has broken things for lix.

### DIFF
--- a/c++/src/capnp/serialize-async.c++
+++ b/c++/src/capnp/serialize-async.c++
@@ -638,7 +638,8 @@ kj::Promise<kj::Maybe<MessageReaderAndFds>> BufferedMessageStream::tryReadMessag
 
   size_t expected = expectedSizeInWordsFromPrefix(data);
 
-  if (!leftoverFds.empty() && expected * sizeof(word) == dataByteSize) {
+  if (!leftoverFds.empty() &&
+      (expected * sizeof(word) == dataByteSize || expected > buffer.size() / 2)) {
     // We're about to return a message that consumes the rest of the data in the buffer, and
     // `leftoverFds` is non-empty. Those FDs are considered attached to whatever message contains
     // the last byte in the buffer. That's us! Let's consume them.
--- a/c++/src/capnp/serialize-async.c++
+++ b/c++/src/capnp/serialize-async.c++
@@ -638,8 +638,7 @@ kj::Promise<kj::Maybe<MessageReaderAndFds>> BufferedMessageStream::tryReadMessag
 
   size_t expected = expectedSizeInWordsFromPrefix(data);
 
-  if (!leftoverFds.empty() &&
-      (expected * sizeof(word) == dataByteSize || expected > buffer.size() / 2)) {
+  if (!leftoverFds.empty() && expected * sizeof(word) >= dataByteSize) {
     // We're about to return a message that consumes the rest of the data in the buffer, and
     // `leftoverFds` is non-empty. Those FDs are considered attached to whatever message contains
     // the last byte in the buffer. That's us! Let's consume them.
--- a/c++/src/capnp/rpc-twoparty-test.c++
+++ b/c++/src/capnp/rpc-twoparty-test.c++
@@ -505,39 +505,45 @@ KJ_TEST("send FD over RPC") {
 
   auto cap = client.bootstrap().castAs<test::TestMoreStuff>();
 
-  int pipeFds[2]{};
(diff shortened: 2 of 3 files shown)


c++/src/capnp/serialize-async.c++
```diff
@@ -638,7 +638,8 @@ kj::Promise<kj::Maybe<MessageReaderAndFds>> BufferedMessageStream::tryReadMessag
 
   size_t expected = expectedSizeInWordsFromPrefix(data);
 
-  if (!leftoverFds.empty() && expected * sizeof(word) == dataByteSize) {
+  if (!leftoverFds.empty() &&
+      (expected * sizeof(word) == dataByteSize || expected > buffer.size() / 2)) {
```
kentonv: I believe the correct fix is actually:

```suggestion
  if (!leftoverFds.empty() && expected * sizeof(word) >= dataByteSize) {
```

In other words, it should be a one-character change from the original code, to change `==` to `>=`.

This aligns with the comment below: The FDs in `leftoverFds` are logically attached to whatever message ends up containing the last byte in the current buffer. So when we've determined the message we are reading includes that last byte, then we should give in the `leftoverFds`.

For some reason, the code previously used `==` here, which meant we only gave it the `leftoverFds` if the message just happened to end exactly at the end of the buffer. If the message consumed the whole buffer _plus more_, then we didn't give it the `leftoverFds`. I suppose they ended up going to the next message, which is not great.

I think checking `expected > buffer.size() / 2` appears to fix the bug only because it happens to catch messages which are large enough to exceed the buffer size, which (at least in the current test) are also the ones that extend past the end of the buffer.

I've already verified my version also fixes your test, but let me know if you disagree with this analysis.

pennae: that sounds correct, but on its own may be inefficient? checking `>= dataByteSize` would open us to setting `fdsSoFar` nonzero for a message that would fit into the buffer, but *not* from the current starting position. this would cause the check in line 705 to match, reading the message into an allocated buffer as though it were oversize even though reclaiming buffer space would let us process the message in the current buffer. not using that path would just have us reclaim space and loop around to the same check until we have a full message.

allocating a buffer for fd-bearing messages does seem to be the intended behavior though as per the comment of the large-message block, and a bit of extra safety marging for rare occurences seems like a good idea too

kentonv: Yeah I think we explicitly have to use that behavior, per the comment: If we tried to do another buffered read in order to read the rest of the message _plus_ possibly some additional messages after it, there's a risk that one of those additional messages has its own file descriptors attached, and we then won't actually be able to tell if those file descriptors were attached to this message or the next message.

PR #2370, opened 2025-08-04
[bazel] Transition to bzlmod

fhanau:
This is required with the upcoming Bazel 9, workspace logic will no longer be supported. As with the master branch, we do not publish capnproto to the bzlmod registry – this is only changing how it is built.


mikea (approved):
LGTM provided you can figure out windows problems?

PR #2372, opened 2025-08-05
RPC bugfix: Properly catch exceptions from ClientHook::call().

kentonv:
Arguably, ClientHooks should not throw from this -- they should return a broken promise. But there are a lot of ClientHook implementations and they could really be doing anything.

Previously, an exception from here would propagate right out of `handleCall()` and kill the connection read loop, aborting the connection. Another side effect is that the `RpcCallContext` constructed in `handleCall()` would be destroyed without having sent a response, which would cause it to send a "canceled" return, despite the client not having actually asked to cancel anything. We are seeing these unexpected "canceled" returns in the wild.

### DIFF
--- a/c++/src/capnp/rpc.c++
+++ b/c++/src/capnp/rpc.c++
@@ -3860,7 +3860,26 @@ private:
       uint64_t interfaceId, uint64_t methodId,
       kj::Own<ClientHook>&& capability, kj::Own<CallContextHook>&& context,
       ClientHook::CallHints hints) {
-    return capability->call(interfaceId, methodId, kj::mv(context), hints);
+    try {
+      return capability->call(interfaceId, methodId, kj::mv(context), hints);
+    } catch (...) {
+      auto exception = kj::getCaughtExceptionAsKj();
+      auto pipeline = newBrokenPipeline(kj::cp(exception));
+
+      // In the past, an exception here would have killed the entire connection (and also caused
+      // us to send back a bogus Finish message claiming the call was canceled). We have evidence
+      // that such exceptions are being thrown, but we're not sure what they are. Hopefully they
+      // will now show up in a more reasonable place, but just to understand what was going on,
+      // let's log.
+      //
+      // TODO(cleanup): Remove this log once investigation is complete.
+      KJ_LOG(WARNING, "NOSENTRY RPC startCall() threw", exception);
+
+      return {
+        .promise = kj::mv(exception),
+        .pipeline = kj::mv(pipeline),
+      };
+    }
   }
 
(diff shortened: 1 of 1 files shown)


c++/src/capnp/rpc.c++
```diff
@@ -3860,7 +3860,26 @@ private:
       uint64_t interfaceId, uint64_t methodId,
       kj::Own<ClientHook>&& capability, kj::Own<CallContextHook>&& context,
       ClientHook::CallHints hints) {
-    return capability->call(interfaceId, methodId, kj::mv(context), hints);
+    try {
+      return capability->call(interfaceId, methodId, kj::mv(context), hints);
```
mikea, line 3864: I wonder: can call be simply made noexcept? Then compiler will simply break downstream on attempt to throw?

kentonv: No, I think the right answer is to catch the exceptions and propagate them as I'm doing here. There's no benefit to making every ClientHook implementation do it themselves.

kentonv: Also the compiler does not catch the problem unless you literally write `throw` inside of a `noexcept` function. If the `noexcept` function calls other functions that throw, then you get a crash, but not a compiler error. And we _never_ write bare `throw`, so we'd always end up with crashes.

mikea: Thanks. I wonder if there is a compiler flag that makes `noexcept` more strict like Java throws signatures: you can only call `noexcept` functions from `noexcept`? Or would such thing be too restrictive?

mikea: btw I found this: https://clang.llvm.org/extra/clang-tidy/checks/bugprone/exception-escape.html#bugprone-exception-escape

It has only one match on capnp sources: https://gist.github.com/mikea/0b66f89ab1286408676e690f017f4218

kentonv: I think in general I disagree with trying to enforce noexcept at compile time, because to me anything that can have a bug may have to throw an exception, and so trying to enforce the code doesn't throw exceptions is essentially trying to enforce that code has no bugs -- obviously impossible. What will end up happening in practice is when people write enforced-noexcept functions, they will simply not write asserts, and hope that there's no bug. That makes things worse: instead of failing loudly, we end up returning wrong answers and corrupting stuff.

I actually frequently use `noexcept` explicitly to mean "if an exception is thrown here, crash the process, because we cannot recover". I actually don't _want_ the compiler to enforce noexcept. I just want to escalate it to a crash when needed. I'd still rather crash than silently corrupt state.

Issue #2293, opened 2025-05-01
Cap'n Proto specification


spektor56:
in the specification it states that: “enumerants must be numbered sequentially starting from zero" and "fields (and enumerants …) must be numbered consecutively starting from zero in the order in which they were added"

So from this it seems to state that enumerals must be sequential and consecutive, but the compiler only enforces consecutive numbers and not sequential.

for non-consecutive numbers in enums the error message states: "test.capnp:4:11-12: error: Skipped ordinal @1.  Ordinals must be sequential with no holes."

because of this discrepancy, tools like [capnproto-dotnetcore](https://github.com/c80k/capnproto-dotnetcore) don't work as expected.

ex)
``
enum Ecu {
    eps @0;
    engine @4;
    unknown @5;
    abs @1;
    fwdRadar @2;
    fwdCamera @3;
}
``
this will result in the c# code
``
public enum Ecu : ushort
{
    eps,
    engine,
    unknown,
    abs,
    fwdRadar,
    fwdCamera
}
``

what is wrong here, the specification or the compiler?

kentonv:
The words "sequentially" and "consecutively" here were both intended to mean the same thing. Both mean "the ordinal numbers must form a sequence starting from zero and not skipping any number." Neither was intended to specify the order in which declarations appear in the schema file.

The intent is that, in your capnp schema files, declarations can appear in any order. The ordering of declarations does not affect the encoding at all. The only thing that affects the encoding is the ordinal numbers (`@` numbers), which must match the order in which the declarations were added to the protocol over time. The goal here is that you should be able to place your declarations in the order that is most natural for a human reader, even if that is not the order in which the protocol originally evolved.

Put another way, all three of these declarations are semantically identical:

```
enum Ecu { eps @0; engine @4; unknown @5; abs @1; fwdRadar @2; fwdCamera @3; }
enum Ecu { unknown @5; fwdRadar @2; engine @4; abs @1; eps @0; fwdCamera @3; }
enum Ecu { engine @4; eps @0; fwdCamera @3; abs @1; fwdRadar @2; unknown @5; }
```

I don't know C#, but I assume that if you don't specify explicit values for enumerants, they end up being assigned in the order they appear. In that case, it appears the capnproto-dotnetcore code generator is buggy.

jclee:
[...]

PR #2373, opened 2025-08-06
Preserve the code order in Compiler

mommsen:
The fields in `Compiler::Node::traverseNodeDependencies` should be traversed in code order. This ensures that, e.g., enumerants preserve the correct ordinal numbers if they are not declared in sequence. This approach is replicated from the C++ code generation in `CapnpcCapnpMain`.

We found this issue when using `Compiler::add` with an enum with out-of-sequence enumerants, e.g.,
```
enum Test {
  zero @0;
  two @2;
  one @1;
}
```
In this case, the numerical values for `one` and `two` are swapped. This MR fixes this problem.

BTW: I have not managed to create a unit test for `compiler` to cover this case.

cc @fabiorossetto

### DIFF
--- a/c++/src/capnp/compiler/compiler.c++
+++ b/c++/src/capnp/compiler/compiler.c++
@@ -716,6 +716,23 @@ void Compiler::Node::traverse(uint eagerness, std::unordered_map<Node*, uint>& s
   }
 }
 
+namespace {
+
+struct OrderByCodeOrder {
+  template <typename T> inline bool operator()(const T &a, const T &b) const {
+    return a.getCodeOrder() < b.getCodeOrder();
+  }
+};
+
+template <typename MemberList>
+kj::Array<decltype(kj::instance<MemberList>()[0])> sortByCodeOrder(MemberList &&list) {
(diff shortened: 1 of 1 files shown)


kentonv:
It looks like you closed this an hour after opening it. Did you decide it isn't needed?

TBH I can't see what difference this makes. This code is just walking the tree to discover dependencies. I don't see why it would matter in what order the dependencies are walked.

I double-checked that the generated code (in C++ at least) does in fact ensure that the numeric values correctly match the ordinals regardless of the code order.

jclee:
Might have been an attempt to fix an issue that was actually a bug in a third-party generator, like: https://github.com/c80k/capnproto-dotnetcore/issues/77 ? (previously seen in https://github.com/capnproto/capnproto/issues/2293)

mommsen:
It's indeed not needed. I first thought it solves the issue at hand, but this is not the case. I'll need to dig deeper.

The use case is a custom Python stub creator, that does not work correctly if the enumerants are not in sequence. I'll try to come up with an example showing the problem, and maybe with a solution.

PR #2376, opened 2025-08-08
Async pipes: Avoid causing a concurrent invocation of tryRead() when calling abortRead()

npaun:
Some of our async streams don't allow for concurrent calls to `tryRead()`.

For example, here is `HttpFixedLengthEntityReader::tryRead`:

```c++

Promise<size_t> tryRead(void* buffer, size_t minBytes, size_t maxBytes) override {
    KJ_REQUIRE(clean, "can't read more data after a previous read didn't complete");
    clean = false;

    size_t alreadyRead = 0;

    for (;;) {
      if (alreadyDone()) {
        clean = true;
        co_return alreadyRead;
      } else if (chunkSize == 0) {
        // ...
      } else {
        // ...
        auto amount = co_await getInner().tryRead(buffer, 1, kj::min(maxBytes, chunkSize));

        chunkSize -= amount;
        if (amount == 0) {
          // ...
          buffer = reinterpret_cast<byte*>(buffer) + amount;
          minBytes -= amount;
          maxBytes -= amount;
          alreadyRead += amount;
          continue;
        }
        clean = true;
        co_return alreadyRead + amount;
      }
    }
```

Now suppose that a `tryRead()` call is blocked waiting for some data, and then we decide to abort reading from this stream.
In order to decide whether the stream was at EOF, `BlockedPipeFrom::abortRead()` reads one more byte of data:

```c++

        canceler.cancel("abortRead() was called");

        // The input might have reached EOF, but we haven't detected it yet because we haven't tried
        // to read that far. If we had not optimized tryPumpFrom() and instead used the default
        // pumpTo() implementation, then the input would not have called write() again once it
        // reached EOF, and therefore the abortRead() on the other end would *not* propagate an
        // exception! We need the same behavior here. To that end, we need to detect if we're at EOF
        // by reading one last byte.
        checkEofTask = kj::evalNow([&]() {
          static char junk;
          return input.tryRead(&junk, 1, 1).then([this](uint64_t n) {
            if (n == 0) {
              fulfiller.fulfill(kj::cp(pumpedSoFar));
            } else {
              fulfiller.reject(KJ_EXCEPTION(DISCONNECTED, "read end of pipe was aborted"));
            }
          }).eagerlyEvaluate([this](kj::Exception&& e) {
            fulfiller.reject(kj::mv(e));
          });
        });
```

It's apparent that this will cause a failed assertion in `HttpFixedLengthEntityReader`. The same logic applies to `HttpChunkedEntityReader`. @harrishancock and I believe this happens sometimes when a `fetch()` is aborted, leading to a frequent Sentry issue, although this has been hard to prove.

This PR fixes this issue by just assuming that if a `tryRead()` is in progress, we are, by definition, not at EOF and we can tell this to the caller.

When reviewing these changes please also have a look at [`AbortedRead::tryPumpFrom`](https://github.com/capnproto/capnproto/blob/83a83aca9b0ffe37af62f6b871ccaa6afb940a99/c%2B%2B/src/kj/async-io.c%2B%2B#L1462) which contains some very similar looking logic. Does it need to also be patched in some way?

### DIFF
--- a/c++/src/kj/async-io-test.c++
+++ b/c++/src/kj/async-io-test.c++
@@ -36,6 +36,7 @@
 #include <kj/time.h>
 #include <sys/types.h>
 #include <kj/filesystem.h>
+
 #if _WIN32
 #include <ws2tcpip.h>
 #include "windows-sanity.h"
@@ -3572,6 +3573,45 @@ KJ_TEST("pump file to socket") {
   // Try with a disk file. Should use sendfile().
   auto fs = kj::newDiskFilesystem();
   doTest(fs->getCurrent().createTemporary());
+}
+
+KJ_TEST("Calling abortRead() while tryRead() is in progress") {
+  // This is a stream that will only permit one call to tryRead() to be in progress at any one time
+  // Any tryRead calls won't ever complete
+  class NonConcurrentStream final: public kj::AsyncInputStream {
+  public:
+    virtual kj::Promise<size_t> tryRead(void* buffer, size_t minBytes, size_t maxBytes) override {
+      numConcurrentCalls++;
+      KJ_DEFER(numConcurrentCalls--);
+      co_await (kj::Promise<void>) kj::NEVER_DONE;
+      co_return 0;
+    }
+
+    uint numConcurrentCalls = 0;
+  };
+ 
+  EventLoop eventLoop;
+  WaitScope ws(eventLoop);
+
+  {
+    auto nonConcurrentStream = kj::heap<NonConcurrentStream>();
+    
+    auto pipe = kj::newOneWayPipe();
+    auto pumpTask = nonConcurrentStream->pumpTo(*pipe.out).ignoreResult();
+    auto readTask = pipe.in->readAllBytes().ignoreResult();
+
+    // Assert that pumpTask and readTask are stuck
+    KJ_EXPECT(!pumpTask.poll(ws));
+    KJ_EXPECT(!readTask.poll(ws));
+
+    // Close the read end of the pipe (will cause an abortRead())
+    pipe.in = nullptr;
+
+    // abortRead() shouldn't invoke tryRead() again because our stream can't handle it
+    // See HttpFixedLengthEntityInputReader and HttpChunkedEntityInputReader for real examples of
+    // such streams.
+    KJ_EXPECT(nonConcurrentStream->numConcurrentCalls == 0);
+  }
 }
 
 }  // namespace
--- a/c++/src/kj/async-io.c++
+++ b/c++/src/kj/async-io.c++
@@ -841,26 +841,32 @@ private:
     }
 
     void abortRead() override {
-      canceler.cancel("abortRead() was called");
+      if (canceler.isEmpty()) {
+        // No read in flight, we must check for EOF.
 
-      // The input might have reached EOF, but we haven't detected it yet because we haven't tried
-      // to read that far. If we had not optimized tryPumpFrom() and instead used the default
-      // pumpTo() implementation, then the input would not have called write() again once it
-      // reached EOF, and therefore the abortRead() on the other end would *not* propagate an
-      // exception! We need the same behavior here. To that end, we need to detect if we're at EOF
-      // by reading one last byte.
-      checkEofTask = kj::evalNow([&]() {
-        static char junk;
-        return input.tryRead(&junk, 1, 1).then([this](uint64_t n) {
-          if (n == 0) {
-            fulfiller.fulfill(kj::cp(pumpedSoFar));
-          } else {
-            fulfiller.reject(KJ_EXCEPTION(DISCONNECTED, "read end of pipe was aborted"));
-          }
-        }).eagerlyEvaluate([this](kj::Exception&& e) {
-          fulfiller.reject(kj::mv(e));
+        // The input might have reached EOF, but we haven't detected it yet because we haven't tried
+        // to read that far. If we had not optimized tryPumpFrom() and instead used the default
+        // pumpTo() implementation, then the input would not have called write() again once it
+        // reached EOF, and therefore the abortRead() on the other end would *not* propagate an
+        // exception! We need the same behavior here. To that end, we need to detect if we're at EOF
+        // by reading one last byte.
+        checkEofTask = kj::evalNow([&]() {
+          static char junk;
+          return input.tryRead(&junk, 1, 1).then([this](uint64_t n) {
+            if (n == 0) {
+              fulfiller.fulfill(kj::cp(pumpedSoFar));
+            } else {
+              fulfiller.reject(KJ_EXCEPTION(DISCONNECTED, "read end of pipe was aborted"));
+            }
+          }).eagerlyEvaluate([this](kj::Exception&& e) {
+            fulfiller.reject(kj::mv(e));
+          });
         });
-      });
+      } else {
+        // Read is in-flight; by definition we are not at EOF.
+        canceler.cancel("abortRead() was called");
+        fulfiller.reject(KJ_EXCEPTION(DISCONNECTED, "read end of pipe was aborted"));
+      }
 
       pipe.endState(*this);
       pipe.abortRead();


c++/src/kj/async-io-test.c++
```diff
@@ -3574,5 +3576,45 @@ KJ_TEST("pump file to socket") {
...
 
+KJ_TEST("Calling abortRead() while tryRead() is in progress") {
+  // This is a stream that will only permit one call to tryRead() to be in progress at any one time
+  // Any tryRead calls won't ever complete
+  class NonConcurrentStream: public kj::AsyncInputStream {
+    public:
+    virtual kj::Promise<size_t> tryRead(void* buffer, size_t minBytes, size_t maxBytes) override {
+      numConcurrentCalls.fetch_add(1);
+      co_await (kj::Promise<void>) kj::NEVER_DONE;
+      numConcurrentCalls.fetch_sub(1);
```
kentonv: This line will never execute since the previous line will never return normally -- exceptions and cancellation will skip this. But you probably want to decrement the counter in those cases, right? You can move this up and wrap it in KJ_DEFER() to make sure it runs no matter how the scope is exited.

c++/src/kj/async-io-test.c++
```diff
@@ -3574,5 +3576,45 @@ KJ_TEST("pump file to socket") {
...
+      co_return 0;
+    }
+
+    virtual ~NonConcurrentStream() {}
```
kentonv: You probably added this to solve a compiler warning. But the better way to fix that warning, which we prefer throughout KJ-style code, is to declare the class final:

```
class NonConcurrentStream final: public kj::AsyncInputStream {
```

c++/src/kj/async-io-test.c++
```diff
@@ -3574,5 +3576,45 @@ KJ_TEST("pump file to socket") {
...
+
+    std::atomic_int numConcurrentCalls = 0;
```
kentonv: This doesn't need to be atomic since all access is from the same thread.

c++/src/kj/async-io-test.c++
```diff
@@ -3574,5 +3576,45 @@ KJ_TEST("pump file to socket") {
...
+  auto& ws = ioContext.waitScope;
+
+  {
+    auto nonConcurrentStream = kj::heap<NonConcurrentStream>();
+    
+    auto pipe = kj::newOneWayPipe();
+    auto pumpTask = nonConcurrentStream->pumpTo(*pipe.out).ignoreResult();
+    auto readTask = pipe.in->readAllBytes().ignoreResult();
+
+    // Wait an arbitrary amount of time for both tasks to get stuck
```
kentonv: You can instead do:

```
KJ_EXPECT(!pumpTask.poll(ws));
KJ_EXPECT(!readTask.poll(ws));
```

Doing `promise.poll(ws)` runs the event loop until either the promise settles (returning true), or there's nothing left to do (returning false). So this actually asserts that the tasks are "stuck".

c++/src/kj/async-io-test.c++
```diff
@@ -3574,5 +3576,45 @@ KJ_TEST("pump file to socket") {
...
+      co_return 0;
+    }
+
+    virtual ~NonConcurrentStream() {}
+
+    std::atomic_int numConcurrentCalls = 0;
+  };
+ 
+  auto ioContext = setupAsyncIo();
+  auto& ws = ioContext.waitScope;
```
kentonv: This test doesn't need any system I/O, so you can do:

```suggestion
  EventLoop eventLoop;
  WaitScope ws(eventLoop);
```

This sets up a purely in-memory event loop with no OS resources, which is much more efficient.

c++/src/kj/async-io.c++
```diff
@@ -841,26 +841,33 @@ private:
...
-            fulfiller.fulfill(kj::cp(pumpedSoFar));
-          } else {
-            fulfiller.reject(KJ_EXCEPTION(DISCONNECTED, "read end of pipe was aborted"));
-          }
-        }).eagerlyEvaluate([this](kj::Exception&& e) {
-          fulfiller.reject(kj::mv(e));
+      if (canceler.isEmpty()) {
+        // No read in flight, we must check for EOF.
+
+        canceler.cancel("abortRead() was called");
```
kentonv: If the canceler is empty, this is a no-op.

You should move this to the other side of the if/else, and remove the `fulfiller.reject()`, since canceling the read will accomplish the same result.

c++/src/kj/async-io.c++
```diff
@@ -841,26 +841,33 @@ private:
...
+              fulfiller.reject(KJ_EXCEPTION(DISCONNECTED, "read end of pipe was aborted"));
+            }
+          }).eagerlyEvaluate([this](kj::Exception&& e) {
+            fulfiller.reject(kj::mv(e));
+          });
         });
-      });
+      } else {
+        // Read is in-flight; by definition we are not at EOF.
+        fulfiller.reject(KJ_EXCEPTION(DISCONNECTED, "read end of pipe was aborted while read in-flight"));
```
kentonv: This would change the exception thrown in this case. Let's keep it as it was, to avoid unintended side effects:

```suggestion
        canceler.cancel("abortRead() was called");
```

c++/src/kj/async-io-test.c++
npaun: True. I originally used an int then got unnecessarily paranoid.

npaun:
Just pushed a rebase because it contains a build fix I need

npaun:
@kentonv Thanks for the review. I've made the changes. 

While you're here, what do you think about [`AbortedRead::tryPumpFrom`](https://github.com/capnproto/capnproto/blob/83a83aca9b0ffe37af62f6b871ccaa6afb940a99/c%2B%2B/src/kj/async-io.c%2B%2B#L1462)? Do I need to change anything there?

c++/src/kj/async-io.c++
```diff
@@ -841,26 +841,31 @@ private:
...
+              fulfiller.reject(KJ_EXCEPTION(DISCONNECTED, "read end of pipe was aborted"));
+            }
+          }).eagerlyEvaluate([this](kj::Exception&& e) {
+            fulfiller.reject(kj::mv(e));
+          });
         });
-      });
+      } else {
+        // Read is in-flight; by definition we are not at EOF.
+        canceler.cancel("abortRead() was called");
```
kentonv, line 867: I looked again and my previous suggested wasn't quite right. We do need to explicitly reject the fulfiller after all. But let's not change any error messages. So it should be:

```suggestion
        canceler.cancel("abortRead() was called");
        fulfiller.reject(KJ_EXCEPTION(DISCONNECTED, "read end of pipe was aborted"));
```

c++/src/kj/async-io-test.c++
```diff
@@ -3574,5 +3575,44 @@ KJ_TEST("pump file to socket") {
...
+  class NonConcurrentStream final: public kj::AsyncInputStream {
+    public:
+    virtual kj::Promise<size_t> tryRead(void* buffer, size_t minBytes, size_t maxBytes) override {
+      numConcurrentCalls++;
+      KJ_DEFER(numConcurrentCalls--);
+      co_await (kj::Promise<void>) kj::NEVER_DONE;
+      co_return 0;
+    }
+
+    int numConcurrentCalls = 0;
```
kentonv: Nitpick, KJ style always uses uint unless negative numbers are expected.

```suggestion
    uint numConcurrentCalls = 0;
```

c++/src/kj/async-io-test.c++
```diff
@@ -3574,5 +3575,44 @@ KJ_TEST("pump file to socket") {
   doTest(fs->getCurrent().createTemporary());
 }
 
+KJ_TEST("Calling abortRead() while tryRead() is in progress") {
+  // This is a stream that will only permit one call to tryRead() to be in progress at any one time
+  // Any tryRead calls won't ever complete
+  class NonConcurrentStream final: public kj::AsyncInputStream {
+    public:
```
kentonv:
```suggestion
  public:
```

npaun:
I can't merge to capnproto, can you please do it for me?

PR #2377, opened 2025-08-10
fix: include libucontext in Requires.private in cmake builds

lf-:
This is required so that static musl actually links to libucontext correctly to get setcontext/etc for fibers.

Using this allows the packager to set up the libucontext pkg-config however they wish so it actually works, rather than requiring that the -L path or other stuff is correct by downstream users.

The current state is that -lucontext isn't set at all on cmake builds, which is unambiguously broken (https://github.com/NixOS/nixpkgs/pull/432323), but it is not the correct approach to copy the autoconf build and merely set `-lucontext` in `Libs` as the latter ignores the pkg-config of libucontext itself (problematic!).

### DIFF
--- a/c++/CMakeLists.txt
+++ b/c++/CMakeLists.txt
@@ -96,6 +96,9 @@ set_property(CACHE WITH_FIBERS PROPERTY STRINGS AUTO ON OFF)
 # CapnProtoConfig.cmake.in needs this variable.
 set(_WITH_LIBUCONTEXT OFF)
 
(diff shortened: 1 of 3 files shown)


kentonv:
Harris is the guy who actually knows cmake but he's out of office this week so it might take a bit to review.

FWIW another option is to disable fibers with `-DKJ_USE_FIBERS=0`. I doubt many people use them, and coroutines make them somewhat obsolete.

PR #2379, opened 2025-08-11
Add support for BAN http method

danlapid:


c++/src/kj/compat/http.h
```diff
@@ -80,7 +80,8 @@ namespace kj {
   MACRO(NOTIFY) \
   MACRO(SUBSCRIBE) \
   MACRO(UNSUBSCRIBE) \
-  MACRO(QUERY)
+  MACRO(QUERY) \
+  MACRO(BAN)
```
kentonv: Looks like both QUERY and BAN have been mislabeled as being from "UPnP" (comment on next line).

QUERY should reference the RFC.

For BAN we should write:

> Non-standard method name requested by a Cloudflare customer.

kentonv (commented):
Can we add a test? (Or is there one that automatically covers this?)

c++/src/kj/compat/http.h
danlapid: Done

kentonv:
> Can we add a test? (Or is there one that automatically covers this?)

Looks like existing tests automatically cover all methods, ok.

kentonv (approved):

PR #2383, opened 2025-08-14
Include header name in invalid value error messages

jasnell:
the extra whitespace edits here are clang-format bits... can remove them if necessary

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -432,7 +432,7 @@ static kj::Maybe<kj::OneOf<HttpMethod, HttpConnectMethod>> consumeHttpMethod(cha
         case 'S': EXPECT_REST(UNS,UBSCRIBE)
         default: return kj::none;
       }
-    
+
     default: return kj::none;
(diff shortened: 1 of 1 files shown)


c++/src/kj/compat/http.c++
```diff
@@ -698,7 +698,7 @@ bool HttpHeaders::isWebSocket() const {
 
 void HttpHeaders::set(HttpHeaderId id, kj::StringPtr value) {
   id.requireFrom(*table);
-  requireValidHeaderValue(value);
+  requireValidHeaderValue(value, id.toString());
```
kentonv: Hmm, this `toString()` function might actually add non-negligible overhead to this method, which is called pretty often.

Can we inline `requireValidHeaderValue` into the call site here so that we can pass `id` directly to `KJ_REQUIRE` such that it will only be stringified if the condition fails?

(Or another alternative would be to make `requireValidHeaderValue` be a template so that the `name` parameter can have any type, so it then doesn't need to be converted to a string prematurely.)

PR #2385, opened 2025-08-15
switch to c++23

mikea:
Doesn't use any features, just a CI change.


c++/src/kj/refcount.h
```diff
@@ -169,8 +169,7 @@ class Rc {
 
 public:
   KJ_DISALLOW_COPY(Rc);
-  Rc() { }
-  Rc(decltype(nullptr)) { }
```
kentonv: Can you put in the PR description an explanation of the code changes needed? I'm curious.

mikea: I think it was unrelated change that snuck its way through. Reverted.

PR #2386, opened 2025-08-15
improved c++23 .as() with derived this

mikea:


c++/src/kj/array.h
```diff
@@ -263,12 +263,12 @@ class Array {
...
 
   template <typename U>
-  inline auto as() { return U::from(this); }
+  inline auto as(this auto& self) { return U::from(&self); }
   // Syntax sugar for invoking U::from.
   // Used to chain conversion calls rather than wrap with function.
 
   template <typename U>
-  inline auto as() const { return U::from(this); }
+  inline auto as(this auto const& self) { return U::from(&self); }
```
kentonv, line 271: Maybe this overload can be deleted since the other one covers it now?

c++/src/kj/array.h
```diff
@@ -263,12 +263,12 @@ class Array {
   // Like Own<T>::attach(), but attaches to an Array.
 
   template <typename U>
-  inline auto as() { return U::from(this); }
+  inline auto as(this auto& self) { return U::from(&self); }
```
kentonv, line 266: What's the motivation for this exactly? Seems like it only makes a difference if someone is subclassing `Array`, and the subclasses should be treated differently?

mikea: I initially thought to have safe conversions for Text::Reader (see https://github.com/cloudflare/workerd-cxx/blob/master/kj-rs/convert.h#L73) but I have reconsidered since Reader doesn't give hard utf8 guanrantees either.

PR #2387, opened 2025-08-15
assert non-negative allocation sizes

mikea:
debugging production issues. Hopefully this will crash.

FWIW we think we found the issue here: https://github.com/capnproto/capnproto/pull/2388

### DIFF
--- a/c++/src/kj/arena.c++
+++ b/c++/src/kj/arena.c++
@@ -135,7 +135,7 @@ void* Arena::allocateBytesInternal(size_t amount, uint alignment) {
   }
 
   // Allocate.
-  byte* bytes = reinterpret_cast<byte*>(operator new(nextChunkSize));
+  byte* bytes = reinterpret_cast<byte*>(operator new(_::requireValidAllocSize(nextChunkSize)));
 
   // Set up the ChunkHeader at the beginning of the allocation.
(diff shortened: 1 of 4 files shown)


c++/src/kj/arena.c++
```diff
@@ -21,6 +21,7 @@
 
 #include "arena.h"
 #include "debug.h"
+#include "kj/memory.h"
```
fhanau: nit: Use angled brackets or plain memory.h. Same with table.c++

fhanau: Sorry, you fixed one but not the other

kentonv:
Wouldn't these all throw std::bad_alloc today? Is there a benefit to adding our own check?

c++/src/kj/memory.h
```diff
@@ -123,6 +123,12 @@ const void* castToConstVoid(T* ptr) {
 
 KJ_NORETURN(void throwWrongDisposerError());
 
+inline size_t requireValidAllocSize(size_t size) {
+  constexpr const std::size_t highest_bit = std::size_t(1) << (sizeof(std::size_t) * 8 - 1);
+  KJ_IREQUIRE(!(size & highest_bit), "Invalid allocation size (negative)");
```
kentonv, line 128: > debugging production issues

Note that IREQUIRE is compiled out in production.

PR #2389, opened 2025-08-17
Fix macOS crash when accept() returns zero-length address

danlapid:
As reported in https://github.com/cloudflare/workerd/issues/4623

On macOS, accept() can return a valid file descriptor with addrlen == 0 when a connection is aborted during the accept (e.g., during "happy eyeballs" dual-stack connection attempts). This was causing workerd to crash with:

  Fatal uncaught kj::Exception: kj/async-io.c++:3126:
  failed: expected addrlen >= sizeof(addr->sa_family) [0 >= 1]

This fix:
1. Checks for addrlen == 0 after accept() in async-io-unix.c++ and treats it as an aborted connection that should be retried (I was able to reproduce the error with the given stress test code before this change and was not able to after).
2. Adds EINVAL to the list of acceptable errors when setting TCP_NODELAY on macOS, as it can be returned for non-TCP sockets. (I was consistently getting this in local dev when trying to repro).

Fixes cloudflare/workerd#4623

🤖 Generated with [Claude Code](https://claude.ai/code)


c++/src/kj/async-io-unix.c++
```diff
@@ -1343,6 +1343,15 @@ public:
 
     if (newFd >= 0) {
       kj::OwnFd ownFd(newFd);
+      
+      // On macOS, accept() can return a valid file descriptor with addrlen == 0 when a
```
danlapid: I can wrap this in an ifdef APPLE if you prefer but this seemed like a generally good idea even for linux so I left this unguarded.

kentonv:
Let's use #2365, it's a more polished fix and includes a test. It just needs a couple comments addressed on the test.

danlapid:
Pushed commit to use the exact code of https://github.com/capnproto/capnproto/pull/2365/files , no code changes as of yet so review comments are not yet addressed.

c++/src/kj/async-io-test.c++
```diff
@@ -3585,17 +3585,17 @@ KJ_TEST("Calling abortRead() while tryRead() is in progress") {
       KJ_DEFER(numConcurrentCalls--);
       co_await (kj::Promise<void>) kj::NEVER_DONE;
       co_return 0;
-    }
```
kentonv: Seems like you misaligned this brace in an unrelated existing test?

danlapid: Yeah, I'll tag you when this is actually ready for a look

PR #2390, opened 2025-08-18
Handle 'Transfer-Encoding: chunked, chunked' as single chunked encoding

danlapid:
This patch addresses a compatibility issue where some clients/proxies incorrectly send "Transfer-Encoding: chunked, chunked" headers. While this is invalid per HTTP spec, rejecting these requests was causing production errors.

The fix treats both "chunked, chunked" and "chunked,chunked" (with or without space) as equivalent to plain "chunked" encoding.

If the sender actually meant double-chunked encoding, the data will be corrupted (only outer layer unwrapped), but this is the sender's fault for violating the HTTP specification.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -2182,6 +2182,18 @@ kj::Own<kj::AsyncInputStream> HttpInputStreamImpl::getEntityBody(
     if (fastCaseCmp<'c','h','u','n','k','e','d'>(te.cStr())) {
       // #3¶1
       return kj::heap<HttpChunkedEntityReader>(*this);
+    } else if (fastCaseCmp<'c','h','u','n','k','e','d',',',
+                      ' ','c','h','u','n','k','e','d'>(te.cStr()) ||
+               fastCaseCmp<'c','h','u','n','k','e','d',',',
+                      'c','h','u','n','k','e','d'>(te.cStr())) {
+      // Handle "chunked, chunked" (with or without space) as equivalent to "chunked"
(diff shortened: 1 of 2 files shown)


c++/src/kj/compat/http.c++
```diff
@@ -2182,6 +2182,14 @@ kj::Own<kj::AsyncInputStream> HttpInputStreamImpl::getEntityBody(
     if (fastCaseCmp<'c','h','u','n','k','e','d'>(te.cStr())) {
       // #3¶1
       return kj::heap<HttpChunkedEntityReader>(*this);
+    } else if (fastCaseCmp<'c','h','u','n','k','e','d',',',
+                      ' ','c','h','u','n','k','e','d'>(te.cStr()) ||
+               fastCaseCmp<'c','h','u','n','k','e','d',',',
+                      'c','h','u','n','k','e','d'>(te.cStr())) {
+      // Handle "chunked, chunked" (with or without space) as equivalent to "chunked"
+      // This is technically invalid per HTTP spec, but we treat it as single chunked encoding
+      // to avoid breaking compatibility with misconfigured clients/proxies
```
kentonv, line 2191: Perhaps we could add:

> Note that this does not create a risk for request smuggling because, in the worst case, if the sender actually did double-chunk the stream, we'll merely end up delivering a corrupted stream (the body will contain chunk framing). We would not end up misinterpreting the outer framing, so we won't desync.

danlapid: Sure, added in the last push.

PR #2392, opened 2025-08-20
Add addPtr/addPtrPtr/setPtr deleted variations

jasnell:
Just for additional safety

### DIFF
--- a/c++/src/kj/compat/http.h
+++ b/c++/src/kj/compat/http.h
@@ -337,6 +337,7 @@ public:
   KJ_DEPRECATED("Use setPtr()")
   void set(HttpHeaderId id, kj::StringPtr value);
   void setPtr(HttpHeaderId id, kj::StringPtr value);
+  void setPtr(HttpHeaderId id, kj::String&& value) = delete;
 
   void set(HttpHeaderId id, kj::String&& value);
   // Sets a header value, overwriting the existing value.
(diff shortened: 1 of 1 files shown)


c++/src/kj/compat/http.h
```diff
@@ -350,10 +351,15 @@ class HttpHeaders {
...
   void addPtrPtr(kj::StringPtr name, kj::StringPtr value);
+  void addPtrPtr(kj::StringPtr name, kj::String&& value) = delete;
+  void addPtrPtr(kj::String&& name, kj::StringPtr value) = delete;
+  void addPtrPtr(kj::String&& name, kj::String&& value) = delete;
 
   KJ_DEPRECATED("Use addPtr()")
   void add(kj::StringPtr name, kj::String&& value);
   void addPtr(kj::StringPtr name, kj::String&& value);
+  void addPtr(kj::StringPtr name, kj::StringPtr value) = delete;
+  void addPtr(kj::String&& name, kj::StringPtr value) = delete;
```
kentonv: I think you only need:

```suggestion
  void addPtr(kj::String&& name, kj::Sting&& value) = delete;
```

Since there's no overload that accepts `kj::StringPtr` for the second argument, an attempt to pass one would already fail.

kentonv: (And without this change I'm not sure if passing two `kj::String&&` would actually select the deleted overload...)

c++/src/kj/compat/http.h
```diff
@@ -350,10 +351,14 @@ class HttpHeaders {
...
   void add(kj::StringPtr name, kj::StringPtr value);
   void addPtrPtr(kj::StringPtr name, kj::StringPtr value);
+  void addPtrPtr(kj::StringPtr name, kj::String&& value) = delete;
+  void addPtrPtr(kj::String&& name, kj::StringPtr value) = delete;
+  void addPtrPtr(kj::String&& name, kj::String&& value) = delete;
 
   KJ_DEPRECATED("Use addPtr()")
   void add(kj::StringPtr name, kj::String&& value);
   void addPtr(kj::StringPtr name, kj::String&& value);
+  void addPtr(kj::String&& name, kj::Sting&& value) = delete;
```
jasnell:
```suggestion
  void addPtr(kj::String&& name, kj::String&& value) = delete;
```

PR #2393, opened 2025-08-21
Do not invoke tryRead() again in abortRead() if a previous call failed

npaun:
Some of our async streams keep track of a `clean` flag in `tryRead()`, and will fail unless `clean = true`.

For example, here is `HttpFixedLengthEntityReader::tryRead`:

```c++

Promise<size_t> tryRead(void* buffer, size_t minBytes, size_t maxBytes) override {
    KJ_REQUIRE(clean, "can't read more data after a previous read didn't complete");
    clean = false;

    size_t alreadyRead = 0;

    for (;;) {
      if (alreadyDone()) {
        clean = true;
        co_return alreadyRead;
      } else if (chunkSize == 0) {
        // ...
      } else {
        // ...
        auto amount = co_await getInner().tryRead(buffer, 1, kj::min(maxBytes, chunkSize));

        chunkSize -= amount;
        if (amount == 0) {
          // ...
          buffer = reinterpret_cast<byte*>(buffer) + amount;
          minBytes -= amount;
          maxBytes -= amount;
          alreadyRead += amount;
          continue;
        }
        clean = true;
        co_return alreadyRead + amount;
      }
    }
```

This prevents any concurrent calls to `tryRead()`. It also prevents any further calls to `tryRead()` if an exception was raised. 

As before, this doesn't work properly because `BlockedPumpFrom::abortRead()` does a final 1-byte `tryRead()` to see if we've reached EOF. This was fixed for the concurrent case in #2376.

This PR also deals with the exception case.

### DIFF
--- a/c++/src/kj/async-io.c++
+++ b/c++/src/kj/async-io.c++
@@ -776,8 +776,12 @@ private:
       auto pumpLeft = amount - pumpedSoFar;
       auto min = kj::min(pumpLeft, minBytes);
       auto max = kj::min(pumpLeft, maxBytes);
+
+      isReadOutstanding = true;
+
       return canceler.wrap(input.tryRead(readBuffer, min, max)
           .then([this,readBuffer,minBytes,maxBytes,min](size_t actual) -> kj::Promise<size_t> {
+        isReadOutstanding = false;
         canceler.release();
         pumpedSoFar += actual;
         KJ_ASSERT(pumpedSoFar <= amount);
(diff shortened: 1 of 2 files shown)


c++/src/kj/async-io.c++
```diff
@@ -776,8 +776,12 @@ private:
       auto pumpLeft = amount - pumpedSoFar;
       auto min = kj::min(pumpLeft, minBytes);
       auto max = kj::min(pumpLeft, maxBytes);
+
+      isReadOutstanding = true;
```
kentonv, line 780: `pumpTo` also counts as a read so we should probably put this there too.

c++/src/kj/async-io.c++
```diff
@@ -841,7 +845,7 @@ private:
     }
 
     void abortRead() override {
-      if (canceler.isEmpty()) {
+      if (canceler.isEmpty() && !isReadOutstanding) {
```
kentonv: The `canceler.isEmpty()` check is probably redundant now as `isReadOutstanding` would always be true if there's an uncanceled operations running.

npaun: Yep no longer needed.

kentonv (approved):

kentonv:
Tests failed because of a problem resolving zlib.net, re-running.

Accidentally clicked to re-run all of them instead of just the failing ones.

PR #2395, opened 2025-08-25
add use_stl_string_interop config setting

anonrig:
Adds a bazel config setting for use_stl_seting_interop


c++/src/kj/string.h
```diff
@@ -57,9 +57,11 @@ constexpr kj::LiteralStringConst operator ""_kjc(const char* str, size_t n);
 
 namespace kj {
 
-// Our STL string SFINAE trick does not work with GCC 4.7, but it works with Clang and GCC 4.8, so
-// we'll just preprocess it out if not supported.
-#if __clang__ || __GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 8) || _MSC_VER
+// Our STL string SFINAE trick does not work with GCC 4.7, but it works with
+// Clang and GCC 4.8, so we'll just preprocess it out if not supported.
```
kentonv: Did you re-wrap this comment for some reason?

c++/src/kj/configure.bzl
```diff
@@ -115,5 +125,8 @@ def kj_configure():
         }) + select({
             "//src/kj:use_kj_enable_irequire": ["KJ_ENABLE_IREQUIRE=1"],
             "//conditions:default": [],
+        }) + select({
+            "//src/kj:use_stl_string_interop": ["KJ_ENABLE_STL_STRING_INTEROP=1"],
+            "//conditions:default": ["KJ_ENABLE_STL_STRING_INTEROP=0"],
```
kentonv: Why should the default be "off"?

AFAICT this has been on by default since GCC 4.8 which is like a decade ago?

anonrig:
> What's the motivation for this change? When would you ever need to disable this?

For 2 different string types (rust::String and kj::StringPtr) that has both c_str() and size() methods implemented, this feature will create ambiguous function call compiler error. Since 2 different string types can implement those methods at the same time.

c++/src/kj/string.h
anonrig: This is just my IDE's formatting. Will revert.

c++/src/kj/configure.bzl
anonrig: Ah yeah, I think this line (line 130) should be `[]`

kentonv:
> For 2 different string types (rust::String and kj::StringPtr) that has both c_str() and size() methods implemented

`StringPtr` doesn't have `c_str()` (it has `cStr()`). Could you describe exactly the situation where there is ambiguity?

That said IMO we could just delete this interoperability if it's not needed anywhere in the workers runtime. It's kind of awkward... like, a StringPtr can implicitly convert to std::string, which means implicit allocation, which goes against KJ style.

anonrig:
@kentonv I agree. I can update the PR to remove this. I think it's a better approach.

kentonv (approved):
LGTM but I guess there are test failures.

kentonv:
(I'm still interested to know what situation was ambiguous before.)

anonrig:
Last commit is made by Claude Code. FYI.

c++/src/kj/string.h
```diff
@@ -610,10 +585,8 @@ struct Stringifier {
   CappedArray<char, 32> operator*(double f) const;
   CappedArray<char, sizeof(const void*) * 2 + 1> operator*(const void* s) const;
 
-#if KJ_COMPILER_SUPPORTS_STL_STRING_INTEROP  // supports expression SFINAE?
   template <typename T, typename Result = decltype(instance<T>().toString())>
```
fhanau, line 588: In the other changes we're removing the `KJ_COMPILER_SUPPORTS_STL_STRING_INTEROP` APIs, but in this case we are always making them available with the last commit? Please make sure to review AI-generated changes before adding them to a PR.

kentonv: Actually this is the right thing to do.

This case isn't actually providing any sort of STL interop. But it does use expression SFINAE to detect the existence of a `toString()` method, similar to how the STL iterop detects a `c_str()` method. Because of this, it had the same compiler requirements as the STL interop, and so it looks like I reused the KJ_COMPILER_SUPPORTS_STL_STRING_INTEROP define to guard it:

```
#if KJ_COMPILER_SUPPORTS_STL_STRING_INTEROP // supports expression SFINAE?
```

In any case, the compiler versions that didn't support this are long long gone so it's safe to remove the guard entirely now.

c++/src/kj/string-test.c++
```diff
@@ -497,6 +497,10 @@ struct OnlyMoves {
   bool operator==(const OnlyMoves&) const = default;
 };
 
+struct Stringable {
```
kentonv: We should also restore `TEST(String, ToString)` that was removed in the previous commit.

kentonv:
> Last commit is made by Claude Code. FYI.

This should be in the commit description.

Also: The commit description should actually explain what is going on in this change, what was the motivation, etc.

And please squash the commits so there's no commit in the history where the tests are broken.

c++/src/kj/string-test.c++
anonrig: Definitely

fhanau (approved):

anonrig:
I simplified and force pushed as @fhanau recommended in private.

PR #2396, opened 2025-08-28
more powerful as<T>() implementation

mikea:
Current design converts `as<T>()`` calls into static members of T, which doesn't make it extensible: T needs to know about every type that can be converted.

Updating the design to convert as<T>() in `into(T{}, this)`, which makes T a simple marker object and you can have as many conversion targets as you like.

This is a very breaking change (for implementation, but not for interface) and will require carefull landing across worker-cxx, workerd and workers repository.

See https://github.com/cloudflare/workerd-cxx/pull/67 for similar improvement for the other direction.
See https://github.com/cloudflare/workerd-cxx/pull/69 for downstream changes.

### DIFF
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -1985,13 +1985,13 @@ public:
   // You must include kj/array.h to call this.
 
   template <typename U>
-  inline auto as() { return U::from(this); }
-  // Syntax sugar for invoking U::from.
+  inline auto as() { return asImpl((U*)nullptr, *this); }
+  // Syntax sugar for invoking asImpl(U*, ArrayPtr&).
   // Used to chain conversion calls rather than wrap with function.
 
   template <typename U>
-  inline auto as() const { return U::from(this); }
-  // Syntax sugar for invoking U::from.
+  inline auto as() const { return asImpl((U*)nullptr, *this); }
+  // Syntax sugar for invoking asImpl(U*, const ArrayPtr&).
   // Used to chain conversion calls rather than wrap with function.
 
   inline void fill(T t) {
--- a/c++/src/kj/array-test.c++
+++ b/c++/src/kj/array-test.c++
@@ -611,12 +611,12 @@ TEST(Array, AttachFromArrayPtr) {
   KJ_EXPECT(destroyed1 == 3, destroyed1);
 }
 
-struct Std {
-  template<typename T>
-  static std::span<T> from(Array<T>* arr) {
-    return std::span<T>(arr->begin(), arr->size());
-  }
-};
+struct Std {};
+
+template<typename T>
+static std::span<T> asImpl(Std*, Array<T>& arr) {
+  return std::span<T>(arr.begin(), arr.size());
+}
 
 KJ_TEST("Array::as<Std>") {
   kj::Array<int> arr = kj::arr(1, 2, 4);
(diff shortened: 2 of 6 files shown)


anonrig (approved):

c++/src/kj/common.h
```diff
@@ -1985,12 +1985,12 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
   // You must include kj/array.h to call this.
 
   template <typename U>
-  inline auto as() { return U::from(this); }
-  // Syntax sugar for invoking U::from.
+  inline auto as() { return into(U{}, this); }
+  // Syntax sugar for invoking into(U, this).
```
kentonv: Nitpick: Why not `*this`? Can't be null so might as well be a reference?

c++/src/kj/common.h
```diff
@@ -1985,12 +1985,12 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
   // You must include kj/array.h to call this.
 
   template <typename U>
-  inline auto as() { return U::from(this); }
-  // Syntax sugar for invoking U::from.
+  inline auto as() { return into(U{}, this); }
```
kentonv: Bikesheds:
* Could the global function just be called `as`? Would probably make it clearer that these are connected.
* Instead of constructing an instance, you could pass a pointer, like `into((U*)nullptr, this)`.
* I guess the reason it's not `into<U>(this)` is because you want to use argument-dependent lookup, so `into` can be declared in the namespace of _either_ `U` or `this`? In theory we could say that it's always `kj::into<U>(this)`, and require people to declare overloads in the `kj` namespace, though that's a bit awkward I guess?

kentonv: (None of these are strong opinions, ignore me if you disagree.)

mikea: - I tried global `as` function but then function resolution breaks: no matter what I do as resolves as a recursive call with wrong parameter count, but maybe there's a syntax trick?
- done
- the motivation is to keep the set of transformable types open. E.g. imagine adding `as` method to WorkerInterface to assist conversion across ffi boundary. We don't want to update workerd-cxx `convert.h` where `Rust` marker structure is defined to know about WorkerInterface. (I am not suggesting adding .as to WorkerInterface but I am contemplating .as in io streams)

kentonv: > I tried global as function but then function resolution breaks: no matter what I do as resolves as a recursive call with wrong parameter count, but maybe there's a syntax trick?

Good point, but that might argue for something like `asImpl` which makes it a little clearer that these are related.

But I don't feel that strongly.

> the motivation is to keep the set of transformable types open. E.g. imagine adding as method to WorkerInterface to assist conversion across ffi boundary. We don't want to update workerd-cxx convert.h where Rust marker structure is defined to know about WorkerInterface. (I am not suggesting adding .as to WorkerInterface but I am contemplating .as in io streams)

Yes I get why you are changing it away from being a method, but I was talking about whether the new freestanding function could take the type as a template argument rather than the awkward pointer-to-type parameter.

But anyway I think it's probably better the way you have it.

mikea: > Good point, but that might argue for something like asImpl which makes it a little clearer that these are related.

asImpl it is.

> I was talking about whether the new freestanding function could take the type as a template argument rather than the awkward pointer-to-type parameter.

Ah, but that would require us to always deal function specialization and declare some weird generic `<typename U> asImpl(kj::String);`  as well as `<Rust>` specialization. Imo while `T*` is a bit ugly, it is a simpler implementation in the end.

PR #2399, opened 2025-08-31
Lazy Zero Memory Allocation

BrianXu0623:
This is still a draft, todo...


kentonv:
Please don't open draft PRs directly on other people's projects. There's no reason to do this -- the code isn't ready for review, and the CI won't run since you aren't a project member. So this just creates noise in the maintainers` inboxes.

Instead, you should start by posting a discussion or issue describing the problem you're facing.

If we decide that it makes sense to make a change to solve the problem, then we may suggest opening a PR. At that time you can post a PR, but it should be ready for review, not a draft.

PR #2404, opened 2025-09-05
make PipelineHook explicitly refcounted

mikea:
All  real implementations are.

This is similar to https://github.com/capnproto/capnproto/pull/1990


kentonv (approved):
I'm OK with this as long as the plan is to convert all the hooks. Let's not stay in a mixed state for long.

mikea:
> I'm OK with this as long as the plan is to convert all the hooks. Let's not stay in a mixed state for long.

ClientHooks are the only one that I don't have PR ready. Will do that soon too.

PR #2405, opened 2025-09-05
converting BrandScope to kj::Rc

mikea:
One more step in eliminating kj::addRef


c++/src/capnp/compiler/generics.h
```diff
@@ -152,11 +152,11 @@ class BrandedDecl {
 
 private:
   Resolver::ResolveResult body;
-  kj::Own<BrandScope> brand;  // null if parameter
+  kj::Rc<BrandScope> brand;  // null if parameter
   Expression::Reader source;
 };
 
-class BrandScope final: public kj::Refcounted {
+class BrandScope final: public kj::Refcounted, public kj::EnableAddRefToThis<BrandScope> {
```
kentonv: Seems like `EnableAddRefToThis` is needed pretty widely. Should we stop forcing people to inherit this wrapper? With C++23 "deducing this" you can get rid of the need for the CRTP I think?

PR #2407, opened 2025-09-08
HttpServer: passing HttpHeaders by value

mikea:
This change replaces pass-by-reference by pass-by-value in `HttpServer` `request` and `connect` method.

To make it easier to review it is broken into two commits:
- mechanical interface/declaration changes
- other fixes related to changed signature

To make pass-by-value possible this also changes all `tryParse*` methods in Headers to store an owned copy of headers data.

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -629,6 +629,7 @@ void HttpHeaders::clear() {
   }
 
   unindexedHeaders.clear();
+  ownedStrings.clear();
 }
 
 size_t HttpHeaders::size() const {
@@ -926,8 +927,10 @@ static kj::Maybe<kj::StringPtr> consumeHeaderName(char*& ptr) {
   return kj::StringPtr(start, end);
 }
 
-static char* trimHeaderEnding(kj::ArrayPtr<char> content) {
+static kj::ArrayPtr<char> trimHeaderEnding(kj::ArrayPtr<char> content) {
   // Trim off the trailing \r\n from a header blob.
+  // Returns ArrayPtr to the new end of the headers _including_ terminating 0.
+  // Returns empty ArrayPtr if the input is invalid.
 
   if (content.size() < 2) return nullptr;
 
@@ -939,7 +942,7 @@ static char* trimHeaderEnding(kj::ArrayPtr<char> content) {
   if (end[-1] == '\r') --end;
   *end = '\0';
 
-  return end;
+  return kj::ArrayPtr<char>(content.begin(), end + 1);
 }
 
 HttpHeaders::RequestOrProtocolError HttpHeaders::tryParseRequest(kj::ArrayPtr<char> content) {
@@ -960,13 +963,19 @@ HttpHeaders::RequestOrProtocolError HttpHeaders::tryParseRequest(kj::ArrayPtr<ch
 
 HttpHeaders::RequestConnectOrProtocolError HttpHeaders::tryParseRequestOrConnect(
     kj::ArrayPtr<char> content) {
-  char* end = trimHeaderEnding(content);
-  if (end == nullptr) {
+  auto headersContent = trimHeaderEnding(content);
+  if (headersContent == nullptr) {
     return ProtocolError { 400, "Bad Request",
         "Request headers have no terminal newline.", content };
   }
 
-  char* ptr = content.begin();
+  auto headersCopy = kj::heapArray(headersContent);
+  auto headersCopyPtr = headersCopy.asPtr();
+  ownedStrings.add(kj::mv(headersCopy));
+
+  char* ptr = headersCopyPtr.begin();
+  char* end = headersCopyPtr.end() - 1;
+  KJ_ASSERT(*end == 0);
 
   HttpHeaders::RequestConnectOrProtocolError result;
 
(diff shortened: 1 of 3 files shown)


c++/src/kj/compat/http.c++
```diff
@@ -960,13 +963,19 @@ HttpHeaders::RequestOrProtocolError HttpHeaders::tryParseRequest(kj::ArrayPtr<ch
...
-  char* end = trimHeaderEnding(content);
-  if (end == nullptr) {
+  auto headersContent = trimHeaderEnding(content);
+  if (headersContent == nullptr) {
     return ProtocolError { 400, "Bad Request",
         "Request headers have no terminal newline.", content };
   }
 
-  char* ptr = content.begin();
+  auto headersCopy = kj::heapArray(headersContent);
```
harrishancock, line 972: If we make a copy here, is there still a point in mutating `content` with `trimHeaderEnding()`? `trimHeaderEnding()` could take an `ArrayPtr<const char>` and return a nul-terminated `Array<char>` copy of it.

I see we use `content` down below in ProtocolError return values, so I guess we'd need to think about it.

c++/src/kj/compat/http.c++
```diff
@@ -1687,7 +1711,9 @@ public:
 
   // Used when suspending a request. HttpinputStream can no longer be used after this.
   kj::HttpHeaders releaseHeaders() {
-    return kj::mv(headers);
+    auto result = kj::mv(headers);
+    headers = HttpHeaders(table);
```
harrishancock, line 1715: I'm curious why this was needed? (Not saying change it.)

c++/src/kj/compat/http.c++
```diff
@@ -7488,7 +7514,37 @@ private:
 
   HttpInputStreamImpl httpInput;
   HttpOutputStream httpOutput;
-  kj::Maybe<kj::OneOf<HttpMethod, HttpConnectMethod>> currentMethod;
+
+  struct WebSocketRequest {
+    WebSocketRequest(const kj::HttpHeaders &requestHeaders)
```
harrishancock, line 7519: Silly style nit (here and elsewhere): `HttpHeaders&`

c++/src/kj/compat/http.h
```diff
@@ -1287,6 +1287,10 @@ class HttpServer final: private kj::TaskSet::ErrorHandler {
   class SuspendableRequest;
   typedef kj::Function<kj::Maybe<kj::Own<HttpService>>(SuspendableRequest&)>
       SuspendableHttpServiceFactory;
+  // Returns none or calls SuspendableRequest.suspend() if it suspends the request.
+  // The reference is valid until the factory returns.
+  // TODO(maybe) .suspend() is not needed and we can totally do it through returning none.
```
harrishancock, line 1292: After reading the code a bit more, I think maybe we can't do this yet. The application code needs to be able to obtain the SuspendedRequest somehow, and `.suspend()` is what returns that.

c++/src/kj/compat/http.c++
```diff
@@ -926,8 +927,10 @@ static kj::Maybe<kj::StringPtr> consumeHeaderName(char*& ptr) {
   return kj::StringPtr(start, end);
 }
 
-static char* trimHeaderEnding(kj::ArrayPtr<char> content) {
+static kj::ArrayPtr<char> trimHeaderEnding(kj::ArrayPtr<char> content) {
   // Trim off the trailing \r\n from a header blob.
+  // Returns ArrayPtr to the new end of the headers _including_ terminating 0.
+  // Returns empty ArrayPtr if the input is invalid.
```
kentonv, line 933: Perhaps this should return a `Maybe`?

c++/src/kj/compat/http.c++
kentonv: It looks like this commit repurposes `releaseHeaders()` to be the way headers are passed off to the app, so new headers need to be allocated for the next request. `releaseHeaders()` is no longer used for request suspension, which instead simply `kj::mv`s the headers directly.

The doc comment needs at update.

c++/src/kj/compat/http.c++
```diff
@@ -6739,7 +6766,7 @@ public:
     // assume that they remain valid until the service handler completes whereas HttpClient callers
     // are allowed to destroy them immediately after the call.
     auto urlCopy = kj::str(url);
-    auto headersCopy = kj::heap(headers.clone());
+    auto headersCopy = headers.clone();
```
kentonv, line 6769: I think we should also update the `HttpClient` interface to pass headers by move, so we can eliminate this.

Note that every subrequest from a worker goes through this code path so we'd probably save a lot by avoiding the copy.

c++/src/kj/compat/http.c++
```diff
@@ -7488,7 +7514,37 @@ private:
 
   HttpInputStreamImpl httpInput;
   HttpOutputStream httpOutput;
-  kj::Maybe<kj::OneOf<HttpMethod, HttpConnectMethod>> currentMethod;
+
+  struct WebSocketRequest {
```
kentonv, line 7518: Please add some comments explaining what this is.

c++/src/kj/compat/http.c++
```diff
@@ -7488,7 +7514,37 @@ private:
...
+      key(requestHeaders.get(HttpHeaderId::SEC_WEBSOCKET_KEY).map([](auto&& key){return kj::str(key);})),
+      extensions(requestHeaders.get(HttpHeaderId::SEC_WEBSOCKET_EXTENSIONS).map([](auto&& extensions){return kj::str(extensions);}))
+      {}
+    
+    bool supportedVersion;
+    kj::Maybe<kj::String> key;
+    kj::Maybe<kj::String> extensions;
+  };
+
+  struct CurrentRequest {
```
kentonv, line 7530: Comments please.

kentonv (commented):
Along with this we probably need to remove the `setPtr()`/`addPtr()` methods from `HttpHeaders`.

As an alternative, we can add versions that accept `LiteralStringConst`, so allocation isn't necessary when it's actually a string literal.

But anything that is not a string literal can no longer reasonably be unowned, since it's no longer possible for a caller to determine how long the `HttpHeaders` object might live.

kentonv:
> As an alternative, we can add versions that accept LiteralStringConst, so allocation isn't necessary when it's actually a string literal.

Actually I guess what we can do is change all the `String`s to `ConstString`s?

PR #2408, opened 2025-09-10
Include relevant variable values in out-of-bounds checks

a-robinson:
So that if any of these do fire, it's easier to get a sense of what's going on.


c++/src/kj/common.c++
```diff
@@ -37,6 +37,27 @@ void inlineRequireFailure(const char* file, int line, const char* expectation,
   }
 }
 
+void inlineRequireFailure(const char* file, int line, const char* expectation,
+                          const char* macroArgs, const char* message, size_t arg1) {
+  Debug::Fault f(file, line, kj::Exception::Type::FAILED, expectation, macroArgs, message, arg1);
```
cacheburner: We don't need to check if `message == nullptr` in these cases?

a-robinson: If you don't provide anything after the expectation, you'll get the original version of `inlineRequireFailure` that has a default `nullptr` value for the `message` parameter and that does the safety check.

If you do provide something after the expectation, it'll be type-checked against the need to provide a `const char*` and one or more `size_t` parameters. So I guess if you passed something that could be implicitly coerced to a `const char*` as the first thing after the expectation then you could get in trouble, but otherwise you'll get a compilation error.

So I guess I should add it in out of paranoia? Although if you're already passing random junk in as the pointer then who knows what data we'll end up printing here.

a-robinson: It feels very unnecessary, but it's hard to justify not having a null check, so I've added it in.

Issue #1448, opened 2022-04-05
Raspbian build fails due to need for libatomic


snej:
I'm building CapnP 0.9.1 from source on a new Raspberry Pi 4 running Raspbian (because apt only has version 0.7.) 

```
$ curl -O https://capnproto.org/capnproto-c++-0.9.1.tar.gz
$ tar zxf capnproto-c++-0.9.1.tar.gz
$ cd capnproto-c++-0.9.1
$ ./configure
$ make -j6 check
$ sudo make install
```

Everything went fine until the final "sudo make install" step:

```
...
libtool: link: g++ -I./src -I./src -DKJ_HEADER_WARNINGS -DCAPNP_HEADER_WARNINGS -DCAPNP_INCLUDE_DIR=\"/usr/local/include\" -pthread -O2 -DNDEBUG -pthread -DKJ_HAS_ZLIB -pthread -o .libs/capnp src/capnp/compiler/module-loader.o src/capnp/compiler/capnp.o  ./.libs/libcapnpc.so ./.libs/libcapnp-json.so ./.libs/libcapnp.so ./.libs/libkj.so -lpthread -pthread
/usr/bin/ld: ./.libs/libcapnp.so: undefined reference to `__atomic_load_8'
/usr/bin/ld: ./.libs/libcapnp.so: undefined reference to `__atomic_store_8’
```

After some googling I found that the fix is to link with libatomic -- I simply changed `LIBS` in the Makefile (`LIBS =  -lpthread -latomic`), but I'm aware the Makefile is generated by autoconf; I'm not sure where the list of libraries actually comes from.

kentonv:
`Makefile` is generated from `Makefile.am`. However, what we probably want here is a `configure` check (which would be defined in `configure.ac`) that checks whether it's necessary to link with libatomic for 64-bit atomics, and adds the library if so.

I'm a bit surprised that these intrinsics aren't automatically added by the compiler, though.

I also wonder if maybe we want to adjust the code to avoid 64-bit atomics on platforms that don't have support for this in hardware... doing the atomic in software is going to be very slow. Maybe we should review where we're using such atomics and see if we can switch to 32-bit numbers on these platforms...

ryanofsky:
> However, what we probably want here is a `configure` check

Debian seems to have patched this in: https://sources.debian.org/patches/capnproto/0.9.1-4/07_libatomic.patch/

kentonv:
[...]

PR #2410, opened 2025-09-10
Check if libatomic is needed

NyCodeGHG:
Should fix https://github.com/capnproto/capnproto/issues/1448
I already struggled with CMake, I have no idea how to do this with autoconf, sorry :/

### DIFF
--- a/c++/src/kj/CMakeLists.txt
+++ b/c++/src/kj/CMakeLists.txt
@@ -84,6 +84,21 @@ target_compile_features(kj PUBLIC cxx_std_20)
 if(UNIX AND NOT ANDROID)
   target_link_libraries(kj PUBLIC pthread)
 endif()
+
+include(CheckCXXSourceCompiles)
+check_cxx_source_compiles("#include <atomic>
+int main() {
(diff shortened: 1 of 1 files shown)


jamessan:
If I read the discussion in #1448 right, the [autoconf portion](https://github.com/thomaslee/capnproto-debian/blob/87c3340f67aac4b8979ed6730cb99ceddb0982cc/debian/patches/07_libatomic.patch#L6-L37) of Debian's patch was fine, so you could use that. It was the CMake part that was broken.

kentonv:
Hmm, reviewing my comment here:

https://github.com/capnproto/capnproto/issues/1448#issuecomment-1148920666

I think this applies to both the autoconf and cmake parts of the Debian patch. It looks to me like in both cases libatomic will be linked if the library exists at all, not just if it is needed. But I could be misunderstanding.

kentonv:
(Sorry for taking forever to reply to this, it got lost in my inbox...)

harrishancock (approved):
The CMake looks reasonable to me.

kentonv:
I had Claude write an autoconf check. #2466

PR #2411, opened 2025-09-11
debug.h: Add KJ_REQUIRE_AT(), etc. variants, to allow overriding the source locations of failures

jclee:
Pulling in some utility functions that have been useful elsewhere, e.g.: https://github.com/cloudflare/workerd/blob/7e2bdefa/src/workerd/util/capnp-mock.h#L22

### DIFF
--- a/c++/src/kj/debug-test.c++
+++ b/c++/src/kj/debug-test.c++
@@ -25,6 +25,7 @@
 
 #include "debug.h"
 #include "exception.h"
+#include "source-location.h"
 #include <kj/compat/gtest.h>
 #include <string>
 #include <stdio.h>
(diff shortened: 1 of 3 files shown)


kentonv:
I wonder if we could reduce the duplication here by changing the signature of `::kj::_::Debug::Fault` itself to take `SourceLocation`, so we don't have to pass `__FILE__` and `__LINE__` to it?

jclee:
(Added TODO(cleanup) comment and rebased)

> I wonder if we could reduce the duplication here by changing the signature of ::kj::_::Debug::Fault itself to take SourceLocation, so we don't have to pass __FILE__ and __LINE__ to it?

Possibly?  I wasn't totally sure that SourceLocation would have the same overhead, but looking closer, I guess it just builds a struct of four numbers (ints/string pointers), and hopefully the compiler would be able to work out that only a few fields were used.

But I'm also not totally sure that the builtin functions have exactly the same meaning as the macros in all contexts.  In the debug test for the _AT members, for instance, the reported line number for the _AT macro is actually one greater than the reported line number for the regular macro -- which is why I had to write `line = SourceLocation().lineNumber;` instead of using the `line = __LINE__;` seen in the other tests.  I assume this is because of some quirk of macro expansion order, but don't know for sure.

(Of course, writing the test that way also has the benefit of smoothly handling the cases when KJ_COMPILER_SUPPORTS_SOURCE_LOCATION might be 0 -- but not sure if that's still important anywhere.)

jclee:
Oh... I guess I can't merge this quite yet.  We have a lint pass on the internal build that fails due to macro redefinition warnings in the places we're already using these macros.  I guess I'll need to wrap those with "#ifndef"s first.

kentonv:
> I guess I'll need to wrap those with "#ifndef"s first.

That or prepare a simultaneous update PR set.

kentonv:
> the reported line number for the _AT macro is actually one greater than the reported line number for the regular macro -- which is why I had to write line = SourceLocation().lineNumber; instead of using the line = __LINE__; seen in the other tests

That's weird! Which one is reporting the correct line and which one is wrong? Do we need to add/subtract 1 to make them correct?

kentonv:
> cases when KJ_COMPILER_SUPPORTS_SOURCE_LOCATION might be 0 -- but not sure if that's still important anywhere.)

No I don't think that's a concern anymore.

jclee:
> That's weird! Which one is reporting the correct line and which one is wrong? Do we need to add/subtract 1 to make them correct?

The `__LINE__` values correctly match their (one-based) line numbers, while the `__builtin_LINE()` values match the line number of the end of the test case lambda.

After some experimentation, I found this is apparently due to the fact that the individual test cases are evaluated using the `EXPECT_LOG_EQ()` macro.  When the preprocessor runs, it rewrites the macro call such that the expected `__LINE__` values are substituted, but it also apparently concatenates the expanded code to a single line.  And when the `__builtin_LINE()` expressions are evaluated, the compiler acts as if that single line is the same as the end of the macro invocation.

It looks like both latest clang and gcc do this, but clang uses the line number at the end of the macro invocation, while gcc picks the beginning?: https://godbolt.org/z/P8o3WEaT6

I guess maybe it would be better if I rewrote the test to evaluate the cases without using a macro, so that the line numbers are accurate.  But it probably also means that `__LINE__` and `__builtin_LINE()` will behave somewhat differently within macros, so `__builtin_LINE()` might report less helpful line numbers in `KJ_IF_SOME()`, `KJ_DEFER()`, etc.

(Oh... actually, `KJ_IF_SOME()` should be generally OK, since the macro isn't taking the code body as a parameter, so it's not part of the macro expansion.)

jclee:
Updated test to use functions instead of macros, such that the line numbers are accurate -- although C++'s verbose lambdas make the test a little less pretty.  Also added a note to SourceLocation to document its inaccuracy within macro calls.

A little further experimentation shows that clang/gcc/msvc all support "std::source_location" in C++20 but not by default, and that clang/msvc both support `__builtin_COLUMN()`, but gcc still does not: https://godbolt.org/z/j5v676TcP

PR #2414, opened 2025-09-18
better support of forward declared classes in Rc and Arc

mikea:
I wasn't able to build a reproducer but empirically if I replace kj::Own<ClientHook> with kj::Rc<ClientHook> then layout.h stops compiling because ClientHook is foward declared.

For some reason this fixes forward compilation.
Any ideas why?

### DIFF
--- a/c++/src/kj/refcount.h
+++ b/c++/src/kj/refcount.h
@@ -174,6 +174,8 @@ public:
   template <typename U, typename = EnableIf<canConvert<U*, T*>()>>
   inline Rc(Rc<U>&& other) noexcept : own(kj::mv(other.own)) { }
 
+  inline ~Rc() { own = nullptr; }
+
   kj::Own<T> toOwn() {
     // Convert Rc<T> to Own<T>.
     // Nullifies the original Rc<T>.
@@ -428,6 +430,8 @@ public:
 
   template <typename U, typename = EnableIf<canConvert<U*, T*>()>>
   inline Arc(Arc<U>&& other) noexcept : own(kj::mv(other.own)) { }
+
+  inline ~Arc() { own = nullptr; }
 
   kj::Own<const T> toOwn() {
     // Convert Arc<T> to Own<const T>.


c++/src/kj/refcount.h
```diff
@@ -174,6 +174,8 @@ class Rc {
   template <typename U, typename = EnableIf<canConvert<U*, T*>()>>
   inline Rc(Rc<U>&& other) noexcept : own(kj::mv(other.own)) { }
 
+  inline ~Rc() { own = nullptr; }
```
kentonv, line 177: It doesn't make sense to me that this would make a difference, given that both `~Own()` and `Own::operator=(nullptr_t)` just call `dispose()`.

What exactly is the error you get otherwise?

mikea:
```
Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
In file included from src/capnp/stream.capnp.c++:4:
In file included from src/capnp/stream.capnp.h:6:
In file included from bazel-out/k8-opt-exec-ST-a828a81199fe/bin/src/capnp/_virtual_includes/capnp/capnp/generated-header-support.h:26:
In file included from bazel-out/k8-opt-exec-ST-a828a81199fe/bin/src/capnp/_virtual_includes/capnp/capnp/raw-schema.h:24:
In file included from bazel-out/k8-opt-exec-ST-a828a81199fe/bin/src/capnp/_virtual_includes/capnp/capnp/common.h:29:
In file included from bazel-out/k8-opt-exec-ST-a828a81199fe/bin/src/kj/_virtual_includes/kj/kj/string.h:25:
In file included from bazel-out/k8-opt-exec-ST-a828a81199fe/bin/src/kj/_virtual_includes/kj/kj/array.h:24:
bazel-out/k8-opt-exec-ST-a828a81199fe/bin/src/kj/_virtual_includes/kj/kj/memory.h:66:10: error: incomplete type 'capnp::ClientHook' used in type trait expression
   66 |   return __is_polymorphic(T);
      |          ^
bazel-out/k8-opt-exec-ST-a828a81199fe/bin/src/kj/_virtual_includes/kj/kj/memory.h:939:17: note: in instantiation of function template specialization 'kj::_kj_internal_isPolymorphic<capnp::ClientHook>' requested here
  939 |   if constexpr (_kj_internal_isPolymorphic((T*)nullptr)) {
      |                 ^
bazel-out/k8-opt-exec-ST-a828a81199fe/bin/src/kj/_virtual_includes/kj/kj/memory.h:365:17: note: in instantiation of function template specialization 'kj::Disposer::dispose<capnp::ClientHook>' requested here
  365 |       disposer->dispose(const_cast<RemoveConst<T>*>(ptrCopy));
      |                 ^
bazel-out/k8-opt-exec-ST-a828a81199fe/bin/src/kj/_virtual_includes/kj/kj/memory.h:267:28: note: in instantiation of member function 'kj::Own<capnp::ClientHook>::dispose' requested here
  267 |   ~Own() noexcept(false) { dispose(); }
      |                            ^
bazel-out/k8-opt-exec-ST-a828a81199fe/bin/src/kj/_virtual_includes/kj/kj/refcount.h:149:7: note: in instantiation of member function 'kj::Own<capnp::ClientHook>::~Own' requested here
  149 | class Rc {
      |       ^
bazel-out/k8-opt-exec-ST-a828a81199fe/bin/src/capnp/_virtual_includes/capnp/capnp/layout.h:61:7: note: forward declaration of 'capnp::ClientHook'
   61 | class ClientHook;
      |       ^
```

I am actually not sure why/how it works now, since layout.h uses `kj::Own<ClientHook>` a lot but does not define it.

If I apply this change to the Rc the forward declaration problem goes away.

kentonv: Hmm, this still doesn't make sense and I'm worried that there's actually some bug here where it _should_ be failing to compile, but isn't. Maybe the `__is_polymorphic` check is improperly returning `false` instead of failing. That could lead to UB.

OTOH, while `layout.h` does mention `Own<ClientHook>` a lot, it's strictly in params or return values of functions that are not inline. It's supposed to be OK to use incomplete types there. The destructor should not be instantiated by any of these uses, since none of them generate any code at all (just declare symbols defined elsewhere).

I wonder if what's making the difference here is just that you're declaring a destructor at all? Maybe for some reason the compiler is eagerly generating the implicit destructor, but wouldn't eagerly generate an explicit one? What if you just do:

```suggestion
  inline ~Rc() noexcept(false) {}
```

PR #2415, opened 2025-09-18
eliminate virtual call in kj::Rc disposer

mikea:
If this works well I'll convert Arc too

### DIFF
--- a/c++/src/kj/refcount.h
+++ b/c++/src/kj/refcount.h
@@ -102,6 +102,9 @@ private:
 
   template <typename T>
   friend class Rc;
+
+  template <typename T>
+  friend struct RcDisposer;
 };
 
 template <typename T, typename... Params>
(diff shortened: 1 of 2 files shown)


mikea:
here is ~MembraneCallContextHook from https://github.com/capnproto/capnproto/pull/1990 https://gist.github.com/mikea/32ffb5fe14e0275d315e1306affc960b

As you can see virtual Own dispose calls are eliminated and only virtual destructor calls remain.

c++/src/kj/refcount.h
```diff
@@ -145,6 +148,14 @@ Rc<T> Refcounted::addRcRefInternal(T* object) {
   return Rc<T>(object);
 }
 
+template<typename T>
+struct RcDisposer {
+  inline static void dispose(void* value) {
```
kentonv, line 153: Static disposers don't need to deal with `void*`.

You can make this:

```suggestion
  inline static void dispose(kj::Refcounted* value) {
```

And then remove the `reinterpret_cast`, and also remove the template argument from `RcDisposer`.

This also lets you use `kj::mv()` when upcasting, so you don't have to deal with `disown`.

PR #2334, opened 2025-06-20
PtmfHelper: Use NOLINT to disable UndefinedBinaryOperatorResult error

ryanofsky:
Disable `clang-analyzer-core.UndefinedBinaryOperatorResult` error in `PtmfHelper` using a NOLINT comment to appease clang-tidy.

A complete description of the error and rationale for the change is in the commit message.

PR #2417, opened 2025-09-23
PromiseDisposer: Use NOLINT to disable ArrayBound warning

ryanofsky:
Disable clang-analyzer-security.ArrayBound warning in [PromiseDisposer::append](https://github.com/capnproto/capnproto/blob/6846dffe3a1e5cff164f868e2999d0d3cfad4bb2/c%2B%2B/src/kj/async-inl.h#L404) to silence clang-tidy.

More information about the warning can be found in the commit message. A similar change was also made recently in https://github.com/capnproto/capnproto/pull/2334.

### DIFF
--- a/c++/src/kj/async-inl.h
+++ b/c++/src/kj/async-inl.h
@@ -401,6 +401,7 @@ public:
       //   code bloat to handle this case.
       next->arena = nullptr;
       T* ptr = reinterpret_cast<T*>(next.get()) - 1;
+      // NOLINTNEXTLINE(clang-analyzer-security.ArrayBound)
       ctor(*ptr, kj::mv(next), kj::fwd<Params>(params)...);
       ptr->arena = arena;
       KJ_IREQUIRE(reinterpret_cast<void*>(ptr) ==


ryanofsky:
In addition to the warning, it's also possible there may be undefined behavior on the [previous line](https://github.com/capnproto/capnproto/blob/6846dffe3a1e5cff164f868e2999d0d3cfad4bb2/c%2B%2B/src/kj/async-inl.h#L403):

```
reinterpret_cast<T*>(next.get()) - 1
```

which ChatGPT suggested changing to something like:

```
reinterpret_cast<T*>(reinterpret_cast<unsigned char*>(next.get()) - sizeof(T))
```

to avoid undefined behavior described https://eel.is/c++draft/expr.add#6 where pointer arithmetic is done with a type not matching the underlying array type (which is [`byte[1024]`](https://github.com/capnproto/capnproto/blob/6846dffe3a1e5cff164f868e2999d0d3cfad4bb2/c%2B%2B/src/kj/async-inl.h#L143)). 

But I'm not a language lawyer and am not sure how to evaluate this.

I also don't know if changing this would affect the ArrayBound warning, so it seemed simplest just to deal with the warning directly.

fhanau (approved):

kentonv:
Sorry for delay. Took me several reads to understand what's going on here, but yeah I guess that makes sense. Thanks!

ryanofsky:
Thanks! Sorry, comments could have used some editing.

To be clear, this PR fixes the ArrayBound warning on line [404](https://github.com/capnproto/capnproto/blob/6846dffe3a1e5cff164f868e2999d0d3cfad4bb2/c%2B%2B/src/kj/async-inl.h#L404) but there may still be undefined behavior on line [403](https://github.com/capnproto/capnproto/blob/6846dffe3a1e5cff164f868e2999d0d3cfad4bb2/c%2B%2B/src/kj/async-inl.h#L403) because the result of the subtraction is undefined.

According to https://eel.is/c++draft/expr.add#6, pointer subtraction is defined if the pointer type and underlying array type are compatible, and in the case the pointer type is `ChainPromiseNode` and array type is `byte`, so it doesn't seem like that is the case. Realistically, it's hard to see how this could be a problem, but I wanted to point it out in case it wasn't intended.

kentonv:
Yeah I don't see how anyone can write a memory allocator without violating that rule...

PR #2422, opened 2025-09-29
Backport commits until 09/28/25 from master

fhanau:
I was interested in the two commits silencing clang static analyzer false positives, but backporting other changes too didn't look overly difficult.

Commits requiring manual changes:
7758c8fc Expose socketpair creation in async-io: This required fixing merge conflicts, **merits close attention in review**.
ec4312f8 Set mainline version to 1.2-dev.: We don't need the version increment itself, but I added a commit for the bump-minor command added here.
12b25e38 Fix release builds on Ubuntu 22.04 action runners.: Backported the no-fiber test setting from this.

Skipped commits:
26adf72a Set mainline version to 1.3-dev.: Self-explanatory
68c07beb Backport #2246 (O_PATH fallback) to 1.x release branch.: Self-explanatory
55a8399c Update 1.x release branch CI to Ubuntu 22.04.: already have 22.04
b72e686b Add `MODULE.bazel` for bzlmod: Excluded for now, when we tried bzlmod on v2 (#2370) it caused CI failures in downstream repos.
8e6ed94a Update c++/src/kj/CMakeLists.txt: sets Cmake to C++17, we already require 20
67591a87 Require CMake 3.10.: we already require 3.16 on v2, I verified that the change does not include other relevant changes
1dc2e607 Minor typo in filesystem-disk-unix.c++: Already covered in prior typo cleanup
b3d4387f ci: Add FreeBSD job to Quick Tests: I think this is nice to have, but setting this up with clang-18 might be difficult. Hopefully we can do a larger CI configuration cleanup when v2 gets released. 
372eda67 ci: Fix "Linux (clang-14)" job: We're beyond clang-14 already

### DIFF
--- a/super-test.sh
+++ b/super-test.sh
@@ -28,6 +28,7 @@ QUICK=
 CPP_FEATURES=
 WERROR="-Werror"
 EXTRA_LIBS=
+CONFIGURE_FLAGS=
 
 PARALLEL=$(nproc 2>/dev/null || echo 1)
 
@@ -75,6 +76,9 @@ while [ $# -gt 0 ]; do
       # Re-run preventing sleep.
       shift
       exec caffeinate -ims $0 $@
+      ;;
+    no-fiber )
+      CONFIGURE_FLAGS="$CONFIGURE_FLAGS --without-fibers"
       ;;
     tmpdir )
       # Clone to a temp directory.
@@ -130,6 +134,7 @@ while [ $# -gt 0 ]; do
       export CC=clang${1#clang}
       export CXX=clang++${1#clang}
       export LIB_FUZZING_ENGINE=-fsanitize=fuzzer
+      $CXX -v
       ;;
     gcc* )
       export CXX=g++${1#gcc}
@@ -150,7 +155,7 @@ while [ $# -gt 0 ]; do
 
       export WINEPATH='Z:\usr\'"$CROSS_HOST"'\lib;Z:\usr\lib\gcc\'"$CROSS_HOST"'\6.3-win32;Z:'"$PWD"'\.libs'
 
-      doit ./configure --host="$CROSS_HOST" --disable-shared CXXFLAGS="-static-libgcc -static-libstdc++ $CPP_FEATURES" LIBS="$EXTRA_LIBS"
+      doit ./configure --host="$CROSS_HOST" --disable-shared CXXFLAGS="-static-libgcc -static-libstdc++ $CPP_FEATURES" LIBS="$EXTRA_LIBS" $CONFIGURE_FLAGS
 
       doit make -j$PARALLEL check
       doit make distclean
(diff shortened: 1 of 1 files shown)


c++/src/kj/async-io-unix.c++
```diff
@@ -2018,24 +2008,15 @@ public:
...
-    auto pipe = lowLevel.wrapSocketFd(fds[0], NEW_FD_FLAGS);
+    auto socketpair = newOsSocketpair();
 
-    auto thread = heap<Thread>([threadFd,startFunc=kj::mv(startFunc)]() mutable {
+    auto pipe = lowLevel.wrapSocketFd(kj::mv(socketpair.fds[0]), NEW_FD_FLAGS);
+    auto thread = heap<Thread>([threadFd=kj::mv(socketpair.fds[1]),startFunc=kj::mv(startFunc)]() mutable {
       UnixEventPort eventPort;
       EventLoop eventLoop(eventPort);
       WaitScope waitScope(eventLoop);
       LowLevelAsyncIoProviderImpl lowLevel(eventPort);
```
fhanau: The original commit uses `LowLevelAsyncIoProviderImpl lowLevelImpl; LowLevelAsyncIoProvider& lowLevel = lowLevelImpl;` here, I did not understand what that was for so I omitted it for now.

kentonv:
Hmm I thought I already cherry-picked everything up to 1.2.0 (June 14) in #2329, but some of these commits look older than that?

kentonv:
LGTM but it appears the tests are failing, looks related to the socketpair change?

fhanau:
> LGTM but it appears the tests are failing, looks related to the socketpair change?

I think what's going wrong is that in `newPipeThread()` we don't manage the lifetime of `socketpair.fds[1]` properly: With the patch we're making this an OwnedFd, and if we prevent it from being freed based on that using `lowLevel.wrapSocketFd(kj::mv(threadFd), NEW_FD_FLAGS);` tests pass again.
However, that is clearly not the right behavior either. It looks like if we use the lowLevelImpl trick mentioned in the other comment that also gets tests to pass somehow? Updated the PR with that, I'm assuming that approach is correct.

kentonv:
> It looks like if we use the lowLevelImpl trick mentioned in the other comment that also gets tests to pass somehow?

Can't see how that would possibly have an effect TBH.

kentonv:
Oh I see why!

`LowLevelAsyncIoProvider` contains a bunch of methods that have overloads for `int` vs. `OwnFd`, where only the former are virtual.

`LowLevelAsyncIoProviderImpl` implements the virtual methods, which hides the non-virtual ones.

Without that, the `kj::mv()` is having no effect, because the `OwnFd` is being implicitly converted to `int` anyway.

kentonv (approved):

Issue #2400, opened 2025-09-01
Skip zeroing for guaranteed-overwritten data in SegmentAllocator


BrianXu0623:
### Background
I'm implementing a custom` MessageBuilder::SegmentAllocator` function to allocate segments within shared memory and populate them in place. And I'm transferring [AVFrame](https://ffmpeg.org/doxygen/2.7/structAVFrame.html?utm_source=chatgpt.com) as a Cap’n Proto message, which consists of roughly 99.9% video frame blob data (`schema::type::DATA`) and about 0.1% other metadata fields.

I found that Cap’n Proto requires developers to zero-initialize the allocated memory before returning the segment to the arena. However, zeroing a 5.93 MB 1080P AVFrame takes around 29.5 µs, while zeroing only the 5 KB of metadata fields takes about 0.7 µs (measured on my local device).

In my use case, zeroing the video frame data is not necessary because I can guarantee that it will be overwritten later. I think this is a common use case in multimedia processing, and possibly in other scenarios as well, where the message is primarily composed of a large Data or List field that is guaranteed to be overwritten.

### Question
Does Cap’n Proto provide options for users, for skipping zeroing of non-essential memory regions? I understand that this could be achieved if users carefully control the memory layout before allocation, but that would require a complete understanding of Cap’n Proto’s wire format. It would be ideal if there's other simple and flexible ways to achieve this.

I previously did a [patch](https://github.com/capnproto/capnproto/pull/2423) to implement a LazyZero approach, which allows developer to choose not to zero out memory during the AllocateSegment stage. Instead, the necessary segment space is lazily zeroed as the message is being constructed, while certain types/fields can be skipped from zeroing.

I’d like to know if there are already better approaches supported, or if you have any suggestions. Thank you!

kentonv:
This might be a good use case for `Orphanage::referenceExternalData()`. With that, you can take a buffer allocated completely separately and make it part of your message, without having to copy it at all. At the very least, this would allow you to do allocation in a different way where you don't have to pre-zero the buffer. But, depending on how your code is structured, it might allow you to avoid a whole copy, which would be an even bigger win. With multi-megabyte buffers this sort of optimization is worthwhile.

Does that work? If so, great! If not, we can consider a patch to do the zeroing lazily, but it may take some care to get right.

BrianXu0623:
Hello author,
I’d like to bring up this issue again.

[...]

PR #2423, opened 2025-09-30
Lazy zero segment allocation strategy

BrianXu0623:
This PR is for resolving the following issue:
[Skip zeroing for guaranteed-overwritten data in SegmentAllocator #2400
](https://github.com/capnproto/capnproto/issues/2400)

## Summary
Introduce an opt-in `BuilderOptions::LazyZeroSegmentAlloc` that lets callers/allocators avoid pre-zeroing entire segments. By default Cap’n Proto requires allocators (including overridden `allocateSegment`) to return zero-initialized memory. When a caller opts in via `BuilderOptions::LazyZeroSegmentAlloc` **and** uses a custom/overridden `allocateSegment`, the allocator may return an unzeroed segment and Cap’n Proto takes responsibility for zeroing during message construction: it will lazily zero the parts that must be zeroed and skip the regions explicitly configured to be safe to leave uninitialized. This moves the zeroing responsibility from the allocator to the builder in an explicit, opt-in way and reduces unnecessary `memset` work for very large buffers (e.g., big `DATA` blobs).

**Important:** A message builder with custom/overridden `allocateSegment` must be used together with `BuilderOptions::LazyZeroSegmentAlloc`. When opted-in, the allocator no longer needs to call `memset(0)` on the whole segment — Cap’n Proto will perform lazy zeroing and will skip configured areas.

## Key changes
- Added `BuilderOptions` with a nested `LazyZeroSegmentAlloc` struct to control lazy zeroing.
- Introduced `SegmentBuilder::doLazyZeroSegment()` to handle conditional zeroing based on type and options.
- Modified memory allocation paths in `BuilderArena` and `WireHelpers` to use the new lazy zeroing logic.
- Added validation logic to ensure only supported types are skipped for zeroing.
- Exposed new `MessageBuilder` methods to set and get builder options.
- **Future extension:** extend to other field types (large lists, text, nested containers) or to per-field control (by `fieldId` / `fieldName`) so callers can mark specific fields to skip lazy-zero.

## Example
```cpp
// Example usage of a custom `MessageBuilder` that overrides the `allocateSegment` method,
// and enables lazy zero segment allocation for DATA type.
BuilderOptions options;
options.lazyZeroSegmentAlloc.enableLazyZero = true;
options.lazyZeroSegmentAlloc.skipLazyZeroTypes.insert(schema::Type::DATA);
MyCustomMessageBuilder builder(options);
```

## Verification & testing
In a media-processing validation using an [AVFrame](https://ffmpeg.org/doxygen/2.7/structAVFrame.html)-style Cap’n Proto message, skipping pre-zero for a 1080p frame raw-bytes DATA field (~5 MB) reduced message-construction zeroing cost from ~20 µs to ~0.4 µs in our measurement.

## Safety
- Users must ensure correctness for lazily-zeroed fields: if a lazily-skipped region is read before being fully overwritten, the read may return random/dirty (uninitialized) memory — avoid reading skipped regions until you have fully written them.


PeakKS:
I know it doesn't match the current definition of `schema::Type::*`, but it would be better to have the zero selectors as bit flags instead of a `std::set`. Assuming it doesn't blow everything up redefining that enum as bit flags would be much more efficient than allocating skipLazyZeroTypes on the heap and having to do a lookup every time you want to check.

PeakKS:
Also somewhat related to this issue it would be nice to have a way to shrink the size of a data blob after initialization in case you didn't end up writing to the whole thing before you want to send it, removing the need for a additional field to specify the "real" size.

c++/src/capnp/message.h
```diff
@@ -145,6 +147,61 @@ class MessageReader {
...
+    //   options.lazyZeroSegmentAlloc.skipLazyZeroTypes.insert(schema::Type::DATA);
+    //   MyCustomMessageBuilder builder(options);  // Must override allocateSegment
+    //
+    // Supported types for skipping zeroing are validated at runtime.
+    // Current supported types for skipping zeroing includes: [schema::Type::DATA].
+
+    bool enableLazyZero = false;
+    // Indicates whether lazy zero segment allocation is enabled.
+
+    std::set<schema::Type::Which> skipLazyZeroTypes;
```
kentonv, line 186: I think this configurability is more trouble than it is worth.

Realistically I don't think anyone would want to configure zeroing on the basis of type. They _might_ want to opt in on a field-by-field basis, but not on a type basis.

If we wanted to support opting in field-by-field, the way to do it would be to support an alternate form of the `initFoo(size)` method, like `allocateFoo(size)` or `uninitializedFoo(size)`. This approach avoids any possible mistakes, as the specific code that allocates the field is clearly opting into the value being uninitialized and therefore taking responsibility for initializing it.

The other extreme is we could just make in a boolean on the overall `BuilderArena` (`MessageBuilder`), so it applies to the whole message. This is simple but runs the risk of mistakes if the application calls some code to initialize a part of the message and that code isn't aware that zero-init is off. I'm not too keen on this approach as it seems too easy to make mistakes and any mistake is likely a critical security flaw (leaks contents of memory).

c++/src/capnp/arena.c++
```diff
@@ -62,6 +62,23 @@ void SegmentBuilder::throwNotWritable() {
...
+  const BuilderArena* arena = getArena();
+  if (!arena) return;
+
+  const auto& lazyZero = arena->getLazyZeroSegmentAlloc();
+
+  // Skip if lazy zero segment alloc is not enabled.
+  if (! lazyZero.enableLazyZero) return;
+
+  // Skip zeroing for types or fields that are configured to be skipped.
+  if (lazyZero.skipLazyZeroTypes.find(type) != lazyZero.skipLazyZeroTypes.end()) return;
```
kentonv, line 76: The other problem with this configurability is that this lookup is probably very expensive to perform in the hot path of serialization like this.

c++/src/capnp/layout.c++
```diff
@@ -1708,7 +1721,7 @@ struct WireHelpers {
       BuilderArena* orphanArena = nullptr)) {
     // Allocate the space.
     word* ptr = allocate(ref, segment, capTable, roundBytesUpToWords(size),
-                         WirePointer::LIST, orphanArena);
+                         WirePointer::LIST, orphanArena, /*type=*/schema::Type::DATA);
```
kentonv, line 1724: If the data blob's size is not evenly divisible by the word size (8 bytes), then there will be some bytes left over after the end of the data blob which will never be initialized here. E.g. if the application allocates a data blob of size 3, then there will be 5 bytes after it which are not part of the blob, and won't be part of the next allocation either; they'll just be skipped.

It is _critical_ that these bytes be zero'd here. Leaving them uninitialized means that the final message sent to the network will contain random uninitialized bytes, which may contain secrets. This is a major security bug.

In general I'm nervous about this change because of the possibility of bugs like this.

c++/src/capnp/message.h
```diff
@@ -145,6 +147,61 @@ class MessageReader {
...
+    // By default, Cap’n Proto requires allocators (including custom overridden allocateSegment methods)
+    // to return zero-initialized memory. When the caller enables lazy zero segment allocation by setting
+    // `BuilderOptions::lazyZeroSegmentAlloc` and uses a custom/overridden `allocateSegment`, the allocator
+    // may return un-zeroed segments. In this case, Cap’n Proto takes responsibility for
+    // zeroing during message construction: it will lazily zero only the parts of the segment that must
+    // be zeroed, and skip the regions that are explicitly configured as safe to leave un-zeroed.
+    // This approach reduces unnecessary memset work for large allocations (e.g., big `DATA` type blobs).
+    //
+    // IMPORTANT:
+    //   - A customised message builder with an overridden `allocateSegment()` must be used together with
```
kentonv, line 165: Why wouldn't we update `MallocMessageBuilder` to support this?

kentonv:
@PeakKS 
> Also somewhat related to this issue it would be nice to have a way to shrink the size of a data blob after initialization in case you didn't end up writing to the whole thing before you want to send it, removing the need for a additional field to specify the "real" size.

`Orphan<Data>::truncate()` lets you do this! Of course, it can only reclaim memory if the blob is still the last thing allocated in the message.

```
auto orphan = messageBuilder.orphanage().newOrphan<capnp::Data>(estimatedSize);
auto actualSize = fill(orphan.get());
orphan.truncate(acutalSize);
parentStrurct.adoptFoo(kj::mv(orphan));
```

c++/src/capnp/message.h
BrianXu0623: Merry Christmas! Thank you for taking the time to review this.

Thanks for the detailed feedback regarding usage, performance and security. Your concerns make total sense. I will take time to work on making the necessary changes to address these points.

BrianXu0623:
@kentonv 

Thanks for the detailed feedback.

To fully address the concerns regarding the usage, allocation contract and safety requirements, the necessary changes involved a significant refactor, so I decided to open a fresh PR that supersedes this one.

Please review the new PR here: #2485

I will close this PR in favor of the new one. Thanks!

kentonv:
OK, closing this PR in favor of the new one.

(I'm just getting back from holidays and am backlogged, so might take me a bit to get to the new PR, sorry.)

PR #2430, opened 2025-10-25
Build samples with bazel

amfern:
Add bazel build files to the sample folder

### DIFF
--- a/c++/src/capnp/BUILD.bazel
+++ b/c++/src/capnp/BUILD.bazel
@@ -84,10 +84,6 @@ cc_library(
     hdrs = [
         "persistent.capnp.h",
         "reconnect.h",
-        "message.h",
-        "common.h",
-        "layout.h",
-        "any.h",
         "rpc.capnp.h",
         "rpc.h",
         "rpc-prelude.h",


c++/src/capnp/BUILD.bazel
```diff
@@ -74,6 +74,10 @@ cc_library(
         "ez-rpc.h",
         "persistent.capnp.h",
         "reconnect.h",
+        "message.h",
```
amfern: ez-rpc.h is using message.h and other headers... but they were not included.

kentonv: They are from `:capnp`, which is listed as a dep, so it should not be necessary to list them again here.

kentonv: Er, also, `ez-rpc.h` was removed a long time ago?

amfern: Thank you for taking a look at this. I was using an older version and had to add this files.
I have rebased this branch on the v2 branch, so these indeed not needed.

Please take another look.

amfern:
ping @kentonv

kentonv:
What's the motivation for this exactly?

I mean I don't see a problem with it but I'm also not sure I understand the purpose. Is it just to help people learn how to build the samples, or is the intent that the samples should be built by CI, or do you have some need to build these regularly in a reproducible way?

amfern:
It's just to help people learn how to build the samples with bazel. Because it's not obvious how to use the `cc_capnp_library` rule.
Would you like to run it also as part of the CI?

kentonv:
Oh interesting. It didn't actually occur to me this was demonstrating the cc_capnp_library rule. Yes that makes sense.

kentonv:
Thanks!

PR #2434, opened 2025-10-28
bazel: use llvm_toolchains from

amfern:
By using toolchains from Bazel, developers can get started more easily without having to install the appropriate toolchains on their host machines. This approach leverages Bazel’s strengths by decoupling the development environment from the host system.

However downstream developers can still override the toolchains to any of their choosing


kentonv:
Thanks, but I think this isn't the direction we want to take. This bazel workspace file is really only here for testing bazel in CI, it's not really intended for people to use directly. Also, this is the `v2` branch, but most people should be using the `master` branch. It's hard to imagine who would be building this that doesn't already have a compiler installed. I worry that adding an explicit toolchain here means another thing we have to maintain and keep updated, so I think we want to leave it as-is.

Though if @fhanau disagrees then I'll let him make that call.

PR #2437, opened 2025-10-30
Add ConstString::clone() method

fhanau:
This allows us to elide memory allocations for literal string constants.

### DIFF
--- a/c++/src/kj/string-test.c++
+++ b/c++/src/kj/string-test.c++
@@ -383,54 +383,6 @@ KJ_TEST("ConstString literal operator") {
   KJ_EXPECT(theString == "it's a const string!");
 }
 
-KJ_TEST("ConstString promotion") {
-  kj::StringPtr theString = "it's a const string!";
-  kj::ConstString constString = theString.attach();
-  KJ_EXPECT(constString == "it's a const string!");
-}
-
-struct DestructionOrderRecorder {
-  DestructionOrderRecorder(uint& counter, uint& recordTo)
-    : counter(counter), recordTo(recordTo) {}
-  ~DestructionOrderRecorder() {
-    recordTo = ++counter;
-  }
-
-  uint& counter;
-  uint& recordTo;
-};
-
-KJ_TEST("ConstString attachment lifetimes") {
-  uint counter = 0;
-  uint destroyed1 = 0;
-  uint destroyed2 = 0;
-  uint destroyed3 = 0;
-
-  auto obj1 = kj::heap<DestructionOrderRecorder>(counter, destroyed1);
-  auto obj2 = kj::heap<DestructionOrderRecorder>(counter, destroyed2);
-  auto obj3 = kj::heap<DestructionOrderRecorder>(counter, destroyed3);
-
-  StringPtr theString = "it's a string!";
-  const char* ptr = theString.begin();
-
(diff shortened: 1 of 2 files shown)


c++/src/kj/array.h
```diff
@@ -275,8 +275,12 @@ class Array {
 private:
   T* ptr;
   size_t size_;
+public:
```
fhanau: I assume we don't actually want this to be public – to check if a ConstString is based on a string literal, we could either add a `bool Array<T>::isNullDisposer()` function here or introduce a simple boolean variable within ConstString that is set when constructing from LiteralStringConst.

fhanau:
Semi-related: `ConstString::attach()` feels like a serious footgun – there are only two uses I was able to find and one of them led to a UaF in workerd when I tried to use the clone() method on a variable derived from it since it also uses the NullArrayDisposer. It is possible to use it safely, but perhaps the potential for misuse is bigger than the potential gain here.

c++/src/kj/array.h
kentonv: `bool Array<T>::hasNullDisposer()` makes sense to me. Or a `getDisposer()` is probably OK too. But yeah I wouldn't make the field itself public.

kentonv:
> Semi-related: ConstString::attach() feels like a serious footgun

I don't see any `attach()` method in `ConstString`.

Maybe you meant `StringPtr::attach()`, which returns a `ConstString`?

Yeah, the no-argument version of this function seems problematic. It's only ever safe to use on a string literal, but if you have a string literal then you don't need it because you can construct `ConstString` directly. The function is problematic even in the absence of `clone()` -- if your StringPtr has limited lifetime, and you turn it into `ConstString`, then that `ConstString` can't safely be moved, which seems pretty broken.

c++/src/kj/array.h
fhanau: Updated to add `hasNullDisposer()`.

fhanau:
> Yeah, the no-argument version of this function seems problematic. It's only ever safe to use on a string literal, but if you have a string literal then you don't need it because you can construct `ConstString` directly. The function is problematic even in the absence of `clone()` -- if your StringPtr has limited lifetime, and you turn it into `ConstString`, then that `ConstString` can't safely be moved, which seems pretty broken.

Added a commit getting rid of `StringPtr::attach()` the version with arguments is wholly unused (except for its unit test), the version without arguments only had two uses downstream which were easy to convert.

c++/src/kj/string-test.c++
```diff
@@ -383,52 +383,16 @@ KJ_TEST("ConstString literal operator") {
...
-  combined = nullptr;
-
-  KJ_EXPECT(destroyed1 == 1, destroyed1);
-  KJ_EXPECT(destroyed2 == 2, destroyed2);
-  KJ_EXPECT(destroyed3 == 3, destroyed3);
+KJ_TEST("ConstString clone") {
+  // Clone from string literal
+  kj::ConstString literalConst = "foo"_kjc;
+  kj::ConstString literalClone = literalConst.clone();
+  KJ_EXPECT(literalClone == "foo");
```
kentonv, line 390: Perhaps worth verifying that the clone of a literal has the same location -- no copy was created.

And then for the heap clone below, verify this isn't the case.

PR #2438, opened 2025-10-31
[nfc] http-bench - helper binary to assist http stack benchmarking

mikea:
Supports 4 modes currently:
- http server
- capnp-over-http server
- http-to-http proxy
- http-to-capnp proxy

### DIFF
--- a/c++/.bazelrc
+++ b/c++/.bazelrc
@@ -12,6 +12,14 @@ build:dbg -c dbg --//src/kj:debug_memory=True
 # Opt build
 build:opt -c opt
 
+build:thin-lto --config=opt
+build:thin-lto --cxxopt='-flto=thin'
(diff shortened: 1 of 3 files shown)


kentonv:
There is also `http-over-capnp-perf-test.c++`, is this significantly different?

I guess one difference is you're actually communicating across a socket between separate processes here, whereas http-over-capnp-perf-test stays in-process. This should more realistically reflect the kernel-space time but that's also time we can't really optimize.

mikea:
The other big difference is presenting everything as http endpoint which makes it possible to measure/compare using http load testing tools.

Otherwise it is essentially the same with some cli glue on top. I also imagine adding different, more realistic kinds of endpoints besides 200 OK once I'm done with it.

kentonv (approved):
Didn't review in much detail, but seems reasonable.

PR #2445, opened 2025-11-11
Add AsyncOutputStream abort method

jasnell:
Intentionally non-op for backwards compatibility.
Provides the means for implementations to support
being notified when the stream is being shutdown
due to an error or explicit cancelation.

### DIFF
--- a/c++/src/kj/async-io.h
+++ b/c++/src/kj/async-io.h
@@ -146,6 +146,14 @@ public:
   //
   // Unlike most other asynchronous stream methods, it is safe to call whenWriteDisconnected()
   // multiple times without canceling the previous promises.
+
+  virtual void abortWrite(kj::Exception&& exception) {}
+  // Communicates to the stream that it should stop accepting writes and should fail any
+  // pending writes with the given exception. This is intended to be used when the stream is
+  // being shutdown due to an error or explicit cancelation. The default implementation, however,
+  // does nothing for backwards compatibilty. Existing implementations of AsyncOutputStream that
+  // want to support this method should override it, existing implementations that don't need it
+  // can just ignore it.
 };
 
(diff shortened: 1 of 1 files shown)


c++/src/kj/async-io.h
```diff
@@ -146,6 +146,14 @@ class AsyncOutputStream: private AsyncObject {
   //
   // Unlike most other asynchronous stream methods, it is safe to call whenWriteDisconnected()
   // multiple times without canceling the previous promises.
+
+  virtual void abort(kj::Exception&& exception) {}
+  // Communicates to the stream that it should stop accepting writes and should fail any
+  // pending writes with the given exception. This is intended to be used when the strema is
```
jasnell:
```suggestion
  // pending writes with the given exception. This is intended to be used when the stream is
```

c++/src/kj/async-io.h
```diff
@@ -146,6 +146,14 @@ class AsyncOutputStream: private AsyncObject {
   //
   // Unlike most other asynchronous stream methods, it is safe to call whenWriteDisconnected()
   // multiple times without canceling the previous promises.
+
+  virtual void abort(kj::Exception&& exception) {}
```
harrishancock: This means `AsyncIoStream` will have the following:
```cpp
  void shutdownWrite();  // clean end of writing
  void abort();  // unclean end of writing
  void abortRead();  // end of reading
```

and if we manage to add an `end()` one day, we'll have that, too. It'd be nice to make these all more consistently named somehow, but I don't want to block on the color of the bikeshed...

kentonv: I think we should call the new method `abortWrite()`.

jasnell: `abortWrite` SGTM. Updated!

Issue #1080, opened 2020-10-09
Allow backtrace to be disabled?


vlovich:
Currently backtrace is forced on on all platforms. I've recently discovered that at Facebook the Linux platform that ASAN runs on has some kind of problem where it reports ASAN. Would exposing a `KJ_ENABLE_BACKTRACE` macro to control this feature at a top level make sense so that I can just disable this on the platforms it's not relevant on? More generally we have infra to capture stack traces in prod & symbolicate them offline, so maybe the backtrace isn't even useful to us & can be globally omitted?

kentonv:
You can turn off stack tracing at runtime by creating a `kj::ExceptionCallback` whose `stackTraceMode()` method returns `StackTraceMode::NONE`. Is that sufficient? Or do you really need it removed at compile time?

vlovich:
At compile time is a bit easier to configure (eg many unit tests so it would require adding some infrastructure to do this globally) but if the preference is that this is a runtime toggle i can try to figure something out. Probably will end up just patching it locally

kentonv:
If you want to rename `KJ_HAS_BACKTRACE` (in `exception.c++`) to `KJ_USE_BACKTRACE` (for consistency with similar macros like `KJ_USE_EPOLL`), and change the logic so that if it's already defined it won't be redefined (so then you can define it to 0 to disable), that works for me.

vlovich:
Should this control the stack tracing in general or just `backtrace` on Linux? I'm thinking it should disable Windows too, in which case maybe `KJ_USE_STACKTRACE` may be more descriptive of a name?

kentonv:
Eh I don't think so:
* It seems like the use case here is that you've found your specific implementation of the backtrace API is broken, so that seems to call for something specific to the backtrace API.
* A `KJ_USE_BACKTRACE` macro could be nice for people to be able to override the current platform auto-detection behavior in general, including overriding it in the positive direction (i.e. assert "my platform has a working backtrace() even though KJ doesn't think it does).
* There's already a `KJ_USE_WIN32_DBGHELP` macro in `exception.c++`, which would make sense to make overrideable if you wanted to disable Windows backtraces...

PR #2446, opened 2025-11-12
kj/exception: KJ_USE_BACKTRACE to configure backtraces at compile

Brian-Williams:
KJ_USE_BACKTRACE can enable (1) or disable (0) backtraces regardless of platform where previously it was decided by platform.

Previously KJ_HAS_BACKTRACE could be set to 1 to enable if auto detection was failing for some reason, but disabling it was not respected.

I doubt that anyone is using KJ_HAS_BACKTRACE, but we maintain the backwards compatibility of KJ_HAS_BACKTRACE and the default platform detection if neither is defined.

See [#1080](https://github.com/capnproto/capnproto/issues/1080).

This is my first issue. It is targeting master because like the linked issue I need to overwrite auto detection in my production use case. Please let me know if I should also target v2 or *only* target v2.

### DIFF
--- a/c++/src/kj/exception.c++
+++ b/c++/src/kj/exception.c++
@@ -58,13 +58,6 @@
 #include <cxxabi.h>
 #endif
 
-// Use KJ_HAS_BACKTRACE; KJ_HAS_BACKTRACE is for backwards compat only.
-#if !defined(KJ_USE_BACKTRACE)
-#if defined(KJ_HAS_BACKTRACE) && ((KJ_HAS_BACKTRACE + 0) == 1)
-#define KJ_USE_BACKTRACE 1
-#endif
-#endif
-
 #ifndef KJ_USE_BACKTRACE
 #if (__linux__ && __GLIBC__ && !__UCLIBC__) || __APPLE__
 #define KJ_USE_BACKTRACE 1
@@ -72,7 +65,7 @@
 #endif
 
 #if KJ_USE_BACKTRACE
-# include <execinfo.h>
+#include <execinfo.h>
 #endif
 
 #if _WIN32 || __CYGWIN__


kentonv:
Out of curiosity, what is the motivation for disabling backtrace?

Brian-Williams:
I can fill in more when I have a better understanding, but the quick and dirty motivation is that gevent's cooperative stacks and unwinding aren't playing nicely.

I intend to do a full debug and get a more complete fix to reenable backtrace. Since I saw another issue requested the same feature I figured it would make sense to add.

kentonv:
Note that it's possible to disable backtracing at runtime too: Define an `ExceptionCallback`, and override `stackTraceMode()` to return `NONE`.

Is that useful in your use case, or is a compile-time option really better?

Brian-Williams:
It's better for my use case, but not required. I'm happy to close this if we want this to only be configurable at runtime.

c++/src/kj/exception.c++
```diff
@@ -58,9 +58,21 @@
 #include <cxxabi.h>
 #endif
 
+// Use KJ_HAS_BACKTRACE; KJ_HAS_BACKTRACE is for backwards compat only.
+#if !defined(KJ_USE_BACKTRACE)
+#if defined(KJ_HAS_BACKTRACE) && ((KJ_HAS_BACKTRACE + 0) == 1)
+#define KJ_USE_BACKTRACE 1
+#endif
+#endif
+
```
kentonv: I don't think we need back-compat with KJ_HAS_BACKTRACE -- it clearly wasn't designed to be set externally, given that it had to #ifndef guard.

```suggestion
```

c++/src/kj/exception.c++
```diff
@@ -58,9 +58,21 @@
...
+#ifndef KJ_USE_BACKTRACE
 #if (__linux__ && __GLIBC__ && !__UCLIBC__) || __APPLE__
-#define KJ_HAS_BACKTRACE 1
-#include <execinfo.h>
+#define KJ_USE_BACKTRACE 1
+#endif
+#endif
+
+#if KJ_USE_BACKTRACE
+# include <execinfo.h>
```
kentonv: Nitpick: This style doesn't appear anywhere else in the codebase, let's be consistent.

```suggestion
#include <execinfo.h>
```

kentonv (commented):
Oh, I'm fine with adding this, just wanted to check if you knew about the other options.

Brian-Williams:
Thanks for the review. Please let me know if you'd like any other changes.

PR #2447, opened 2025-11-12
fix InFlightExceptionIterator to skip consumed exceptions

mikea:


c++/src/kj/exception.c++
```diff
@@ -1205,6 +1205,11 @@ Maybe<const Exception&> InFlightExceptionIterator::next() {
 
   const ExceptionImpl* result = static_cast<const ExceptionImpl*>(ptr);
   validateExceptionPointer(result);
+  if (!result->isValid()) {
+    // caught exception was consumed, so there is no in-flight exception
+    ptr = nullptr;
```
kentonv: If there are more exceptions in the list (result->nextCurrentException is non-null) shouldn't we try them too?

(Happens when there are multiple layers of thrown exceptions -- probably rare.)

mikea: fixed by looping to find first valid exception

c++/src/kj/exception.c++
```diff
@@ -1201,12 +1201,17 @@ InFlightExceptionIterator::InFlightExceptionIterator()
     : ptr(currentException) {}
 
 Maybe<const Exception&> InFlightExceptionIterator::next() {
-  if (ptr == nullptr) return kj::none;
+  while (ptr != nullptr) {
+    const ExceptionImpl *result = static_cast<const ExceptionImpl *>(ptr);
+    validateExceptionPointer(result);
+    ptr = result->nextCurrentException;
+    if (result->isValid()) {
```
kentonv: Maybe not in this PR since it's not new, but I think `isMovedAway()` would be a clearer name than `isValid()` (with opposite semantics).

PR #2449, opened 2025-11-14
async-bench

mikea:
bazel-only since it depends on google_benchmark.

### DIFF
--- a/c++/src/kj/BUILD.bazel
+++ b/c++/src/kj/BUILD.bazel
@@ -279,3 +279,13 @@ cc_test(
         ":kj-test",
     ],
 )
+
+cc_test(
+    name = "async-bench",
+    srcs = ["async-bench.c++"],
+    tags = ["google_benchmark"],
+    deps = [
+        ":kj-async",
+        "@google_benchmark//:benchmark",
+    ],
+)
--- a/c++/src/kj/async-bench.c++
+++ b/c++/src/kj/async-bench.c++
@@ -1,0 +1,82 @@
+// Copyright (c) 2025 Cloudflare, Inc. and contributors
+// Licensed under the MIT License:
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// Micro-benchmarks for KJ async stack.
+
+#include <benchmark/benchmark.h>
+
+#include <kj/async.h>
+
+kj::Promise<size_t> immediate() { return 42; }
+
+static void bm_ReadyNow(benchmark::State &state) {
+  // Benchmark waiting for a kj::READ_NOW promise.
+  kj::EventLoop loop;
+  kj::WaitScope waitScope(loop);
+
+  for (auto _ : state) {
+    auto promise = []() -> kj::Promise<void> { return kj::READY_NOW; }();
+    promise.wait(waitScope);
+  }
+}
+
+BENCHMARK(bm_ReadyNow);
+
+static void bm_Immediate(benchmark::State &state) {
+  // Benchmark waiting for an immediate promise.
+  kj::EventLoop loop;
+  kj::WaitScope waitScope(loop);
+
+  for (auto _ : state) {
+    auto promise = immediate();
+    promise.wait(waitScope);
+  }
+}
+
+BENCHMARK(bm_Immediate);
+
+static void bm_coroCoReturn(benchmark::State &state) {
+  // Benchmark waiting for a co_return coroutine.
+  kj::EventLoop loop;
+  kj::WaitScope waitScope(loop);
+
+  for (auto _ : state) {
+    auto promise = []() -> kj::Promise<void> { co_return; }();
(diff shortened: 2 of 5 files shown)


mikea:
In case you're wondering here's the benchmark result (profile build)

```
------------------------------------------------------------------
Benchmark                        Time             CPU   Iterations
------------------------------------------------------------------
bm_ReadyNow                   20.7 ns         20.7 ns     34587970
bm_Immediate                  30.6 ns         30.6 ns     22604456
bm_coroCoReturn               49.2 ns         49.1 ns     14368533
bm_coroCoAwaitImmediate       80.0 ns         80.0 ns      7635680
```

c++/src/kj/BUILD.bazel
```diff
@@ -138,13 +138,15 @@ cc_library(
...
     hdrs = [
+        "benchmark.h",
         "test.h",
     ],
     include_prefix = "kj",
     visibility = ["//visibility:public"],
     deps = [
         ":kj",
         "//src/kj/compat:gtest",
+        "@google_benchmark//:benchmark",
```
fhanau: Make sure that this is handled properly with downstream repos (workerd doesn't directly use google_benchmark directly or overrides it IIRC), if the build still works with all downstream projects this can be merged.

fhanau (approved):

c++/src/kj/benchmark.h
```diff
@@ -0,0 +1,74 @@
...
+
+#define KJ_BENCHMARK(x) KJ_TEST(#x) { kj::BenchmarkState s{kj::TestCase::iterCount()}; x(s); }
+#define KJ_BENCHMARK_MAIN()
+
+namespace kj {
+    struct BenchmarkState {
+        size_t iterCount;
+
+        struct Iterator {
+            size_t i;
```
harrishancock: Nit: maybe make `const`?

c++/src/kj/async-bench.c++
```diff
@@ -0,0 +1,66 @@
...
+
+  for (auto _ : state) {
+    auto promise = [] () -> kj::Promise<void> {
+      return kj::READY_NOW;
+    }();
+    promise.wait(waitScope);
+  }
+}
+
+KJ_BENCHMARK(bm_ReadyNow);
```
harrishancock: `KJ_TEST` has something that causes a linker failure (I think) when it's defined outside of an anonymous namespace. So, I'd expect this to fail when benchmark/benchmark.h isn't available. Does it actually work, though? I see CI is green...

mikea: I does run on CI: https://github.com/capnproto/capnproto/actions/runs/19443340069/job/55631808569?pr=2449 so it doesn't seem to be an issue?

```
[ TEST ] kj/async-bench.c++:21: bm_ReadyNow
[ PASS ] kj/async-bench.c++:21: bm_ReadyNow (18.505μs)
[ TEST ] kj/async-bench.c++:34: bm_Immediate
[ PASS ] kj/async-bench.c++:34: bm_Immediate (3.556μs)
[ TEST ] kj/async-bench.c++:49: bm_coroCoReturn
[ PASS ] kj/async-bench.c++:49: bm_coroCoReturn (11.211μs)
[ TEST ] kj/async-bench.c++:64: bm_coroCoAwaitImmediate
[ PASS ] kj/async-bench.c++:64: bm_coroCoAwaitImmediate (6.752μs)
```

mikea:
> LGTM, but I'm curious about the anonymous namespace expectations. Even if `KJ_BENCHMARK` when falling back to `KJ_TEST` currently works outside of an anonymous namespace, that seems a bit confusing to me -- I'd prefer if both `KJ_BENCHMARK` and `KJ_TEST` followed the same rules.

apparently global-level KJ_TEST works fine: https://github.com/capnproto/capnproto/pull/2454

harrishancock:
> apparently global-level KJ_TEST works fine: https://github.com/capnproto/capnproto/pull/2454

All that shows is that our safeguard (https://github.com/capnproto/capnproto/blob/e2881d1d8179c097558718dda4ca9139ea0e1d28/c%2B%2B/src/kj/test.h#L75) is no longer working.

The kj-test "library" is designed so that you can link multiple `*-test.c++` files together, each with different KJ_TEST cases. The `main()` function is provided by `kj/test.c++` (which is why I said it's a "library" in quotes). If you link multiple test files together, and they share the same line numbers, and they are _not_ wrapped in anonymous namespaces, you end up with linker errors.

For example, if I copy async-bench.c++ to async-bench2.c++, fix its `immediate()` function to be `static`, and add async-bench2.c++ to `kj-heavy-tests` in the relevant CMakeLists.txt, I end up with this:

```
[ 32%] Linking CXX executable kj-heavy-tests
/usr/bin/ld: CMakeFiles/kj-heavy-tests.dir/async-bench2.c++.o:(.bss+0x0): multiple definition of `testCase21'; CMakeFiles/kj-heavy-tests.dir/async-bench.c++.o:(.bss+0x0): first defined here
/usr/bin/ld: CMakeFiles/kj-heavy-tests.dir/async-bench2.c++.o:(.data.rel.ro+0x0): multiple definition of `vtable for TestCase21'; CMakeFiles/kj-heavy-tests.dir/async-bench.c++.o:(.data.rel.ro+0x0): first defined here
/usr/bin/ld: CMakeFiles/kj-heavy-tests.dir/async-bench2.c++.o: in function `TestCase21::run()':
async-bench2.c++:(.text+0x0): multiple definition of `TestCase21::run()'; CMakeFiles/kj-heavy-tests.dir/async-bench.c++.o:async-bench.c++:(.text+0x0): first defined here
```

At the very least, this proves that this PR adds externally-visible symbols to the CMake `kj-heavy-tests` target which should not be externally-visible. It works as-is, because all of the other `*-test.c++` files that are linked into `kj-heavy-tests` are well-behaved `KJ_TEST` users, and wrap everything in an anonymous namespace.

What I'm saying is: if `KJ_BENCHMARK` falls back to `KJ_TEST`, it should also use `KJ_TEST` correctly.

In the short-term, if you want to just land this so you are unblocked, you might drop the `KJ_TEST` fallback. Is it really bad if we require Google Benchmark in our downstream projects for now?

mikea:
> In the short-term, if you want to just land this so you are unblocked, you might drop the `KJ_TEST` fallback. Is it really bad if we require Google Benchmark in our downstream projects for now?

I really didn't want to figure out google benchmark in autotools and cmake, that's why. But maybe I'm overthinking things.

Let me try to see if I can figure out anonymous namespaces. I'll add another micro-benchmark to source code to verify that it works.

harrishancock:
> I really didn't want to figure out google benchmark in autotools and cmake, that's why. But maybe I'm overthinking things.

I think it's reasonable to add Bazel-only library components if we need to.

> I'll add another micro-benchmark to source code to verify that it works.

Remember that the test cases will need to be defined on the same line in their respective files, so that they collide to the same identifier.

mikea:
> I think it's reasonable to add Bazel-only library components if we need to.

I'll go this way. I'll update the PR not to have benchmark.h and for bench to be bazel-only target.

mikea:
Done. I'll resurrect benchmark.h if we decided that we want to support it outside of bazel as well.

c++/src/kj/test.h
```diff
@@ -57,6 +57,8 @@ class TestCase {
     }
   }
 
+  static size_t iterCount();
```
kentonv, line 60: Why did you move this from private to protected? It doesn't appear that anything else in this PR uses this?

PR #2450, opened 2025-11-14
Remove KJ_UNREACHABLE calls from exception tests

mikea:
triggers warnings downstream


c++/src/kj/exception-test.c++
```diff
@@ -324,7 +324,6 @@ KJ_TEST("getDestructionReason returns default exception if exception wasn't thro
 KJ_TEST("getDestructionReason returns thrown exception if it wasn't consumed") {
   try {
     kj::throwFatalException(KJ_EXCEPTION(DISCONNECTED, "test exception"));
-    KJ_UNREACHABLE("no exception");
```
kentonv, line 327: Presumably this triggers warnings because KJ_UNREACHABLE does not take an argument? So this expands to:

```
::kj::_::unreachable();("no exception");
```

mikea: yeah, I'm surprised this is doesn' result in the error.

PR #2451, opened 2025-11-14
do not suspend immediate coroutine

mikea:


c++/src/kj/async-coroutine-test.c++
```diff
@@ -103,24 +98,21 @@ KJ_TEST("co_awaiting initial immediate promises suspends even if event loop is e
...
     size_t wind = 0, unwind = 0;
     bool evalLaterRan = false;
 
     auto promise = kj::evalLater([&]() { evalLaterRan = true; });
     auto coroPromise = countAroundAwait(wind, unwind, kj::mv(promise));
 
-    KJ_EXPECT(evalLaterRan == false);
-    KJ_EXPECT(wind == 1);
-    KJ_EXPECT(unwind == 0);
+    KJ_EXPECT(evalLaterRan == true);
```
harrishancock, line 113: @mikea this is the test case that asserted `evalLater()` semantics that I mentioned in our call.

PR #2454, opened 2025-11-18
tests in root namespace

mikea:
Removed unnecessary namespace declarations in test-test.c

### DIFF
--- a/c++/src/kj/test-test.c++
+++ b/c++/src/kj/test-test.c++
@@ -30,10 +30,6 @@
 #include <unistd.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


kentonv:
why?

mikea:
> why?

sorry, this was to verify if existing check for KJ_TEST outside of anonymouse namespace works. (it doesn't)

kentonv:
You would need to have _multiple_ test files that don't use anonymous namespaces, and then you need to link them together into a single binary. Then you'll see symbol conflicts.

Issue #1356, opened 2021-10-08
pkg-config generation when using MSVC would be nice


FrancisRussell:
I see that when compiling with MSVC, that pkg-config generation is disabled. It's unclear to me whether this is because something doesn't work, or it's an assumption that users compiling with MSVC wouldn't want them.

If it's the former, it would be nice to document this in the `CMakeLists.txt`. If it's the latter, it would be nice to be able to enable generation via an option passed to CMake. In particular it would enable us to remove platform specific logic from our build scripts, since we have issues with linking using the generated CMake under Linux.

munechika-koyo:
I would like to ask maintainers to consider this issue, because I ran into a build error in Windows in the conda context when building another package that depends on `capnp`, which tried to resolve the dependency with pkg-config.
I think the pkg-config file should also be provided on Windows platforms, or at least options should be made available to users.

kentonv:
The cmake files are contributed by the community. I actually don't have any idea how they work myself. If you want any changes here you'll have to send a PR.

munechika-koyo:
@kentonv Thank you for your comment!
I attempted to create PR #2456 by simply removing the MSVC conditional statement for pkg-config file creation.

munechika-koyo:
I noticed that the pkg-config files are useless if we didn't build the shared version for Windows.
Some guys have already mentioned it (#480, #1901).

When I tried to tweak the build configuration, I found it wasn't easy to fix. (presumably we need to change source codes as well, not only adding `-D BUILD_SHARED_LIBS=ON` and `-D CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS=ON` options.)

PR #2456, opened 2025-11-28
Enable pkg-config file installation on Windows

munechika-koyo:
This update allows the installation of pkg-config files on Windows, specified at #1356

Key changes include:

- Removal of the MSVC condition for pkg-config file installation

This change will distribute additional configuration files, so it doesn't seem to be related to v2 release. 
But I am not really sure if I should merge this PR to master branch.

### DIFF
--- a/c++/CMakeLists.txt
+++ b/c++/CMakeLists.txt
@@ -213,7 +213,9 @@ set(prefix "${CMAKE_INSTALL_PREFIX}")
 set(exec_prefix "") # not needed since we use absolute paths in libdir and includedir
 set(libdir "${CMAKE_INSTALL_FULL_LIBDIR}")
 set(includedir "${CMAKE_INSTALL_FULL_INCLUDEDIR}")
-set(PTHREAD_CFLAGS "-pthread")
+if(NOT MSVC)
+  set(PTHREAD_CFLAGS "-pthread")
+endif()
 set(STDLIB_FLAG)  # TODO: Unsupported
 set(CAPNP_PKG_CONFIG_FILES
   pkgconfig/kj.pc
@@ -240,8 +242,10 @@ foreach(pcfile ${CAPNP_PKG_CONFIG_FILES})
 endforeach()
 
 unset(STDLIB_FLAG)
-unset(PTHREAD_CFLAGS)
 unset(includedir)
 unset(libdir)
 unset(exec_prefix)
 unset(prefix)
+if(NOT MSVC)
+  unset(PTHREAD_CFLAGS)
+endif()


c++/CMakeLists.txt
```diff
@@ -208,43 +208,40 @@ install(FILES
...
-      pkgconfig/capnp-websocket.pc
-      pkgconfig/capnp-json.pc
-    )
-  endif()
+# Variables for pkg-config files
+set(prefix "${CMAKE_INSTALL_PREFIX}")
+set(exec_prefix "") # not needed since we use absolute paths in libdir and includedir
+set(libdir "${CMAKE_INSTALL_FULL_LIBDIR}")
+set(includedir "${CMAKE_INSTALL_FULL_INCLUDEDIR}")
+set(PTHREAD_CFLAGS "-pthread")
```
harrishancock: If the libraries were built with MSVC, I would not expect them to depend on pthread. So it seems odd to me that we'd tell dependents to set `-pthread` when linking with MSVC-built libraries.

But, if these pkg-config files work for you locally, there's probably no harm, either.

munechika-koyo: I understood.
I will set up an MSVC if-statement regarding this flag.

harrishancock: Thanks

harrishancock:
FreeBSD failures look clearly unrelated, ignoring those.

munechika-koyo:
Thank you for addressing this issue really promptly!

munechika-koyo:
Do you plan to publish a patched release, v1.2.1, that includes this change?

kentonv:
@munechika-koyo I have just released v1.3.0 with this fix and others.

munechika-koyo:
@kentonv Thank you for releasing the minor update.
However, as I mentioned at https://github.com/capnproto/capnproto/issues/1356#issuecomment-3633309026, we need to implement code to create Windows DLLs to make the pkg-config files effective.
I am just thinking of adding `*.def` files, not to make existing code messy, but I am not familiar with Windows development, so any advice would help me.

kentonv:
I'm not an expert on Windows either.

I'm vaguely familiar with the need to declare `__declspec(dllexport)` / `__declspec(dllimport)` on stuff and how people set up macros to handle that. I'm not very excited about littering our headers with such macros, they are quite noisy and annoying to deal with.

If there's some sort of tool that can just generate `.def` files covering all the declarations in some list of headers, that would be a whole lot nicer.

The Cloudflare Workers Runtime (my job) doesn't really need any of this, though, so I'm afraid I won't be able to work on it it myself. Happy to accept contributions if they are not too invasive.

munechika-koyo:
I understood your concern.

I would try to make `.def` files first, then post the PR later.

I am curious about the procedure of `capnproto` in the conda ecosystem, and would try to generate DLLs at https://github.com/conda-forge/capnproto-feedstock/pull/62, because conda-forge recommends uploading shared libraries (https://github.com/conda-forge/cfep/blob/main/cfep-18.md) even though on Windows, they are not necessarily needed.

alexshpilkin:
@kentonv If it helps, as far as I understand the declarations that need `__declspec(export)` when compiling a DLL on Windows are the same ones that need `__attribute__((visibility("default")))` when compiling a shared object on Linux with `-fvisibility=hidden` (which you probably should use because otherwise every globally visible symbol is part of your ABI and none can be inlined even within the same compilation unit).

kentonv:
> which you probably should use because otherwise every globally visible symbol is part of your ABI

Cap'n Proto generally declares module-internal functions with internal linkage (via `static` or wrapping in anonymous namespaces). It has almost no headers that aren't public. So "all globally visible symbols" is pretty close to being "all public symbols".

Also, Cap'n Proto doesn't attempt to be ABI-compatible between versions. The nature of C++ makes ABI compatibility too difficult to maintain.

> none can be inlined even within the same compilation unit

This surprised me a bit because this is not the behavior I see in practice.

Turns out, though, that GCC specifically will inhibit inlining of symbols with external linkage if they are not declared "hidden", in order to allow those symbols to be overridden by dynamic linking.

Clang (which is what I use pretty much exclusively) does not do this. It'll happily inline functions that have "default" visibility. This seems valid to me, under the One Definition Rule.

Both compilers will, of course, allow inlining functions that  are declared `inline`, or which are implicitly inline by virtue of having their body defined nested within the parent class's body. Nearly all functions in Cap'n Proto that you'd want to inline fall into this category.

Both compilers will also always allow inlining functions that have internal linkage due to being declared `static` or within an anonymous namespace.

But indeed, I wasn't aware that GCC tries to make all non-inline symbols able to be overridden by dynamic linking by default. That's interesting and indeed makes me more willing to accept littering the code with export macros...

(Though it still doesn't affect Cloudflare Workers, I think, since we compile with Clang and we statically link against Cap'n Proto.)

alexshpilkin:
@kentonv Yes, you could argue the ODR outright prohibits this “semantic interposition”. Following the example of static libraries (which after all is what this all rather ill-conceived dynamic linking behaviour was attempting to emulate), you could say that if the executable or a (BFS-)closer library in the dependency graph wants to override something in a file, it will need to override everything defined in that file. I’m guessing that’s what Clang has decided here.

You could also argue that it is least surprising to follow the example of C, where the established precedent all the way to the birth of this mechanism in Solaris is that interposition can happen symbol by symbol, unlike in static libraries. (This is, once again, not smart, and Solaris soon realized that and at least stopped doing it by default, as eventually did OS X, only Glibc didn’t.) I’m guessing that’s the GCC reasoning.

I didn’t know of this difference for C++ (but can now confirm it). Either way, however, if the call to a default-visibility symbol is _not_ inlined, both compilers go via the PLT, so even your non-virtual calls are now indirect call instructions. To actually make either compiler go ODR all the way, you need to pass `-fno-semantic-interposition` (which does not prevent interposition, only tells the compiler to assume that it does not happen). There’s also the [link-time flag][] `-symbolic` (aka `-Wl,-Bsymbolic`) but I’ve always had problems understanding precisely what it does.

[link-time flag]: https://maskray.me/blog/2021-05-16-elf-interposition-and-bsymbolic

PR #2457, opened 2025-12-01
Return a protocol error when web socket receives invalid compressed data

fhanau:


c++/src/kj/compat/http.c++
```diff
@@ -3406,6 +3406,13 @@ private:
           break;
         case Mode::DECOMPRESS:
           result = inflate(&ctx, Z_SYNC_FLUSH);
+          // inflate() returns Z_DATA_ERROR if the provided input is not valid compressed data.
+          // Return a protocol error in that case.
+          if (result == Z_DATA_ERROR) {
+              return ProtocolError {
```
fhanau, line 3412: Open to returning a different error message here – RFC 7692 which I believe describes the web socket compression we're doing here does not say anything about error handling so there might not be a standard error message we could use.

c++/src/kj/compat/http.c++
```diff
@@ -3406,6 +3406,13 @@ private:
...
+                  .statusCode = 1002,
+                  .description = "invalid compressed data"};
```
kentonv: We should be consistent with the style of other protocol errors.

```suggestion
                  .description = "Invalid compressed data"};
```

PR #2464, opened 2025-12-16
Clarify comment about `scopeId`.

### LINKED ISSUE #2427: Misleading description of schema.capnp:Type.anyPointer.parameter.scopeId
From the description, the `scopeId` of the generic parameter should be "a parent of the current scope".
https://github.com/capnproto/capnproto/blob/d0bab02e6cc4a962f0aeef6012162a025dfc8b77/c%2B%2B/src/capnp/schema.capnp#L369-L371

However, for direct generic parameter referencing (no inheritance), like:
```capnp
struct Foo(T) {
    value @0 :T;
}
```

The parameter's scopeId of the field `value` will be the type ID of `Foo`, which is not the parent of the current scope; instead, this is the ID of the current scope.

Another example is:
```capnp
struct Foo(T) {
    struct Bar {
        struct Qux {
            value @0 :T;
        }
    }
}
```

Again, the parameter's scopeId of the field `value` will be the type ID of `Foo`, which is not the parent of the current scope `Qux`; instead, this is the ancestor of the current scope.

The first statement of this description is always true, but the latter one is somewhat misleading.

kentonv:
Fixes #2427.

### DIFF
--- a/c++/src/capnp/schema.capnp
+++ b/c++/src/capnp/schema.capnp
@@ -369,8 +369,8 @@ struct Type {
         # This is actually a reference to a type parameter defined within this scope.
 
         scopeId @19 :Id;
-        # ID of the generic type whose parameter we're referencing. This should be a parent of the
-        # current scope.
+        # ID of the generic type whose parameter we're referencing. This is always either the
+        # current scope's type ID or one of its ancestors' IDs.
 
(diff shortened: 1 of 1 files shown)


kentonv:
Comment-only change, not going to waste anyone's time asking for a review.

PR #2466, opened 2025-12-16
Extend configure.ac to check if -latomic is needed.

### LINKED ISSUE #1448: Raspbian build fails due to need for libatomic

kentonv:
Written using Claude Code.

Fixes #1448 (the rest of the way)

### DIFF
--- a/c++/configure.ac
+++ b/c++/configure.ac
@@ -113,6 +113,39 @@ AC_SUBST([CAPNP_LITE_FLAG])
 
 AC_SEARCH_LIBS(sched_yield, rt)
 
+# Check if we need to link against libatomic for atomic operations.
+# On some architectures (e.g., armel, mipsel, powerpc), GCC requires linking
+# with libatomic to support certain atomic operations.
+AC_MSG_CHECKING([if libatomic is needed])
+AC_LINK_IFELSE([
+  AC_LANG_PROGRAM(
+    [[#include <atomic>
+      #include <cstdint>]],
+    [[std::atomic<std::int64_t> a{0};
+      a.fetch_add(1, std::memory_order_relaxed);
+      return 0;]])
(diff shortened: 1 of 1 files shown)


kentonv:
We don't even use autoconf in workers so I'm not going to worry about formal code review requirements so much.

PR #2467, opened 2025-12-16
Cherry-pick some fixes from v2 into 1.x

kentonv:


kentonv:
I had Claude research the FreeBSD 15 failure and it's a kernel bug that was fixed recently. Hah!

## Claude's analysis

Based on my research, I found the likely cause of the test failure on FreeBSD 15. Here's what happened:

  The Root Cause

  FreeBSD 15 introduced a kernel bug in the kqueue implementation that affected EVFILT_WRITE event handling for Unix domain sockets (including socketpairs).

  Timeline of Events:

  1. Original Bug (eafe5967ac558): A commit was made to fix https://www.mail-archive.com/freebsd-bugs@freebsd.org/msg89678.html, which addressed an issue where sway compositor was getting into an infinite loop. This commit fixed the case when the peer side closes a Unix stream socket connection.
  2. Regression Introduced: However, this same commit inadvertently broke a different code path - when the local side closes its end of the socket. The problem was that the fix removed the line kn->kn_flags |= EV_EOF; which is needed to signal EOF on write disconnect events.
  3. The Fix (https://www.mail-archive.com/dev-commits-src-main@freebsd.org/msg41873.html): On September 2, 2025, Gleb Smirnoff committed a fix that restored the proper EV_EOF flag handling for the shutdown/close case. This fix was discovered with help from the tokio-rs/mio project.
  4. FreeBSD 15.0 Release: FreeBSD 15.0-RELEASE was released on December 2, 2025.

  What This Means for Your Test

  The whenWriteDisconnected() test relies on kqueue's EVFILT_WRITE filter returning with the EV_EOF flag set when the remote end of a socketpair is closed. During the buggy period:

  - The promise returned by whenWriteDisconnected() would never resolve because the kernel wasn't setting EV_EOF on the write filter
  - This would cause the test to hang or timeout

  Why It Fails on FreeBSD 15 but Not 14.2/13.4

  FreeBSD 14.2 and 13.4 don't have the buggy commit, so they work correctly. FreeBSD 15 development snapshots between the introduction of the bug and September 2, 2025 would exhibit the failure.

  Solution

  If you're testing on:
  - FreeBSD 15.0-RELEASE or later: The fix should be included, and the test should pass
  - FreeBSD 15 snapshots/CURRENT before September 2, 2025: Update to a newer snapshot that includes commit 60d5a4216b34
  - FreeBSD 15 pre-release builds: These might have the bug; wait for 15.0-RELEASE or update to the latest snapshot

  You can verify if the fix is present by checking if commit 60d5a4216b34 is in your FreeBSD source tree.

  Sources:

  - https://www.mail-archive.com/dev-commits-src-main@freebsd.org/msg41873.html
  - https://www.mail-archive.com/dev-commits-src-main@freebsd.org/msg42054.html
  - https://www.freebsd.org/releases/15.0R/relnotes/
  - https://www.freebsd.org/releases/

PR #2473, opened 2025-12-19
[nfc] integrating codspeed benchmarks

mikea:
- will run all google benchmarks on every pr as workerd does
- flag tweaks/cleanup

### DIFF
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -1,0 +1,56 @@
+name: CodSpeed
+
+on:
+  pull_request:
+  merge_group:
+  push:
+    branches:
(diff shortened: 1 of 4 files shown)


mikea:
@art049 do you think you can enable capnproto repo on codspeed side for data collection?

art049:
Hey @mikea, just checked, and I don't see the org in our app. You need to enable the GitHub/CodSpeed app like shown [here](https://codspeed.io/docs#connect-your-repository). This will allow us to access the repo to put the performance checks and comments
Someone with admin permissions on the organization will be allowed to enable it.

anonrig (approved):

PR #2475, opened 2025-12-22
Add CapabilityServerSet::tryGetLocalServerSync().

kentonv:
This is a bit of a hack, but works in certain cases where we know for sure that promises are already resolved, including a case I'm working on in: https://github.com/cloudflare/workerd/pull/5738

### DIFF
--- a/c++/src/capnp/capability.h
+++ b/c++/src/capnp/capability.h
@@ -699,6 +699,7 @@ class CapabilityServerSetBase {
 public:
   Capability::Client addInternal(kj::Own<Capability::Server>&& server, void* ptr);
   kj::Promise<void*> getLocalServerInternal(Capability::Client& client);
+  void* tryGetLocalServerSyncInternal(Capability::Client& client);
 };
 
 }  // namespace _ (private)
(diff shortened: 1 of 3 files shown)


c++/src/capnp/capability.h
```diff
@@ -699,6 +699,7 @@ class CapabilityServerSetBase {
 public:
   Capability::Client addInternal(kj::Own<Capability::Server>&& server, void* ptr);
   kj::Promise<void*> getLocalServerInternal(Capability::Client& client);
+  void* tryGetLocalServerSyncInternal(Capability::Client& client);
```
jasnell, line 702: Nit: the `void*` return here is a bit unfortunate but obviously isn't new given the `getLocalServerInternal`. A comment here explaining the expectations would be nice but optional.

c++/src/capnp/capability.h
```diff
@@ -730,6 +731,15 @@ class CapabilityServerSet: private _::CapabilityServerSetBase {
...
   // accepts an lvalue input).
+
+  kj::Maybe<typename T::Server&> tryGetLocalServerSync(typename T::Client& client);
+  // Like getLocalServer() but attempts to unwrap synchronously. A null return value does not
+  // necessarily mean that the capability isn't part of this set, just that it isn't known yet;
+  // in this case you must fall back to `getLocalServer()`. (In particular, if the capability is
+  // a promise, it's necessary to wait for it to resolve.)
+  //
+  // DANGER: Capabilities can be promises more often than you think. Be very careful about using
+  // this. If in doubt, don't.
```
jasnell, line 742: Nit: adding something like an `Unsafe` label to the API name (e.g. `tryGetLocalServerSyncUnsafe(...)` might be useful to encourage folks to be more careful about using this. Optional tho.

jasnell (approved):

c++/src/capnp/capability.h
kentonv: Sorry, was so eager to finally land my three-part PR that I missed the comment.

That said, this is an internal interface, not exposed to applications at all. It is used in the implementation of the more type-safe interface below. The reason this deals in void* is just so we don't need so much template code.

PR #2486, opened 2025-12-30
Allow late -DCMAKE_INSTALL_PREFIX

majg0:
# Problem

The following will error with lacking permission:

```sh
mkdir -p build && cmake -DCMAKE_BUILD_TYPE=Release -B build -G Ninja && cmake --build build && cmake -DCMAKE_INSTALL_PREFIX=install -P build/cmake_install.cmake
```

because if the user doesn't **configure** with -DCMAKE_INSTALL_PREFIX, the resulting installation becomes:

```cmake
execute_process(COMMAND "/usr/bin/cmake" -E create_symlink capnp "$ENV{DESTDIR}/usr/local/bin/capnpc")
```

# Solution

With the proposed change, the symlink line becomes:

```diff
- execute_process(COMMAND "/usr/bin/cmake" -E create_symlink capnp "$ENV{DESTDIR}/usr/local/bin/capnpc")
+ execute_process(COMMAND "/usr/bin/cmake" -E create_symlink capnp "$ENV{DESTDIR}${CMAKE_INSTALL_PREFIX}/bin/capnpc")
```

which matches the other commands generated by cmake, like:

```cmake
execute_process(COMMAND "/usr/bin/strip" "$ENV{DESTDIR}${CMAKE_INSTALL_PREFIX}/bin/capnpc-capnp")
```

allowing the user the late-bind the installation prefix!

# Workaround

Early binding removes the error:

```diff
 mkdir -p build && \
     cmake \
+        -DCMAKE_INSTALL_PREFIX=install \
         -DCMAKE_BUILD_TYPE=Release -B build -G Ninja && \
     cmake --build build && \
     cmake \
-       -DCMAKE_INSTALL_PREFIX=install \
        -P build/cmake_install.cmake

```

# Verification

I have **NOT** tested this on Windows! It *should* work, but I can't be certain.

### DIFF
--- a/c++/src/capnp/CMakeLists.txt
+++ b/c++/src/capnp/CMakeLists.txt
@@ -213,10 +213,10 @@ if(NOT CAPNP_LITE)
   if(WIN32)
     # On Windows platforms symlinks are not guaranteed to support. Also different version of CMake handle create_symlink in a different way.
     # The most portable way in this case just copy the file.
-    install(CODE "execute_process(COMMAND \"${CMAKE_COMMAND}\" -E copy \"\$ENV{DESTDIR}${CMAKE_INSTALL_FULL_BINDIR}/capnp${CMAKE_EXECUTABLE_SUFFIX}\" \"\$ENV{DESTDIR}${CMAKE_INSTALL_FULL_BINDIR}/capnpc${CMAKE_EXECUTABLE_SUFFIX}\")")
+    install(CODE "execute_process(COMMAND \"${CMAKE_COMMAND}\" -E copy \"\$ENV{DESTDIR}\${CMAKE_INSTALL_PREFIX}/bin/capnp${CMAKE_EXECUTABLE_SUFFIX}\" \"\$ENV{DESTDIR}\${CMAKE_INSTALL_PREFIX}/bin/capnpc${CMAKE_EXECUTABLE_SUFFIX}\")")
   else()
     # Symlink capnpc -> capnp
-    install(CODE "execute_process(COMMAND \"${CMAKE_COMMAND}\" -E create_symlink capnp${CMAKE_EXECUTABLE_SUFFIX} \"\$ENV{DESTDIR}${CMAKE_INSTALL_FULL_BINDIR}/capnpc${CMAKE_EXECUTABLE_SUFFIX}\")")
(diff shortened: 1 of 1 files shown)


kentonv:
Hmm, I think this has a subtle problem: The install location is embedded into the `capnp` binary at compile time, as CAPNP_INCLUDE_DIR, the location where the capnp compiler tool will search for imports. If you change the install prefix at the install step, even with your fix, you'll still end up with a binary that is searching the wrong import path.

Moreover, the fact that cmake computes CMAKE_INSTALL_FULL_BINDIR based on CMAKE_INSTALL_PREFIX at build time, ignoring any changes at install time, suggests to me that cmake does not intend for people to be able to change the install prefix at install time generally. Maybe safer to stick with what cmake is designed for?

Disclaimer: I don't actually know anything about cmake myself.

Sarcasm:
I think install prefix was a configuration time thing for a long time but CMake has improved support for changing the prefix at install time.

There is the `cmake --install <dir>` subcommand:
- https://cmake.org/cmake/help/latest/manual/cmake.1.html#install-a-project

On my CMake project, where I also depend on locating resources, but I want my installs to be relocatable.
Just have to be careful not using the `CMAKE_INSTALL_FULL_<XXX>` variables.
I find the resources relative to the .so / executable (platform specific, via `dladdr` and the `Dl_info.dli_fname` and compiling in position independent code mode).

majg0:
Great catch @kentonv !

How about using @Sarcasm 's method : finding the filepath of the binary (exe or lib alike) containing some symbol of choice at runtime, and using this as the basis for path resolution?

We would require separate implementations for unix and windows, but then we'd decouple the binary location from compilation.

kentonv:
> How about using Sarcasm's method : finding the filepath of the binary (exe or lib alike) containing some symbol of choice at runtime, and using this as the basis for path resolution?

I'm open to this but there's some work involved, including some research / design work:
* On Unix, I would want to exactly match whatever behavior GCC and Clang have, for consistency. So someone has to figure that out. E.g. what happens if the binary is a symlink? Do we look up relative to the link or the target?
* What exactly happens on Windows? Keep in mind we should behave reasonably both in MSYS (unix-like) and idiomatic Windows installations. Answer unclear.

See also #2364 which tried to solve this problem specifically for MSYS but I wasn't very excited about accepting a narrow solution.

PR #2489, opened 2025-12-31
capnp-mode.el: fix comment configuration

isker:
The existing approach only works if you use `comment-dwim`, but that function is just a wrapper around lower-level ones that actually consume the various `comment-` vars. So, if you don't use `comment-dwim`, you are not helped by capnp-mode in this regard.

So, drop the existing approach and just set an appropriate value for `comment-start` in all capnp-mode buffers. This is typically how major modes configure this kind of thing. I can't say why it was done a different way to start with. (The configuration of `comment-end` in `capnp-comment-dwim` was redundant, as `""` is the default value of this var.)

### DIFF
--- a/highlighting/emacs/capnp-mode.el
+++ b/highlighting/emacs/capnp-mode.el
@@ -20,16 +20,6 @@
 
 ;;; Code:
 
-;; command to comment/uncomment text
-(defun capnp-comment-dwim (arg)
-  "Comment or uncomment current line or region in a smart way.
-For detail, see `comment-dwim' for ARG explanation."
-  (interactive "*P")
-  (require 'newcomment)
-  (let (
-        (comment-start "#") (comment-end ""))
-    (comment-dwim arg)))
-
 (defvar capnp--syntax-table
   (let ((syn-table (make-syntax-table)))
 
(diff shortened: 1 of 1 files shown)


highlighting/emacs/capnp-mode.el
```diff
@@ -65,11 +55,9 @@ For detail, see `comment-dwim' for ARG explanation."
   "capn-mode is a major mode for editing capnp protocol files"
   :syntax-table capnp--syntax-table
 
+  (setq-local comment-start "#")
```
hpfr:
```suggestion
  (setq-local comment-start "# ")
```

hpfr (changes requested):
Should probably set up another standard variable (from `python.el`):
```elisp
(setq-local comment-start-skip "#+\\s-*")
```

That allows comment commands to jump to the body even when there are multiple leading `#` (commonly used in `outline-minor-mode` headers) and spaces.

highlighting/emacs/capnp-mode.el
isker: Why? What I’ve got already works as far as I can tell, due to `comment-padding`. https://www.gnu.org/software/emacs/manual/html_node/emacs/Options-for-Comments.html

hpfr: I haven’t read the `newcomment` code enough to offer a full explanation, I’m just trying to follow existing examples, but if you try `comment-dwim` on a line with code to add a comment at the end of the line, the padding doesn’t get used.

isker: OK, thanks for the review, I applied your suggestions.

hpfr (approved):

kentonv:
@justin-mp as the emacs guy on the team can you take a look?

PR #2490, opened 2025-12-31
[coro] remove unwind detector from PromiseAwaiter

mikea:
By disposing node in resume we get the same behavior and pay the exception handling price only if coro resolved into exception.

It is import to optimize this path since every `co_await` incurs the cost.

There's another UnwindDisposer in CoroutineBase that I'm looking at, but one tiny step at a time I guess.

### DIFF
--- a/c++/src/kj/async.c++
+++ b/c++/src/kj/async.c++
@@ -3304,11 +3304,6 @@ PromiseAwaiterBase::~PromiseAwaiterBase() noexcept(false) {
   KJ_IF_SOME(coroutine, maybeCoroutine) {
     coroutine.clearPromiseNodeForTrace();
   }
-
-  unwindDetector.catchExceptionsIfUnwinding([this]() {
-    // No need to check for a moved-from state, node will just ignore the nullification.
-    node = nullptr;
-  });
 }
 
 void PromiseAwaiterBase::awaitResumeImpl(ExceptionOrValue& result, void* awaitedAt) {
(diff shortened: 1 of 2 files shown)


c++/src/kj/async.c++
```diff
@@ -3324,11 +3319,20 @@ void PromiseAwaiterBase::awaitResumeImpl(ExceptionOrValue& result, void* awaited
     // comment there).
     exception.addTrace(reinterpret_cast<void*>(reinterpret_cast<uintptr_t>(awaitedAt) - 1));
 
+    try {
```
harrishancock: I want to suggest something like this:

```cpp
node->get(result);

try {
  node = nullptr;
} catch (...) {
  result.addException(kj::mv(getCaughtExceptionAsKj());
}

KJ_IF_SOME(exception, result.exception) {
  // extend stack trace and rethrow
}
```

`result.addException()` only sets the first exception, so the end result should be (almost) the same as this code, but the control flow is slightly simpler, and it matches a very standard pattern in the rest of kj-async (search for `addException` in async.c++, and you'll see it all over the place).

The only slight change in behavior between my snippet above and the current PR (and existing code) is the stack trace extension -- in my snippet, we would extend the stack trace for exceptions thrown during destruction, whereas in your PR, and the existing code, we do not. But maybe we should? It would probably make sense to, but I'm not sure.

mikea: awesome suggestion, thanks.

kentonv:
I think that technically this isn't equivalent in the case of cancellation.

mikea:
> I think that technically this isn't equivalent in the case of cancellation.

let's revert this (https://github.com/capnproto/capnproto/pull/2504) and see if I can write/identify a test (I was under impression there already is one)

PR #2491, opened 2026-01-05
allow Event.fire to delete itself

mikea:
Removes `firing` tracking. Besides making self-deletion possible this simplification improves performance across the board: https://gist.github.com/mikea/6216a6512423e6eaa2784e1e94594cc9 (`bm_Http_FullProtocol` is ~4.8%)

### DIFF
--- a/c++/src/kj/async.c++
+++ b/c++/src/kj/async.c++
@@ -326,8 +326,10 @@ protected:
     }
 
     // Remove from the task list. Do this before calling taskFailed(), so that taskFailed() can
-    // safely call clear(). Will be destroyed on scope exit.
+    // safely call clear().
     auto self = pop();
+    // self will be destroyed on scope exit.
+    self->permitSelfDestruction();
 
     // We'll also process onEmpty() now, just in case `taskFailed()` actually destroys the whole
     // `TaskSet`.
@@ -1114,8 +1116,9 @@ void XThreadEvent::fire() {
   KJ_IF_SOME(n, promiseNode) {
     n->get(result);
     promiseNode = kj::none;  // make sure to destroy in the thread that created it
-    done(); // possible deletes this
-    return;
+    // Done might delete this.
+    permitSelfDestruction();
+    done();
   } else {
     KJ_IF_SOME(exception, kj::runCatchingExceptions([&]() {
       promiseNode = execute();
@@ -1125,8 +1128,9 @@ void XThreadEvent::fire() {
     KJ_IF_SOME(n, promiseNode) {
       n->onReady(this);
     } else {
-      done(); // possible deletes this
-    return;
+      // Done might delete this.
+      permitSelfDestruction();
+      done();
     }
   }
 }
@@ -2166,11 +2170,15 @@ kj::EventLoop& Event::requireEventLoop() {
 }
 
 Event::~Event() noexcept {  // intentionally noexcept
-  if (threadLocalCurrentlyFiring == this) {
-    threadLocalCurrentlyFiring = nullptr;
+  if (KJ_UNLIKELY(threadLocalCurrentlyFiring == this)) {
+    KJ_FAIL_REQUIRE("Promise callback destroyed itself.");
   }
-
   disarm();
+}
+
+void Event::permitSelfDestruction() {
+  KJ_REQUIRE(threadLocalCurrentlyFiring == this, "Event needs to be firing.");
+  threadLocalCurrentlyFiring = nullptr;
 }
 
 void Event::armDepthFirst() {
@@ -2230,8 +2238,6 @@ String TraceBuilder::toString() {
 }  // namespace _ (private)
 
 ArrayPtr<void* const> getAsyncTrace(ArrayPtr<void*> space) {
-  EventLoop* loop = threadLocalEventLoop;
-  if (loop == nullptr) return nullptr;
   if (threadLocalCurrentlyFiring == nullptr) return nullptr;
 
   _::TraceBuilder builder(space);
@@ -2566,8 +2572,11 @@ void ChainPromiseNode::fire() {
   state = STEP2;
 
   if (selfPtr != nullptr) {
-    // Hey, we can shorten the chain here. Will be destroyed on scope exit.
+    // Hey, we can shorten the chain here.
     auto chain = selfPtr->downcast<ChainPromiseNode>();
+    // self will be destroyed on scope exit.
+    chain->permitSelfDestruction();
+
     *selfPtr = kj::mv(inner);
     selfPtr->get()->setSelfPointer(selfPtr);
     if (onReadyEvent != nullptr) {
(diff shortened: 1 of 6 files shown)


c++/src/kj/async.c++
```diff
@@ -1838,11 +1838,10 @@ bool EventLoop::turn() {
 
     Maybe<Own<_::Event>> eventToDestroy;
     {
-      event->firing = true;
-      KJ_DEFER(event->firing = false);
       currentlyFiring = event;
       KJ_DEFER(currentlyFiring = nullptr);
       eventToDestroy = event->fire();
+      currentlyFiring = nullptr;
```
kentonv: The KJ_DEFER above already does this.

But probably we need the event's destructor to null out `currentyFiring` if it points to the event, or we need to rethink how `currentlyFiring` works altogether, since we don't want to leave it pointing at an event that has been destroyed.

mikea: ~fixed in latest version. I'll look at .detach next.~

not quite yet.

mikea: XThreadEvent complicates everything since it outlives theEventLoop.

I gave it a go in the 3rd commit (https://github.com/capnproto/capnproto/pull/2491/changes/6be6af0c8ec5c38aab982b10111d9f061060ce3f) by making EventLoop& a Maybe.

c++/src/kj/async.c++
```diff
@@ -1038,6 +1036,14 @@ void XThreadEvent::ensureDoneOrCanceled() {
     }
   }
 
+  // Check if the target loop is still alive. If not, we need to detach from it to prevent
+  // ~Event() from accessing invalid memory. We need to check this even if state was already DONE
+  // when we entered, because the loop could have been destroyed and transitioned us to DONE.
+  auto lock = targetExecutor->impl->state.lockExclusive();
```
harrishancock: This should probably be held in a new lexical scope, otherwise we'll end up holding two locks at once if we have a `replyExecutor` (line 1052).

c++/src/kj/async.c++
```diff
@@ -2128,12 +2139,18 @@ void detach(kj::Promise<void>&& promise) {
 }
 
 Event::Event(SourceLocation location)
-    : loop(currentEventLoop()), next(nullptr), prev(nullptr), location(location) {}
+    : loop(&currentEventLoop()), next(nullptr), prev(nullptr), location(location) {}
```
harrishancock: Non-blocking nit: I'd expect the previous code to have worked, too -- `Maybe<T&>` should be instantiable from `T&`.

kentonv:
Is there a way we can still catch unintentional self-cancellations?

I am worried about application code that e.g. accidentally cancels the coroutine that is currently running, which will almost certainly lead to a UAF.

mikea:
This change works much better on top of https://github.com/capnproto/capnproto/pull/2546 because it removes EventLoop& from Event, so no additional complexities in cross-thread resolvers. I'll hold off this change and focus on landing event list improvements first.

c++/src/kj/async.c++
```diff
@@ -1151,11 +1125,10 @@ Maybe<Own<Event>> XThreadEvent::fire() {
     KJ_IF_SOME(n, promiseNode) {
       n->onReady(this);
     } else {
-      return Own<Event>(this, DISPOSER);
+      done(); // possible deletes this
+    return;
```
kentonv:
```suggestion
```

c++/src/kj/async.c++
```diff
@@ -1114,34 +1110,12 @@ void XThreadEvent::setDisconnected() {
...
-Maybe<Own<Event>> XThreadEvent::fire() {
-  static constexpr DelayedDoneHack DISPOSER {};
-
+void XThreadEvent::fire() {
   KJ_IF_SOME(n, promiseNode) {
     n->get(result);
     promiseNode = kj::none;  // make sure to destroy in the thread that created it
-    return Own<Event>(this, DISPOSER);
+    done(); // possible deletes this
+    return;
```
kentonv:
```suggestion
```

c++/src/kj/async.c++
```diff
@@ -2197,74 +2166,40 @@ kj::EventLoop& Event::requireEventLoop() {
...
-  std::atomic_signal_fence(std::memory_order_acq_rel);
+  if (threadLocalCurrentlyFiring == this) {
+    threadLocalCurrentlyFiring = nullptr;
+  }
 
   disarm();
-
-  // If this fails, we'll abort due to `noexcept`. That's good because otherwise we're likely to
-  // be in a use-after-free situation.
-  KJ_REQUIRE(!firing, "Promise callback destroyed itself.");
```
kentonv, line 2229: I still think we need to re-introduce this check elsewhere. Despite your changes, the promise implementation today is not prepared to allow self-cancellation from application code (I checked and several callers of `get(ExceptionOrValue& output)` would crash if it caused `this` to be deleted).

Maybe the easiest thing to do is:
* Have `~Event()` throw if the event is currently pointed to by `threadLocalCurrentlyFiring`.
* Add a method `permitSelfDestruction()` that unsets the event as `threadLocalCurrentlyFiring`, and call it in the specific places where we know it is safe to delete the event.

c++/src/kj/async.c++
```diff
@@ -2297,10 +2232,10 @@ String TraceBuilder::toString() {
 ArrayPtr<void* const> getAsyncTrace(ArrayPtr<void*> space) {
   EventLoop* loop = threadLocalEventLoop;
   if (loop == nullptr) return nullptr;
```
kentonv: Seems like you can delete these lines now as the loop isn't needed.

c++/src/kj/async.c++
mikea: Done. Introduced permitSelfDestruction. Without the call deleting in fire() will result Promise callback destroyed itself. I made a change in a separate commit for easier review.

mikea:
all done, ptal @kentonv

kentonv:
> mikea force-pushed the maizatskyi/2026-01-05-event-fire branch from 3f65766 to 80bfeb5 

Contains both a rebase and new changes, hard to incrementally review. :/

c++/src/kj/async.c++
```diff
@@ -2197,74 +2170,44 @@ kj::EventLoop& Event::requireEventLoop() {
...
-  // any instructions, it just blocks compiler optimizations.
-  std::atomic_signal_fence(std::memory_order_acq_rel);
-
+  if (KJ_UNLIKELY(threadLocalCurrentlyFiring == this)) {
+    KJ_FAIL_REQUIRE("Promise callback destroyed itself.");
+  }
   disarm();
+}
 
-  // If this fails, we'll abort due to `noexcept`. That's good because otherwise we're likely to
```
kentonv, line 2227: Maybe transfer this comment to the new assertion.

kentonv (approved):

mikea:
re-measured performance after rebase: https://gist.github.com/mikea/7151c8df99baba86e3afd73e1730338b
-4.6% on HttpFullProtocol

PR #2492, opened 2026-01-06
WIP: disable unwindDetector in CoroutineBase::destroy

mikea:
** not suggesting we do this, just measuring the impact **


mikea:
codspeed is completely wrong here. bm_Promise_ReadyNow codepath isn't even touched

PR #2493, opened 2026-01-06
[coro] explicit coroutine allocator argument

mikea:
This introduces `CoroutineAllocator` marker type and parametrizes `Coroutine` by `Allocator`. This will allow me to experiment with allocators (PRs coming) without any performance overhead before enabling anything by default.

Adds `DebugCoroutineAllocator` to be able to perform coroutine size measurements. Our current empty coro is 192 bytes.

I tried to put as much code as possible into new header not to pile up on async-inl.h LMK if you want me to rearrange.

performance neutral: https://gist.github.com/mikea/2749fff1fd409fec57e154696b05604f

### DIFF
--- a/c++/src/kj/async-coroutine-alloc.h
+++ b/c++/src/kj/async-coroutine-alloc.h
@@ -1,0 +1,105 @@
+// Copyright (c) 2026 Cloudflare, Inc. and contributors
+// Licensed under the MIT License:
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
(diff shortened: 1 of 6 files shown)


c++/src/kj/async-coroutine-alloc.h
```diff
@@ -0,0 +1,104 @@
...
+
+#pragma once
+
+#include <kj/async.h>
+
+#include <new> // for placement new
+
+namespace kj {
+
+template <typename Allocator> struct CoroutineFrame {
```
kentonv:
```suggestion
template <typename Allocator>
struct CoroutineFrame {
```

c++/src/kj/async-inl.h
```diff
@@ -2458,6 +2527,32 @@ class Coroutine final: public CoroutineBase,
...
+      return CoroutineAllocator::getAllocator(args...).alloc(frameSize);
+    } else {
+      return Allocator::alloc(frameSize);
+    }
+  }
+
+
+#if defined(__clang__) && __clang_major__ > 18
+  #define KJ_ASYNC_CORO_SIZED_DELETE 1
+  // Sized delete for coroutines are supported since clang-19
```
kentonv: Did you check if GCC and MSVC support it? If they do, maybe we should just require new enough versions of all compilers rather than maintain separate code?

mikea: MSVC is what broke when I didn't have this. It seems that `#if defined(__cpp_sized_deallocation)` works too and is much more elegant so I don't mind keeping it around.

PR #2495, opened 2026-01-07
Implement coroutine-friendly KJ_TRY / KJ_CATCH macros

harrishancock:
This implements separate macros of the forms:

```cpp
KJ_TRY {
  // throws
} KJ_CATCH(e) {
  // handle
}
```

The big killer features over our existing utilities are that:

- It coerces all exceptions to kj::Exception for you.
- You can co_await in both the "try" and the "catch" blocks, if you're in a coroutine.
- You can use other standard control flow keywords in either block, such as `return`ing from the enclosing function, `break`ing an outer loop, etc.

The macros rely on a trick: you can `goto` from an `if` statement's true branch to its false branch. This allows us (the implementor) to execute a compound statement inline first, like a `try/catch` block, and only conditionally jump to the user's provided subsequent block. I learned the trick from this article by Simon Tatham: https://www.chiark.greenend.org.uk/~sgtatham/mp/

The separate macro approach requires hard-coding a constant identifier to serve as the exception storage (`_kjTryCatchException`). This causes benign variable shadowing in nested KJ_TRY/KJ_CATCH blocks, which we suppress with compiler-specific warning pragmas.

I did some casual benchmarking on an aarch64 machine (my Pixel 9a... it's what I had available at the time). The happy path appears to have no overhead. The sad path seems to incur 0% - 5% overhead. If we inline the implementation of kj::getCaughtExceptionAsKj() into the macro, then the sad path overhead instead becomes negative 50% -- it's twice as fast. But, I don't know what impact that may have on code size.

I tried optimizing it further by storing the exception in an anonymous union and setting up a dtor-only RAII guard whose lifetime encompasses the user's catch handler, but this showed no advantage over the existing Maybe-based solution. In most of my benchmark runs, in fact, the anonymous union / dtor RAII guard experiment was slower than just using a Maybe<Exception>.

Claude wrote all of the tests and did most of my typing for me.

### DIFF
--- a/c++/src/kj/async-inl.h
+++ b/c++/src/kj/async-inl.h
@@ -1573,9 +1573,9 @@ inline PromiseForResult<Func, void> evalLast(Func&& func) {
 template <typename Func>
 inline PromiseForResult<Func, void> evalNow(Func&& func) {
   PromiseForResult<Func, void> result = nullptr;
-  KJ_IF_SOME(e, kj::runCatchingExceptions([&]() {
+  KJ_TRY {
     result = func();
-  })) {
+  } KJ_CATCH(e) {
     result = kj::mv(e);
   }
   return result;
@@ -1783,21 +1783,23 @@ private:
 template <typename T>
 template <typename Func>
 bool PromiseFulfiller<T>::rejectIfThrows(Func&& func) {
-  KJ_IF_SOME(exception, kj::runCatchingExceptions(kj::mv(func))) {
+  KJ_TRY {
+    func();
+    return true;
+  } KJ_CATCH(exception) {
     reject(kj::mv(exception));
     return false;
-  } else {
-    return true;
   }
 }
 
 template <typename Func>
 bool PromiseFulfiller<void>::rejectIfThrows(Func&& func) {
-  KJ_IF_SOME(exception, kj::runCatchingExceptions(kj::mv(func))) {
+  KJ_TRY {
+    func();
+    return true;
+  } KJ_CATCH(exception) {
     reject(kj::mv(exception));
     return false;
-  } else {
-    return true;
   }
 }
 
--- a/c++/src/capnp/compat/http-over-capnp.c++
+++ b/c++/src/capnp/compat/http-over-capnp.c++
@@ -181,14 +181,13 @@ private:
       }
     });
 
-    try {
+    KJ_TRY {
       KJ_IF_SOME(ws, webSocket) {
         co_await canceler.wrap(func(ws));
       } else {
         kj::throwFatalException(KJ_EXCEPTION(DISCONNECTED, "request canceled"));
       }
-    } catch (...) {
-      auto e = kj::getCaughtExceptionAsKj();
+    } KJ_CATCH(e) {
       error = kj::heap(kj::cp(e));
       kj::throwFatalException(kj::mv(e));
     }
@@ -595,16 +594,15 @@ public:
     }
 
     // Finish pumping the response or WebSocket. (Probably it's already finished.)
-    try {
+    KJ_TRY {
       co_await context.finishPump();
-    } catch (...) {
+    } KJ_CATCH(exception) {
       // Ignore DISCONNECTED exceptions from this pump, because it should have been the server's
       // responsibility to propagate any exceptions from pushing the response. If this were a local
       // HttpService (with no RPC layer), such exceptions would not propagate here, so we want to
       // do the same. Actually, technically, even non-DISCONNECTED exceptions arguably shouldn't
       // propagate here for the same reason. But, non-DISCONNECTED exceptions are more likely to
       // flag some real bug, so I'm leaving them alone for now. This could be revisited later.
-      auto exception = kj::getCaughtExceptionAsKj();
       if (exception.getType() != kj::Exception::Type::DISCONNECTED) {
         kj::throwFatalException(kj::mv(exception));
       }
--- a/c++/src/capnp/rpc.c++
+++ b/c++/src/capnp/rpc.c++
@@ -3879,10 +3879,9 @@ private:
       uint64_t interfaceId, uint64_t methodId,
       kj::Own<ClientHook>&& capability, kj::Own<CallContextHook>&& context,
       ClientHook::CallHints hints) {
-    try {
+    KJ_TRY {
       return capability->call(interfaceId, methodId, kj::mv(context), hints);
-    } catch (...) {
-      auto exception = kj::getCaughtExceptionAsKj();
+    } KJ_CATCH(exception) {
       auto pipeline = newBrokenPipeline(kj::cp(exception));
 
       // In the past, an exception here would have killed the entire connection (and also caused
--- a/c++/src/kj/compat/http-test.c++
+++ b/c++/src/kj/compat/http-test.c++
@@ -4353,12 +4353,11 @@ KJ_TEST("HttpServer bad requests") {
 
     auto request = testCase.request;
     auto writePromise = pipe.ends[1]->write(request.asBytes());
-    try {
+    KJ_TRY {
       auto response = pipe.ends[1]->readAllText().wait(waitScope);
       auto expectedResponse = testCase.expectedResponse;
       KJ_EXPECT(expectedResponse == response, expectedResponse, response);
-    } catch (...) {
-      auto ex = kj::getCaughtExceptionAsKj();
+    } KJ_CATCH(ex) {
       KJ_FAIL_REQUIRE("not supposed to happen", ex);
     }
 


c++/src/kj/compat/http.c++
```diff
@@ -3828,14 +3828,18 @@ static kj::Promise<void> pumpWebSocketLoop(WebSocket& from, WebSocket& to) {
       }
       // continue the loop
     }
-  } catch (...) {
+  }) {
     // We don't know if it was a read or a write that threw. If it was a read that threw, we need
     // to send a disconnect on the destination. If it was the destination that threw, it
     // shouldn't hurt to disconnect() it again, but we'll catch and squelch any exceptions.
-    kj::runCatchingExceptions([&to]() { to.disconnect(); });
+    KJ_TRY_CATCH(exception [[maybe_unused]], {
```
harrishancock: We might want a `KJ_TRY_CATCH_AND_IGNORE(code)` variant, as the `name [[maybe_unused]]` junk is a lot of extra typing.

c++/src/kj/exception.h
```diff
@@ -319,6 +319,49 @@ KJ_NOINLINE void throwRecoverableException(kj::Exception&& exception, uint ignor
 
 // =======================================================================================
 
+// KJ_TRY_CATCH macro
+//
+// A convenient macro for exception handling that automatically coerces all exceptions to kj::Exception.
+// This can replace both raw try/catch(...) and runCatchingExceptions() usage.
+//
+// Use like so:
+//   KJ_TRY_CATCH(e, {
```
kentonv: Hmm, I think you could make this macro work like:

```
KJ_TRY {
  ...
} KJ_CATCH(err) {
  ...
}
```

You just split the macro at the `__VA_ARGS__`...

Instead of `KJ_UNIQUE_NAME(storage)` you'd have to use something like `_kjTryCatchException`, but I think that's OK since it's scoped to the if() block and a nested try will shadow it which is fine. You can still use `KJ_UNIQUE_NAME(handler)` for the goto label since it's entirely in the `KJ_CATCH` part.

Would be more intuitive syntax I think?

harrishancock: Definitely. I initially started with that version, in fact, but ended up liking the single macro version more, though I don't feel strongly either way.

In addition to the unique name wart (we'd have to inject some disable-warning directives, but doable), clang-format will format it like:

```cpp
KJ_TRY {
  ...
}
KJ_CATCH {
  ...
}
```

We could submit a patch to clang-format, though, to teach it to treat certain macros as `try` and `catch` keywords.

harrishancock:
MSVC seems to have spotted a legitimate problem: https://github.com/capnproto/capnproto/actions/runs/20785170635/job/59692560780?pr=2495#step:5:721

Given:

```cpp
KJ_TRY_CATCH(e, {
  KJ_TRY_CATCH(e2, {
  }) {]
}) {}
```

Both KJ_TRY_CATCH macros are instantiated on the same line, so their handler labels conflict. Interesting that this works, with tests passing under clang.

**Edit:** Pushed an attempt at a fix using \_\_COUNTER\_\_ instead of KJ_UNIQUE_NAME: https://github.com/capnproto/capnproto/pull/2495/commits/c85e99d319e9955e2e2e04a4a3ca8dd7297836d2

c++/src/kj/compat/http.c++
```diff
@@ -3907,14 +3907,18 @@ static kj::Promise<void> pumpWebSocketLoop(WebSocket& from, WebSocket& to) {
       }
       // continue the loop
     }
-  } catch (...) {
+  } KJ_CATCH(pumpError) {
```
harrishancock: I wonder if we should have a `KJ_CATCH_AND_RETHROW`?

```cpp
KJ_TRY {
  throw exception;
} KJ_CATCH_AND_RETHROW {
  // cleanup code
}
// unreachable
```

I guess it's equivalent to
```cpp
KJ_ON_SCOPE_FAILURE(cleanup code);
throw exception;
```

c++/src/kj/exception-test.c++
```diff
@@ -313,6 +314,330 @@ KJ_TEST("copy constructor") {
...
+  } KJ_CATCH_AND_IGNORE {
+    cleanupCalled = true;
+    cleanupValue = 42;
+  }
+
+  KJ_EXPECT(cleanupCalled);
+  KJ_EXPECT(cleanupValue == 42);
+}
+
+KJ_TEST("KJ_TRY/KJ_CATCH does not catch CanceledException") {
```
harrishancock, line 501: Note that there is CanceledException and CleanShutdownException test coverage for both KJ_CATCH and KJ_CATCH_AND_IGNORE.

c++/src/kj/exception.h
```diff
@@ -319,6 +319,108 @@ KJ_NOINLINE void throwRecoverableException(kj::Exception&& exception, uint ignor
...
+      catch (...) { \
+        _kjTryCatchException_8908013cdf4952df = ::kj::getCaughtExceptionAsKj(); \
+        goto KJ_UNIQUE_NAME(_kjTryCatchHandler); \
+      } \
+    else \
+      KJ_UNIQUE_NAME(_kjTryCatchHandler): \
+      if (auto& exception = *::kj::_::readMaybe(_kjTryCatchException_8908013cdf4952df); false) {} \
+      else
+
+// TODO(now): Bikeshed the name.
```
harrishancock: Thoughts?

c++/src/kj/compat/http.c++
jasnell: For now I'd skip it. We can add it later to eliminate boilerplate but for now let's err on the side of less.

c++/src/kj/compat/http.c++
```diff
@@ -3907,14 +3907,18 @@ static kj::Promise<void> pumpWebSocketLoop(WebSocket& from, WebSocket& to) {
...
     // We don't know if it was a read or a write that threw. If it was a read that threw, we need
     // to send a disconnect on the destination. If it was the destination that threw, it
     // shouldn't hurt to disconnect() it again, but we'll catch and squelch any exceptions.
-    kj::runCatchingExceptions([&to]() { to.disconnect(); });
+    KJ_TRY {
+      to.disconnect();
+    } KJ_CATCH_AND_IGNORE {
+      // Squelch.
+    }
```
jasnell: Is the trailing block `{ ... }` necessary here? If we're ignoring it would be nice to not require the additional empty block.

c++/src/kj/exception.h
jasnell: the name is fine to me.

c++/src/kj/compat/http.c++
harrishancock: It could also just be a semicolon, like

```cpp
    KJ_TRY {
      to.disconnect();
    } KJ_CATCH_AND_IGNORE;
```

c++/src/kj/exception.h
```diff
@@ -319,6 +319,108 @@ KJ_NOINLINE void throwRecoverableException(kj::Exception&& exception, uint ignor
...
+
+// TODO(soon): Inline getCaughtExceptionAsKj()'s logic here and use KJ_TRY / KJ_CATCH to implement
+//   getCaughtExceptionAsKj(). This should reduce sad path overhead by 50% (no re-throw), but may
+//   increase code size. Experiment with this after KJ_TRY / KJ_CATCH has been adopted at large.
+#define KJ_CATCH(exception) \
+      catch (...) { \
+        _kjTryCatchException_8908013cdf4952df = ::kj::getCaughtExceptionAsKj(); \
+        goto KJ_UNIQUE_NAME(_kjTryCatchHandler); \
+      } \
+    else \
```
jclee, line 552: Hmm... Does this end up manifesting as a dangling-else warning, if the try/catch appears after an `if(...) {...}` block in code?  I know we've needed suppression for such warnings in [other macro constructions](https://github.com/capnproto/capnproto/blob/07764022/c%2B%2B/src/kj/common.h#L1359).

jclee: Oh... I suppose it would have to be something like `if (...) KJ_TRY { ... } KJ_CATCH (...) { ...}`, which seems like an unlikely code pattern, and one that we'd want to discourage anyway.

c++/src/kj/compat/http.c++
kentonv: TBH I prefer just KJ_ON_SCOPE_FAILURE.

If we had KJ_CATCH_AND_RETHROW then we should have KJ_FINALLY, too... but again, KJ_DEFER seems strictly better.

c++/src/kj/compat/http.c++
```diff
@@ -3907,14 +3907,18 @@ static kj::Promise<void> pumpWebSocketLoop(WebSocket& from, WebSocket& to) {
...
 
     // In any case, this error broke the pump. We should propagate it out as the pump result.
-    throw;
+    throw kj::mv(pumpError);
```
kentonv: Ah this is actually a regression here.

A kj::Exception should never be thrown "raw" with `throw` like this. You need to use `kj::throwFatalException()`, as it adds the stack trace, and does some other bookkeeping.

You might say "but we just caught this exception, it already has a stack trace". Not quite: kj::getCaughtExceptionAsKj() actually _removes_ the prefix of the stack that matches the catch location, deeming it as irrelevant. But if you rethrow it raw, then that part of the stack is lost.

(There is also the "other bookkeeping"... check out the code in throwFatalException() for details.)

The previous code here, though, was fine: using `throw` with no argument to rethrow the caught exception maintains the stack trace and all the bookkeeping works as intended. It's also more efficient that way, vs. removing the stack and re-adding it. IMO we should just leave it as it was.

c++/src/kj/exception-test.c++
```diff
@@ -21,6 +21,7 @@
 
 #include "exception.h"
 #include "debug.h"
+#include "main.h"
```
kentonv, line 24: Why this include?

c++/src/kj/exception.h
```diff
@@ -319,6 +319,108 @@ KJ_NOINLINE void throwRecoverableException(kj::Exception&& exception, uint ignor
...
+#endif
+
+// Since we have two macros -- KJ_TRY and KJ_CATCH -- which must both access the same exception
+// storage variable, we must choose a hard-coded name for it. This will cause variable shadowing in
+// nested KJ_TRY/KJ_CATCHes, but that is benign, so we disable shadowing warnings. We suffix the
+// variable name with a randomly-chosen number to make collision with user code virtually
+// impossible.
+#define KJ_TRY \
+    KJ_SILENCE_SHADOWING_BEGIN \
+    if (::kj::Maybe<::kj::Exception> _kjTryCatchException_8908013cdf4952df; true) \
```
kentonv: The `_kj` prefix is already (informally) reserved for KJ so the suffix isn't _really_ needed.

c++/src/kj/exception.h
kentonv: Do we really need it? Maybe people should just write `KJ_CATCH(_)` (which I think suppresses the unused variable warning)?

harrishancock: You're right, `_` suppresses the warning! I didn't realize that. Will delete this.

c++/src/kj/compat/http.c++
harrishancock: Nice catch. I'll revert this.

It occurred to me that there is actually a second, subtle behavior change here: previously, `to.disconnect()` would run for all exceptions, including `CanceledException` and `CleanShutdownException`. With my change here, those exceptions would be rethrown before `to.disconnect()` could run.

To compare the various "rethrow" patterns we've discussed in this PR:

```cpp
// Exhibit A, try/catch
try {
  // throw
} catch (...) {
  // code RUNS during fiber/proces cancellation
  throw;
}

// Exhibit B, KJ_TRY/KJ_CATCH
KJ_TRY {
  // throw
} KJ_CATCH(e) {
  // code does NOT run during fiber/process cancellation
  kj::throwFatalException(kj::mv(e));
}

// Exhibit C, KJ_ON_SCOPE_FAILURE
{
  KJ_ON_SCOPE_FAILURE(/* code RUNS during fiber/process cancellation */);
  // throw
}
```

I'm not sure it matters very much, but it's an interesting quirk. I've documented the behavior in `KJ_CATCH`'s doc comment.

c++/src/kj/exception-test.c++
harrishancock: To test that `TopLevelProcessContext::CleanShutdownException` is not caught by `KJ_CATCH`. See the "KJ_TRY/KJ_CATCH does not catch CleanShutdownException" test case.

c++/src/kj/exception.h
harrishancock: Removed.

c++/src/kj/async-inl.h
```diff
@@ -1783,21 +1783,23 @@ class PromiseAndFulfillerAdapter {
 template <typename T>
 template <typename Func>
 bool PromiseFulfiller<T>::rejectIfThrows(Func&& func) {
-  KJ_IF_SOME(exception, kj::runCatchingExceptions(kj::mv(func))) {
+  KJ_TRY {
+    func();
```
harrishancock, line 1787: This is technically a subtle behavior change. Previously, `func` would be moved, then invoked as an lvalue. Now, I think it's simply invoked as an lvalue.

We could preserve the previous behavior with
```cpp
auto f = kj::mv(func);
f();
```

But I doubt it matters.

harrishancock:
Addressed all comments, added a migration guide in the doc comment for the macros, and a little mention in kjdoc/tour.md. PTAL.

harrishancock:
CodSpeed shows a slight regression in [bm_Http_OverCapnpFullRPC](https://codspeed.io/capnproto/capnproto/branches/harris%2F2026-01-07-kj-try-catch-macro?uri=src%2Fcapnp%2Fcompat%2Fhttp-over-capnp-bench.c%2B%2B%3A%3Abm_Http_OverCapnpFullRPC&runnerMode=Instrumentation&utm_source=github&utm_medium=comment-v2&utm_content=benchmark), and I did touch that code with some KJ_TRY conversions in my latest push.

I'm going to retry my anonymous union optimization, which should make the KJ_TRY happy path completely branchless, and see if that helps.

c++/src/kj/exception.h
harrishancock: I tested this, and you're right, the warning pops up with that syntax.

I agree, we would want to discourage its use anyway, so I'll leave the warning unsuppressed, to encourage authors to reconsider.

harrishancock:
CodSpeed didn't flag the latest push with the anonymous union, but when I go into that benchmark's dashboard, I still see a slight regression. But I don't really know how to read CodSpeed.

I'm going to try reverting the conversions I made in http-over-capnp.c++ and see if that makes it go back to baseline. If it does, then that's a bit concerning for the future of this macro. :/

harrishancock:
Uggghh, reverting the http-over-capnp.c++ conversions made the CodSpeed benchmark go back to baseline.

I re-pushed the conversions so @kentonv can take a look: https://github.com/capnproto/capnproto/pull/2495/commits/82feca55217bc56a130e3d11180d089468011cb3#diff-3c48d784e4a67f5640101e8f8a5662b963e9cb5d7f5f30c0b587dec440596b7f

I wonder if that benchmark exercises exception-handling code paths more heavily than happy paths. I did notice locally that exception-handling was slightly slower with the new macros. If so, inlining the `getCaughtExceptionAsKj()` logic could help.

c++/src/kj/async-inl.h
kentonv: There are three uses of `rejectIfThrows` across the entirety of the Cap'n Proto and Workers Runtime codebases and they all look fine. (Also they are all in tests.)

kentonv (approved):

harrishancock:
Latest push is a rebase on origin/v2, minor reordering of commits, and revert from the anonymous union "optimized" implementation back to the original, simpler Maybe-based implementation.

CodSpeed now shows 0% perf change on bm_Http_OverCapnpFullRPC with the latest push. I separately analyzed the change in cachegrind, and there are no smoking guns pointing at the KJ_TRY / KJ_CATCH macro implementation. So, I'll go ahead and merge.

mikea:
The utility is nice, but it increases the size of coro frames by 8 bytes for every KJ_TRY usage. I believe this is why we see more malloc pressure when we run benchmarks under cachegrind.

Is there any way to get rid of this memory overhead? I am working really hard to reduce the size of coro frames, and this goes in the opposite direction.

kentonv:
> Is there any way to get rid of this memory overhead?

Since `Exception` just contains `Own<Storage>`, we could optimize `Maybe<Exception>` the same way we've optimized `Maybe<Own<T>>`, so that it uses nullness of the pointer instead of storing a separate boolean (where that boolean gets padded to 8 bytes).

harrishancock:
There's another alternative implementation with zero coroutine frame overhead:

```cpp
#define KJ_TRY                                                                                     \
  try {                                                                                            \
    try

#define KJ_CATCH(name)                                                                             \
    catch (...) {                                                                                  \
      ::kj::throwFatalException(::kj::getCaughtExceptionAsKj());                                   \
    }                                                                                              \
  } catch (::kj::Exception& name)
```

I didn't like it mostly because of the nested try-catch, which I thought would be high overhead. However, that overhead probably only appears on the sad path. I re-ran cachegrind on this implementation, and http-over-capnp-bench shows no new malloc pressure, and essentially identical instruction counts.

A downside is that you can't `co_await` from the `catch` block in this implementation, but if we really want syntax sugar for that, we could make a `KJ_CO_TRY` / `KJ_CO_AWAIT` which uses the as-merged technique.

Edit: I also like the idea of optimizing `Maybe<Exception>`. It occurs to me that we could probably define a concept, `MaybeNoneOptimizable`, for types which implement a special niche value that `Maybe` can use as a `kj::none` marker. Maybe a constructor that accepts `kj::none` and `operator==(kj::none)`? Then we could replace the `Maybe<Own<T>>` specialization with a partial specialization on types of that concept.

harrishancock:
I tried optimizing `Maybe<Exception>` to 4 bytes over here: https://github.com/capnproto/capnproto/pull/2514

Benchmark results: https://gist.github.com/harrishancock/55619b744a8857343767a0157472ff54

cachegrind diff shows that optimizing `Maybe<Exception>` to 4 bytes essentially eliminates the malloc overhead. On some measurements (e.g. I refs), the KJ_TRY in http-over-capnp with 4-byte `Maybe<Exception>` actually beats naked `try` with any size `Maybe<Exception>`. I don't have any theories as to why.

kentonv:
How can anything be optimized to 4 bytes when pointers are 8 bytes?

harrishancock:
Heh, I think my brain went back in time to 32-bit mode. I meant 8-bytes and 12-bytes, or whatever the bool forces padding to.

kentonv:
Note that `Own<T>` is two pointers, so I think you actually have 16 bytes?

PR #2496, opened 2026-01-08
Implement HttpFixedLengthEntityReader::pumpTo().

kentonv:
This enables optimize pumps (using splice() on Linux) in the most important use case imaginable: proxying large HTTP response bodies through Workers.

I totally thought this optimization was already happening. I remember measuring the performance gain -- in Workers -- when I implemented it. But it turns out... it never worked, because we were missing this small amount of code. :(

Per @mhart (see below), this optimization reduces the CPU overhead for streaming a 300MB response body from ~1s to ~0.

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -1663,6 +1663,28 @@ public:
     }
   }
 
+  Promise<uint64_t> pumpTo(AsyncOutputStream& output, uint64_t amount) {
+    KJ_REQUIRE(onMessageDone != kj::none);
+
+    if (leftover == nullptr) {
+      // No leftovers. Forward directly to inner stream.
+      co_return co_await inner.pumpTo(output, amount);
+    } else if (leftover.size() >= amount) {
+      // Didn't even read the entire leftover buffer.
+      co_await output.write(leftover.first(amount).asBytes());
+      leftover = leftover.slice(amount, leftover.size());
+      co_return amount;
+    } else {
+      // Read the entire leftover buffer, plus some.
+      co_await output.write(leftover.asBytes());
+      size_t firstWrite = leftover.size();
+      leftover = nullptr;
+
+      auto rest = co_await inner.pumpTo(output, amount - firstWrite);
+      co_return firstWrite + rest;
+    }
+  }
+
   enum RequestOrResponse {
     REQUEST,
     RESPONSE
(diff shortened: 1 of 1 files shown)


kentonv:
Note: I verified manually that workerd pass-through pumps are now optimized with this change, but weren't before.

It's tricky to write a test since it's an optimization, and even though I could perhaps test that the HTTP layer itself passes through pumpTo() correctly, what I really want to prove is that the optimization works end-to-end. So a test in the workers runtime somewhere is more appropriate.

(Needless to say if this code is _wrong_, all kinds of existing workers tests will break, so I'm not worried about that... worried about testing that the optimization actually kicks in.)

mhart:
> Merging this PR will **improve performance by 1.15%**

I wonder if another benchmark is needed here – or whether it should be in workerd? I see an improvement from 1370ms to 303ms on a 300MB file benchmark (and even then, that 303ms is either http client or origin server overhead as I get the same result running the benchmark without workerd in the middle)

c++/src/kj/compat/http.c++
```diff
@@ -2053,9 +2075,66 @@ public:
...
+    auto actual = co_await getInner().pumpTo(output, amount);
+    length -= actual;
+
+    if (length == 0) {
+      doneReading();
+    } else if (actual < amount) {
+      // We hit EOF before pumping what was expected, but this means the stream ended prematurely
+      // without reaching the expected content-length, so throw an exception instead.
+      size_t expectedLength = length + actual;
+      kj::throwRecoverableException(KJ_EXCEPTION(
```
harrishancock, line 2127: Non-blocking, but I'm curious, why recoverable? It looks like the default recoverable exception callback logs errors rather than throws them if a current exception is in-flight, which is the main difference compared to fatal exceptions. But how would that come about? We're after a `co_await`, meaning a caller would have to resume this coroutine -- during an exception unwind -- to get here.

kentonv: This is copy-pasted from tryRead(), above (line 2055).

KJ no longer supports no-exceptions mode, so fatal vs. recoverable are equivalent (except in the destructor-during-unwind case, which as you note doesn't apply here).

With that said, in theory, if we did still support no-exceptions, then "recoverable" is the right choice here because it's safe to continue execution through the co_return below, at which point the async framework will pick up the exception via the exception callback and propagate it.

PR #2501, opened 2026-01-12
Use more precise AnyList and AnyStruct in the Value schema

RReverser:
This changes Value.list and Value.struct from AnyPointer to its narrower counterparts.

This improves type safety for external codegens, but of course requires some changes in C++ as well. In particular, I'm not 100% sure about my  `getValuePointer` approach to getting the default value as non-dereferenced raw pointer again after this change, if there's a better way, I'm happy to change it.

As far as I can tell, this schema change is backward-compatible and no renumbering is necessary.


c++/src/capnp/pointer-helpers.h
```diff
@@ -149,6 +149,13 @@ struct PointerHelpers<UncheckedMessage> {
   static inline const word* get(PointerReader reader) {
     return reader.getUnchecked();
   }
+
+  template <typename T>
+  static inline const word* getValuePointer(const T& value) {
```
kentonv, line 156: `PointerHelpers<UncheckedMessage>` is structured as a specialization of `PointerHelpers` specifically as a trick to make `AnyPointer::getAs<UncheckedMesasge>()` "just work".

But you've eliminated all uses of that in this PR, so we can remove `PointerHelpers<UncheckedMessage>` entirely.

I guess the reason you've shoved `getValuePointer()` here is because `PointerHelpers` is a friend of all struct readers so can access `value._reader`. But it's weird to shove a helper function here that isn't actually related to the `PointerHelpers` template.

I think instead of this helper you could write code like:

`AnyStruct::Reader(dval).getPointerSection()[0].getAs<Unchecked>()`

(This actually relies on the existing `PointerHelpers<UncheckedMessage>` so in that case it can't be removed, but the new method is no longer needed.)

c++/src/capnp/dynamic.c++
```diff
@@ -234,15 +234,14 @@ DynamicValue::Reader DynamicStruct::Reader::get(StructSchema::Field field) const
           auto elementType = type.asList().getElementType();
           return DynamicList::Reader(type.asList(),
               reader.getPointerField(assumePointerOffset(slot.getOffset()))
-                    .getList(elementSizeFor(elementType.which()), dval.isAnyPointer() ? nullptr :
-                        dval.getList().getAs<_::UncheckedMessage>()));
+                    .getList(elementSizeFor(elementType.which()),
+                        _::PointerHelpers<_::UncheckedMessage>::getValuePointer(dval)));
```
kentonv, line 238: The behavior here seems to have changed in the case of `dval.isAnyPointer()`, where previously we returned `nullptr`?

Though the old behavior on its face looks wrong, and I can't remember any reason for it to be that way... is that your thinking, that this is an improvement?

RReverser: IIRC my thinking was that here we just want to get an unchecked pointer, and it shouldn't matter which subkind it came from - if it's `nullptr`, `getValuePointer` will return that to us.

It's a bit murky now (I should've returned to this PR sooner) so I'm happy to revert if it looks suspicious.

c++/src/capnp/pointer-helpers.h
RReverser: > But it's weird to shove a helper function here that isn't actually related to the `PointerHelpers` template.

Well this new helper function is literally a pointer helper, so it made sense semantically to put it inside a struct literally called "pointer helpers".

But yeah I can go a step further and remove it I guess.

c++/src/capnp/dynamic.c++
RReverser: Ah no, wait, I think it was specifically because we now know precise subtypes instead of `getList()` etc all being anypointers so we don't need to check for `.isAnyPointer()` separately?

c++/src/capnp/pointer-helpers.h
RReverser: > I think instead of this helper you could write code like:
> 
> `AnyStruct::Reader(dval).getPointerSection()[0].getAs<Unchecked>()`

I tried doing this, but it looks pretty verbose to repeat that everywhere IMO compared to the helper.

RReverser:
Btw another alternative to this PR could be to eliminate struct/list/interface variants from `Value` altogether - `AnyPointer` type already has its own nested information about which kind of pointer it is, so having variants at the `Value` level as well seems redundant, `Value.anyPointer` variant would be enough to represent all of them.

Or vice versa, remove `Value.anyPointer` and keep only the strongly-typed variants - a bit more cumbersome but also avoids redundancy.

But I guess it's too much of a breaking change either way.

c++/src/capnp/dynamic.c++
kentonv: Actually now I see, there's a comment above saying that dval may be `anyPointer` in certain cases with generics, but it'll always be null then.

I agree that `getValuePointer()` should handle it correctly.

c++/src/capnp/pointer-helpers.h
kentonv: OK well this needs some commentary:

> Assuming `T` is `schema::Value::Reader` for a pointer-typed value, return a (native) pointer pointing at the value's (capnp) pointer, which can then be used as an unchecked message. This is used in default value handling in `dynamic.c++`.
>
> This is wrapped in a one-of specialization of `PointerHelpers` primarily to make it a "friend" of all pointer types, which is needed specifically so that it can become a friend of `schema::Value::Reader`, in order to extract its underlying `StructReader`.
>
> TODO(cleanup): This is quite a hack and there's probably a better way to structure this but we're too lazy to come up with something at the moment.

Also, if `get(PointerReader)` is no longer used can we remove it?

RReverser: > Also, if `get(PointerReader)` is no longer used can we remove it?

Oh of course, good catch.

Added the comment too.

kentonv (approved):

PR #2502, opened 2026-01-12
Fix spurious exceptions in kj::ExceptionCallback when using HWASan

### LINKED ISSUE #2500: Creating an exception handler aborts when using HWASan
`kj::ExceptionCallback` throws an exception if it detects that the callback is not stack allocated. This check is disabled when ASan is enabled, but not when [HWASan](https://clang.llvm.org/docs/HardwareAssistedAddressSanitizerDesign.html) is enabled, leading to an abort despite a correct usage:

```
Abort message: 'terminating due to uncaught exception of type kj::ExceptionImpl: kj/exception.c++:1040: failed: expected offset < 65536 && offset > -65536; ExceptionCallback must be allocated on the stack.'
```

Not entirely sure how this should be correctly detected, in my case (compiling for Android) it seems like `__has_feature(hwaddress_sanitizer)` is true, but `__SANITIZE_ADDRESS__` and `__SANITIZE_HWADDRESS__` are undefined. Changing the check to this fixed the crash:

```cpp
#if defined(FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION) || \
    KJ_HAS_COMPILER_FEATURE(address_sanitizer) || \
    KJ_HAS_COMPILER_FEATURE(hwaddress_sanitizer) || \
    defined(__SANITIZE_ADDRESS__)
```

dankmeme01:
Fixes #2500, HWASan on Android does not seem to define the regular ASan macros / features, and this is the only way I found to detect it

### DIFF
--- a/c++/src/kj/exception.c++
+++ b/c++/src/kj/exception.c++
@@ -1281,6 +1281,7 @@ thread_local ExceptionCallback* threadLocalCallback = nullptr;
 void requireOnStack(void* ptr, kj::StringPtr description) {
 #if defined(FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION) || \
     KJ_HAS_COMPILER_FEATURE(address_sanitizer) || \
+    KJ_HAS_COMPILER_FEATURE(hwaddress_sanitizer) || \
     defined(__SANITIZE_ADDRESS__)
   // When using libfuzzer or ASAN, this sanity check may spurriously fail, so skip it.
 #else


kentonv:
Thanks!

Do you care about this being backported to the next 1.x release?

dankmeme01:
That would be great, yeah :)

kentonv:
This will be in the 1.4.0 release which should go out tomorrow.

PR #2503, opened 2026-01-12
[coro] optimize unwind detection in CoroutineBase

mikea:


c++/src/kj/async-inl.h
```diff
@@ -2351,8 +2351,6 @@ class CoroutineBase: public PromiseNode,
   // ability to destroy themselves, replacing their own Own, we hold a reference to the owning Own
   // instead of directly to the PromiseNode.
 
-  UnwindDetector unwindDetector;
```
kentonv, line 2354: There's also one in PromiseAwaiterBase, should that be removed too?

mikea: https://github.com/capnproto/capnproto/pull/2490

mikea:
codspeed is drunk, bm_Promise_Shift_20 is not touched at all. I'm feeling we need to disable it since it is very noisy _despite_ being a simulator.

PR #2504, opened 2026-01-12
Revert "[coro] remove unwind detector from PromiseAwaiter"

mikea:
Reverts capnproto/capnproto#2490


mikea:
> FWIW I think it's extremely unlikely that this impacts any real use case, so I could be convinced that it's fine and worth it.
> 
> Another thing you could do is check if `node` is already `nullptr` in the destructor, and only do anything if it's non-null. In the common case (no cancellation) it should always be null already.

I still want to make it right and reverts are free...

Issue #2506, opened 2026-01-13
C++ build failure, missing cstdint header in async.h


c3d:
The build failure looks like this
```
In file included from $ROOT/capnproto/c++/src/kj/async.h:1409:
$ROOT/capnproto/c++/src/kj/async-inl.h:39:1: note: ‘uintptr_t’ is defined in header ‘<cstdint>’; this is probably fixable by adding ‘#include <cstdint>’
   38 | #include <kj/list.h>
  +++ |+#include <cstdint>
 ```

This is not a regression, saw in on `v2` as well as `v1.0.2`

Issue #1949, opened 2024-02-22
Error while using coroutines


thegreathir:
I'm using v1.0.2, this is my code:

```
#include "kj/async-io.h"

kj::Promise<void> loop(kj::AsyncIoContext &io) {
  auto addr = co_await io.provider->getNetwork().parseAddress("", 12);
  auto receiver = addr->listen();
  auto conn = co_await receiver->accept();
  co_return;
}

int main() {
  auto io = kj::setupAsyncIo();
  loop(io).wait(io.waitScope);
  return 0;
}
```

This is how I compile it:
```
set(CMAKE_CXX_FLAGS "${CMAKE_C_FLAGS} -freport-bug -fcoroutines")
add_executable(
    main main.cpp
)

set_property(TARGET main PROPERTY CXX_STANDARD 20)
target_link_libraries(main kj)
```

Am I doing something obviously wrong?
Because compiler crashes:
```
main.cpp:8:1: internal compiler error: in gimplify_var_or_parm_decl, at gimplify.cc:3058
    8 | }
      | ^
0x1ad33c8 internal_error(char const*, ...)
	???:0
0x6b7b63 fancy_abort(char const*, int, char const*)
	???:0
0xb96497 gimplify_expr(tree_node**, gimple**, gimple**, bool (*)(tree_node*), int)
	???:0
0xb96546 gimplify_expr(tree_node**, gimple**, gimple**, bool (*)(tree_node*), int)
	???:0
0xb96a34 gimplify_expr(tree_node**, gimple**, gimple**, bool (*)(tree_node*), int)
	???:0
0xb965ff gimplify_expr(tree_node**, gimple**, gimple**, bool (*)(tree_node*), int)
	???:0
0xb97cbf gimplify_expr(tree_node**, gimple**, gimple**, bool (*)(tree_node*), int)
	???:0
0xb96fc5 gimplify_expr(tree_node**, gimple**, gimple**, bool (*)(tree_node*), int)
	???:0
0xb97c5f gimplify_expr(tree_node**, gimple**, gimple**, bool (*)(tree_node*), int)
	???:0
0xb96fc5 gimplify_expr(tree_node**, gimple**, gimple**, bool (*)(tree_node*), int)
	???:0
0xb96326 gimplify_expr(tree_node**, gimple**, gimple**, bool (*)(tree_node*), int)
	???:0
0xb96fc5 gimplify_expr(tree_node**, gimple**, gimple**, bool (*)(tree_node*), int)
[...]

PR #2507, opened 2026-01-13
c++/src/kj: Fix build failure on GCC 15.2.1

c3d:
With GCC 15.2.1, the `<cstdint>` header is required for `uintptr_t`:

```
In file included from $ROOT/capnproto/c++/src/kj/async.h:1409:
$ROOT/capnproto/c++/src/kj/async-inl.h:39:1: note: ‘uintptr_t’ is defined in header ‘<cstdint>’; this is probably fixable by adding ‘#include <cstdint>’
   38 | #include <kj/list.h>
  +++ |+#include <cstdint>
```

Fixes: #2506

### DIFF
--- a/c++/src/kj/async-inl.h
+++ b/c++/src/kj/async-inl.h
@@ -35,6 +35,8 @@
 #include <intrin.h>  // _ReturnAddress
 #endif
 
+#include <cstdint>
+
 #include <kj/list.h>
 
 KJ_BEGIN_HEADER


Sarcasm:
FWIW, in https://github.com/capnproto/capnproto/issues/1949#issuecomment-3664524463 I was wondering if official GCC support could come back.

It looked like to me GCC >= 14.2 ought to work.
Your PR seems to confirm GCC >= 15.2.1 it's ok, if that's the only issue you noticed in v2 and v1.

I guess CI update is still required to make the support more official / avoid regressions.

kentonv:
The error message you quoted is only half the error, it's showing the suggested fix but not the error itself.

AFAICT `uintptr_t` does not appear in `async-inl.h` so I don't quite understand what this is fixing.

FWIW we prefer to include `<stdint.h>` rather than `<cstdint>`. But I would like to understand where the actual error is before doing anything here.

fhanau:
I may have inadvertently solved this: With an older but still supported clang version, I got errors about uintptr_t missing from `src/kj/exception.h` and accordingly added the header in [#2523](https://github.com/capnproto/capnproto/pull/2523). exception.h is transitively included in async.h, so if uintptr_t was missing in async.h some other way, it should be available now.

kentonv:
Yes, it appears the real culprit was the `KJ_CALLING_ADDRESS` which is defined in exception.h and used in async-inl.h. In any case #2523 fixes this so I'll close this PR.

PR #2509, opened 2026-01-13
[coro] remove unwind detector from PromiseAwaiter

mikea:
This is #2490 with additional fix in ~PromiseAwaiterBase together with new tests.

### DIFF
--- a/c++/src/kj/async-coroutine-test.c++
+++ b/c++/src/kj/async-coroutine-test.c++
@@ -281,8 +281,9 @@ KJ_TEST("Coroutines can be canceled while suspended") {
   KJ_EXPECT(unwind == 3);
 }
 
-kj::Promise<void> deferredThrowCoroutine(kj::Promise<void> awaitMe) {
-  KJ_DEFER(kj::throwFatalException(KJ_EXCEPTION(FAILED, "thrown during unwind")));
+Promise<void> deferredThrowCoroutine(Promise<void> awaitMe, StringPtr message) {
+  KJ_DEFER(throwFatalException(
+      Exception(Exception::Type::FAILED, __FILE__, __LINE__, str(message))));
   co_await awaitMe;
   co_return;
 };
@@ -292,7 +293,7 @@ KJ_TEST("Exceptions during suspended coroutine frame-unwind do not propagate via
   WaitScope waitScope(loop);
 
   auto exception = kj::runCatchingExceptions([&]() {
-    (void)deferredThrowCoroutine(kj::NEVER_DONE);
+    (void)deferredThrowCoroutine(kj::NEVER_DONE, "thrown during unwind");
   });
 
   KJ_EXPECT(exception == kj::none);
@@ -331,7 +332,7 @@ KJ_TEST("Exceptions during suspended coroutine frame-unwind coro destructor can 
   bool caught = false;
   bool resolved = false;
   auto exception = kj::runCatchingExceptions([&]() {
-    auto coro1 = deferredThrowCoroutine(kj::NEVER_DONE);
+    auto coro1 = deferredThrowCoroutine(kj::NEVER_DONE, "thrown during unwind");
     (void)tryCatchCoAwaitCoroutine(kj::mv(coro1), &resolved, &caught);
   });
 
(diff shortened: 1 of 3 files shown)


c++/src/kj/async.c++
```diff
@@ -3300,16 +3300,22 @@ void CoroutineBase::destroy() {
 PromiseAwaiterBase::PromiseAwaiterBase(OwnPromiseNode&& node): node(kj::mv(node)) {}
 PromiseAwaiterBase::PromiseAwaiterBase(PromiseAwaiterBase&&) = default;
 PromiseAwaiterBase::~PromiseAwaiterBase() noexcept(false) {
-  // Make sure it's safe to generate an async stack trace between now and when the Coroutine is
-  // destroyed.
-  KJ_IF_SOME(coroutine, maybeCoroutine) {
+  if (node) {
```
kentonv: Nitpick: KJ style prefers `if (node.get() == nullptr)` rather than testing for truthiness.

c++/src/kj/async.c++
```diff
@@ -3300,16 +3300,22 @@ void CoroutineBase::destroy() {
...
-  }
 
-  unwindDetector.catchExceptionsIfUnwinding([this]() {
-    // No need to check for a moved-from state, node will just ignore the nullification.
-    node = nullptr;
-  });
+    try {
+      node = nullptr;
+    } catch (...) {
+      // propagate this exception as coroutine result
```
kentonv: What if the coroutine itself wrapped this await in a try/catch? I think this will skip it?

Previously we would let the exception propagate out of `~PromiseAwaiterBase()` (if not already unwinding), could we stick with that?

mikea: > Previously we would let the exception propagate out of ~PromiseAwaiterBase() (if not already unwinding), could we stick with that?

This code is not correct. Currently the code aborts on new tests II & III. 

Throwing from ~PromiseAwaiterBase is not correct, and existing test does not exercise this behavior.

mikea: It is also worth noting that uncaughtExceptions == 0 always in ~PromiseAwaiter and previously unwind detector didnt do anything in the place. 

I'll add additional test to see how it works.

mikea: here: https://github.com/capnproto/capnproto/pull/2510

you can see that new tests crash with existing approach of throwing in ~PromiseAwaiterBase

mikea: > What if the coroutine itself wrapped this await in a try/catch? I think this will skip it?

I've added new test: https://github.com/capnproto/capnproto/pull/2511

I'll try to figure out how to fix the exception in destructor and this one at the same time.

kentonv: Oh I see, since we're strictly in the cancellation path here there's actually nowhere for the exception to go.

Does calling `coroutine.unhandled_exception()` even accomplish anything? Since the coroutine is being canceled anyway.

I think it would probably be fine to ignore the exception here.

harrishancock: > This code is not correct. Currently the code aborts on new tests II & III.

I did some testing locally, and I think those test cases abort because:
- On the first exception during frame-unwind, clang calls `unhandled_exception()`, and we store the exception. There is now a zero uncaught exception count.
- `coroutine.destroy()` continues destroying the rest of the frame! This behavior is news to me -- I thought it stopped after calling `unhandled_exception()`, and we had to keep calling `coroutine.destroy()`.
- On the second exception during frame-unwind, clang does _not_ call `unhandled_exception()`, and instead aborts, seemingly because something inside its `__builtin_coro_destroy()` is `noexcept` (though `coroutine.destroy()` is not itself `noexcept`).

From one perspective, the new tests are buggy. That is:

```cpp
kj::Promise<void> deferredThrowCoroutine2(kj::Promise<void> awaitMe) {
  KJ_DEFER(kj::throwFatalException(KJ_EXCEPTION(FAILED, "thrown during unwind 2")));
  co_await awaitMe;
  co_return;
};
```

If that were a normal function, it'd be obvious that unconditionally throwing from the `KJ_DEFER` is dangerous, because the `awaitMe` promise destructor might have thrown, and we'd expect the terminate handler to be called. If it were a normal function, we'd have to wrap it in an `UnwindDetector` to be safe.

But, since it's a coroutine, we know that the `awaitMe` destructor exception will be caught and set off to the side, so when the KJ_DEFER executes, there's no uncaught exception count to tip us off that there's an in-flight exception -- even though apparently clang's coroutine destruction builtin machinery _does_ seem to know, as it refuses to call `unhandled_exception()` a second time.

The solution seems to be to rethrow from `unhandled_exception()` if it's called during destruction. It appears that if we do that, the rest of the frame _does_ still get unwound, with the exception in-flight, just like a normal function's stack. The only thing that doesn't happen is that `~CoroutineBase()` is not called, forcing us to still call `coroutine.destroy()` a second time to avoid a leak.

I put up a very hacky PR based on #2510 which makes the above change: https://github.com/capnproto/capnproto/pull/2512 It passes the new tests, and the new one in #2511.

It's possible that `unhandled_exception()`'s behavior was modified to allow throwing (or rather, define its semantics properly if you do) sometime in C++20: https://ibob.bg/blog/2024/10/06/coro-throw/ So maybe things have shifted underneath our feet.

kentonv: > coroutine.destroy() continues destroying the rest of the frame! This behavior is news to me -- I thought it stopped after calling unhandled_exception(), and we had to keep calling coroutine.destroy().

Perhaps a bug that they've since fixed?

c++/src/kj/async-coroutine-test.c++
```diff
@@ -19,6 +19,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
+#include "kj/exception.h"
```
kentonv:
```suggestion
#include <kj/exception.h>
```

c++/src/kj/async-coroutine-test.c++
```diff
@@ -353,9 +354,34 @@ KJ_TEST("Exceptions during suspended coroutine frame-unwind node destructor can
   });
 
   KJ_EXPECT(exception == kj::none);
+  // exceptions are ignored during cancellation
   KJ_EXPECT(!resolved);
-  // TODO(mikea): this will be fixed to be consistent with above in #2509
-  KJ_EXPECT(caught);
+  KJ_EXPECT(!caught);
```
mikea, line 359: @harrishancock as expected this has fixed the test.

harrishancock: I guess "Exceptions during suspended coroutine frame-unwind node destructor can be caught" is no longer an accurate name for the test...

"Exceptions thrown during co_await cancellation cannot be caught"?

PR #2511, opened 2026-01-14
[nfc] coro try/catch of destructor exception test

mikea:


fhanau (approved):

c++/src/kj/async-coroutine-test.c++
```diff
@@ -297,6 +297,24 @@ KJ_TEST("Exceptions during suspended coroutine frame-unwind propagate via destru
   KJ_EXPECT(exception.getDescription() == "thrown during unwind");
 };
 
+kj::Promise<void> tryCatchCoAwaitCoroutine(kj::Promise<void> awaitMe) {
+  try {
+    co_await awaitMe;
+  } catch (...) { }
```
harrishancock, line 303: The test case says, "Exceptions ... can be caught". Is it referring to this `catch` block? If so, should the `catch` block set a boolean proving that it ran?

PR #2512, opened 2026-01-14
More coro destructor exceptions tests

harrishancock:


c++/src/kj/async-coroutine-test.c++
```diff
@@ -281,7 +281,28 @@ KJ_TEST("Coroutines can be canceled while suspended") {
 }
 
 kj::Promise<void> deferredThrowCoroutine(kj::Promise<void> awaitMe) {
-  KJ_DEFER(kj::throwFatalException(KJ_EXCEPTION(FAILED, "thrown during unwind")));
+  UnwindDetector detector;
+  KJ_DEFER(detector.catchExceptionsIfUnwinding([&]() {
+    kj::throwFatalException(KJ_EXCEPTION(FAILED, "thrown during unwind"));
+  }));
+  co_await awaitMe;
```
mikea, line 288: I think that you are technically correct. However this means that KJ_DEFER is not safe in coroutines and should always be accompanied by UnwindDetector.

Maybe KJ_DEFER needs to be fixed? However it would also make it slower?

harrishancock: I think KJ_DEFER is unsafe in regular functions, too, not just coroutines. For example, this regular function contains equally buggy code:

```cpp
void foo() {
  KJ_DEFER(KJ_FAIL_ASSERT("throw"));
  KJ_DEFER(KJ_FAIL_ASSERT("throw 2"));
}
```

We don't seem to have major problems with double-throwing in production, though, so I'm not sure we should fix it. I'm not confident we even can fix it perfectly, because I think storing UnwindDetector on the heap (like in coroutine frames) might be generally incorrect. (Despite my use of them in CoroutineBase, and now `deferredThrowCoroutine()`.) Probably this is not news to you. :) Consider this contrived, non-coroutine example:

```cpp
void doubleThrow() {
  Own<void> ptr;
  try {
    // On scope exit, create a KJ_DEFER stored in ptr. The KJ_DEFER will
    // throw if it thinks we're _not_ unwinding.
    KJ_DEFER({
      ptr = kj::heap(kj::defer([detector = UnwindDetector()]() {
        detector.catchExceptionsIfUnwinding([&]() {
          KJ_FAIL_ASSERT("thrown during unwind");
        });
      }));
    });
    // ptr's detector stores uncaughtCount = 1 due to this exception.
    KJ_FAIL_ASSERT("ignored");
  } catch (...) {}

  // ptr's detector incorrectly thinks we're not unwinding during unwind for this exception.
  // So it throws, calling std::terminate().
  KJ_FAIL_ASSERT("test exception");
}
```

More on topic, perhaps rethrowing from `unhandled_exception()` during destruction is the best change we can make, under the cirumstances? That would allow object destructors in coroutine frames to see an accurate uncaught exception count, so code like this would work:

```cpp
Promise<void> foo() {
  KJ_DEFER({
    if (UnwindDetector::uncaughtExceptionCount() == 0) {
      KJ_FAIL_ASSERT("throw");
    }
  });
  KJ_DEFER({
    if (UnwindDetector::uncaughtExceptionCount() == 0) {
      KJ_FAIL_ASSERT("throw 2");
    }
  });
  auto p = deferredThrowCoroutine(kj::NEVER_DONE);
  co_await deferredThrowCoroutine(kj::NEVER_DONE);
}
```

If instead we have `~PromiseAwaiter()` call `unhandled_exception()` manually during `coroutine.destroy()`, like in #2509, and keep `unhandled_exception()`'s current behavior, that special-cases the last line (the `co_await`), but the previous three lines would all still throw.

Separately, I also wonder if we just can unconditionally rethrow from `unhandled_exception()`, and catch them at the `coroutine.resume()` call site nowadays? That might simplify our code a bit.

PR #2513, opened 2026-01-15
Add self-reference in RpcConnectionState::handleMessage

joshthoward:
I believe that addresses a segfault (STOR-4876). What happens is:
1. In handleFinish pipelineToRelease, goes out of scope and triggers the PostReturnRpcPipeline destructor
2. PostReturnRpcPipeline owns a RpcCallContext, which then goes out of scope and triggers its destructor
3. RpcCallContext might contain the last reference to RpcConnectionState without the self-reference

I do not know much about RPC internals, so please let me know if there is a better way to fix this.


kentonv:
This doesn't seem like the right solution. `handleMessage()` is called by `messageLoop()`, which runs as a task in the `TaskeSet` of the `RpcConnectionState`. If we destroy the `RpcConnectionState` from inside `messageLoop()`, we are self-cancelling, which is UAF city. This change would only delay the inevitable a little bit.

Normally, though, `RpcConnectionState` should always have a strong reference held by the `RpcSystem` while the connection is live. Obviously we don't expect `handleMessage()` to be called when the connection is lost. Is the destructor chain actually destroying the `RpcSystem`, too?

PR #2514, opened 2026-01-15
Implement generalized niche value optimization for Maybe

harrishancock:
This PR implements the following new features for `Maybe<T>`.

- Generalized niche value optimization.
- A converting constructor and converting assignment operator from `U&&` for any type U which can convert to T.
- An implicit conversion operator to `Maybe<U&>`, where `U&` is returned from T's dereference operator (unary `operator*()`).

Types must explicitly opt-in to benefit from the new features. Niche value optimization must be opt-in for obvious reasons; the converting constructor/assignment operator must be because suddenly implementing them for all types would break existing code; the implicit dereferencing conversion operator could be implemented unconditionally, but it seemed prudent not to.

All of these features were already implemented by  the `Maybe<Own<T>>` partial specialization. Now that `Maybe<T>` supports them, too, this PR deletes the `Maybe<Own<T>>` specialization in favor of opting `Own<T>` into the new features.

Additionally, this PR opts `Exception` into niche value optimization, making `sizeof(Maybe<Exception>) == sizeof(Exception)`.

As part of this work, I moved `Maybe`-related tests to their own file, maybe-test.c++, and converted them to `KJ_TEST` (they were originally written with gtest macros). I also added a `Maybe<Own<T>>` test suite to memory-test.c++ to help avoid regressions during the `Maybe<Own<T>>` conversion.

Optimizing `Maybe<Exception>` had the side effect of making the recent `KJ_TRY` / `KJ_CATCH` macros zero-cost in terms of frame size when used inside a coroutine. That is, previously `KJ_TRY` would add 8 bytes to the size of coroutine frames which `co_await`ed in a `KJ_TRY` block, when compared to the equivalent `try` block. Now that overhead is gone. That said, in my testing, if a `co_await` is used in a `KJ_CATCH` block, the macros have 60 bytes of overhead more than the equivalent `try` / `catch` formulation which stores the exception off to the side. I don't yet know why the difference is so extreme, but such code is relatively rare.

This PR is best reviewed commit by commit. The first several commits are no-functional-change refactors, followed by one large commit which implements the three new features. After that are two more commits for `Maybe<Own<T>>` and `Maybe<Exception>` to take advantage of the new features. Lastly, there is a commit which implements a minor change to the niche value optimization mechanism which I am undecided about (skipping "none" value destruction), but wanted to offer for review.

Claude wrote all the code at first, under my wise tutelage, then I revised it. I assert that it is not slop.

# Traits class

To make the new features opt-in, I added a `kj::MaybeTraits<T>` traits class which authors can specialize. Niche value optimization requires implementing two helper functions in the traits class (`initNone()`, `isNone()`), and the other two features are as simple as setting a boolean to true. Traits classes are not a common design pattern in KJ code, but the number of customization points seemed to warrant such an approach. I originally tried an approach similar to `KJ_STRINGIFY()` / `KJ_HASHCODE()`, but I thought the result was harder to read.

# Niche optimization

Niche optimization is terminology borrowed from Rust -- a more C++ name might be "sentinel value optimization". The general idea is that types which have an invalid value state can use this state to represent `kj::none`, allowing `Maybe<T>` to avoid having to store a separate boolean off to the side. `Maybe<T&>` and `Maybe<Own<T>>` have long had this optimization, though hand-rolled.

To opt into this feature, authors must implement:

- `MaybeTraits<T>::initNone(T*) -> void`
- `MaybeTraits<T>::isNone(const T&) -> bool`

Any type T which implements these functions is said to be `NicheOptimizable`. A new partial specialization of `kj::_::NullableValue<T>` for `NicheOptimizable<T>` allows such types to be instantiated using the main `Maybe<T>` template. The specialized `NullableValue` template uses `!isNone(value)` in place of its `isSet` boolean, and `initNone(&value)` in place of `isSet = false; dtor(value);`, otherwise its semantics are identical to the default `NullableValue<T>` implementation. The two implementations are still convertible to each other if T and U are themselves convertible.

There are now only two `Maybe` template implementations: the default `Maybe<T>`, and the `Maybe<U&>` partial specialization, reducing maintenance burden.

# Conversions

The new converting constructor, converting assignment operator, and implicit dereferencing conversion operator templates are more straightforward.

I originally tried to simply add the features unconditionally to the default `Maybe<T>` implementation. This worked fine for the dereferencing conversion operator (`Maybe<T> -> Maybe<U&>` if `T::operator*() -> U&` exists). At least, all of our code compiled when I added it. However, it seemed unwise to make that the default behavior for types which can be dereferenced, so I made it opt-in anyway.

I think there is a stronger rationale for unconditionally adding a converting constructor and converting assignment operator from `U&&` for types U which can be converted to T. After all, `Maybe<T>` already has such a constructor and operator for `Maybe<U>`. Probably if we had thought to add this long ago, we could have done so, but doing so now breaks code downstream.

### DIFF
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -1135,9 +1135,11 @@ struct MaybeTraits {
   // Maybe's assignment operators - if a constructor throws, we need to be able to safely
   // leave the Maybe in the none state.
   //
-  // IMPORTANT: T's destructor must be able to handle destroying a none value. The Maybe
-  // implementation may destroy the none value at any time (e.g., when transitioning from
-  // none to a real value, or when the Maybe itself is destroyed).
+  // IMPORTANT: The none value must not require destruction. Maybe will NEVER call T's destructor
+  // on a none value. This allows T's destructor to assume it is always destroying a valid
+  // (non-none) value, simplifying the implementation of RAII types. For example, a guard type
+  // that stores a pointer and calls ptr->exit() in its destructor can use nullptr as its none
+  // state without checking for null in the destructor.
   //
   // Choosing a none value: If your type T has a moved-from state which is practically
   // unusable (supporting only destruction and assignment), then that state is likely a
@@ -1487,7 +1489,8 @@ public:
       noexcept(noexcept(instance<T&>().~T()))
 #endif
   {
-    dtor(value);
+    // Never destroy none values - they are treated as trivially destructible sentinels.
+    if (!isNone(value)) dtor(value);
   }
 
   inline T& operator*() & { return value; }
@@ -1502,8 +1505,9 @@ public:
   template <typename... Params>
   inline T& emplace(Params&&... params) {
     // Exception safety: if dtor or ctor throws, leave in none state.
+    // Note: we never destroy none values - they are trivially destructible sentinels.
     try {
-      dtor(value);
+      if (!isNone(value)) dtor(value);
       ctor(value, kj::fwd<Params>(params)...);
     } catch (...) {
       initNone(&value);  // noexcept - leave in none state
@@ -1598,14 +1602,17 @@ public:
     return *this;
   }
   inline NullableValue& operator=(decltype(nullptr)) {
-    // Exception safety: if dtor throws, leave in none state.
-    try {
-      dtor(value);
-    } catch (...) {
-      initNone(&value);  // noexcept
-      throw;
+    // Never destroy none values - they are trivially destructible sentinels.
+    if (!isNone(value)) {
+      // Exception safety: if dtor throws, leave in none state.
+      try {
+        dtor(value);
+      } catch (...) {
+        initNone(&value);  // noexcept
+        throw;
+      }
+      initNone(&value);
     }
-    initNone(&value);
     return *this;
   }
 
--- a/c++/src/kj/maybe-test.c++
+++ b/c++/src/kj/maybe-test.c++
@@ -150,6 +150,42 @@ struct NoneThrowingNiche {
   friend struct ::kj::MaybeTraits<NoneThrowingNiche>;
 };
 
+struct NoneDestructorCounter {
+  // A niche-optimized type that counts destructor calls and asserts its destructor is never
+  // called with the none value. This verifies that Maybe never destroys none values.
+  int value;
+
+  static int nonNoneDestroyCount;
+  static int noneDestroyCount;  // Should always remain 0
+
+  explicit NoneDestructorCounter(int v): value(v) {}
+
+  NoneDestructorCounter(NoneDestructorCounter&& other): value(other.value) {
+    other.value = -1;  // moved-from state, NOT none (which is 0)
+  }
+
+  NoneDestructorCounter& operator=(NoneDestructorCounter&& other) {
+    value = other.value;
+    other.value = -1;  // moved-from state, NOT none
+    return *this;
+  }
+
+  ~NoneDestructorCounter() {
+    if (value == 0) {
+      ++noneDestroyCount;  // This should never happen!
+    } else {
+      ++nonNoneDestroyCount;
+    }
+  }
+
+  KJ_DISALLOW_COPY(NoneDestructorCounter);
+
+  friend struct ::kj::MaybeTraits<NoneDestructorCounter>;
+};
+
+int NoneDestructorCounter::nonNoneDestroyCount = 0;
+int NoneDestructorCounter::noneDestroyCount = 0;
+
 }  // namespace
 
 // MaybeTraits specializations for test types
@@ -179,6 +215,13 @@ struct MaybeTraits<NoneThrowingNiche> {
   // Niche optimization: value == 0 is the "none" state
   static void initNone(NoneThrowingNiche* ptr) noexcept { kj::ctor(*ptr, 0); }
   static bool isNone(const NoneThrowingNiche& m) noexcept { return m.value == 0; }
+};
+
+template <>
+struct MaybeTraits<NoneDestructorCounter> {
+  // Niche optimization: value == 0 is the "none" state
+  static void initNone(NoneDestructorCounter* ptr) noexcept { kj::ctor(*ptr, 0); }
+  static bool isNone(const NoneDestructorCounter& m) noexcept { return m.value == 0; }
 };
 
 namespace {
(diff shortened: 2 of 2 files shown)


kentonv:
If we can generalize the `Maybe<Own<T>>` optimization that seems great. Let's make sure to replace it rather than just add another `Maybe` variant as it's kind of painful to maintain all these separate variants.

harrishancock:
Edit: this comment is outdated, see PR description

~~@kentonv I agree. It looks doable, but there are a couple awkward details:~~

~~**`Maybe<Own<T>>` supports implicit conversion to `Maybe<T&>`:** https://github.com/capnproto/capnproto/blob/v2/c%2B%2B/src/kj/memory.h#L560-L567~~

~~To solve this, we could add support into the main `Maybe<T>` implementation for "deref mapping". That is, if a `T::operator*() -> U&` exists, `Maybe<T>` could support conversion to `Maybe<U&>`. It would mean, for instance, that `Maybe<Locked<T>>` could be converted to `Maybe<T&>`. This makes me uneasy, though, because if we removed the Maybes, the same conversion would not occur implicitly -- that is, C++ does not consider the presence of a deref operator to imply convertibility.~~

~~We could perhaps implement the feature, but immediately mark it deprecated with a message steering people toward a new, explicit `mapDeref()` helper.~~

~~At any rate, it looks like if we add implicit deref mapping to the main `Maybe<T>` implementation, our downstream codebases compile just fine.~~

~~**`Maybe<Own<T>>` supports constructing from `Maybe<Own<U>>`, where `U` is convertible to `T`:** https://github.com/capnproto/capnproto/blob/v2/c%2B%2B/src/kj/memory.h#L543-L546~~

~~Similar to the above, we could add converting constructors to `Maybe<T>` from `Maybe<U>` where `U` is convertible to `T`. It tends to introduce ambiguity, though -- e.g. it's easy for call sites of this set of overloads to become ambiguous: https://github.com/cloudflare/workerd/blob/main/src/workerd/jsg/url.h#L63-L66~~

~~I also ran into trouble where the constructor for `Maybe<T>` would try to aggregate initialize `T` from `Maybe<T>` when `T` itself contained an initial Maybe member, [like in this line of code](https://github.com/cloudflare/workerd/blob/02cf171e25bd4632dd33a9ba088c9ecbec6886a4/src/workerd/api/r2-bucket.h#L307). Of course it would fail to compile. This was solvable by only considering converting from types which are not themselves derived from Maybe. It all seemed quite brittle, though.~~

~~If we choose not to support converting constructors in the main `Maybe<T>` implementation, then that will necessitate some adjustments downstream to manually cast `Own<T>` to `Own<U>` (e.g., `Own<ReadableDirectory>` to `Own<const ReadableDirectory>`, or `Own<AsyncTee::Branch>` to `Own<AsyncInputStream>`). It's not too bad, though: 5 locations in capnproto, 1 in workerd, and 1 in our internal codebase.~~

c++/src/kj/maybe-test.c++
```diff
@@ -0,0 +1,557 @@
...
+// THE SOFTWARE.
+
+#include "common.h"
+#include "test.h"
+#include <kj/compat/gtest.h>
+
+namespace kj {
+namespace {
+
+struct ImplicitToInt {
```
harrishancock, line 29: Claude did this code move, but I verified manually that it was, indeed, a verbatim code move.

c++/src/kj/common.h
```diff
@@ -1179,11 +1179,10 @@ class NullableValue {
     }
   }
   template <typename U>
-  inline NullableValue(const NullableValue<U&>& other)
-      : isSet(other.isSet) {
-    if (isSet) {
-      ctor(value, *other.ptr);
-    }
+    requires (!canConvert<U&, kj::Maybe<T>&>()) &&  // Don't use if copy ctor can handle it
```
harrishancock: This should have gone into a later commit.

c++/src/kj/memory-test.c++
```diff
@@ -803,6 +803,57 @@ KJ_TEST("Maybe<Own<T>> implicit conversion to Maybe<T&>") {
   }
 }
 
+// Types for testing Maybe<Own<Derived>> -> Maybe<Own<Base>> converting constructor
```
harrishancock, line 806: This should have gone in an earlier commit.

c++/src/kj/memory.h
```diff
@@ -499,152 +498,32 @@ class Own {
...
-template <typename T, typename D>
-const Own<T, D>* readMaybe(const Maybe<Own<T, D>>& maybe) {
-  return maybe.ptr ? &maybe.ptr : nullptr;
-}
-
-}  // namespace _ (private)
-
+// MaybeTraits specialization for Own<T, D>.
+// This enables:
+// 1. Niche optimization: Maybe<Own<T>> uses ptr == nullptr as "none", reducing size from 16 to 8.
```
harrishancock: "24 to 16"

c++/src/kj/maybe-test.c++
```diff
@@ -47,6 +47,153 @@ struct CopyOrMove {
...
+// =======================================================================================
+// Test types for converting constructor tests
+
+struct Base {
+  int value;
+  explicit Base(int v): value(v) {}
+  virtual ~Base() = default;
+};
+
+struct Derived: Base {
```
harrishancock, line 61: Unused structs

c++/src/kj/maybe-test.c++
```diff
@@ -47,6 +47,153 @@ struct CopyOrMove {
...
+    new (ptr, _::PlacementNew()) MoveOnlyNiche(0);
+  }
+  static constexpr bool isNone(const MoveOnlyNiche& m) { return m.value == 0; }
+};
+
+template <>
+struct MaybeTraits<NonMoveableNiche> {
+  // Niche optimization: value == 0 is the "none" state
+  static constexpr void initNone(NonMoveableNiche* ptr) {
+    new (ptr, _::PlacementNew()) NonMoveableNiche(0);
```
harrishancock: kj::ctor, here and elsewhere

c++/src/kj/common.h
```diff
@@ -1091,6 +1092,143 @@ constexpr void dtor(T& location) {
...
+concept DerefConvertsTo = hasDereferencingConversion<T>() && requires(T& t) {
+  requires canConvert<decltype(*t), U&>();
+};
+
+}  // namespace _ (private)
+
+// Concept for types that support niche optimization in Maybe<T>.
+// Requires all three niche functions to be defined in MaybeTraits<T>.
+template <typename T>
+concept NicheOptimizable = _::HasAllNicheFunctions<T>;
```
harrishancock: Perhaps relax to HasAnyNicheFunctions, so the static assert in the NullableValue specialization fires if it doesn't have all.

c++/src/kj/common.h
```diff
@@ -1091,6 +1092,143 @@ constexpr void dtor(T& location) {
...
+concept HasNicheIsNone = requires(const T& v) { { MaybeTraits<T>::isNone(v) } -> SameAs<bool>; };
+
+template <typename T>
+concept HasAnyNicheFunction = HasNicheInitNone<T> || HasNicheIsNone<T>;
+template <typename T>
+concept HasAllNicheFunctions = HasNicheInitNone<T> && HasNicheIsNone<T>;
+
+// Helper to detect convertingConstructor
+template <typename T>
+concept HasConvertingConstructor =
```
harrishancock: Rename with `-Flag` suffix for clarity.

Or just delete and use the same kind of requires clause as `hasDereferencingConversion()` uses.

c++/src/kj/common.h
```diff
@@ -1091,6 +1092,143 @@ constexpr void dtor(T& location) {
...
+//    See the detailed discussion in struct MaybeTraits below.
+//
+// 3. REFERENCE CONVERSION: When dereferencingConversion is true, Maybe<T> can implicitly
+//    convert to Maybe<U&> when *T returns a reference convertible to U&. This is useful for
+//    smart pointer types like Own<T>.
+//
+// WARNING: kj-rs (the Rust bindings for KJ) has its own knowledge of Maybe<T> memory layout
+// for interoperability with Rust. As of this writing, kj-rs supports Maybe<T&> and Maybe<Own<T>>
+// niche-optimized types, but not others. When adding niche optimization for a new type, you may
+// need to add corresponding support in kj-rs.
```
harrishancock, line 1117: Need to remember to add a static_assert to kj-rs.

c++/src/kj/common.h
```diff
@@ -1284,6 +1445,213 @@ class NullableValue {
   friend NullableValue<U>&& readMaybe(Maybe<U>&& maybe);
 };
 
+template <typename T>
+  requires NicheOptimizable<T>
+class NullableValue<T> {
+  // Partial specialization of NullableValue for niche-optimizable types.
+  // Instead of storing a separate bool flag, this uses the type's "none" state.
+
+// Verify that if any niche function is defined, all are defined.
```
harrishancock: Bad indentation

c++/src/kj/common.h
```diff
@@ -1367,7 +1735,6 @@ static constexpr None none;
 // A "none" value solely for use in comparisons with and initializations of Maybes. `kj::none` will
 // compare equal to all empty Maybes, and will compare not-equal to all non-empty Maybes. If you
 // construct or assign to a Maybe from `kj::none`, the constructed/assigned Maybe will be empty.
-
```
harrishancock, line 1377: Inadvertent deletion, will fix

c++/src/kj/common.h
```diff
@@ -1466,6 +1833,27 @@ class Maybe {
...
+  // for Maybe<T>-derived types, because taking U&& directly is a better match than derived-to-base
+  // slicing. Why would would this matter? If a Maybe-like class is implemented using
+  // inheritance-by-implementation, and has a sloppy variadic constructor template which
+  // perfectly forwards to Maybe, then that constructor will end up choosing to forward to this
+  // converting constructor, rather than to the Maybe copy constructor. When compiling the expression
+  // MaybeDerived<T>(MaybeDerived<T>(...)), the compiler ends up trying to construct T with the
+  // Maybe<T>-derived argument, which is either a compiler error or, if T actually has a compatible
+  // constructor, highly surprising. This comment also applies to the converting assignment
+  // operator, defined elsewhere. The correct solution is to avoid implementation-by-inheritance,
+  // or at least use `using Maybe<T>::Maybe` for such Maybe-derived types.
```
harrishancock, line 1852: To boil down this comment: if we want to support the `convertingConstructor` feature for T in `jsg::Optional<T>`, we have to implement `jsg::Optional` more carefully first.

harrishancock:
Rebased on v2.

harrishancock:
Another rebase on v2, about to merge.

c++/src/kj/common.h
harrishancock: static_assert added here: https://github.com/cloudflare/workerd-cxx/blob/master/gen/src/write.rs#L1761

Now, if someone tries to niche-optimize a shared Rust / C++ struct, they'll get a build failure, forcing them to manually implement support in kj-rs.

PR #2518, opened 2026-01-16
[coro] do not propagate exception during cancellation

mikea:
As discussed, the reasonings are:

- promise is destroyed there is no place to report an error
- clients have no good recourse to cancellation errors
- promise is usually cancelled as a result of error handling path

performance-neutral according to cachegrind

### DIFF
--- a/c++/src/kj/async-coroutine-test.c++
+++ b/c++/src/kj/async-coroutine-test.c++
@@ -23,6 +23,7 @@
 #include <kj/array.h>
 #include <kj/compat/http.h>
 #include <kj/debug.h>
+#include <kj/exception.h>
 #include <kj/test.h>
 
 namespace kj {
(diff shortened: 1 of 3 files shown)


c++/src/kj/async-coroutine-test.c++
```diff
@@ -286,37 +286,41 @@ kj::Promise<void> deferredThrowCoroutine(kj::Promise<void> awaitMe) {
...
   try {
     co_await awaitMe;
+    *resolved = true;
   } catch (...) { 
     *caught = true;
   }
 };
 
-KJ_TEST("Exceptions during suspended coroutine frame-unwind destructor can be caught") {
+KJ_TEST("Exceptions during suspended coroutine frame-unwind destructor can not be caught") {
```
harrishancock: Since the name of this test case's helper function is `tryCatchCoAwaitCoroutine()`, and it is implemented with a try-catch around a `co_await`, I _think_ the test case's intent is to test what happens when `co_await` throws. It's not entirely clear if that's the intent from the name of the test case, though (this line, 309), so I may be misreading.

Assuming I'm not misreading ...

This PR doesn't change any behavior of `co_await`, though. It changes the behavior of coroutine destrutors -- our helper function just happens to be `co_await`ing a coroutine, so the test case's behavior changed due to coroutine implementation details.

If it were `co_await`ing a different promise, which does not yet implement cancellation exception suppression, then the test case would still work as it was.

Since `co_await`'s behavior necessarily depends on the implementation of the PromiseNode destructor it's awaiting, it might not be possible to test what I think the test case is trying to test, unless we were to implement an explicitly-throwy `ThrowyPromiseNodeForTest` class.

What should we do about this? Reasonable options I can think of:

- Implement a special PromiseNode which always throws from its dtor, so that we can isolate `co_await`'s frame-unwind behavior using that (arguably buggy) PromiseNode.
- Change the intent of the test to just show that `catch` blocks run during frame-unwind (e.g. use KJ_DEFER instead of `co_await`).
- Delete the test.

On a separate tangent: I wonder how the new KJ_TRY / KJ_CATCH macro behaves under frame-unwind? I made it `goto` out of the `catch` handler. I ... can't imagine that works out very well during destruction, as it might hit another `co_await` during destruction. :| I'll test...

harrishancock: > On a separate tangent: I wonder how the new KJ_TRY / KJ_CATCH macro behaves under frame-unwind? I made it goto out of the catch handler. I ... can't imagine that works out very well during destruction, as it might hit another co_await during destruction. :| I'll test...

So, this code

```cpp
kj::Promise<void> tryCatchCoAwaitCoroutine(kj::Promise<void> awaitMe) {
  try {
    co_await awaitMe;
  } catch(...) {
    KJ_DBG("before");
  }
  KJ_DBG("after");
};
```

prints out "before" and "after" if `awaitMe` throws during frame-unwind. So, KJ_TRY / KJ_CATCH's use of `goto` to escape the `catch` block doesn't introduce any special concern: it's already possible that throwing during frame-unwind can resume the coroutine, in the sense that it can start executing more of the body of the coroutine (up to the next suspension point), not just destructors and catch blocks.

mikea: Thanks, I have added `DeferredThrowPromiseNode` and tested its behavior. As expected the exception now propagates through co_await but I expect this to be normalized with #2509

PR #2519, opened 2026-01-17
Fix GCC 15.2.1 compatibility

Sarcasm:
Fixes https://github.com/capnproto/capnproto/issues/2506.
Takes over https://github.com/capnproto/capnproto/pull/2507.

Fixed some compilations errors and warnings.
Fixed a failing test by using the same workaround as MSVC.

The tests seems to pass now:

```
 ctest --output-on-failure                                                                                                                                                                        (v2)*
Test project /home/papin/dev/github.com/capnproto/capnproto/build
    Start 1: kj-tests-run
1/5 Test #1: kj-tests-run .....................   Passed    1.61 sec
    Start 2: kj-heavy-tests-run
2/5 Test #2: kj-heavy-tests-run ...............   Passed    3.46 sec
    Start 3: capnp-tests-run
3/5 Test #3: capnp-tests-run ..................   Passed   36.05 sec
    Start 4: capnp-heavy-tests-run
4/5 Test #4: capnp-heavy-tests-run ............   Passed    0.83 sec
    Start 5: capnp-evolution-tests-run
5/5 Test #5: capnp-evolution-tests-run ........   Passed    1.28 sec

100% tests passed, 0 tests failed out of 5

Total Test time (real) =  43.24 sec
```

I understand that for real GCC support a CI update is desired, but it's a more involved PR than these simple fixes.
Eventually I'd like to test with GCC 14.2 which is hypothetically the first GCC release capable of really supporting KJ coroutines (and the compiler I will depend on soon).

Edit: after some testing, unfortunately GCC 14.2 isn't really the release that fixes coroutines, it seems to be GCC 14.3 which is not easily available unlike 14.2.

### DIFF
--- a/c++/src/kj/debug-test.c++
+++ b/c++/src/kj/debug-test.c++
@@ -271,7 +271,7 @@ void myFailRequire(kj::SourceLocation location = {}) {
   KJ_FAIL_REQUIRE_AT(location, "a_message");
 }
 void myRequireNonnull(kj::Maybe<int> value, kj::SourceLocation location = {}) {
-  KJ_REQUIRE_NONNULL_AT(value, location, "a_message");
+  (void)KJ_REQUIRE_NONNULL_AT(value, location, "a_message");
 }
 void myAssert(bool cond, kj::SourceLocation location = {}) {
   KJ_ASSERT_AT(cond, location, "a_message");
@@ -280,7 +280,7 @@ void myFailAssert(kj::SourceLocation location = {}) {
   KJ_FAIL_ASSERT_AT(location, "a_message");
 }
 void myAssertNonnull(kj::Maybe<int> value, kj::SourceLocation location = {}) {
-  KJ_ASSERT_NONNULL_AT(value, location, "a_message");
+  (void)KJ_ASSERT_NONNULL_AT(value, location, "a_message");
 }
 void myLog(kj::SourceLocation location = {}) {
   KJ_LOG_AT(WARNING, location, "a_message");


c++/src/kj/async-inl.h
```diff
@@ -2275,7 +2275,7 @@ class CoroutineBase: public PromiseNode,
 
   auto initial_suspend() { return stdcoro::suspend_never(); }
   auto final_suspend() noexcept {
-#if _MSC_VER && !defined(__clang__)
+#if _MSC_VER && !defined(__clang__) || __GNUC__
```
kentonv: `__GNUC__` is defined by Clang (on Unix), so this will incorrectly match Clang.

Since Clang, MSVC, and GCC are our only supported compilers, perhaps we should just reduce this to:

```
#if !defined(__clang__)
```

c++/src/kj/exception.h
kentonv: Seems this comment now needs updating.

Does GCC produce the shadowing warnings that these were meant to suppress?

Sarcasm: I don't know.
If I compile with `-Wshadow`, with GCC and Clang (CMake build sorry), I have a bunch of warnings, so I'm not sure what was supposed to happen here.
If I remove the Clang case here, so the pragma is not used, I do not have a warning either.
If I add a test with nested KJ_TRY, I do not get a warning.
But I did not try to understand the code.

I glimpsed at the KJ_TRY PR and I'm not sure if this code is still relevant or later modifications made it obsolete.

I attempted an update to the comment but I'm not sure it's good.

c++/src/kj/async-inl.h
Sarcasm: Indeed, I had a thought about this after leaving the computer, sorry.
Fixed.

Sarcasm:
I tried building with `./super-test.sh quick` instead of CMake and I get some other errors (due to `-Werror` and different warning flags), e.g:

```
libtool: compile:  g++ -DHAVE_CONFIG_H -I. -I./src -I./src -DKJ_HEADER_WARNINGS -DCAPNP_HEADER_WARNINGS -DCAPNP_INCLUDE_DIR=\"/home/papin/dev/github.com/capnproto/capnproto/tmp-staging/include\" -O2 -DDEBUG -Wall -Wextra -Werror -Werror -Wsuggest-override -Wno-strict-aliasing -Wno-sign-compare -Wno-unused-parameter -DCAPNP_EXPENSIVE_TESTS=1 -Wimplicit-fallthrough -Wno-maybe-uninitialized -Wno-strict-overflow -Wno-misleading-indentation -std=gnu++23 -DKJ_HAS_ZLIB -DKJ_HAS_OPENSSL -DKJ_USE_FIBERS -MT src/capnp/compiler/type-id.lo -MD -MP -MF src/capnp/compiler/.deps/type-id.Tpo -c src/capnp/compiler/type-id.c++ -o src/capnp/compiler/type-id.o >/dev/null 2>&1
In file included from ./src/kj/parse/common.h:41,
                 from src/capnp/compiler/parser.h:26,
                 from src/capnp/compiler/parser.c++:26:
./src/kj/parse/../tuple.h:58:23: error: self-comparison always evaluates to false [-Werror=tautological-compare]
   58 |   static_assert(index != index, "Index out-of-range.");
      |                 ~~~~~~^~~~~~~~


libtool: link: g++  -fPIC -DPIC -shared -nostdlib /usr/lib/gcc/x86_64-pc-linux-gnu/15.2.1/../../../../lib/crti.o /usr/lib/gcc/x86_64-pc-linux-gnu/15.2.1/crtbeginS.o  src/capnp/compiler/.libs/type-id.o src/capnp/compiler/.libs/error-reporter.o src/capnp/compiler/.libs/lexer.capnp.o src/capnp/compiler/.libs/lexer.o src/capnp/compiler/.libs/grammar.capnp.o src/capnp/compiler/.libs/parser.o src/capnp/compiler/.libs/generics.o src/capnp/compiler/.libs/node-translator.o src/capnp/compiler/.libs/compiler.o src/capnp/.libs/schema-parser.o src/capnp/.libs/serialize-text.o   -Wl,-rpath -Wl,/home/papin/dev/github.com/capnproto/capnproto/c++/.libs -Wl,-rpath -Wl,/home/papin/dev/github.com/capnproto/capnproto/tmp-staging/lib -L/home/papin/dev/github.com/capnproto/capnproto/c++/.libs ./.libs/libcapnp.so /home/papin/dev/github.com/capnproto/capnproto/c++/.libs/libkj.so ./.libs/libkj.so -L/usr/lib/gcc/x86_64-pc-linux-gnu/15.2.1 -L/usr/lib/gcc/x86_64-pc-linux-gnu/15.2.1/../../../../lib -L/lib/../lib -L/usr/lib/../lib -L/usr/lib/gcc/x86_64-pc-linux-gnu/15.2.1/../../.. -L/lib -L/usr/lib -lstdc++ -lm -lgcc_s -lc -lgcc_s /usr/lib/gcc/x86_64-pc-linux-gnu/15.2.1/crtendS.o /usr/lib/gcc/x86_64-pc-linux-gnu/15.2.1/../../../../lib/crtn.o  -O2 -Wall -Wextra -Werror -Werror -Wsuggest-override -Wno-strict-aliasing -Wno-sign-compare -Wno-unused-parameter -Wimplicit-fallthrough -Wno-maybe-uninitialized -Wno-strict-overflow -Wno-misleading-indentation   -Wl,-soname -Wl,libcapnpc-2.0-dev.so -o .libs/libcapnpc-2.0-dev.so
In file included from ./src/capnp/common.h:29,
                 from ./src/capnp/raw-schema.h:24,
                 from ./src/capnp/generated-header-support.h:26,
                 from ./src/capnp/schema.capnp.h:6,
                 from src/capnp/compiler/capnpc-c++.c++:28:
In function ‘char* kj::_::fill(char*, const First&, Rest&& ...) [with First = kj::ArrayPtr<const char>; Rest = {}]’,
    inlined from ‘void kj::StringTree::fill(char*, size_t, First&&, Rest&& ...) [with First = kj::ArrayPtr<const char>; Rest = {}]’ at ./src/kj/string-tree.h:182:16,
    inlined from ‘void kj::StringTree::fill(char*, size_t, First&&, Rest&& ...) [with First = kj::CappedArray<char, 14>; Rest = {kj::ArrayPtr<const char>}]’ at ./src/kj/string-tree.h:183:7,
    inlined from ‘void kj::StringTree::fill(char*, size_t, First&&, Rest&& ...) [with First = kj::ArrayPtr<const char>; Rest = {kj::CappedArray<char, 14>, kj::ArrayPtr<const char>}]’ at ./src/kj/string-tree.h:183:7,
    inlined from ‘static kj::StringTree kj::StringTree::concat(Params&& ...) [with Params = {kj::ArrayPtr<const char>, kj::CappedArray<char, 14>, kj::ArrayPtr<const char>}]’ at ./src/kj/string-tree.h:208:14,
    inlined from ‘kj::StringTree kj::strTree(Params&& ...) [with Params = {const char (&)[45], unsigned int&, const char (&)[35]}]’ at ./src/kj/string-tree.h:214:86,
    inlined from ‘capnp::{anonymous}::CapnpcCppMain::makeFieldText(kj::StringPtr, capnp::StructSchema::Field, const TemplateContext&)::<lambda(kj::_::Mapper<kj::Array<capnp::{anonymous}::CapnpcCppMain::Slot>&>::Element)>’ at src/capnp/compiler/capnpc-c++.c++:1259:68,
    inlined from ‘kj::Array<decltype (func((*((kj::_::Mapper<T>*)this)->kj::_::Mapper<T>::array.begin())))> kj::_::Mapper<T>::operator*(Func&&) [with Func = capnp::{anonymous}::CapnpcCppMain::makeFieldText(kj::StringPtr, capnp::StructSchema::Field, const TemplateContext&)::<lambda(kj::_::Mapper<kj::Array<capnp::{anonymous}::CapnpcCppMain::Slot>&>::Element)>; T = kj::Array<capnp::{anonymous}::CapnpcCppMain::Slot>&]’ at ./src/kj/array.h:715:23:
./src/kj/string.h:474:15: error: ‘void* __builtin_memcpy(void*, const void*, long unsigned int)’ offset [0, 33] is out of the bounds [0, 0] [-Werror=array-bounds=]
  474 |     *target++ = *i++;
      |     ~~~~~~~~~~^~~~~~

g++ -DHAVE_CONFIG_H -I.  -Wno-deprecated-declarations  -I./src -I./src -DKJ_HEADER_WARNINGS -DCAPNP_HEADER_WARNINGS -DCAPNP_INCLUDE_DIR='"/home/papin/dev/github.com/capnproto/capnproto/tmp-staging/include"'  -O2 -DDEBUG -Wall -Wextra -Werror -Werror -Wsuggest-override -Wno-strict-aliasing -Wno-sign-compare -Wno-unused-parameter -DCAPNP_EXPENSIVE_TESTS=1  -Wimplicit-fallthrough -Wno-maybe-uninitialized -Wno-strict-overflow -Wno-misleading-indentation -std=gnu++23  -DKJ_HAS_ZLIB -DKJ_HAS_OPENSSL -DKJ_USE_FIBERS -MT src/kj/capnp_test-exception-test.o -MD -MP -MF src/kj/.deps/capnp_test-exception-test.Tpo -c -o src/kj/capnp_test-exception-test.o `test -f 'src/kj/exception-test.c++' || echo './'`src/kj/exception-test.c++
In static member function ‘static void kj::{anonymous}::TestCase517::run()::MyDisposer::dispose(int*)’,
    inlined from ‘void kj::Own<T, StaticDisposer>::dispose() [with T = int; StaticDisposer = kj::{anonymous}::TestCase517::run()::MyDisposer]’ at ./src/kj/memory.h:491:30,
    inlined from ‘kj::Own<T, StaticDisposer>::~Own() [with T = int; StaticDisposer = kj::{anonymous}::TestCase517::run()::MyDisposer]’ at ./src/kj/memory.h:421:35,
    inlined from ‘virtual void kj::{anonymous}::TestCase517::run()’ at src/kj/memory-test.c++:553:3:
src/kj/memory-test.c++:522:19: error: storing the address of local variable ‘i’ in ‘disposedPtr’ [-Werror=dangling-pointer=]
  522 |       disposedPtr = value;
      |       ~~~~~~~~~~~~^~~~~~~
src/kj/memory-test.c++: In member function ‘virtual void kj::{anonymous}::TestCase517::run()’:
src/kj/memory-test.c++:526:7: note: ‘i’ declared here
  526 |   int i;
      |       ^
src/kj/memory-test.c++:518:15: note: ‘disposedPtr’ declared here
  518 |   static int* disposedPtr = nullptr;
      |               ^~~~~~~~~~~
```

Ignoring these warnings the build and the tests pass.

Sarcasm:
Not sure my changes are responsible of the macOs test failures:

```
[ TEST ] kj/async-io-test.c++:670: legacy test: AsyncIo/CapabilityPipeBlockedSendStream
kj/async-io-test.c++:708: failed: expected endpoint2->readAllText().wait(io.waitScope) == "foo" [ == foo]
[ FAIL ] kj/async-io-test.c++:670: legacy test: AsyncIo/CapabilityPipeBlockedSendStream (204.236ms)
[ TEST ] kj/async-io-test.c++:711: legacy test: AsyncIo/CapabilityPipeMultiStreamMessage
[ PASS ] kj/async-io-test.c++:711: legacy test: AsyncIo/CapabilityPipeMultiStreamMessage (201μs)
stack: 100085e8f 10027c8f7 100a75a3b 100a751df 100ffc62b 100ffabf3 100ffa993 100a747bf 18bc86b97
[ TEST ] kj/async-io-test.c++:746: legacy test: AsyncIo/ScmRightsTruncatedOdd
    <deduplicated_symbol> (in capnp-test) + 323: returning here
[ PASS ] kj/async-io-test.c++:746: legacy test: AsyncIo/ScmRightsTruncatedOdd (33μs)
    kj::(anonymous namespace)::TestCase670::run() (in capnp-test) + 1155: returning here
[ TEST ] kj/async-io-test.c++:904: legacy test: AsyncIo/PipeThread
    kj::TestRunner::run() (in libkj-test-2.0-dev.dylib) + 531: returning here
[ PASS ] kj/async-io-test.c++:904: legacy test: AsyncIo/PipeThread (236μs)
    kj::MainBuilder::MainImpl::operator()(kj::StringPtr, kj::ArrayPtr<kj::StringPtr const>) (in libkj-2.0-dev.dylib) + 1811: returning here
    kj::runMainAndExit(kj::ProcessContext&, kj::Function<void (kj::StringPtr, kj::ArrayPtr<kj::StringPtr const>)>&&, int, char**) (in libkj-2.0-dev.dylib) + 407: returning here
    main (in libkj-test-2.0-dev.dylib) + 123: returning here
    start (in dyld) + 6075: returning here
Test failed due to logging an unexpected ERROR (above). Expected error logs must be captured via KJ_EXPECT_LOG() or an ExceptionCallback.
```

c++/src/kj/exception.h
harrishancock: Interestingly we have a similar usage of _Pragma here, which GCC still uses: https://github.com/capnproto/capnproto/blob/v2/c%2B%2B/src/kj/common.h#L1357-L1379

The main difference between KJ_IF_SOME's usage and KJ_TRY's usage of _Pragma is that the _END macro is on an independent "line" of the macro for KJ_IF_SOME.

```cpp
#define KJ_IF_SOME(name, exp) \
    KJ_SILENCE_DANGLING_ELSE_BEGIN \
    if (auto KJ_UNIQUE_NAME(_##name) = ::kj::_::readMaybe(exp)) \
      if (auto& name = *KJ_UNIQUE_NAME(_##name); false) {} else \
    KJ_SILENCE_DANGLING_ELSE_END

#define KJ_TRY \
    KJ_SILENCE_SHADOWING_BEGIN \
    if (::kj::Maybe<::kj::Exception> _kjTryCatchException; true) \
      try KJ_SILENCE_SHADOWING_END
```

I wonder if we changed it to

```cpp
#define KJ_TRY \
    KJ_SILENCE_SHADOWING_BEGIN \
    if (::kj::Maybe<::kj::Exception> _kjTryCatchException; true) \
      try
    KJ_SILENCE_SHADOWING_END
```

if it would work?

By the way, if GCC did emit a shadow warning for this macro, you'd see it emitted here: https://github.com/capnproto/capnproto/blob/v2/c%2B%2B/src/kj/exception-test.c%2B%2B#L397

Sarcasm: Unfortunately, I'm not able to get a shadow warning out of the macro, whether I test with GCC (15.2.1) or Clang (21.1.6).

I disabled `KJ_SILENCE_SHADOWING_BEGIN` and `KJ_SILENCE_SHADOWING_END` to test with Clang:

```diff
-#if __clang__
+#if 0
 // clang understands the GCC set of pragma directives.
 // Not applied to GCC due to Bug 78657 (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78657).
 #define KJ_SILENCE_SHADOWING_BEGIN \
    _Pragma("GCC diagnostic push") \
    _Pragma("GCC diagnostic ignored \"-Wshadow\"")
```

Do you still have the warning?
I tested with: 

```sh
$ g++ -Ic++/src -fdiagnostics-color -std=c++20 -Wall -Wextra -Wshadow -fsyntax-only c++/src/kj/exception-test.c++
$ clang++ -Ic++/src -fdiagnostics-color -std=c++20 -Wall -Wextra -Wshadow -fsyntax-only c++/src/kj/exception-test.c++
```

I only get `-Wshadow` to trigger a warning if I expand the macros like this:

```c++
KJ_TEST("KJ_TRY/KJ_CATCH nested usage") {
  bool outerCaught = false;
  bool innerCaught = false;

  if (::kj::Maybe<::kj::Exception> _kjTryCatchException; true)
  try { 
    if (::kj::Maybe<::kj::Exception> _kjTryCatchException; true)
    try { 
      KJ_FAIL_ASSERT("inner exception");
    } KJ_CATCH(innerEx) {
```

It is as if shadowing was silenced for macros.

Sarcasm: Regarding the location of the pragma with GCC, the issue is not able being on a single ligne but more about being after the `{` it seems.

Bad:
```c++
KJ_TEST("KJ_TRY/KJ_CATCH nested usage") {
  bool outerCaught = false;
  bool innerCaught = false;

#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wshadow"
  if (::kj::Maybe<::kj::Exception> _kjTryCatchException; true) try 
#pragma GCC diagnostic pop
         {
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wshadow"
    if (::kj::Maybe<::kj::Exception> _kjTryCatchException; true) try 
#pragma GCC diagnostic pop
           {
      KJ_FAIL_ASSERT("inner exception");
``` 

Good:
```c++
KJ_TEST("KJ_TRY/KJ_CATCH nested usage") {
  bool outerCaught = false;
  bool innerCaught = false;

#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wshadow"
  if (::kj::Maybe<::kj::Exception> _kjTryCatchException; true) try 
         {
#pragma GCC diagnostic pop
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wshadow"
    if (::kj::Maybe<::kj::Exception> _kjTryCatchException; true) try 
           {
#pragma GCC diagnostic pop
      KJ_FAIL_ASSERT("inner exception");
```

Sarcasm:
I updated the commits:
- dropped the addition stdint.h to exception.h, recently fixed https://github.com/capnproto/capnproto/pull/2523/changes/88a1c50aee4cf3b5abfe69b0d82b0c45f06e7d9f
- adapted to the recently updated coroutine isDone logic

c++/src/kj/exception.h
harrishancock: > Do you still have the warning?

I tested with clang 21 and ... I don't. I'm certain I remember seeing it in development, though, because I didn't actually want to write these pragmas. :) I don't know what to say!

> the issue is not able being on a single ligne but more about being after the {

Interesting! I suppose we could add a brace around the entire first `if` statement

```cpp
#define KJ_TRY \
    KJ_SILENCE_SHADOWING_BEGIN \
    if (::kj::Maybe<::kj::Exception> _kjTryCatchException; true) { \
      KJ_SILENCE_SHADOWING_END try

#define KJ_CATCH(exception) \
      ...
    } else \
      ...
```

But it doesn't seem worth spending time on, if the warning doesn't even appear with clang anymore.

Thanks for investigating this, I appreciate it.

harrishancock:
> Not sure my changes are responsible of the macOs test failures:

Almost certainly not.

c++/src/kj/async-inl.h
```diff
@@ -2334,9 +2334,9 @@ class CoroutineBase: public PromiseNode,
 
   bool hasSuspendedAtLeastOnce = false;
 
-#if _MSC_VER && !defined(__clang__)
+#if !defined(__clang__)
   bool finalSuspendCalled = false;
-  // MSVC erroneously reports the coroutine as done (that is, `coroutine.done()` returns true)
+  // MSVC and GCC erroneously report the coroutine as done (that is, `coroutine.done()` returns true)
```
harrishancock, line 2339: This makes me wonder if clang is not actually the erroneous one. Probably doesn't matter, though.

kentonv:
Can you update the GitHub Action config to run tests on Linux-GCC? Otherwise I'm sure support will just break again soon enough.

Sarcasm:
I understand that without GCC in CI there is no guarantees for compatibility.
This was just a quick check on my part to see how much work would be needed (not much).
I'm slightly limited in free time / energy at this time, so not super enthusiast spending time on adding GCC to the CI as I think it's more work than these commits.

That said, for CI some questions:
- Do you accept a CMake build ?
  I think it's also untested but seems to be working and I'm pretty sure some people package with that (found a hit on first try: https://gitlab.archlinux.org/archlinux/packaging/packages/capnproto/-/blob/76babe2cf6de9ad44f8436ed9815754e4376261e/PKGBUILD)
  If using configure via `./super-test.sh quick` there is more work to make it really work with GCC, see my comment here: https://github.com/capnproto/capnproto/pull/2519#issuecomment-3765392983
- It looks like the first working of GCC is 14.3 which is unfortunately not readily available on Ubuntu or other mainstream distribs, there seems stuck with GCC 14.2 or the next major version:
  - Is it ok to use the docker image gcc:14.3 (https://hub.docker.com/_/gcc) or do you prefer a GCC 15 that can be found on Ubuntu ?

kentonv:
OK, I'll just merge this for now and not make you do CI. Might break until someone gets around to it, though.

To answer the questions, though:
* I would really prefer to make super-test work, it covers a lot more than cmake. Sounds like it's just a matter of figuring out which warnings need to be ignored these days. There's a place in the script where we apply a bunch of such flags depending on which compiler we're using.
* Docker is fine if it works.

kentonv:
Thanks!

PR #2522, opened 2026-01-20
[coro] coroutine operator new & delete

mikea:
It was tricky to get exact calls right to match current performance:

https://gist.github.com/mikea/47856a20212643b7eda1f5e15d36c645

Curiously new[]/delete[] pair is measurably slower.

This has no functionality change, getting this in to start building on top of it.


mikea:
~looks like old compilers do not support sized delete for coros, I'll add an ifdef fix in an hour or so.~

done

PR #2525, opened 2026-01-21
[coro] optimize CoroutineBase::waiting flag out

mikea:


c++/src/kj/async.c++
```diff
@@ -3188,26 +3192,13 @@ void CoroutineBase::unhandledExceptionImpl(ExceptionOrValue& resultRef) {
...
-    //
-    // Since we know we're unwinding due to a successful completion, we also know that whatever
-    // Event we may have armed has not yet fired, because we haven't had a chance to return to
-    // the event loop.
-
-    KJ_IASSERT(!isDone());
-
-    // Since final_suspend() hasn't been called, whatever Event is waiting on us has not fired,
-    // and will see this exception.
+  } else{
```
harrishancock:
```suggestion
  } else {
```

c++/src/kj/async.c++
```diff
@@ -3188,26 +3192,13 @@ void CoroutineBase::unhandledExceptionImpl(ExceptionOrValue& resultRef) {
...
     resultRef.addException(kj::mv(exception));
+    if (!onReadyEvent.armed()) {
+      // Exception during coroutine execution.
+      onReadyEvent.arm();
+    } 
+    // Otherwise this is an exception during coroutine destructor.
```
harrishancock: I think it might be more accurate to say, "during coroutine frame-unwind between `co_return` and `final_suspend()`. I think the destructor case is handled in the `KJ_IF_SOME` block above.

PR #2528, opened 2026-01-22
HTTP: Return 405 for unrecognized methods instead of 501.

danlapid:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/405


kentonv:
The long-term proper solution here is to give up on the enum and instead represent the method as an arbitrary string.

PR #2529, opened 2026-01-22
[coro] throwy parameter test

mikea:


c++/src/kj/async-coroutine-test.c++
```diff
@@ -468,6 +468,48 @@ KJ_TEST("co_await only sees coroutine destruction exceptions if promise was not
...
+
+KJ_TEST("Exception thrown by parameter destructor rejects the promise") {
+  EventLoop loop;
+  WaitScope waitScope(loop);
+
+  auto promise = fortyTwo(ThrowInDestructor());
+  auto exception = KJ_ASSERT_NONNULL(kj::runCatchingExceptions([&]() {
+    return promise.wait(waitScope);
+  }));
+  KJ_EXPECT(exception.getDescription() == "~ThrowInDestructor()");
```
harrishancock: Nit: I think this pattern could be `KJ_EXPECT_THROW_MESSAGE("~ThrowInDestructor()", promise.wait(waitScope))`.

c++/src/kj/async-coroutine-test.c++
```diff
@@ -468,6 +468,48 @@ KJ_TEST("co_await only sees coroutine destruction exceptions if promise was not
...
+}
+
+// TODO(mikea): this behavior is wrong, we want to ignore all exceptions during cancellation.
```
harrishancock, line 499: Idea: Before calling `coroutine.destroy()`, save the value of `isWaiting()` and the state of `Coroutine::result`.

If `isWaiting()` is true, then the coroutine never produced a result. Otherwise, if `Coroutine::result::value != kj::none || Coroutine::result::exception != kj::none` the coroutine produced a result, but it was never consumed by anyone waiting on the promise. In either case, the destruction must be a cancellation, so we could suppress exceptions.

Alternatively, we could figure out how to call `coroutine.destroy()` from `Coroutine::get()`, but that would destroy `this` from within `get()`.

PR #2530, opened 2026-01-22
[nfc] fix unreachable code warning in test

mikea:


c++/src/kj/async-coroutine-alloc-test.c++
```diff
@@ -140,10 +140,10 @@ KJ_TEST("Coroutine Frame sizes") {
   // We intentionally keep the upper bound open to detect when production compiler deviates.
 #if !(defined(__clang__) && __clang_major__ >= 20 && defined(NDEBUG))
   return;
-#else
 #endif
 
-
+#pragma clang diagnostic push
```
fhanau: Why ignore the warning here instead of moving the following block under `#else`?

mikea: It makes it fully gray in my ide if I do that, which implies this is some "secondary" code. idk

mikea: I redid this not to have unreachable code. but I'm going to keep -Wunreachable anyway since it is enabled downstream and has broken update many times. @fhanau

fhanau: Sounds good to me

PR #2533, opened 2026-01-26
Support customizing GzipAsyncInputStream buffer size

a-robinson:
In local benchmarking, increasing this buffer size provides a very large speedup for what seem like pretty minimal increases in memory usage.

For now I've (A) left the default size alone, and (B) only applied the change to GzipAsyncInputStream, to avoid rocking the boat too much. But we may want to change all of them and/or change the default given the performance increase available, e.g. a nearly ~2.5x increase in speed / 60% decrease in CPU usage at 16 KiB.

### DIFF
--- a/c++/src/kj/compat/gzip.h
+++ b/c++/src/kj/compat/gzip.h
@@ -31,6 +31,10 @@ namespace kj {
 
 namespace _ {  // private
 
+// Default buffer size for gzip streams. Larger buffers reduce per-call overhead but use more
+// memory. 4 KiB is the historical default, but note that many uses may want to increase this.
+// Testing on an amd64 Linux machine gets about 2 GiB/s at 16 KiB vs 850 MiB/s at 4 KiB.
+// TODO(perf): Should we just bump this to 16 KiB by default?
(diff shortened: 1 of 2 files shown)


c++/src/kj/compat/gzip.h
```diff
@@ -31,6 +31,10 @@ namespace kj {
 
 namespace _ {  // private
 
+// Default buffer size for gzip streams. Larger buffers reduce per-call overhead but use more
+// memory. 4 KiB is the historical default, but note that many uses may want to increase this.
+// Testing on an amd64 Linux machine gets about 2 GiB/s at 16 KiB vs 850 MiB/s at 4 KiB.
+// TODO(perf): Should we just bump this to 16 KiB by default?
```
kentonv, line 37: Let's go ahead and change this to whatever you've found is a good trade-off. We probably want the change everywhere where we use gzip anyway, no reason to go around configuring each spot separately.

kentonv: We can change the default in a separate PR, I'll go ahead and merge this to address the urgent issue.

PR #2535, opened 2026-01-26
[ci] add arm build

mikea:
driveby: decreasing total number of builds to run


mikea:
> nit: `--config=ci` should also be used in the Windows-bazel-clang-cl config. nit: This is an existing issue but `--verbose_failures` is already being set for all platforms in .bazelrc, no need to have it in the ci config.

I cleaned this up a bit.

PR #2536, opened 2026-01-26
[async] allow observation of EventLoop

mikea:
Introduces `EventLoopObserver` that allows monitoring of of event port events and thus measure event queue saturation.

### DIFF
--- a/c++/src/kj/async-io.h
+++ b/c++/src/kj/async-io.h
@@ -978,7 +978,7 @@ struct AsyncIoContext {
 #endif
 };
 
-AsyncIoContext setupAsyncIo();
+AsyncIoContext setupAsyncIo(kj::Maybe<EventLoopObserver&> observer = kj::none);
 // Convenience method which sets up the current thread with everything it needs to do async I/O.
 // The returned objects contain an `EventLoop` which is wrapping an appropriate `EventPort` for
 // doing I/O on the host system, so everything is ready for the thread to start making async calls
--- a/c++/src/kj/async-test.c++
+++ b/c++/src/kj/async-test.c++
@@ -1763,6 +1763,18 @@ KJ_TEST("fiber pool limit") {
   // likelihood that the new stack would be allocated in the same location.
 }
 
+struct DebugObserver: public kj::EventLoopObserver {
+  void onWaitStart() override {
+    events.add(kj::str("waitStart"));
+  }
+
+  void onWaitEnd() override {
+    events.add(kj::str("waitEnd"));
+  }
+
+  Vector<String> events;
+};
+
 #if __GNUC__ >= 12 && !__clang__
 // The test below intentionally takes a pointer to a stack variable and stores it past the end
 // of the function. This seems to trigger a warning in newer GCCs.
@@ -1802,7 +1814,8 @@ KJ_TEST("run event loop on freelisted stacks") {
   };
 
   MockEventPort port;
-  EventLoop loop(port);
+  DebugObserver observer;
+  EventLoop loop(port, observer);
   WaitScope waitScope(loop);
   waitScope.runEventCallbacksOnStackPool(pool);
 
@@ -1839,6 +1852,11 @@ KJ_TEST("run event loop on freelisted stacks") {
       KJ_EXPECT(notOnOurStack(port.waitStack));
     });
   }
+
+  KJ_EXPECT(2 == observer.events.size());
+  KJ_EXPECT("waitStart"_kj == observer.events[0]);
+  KJ_EXPECT("waitEnd"_kj == observer.events[1]);
+  observer.events.clear();
 
   port.waitStack = nullptr;
   port.pollStack = nullptr;
@@ -1878,8 +1896,10 @@ KJ_TEST("run event loop on freelisted stacks") {
       KJ_EXPECT(onOurStack(port.pollStack));
     });
   }
+
+  KJ_EXPECT(0 == observer.events.size());
 }
-#endif
+#endif // KJ_USE_FIBERS
 
 KJ_TEST("retryOnDisconnect") {
   EventLoop loop;
@@ -2028,6 +2048,21 @@ KJ_TEST("EventLoopLocal") {
   // Destroying the event loop destoys all locals, so these are no longer shared.
   KJ_EXPECT(!rc1->isShared());
   KJ_EXPECT(!rc2->isShared());
+}
+
+KJ_TEST("EventLoopObserver") {
+  DebugObserver observer;
+  EventLoop loop(observer);
+  WaitScope waitScope(loop);
+
+  auto paf = newPromiseAndFulfiller<void>();
+  KJ_ASSERT(!paf.promise.poll(waitScope));
+  paf.fulfiller->fulfill();
+  KJ_ASSERT(paf.promise.poll(waitScope));
+  paf.promise.wait(waitScope);
+
+  // no actual call to wait() will happen
+  KJ_EXPECT(0 == observer.events.size());
 }
 
 }  // namespace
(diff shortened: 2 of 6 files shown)


c++/src/kj/async-io.h
```diff
@@ -978,7 +978,27 @@ struct AsyncIoContext {
...
+
+struct EventPortObserver {
+  // Observer interface to receive callbacks on EventPort calls.
+
+  EventPortObserver() = default;
+  EventPortObserver(EventPortObserver&&) = default;
+  virtual ~EventPortObserver() = default;
+
+  virtual void onWait() {}
+  // EventPort's `wait()` method was called.
```
kentonv: In order to compute saturation metrics we will need to know both when the wait starts and when it ends.

c++/src/kj/async-unix.h
```diff
@@ -89,7 +90,7 @@ class UnixEventPort final: public EventPort, private TimerImpl::SleepHooks {
   //   processes!
 
 public:
-  UnixEventPort();
+  UnixEventPort(kj::Maybe<EventPortObserver&> observer = kj::none);
```
kentonv: Instead of having every EventPort implementation accept and call an observer, could EventLoop be responsible for it?

c++/src/kj/async-unix.c++
kentonv, line 1: It looks like you updated the epoll and kqueue variants of UnixEventPort, but not the plain-poll() variant.

mikea:
I moved the observer into the EventLoop. ptal @kentonv

c++/src/kj/async-test.c++
```diff
@@ -1763,6 +1763,34 @@ KJ_TEST("fiber pool limit") {
...
+
+  void onWaitEnd() override {
+    events.add(kj::str("waitEnd"));
+  }
+
+  void onPollStart() override {
+    events.add(kj::str("pollStart"));
+  }
+
+  void onPollEnd() override {
```
kentonv: poll() and wake() are non-blocking, so they don't need separate start/end events.

c++/src/kj/async.c++
```diff
@@ -1892,6 +1885,15 @@ void EventLoop::wait() {
     return;
   }
 
+  KJ_IF_SOME(observer, this->observer) {
+    observer.onWaitStart();
+  }
+  auto onWaitEnd = kj::defer([this]() {
```
kentonv:
```suggestion
  KJ_DEFER({
```

c++/src/kj/async.c++
```diff
@@ -1907,6 +1909,15 @@ void EventLoop::wait() {
 }
 
 void EventLoop::poll() {
+  KJ_IF_SOME(observer, this->observer) {
+    observer.onPollStart();
+  }
+  auto onPollEnd = kj::defer([this]() {
```
kentonv:
```suggestion
  KJ_DEFER({
```

c++/src/kj/async.c++
```diff
@@ -1926,6 +1937,21 @@ void EventLoop::poll() {
   }
 }
 
+void EventLoop::wake() const {
+  KJ_IF_SOME(observer, this->observer) {
+    observer.onWakeStart();
+  }
+  auto onWakeEnd = kj::defer([this]() {
```
kentonv:
```suggestion
  KJ_DEFER({
```

c++/src/kj/async-test.c++
kentonv: (I'm not actually aware of any use case for onPoll() or onWake() at all right now, so we could also just skip them entirely until they are needed.)

c++/src/kj/async.c++
mikea: KJ_DEFER here leads to

```
src/kj/async.c++:1945:5: warning: add explicit braces to avoid dangling else [-Wdangling-else]
 1945 |     KJ_IF_SOME(observer, this->observer) {
      |     ^
bazel-out/k8-dbg/bin/src/kj/_virtual_includes/kj/kj/common.h:1369:60: note: expanded from macro 'KJ_IF_SOME'
 1369 |       if (auto& name = *KJ_UNIQUE_NAME(_##name); false) {} else \
      |                                                            ^
```

I faught with it for half an hour, no idea why it is happening here.

```
  KJ_DEFER({
    KJ_IF_SOME(observer, this->observer) {
      observer.onWakeEnd();
    }
  });
```

kentonv: That doesn't make sense, it expands to essentially the same code you have here.

You can use a pragma to disable the warning.

c++/src/kj/async.h
```diff
@@ -1183,6 +1183,31 @@ class EventPort {
   // The default implementation throws an UNIMPLEMENTED exception.
 };
 
+struct EventLoopObserver {
```
kentonv: KJ style generally always defines abstract interfaces as classes, not structs:

```suggestion
class EventLoopObserver {
public:
```

I know it's equivalent but "struct" has a very different abstract meaning.

c++/src/kj/async-test.c++
kentonv: You marked this resolved but AFAICT it is not resolved?

mikea: > You marked this resolved but AFAICT it is not resolved?

Did I? I don't think I did. These were the only resolved conversations:
I have just pushed a change 2 minutes ago that addresses it though.


image

kentonv: This thread was marked resolved. I unresolved it just before my previous comment.

Maybe a misclick.

c++/src/kj/async.c++
mikea: > That doesn't make sense, it expands to essentially the same code you have here.

Agree. I wasn't able to figure it out why.

> You can use a pragma to disable the warning.

Done. Curiously pragma _inside_ KJ_DEFER doesn't help either:

```
  KJ_SILENCE_DANGLING_ELSE_BEGIN
  // For some reason silence inside KJ_IF_SOME doesn't work.
  // Putting this silence inside KJ_DEFER doesn't help either.
  KJ_DEFER({
    KJ_IF_SOME(observer, this->observer) {
      observer.onWaitEnd();
    }
  });
  KJ_SILENCE_DANGLING_ELSE_END
```

c++/src/kj/async.c++
```diff
@@ -1926,6 +1928,12 @@ void EventLoop::poll() {
   }
 }
 
+void EventLoop::wake() const {
```
kentonv, line 1936: I guess this refactoring is technically no longer needed but OTOH it's still an improvement so whatever.

c++/src/kj/async-test.c++
mikea: np, I'm pretty sure that wasn't my intent either, since I didn't see your comment until I came back from lunch.

This time it is resolved: I've removed onWake & onPoll.

c++/src/kj/async.c++
kentonv: Huh... I wonder if it's actually complaining about the KJ_IF_SOME from _before_ the KJ_DEFER?

kentonv (approved):

c++/src/kj/async.c++
harrishancock: I'm pretty sure this always happens when KJ_IF_SOME is used in a macro argument. Claude says it's because the _Pragma is evaluated immediately when the compiler sees it (during "argument prescan", the phase where the KJ_DEFER's arguments have their macros expanded). I guess the compiler doesn't have any sense for whether the expanded argument has any semantic meaning in C++ at that point.

I think we could make something like this work with the KJ_MAP `operator*()` trick:

```cpp
KJ_DEFER2 {
  KJ_IF_SOME(...) { ... }
};

struct DeferHelper {
  template <typename Func>
  Deferred<Func> operator*(Func&& func) {
    return Deferred<Func>(kj::fwd<Func>(func));
  }
};
#define KJ_DEFER2 auto KJ_UNIQUE_NAME(_kjDefer) = ::kj::_::DeferHelper{} * [&]()
```

We could re-use the same name as `KJ_DEFER` with more tricks, but I think we'd need to preserve `()`, like `KJ_DEFER() { ... };`.

PR #2541, opened 2026-01-28
Reapply "[coro] optimize CoroutineBase::waiting flag out (#2525)" (#2534)

mikea:
Original change + clang-tidy fix in a separate commit (not sure why it gets triggered here). No functionality differences.


mikea:
> * non-blocking: Fix trailing whitespace in async-inl.h (or don't, we currently have a bunch of that since there's no formatter being used)
> * non-blocking: It looks like destroy() is going to be frequently used as NullableValue is templated – it might benefit from KJ_ALWAYS_INLINE to ensure that inlining happens in debug builds. That being said I don't like how KJ_ALWAYS_INLINE needs to wrap the entire function and there's likely worse offenders here, could also be done another time.

done

PR #2544, opened 2026-01-30
define virtual destructor for TaskSet::ErrorHandler

mikea:
needs to be noexcept(false) since some implementations (HttpServer) are noexcept(false)


kentonv:
What is the purpose of this PR?

The PR description explains why the destructor needs to be `noexcept(false)` but does not describe why a virtual destructor needs to be declared in the first place.

In general in KJ style we never declare virtual destructors because the disposer system makes them unnecessary.

PR #2546, opened 2026-01-31
[async] optimize event list

mikea:
This PR replaces EventLoop::head, tail, etc... event pointers with sentinel nodes that always exist in the list.

As a result of this refactoring:

- all arm* operations become branchless
- disarm operation becomes branchless
- disarm call does not require access to EventLoop&. This is the only operation where current event loop can't be used
- drops EventLoop& loop completely from the Event (will help delete-from-fire PR)

The `turn()` function becomes more complicated since it needs to move insertion point to the head of the list, but I believe this trade-off to be worth it: the more we move toward depth-first arming, the overhead will become less. And since we save instructions per event (both arming and disariming), this will become more and more benefitial.

Current benchmark measurements (https://gist.github.com/mikea/2a1cf361abe17e0a552554225e50509c) show consistent improvements across all benchmarks, cherry-picking few important ones:

- cachegrind.bm_Coro_Fib10.diff -9.4% (many co-awaits in single coro)
- http benchmarks vary from -0.7% to -1.6%
- cachegrind.bm_Promise_Fib10.diff -3.4% (long promise.then chain).

To simplify the review I have separated moving Event to async.h in its own commit. Actual changes are in the second commit only.

### DIFF
--- a/c++/src/kj/async.c++
+++ b/c++/src/kj/async.c++
@@ -902,7 +902,7 @@ namespace _ {  // (private)
 XThreadEvent::XThreadEvent(
     ExceptionOrValue& result, const Executor& targetExecutor, EventLoop& loop,
     void* funcTracePtr, SourceLocation location)
-    : Event(loop, location), result(result), funcTracePtr(funcTracePtr),
+    : Event(location), result(result), funcTracePtr(funcTracePtr),
       targetExecutor(targetExecutor.addRef()) {}
 
 void XThreadEvent::tracePromise(TraceBuilder& builder, bool stopAtNextEvent) {
@@ -1751,20 +1751,39 @@ void EventPort::wake() const {
 }
 
 EventLoop::EventLoop(kj::Maybe<EventLoopObserver&> observer)
-    : observer(observer), daemons(kj::heap<TaskSet>(_::LoggingErrorHandler::instance)) {}
+    : observer(observer), daemons(kj::heap<TaskSet>(_::LoggingErrorHandler::instance)) {
+  auto link = [](_::Event& prev, _::Event& next) {
+    prev.next = &next;
+    next.prev = &prev.next;
+  };
+
+  // head -> depthFirstInsertPoint -> breadthFirstInsertPoint -> tail
+  link(headSentinel, depthFirstInsertPoint);
+  link(depthFirstInsertPoint, breadthFirstInsertPoint);
+  link(breadthFirstInsertPoint, tailSentinel);
+
+  // wouldSleepHead -> wouldSleepTail
+  link(wouldSleepHead, wouldSleepTail);
+}
 
 EventLoop::EventLoop(EventPort& port, kj::Maybe<EventLoopObserver&> observer)
-    : port(port),
-      observer(observer),
-      daemons(kj::heap<TaskSet>(_::LoggingErrorHandler::instance)) {}
+    : EventLoop(observer) {
+  this->port = port;
+}
 
 EventLoop::~EventLoop() noexcept(false) {
-  // Destroy all "daemon" tasks, noting that their destructors might register more daemon tasks.
-  while (!daemons->isEmpty()) {
-    auto oldDaemons = kj::mv(daemons);
-    daemons = kj::heap<TaskSet>(_::LoggingErrorHandler::instance);
+  {
+    // Mark event loop as current to allow tasks to arm/disarm events in the queue.
+    enterScope();
+    KJ_DEFER(leaveScope());
+
+    // Destroy all "daemon" tasks, noting that their destructors might register more daemon tasks.
+    while (!daemons->isEmpty()) {
+      auto oldDaemons = kj::mv(daemons);
+      daemons = kj::heap<TaskSet>(_::LoggingErrorHandler::instance);
+    }
+    daemons = nullptr;
   }
-  daemons = nullptr;
 
   KJ_IF_SOME(e, executor) {
     // Cancel all outstanding cross-thread events.
@@ -1773,11 +1792,15 @@ EventLoop::~EventLoop() noexcept(false) {
 
   // The application _should_ destroy everything using the EventLoop before destroying the
   // EventLoop itself, so if there are events on the loop, this indicates a memory leak.
-  KJ_REQUIRE(head == nullptr, "EventLoop destroyed with events still in the queue.  Memory leak?",
-             head->traceEvent()) {
+  _::Event* firstEvent = head();
+
+  KJ_REQUIRE(firstEvent == &tailSentinel,
+             "EventLoop destroyed with events still in the queue.  Memory leak?",
+             firstEvent->traceEvent()) {
+
     // Unlink all the events and hope that no one ever fires them...
-    _::Event* event = head;
-    while (event != nullptr) {
+    for (_::Event* event = firstEvent; event != &tailSentinel;) {
+      // no need to handle sentinels separately - unlinking them doesn't hurt.
       _::Event* next = event->next;
       event->next = nullptr;
       event->prev = nullptr;
@@ -1786,11 +1809,7 @@ EventLoop::~EventLoop() noexcept(false) {
     break;
   }
 
-  KJ_REQUIRE(threadLocalEventLoop != this,
-             "EventLoop destroyed while still current for the thread.") {
-    threadLocalEventLoop = nullptr;
-    break;
-  }
+  threadLocalEventLoop = nullptr;
 }
 
 void EventLoop::run(uint maxTurnCount) {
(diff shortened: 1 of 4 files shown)


c++/src/kj/async-win32.c++
```diff
@@ -98,6 +98,10 @@ public:
   void done(IoResult result) {
     KJ_ASSERT(handle != INVALID_HANDLE_VALUE);
     handle = INVALID_HANDLE_VALUE;
+    if (!EventLoop::currentThreadHasEventLoop()) {
+      // the loop is already destroyed
```
kentonv: Hmm, the fact that the current thread doesn't have an EventLoop does *not* necessarily mean the EventLoop has been destroyed. It only means the WaitScope was destroyed. Consider green threads, where EventLoops are allowed to switch threads.

So at first glance this seems like it could be wrong.

Is this function actually ever being called without a current event loop? I suppose it must be for you to have changed this.

Looks like it's only called from waitIocp(), which in turn is called from Win32IocpEventPort::{wait,poll}() (in which the EventLoop must be current), as well as by `~IoPromiseAdapter()`, at line 88 above. I guess this last call site must be the culprit, although it seems problematic for that to be called without a current event loop, too. I think this line may be covering for some other bug somewhere -- we shouldn't be destroying IoPromiseAdapters while the EventLoop is not current.

c++/src/kj/async.c++
```diff
@@ -1807,44 +1819,45 @@ void EventLoop::run(uint maxTurnCount) {
...
+
+  auto resetDepthFirstInsertPoint = [this]() {
+    // move depthFirstInsertPoint to the head of the list
+    if (headSentinel.next == &depthFirstInsertPoint) return;
+    // Unlink depthFirstInsertPoint from current position
+    depthFirstInsertPoint.linkPrevToNext();
+    // Relink after headSentinel
+    _::Event* oldNext = headSentinel.next;
+    headSentinel.linkNext(depthFirstInsertPoint);
+    depthFirstInsertPoint.linkNext(*oldNext);
```
kentonv: I think this could be expressed as:

```
depthFirstInsertPoint.insertBefore(*headSentinel.next);
```

Or we could add an `insertAfter()`:

```
depthFirstInsertPoint.insertAfter(headSentinel);
```

Either way feels a little clearer than using `linkNext()`.

c++/src/kj/async.c++
```diff
@@ -2199,153 +2217,75 @@ Event::~Event() noexcept {  // intentionally noexcept
...
-    if (next != nullptr) {
-      next->prev = &next;
-    }
-
-    if (loop.wouldSleepTail == prev) {
-      loop.wouldSleepTail = &next;
-    }
+    Event* oldNext = loop.wouldSleepHead.next;
+    loop.wouldSleepHead.linkNext(*this);
+    linkNext(*oldNext);
```
kentonv: Could this be:

```
  insertBefore(*loop.wouldSleepHead.next);
```

or:

```
  insertAfter(loop.wouldSleepHead);
```

c++/src/kj/async.h
```diff
@@ -1258,9 +1257,30 @@ class Event: private AsyncObject {
 
 private:
   friend class kj::EventLoop;
-  EventLoop& loop;
+  kj::EventLoop& assertEventLoop();
+
+  inline void linkNext(Event& other) {
```
kentonv: This `linkNext` strikes me as a little fragile in that it only links the event in one direction, leaving the other direction uninitialized. It looks like the only real place this is needed is for initializing the sentinels -- and the head and tail sentinels are in fact, intentionally, left half-initialized. Perhaps a comment to that effect would help:

```
// Only used to initialize the sentinels.
```

c++/src/kj/async.h
```diff
@@ -1258,9 +1257,30 @@ class Event: private AsyncObject {
...
+    next = &other;
+    other.prev = &this->next;
+  }
+
+  inline void linkPrevToNext() {
```
kentonv: Maybe this should be called `unlink()`?

Should it be nulling out `prev` and `next`? Feels a little dangerous to leave the pointers dangling.

c++/src/kj/async.c++
```diff
@@ -2175,10 +2192,11 @@ void detach(kj::Promise<void>&& promise) {
 }
 
 Event::Event(SourceLocation location)
-    : loop(currentEventLoop()), next(nullptr), prev(nullptr), location(location) {}
+    : next(nullptr), prev(nullptr), location(location) {}
 
-Event::Event(kj::EventLoop& loop, SourceLocation location)
-    : loop(loop), next(nullptr), prev(nullptr), location(location) {}
+kj::EventLoop& Event::assertEventLoop() {
```
kentonv: Arguably should be called `requireEventLoop()`

c++/src/kj/async.c++
```diff
@@ -3154,7 +3094,10 @@ void WeakFulfillerBase::disposeImpl(void* pointer) const {
     // Already detached.
     delete this;
   } else {
-    if (inner->isWaiting()) {
+    if (inner->isWaiting() && threadLocalEventLoop != nullptr) {
+      // Do not reject if there is no event loop running anymore - there is no way to handle it
```
kentonv: Do you actually see this being run with no event loop? I don't think it's necessarily safe to skip the rejection -- the EventLoop may still be alive, just descheduled from the thread. It seems like an error if this is ever reached.

This seems similar to my comment in async-win32.c++. It's probably just not safe to be destroying these objects when the EventLoop is not active. I don't think we can just skip the rejection.

c++/src/kj/async.h
```diff
@@ -1346,7 +1366,35 @@ class EventLoop {
   // Same as WaitScope::cancelAllDetached(). Sometimes it's easier to call on the EventLoop. (A
   // WaitScope still must exist, i.e., this EventLoop must be current.)
 
+  static bool currentThreadHasEventLoop();
```
kentonv: I guess this is only used by async-win32.c++ but I questioned its use there -- if it goes away we can delete this.

c++/src/kj/async.h
mikea: > Should it be nulling out prev and next? Feels a little dangerous to leave the pointers dangling.

done. I have inlined it in turn() though: my measurements show that compiler doesn't optimize the dead `next = nullptr` store out.

c++/src/kj/async.h
mikea: I have removed the method and replaced it with a little lambda in EventLoop constructor.

c++/src/kj/async.c++
mikea: yes, byte-stream-test fails like this: https://gist.github.com/mikea/4f4d37fd1a8a8d6eb62caa6aa31e6bac

I'll see if I can figure out what's wrong with destructor order.

c++/src/kj/async-win32.c++
mikea: >  I think this line may be covering for some other bug somewhere -- we shouldn't be destroying IoPromiseAdapters while the EventLoop is not current.

I'll try to reproduce it. Frustratingly it was very flaky - only 1 in 100 local runs with wine would crash (but reliably did crash on CI). But hopefully it is related to wrong destruction order like in unix case.

c++/src/kj/async.c++
mikea: updated the gist with full stacktrace from debugger, but the summary is: we're trying to arm an event during EventLoop destructor.

mikea: I've identified the issue: fulfiller/promise field order was wrong in ByteStreamFactory::PathProber.

c++/src/kj/async-win32.c++
mikea: As far as I can tell the chain of destruction that arms the event is the following:

- EventLoop is destroyed
- It destroys a TaskSet
- The task set through a very long chain of promises destroys the ~IoPromiseAdapter
- ~IoPromiseAdapter tries to arm the event.

Do you think this destruction sequence is wrong, or do we need additional care destroying EventLoop?
I can imagine setting EventLoop as current while destroying events and keep destroying them untile they are empty? This is the current behavior with explicit EventLoop& field I think?

@kentonv

kentonv: Yeah I think it makes sense that the EventLoop destructor should set itself as current while destroying any unprocessed events.

c++/src/kj/async.c++
kentonv: Ah, that's a good fix, it was wasting cycles rejecting the promise before!

But of course the code here shouldn't really be predicated on such destruction order in higher-level code.

I suspect setting the EventLoop current during its destructor might solve this case, too?

c++/src/kj/async.h
mikea: this is gone.

c++/src/kj/async.c++
mikea: move to https://github.com/capnproto/capnproto/pull/2574

Yes, destroying task set with EventLoop marked as current solved the problem.

c++/src/kj/async-win32.c++
mikea: done. This is needed only for daemon destroy.

mikea:
all issues addressed, ptal.

PR #2550, opened 2026-02-03
Add KJ_DEFER() block syntax to avoid -Wdangling-else warnings

harrishancock:
Possible fix for the dangling else issue uncovered in https://github.com/capnproto/capnproto/pull/2536#discussion_r2755945147.

I asked Claude to cook up a backwards-compatible block-style KJ_DEFER implementation. I think it might be worthwhile.

--- 

KJ_IF_SOME uses a deliberate dangling-else pattern with _Pragma to suppress warnings. However, _Pragma doesn't work correctly when KJ_IF_SOME is nested inside another macro's arguments (like KJ_DEFER), because the pragma is processed during argument prescan before the outer macro completes expansion.

This adds a new block syntax: KJ_DEFER() { code; };

The empty parentheses signal the new syntax, which places the lambda body outside the macro argument, allowing _Pragma to work correctly. The traditional KJ_DEFER(code) syntax remains supported for backward compatibility.

### DIFF
--- a/c++/src/kj/common-test.c++
+++ b/c++/src/kj/common-test.c++
@@ -1107,6 +1107,31 @@ KJ_TEST("kj::defer()") {
 
 }
 
+KJ_TEST("KJ_DEFER() block syntax") {
+  // Test the KJ_DEFER() { ... }; block syntax, which allows statement macros like KJ_IF_SOME
+  // inside the deferred code without triggering -Wdangling-else warnings.
+  uint i = 0;
+
+  {
+    KJ_DEFER() { ++i; };
+    KJ_EXPECT(i == 0);
+  }
+  KJ_EXPECT(i == 1);
+
+  // Test that KJ_IF_SOME works inside KJ_DEFER() without warnings.
+  Maybe<int> maybeVal = 42;
+  int captured = 0;
+  {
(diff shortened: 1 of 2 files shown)


jclee:
Oh nice... This probably also has the benefit of producing better code locations for stack traces and uses of SourceLocation, since the compiler will otherwise [use the code location at the beginning or end of the macro expansion](https://github.com/capnproto/capnproto/pull/2411#issuecomment-3325783420) for all code within the expansion.

jclee:
...A bit of a pity that the pattern requires the semicolon after the block, though; I could imagine it would be easy to accidentally omit and get confusing errors.  Not sure if the errors would typically point to a particular location, like in DeferHelper or Deferred, but if they did, it might be helpful to have a comment there to hint at the problem.

kentonv:
Eh, not sure if it's really worth it. There's other macros with internal blocks, like KJ_UNWRAP_OR... do we have to fix them too?

What if we just blanket disabled those warnings in KJ_DEFER itself? Or maybe just disable the warning globally? I have never in my life actually benefited from a dangling-else warning but I have fought with so many spurious ones, maybe this warning just sucks generally.

PR #2556, opened 2026-02-06
Kj/http: avoid optimized WebSocket pumping through adapters

ghostwriternr:
## Background

I ran into this while debugging a local-dev issue in the [Sandbox SDK](https://github.com/cloudflare/sandbox-sdk). For a new feature that exposes PTY support to connect xterm-like terminals to a sandbox, the SDK proxies a WebSocket from a container port (via workerd’s `tcpPort.fetch()` upgrade path) and then returns that WebSocket to an external client as `Response.webSocket`.

When I test this, messages proxy correctly in both directions, but closing the WebSocket doesn’t reliably finish. The external client never gets a `'close'` event, and my e2e tests hang waiting for the connection to shut down. Upon inspection, on the container side it looks like the close handshake never completes unless the connection is forcibly terminated. I wasn’t able to reproduce this behaviour in production; it seems specific to the local-dev stack.

**Note:**  I’m coming at this from the Sandbox SDK side rather than deep KJ internals, so I’d appreciate a sanity check on whether this matches KJ’s intended semantics. In the likelihood that this approach is incorrect, I still hope this debugging trail is still useful to reproduce the issue with potentially a right fix made some other way.

## Investigation

I didn’t initially know whether this was a bug in the SDK, in workerd’s WebSocket coupling, or in the container networking path, so I tried to rule out the obvious layers first:

1. I verified that basic WebSocket cases under wrangler dev close correctly (without involving container proxying).
2. I verified that direct connections to the container close correctly; the issue only appears when a `tcpPort.fetch()`-originated WebSocket is returned outward and coupled to an external client.
3. I added a worker-internal probe: accepting/driving the `tcpPort.fetch()` WebSocket inside the worker observes a clean close, but returning it outward and letting the runtime couple it can hang.

After that, it didn’t look like a TypeScript-layer bug so much as something about how the runtime couples/pumps this particular kind of WebSocket.

I reduced this to a deterministic repro in workerd’s container WebSocket proxying path (real listening socket + external WebSocket client). When I disable KJ’s optimised pumping and force the default pump loop, the repro passes.

That suggested the optimised pump path isn’t safe to use through wrapper WebSockets in this configuration, because it can affect whether close propagation actually completes.

## Fix

Avoid the optimised pumping path when wrapper WebSockets are involved, so `pumpTo()` uses the default pump loop for those wrappers.

This is the smallest change I found that makes the deterministic repro pass without disabling the optimisation globally, and it keeps the optimised path available for raw WebSocket implementations where it’s safe.

## Tests

I added a small regression test (`"WebSocket pumpTo completes on close through WebSocketPipe"`) and ran:

`bazel test //src/kj/compat:http-test`

## AI usage disclosure
I used GPT-5.2 and Opus 4.5 while working on this task, for creating reproductions of the issue at every step and finally narrowing down on the fix. But I've verified every step myself to the best of my understanding, including the final changes that have been proposed.

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -4596,8 +4596,8 @@ public:
     return out->whenAborted();
   }
   kj::Maybe<kj::Promise<void>> tryPumpFrom(WebSocket& other) override {
-    // WebSocketPipeEnd provides message-level semantics (send/receive/close). Optimized pumping can
-    // bypass close-frame parsing, so always use the default frame-aware pump.
+    // This wrapper depends on message-level close semantics, so it must not participate in
+    // optimized pumping via tryPumpFrom().
     return kj::none;
   }
 
@@ -7098,8 +7098,8 @@ private:
       co_await afterReceiveClosed();
     }
     kj::Maybe<kj::Promise<void>> tryPumpFrom(WebSocket& other) override {
-      // Default pumping ensures close frames are forwarded through close()/receive(), which keeps
-      // DelayedCloseWebSocket's close tracking consistent.
+      // This wrapper tracks close state via close()/receive(), so it must not participate in
+      // tryPumpFrom().
       return kj::none;
     }
 


ghostwriternr:
The rationale for disabling optimisations being a no-go makes sense to me. And agreed, the difference you've described between the optimised pumps vs otherwise is what I'm seeing too (and well in my case, understood 0.01% of all this for the very first time here, which has been fun).

Closing this PR in favour of the workerd regression test PR, and will continue tracking the issue there.

PR #2557, opened 2026-02-06
Add .clang-format configuration

mikea:
I do not suggest we reformat anything, but not having  .clang-format is really annoying when writing new code.

This is workerd's clang-format without workerd macros definitions.

### DIFF
--- a/c++/.clang-format
+++ b/c++/.clang-format
@@ -1,0 +1,78 @@
+Language: Cpp
+Standard: c++20
+ColumnLimit: 100
+
+SortIncludes: false
+
+AllowShortBlocksOnASingleLine: Always
+AllowShortFunctionsOnASingleLine: Inline
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false
+
+IndentWidth: 2
+IndentCaseBlocks: false
+IndentCaseLabels: true
+PointerAlignment: Left
+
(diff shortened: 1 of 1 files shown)


kentonv:
This isn't going to cause people's editors to automatically reformat files on save?

mikea:
> This isn't going to cause people's editors to automatically reformat files on save?

It shouldn't. Every editor I have ever used has a separate "format on save" option. I assume everyone who works on this repo has it disabled anyway since there is no formatter.

kentonv:
Hmm, I always use the same IDE instance for runtime and capnp changes. I don't have "format on save" enabled currently but I probably should (for runtime code).

c++/.clang-format
```diff
@@ -0,0 +1,92 @@
+Language: Cpp
```
harrishancock, line 1: I'd suggest adding a top-level comment explaining its purpose:

> This .clang-format configuration approximates, but not does not perfectly match, the C++ style used in the capnproto repository. This being the case, we do not generally run clang-format on the codebase. Nevertheless, contributors may find this configuration useful when contributing _new_ code, which their fingers, or their editor, did not organically format to perfection.

kentonv:
OK, you can merge this but the first time I see a PR where someone has applied a formatter to code they didn't otherwise modify I'm gonna delete it...

PR #2558, opened 2026-02-06
[coro] simple stack allocator

mikea:
Mainly to benchmark coros without tcmalloc overhead (and maybe to build upon)

```
-----------------------------------------------------------------------------
Benchmark                                   Time             CPU   Iterations
-----------------------------------------------------------------------------
bm_Promise_ReadyNow                      31.5 ns         31.5 ns       100000
bm_Promise_Immediate                     30.6 ns         30.6 ns       100000
bm_Coro_Immediate                        43.4 ns         43.4 ns       100000
bm_Promise_ImmediatePromise_Then         35.6 ns         35.6 ns       100000
bm_Coro_CoAwait_ImmediatePromise         66.4 ns         66.4 ns       100000
bm_Coro_CoAwait_ImmediateCoroutine       82.8 ns         82.8 ns       100000
bm_Promise_Pow2_20                        563 ns          562 ns       100000
bm_Coro_Pow2_20                          1079 ns         1078 ns       100000
bm_Stack_Pow2_20                          777 ns          777 ns       100000
bm_Promise_Shift_20                       595 ns          595 ns       100000
bm_Coro_Shift_20                         1055 ns         1054 ns       100000
bm_Stack_Shift_20                         792 ns          792 ns       100000
bm_Promise_Fib10                          984 ns          984 ns       100000
bm_Coro_Fib10                             787 ns          787 ns       100000
bm_Stack_Fib10                            686 ns          686 ns       100000
```

### DIFF
--- a/c++/src/kj/exception-test.c++
+++ b/c++/src/kj/exception-test.c++
@@ -536,6 +536,66 @@ KJ_TEST("KJ_TRY/KJ_CATCH does not catch CleanShutdownException") {
   KJ_EXPECT(outerCatchCalled);
 }
 
+KJ_TEST("addTrace grows dynamically") {
+  kj::Exception e(kj::Exception::Type::FAILED, __FILE__, __LINE__, kj::str("test"));
+
+  for (int i = 0; i < 100; i++) {
+    e.addTrace(reinterpret_cast<void*>(static_cast<uintptr_t>(i + 1)));
+  }
+
+  auto trace = e.getStackTrace();
+  KJ_EXPECT(trace.size() == 100);
+  for (int i = 0; i < 100; i++) {
+    KJ_EXPECT(trace[i] == reinterpret_cast<void*>(static_cast<uintptr_t>(i + 1)));
+  }
+}
+
+KJ_TEST("extendTrace appends stack frames") {
+  kj::Exception e(kj::Exception::Type::FAILED, __FILE__, __LINE__, kj::str("test"));
+
+  // Add some initial trace entries.
+  e.addTrace(reinterpret_cast<void*>(uintptr_t(0x1)));
+  e.addTrace(reinterpret_cast<void*>(uintptr_t(0x2)));
+  KJ_EXPECT(e.getStackTrace().size() == 2);
+
+  // extendTrace should append current stack frames.
+  e.extendTrace(0);
+  KJ_EXPECT(e.getStackTrace().size() > 2);
+
+  // The first two entries should still be our original ones.
+  KJ_EXPECT(e.getStackTrace()[0] == reinterpret_cast<void*>(uintptr_t(0x1)));
+  KJ_EXPECT(e.getStackTrace()[1] == reinterpret_cast<void*>(uintptr_t(0x2)));
+}
+
+
+KJ_TEST("truncateCommonTrace removes common suffix") {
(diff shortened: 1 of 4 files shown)


c++/src/kj/async-coroutine-alloc.h
```diff
@@ -65,7 +80,7 @@ struct DebugCoroutineAllocator : public _::CoroutineAllocator {
 
   using Frame = CoroutineFrame<DebugCoroutineAllocator>;
```
harrishancock: Nit: calling this the FrameHeader would make a bit more sense to me. Then we could say "`alloc()`'s `frameSize` parameter refers to the size of the data after the `CoroutineFrameHeader` object.". Right now `frameSize` refers to the data after the `CoroutineFrame` object, which doesn't really make sense to me.

c++/src/kj/async-coroutine-alloc.h
```diff
@@ -25,35 +25,50 @@
...
-  Allocator &allocator;
+#endif
+  Allocator& allocator;
   kj::byte data[];
 
-  inline constexpr kj::byte *dataBegin() { return data; }
-  inline constexpr size_t allocSize() {
-    return sizeof(CoroutineFrame) + dataSize;
+  inline constexpr kj::byte* dataBegin() {
+    return data;
```
harrishancock, line 66: We need to make sure this is aligned correctly somehow, right?

If `sizeof(CoroutineFrame)` is always a multiple of `__STDCPP_DEFAULT_NEW_ALIGNMENT__`, then we're probably good, and `data` will naturally be aligned to the default `new` alignment. Perhaps we could just static assert that.

I'm curious if the compiler arranges to call overloads of [`operator new` that accept `std::align_val_t`](https://eel.is/c++draft/basic.stc.dynamic.allocation#3.1) arguments, though. If they do, it'd be nice to honor those alignment requests.

c++/src/kj/async-coroutine-alloc.h
```diff
@@ -77,29 +92,91 @@ struct DebugCoroutineAllocator : public _::CoroutineAllocator {
...
+    kj::byte* alloc(size_t n) {
+      KJ_IREQUIRE(offset + n < size(), "stack overflow: not implemented");
+      auto ptr = bytes + offset;
+      offset += n;
+      ASAN_UNPOISON_MEMORY_REGION(ptr, n);
+      return ptr;
+    }
+
+    void free(kj::byte* ptr, size_t n) {
+      KJ_IREQUIRE(ptr + n == bytes + offset, "out of order free");
```
harrishancock, line 161: This would catch mistakes like this?

```cpp
template <typename Alloc>
Promise<void> foo(Alloc& alloc) {
  auto p1 = bar(alloc);
  co_await baz(alloc);
  co_await p1;
}
```

mikea: yes. Only immediately awaitables coroutines can use the stack as of now.

c++/src/kj/async-coroutine-alloc.h
mikea: added proper alignment (turns out __STDCPP_DEFAULT_NEW_ALIGNMENT__ == 16) with assert. I've added a todo to investigate aligned new call.

PR #2559, opened 2026-02-08
Implement RpcFlowController that adaptively adjusts for BDP.

kentonv:
The controller measures the stream's RTT and bandwidth and tries to keep the window equal to RTT * bandwidth, plus some margin so that if more bandwidth is available, it'll notice and adjust.

This is a transliteration of the algorithm introduced in Cap'n Web here: https://github.com/cloudflare/capnweb/pull/132

The transliteration was performed with the help of Claude + Opencode: https://share.opencode.cloudflare.dev/share/xZgIlJe4

### DIFF
--- a/c++/src/capnp/rpc-twoparty.c++
+++ b/c++/src/capnp/rpc-twoparty.c++
@@ -81,6 +81,12 @@ TwoPartyVatNetwork::TwoPartyVatNetwork(kj::AsyncCapabilityStream& stream, uint m
 
 TwoPartyVatNetwork::~TwoPartyVatNetwork() noexcept(false) {};
 
+static bool adaptiveFlowControlEnabled = false;
+
+void TwoPartyVatNetwork::useAdaptiveFlowControl() {
+  adaptiveFlowControlEnabled = true;
+}
+
 MessageStream& TwoPartyVatNetwork::getStream() {
   KJ_SWITCH_ONEOF(stream) {
     KJ_CASE_ONEOF(s, MessageStream*) {
@@ -255,7 +261,11 @@ private:
 };
 
 kj::Own<RpcFlowController> TwoPartyVatNetwork::newStream() {
-  return RpcFlowController::newVariableWindowController(*this);
+  if (adaptiveFlowControlEnabled) {
+    return RpcFlowController::newAdaptiveController(getWindow(), clock);
+  } else {
+    return RpcFlowController::newVariableWindowController(*this);
+  }
 }
 
 size_t TwoPartyVatNetwork::getWindow() {
--- a/c++/src/capnp/rpc-twoparty.h
+++ b/c++/src/capnp/rpc-twoparty.h
@@ -51,16 +51,16 @@ class TwoPartyVatNetwork: public TwoPartyVatNetworkBase,
 public:
   TwoPartyVatNetwork(MessageStream& msgStream,
                      rpc::twoparty::Side side, ReaderOptions receiveOptions = ReaderOptions(),
-                     const kj::MonotonicClock& clock = kj::systemCoarseMonotonicClock());
+                     const kj::MonotonicClock& clock = kj::systemPreciseMonotonicClock());
   TwoPartyVatNetwork(MessageStream& msgStream, uint maxFdsPerMessage,
                      rpc::twoparty::Side side, ReaderOptions receiveOptions = ReaderOptions(),
-                     const kj::MonotonicClock& clock = kj::systemCoarseMonotonicClock());
+                     const kj::MonotonicClock& clock = kj::systemPreciseMonotonicClock());
   TwoPartyVatNetwork(kj::AsyncIoStream& stream, rpc::twoparty::Side side,
                      ReaderOptions receiveOptions = ReaderOptions(),
-                     const kj::MonotonicClock& clock = kj::systemCoarseMonotonicClock());
+                     const kj::MonotonicClock& clock = kj::systemPreciseMonotonicClock());
   TwoPartyVatNetwork(kj::AsyncCapabilityStream& stream, uint maxFdsPerMessage,
                      rpc::twoparty::Side side, ReaderOptions receiveOptions = ReaderOptions(),
-                     const kj::MonotonicClock& clock = kj::systemCoarseMonotonicClock());
+                     const kj::MonotonicClock& clock = kj::systemPreciseMonotonicClock());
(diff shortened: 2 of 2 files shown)


c++/src/capnp/rpc.c++
```diff
@@ -4849,6 +4849,298 @@ private:
...
+private:
+  struct SendSnapshot {
+    // Records the state of the world at the time that a particular message was sent. We hold onto
+    // this information until the message is acked, then use it to update the window size.
+
+    kj::TimePoint sentTime = kj::origin<kj::TimePoint>();
+    // Time when tnhis message was sent.
+
+    size_t size = 0;
+    // Size of the mesasge in bytes.
```
justin-mp: nit: spelling.

c++/src/capnp/rpc.c++
```diff
@@ -4849,6 +4849,298 @@ private:
...
+
+    uint64_t deliveredAtSend = 0;
+    kj::Maybe<kj::TimePoint> deliveredTimeAtSend;
+    // At the time of send, what was the time of the last ack (if any), and the total bytes acked?
```
justin-mp, line 4941: Why is this a question?

c++/src/capnp/rpc.c++
```diff
@@ -4849,6 +4849,298 @@ private:
...
+
+        // Check startup exit: has the window stopped growing meaningfully?
+        if (inStartupPhase) {
+          bool newRound = true;
+          KJ_IF_SOME(rst, roundStartTime) {
+            newRound = (snapshot.sentTime >= rst);
+          }
+
+          if (newRound) {
+            if (window > lastRoundWindow * 5 / 4) {
```
justin-mp: It's annoying that this `* 5 / 4` construction appears in 2 places.  Is there a way to refactor `applyGrowth` so we can re-use the same function here and there?  Maybe `grow()` does `* 5 / 4` and `applyGrowth` calls `grow()` if `!inStartupPhase`?

c++/src/capnp/rpc.c++
```diff
@@ -4849,6 +4849,298 @@ private:
...
+    }
+
+    // Release blocked senders if the window now has room.
+    KJ_SWITCH_ONEOF(state) {
+      KJ_CASE_ONEOF(blockedSends, Running) {
+        if (isReady()) {
+          for (auto& fulfiller: blockedSends) {
+            fulfiller->fulfill();
+          }
+          blockedSends.clear();
```
justin-mp, line 5110: Why is it OK to release all of the senders once we have room?  If the (number of senders) * (the size of each send()) is much greater than the window, won't the senders (potentially) immediately send data and we'll we wind up sending way more data than the window would allow?

c++/src/capnp/rpc.c++
kentonv: The question is "what was ... the total bytes acked?", but I can rephrase it.

c++/src/capnp/rpc.c++
kentonv: In the common case there is only one sender. It generally doesn't make sense to have multiple writers to a single stream.

But if for some reason multiple concurrent calls to `send()` were made, all of which blocked, we have no idea which of those calls plan to send again once the promise resolves. If we only unblocked one sender, it might be that that sender has nothing else to do, and the other senders just remain blocked. So we need to unblock all of them to give all of them a chance to proceed.

This is how the existing fixed-widow controller has always worked FWIW.

c++/src/capnp/rpc.c++
kentonv: Thought about it for a while and the current phrasing seems best to me.

c++/src/capnp/rpc.c++
kentonv: Split out `applyStartupGrowth()` and `applySteadyGrowth()`.

(Original code in Cap'n Web just used floating-point math... it all gets harder with integers...)

c++/src/capnp/rpc.c++
```diff
@@ -4849,6 +4849,305 @@ private:
...
+      }
+    }
+    KJ_UNREACHABLE;
+  }
+
+  kj::Promise<void> waitAllAcked() override {
+    KJ_IF_SOME(q, state.tryGet<Running>()) {
+      if (!q.empty()) {
+        auto paf = kj::newPromiseAndFulfiller<kj::Promise<void>>();
+        emptyFulfiller = kj::mv(paf.fulfiller);
```
zachlefevre, line 4921: Is there a bug here with calling waitAllAcked multiple times?
One could do:
```
auto promise1 = controller->waitAllAcked();
auto promise2 = controller->waitAllAcked();

co_await promise1; //   <- this should then _fail_ (I think) because the fulfiller is dropped.
```

c++/src/capnp/rpc.c++
```diff
@@ -4849,6 +4849,305 @@ private:
...
+  }
+
+  void taskFailed(kj::Exception&& exception) override {
+    KJ_SWITCH_ONEOF(state) {
+      KJ_CASE_ONEOF(blockedSends, Running) {
+        // Fail out all pending sends.
+        for (auto& fulfiller: blockedSends) {
+          fulfiller->reject(kj::cp(exception));
+        }
+        // Fail out all future sends.
```
zachlefevre, line 5133: This says `fail out all future sends` but it won't. One other send will go through before the state is checked and the exception is thrown. One would need to check the state before `message->send()`  to block all future sends.

c++/src/capnp/rpc.c++
kentonv: The existing `WindowFlowController` has exactly the same logic.

The RPC system never makes multiple overlapping calls to this, so it's not worth the extra complexity / allocations to handle "correctly".

c++/src/capnp/rpc.c++
kentonv: Actually, all future `send()`s will fail (throw an exception), but they will also all actually send the message before failing.

The behavior is debatable, but `WindowFlowController` works the same way. Ultimately it doesn't matter that much because the receiving end will fail all future stream calls either way, and avoiding the send on this end is just an optimization. Meanwhile, if we did actually skip the `message->send()`, we have to be careful to unwind some state in the RPC system that assumes the send actually happened. I'd rather just leave it as-is since the current logic has worked for years.

PR #2560, opened 2026-02-09
Re-land niche value optimized Maybe<T>

harrishancock:
This PR cherry-picks the original 8 commits, with no changes, from my previous branch #2514.

I've added one commit on top of that to fix Maybe's assignment operators.

The issue turns out to be both a pre-existing bug in the main Maybe template implementation, and a new bug in the new NullableValue specialization I wrote. Since fixing it would require changing the Maybe assignment operator implementations anyway, I opted to just delete the NullableValue assignment operators entirely. This made things easier to reason about, and reduces the surface area for bugs.

### DIFF
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -1341,74 +1341,10 @@ public:
   }
   inline NullableValue(decltype(nullptr)): isSet(false) {}
 
-  inline NullableValue& operator=(NullableValue&& other) {
-    if (&other != this) {
-      // Careful about throwing destructors/constructors here.
-      if (isSet) {
-        isSet = false;
-        dtor(value);
-      }
-      if (other.isSet) {
-        ctor(value, kj::mv(other.value));
-        isSet = true;
-      }
-    }
-    return *this;
-  }
+  // Note: Assignment operators (except nullptr) are intentionally not provided here.
+  // Maybe<T> implements its own assignment operators that are safe against the case where
+  // `this` owns `other` (e.g., head = kj::mv(head->next) in a linked list).
 
-  inline NullableValue& operator=(NullableValue& other) {
-    if (&other != this) {
-      // Careful about throwing destructors/constructors here.
-      if (isSet) {
-        isSet = false;
-        dtor(value);
-      }
-      if (other.isSet) {
-        ctor(value, other.value);
-        isSet = true;
-      }
-    }
-    return *this;
-  }
-
-  inline NullableValue& operator=(const NullableValue& other) {
-    if (&other != this) {
-      // Careful about throwing destructors/constructors here.
-      if (isSet) {
-        isSet = false;
-        dtor(value);
-      }
-      if (other.isSet) {
-        ctor(value, other.value);
-        isSet = true;
-      }
-    }
-    return *this;
-  }
-
-  inline NullableValue& operator=(T&& other) { emplace(kj::mv(other)); return *this; }
-  inline NullableValue& operator=(T& other) { emplace(other); return *this; }
-  inline NullableValue& operator=(const T& other) { emplace(other); return *this; }
-  template <typename U>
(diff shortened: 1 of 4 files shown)


harrishancock:
I confirmed that this resolves the issue uncovered in workerd tests.

c++/src/kj/common.h
```diff
@@ -1865,30 +1754,78 @@ class Maybe {
...
+  // 3. If emplace() throws, `this` is left in the none state (emplace has exception safety).
+
+  inline Maybe& operator=(T&& other) {
+    T temp(kj::mv(other));
+    ptr.emplace(kj::mv(temp));
+    return *this;
+  }
+  inline Maybe& operator=(T& other) {
+    T temp(other);
+    ptr.emplace(temp);
```
kentonv: I think this should still do `mv(temp)`?

Although `temp` itself was constructed using a copy-constructor, that doesn't mean we have to use the copy constructor a second time here.

c++/src/kj/common.h
```diff
@@ -1865,30 +1754,78 @@ class Maybe {
...
+    return *this;
+  }
+  inline Maybe& operator=(const T& other) {
+    T temp(other);
+    ptr.emplace(temp);
```
kentonv: Same here.

c++/src/kj/common.h
```diff
@@ -1865,30 +1754,78 @@ class Maybe {
...
+    } else {
+      *this = kj::none;
+    }
+    return *this;
+  }
+  inline Maybe& operator=(Maybe& other) {
+    if (this == &other) return *this;
+    Maybe temp(other);
+    KJ_IF_SOME(value, temp) {
+      ptr.emplace(value);
```
kentonv: mv(value)

c++/src/kj/common.h
```diff
@@ -1865,30 +1754,78 @@ class Maybe {
...
+    } else {
+      *this = kj::none;
+    }
+    return *this;
+  }
+  inline Maybe& operator=(const Maybe& other) {
+    if (this == &other) return *this;
+    Maybe temp(other);
+    KJ_IF_SOME(value, temp) {
+      ptr.emplace(value);
```
kentonv: mv(value)

c++/src/kj/common.h
```diff
@@ -1865,30 +1754,78 @@ class Maybe {
...
+    T temp(other);
+    ptr.emplace(temp);
+    return *this;
+  }
 
-  inline Maybe& operator=(Maybe&& other) { ptr = kj::mv(other.ptr); other = kj::none; return *this; }
-  inline Maybe& operator=(Maybe& other) { ptr = other.ptr; return *this; }
-  inline Maybe& operator=(const Maybe& other) { ptr = other.ptr; return *this; }
+  inline Maybe& operator=(Maybe&& other) {
+    if (this == &other) return *this;
```
kentonv: This check isn't strictly necessary, right? The code below will work correctly even in this case.

(Same comment for two other assignment operators below.)

c++/src/kj/common.h
```diff
@@ -1865,30 +1754,78 @@ class Maybe {
...
     }
     return *this;
   }
   template <typename U>
   Maybe& operator=(const Maybe<U>& other) {
-    KJ_IF_SOME(val, other) {
-      ptr.emplace(val);
+    Maybe<U> temp(other);
+    KJ_IF_SOME(value, temp) {
+      ptr.emplace(value);
```
kentonv: mv(value)

c++/src/kj/common.h
harrishancock: Good point, fixed.

c++/src/kj/common.h
harrishancock: Removed.

harrishancock:
Amended the last commit to address feedback.

harrishancock:
Rebased on v2.

PR #2563, opened 2026-02-12
Fix cross-type Maybe assignment to not copy source Maybe

harrishancock:
I managed to screw up again -- my previous fix broke "copy" assignment from `Maybe<String>` to `Maybe<StringPtr>`. I'm not sure how I missed this; I distinctly remember running all downstream tests. :/

Claude's prose follows.

---

When assigning `Maybe<T>` from `Maybe<U>` (cross-type), the assignment operators were constructing a Maybe<U> temporary, which requires U to be copyable. This broke cases like assigning `Maybe<StringPtr>` from `Maybe<String>`, since String is not copyable.

The fix is to construct a `Maybe<T>` temporary instead, using the existing converting constructors. This converts U -> T directly without needing to copy the `Maybe<U>`. The converting constructors already handle nullifying the source for move operations.

Added tests for:
- `Maybe<StringPtr>` assigned from `Maybe<String>` (the original bug)
- Cross-type assignment with move-only types (`Own<Derived>` -> `Own<Base>`)
- Self-assignment (m = m, m = kj::mv(m))

### DIFF
--- a/c++/src/kj/maybe-test.c++
+++ b/c++/src/kj/maybe-test.c++
@@ -1691,5 +1691,83 @@ KJ_TEST("Maybe<T> T-value copy-assignment is safe when this owns other") {
   KJ_EXPECT(KJ_ASSERT_NONNULL(head).value == 2);
 }
 
+// =======================================================================================
+// Cross-type assignment tests
+
+KJ_TEST("Maybe<StringPtr> assigned from Maybe<String>") {
+  // String is not copyable, so cross-type assignment must not try to copy the Maybe<String>.
+  // Instead it should convert String -> StringPtr directly.
+
+  // Copy-assignment from const Maybe<String>&
+  Maybe<String> s = kj::str("hello");
+  Maybe<StringPtr> sp;
+  sp = s;
+  KJ_EXPECT(KJ_ASSERT_NONNULL(sp) == "hello");
+  KJ_EXPECT(s != kj::none);  // source unchanged
+
+  // Note: Move-assignment (sp = kj::mv(s)) would leave sp with a dangling StringPtr because
+  // StringPtr is non-owning and the String would be destroyed. This is expected behavior.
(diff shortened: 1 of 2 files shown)


c++/src/kj/maybe-test.c++
```diff
@@ -1691,5 +1691,83 @@ KJ_TEST("Maybe<T> T-value copy-assignment is safe when this owns other") {
...
+  // Instead it should convert String -> StringPtr directly.
+
+  // Copy-assignment from const Maybe<String>&
+  Maybe<String> s = kj::str("hello");
+  Maybe<StringPtr> sp;
+  sp = s;
+  KJ_EXPECT(KJ_ASSERT_NONNULL(sp) == "hello");
+  KJ_EXPECT(s != kj::none);  // source unchanged
+
+  // Note: Move-assignment (sp = kj::mv(s)) would leave sp with a dangling StringPtr because
```
harrishancock, line 1708: I was initially surprised that this was not a compile failure. I realized that it's because Maybe's assignment operators do not delegate to its wrapped type's assignment operators. That is, `Maybe<StringPtr>::operator=(Maybe<String>&&)` does not invoke, or even require to exist, `StringPtr::operator=(String&&)`, which is (correctly) deleted.

I initially wanted to add a `requires` clause to Maybe's assignment operators to require that `T::operator=(U&&)` exists for the cross-type move assignment operator, and similar clauses for the others. I tried it, and it indeed made this case a compile failure, which would be really nice.

I held off because I worried that it would break downstream code. It didn't break anything in capnproto, but when I tried adding similar `requires` clauses to the non-cross-type assignment operators, it did break some code in KJ, which made me worry.

As an aside, it'd be nice if we could have Maybe's assignment operators actually invoke T's assignment operators, when both Maybes are value-ful. It seems a bit tricky, though.

PR #2564, opened 2026-02-13
Always use CRC32C for intHash

fhanau:
This addresses a footgun in hash.h – admittedly something from a change I suggested myself: If hash.h is included in several source files that are compiled with different compilation options (e.g. some with -msse4.2 and some without for x64), the binary may contain several versions of intHash32/intHash64 that return different values. This may have caused https://github.com/cloudflare/workerd/issues/5977, as reported in https://github.com/cloudflare/workerd/pull/6078 (I did not confirm this myself yet, but it sounds plausible).
One way to fix that is to always use the same hash function: CRC32C works well for this purpose. Accordingly, we switch from plain CRC32 to CRC32C for arm64 and use a bitwise implementation as a fallback. If more performance is desired for the fallback version, I can prepare that too (http://create.stephan-brumme.com/crc32/ lists many implementations with varying levels of complexity, I'd just have to rebuild the lookup table for CRC32C). Note that the downstream production environment is not affected by this bug, there we use consistent compile flags.

We can easily fix the issue in workerd by enabling the right compile flags more widely, but I think we'll want to fix the underlying footgun. I'm not attached to this approach, there are some alternative approaches to address this problem:
- We could mandate that CRC32C is available for arm64/x64 (which is the case for all x64 CPUs in the last 15 years and all recent arm64 CPUs) and only use the fallback on other architectures, that way we get the same implementation each time as long as we don't mix architectures.
- We could mandate that consistent compile flags are used, for Bazel this is relatively easy by adding a copts parameter to the KJ cc_library target, which will be inherited by all dependent targets. This is more difficult for cmake though.
- We could move intHash32/intHash64 to src/kj/hash.c++ so that there is only one copy of it, but I assume it was placed in the header for easier inlining/to avoid the function call overhead
- Detecting whether hardware support is available at runtime by looking up CPUID can also be a good option, but it feels slightly overkill for capnproto

### DIFF
--- a/c++/src/kj/hash.h
+++ b/c++/src/kj/hash.h
@@ -223,6 +223,59 @@ inline uint HashCoder::operator*(T e) const {
   return operator*(static_cast<__underlying_type(T)>(e));
 }
 
+#ifndef __CRC32__
+#ifndef __ARM_FEATURE_CRC32
+// CRC32 implementation taken from https://create.stephan-brumme.com/crc32/, provided under Zlib
+// license:
+/*
+This software is provided 'as-is', without any express or implied
+warranty. In no event will the authors be held liable for any damages
+arising from the use of this software.
+
+Permission is granted to anyone to use this software for any purpose,
+including commercial applications, and to alter it and redistribute it
+freely, subject to the following restrictions:
+
+1. The origin of this software must not be misrepresented; you must not
+   claim that you wrote the original software. If you use this software
+   in a product, an acknowledgment in the product documentation would be
+   appreciated but is not required.
+2. Altered source versions must be plainly marked as such, and must not be
+   misrepresented as being the original software.
+3. This notice may not be removed or altered from any source distribution.zlib License
+*/
+
+inline uint32_t crc32_bitwise(const void* data, size_t length, uint32_t previousCrc32)
+{
+  // The polynomial for CRC32C. We use this since hardware implementations are available for x64 and
+  // arm64, zlib CRC32 (polynomial 0xEDB88320) is only available on arm64.
+  uint32_t Polynomial = 0x82F63B78;
+
+  uint32_t crc = ~previousCrc32; // same as previousCrc32 ^ 0xFFFFFFFF
+  const uint8_t* current = (const uint8_t*) data;
+
+  while (length-- != 0)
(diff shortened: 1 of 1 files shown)


kentonv:
I wonder if we could somehow detect inconsistent build flags in such a way that it creates a linker error?

No one would ever want _some_ of their translation units to use hardware acceleration while others use software. I also worry that it's difficult to prove the software version is equivalent to the hardware version (could there be different variants of crc32?), and really hard to test this. So I'd rather just make it impossible to use inconsistent hash implementations across translation units.

fhanau:
> I wonder if we could somehow detect inconsistent build flags in such a way that it creates a linker error?
> 
> No one would ever want _some_ of their translation units to use hardware acceleration while others use software. I also worry that it's difficult to prove the software version is equivalent to the hardware version (could there be different variants of crc32?), and really hard to test this. So I'd rather just make it impossible to use inconsistent hash implementations across translation units.

Throwing around around a few more ideas:
I can't think of a way to enforce a linker error here; another approach could be to have the linker always choose the accelerated version if both are present, weak symbols go in that direction but aren't sufficient I think. Another approach for problems like this could be to use compiler builtins, but those are only defined if hardware support is enabled.
As for different CRC implementations, there are two commonly used variants (CRC32C and zlib CRC32), so yeah that is something we need to get right.

Maybe one solution would be to add a `KJ_USE_CRC_HASH` define so that hardware-based crc32 will always be used when the define is enabled (resulting in a compile error if hardware support is not enabled), and the bit shift hash is used otherwise. That way we always get the same hash as long as the define is enabled across all compile units, which is easy to do in Bazel. Unfortunately, setting architecture-dependent flags for all compile units is hard in Bazel, as mentioned we could set the flags for all targets that depend on libkj though which should also be sufficient.

kentonv:
Can we just set the flag on for all users of KJ's bazel build?

Or offer a config option that does that, which workerd should use?

But if there's really no way to do a config option, then I'm fine hard-coding the flag on for everyone. Our bazel build is really intended for consumption by workerd, if other people have a problem with it they can figure out a better solution.

fhanau:
Closing this, in favor of #2582.

PR #2580, opened 2026-02-25
arbitrary-sized stacktraces

mikea:
Exposes kj::Vector as storage (rather than array), to allow modifications in addTrace etal and reduce number of copies.

Supercedes https://github.com/capnproto/capnproto/pull/2569


c++/src/kj/exception.c++
```diff
@@ -385,30 +390,34 @@ String stringifyStackTraceWithLlvm(ArrayPtr<void* const> trace) {
...
-  for (auto& addr: space.first(size)) {
-    // The addresses produced by backtrace() are return addresses, which means they point to the
-    // instruction immediately after the call. Invoking addr2line on these can be confusing because
-    // it often points to the next line. If the next instruction is inlined from another function,
-    // the trace can be extra-confusing, since now it claims to be in a function that was not
-    // actually on the call stack. If we subtract 1 from each address, though, we get a much more
-    // reasonable trace. This may cause the addresses to be invalid instruction pointers if the
-    // instructions were multi-byte, but it appears addr2line is able to cope with this.
-    addr = reinterpret_cast<void*>(reinterpret_cast<uintptr_t>(addr) - 1);
+  kj::Vector<void*> vec(128);
```
kentonv: Note that vectors never shrink, so as written we now allocate 1k for every trace. I think the vast majority of traces are much smaller so maybe we should start with a smaller number?

c++/src/kj/exception.c++
```diff
@@ -385,30 +390,34 @@ String stringifyStackTraceWithLlvm(ArrayPtr<void* const> trace) {
...
+  for (;;) {
+    vec.resize(vec.capacity());
+    size_t size = backtrace(vec.begin(), vec.size());
+    if (size < vec.size()) {
+      // Adjust addresses: backtrace() returns return addresses, subtract 1 to point to the call
```
kentonv: The old code contained a much lengthier comment to explain this. Why shorten it? Can we keep the old comment?

c++/src/kj/exception.c++
```diff
@@ -385,30 +390,34 @@ String stringifyStackTraceWithLlvm(ArrayPtr<void* const> trace) {
...
-    // The addresses produced by backtrace() are return addresses, which means they point to the
-    // instruction immediately after the call. Invoking addr2line on these can be confusing because
-    // it often points to the next line. If the next instruction is inlined from another function,
-    // the trace can be extra-confusing, since now it claims to be in a function that was not
-    // actually on the call stack. If we subtract 1 from each address, though, we get a much more
-    // reasonable trace. This may cause the addresses to be invalid instruction pointers if the
-    // instructions were multi-byte, but it appears addr2line is able to cope with this.
-    addr = reinterpret_cast<void*>(reinterpret_cast<uintptr_t>(addr) - 1);
+  kj::Vector<void*> vec(128);
+  for (;;) {
```
kentonv, line 447: In case of a stack overflow from an infinite loop, the trace could be excessively large. Should we impose a limit?

c++/src/kj/exception.c++
```diff
@@ -760,9 +762,9 @@ namespace {
   static_assert(sizeof(ucontext->uc_mcontext) >= sizeof(win32Context),
       "mcontext_t should be an extension of CONTEXT");
   memcpy(&win32Context, &ucontext->uc_mcontext, sizeof(win32Context));
-  auto trace = getStackTrace(traceSpace, 0, GetCurrentThread(), win32Context);
+  auto trace = getStackTrace(0, GetCurrentThread(), win32Context);
 #elif __linux__ && __x86_64__
-  kj::ArrayPtr<void* const> trace;
+  kj::Vector<void*> trace;
```
kentonv: This is a signal handler. We cannot use `Vector` here.

We also use kj::getStackTrace() in signal handlers in the production runtime, so we probably need to keep this signal-safe API available.

c++/src/kj/exception.c++
mikea: sorry, oversight. I reverted the old implementation anyway.

mikea:
I kept the low-level versions, always start with 32-traces buffer on stack and switch to Vector if that is not enough. Added upper bound of 1024 frames. 

ptal @kentonv

c++/src/kj/exception.c++
```diff
@@ -605,18 +643,15 @@ StringPtr stringifyStackTraceAddresses(ArrayPtr<void* const> trace, ArrayPtr<cha
...
 }
 
 namespace {
 
 [[noreturn]] void terminateHandler() {
-  void* traceSpace[32]{};
-
   // ignoreCount = 3 to ignore std::terminate entry.
-  auto trace = kj::getStackTrace(traceSpace, 3);
+  auto trace = kj::getStackTrace(3);
```
mikea, line 663: terminateHandler is not a signal, we can allocate here, right?

kentonv: I believe that's correct.

kentonv (approved):

PR #2582, opened 2026-03-04
[build] Add define to consistently enable CRC32 hash

fhanau:
This should eliminate the footgun here. We use a define that basically means: I promise to use this define on all code that ends up including KJ headers; and I will enable the required hardware extensions (this part is enforced through a compiler error). This way, we can enable CRC extensions in the downstream project (where we can configure hardware extensions well) and for workerd release builds (hard to set in the general case, but easy enough if we only have to do it for release CI).

### DIFF
--- a/c++/src/kj/hash.h
+++ b/c++/src/kj/hash.h
@@ -24,9 +24,12 @@
 #include "string.h"
 #include <stdint.h>
 
-// clang has dedicated builtins for crc32 on arm64, for GCC we fall back to ARM ACLE intrinsics.
+// When crc32 is enabled and we have clang, use the dedicated builtins for crc32 on arm64, for GCC
+// we fall back to ARM ACLE intrinsics.
+#if KJ_HAS_CRC32
 #if __ARM_FEATURE_CRC32 && !__clang__
 #include <arm_acle.h>
+#endif
 #endif
 
 KJ_BEGIN_HEADER
@@ -237,13 +240,20 @@ inline uint intHash32(uint32_t i) {
 
   // On architectures with a hardware CRC32 instruction, use it. Otherwise fall back to a
   // reasonable shifty hash.
+  // To avoid mixing code using CRC32 and not using it, we either don't use crc32 (when KJ_HAS_CRC32
+  // is undefined) or require that hardware support for it is available. The define should always
+  // be used consistently, if it is not the link-time kjCrcHashGuard check will catch it.
+#if KJ_HAS_CRC32
 #if __CRC32__
   return __builtin_ia32_crc32si(0, i);
 #elif __ARM_FEATURE_CRC32
 #ifdef __clang__
-  return __builtin_arm_crc32w(0, i);
+  return __builtin_arm_crc32cw(0, i);
 #else
   return __crc32w(0, i);
+#endif
+#else
+#error "CRC32 hardware support must be available when KJ_HAS_CRC32 is set."
 #endif
 #else
   // Thomas Wang 32 bit integer hash function from https://gist.github.com/badboy/6267743
@@ -270,13 +280,17 @@ inline uint intHash64(uint64_t i) {
     return intHash32(i);
   }
 
+#if KJ_HAS_CRC32
 #if __CRC32__
   return __builtin_ia32_crc32di(0, i);
 #elif __ARM_FEATURE_CRC32
 #ifdef __clang__
-  return __builtin_arm_crc32d(0, i);
+  return __builtin_arm_crc32cd(0, i);
 #else
   return __crc32d(0, i);
+#endif
+#else
+#error "CRC32 hardware support must be available when KJ_HAS_CRC32 is set."
 #endif
 #else
   // Thomas Wang hash6432shift() from https://gist.github.com/badboy/6267743
--- a/c++/src/kj/hash.h
+++ b/c++/src/kj/hash.h
@@ -226,6 +226,19 @@ inline uint HashCoder::operator*(T e) const {
   return operator*(static_cast<__underlying_type(T)>(e));
 }
 
+// Ensure we don't mix different hash implementations. The template will only be instantiated in
+// translation units where intHash32 or intHash64 are called, so there is no overhead otherwise.
+extern const char kjCrc32HashEnabled;
+extern const char kjCrc32HashDisabled;
+
+template <bool enabled>
+struct kjCrcHashGuard {
+    kjCrcHashGuard() {
+    KJ_USED static constexpr const char* value =
+        enabled ? &kjCrc32HashEnabled : &kjCrc32HashDisabled;
+    }
+};
+
 inline uint intHash32(uint32_t i) {
   // Basic 32-bit integer hash function.
   //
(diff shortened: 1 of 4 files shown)


fhanau:
Will rework this entirely – discussed with Mike and I think we'll end up making the functionality opt-in so that it is limited to builds in the downstream project/workerd release builds.

c++/src/kj/hash.h
```diff
@@ -24,10 +24,13 @@
 #include "string.h"
 #include <stdint.h>
 
-// clang has dedicated builtins for crc32 on arm64, for GCC we fall back to ARM ACLE intrinsics.
+// When crc32 is enabled and we have clang, use the dedicated builtins for crc32 on arm64, for GCC
+// we fall back to ARM ACLE intrinsics.
+#if KJ_HAS_CRC32
```
kentonv, line 29: I don't feel like this guard accomplishes much as it seems just as easy to be inconsistent about the `-DKJ_HAS_CRC32` flag as about the hardware feature flags. I know that in our particular setup, that turns out not to be the case, but I'm not convinced that's a general problem.

What I'd much rather do is figure out a way to force a linker error when different translation units have conflicting flags. Can we output some sort of symbol that will merge with other definitions as long as they are all the same but cause the linker to error out if they differ?

In the absence of that, I think we should just leave the code alone and force the hardware flags on in the bazel build.

fhanau: The motivation for doing it this way is that under Bazel, it is easy to apply a given define to all code that uses KJ (`defines` on a cc_library are inherited by all dependent targets) but difficult to apply compiler flags to all code that uses KJ, or to all C++ code – the same does not apply for the `copts` parameter which is not inherited. That means we unfortunately can't easily enable the required hardware extensions across build configurations, we simply don't know if we're on x64 or arm64 within Bazel (unless we tell users to pass a flag with the architecture they have which is cumbersome).
I'm not aware of a way to solve this at the linker level.

kentonv: > I'm not aware of a way to solve this at the linker level.

Claude has a couple of ideas: https://claude.ai/share/e79b6616-0a6b-4b08-a2f4-369360c96f6d

fhanau: Finally got back to this – at least some of the suggestions were workable and I implemented a template-based approach that enforces that we never mix the two hash implementations. Not sure if this is the best approach (feels kind of overengineered) but it works as intended locally.
I kept the previous commit adding the KJ_HAS_CRC32 since at least with workerd non-release builds, we'll likely have to support building with inconsistent flags – this is not great but also hard to avoid – by only enabling that define for release builds there, we still get this issue right.

fhanau: Also turns out abseil (and by extension Chromium) had a very similar issue recently, but it looks like they didn't have a perfect solution for it either. (Dan Carney also noticed this and commented in the thread) https://issues.chromium.org/issues/475029685#comment11

c++/src/kj/hash.h
```diff
@@ -223,6 +226,19 @@ inline uint HashCoder::operator*(T e) const {
...
+// translation units where intHash32 or intHash64 are called, so there is no overhead otherwise.
+extern const char kjCrc32HashEnabled;
+extern const char kjCrc32HashDisabled;
+
+template <bool enabled>
+struct kjCrcHashGuard {
+    kjCrcHashGuard() {
+    KJ_USED static constexpr const char* value =
+        enabled ? &kjCrc32HashEnabled : &kjCrc32HashDisabled;
+    }
```
kentonv:
```suggestion
  kjCrcHashGuard() {
    KJ_USED static constexpr const char* value =
        enabled ? &kjCrc32HashEnabled : &kjCrc32HashDisabled;
  }
```

c++/src/kj/hash.h
```diff
@@ -237,15 +253,26 @@ inline uint intHash32(uint32_t i) {
...
+  // be used consistently, if it is not the link-time kjCrcHashGuard check will catch it.
+#if KJ_HAS_CRC32
+  kjCrcHashGuard<true> guard;
+  (void)guard;
 #if __CRC32__
   return __builtin_ia32_crc32si(0, i);
 #elif __ARM_FEATURE_CRC32
 #ifdef __clang__
-  return __builtin_arm_crc32w(0, i);
+  return __builtin_arm_crc32cw(0, i);
```
kentonv, line 266: Seems you changed this from crc32w to crc32cw, but didn't change the non-clang branch two lines down. I think we had better be consistent between clang and non-clang on the same architecture, in case people use different compilers for different components. Apparently cw is faster than w so I guess it's a good change as long as we apply it to both.

c++/src/kj/hash.h
```diff
@@ -270,15 +297,23 @@ inline uint intHash64(uint64_t i) {
...
 
+#if KJ_HAS_CRC32
+  kjCrcHashGuard<true> guard;
+  (void)guard;
 #if __CRC32__
   return __builtin_ia32_crc32di(0, i);
 #elif __ARM_FEATURE_CRC32
 #ifdef __clang__
-  return __builtin_arm_crc32d(0, i);
+  return __builtin_arm_crc32cd(0, i);
```
kentonv, line 307: Similar to above: If we're changing this to -cd, better change the one two lines down as well.

c++/src/kj/hash.h
kentonv: > with workerd non-release builds, we'll likely have to support building with inconsistent flags

Why?

fhanau: > > with workerd non-release builds, we'll likely have to support building with inconsistent flags
> 
> Why?

To my knowledge, there's no convenient way to set platform-dependent compiler flags, which we need to enable the CRC32 extensions (-mcrc for arm64, -msse4.2 for x64). For OS-dependent flags, this is easy using the [--enable_platform_specific_config](https://bazel.build/reference/command-line-reference#common_options-flag--enable_platform_specific_config) flag, but there's no architecture-specific equivalent. Internally, we set these extensions using a custom Bazel toolchain, where we can reliably set the flags we need. But in workerd we don't have a custom toolchain, which leaves us with a couple of inconvenient options:
1. Enable the flags we need (or a config defined in .bazelrc that contains them) on the command line – for workerd release builds we can do this (see https://github.com/cloudflare/workerd/pull/6243), but it's inconvenient for human developers
2. Tell developers that they are required to enable the configuration themselves by putting it into their global .bazelrc file – this would mean having a one-time setup step before building workerd. We could use githooks to enforce that the config option is present and provide a simple shell command to automatically set the right config, but this still feels like an additional step that makes it harder for developers to get started with workerd.
3. Patch rules_cc (where the default Bazel C++ toolchains live) to include the flags we need – will make maintenance harder, we'd likely need to freeze rules_cc since the patch will need to be rebased after new releases
4. Apply the required flags to each individual cc_library target – this is conceptually easy using Starlark's select() function, and for code within workerd we can easily accomplish this by changing the wd_cc_library() definition, but to set the required flags for *every* source file, we'd need to patch many of our dependencies to add the flags to all the cc_library targets there. It would require lots of patches that need to be maintained.

2 and 3 are not terribly inconvenient, but still sound like a hassle overall

kentonv (approved):

PR #2588, opened 2026-03-10
Avoid signal starvation caused by rapidly repeating timers

vaci:
A kj::Timer of duration less than 1ms prevents signals from being delivered. Repeating timers of such duration prevent signals from ever being delivered.

### DIFF
--- a/c++/src/kj/async-unix.c++
+++ b/c++/src/kj/async-unix.c++
@@ -597,6 +597,13 @@ bool UnixEventPort::wait() {
     //   the parent process exit while the child thread lives on. In this case, if a UnixEventPort
     //   had been created before daemonizing, signal handling would be forever broken in the child.
 
+    // If the timeout is zero, use the alternative approach implemented by poll()
+    // to deliver pending signals. Otherwise a repeating timer with a timeout less than
+    // 1ms resolution will prevent signals from ever being delivered.
+    if (timeout == 0) {
+      return poll();
+    }
+
     sigset_t waitMask = originalMask;
 
     // Unblock the signals we care about.
--- a/c++/src/kj/async-unix-test.c++
+++ b/c++/src/kj/async-unix-test.c++
@@ -1025,6 +1025,26 @@ KJ_TEST("UnixEventPort poll for signals") {
   promise2.wait(waitScope);
 }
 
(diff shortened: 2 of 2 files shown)


c++/src/kj/async-unix.c++
```diff
@@ -597,6 +597,13 @@ bool UnixEventPort::wait() {
     //   the parent process exit while the child thread lives on. In this case, if a UnixEventPort
     //   had been created before daemonizing, signal handling would be forever broken in the child.
 
+    // If the timeout is zero, use the alternative approach implemented by poll()
+    // to deliver pending signals. Otherwise a repeating timer with a timeout less than
+    // 1ms resolution will prevent signals from ever being delivered.
+    if (timeout == 0) {
```
kentonv, line 603: Hmm. Arguably even in the absence of signals, it's problematic that a short timeout causes a busy loop. Or that even for a longer timeout, like 100.5ms, we will wake up a little bit too early, and then busy-loop waiting for it to be ready.

Perhaps we need to adjust the timeout code to round up instead of down? epoll_wait itself is documented as rounding the timeout up to the clock resolution -- it makes sense that we should also round up the value we pass to epoll_wait.

Also, there is an epoll_pwait2 now which accepts nanosecond resolution, introduced in Linux 5.11. Maybe we should be using that when available?

vaci: Rounding up would probably be worse for my scenarios, at least, because then I couldn't use timers to busy-wait. (I configure the delay of repeating timers according to how reactive I want our application to be). 

I'm actually using v1.3.0, otherwise I'd probably use `kj::yieldUntilWouldSleep()` instead.

kentonv: Well... I think rounding down is actually a bug here, in general. Seems like it would cause us to busy-loop in KJ for an average of 0.5 ms on every timer wakeup. Kind of surprised this hasn't caused more trouble TBH.

But with epoll_pwait2 I think you get everything you want?

vaci: It makes sense that you can only really schedule timers at the resolution that the OS offers, I think that's the usual case in event loop implementations. If caller really wants busy-wait behaviour, they can supply a timeout of zero explicitly.

PR #2591, opened 2026-03-10
optimize ExceptionOr<T> operator=

mikea:
(This build on top of https://github.com/capnproto/capnproto/pull/2592)

As noticed in https://github.com/capnproto/capnproto/pull/2592 this _typical_ code:

```
  void get(ExceptionOrValue& output) noexcept override {
    output.as<FixVoid<T>>() = kj::mv(result);
  }
```

results in compiler generating ~T and ~Exception calls.

This change makes operator= valid only with empty ExceptionOr and uses emplaceInit methods to generate more efficient code. This gives a free perf boost (https://gist.github.com/mikea/a6f11a3e082aa8d35c282a47dff49cbd):

- 3.6% on bm_Http_FullProtocol
- 5.5% is the biggest improvement for bm_Promise_Fib10

I have explored other design options, in particular, making get look like `get(kj::SpaceFor<ExceptionOrValue>&)` with various attempts to make this type safer, but 

a) all the bloat didn't materialize in performance gain
b) the changes are huge and risky and hard to land

so I think this represents a pragmatic middle ground. I'm curious what you think @kentonv 

Also worth mentioning that this PR is required to make full use of https://github.com/capnproto/capnproto/pull/2586 (otherwise inlining blows up ir cache, and I'd rather kept inlining since it is the whole purpose of optimization)

### DIFF
--- a/c++/src/kj/async-inl.h
+++ b/c++/src/kj/async-inl.h
@@ -67,7 +67,7 @@ protected:
   // Allow subclasses to have move constructor / assignment.
   ExceptionOrValue() = default;
   ExceptionOrValue(ExceptionOrValue&& other) = default;
-  ExceptionOrValue& operator=(ExceptionOrValue&& other) = default;
+  ExceptionOrValue& operator=(ExceptionOrValue&&) = delete;
 };
 
 template <typename T>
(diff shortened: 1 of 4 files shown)


c++/src/kj/async-inl.h
```diff
@@ -77,7 +77,14 @@ class ExceptionOr: public ExceptionOrValue {
   ExceptionOr(T&& value): value(kj::mv(value)) {}
   ExceptionOr(bool, Exception&& exception): ExceptionOrValue(false, kj::mv(exception)) {}
   ExceptionOr(ExceptionOr&&) = default;
-  ExceptionOr& operator=(ExceptionOr&&) = default;
+
+  inline ExceptionOr& operator=(ExceptionOr&& other) {
+    KJ_IREQUIRE(value == kj::none && exception == kj::none,
```
kentonv, line 82: I suppose this IREQUIRE is a bit redundant as the two emplaceInit()s below have their own IREQUIREs.

Though maybe the compiler is smart enough to optimize out the inner ones after seeing this one passed.

kentonv (approved):

c++/src/kj/async-inl.h
mikea: > Though maybe the compiler is smart enough to optimize out the inner ones after seeing this one passed.

It is a standard optimization. I double-checked and release-version of async-bench doesn't have low-level checks compiled in (we need this btw https://github.com/capnproto/capnproto/pull/2593 to match our production).

PR #2592, opened 2026-03-11
[maybe] emplaceInit

mikea:
The following expression `maybe = kj::mv(otherMaybe)` results in generationg code for ~T in case target maybe is not empty.

Sometimes (often) compiler can't prove the emptiness of maybe and fails to eliminate ~T branch. Not only this generates longer code (bad for instruction cache), it also prevents further optimizations from happening. This is particularly bad for ExceptionOr case (will be addressed in later PRs).

This pr introduces emplaceInit operations to be used when engineer can prove the empty target.

### DIFF
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -1318,6 +1318,13 @@ public:
     return value;
   }
 
+  template <typename... Params>
+  inline T& emplaceInit(Params&&... params) {
+    ctor(value, kj::fwd<Params>(params)...);
+    isSet = true;
+    return value;
+  }
+
   inline NullableValue(): isSet(false) {}
   inline NullableValue(T&& t)
       : isSet(true) {
@@ -1522,6 +1529,12 @@ public:
       initNone(&value);  // noexcept - leave in none state
       throw;
     }
+    return value;
+  }
+
+  template <typename... Params>
+  inline T& emplaceInit(Params&&... params) {
+    ctor(value, kj::fwd<Params>(params)...);
     return value;
   }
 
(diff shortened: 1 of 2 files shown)


harrishancock (approved):
Non-blocking nit: `emplaceEmpty()` could be interpreted as meaning, "emplace an empty value into this Maybe", rather than, "emplace a value into this empty Maybe".

Ideas: `uninitializedEmplace()`, `assumeUninitEmplace()`, `emplaceAssumingEmpty()`

But, naming is hard.

c++/src/kj/common.h
```diff
@@ -1318,6 +1318,13 @@ class NullableValue {
     return value;
   }
 
+  template <typename... Params>
+  inline T& emplaceEmpty(Params&&... params) {
```
kentonv: Should this `KJ_IREQUIRE(!isSet)`?

kentonv: nm, I guess you already IREQUIRE in the public-facing wrapper.

kentonv:
How about `emplaceInit()` as the method name?

kentonv:
And the move assignment could be `moveInit()` or could even just be an overload of `emplaceInit()`...

c++/src/kj/common.h
mikea: yep, I decided to have checks in public API to make sure impl doesn't have any overheads itself.

mikea:
> How about `emplaceInit()` as the method name?

done.

> And the move assignment could be `moveInit()` or could even just be an overload of `emplaceInit()`...

imo overload is a great idea. done.

ptal @kentonv

kentonv (approved):

PR #2595, opened 2026-03-16
revert to google_benchmark

mikea:
Also disable third-party warnings since benchmark generates some.


c++/.bazelrc
```diff
@@ -45,6 +45,10 @@ build:unix --cxxopt='-Wno-sign-compare' --host_cxxopt='-Wno-sign-compare'
 build:unix --cxxopt='-Wno-unused-parameter' --host_cxxopt='-Wno-unused-parameter'
 build:unix --cxxopt='-DKJ_HEADER_WARNINGS'
 build:unix --cxxopt='-DCAPNP_HEADER_WARNINGS'
+
+# do not show warnings from external libraries
+build:unix --per_file_copt='external@-Wno-everything'
```
fhanau: nit: Isn't `-w` (lower case) the more idiomatic option for this?

fhanau (approved):

PR #2596, opened 2026-03-16
ubsan build configuration

mikea:
more or less matches our downstream config

### DIFF
--- a/c++/.bazelrc
+++ b/c++/.bazelrc
@@ -95,6 +95,7 @@ build --//src/kj:brotli=True
 
 build:sanitizers-common --copt="-Og"
 build:sanitizers-common --copt="-fno-optimize-sibling-calls"
+build:sanitizers-common --copt="-fno-sanitize-merge"
 build:sanitizers-common --config=max-debug-info
 # sanitizers often do work on shutdown, do it cleanly
 build:sanitizers-common --test_env=KJ_CLEAN_SHUTDOWN=1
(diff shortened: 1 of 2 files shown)


c++/.bazelrc
```diff
@@ -95,6 +95,7 @@ build --//src/kj:brotli=True
 
 build:sanitizers-common --copt="-Og"
 build:sanitizers-common --copt="-fno-optimize-sibling-calls"
+build:sanitizers-common --copt="-fno-sanitize-merge"
```
fhanau, line 98: nit: I've never had issues with fsanitize-merge, my impression is that disabling it will just give away performance. Not blocking though, for the capnproto standalone build that's fine if you see a reason to disable it

c++/.bazelrc
```diff
@@ -103,6 +104,21 @@ build:asan --config=sanitizers-common
...
+# Enable non-default components of UBSan: https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html#available-checks
+# Many of these do not represent undefined behavior based on the C/C++ standards per se, but often
+# indicate bugs.
+build:ubsan --copt="-fsanitize=alignment" --copt="-fno-sanitize-recover=alignment"
+build:ubsan --copt="-fsanitize=float-divide-by-zero" --copt="-fno-sanitize-recover=float-divide-by-zero"
+build:ubsan --copt="-fsanitize=implicit-bitfield-conversion" --copt="-fno-sanitize-recover=implicit-bitfield-conversion"
+build:ubsan --copt="-fsanitize=local-bounds" --copt="-fno-sanitize-recover=local-bounds"
+build:ubsan --copt="-fsanitize=nullability" --copt="-fno-sanitize-recover=nullability"
+build:ubsan --copt="-fsanitize=signed-integer-overflow" --copt="-fno-sanitize-recover=signed-integer-overflow"
+# todo: doesn't link
```
fhanau, line 120: I think the link issues here might be due to different clang versions having different handling of vptr – some recent clang release made it so that vptr is no longer included under fsanitize=undefined by default, it needs to be specified separately at compile and link time. Maybe in this case there's mismatching versions where clang includes vptr but at link time compiler-rt doesn't know that it's needed? Otherwise I don't see how this internally consistent set of flags would cause the missing symbols issues.

fhanau:
(changes so far LGTM, not approving yet since there might be a bunch of UBSan warnings to fix after addressing the linker errors, so the PR will likely include more changes)

PR #2597, opened 2026-03-18
Add overload of OneOf::tryGet() when the OneOf is an rvalue.

kentonv:
Note this is actually a breaking change. There are 6 places in our edge runtime (zero in workerd) that do:

    kj::mv(KJ_ASSERT_NONNULL(kj::mv(someOneOf).tryGet<T>()))

The outer `kj::mv` is not just to longer necessary, it becomes an error. I have updated them in a concurrent PR.

### DIFF
--- a/c++/src/kj/one-of.h
+++ b/c++/src/kj/one-of.h
@@ -425,7 +425,7 @@ public:
   }
 
   template <typename T>
-  Maybe<T&> tryGet() {
+  Maybe<T&> tryGet() & {
     if (is<T>()) {
       return *reinterpret_cast<T*>(space);
     } else {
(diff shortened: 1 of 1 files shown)


c++/src/kj/one-of.h
harrishancock: MSVC says:
```
D:\a\capnproto\capnproto\c++\src\kj\one-of.h(436,19): error C2559: 'kj::Maybe<const U&> kj::OneOf<Variants...>::tryGet(void) const': cannot overload a member function without ref-qualifier with a member function with ref-qualifier [D:\a\capnproto\capnproto\build-output\src\kj\kj.vcxproj]
```

I think we just need `const &` here. (I tried suggesting it, but the suggestion tool only seems to work correctly for lines which were changed.)

harrishancock (approved):

kentonv:
MacOS flake is just the usual one with timers...

PR #2602, opened 2026-03-27
Add AsyncIoStreamWithGuards::getFd delegating to inner AsyncIoStream

joshthoward:


mikea (approved):

kentonv:
Hey I'm not sure this is correct. The raw FD doesn't enforce the "guards" that the stream is enforcing. Generally you cannot have `getFd()` return the underlying FD unless that FD is equivalent to the stream -- which in this case it is not.

kentonv:
Can you tell me what you're trying to accomplish?

Issue #2603, opened 2026-03-27
KJ_NONBLOCKING_SYSCALL leaks errno = EAGAIN after successful non-blocking operations


lilitjanpoladyan:
getOsErrorNumber() in kj/debug.c++ returns 0 when errno is EAGAIN/EWOULDBLOCK in non-blocking mode, correctly treating it as a non-error. However, it doesn't clear errno, so it remains set to EAGAIN after the macro completes successfully.

This causes errno to leak into application code. For example, after establishing a TwoPartyVatNetwork connection over a non-blocking Unix socket, errno stays set to EAGAIN from internal I/O operations. Any subsequent code that logs strerror(errno) 
sees spurious "Resource temporarily unavailable" messages.

Reproduction: Use TwoPartyVatNetwork with AsyncCapabilityStream (via wrapUnixSocketFd), exchange a few RPC messages, then check errno — it will be 11 (EAGAIN).

Proposed fix: Clear errno in getOsErrorNumber() when the EAGAIN/EWOULDBLOCK case is handled as non-error. I have a one-line fix ready and can open a PR if this approach is acceptable.

kentonv:
KJ never clears errno at all, EAGAIN or otherwise.

This is standard practice. Most code doesn't clear errno after handling an error. It just leaves it alone, and lets it get overwritten on the next error.

You can't detect if an error occurred by checking errno alone. You have to first observe a function return a value that indicates an error. Then you check errno to see what error it was.

There are a small number of functinos, like strtol(), whose "error" return value is also technically a valid result. The man page for strtol() says:

> Since strtol() can legitimately return 0, LONG_MAX, or LONG_MIN (LLONG_MAX or LLONG_MIN for strtoll()) on both success and failure, the calling program should set errno to 0 before the call, and then determine if an error occurred by checking whether errno == ERANGE after the call.

So, as you can see, it's expected that errno basically always contains a stale error value. If you are calling a function where you need to detect whether it sets errno, you have to manually set it back to zero first. The last code you called that handled an error is not expected to do this for you.

PR #2604, opened 2026-03-27
Fix errno leak nonblocking

lilitjanpoladyan:
Fixes https://github.com/capnproto/capnproto/issues/2603

`KJ_NONBLOCKING_SYSCALL` uses `getOsErrorNumber(true)` which returns 0
when `errno` is `EAGAIN`/`EWOULDBLOCK`, correctly treating it as a
successful non-blocking operation. However, `errno` itself is left set
to `EAGAIN`, causing it to leak into application code.

For example, after a `TwoPartyVatNetwork` connection is established over
a non-blocking Unix socket and messages are exchanged, `errno` remains
set to `EAGAIN` from internal I/O operations. Any code that subsequently
checks `errno` or calls `strerror(errno)` sees a spurious "Resource
temporarily unavailable".

The fix clears `errno` in `getOsErrorNumber()` when the
`EAGAIN`/`EWOULDBLOCK` case is handled as non-error.


kentonv:
Per my comment on the issue, this is not a bug.

PR #2606, opened 2026-03-28
Fix bug in membranePipeline() helper added in #2598.

kentonv:
Not sure how I got away with this not failing earlier. :facepalm:


kentonv:
Just gonna merge since this is a no-brainer.

PR #2609, opened 2026-03-31
mark ~Exception as KJ_NOINLINE

mikea:
This reduces the happy case code pressure across the code.

https://gist.github.com/mikea/903de5a262ee5b6a7d87f88d0c34aa85

Biggest winners are bm_Promise_Pow2_20 (-1.4%) and bm_Promise_Shift_20 (-1.3%) since compiler can better optimise `SimpleTransformPromiseNode::getImpl` now.

This micro-optimization is required to reduce inlining storm when switching Exception::storage to static disposer. (https://github.com/capnproto/capnproto/pull/2586)

### DIFF
--- a/c++/src/kj/exception.h
+++ b/c++/src/kj/exception.h
@@ -74,7 +74,7 @@ public:
   Exception(Type type, String file, int line, String description = nullptr) noexcept;
   Exception(const Exception& other) noexcept;
   Exception(Exception&& other) noexcept = default;
-  ~Exception() noexcept;
+  KJ_NOINLINE ~Exception() noexcept;
 
   const char* getFile() const { return storage->file; }
   int getLine() const { return storage->line; }


harrishancock (approved):

kentonv:
Hmm, does this mean the compiler can no longer tell that it can elide the destructor for an exception that was previously moved away? I would have expected that to be an important optimization...

mikea:
Turns out that optimization is not that important. Looks like having shorter code wins here. 


FWIW I separately work on bringing inline back here. There are couple more inefficiencies I have to fix before  enable it back.

mikea:
> Hmm, does this mean the compiler can no longer tell that it can elide the destructor for an exception that was previously moved away? I would have expected that to be an important optimization...

here's a specific example where we assume compiler is smart, but it isn't: https://github.com/capnproto/capnproto/blob/v2/c%2B%2B/src/kj/async-inl.h#L662C8-L662C15

```
  void getImpl(ExceptionOrValue& output) override {
    // Derive return type from DepT to reduce templating.
    typedef _::FixVoid<_::ReturnType<Func, _DepT>> T;
    typedef _::FixVoid<_DepT> DepT;
    ExceptionOr<DepT> depResult;
    getDepResult(depResult);
    KJ_IF_SOME(depException, depResult.exception) {
      output.as<T>() = handle<T>(
          MaybeVoidCaller<Exception, FixVoid<ReturnType<ErrorFunc, Exception>>>::apply(
              errorHandler, kj::mv(depException)));
    } else KJ_IF_SOME(depValue, depResult.value) {
      output.as<T>() = handle(MaybeVoidCaller<DepT, T>::apply(func, kj::mv(depValue)));
    }
  }
  ```
  
  This code does not destroy Exceptions. Yet, the decompiled version looks like this:

https://gist.github.com/mikea/36872f1fac6d25ee2292ffce75d7c06b

It will check the depResult.super_ExceptionOrValue.exception.ptr.field_0 for 0 again despite the fact that previous `if` has handled it.

Another source of missing optimization is when we take `Exception&&` parameter instead of `Exception` - compiler doesn't see move constructor if function call is not inlined.

PR #2614, opened 2026-04-02
Fix RpcFlowController to fulfill blocked senders with DISCONNECTED on destruction

shrima-cf:
When an AdaptiveFlowController or WindowFlowController is destroyed while senders are blocked waiting for window space, the pending PromiseFulfillers were being destroyed without being fulfilled or rejected. This caused KJ's WeakFulfillerBase::disposeImpl to reject the corresponding promises with a hardcoded FAILED exception type ("PromiseFulfiller was destroyed without fulfilling the promise"), regardless of the actual cause.

A flow controller being destroyed while senders are blocked means the underlying connection is going away — this is a disconnection, not a logic bug. The FAILED exception type is misleading and causes callers that distinguish between exception types to mishandle the error. Add explicit destructors to both AdaptiveFlowController and WindowFlowController that reject any pending blocked-sender fulfillers with DISCONNECTED before the implicit member destruction runs. This matches the semantics already used by taskFailed() when an ack promise fails.

Also make StreamSimulator accept an optional RpcFlowController parameter so it can be used to test any flow controller implementation.

### DIFF
--- a/c++/src/capnp/rpc-test.c++
+++ b/c++/src/capnp/rpc-test.c++
@@ -2691,6 +2691,52 @@ KJ_TEST("AdaptiveFlowController: minimum window is enforced") {
   KJ_EXPECT(estimatedWindow >= 64 * 1024, estimatedWindow);
 }
 
+KJ_TEST("AdaptiveFlowController: destroying with blocked senders fulfills them") {
+  kj::EventLoop loop;
+  kj::WaitScope waitScope(loop);
+  TestClock clock;
+  auto fc = RpcFlowController::newAdaptiveController(256 * 1024, clock);
+
+  // Send a 256KB message to fill most of the window. The ack never arrives (NEVER_DONE),
+  // simulating a dead follower.
+  auto msg1 = kj::heap<MockMessage>(256 * 1024);
+  fc->send(kj::mv(msg1), kj::NEVER_DONE).poll(waitScope);
+
+  // Second send blocks — window is full since the first ack never arrived.
+  auto msg2 = kj::heap<MockMessage>(256 * 1024);
+  auto blockedPromise = fc->send(kj::mv(msg2), kj::NEVER_DONE);
+  KJ_EXPECT(!blockedPromise.poll(waitScope));
+
+  // Destroy the flow controller while a sender is blocked.
+  fc = nullptr;
+
+  // The blocked promise should be fulfilled (not rejected).
+  blockedPromise.wait(waitScope);
+}
+
+KJ_TEST("WindowFlowController: destroying with blocked senders fulfills them") {
+  kj::EventLoop loop;
+  kj::WaitScope waitScope(loop);
+  auto fc = RpcFlowController::newFixedWindowController(256 * 1024);
+
+  // Send a 256KB message to fill most of the window. The ack never arrives,
+  // simulating a dead follower.
+  auto msg1 = kj::heap<MockMessage>(256 * 1024);
+  fc->send(kj::mv(msg1), kj::NEVER_DONE).poll(waitScope);
+
+  // Second send blocks — window is full since the first ack never arrived.
+  auto msg2 = kj::heap<MockMessage>(256 * 1024);
+  auto blockedPromise = fc->send(kj::mv(msg2), kj::NEVER_DONE);
+  KJ_EXPECT(!blockedPromise.poll(waitScope));
+
+  // Destroy the flow controller while a sender is blocked.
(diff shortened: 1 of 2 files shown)


shrima-cf:
@a-robinson @kentonv @justin-mp 
Tagging for review, for some reason, I am not able to do this through UI

a-robinson (commented):
Looks reasonable, but @kentonv will have a more intuitive sense of whether this is the best fix. If he's available to review it soon then that's ideal, otherwise I can take a closer look.

kentonv:
This isn't the right solution. The implication is that someone called flowController->send(), got a Promise back, and then destroyed the flowController while that Promise still exists. That's a violation of the KJ style guide which says that any Promise returned by a method of an object must either complete or be canceled before the object can be destroyed.

So we really need to examine the cases where this is happening and fix them to either cancel the promise or attach the flow controller to the promise.

a-robinson:
Makes sense, thanks.

Going by the stack trace on the ticket I tagged you on, it looks like it's downstream of `RpcClient` being destructed: https://github.com/capnproto/capnproto/blob/fb4e4ac9ff13ada088f40d75b640638435600044/c%2B%2B/src/capnp/rpc.c%2B%2B#L927-L938

But you're saying the issue goes even higher up, and that the promise chain that sent the RPC request and was awaiting its response needs to be canceled before anything can be done with the rpc client? It's surprising that that isn't already the case, but there is some funny business going on there with timeouts and activationState that might be able to cause some kind of issue like this I guess.

shrima-cf:
> This isn't the right solution. The implication is that someone called flowController->send(), got a Promise back, and then destroyed the flowController while that Promise still exists. That's a violation of the KJ style guide which says that any Promise returned by a method of an object must either complete or be canceled before the object can be destroyed.
> 
> So we really need to examine the cases where this is happening and fix them to either cancel the promise or attach the flow controller to the promise.

Is it because of the check here - https://github.com/capnproto/capnproto/blob/v2/c%2B%2B/src/capnp/rpc.c%2B%2B#L913

```
  void waitAllAckedInBackground(kj::Own<RpcFlowController> flowController) {
    // Calls `waitAllAcked()` on `flowController` as a background task. This is needed when an
    // `RpcFlowController` is being abandoned but we don't actually intend to cancel the calls
    // that were made on it.

    // If we're disconnected, there's no need to wait for anything, so we'll just drop the
    // flowController in that case.
    if (connection.is<Connected>()) {
      tasks.add(flowController->waitAllAcked().attach(kj::mv(flowController)));
    }
  }
```

kentonv:
Hmm, `waitAllAckedInBackground` looks like it is correctly attaching the flow controller to the promise, so it's impossible for the flow controller to be destroyed before the promise is either resolved or has been canceled.

But maybe the real problem is:
* One or more callers are still waiting on promises returned by send().
* waitAllAckedInBackground() is called, but the inner waitAllAcked() call is actually being canceled before it finishes waiting (someone is clearing the TaskSet).
* The attached `flowController` reference is the last reference to the flow controller, so this destroys the flow controller.
* The callers of send() see the resulting exception.

The bug here would be that there are callers who have called send() and then allowed the flow controller to be destroyed before send() finished.

This in turn is probably because someone is sending a request to the `RpcClient` and then dropping it before the request's returned promise resolves.

This bug must be somewhere in the RPC system, since `Client` objects normally automatically attach a reference to the underlying `ClientHook` to any RPC promise. That is, it shouldn't be possible for application code to make this mistake. But the RPC internals could make this mistake.

shrima-cf:
Is it because the flowPromise does not attach to the client? - https://github.com/capnproto/capnproto/blob/v2/c%2B%2B/src/capnp/rpc.c%2B%2B#L2667

kentonv:
@shrima-cf Yes I think that might be it. But attaching the client isn't quite right, as that would delay sending a release message to the other side in certain cases.

After closer inspection I think the right thing to do might just be to fulfill all the `blockedSends`.

For streaming sends, the promise resolving doesn't mean the message was delivered, it only means "now is a good time to send the next message". If we just fulfill, the caller will try to send the next message, and that will fail appropriately.

Not that it's _wrong_ to reject either... but I like fulfilling because the error the caller gets on the next send will be more specific. If we reject in the flow controller, I feel like we need to plumb through the root cause error rather than just stay "RpcFlowController destroyed".

kentonv (approved):

shrima-cf:
Thanks for the merge! I was just about to ping!

PR #2615, opened 2026-04-03
add Exception::clone() and remove exception copy constructor

mikea:
as discussed in https://github.com/capnproto/capnproto/pull/2613

it will take me a bit to align downstream changes.

### DIFF
--- a/c++/src/kj/exception.h
+++ b/c++/src/kj/exception.h
@@ -72,9 +72,13 @@ public:
 
   Exception(Type type, const char* file, int line, String description = nullptr) noexcept;
   Exception(Type type, String file, int line, String description = nullptr) noexcept;
-  Exception(const Exception& other) noexcept;
   Exception(Exception&& other) noexcept = default;
   KJ_NOINLINE ~Exception() noexcept;
+
+  Exception(const Exception& other) noexcept = delete;
+  // Use clone() instead.
+
+  Exception clone() const noexcept;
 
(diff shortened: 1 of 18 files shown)


c++/src/kj/exception.h
```diff
@@ -162,6 +163,9 @@ class Exception {
   bool isMovedAway() const { return storage == nullptr; }
   // When exception is moved away from, accessing any of its data will throw.
 
+protected:
+  Exception(const Exception& other) noexcept;
```
kentonv: Why protected and not deleted?

c++/src/kj/common.h
```diff
@@ -2161,6 +2161,14 @@ class Maybe {
     }
   }
 
+  Maybe<T> clone() {
```
kentonv: Is the plan to remove this commit in favor of #2616?

kentonv (commented):
Seems like it'd be easier to manage downstream changes if we added clone() first, then updated everyone, then removed the copy constructor, in separate PRs? But up to you.

c++/src/kj/common.h
mikea: correct, this is now rebased and cleaned-up.

c++/src/kj/exception.h
mikea: there was ExceptionImpl copy constructor, but nothing complained when I removed that, so done.

kentonv (approved):

PR #2616, opened 2026-04-03
Cloneable<T> and clone() method in basic structures

mikea:
- StringPtr, String
- ArrayPtr, Array
- Maybe

Supports only types with explicit clone() for now. I'll address primitive T in followups.

### DIFF
--- a/c++/src/kj/array.h
+++ b/c++/src/kj/array.h
@@ -272,6 +272,10 @@ public:
   // Syntax sugar for invoking asImpl(U*, const Array&).
   // Used to chain conversion calls rather than wrap with function.
 
+  auto clone() const requires Cloneable<T>;
+  // Deep-clone to a new heap array by cloning every element.
+  // Returns Array<decltype(t.clone())>
+
   inline bool hasNullDisposer() const {return disposer == &NullArrayDisposer::instance; }
   // Returns true if array uses NullArrayDisposer, intended for use with string literal
   // ConstStrings.
@@ -947,6 +951,21 @@ heapArray(Iterator begin, Iterator end) {
 template <typename T>
 inline Array<T> heapArray(std::initializer_list<T> init) {
   return heapArray<T>(init.begin(), init.end());
+}
+
+template <typename T>
+inline auto ArrayPtr<T>::clone() const requires Cloneable<T> {
+  using U = decltype(instance<const T&>().clone());
+  auto builder = heapArrayBuilder<U>(size());
+  for (auto& value: *this) {
+    builder.add(value.clone());
+  }
+  return builder.finish();
+}
+
+template <typename T>
+inline auto Array<T>::clone() const requires Cloneable<T> {
+  return asPtr().clone();
 }
 
 template <typename T, typename... Params>
--- a/c++/src/kj/common-test.c++
+++ b/c++/src/kj/common-test.c++
@@ -29,6 +29,16 @@
 namespace kj {
 namespace {
 
+struct ClonesToInt { int clone() const { return 123; } };
+struct ClonesToStringPtr { StringPtr clone() const { return "foo"; } };
+struct NoClone {};
+struct NonConstClone { int clone() { return 123; } };
+
+static_assert(Cloneable<ClonesToInt>);
+static_assert(Cloneable<ClonesToStringPtr>);
+static_assert(!Cloneable<NoClone>);
+static_assert(!Cloneable<NonConstClone>);
+
 KJ_ASSERT_CAN_MEMCPY(char);
 KJ_ASSERT_CAN_MEMCPY(byte);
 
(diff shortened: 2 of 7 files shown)


c++/src/kj/common.h
```diff
@@ -2124,6 +2124,15 @@ class Maybe {
     }
   }
 
+  Maybe<T> clone() const {
```
kentonv: * Can this be guarded with `requires` to check T has a clone() method, similar to what you did for `Maybe<T&>`? I think this is important so that other types can correctly check whether this type has a clone method.
* The return value should probably be something like `Maybe<decltype(t.clone())>`. Consider a type like `Maybe<StringPtr>` -- cloning it should return `Maybe<String>`.

c++/src/kj/common.h
```diff
@@ -2248,6 +2257,17 @@ class Maybe<T&> {
     }
   }
 
+  auto clone() const -> Maybe<T>
```
kentonv: I guess it's an interesting question whether cloning `Maybe<T&>` should do a shallow copy or a deep copy.

Note we should also add clone to Array and ArrayPtr.

But this implies cloning `ArrayPtr<Maybe<T&>>` produces `Array<Maybe<T>>`... but what if you really only wanted to copy the T pointers not the T values? Well, I guess then you have to use KJ_MAP yourself.

So I think I'm agreed with what you have here, but we'll have to see how it works out in practice.

mikea:
Alright, I decided to go full mile and implement clone() everywhere. It is important to make everything work together, hence I extended the scope.

Note that containers clone() definition is `auto` and doesn't have explicit `-> decltype`. It breaks because apperently compiler tries to compute return type even if concept does not match.

```
  auto clone() const -> Array<decltype(instance<const T&>().clone())> requires Cloneable<T>;
```

becomes `error: member reference base type 'const char' is not a structure or union`.

It is possible to fix it by defining

```
template <typename U = T>
auto clone() const -> Array<decltype(instance<const U&>().clone())>
    requires Cloneable<U>;
```

but this is way to ugly imo.

PTAL @kentonv

c++/src/kj/common.h
mikea: I think `T& -> T` cloning is consistent with `StringPtr -> String` and `ArrayPtr -> Array` cloning.

If one doesn't need deep cloning - copy constructor is still avaiable. And for more complex scenarious - KJ_MAP.

c++/src/kj/common-test.c++
```diff
@@ -29,6 +29,16 @@
 namespace kj {
 namespace {
 
+struct ClonesToInt { int clone() const { return 123; } };
+struct ClonesToStringPtr { StringPtr clone() const { return "foo"; } };
+struct NoClone {};
+struct NonConstClone { int clone() { return 123; } };
```
kentonv, line 35: I think we'll need non-const clone() to be allowed. Consider refcounted pointers or capnp `Client` objects that are rightly cloned by incrementing the refcount, but that's not const.

Note we have a TON of use cases for exactly this in the workers codebase. We could get rid of `mapAddRef` and replace it with `clone()`.

c++/src/kj/array.h
```diff
@@ -272,6 +272,10 @@ class Array {
   // Syntax sugar for invoking asImpl(U*, const Array&).
   // Used to chain conversion calls rather than wrap with function.
 
+  auto clone() const requires Cloneable<T>;
```
kentonv, line 275: Already mentioned elsewhere: I think e.g. `Array<int>`, or `Array<SomeTrivialStruct>`, really ought to be clonable. That means T either needs to be clonable, _or_ copy-constructable. The latter case can delegate to `heapArray(array)`, which is already optimized to use memcpy() for trivial types, etc.

kentonv (approved):
OK with merging as-is and iterating but I think we really do want to support the two things I commented on.

c++/src/kj/common.h
```diff
@@ -1249,6 +1249,10 @@ concept ConstructibleFrom = requires(U&& u) { T(kj::fwd<U>(u)); };
 
 }  // namespace _ (private)
 
+template <typename T>
+concept Cloneable = requires(const T& value) { value.clone(); };
```
harrishancock, line 1253: Bikeshed:
```cpp
template<typename T>
concept Cloneable = requires(const T& value) {
  { value.clone() } -> std::convertible_to<T>;
  // the result can be different from T, but must convert back
};
```

or

```cpp
template<typename T>
concept Cloneable = requires(const T& value) {
  value.clone();
  requires !kj::isSameType<decltype(value.clone()), void>();
  // must return any non-void result
};
```

harrishancock (approved):

mikea:
todo list for followup

- non-const clone()
- clone() in Rc, Arc, CapabilityClient
- Array/ArrayPtr clone support for primitive types and for types with copy constructor

PR #2617, opened 2026-04-03
Verify not reading hole-punched regions on file reads

zachlefevre:


kentonv (changes requested):
We can't make this change:
* It adds an additional syscall to every read. That is actually quite expensive! This quite possibly makes most reads twice as expensive, possibly even worse depending on what data structures lseek needs.
* The syscall in question has the side effect of seeking the file descriptor's current offset, which may be a breaking change for some apps.
* The warning being printed is not necessarily detecting erroneous behavior, or even very suspicious behavior. Holes are punched in order to implement the `zero()` method, whose job is to write zeros to the file. An app may have called `zero()` merely as an efficient way of zeroing out some bytes, having no idea that this punches holes. It may legitimately read those bytes later for any number of reasons.

Sorry, but this is not the right way to detect this sort of bug.

I think what you should do is write a wrapper around files in SRS (or maybe use the existing ShiftedFile wrapper since it's already wrapping everything relevant?) and have it track known holes in memory, and then compare reads against those holes. That will be much more efficient, more targeted, and doesn't impact other users of the KJ filesystem APIs.

PR #2618, opened 2026-04-06
TransformPromiseNode::getImpl micro optimization

mikea:
I was hunting down cases of missed optimization opportunities related to ~Exception, which are uncovered by making it inline.

This one stood out. If you take a look at v2 decompiled code: https://gist.github.com/mikea/1c5ef39c5c0d1b70a4ade15b70bd5cfe

there's ~Exception branch present.

This change is the smallest transformation of existing code I was able to find that eliminates the branch: https://gist.github.com/mikea/d7e4d0a1293c1d6fe1b8130048eb7310

It does this by strictly limiting Exception scope.

This change has some minor performance improvement for Promise_Fib10 benchmark (-0.2%) that uses this code path a lot. This optimization becomes even more important if you remove NOINLINE for ~kj::Exception.

https://gist.github.com/mikea/409bf964945f43d80b85e7cf47b37097

As well as landing this I'm curious if we can identify some code patterns/helpers that could make this more elegant?

### DIFF
--- a/c++/src/kj/async-inl.h
+++ b/c++/src/kj/async-inl.h
@@ -663,14 +663,27 @@ private:
     // Derive return type from DepT to reduce templating.
     typedef _::FixVoid<_::ReturnType<Func, _DepT>> T;
     typedef _::FixVoid<_DepT> DepT;
-    ExceptionOr<DepT> depResult;
-    getDepResult(depResult);
-    KJ_IF_SOME(depException, depResult.exception) {
-      output.as<T>() = handle<T>(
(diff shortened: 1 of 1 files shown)


harrishancock (approved):
I struggle to think of how to improve this or make it clearer, mostly because I can't understand why it makes the compiler produce better code.

Since it's a measurable improvement, though, seems like we should use it. I'd suggest adding some code comments explaining what the important bit is (controlling the scope of `depResult` I guess).

PR #2619, opened 2026-04-08
kj::isNoThrowMoveConstructible

mikea:
std::is_nothrow_move_constructible also checks for noexcept destructor which is never true for us. Our implementation considers only move constructor.

The plan is to use it inside ExceptionOr and other places for safety.

### DIFF
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -1066,6 +1066,13 @@ inline void dtor(T& location) {
   location.~T();
 }
 
+template <typename T>
+constexpr bool isNoThrowMoveConstructible() {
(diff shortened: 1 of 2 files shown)


c++/src/kj/common.h
```diff
@@ -1043,9 +1060,6 @@ class MappedIterable: private Mapping {
 // are compatible, and there are not segfaults trying to call clang operator new/delete from GCC or
 // vice versa.
```
harrishancock, line 1044: This comment should probably be moved along with `_::PlacementNew`.

harrishancock (approved):

c++/src/kj/common.h
mikea: I moved my definition below instead.

PR #2620, opened 2026-04-08
move constructor noexcept hygeine

mikea:
Bunch of noexcept specifiers driven by attempt to make ExceptionOr move constructor noexcept

### DIFF
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -1066,13 +1066,6 @@ inline void dtor(T& location) {
   location.~T();
 }
 
-template <typename T>
-constexpr bool isNoThrowMoveConstructible() {
-  // like std::is_nothrow_move_constructible but that ignores noexcept(false) destructors
-  if constexpr (isReference<T>()) return true;
-  else return noexcept(new ((void*)nullptr, _::PlacementNew()) T(instance<T&&>()));
-}
-
 // =======================================================================================
 // Maybe
 //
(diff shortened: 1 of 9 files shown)


fhanau:
If you haven't already used it: clang-tidy [performance-noexcept-move-constructor](https://clang.llvm.org/extra/clang-tidy/checks/performance/noexcept-move-constructor.html) can be used to enforce that noexcept is used here.

c++/src/kj/common.h
```diff
@@ -1066,11 +1066,11 @@ inline void dtor(T& location) {
   location.~T();
 }
 
-template <typename T>
+template <typename T, typename U = T>
 constexpr bool isNoThrowMoveConstructible() {
```
harrishancock: Non-blocking bikeshed: It occurs to me this could be a `NoThrowConstructibleFrom` concept, dove-tailing with this one: https://github.com/capnproto/capnproto/blob/v2/c%2B%2B/src/kj/common.h#L1253-L1255

I guess that wouldn't support `T`s that are reference types, though.

harrishancock (approved):

c++/src/kj/common.h
mikea: NP, done.

mikea:
> If you haven't already used it: clang-tidy [performance-noexcept-move-constructor](https://clang.llvm.org/extra/clang-tidy/checks/performance/noexcept-move-constructor.html) can be used to enforce that noexcept is used here.

yeah, I know but I wanted something stricter (it will be compiler-checked) and we don't run clang-tidy here.

PR #2621, opened 2026-04-13
non-const Cloneable<T>

mikea:
This removes const requirement from clone() method and adds corresponding const overrides to containers.

Adds clone() to Rc and Arc to demonstrate the feature (in Rc case)

### DIFF
--- a/c++/src/kj/array.h
+++ b/c++/src/kj/array.h
@@ -272,7 +272,8 @@ public:
   // Syntax sugar for invoking asImpl(U*, const Array&).
   // Used to chain conversion calls rather than wrap with function.
 
-  auto clone() const requires Cloneable<T>;
+  auto clone() requires Cloneable<T>;
+  auto clone() const requires Cloneable<const T>;
   // Deep-clone to a new heap array by cloning every element.
   // Returns Array<decltype(t.clone())>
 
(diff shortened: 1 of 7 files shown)


c++/src/kj/array.h
```diff
@@ -964,7 +965,22 @@ inline auto ArrayPtr<T>::clone() const requires Cloneable<T> {
...
 
 template <typename T>
-inline auto Array<T>::clone() const requires Cloneable<T> {
+inline auto ArrayPtr<T>::clone() const requires Cloneable<const T> {
+  using U = decltype(instance<const T&>().clone());
+  auto builder = heapArrayBuilder<U>(size());
+  for (const auto& value: *this) {
+    builder.add(value.clone());
+  }
+  return builder.finish();
```
harrishancock: I wonder if this could be `return KJ_MAP(value, *this) { return value.clone(); };`.

harrishancock (approved):
I suppose ForkedPromise could make good use of Cloneable, too: https://github.com/capnproto/capnproto/blob/d0e9605b288ac4d8b86ee3201a23c43c017a6b61/c%2B%2B/src/kj/async-inl.h#L749-L763

PR #2622, opened 2026-04-14
clone() for ForkedPromise

mikea:


harrishancock:
This wasn't what I meant by this comment over here, just to be clear: https://github.com/capnproto/capnproto/pull/2621#pullrequestreview-4104702567

I mean that ForkedPromise could support any `Cloneable<T>`, rather than any T that supports `.addRef()` or `->addRef()`.

But obviously this change makes sense on its own. :)

harrishancock (approved):

c++/src/kj/async.h
```diff
@@ -393,7 +393,10 @@ class ForkedPromise {
   inline ForkedPromise(decltype(nullptr)) {}
 
   Promise<T> addBranch();
-  // Add a new branch to the fork.  The branch is equivalent to the original promise.
+  // Add a new branch to the fork. The branch is equivalent to the original promise.
+
+  inline Promise<T> clone() { return addBranch(); }
```
kentonv, line 398: I don't think this is a valid definition of `clone()` since it is returning a different type.

You _could_ pretty easily make this return `ForkedPromise<T>` instead by constructing it on `hub->addRef()`.

kentonv (changes requested):

kentonv:
Hmm, I am sort of wondering if, instead of adding `clone()` methods to refcounted pointer types, we should instead recognize `addRef()` as a valid way to implement `copyOrClone()`, so it can be used when cloning arrays.

Pros:
* It may be confusing that calling `clone()` on a refcounted type directly is just incrementing the refcount rather than actually cloning. If we continue to use `addRef()` in those scenarios, it's a lot clearer.

Cons:
* This approach implies that `Maybe<T>` now needs an `addRef()` method that forwards to the inner `addRef()`, instead of using `clone()` for that. More complexity in Maybe, but not too terrible?

mikea:
FWIW Rust uses clone https://doc.rust-lang.org/std/rc/struct.Rc.html#impl-Clone-for-Rc%3CT,+A%3E which was an inspiration for our Rc/Arc. I don't mind using clone consistently everywhere.

kentonv:
Ah if Rust does it this way already then I guess we should just use `clone()`.

Point still stands that `ForkedPromise<T>::clone()` should return `ForkedPromise<T>`.

PR #2623, opened 2026-04-14
Copyable<T> and its support in clone() of ArrayPtr, Array

mikea:


c++/src/kj/common.h
```diff
@@ -2142,25 +2160,23 @@ class Maybe {
     }
   }
 
-  auto clone() requires Cloneable<T> {
-    // Clones the value if it is not none.
-    // Returns Maybe<decltype(t.clone())>
-    using U = decltype(instance<T&>().clone());
+  auto clone() requires (Cloneable<T> || Copyable<T>) {
```
kentonv: For Maybe I don't think we really need to support clone() for copyable types since the Maybe itself is already copyable if T is copyable. It's a bit different from Array<T> where Array is never copyable even if T is.

mikea: reverted this part.

c++/src/kj/common.h
```diff
@@ -1267,6 +1267,24 @@ template <typename T>
 concept Cloneable = requires(T& value) { value.clone(); };
 // Concept: T has a `clone()` member callable on a `T&`.
 
+template <typename T>
+concept Copyable = _::ConstructibleFrom<T, T&>;
+// Concept: T has a copy constructor callable on a `T&`.
```
harrishancock, line 1275: Since Rust has similarly-named traits (`Clone` and `Copy`), it might be helpful for people coming from Rust to contrast these concepts with those traits in doc comments.

My own understanding is:

- KJ `Cloneable` is like Rust `Clone`, but must support mutable cloning, because C++ can't safely hide interior mutability like Rust can.
- KJ `Copyable` is NOT like Rust `Copy`. Rust `Copy` is the same as C++'s concept of "trivially copyable" (e.g. memcpy-able). KJ `Copyable` just tells us if there's a copy constructor, for which there is no Rust equivalent.

Would you agree? Maybe an AI can generate a more succinct comparison?

harrishancock (approved):

c++/src/kj/common.h
mikea: done.

kentonv (approved):

PR #2624, opened 2026-04-16
static disposer for kj::Rc

mikea:
Makes no sense to go through virtual calls.
If this works well I'll convert Arc too.

### DIFF
--- a/c++/src/kj/refcount.h
+++ b/c++/src/kj/refcount.h
@@ -172,7 +172,7 @@ public:
   inline Rc(Rc&& other) noexcept = default;
 
   template <typename U, typename = EnableIf<canConvert<U*, T*>()>>
-  inline Rc(Rc<U>&& other) noexcept : own(kj::mv(other.own)) { }
+  inline Rc(Rc<U>&& other) noexcept : own(other.own.template disown<Rc<U>>()) { }
 
   kj::Own<T> toOwn() {
     // Convert Rc<T> to Own<T>.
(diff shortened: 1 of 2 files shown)


c++/src/kj/refcount.h
```diff
@@ -215,13 +217,22 @@ class Rc {
...
-  Rc(T* t) : own(t, *t) { }
-  Rc(Own<T>&& t) : own(kj::mv(t)) { }
+  Rc(T* t) : own(t) { }
+  Rc(Own<T, Rc>&& t) : own(kj::mv(t)) { }
 
-  Own<T> own;
+  Own<T, Rc> own;
+  // Rc is its own disposer.
+
+  inline static void dispose(T* t) {
```
kentonv, line 226: I guess having a separate dispose() implementation for every T rather than a shared one that takes `kj::Refcounted` potentially avoids virtual dispatch on the destructor? Otherwise I'd suggest having a single disposer for less codegen but avoiding the virtual destructor call is nice.

kentonv (approved):

c++/src/kj/refcount.h
mikea: the biggest issue with virtual call is that compiler does not see refcount decrements. I think it is important to give compiler a chance to optimize them out? (at least to try)

kentonv: That's not what I meant. I'm talking about the _destructor_ being virtual.

If we had one dispose function `dispose(Refcounted* t)`, decrementing the refcount doesn't require any virtual calls, only when the refcount reaches zero and the destructor is called, then it would be calling `Refcounted`'s destructor which is virtual. Whereas by using a `T`-specific instantiation as you do here, then if `T` is marked `final` (often the case), no virtual call is needed for the destructor itself.

PR #2626, opened 2026-04-17
ArrayPtr::split()

mikea:
Non-allocating version returning an iterator.

Supercedes https://github.com/capnproto/capnproto/pull/2050

### DIFF
--- a/c++/src/kj/common.h
+++ b/c++/src/kj/common.h
@@ -2452,6 +2452,16 @@ struct Mapper<Maybe<T>> {
 template <typename T>
 class Array;
 
+namespace _ {  // private
+class SplitIteratorEnd;
+
+template <typename T>
+class SplitIterator;
+
+template <typename T>
+class SplitIterable;
+}  // namespace _ (private)
+
 template <typename T>
 class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
   // A pointer to an array.  Includes a size.  Like any pointer, it doesn't own the target data,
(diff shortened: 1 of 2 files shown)


c++/src/kj/common.h
```diff
@@ -2734,6 +2748,65 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
...
+    nextSegment();
+  }
+
+  inline ArrayPtr<T> operator*() { return array.first(end); }
+
+  inline SplitIterator& operator++() {
+    if (end == array.size()) {
+      end = array.size() + 1;
+    } else {
+      array = array.slice(end + 1);
```
harrishancock, line 2768: I guess if we `++` an end iterator, this will throw, assuming `KJ_IREQUIRE` is enabled?

c++/src/kj/common.h
```diff
@@ -2734,6 +2748,65 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
...
+    } else {
+      end = array.size();
+    }
+  }
+};
+
+template <typename T>
+class SplitIterable {
+public:
+  inline SplitIterable(ArrayPtr<T> array, T&& delim) : array(array), delim(kj::mv(delim)) {}
```
harrishancock, line 2799: Non-blocking nit: inconsistent `delim` type with `SplitIterator` (`T` vs `T&&`). Not sure it matters.

harrishancock (approved):

c++/src/kj/common.h
mikea: yes. It will provide the same level of protection as arrayptr does. We also have KJ_IREQUIRE enabled in production.

c++/src/kj/common.h
mikea: it does make a copy in `begin()`. Do you think I shouldn't support calling begin more than once?

harrishancock: Ah good point, I missed that subtlety.

I suppose if you wanted to avoid that copy, you could take a reference to the value in the SplitIterable. That might be more expensive than just copying the value, though, for the types of values we'll be iterating over.

PR #2630, opened 2026-04-21
Replace c++/LICENSE.txt symlink with real file

pgerell:
## Problem
`c++/LICENSE.txt` is a symlink to `../LICENSE`. This breaks tools that extract only the `c++/` subtree via `strip_prefix`, as the symlink target resolves outside the extracted directory.

## Solution
- Replace the symlink with a real copy of the file.


kentonv:
Oh man I realize now looking at it the first line desperately needs an update, heh. I'll take care of that at some point.

kentonv:
Oh crap, I just realized you targeted the release branch.

That won't accomplish anything, as release 1.4.0 is already shipped, and the next release will be cut from `master`.

You need to target `master` (for 1.x releases) or `v2` (for the development branch)...

PR #2631, opened 2026-04-21
Fix bare includes in ez-rpc.h

pgerell:
Bare includes in `ez-rpc.h` break builds with sandboxed include paths. See #1942.


kentonv:
I think only `message.h` needs changing since `rpc.h` is in the same library?

This will need to target the master branch in any case.

pgerell:
@kentonv Now targets `master`

kentonv:
Thanks

PR #2633, opened 2026-04-23
kj: tolerate ERROR_INVALID_FUNCTION from GetFileInformationByHandleEx on Wine

Sjors:
I tested this patch as part of cross-compile-then-run-in-wine PR: https://github.com/bitcoin-core/libmultiprocess/pull/272.

However, I didn't check the (LLM generated) reasoning in the commit message.

### DIFF
--- a/c++/src/kj/filesystem-disk-win32.c++
+++ b/c++/src/kj/filesystem-disk-win32.c++
@@ -380,6 +380,11 @@ public:
       case ERROR_CALL_NOT_IMPLEMENTED:
         // Probably WINE.
       	break;
+      case ERROR_INVALID_FUNCTION:
+        // Also WINE: NtQueryInformationFile returns STATUS_NOT_IMPLEMENTED for
+        // unsupported FileInfoClass values, which RtlNtStatusToDosError maps to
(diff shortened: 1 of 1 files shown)


kentonv:
The fix seems reasonable but I am slightly offended by the LLM's assumptions about me! (lol)

> That arm appears to never have actually fired under Wine; it has just been an unverified guess.

I certainly did actually test this under WINE! But it was over eight years ago. Perhaps WINE has changed since then.

Would you mind adjusting the commit message to say the error seems to have changed?

Sjors:
@kentonv sorry about that :-)

It was probably my fault: I asked it figure out which version of Wine introduced a change. It couldn't find the answer and so it blamed you.

I'll adjust the test and look into the CI failure.

kentonv:
Don't worry about the CI failure. FreeBSD 15 has a bug (in FreeBSD, not Cap'n Proto) which has made it broken fro some time.

And I think windows-2019 isn't running because they don't have action runners for it anymore.

Sjors:
Ok in that case: I pushed a modified commit message.

kentonv (approved):

kentonv:
thanks!

PR #2634, opened 2026-04-23
explicit refcounted PipelineHook and MembranePolicy

mikea:
supercedes https://github.com/capnproto/capnproto/pull/2404 and https://github.com/capnproto/capnproto/pull/1990

These both now have to be done at the same time because of membrainePipeline/reverseMembrainePipeline.

The issues with too many increments in #1990 is resolved by adding _non-virtual_ kj::Rc<MembranePolicy> MembranePolicy::addRef() so that MembraneCapTableReader can keep its naked references. I think this not worth it and extra increment is not a real performance issue, but it is not worse than now. Unfortunately I don't think http over capnp benchmark exercises this codepath, so I can't measure the impact. Let me know if you have ideas for MembraneCall bench.

### DIFF
--- a/c++/src/capnp/membrane.h
+++ b/c++/src/capnp/membrane.h
@@ -49,16 +49,19 @@
 
 #include "capability.h"
 #include <kj/map.h>
+#include <kj/refcount.h>
 
 CAPNP_BEGIN_HEADER
 
 namespace capnp {
 
-class MembranePolicy {
+class MembranePolicy: public kj::Refcounted {
   // Applications may implement this interface to define a membrane policy, which allows some
   // calls crossing the membrane to be blocked or redirected.
 
 public:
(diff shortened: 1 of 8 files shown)


c++/src/capnp/membrane.h
```diff
@@ -180,14 +172,14 @@ class MembranePolicy {
   // importInternal() and exportExternal() will always be references to *this.
 
   virtual Capability::Client importInternal(
-      Capability::Client internal, MembranePolicy& exportPolicy, MembranePolicy& importPolicy);
+      Capability::Client internal, kj::Rc<MembranePolicy> exportPolicy, kj::Rc<MembranePolicy> importPolicy);
```
mikea, line 175: I opted into converting this to kj::Rc too. I can revert it if you think the extra increment here is critical.

kentonv (approved):

harrishancock (approved):

PR #2637, opened 2026-04-24
Fix the trouble with Tribbles.

kentonv:
TribbleRaceBlocker had a TODO describing a case where it blocks promise resolution. It described this as TODO(perf). But it's actually a correctness issue when combined with CapabilityServerSet, which needs all promises to resolve in order to unwrap a local capability.

Two tests are added, one to cover the bug that is fixed, and another than verifies the general TribbleRaceBlocker behvaior, because I noticed that removing it entirely didn't break any tests.

This was written entirely by GPT 5.5, which can also take credit for figuring out that this bug was manifesting in an obscure case in production in the Cloudflare Workers Runtime.

The first commit was GPT's first attempt to fix the specific bug, which specifically solved the CapabilityServerSet issue but not the broader performance issue. I then pushed it to actually fully solve the issue, which produced the second commit.

**I HAVE NOT REVIEWED THIS YET.** I honestly have not yet been able to fully remember WTF is going on here. This is THE most complicated part of Cap'n Proto. When I get a chance (definitely before asking anyone else to review) I'm going to spend an hour or two paging this all back into my head and decide if this fix is actually the fix I want...

### DIFF
--- a/c++/src/capnp/rpc-test.c++
+++ b/c++/src/capnp/rpc-test.c++
@@ -1356,6 +1356,81 @@ KJ_TEST("embargos block CapabilityServerSet") {
   KJ_EXPECT(unwrappedAt >= 3, unwrappedAt);
 }
 
+KJ_TEST("TribbleRaceBlocker blocks CapabilityServerSet unwrap") {
+  TestContext context;
+
+  capnp::CapabilityServerSet<test::TestCallOrder> capSet;
+
+  auto client = context.connect().getTestMoreStuffRequest().send().getCap();
+
+  TestCallOrderImpl* ptr;
+  auto ownCap = kj::heap<TestCallOrderImpl>();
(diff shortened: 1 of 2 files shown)


kentonv:
I've determined that neither commit actually fixes the bug in prod lol.

Guess I'll have to use my old human brain on this one.

ocdtrekkie:
I have no point, I just want to say that I enjoy this commit and branch name existing immensely.

HenkPoley:
There is a possibility for `CapabilityServerSet::getLocalServer(receiverAnswer-type promise)` to return `null` (meaning: not local) for local service objects when it is referenced by remotes. Even though the interface specifies it waits to resolve promises.

Tiny chance that this is your remaining bug.

kentonv:
Proper, working fix in #2640.

PR #2640, opened 2026-04-29
Actually fix the trouble with tribbles.

kentonv:
This is a second attempt at #2637.

This time:
* I spent all day trying to understand what's actually going on.
* I wrote the solution by hand.
* I kept the tests but also prompted GPT 5.5 to improve them further.
* I also verified this fixes the real issue seen in prod, which I think is still not covered properly by the tests.

To understand the problem and solution, see the updated comment in rpc.c++ around 2101. Then spend all day trying to understand WTF it is talking about. Then give up and just trust it's fixed now.

### DIFF
--- a/c++/src/capnp/rpc.c++
+++ b/c++/src/capnp/rpc.c++
@@ -941,13 +941,23 @@ private:
       return connectionState.get() == &expected;
     }
 
-    virtual void promiseResolvedToThis(bool receivedCall) {}
+    virtual bool promiseResolvedToThis(bool receivedCall) { return false; }
     // Called to indicate that an RPC promise was resolved (via a `Resolve` message we received)
     // to point to this client (i.e. the `CapDescriptor` for the resolution pointed to another
     // import table entry). `receivedCall` is true if the resolving promise had received calls
     // in the past, which had been pipelined. Some RpcClient derivatives need to know when this
     // happens in order to arrange to use embargoes where appropriate. Others don't care, so can
     // use the default implementation.
+    //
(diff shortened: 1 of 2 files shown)


c++/src/capnp/rpc.c++
```diff
@@ -941,13 +941,23 @@ private:
...
     // in the past, which had been pipelined. Some RpcClient derivatives need to know when this
     // happens in order to arrange to use embargoes where appropriate. Others don't care, so can
     // use the default implementation.
+    //
+    // The method returns true if the caller needs to apply an embargo to the resolution.
+    // Normally, when a promised import or answer resolves to another import on the same
+    // connection, no embargo is necessary since it points along the same path. However, if a
+    // promised import/answer resolves to what the sending side thinks is a capability hosted on
+    // the receiving end, but it just so happens that said capability itself loops back to point
+    // to the sender, then we have a Tribble 4-way race an an embargo is still needed. In this case
```
harrishancock, line 957:
```suggestion
    // to the sender, then we have a Tribble 4-way race and an embargo is still needed. In this case
```

harrishancock (approved):

c++/src/capnp/rpc.c++
kentonv: Oh well.

kentonv:
I hope @ocdtrekkie enjoyed the branch name once again.

ocdtrekkie:
I, in fact, guessed this might be the branch name based on your comment. I am delighted.

PR #2642, opened 2026-04-30
Add RevokerMembrane utility class

jclee:
RevokerMembrane has been a useful utility in workerd and elsewhere.

### DIFF
--- a/c++/src/capnp/membrane-test.c++
+++ b/c++/src/capnp/membrane-test.c++
@@ -401,6 +401,40 @@ KJ_TEST("revoke membrane") {
       thing.passThroughRequest().send().wait(env.waitScope));
 }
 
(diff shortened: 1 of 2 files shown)


jclee:
Hmm... So this was basically importing the [workerd implementation](https://github.com/cloudflare/workerd/blob/main/src/workerd/util/completion-membrane.h#L43-L44).

But having worked more on the problem that I thought needed this, I found out it also needed an allowFdPassthrough() override, so maybe it is not as universally useful as I thought, as-is.

(And similarly for [workerd's CompletionMembrane](https://github.com/cloudflare/workerd/blob/main/src/workerd/util/completion-membrane.h#L10-L11), which I'd initially omitted because it has grown a reject() method that I'm not sure is as broadly useful.)

Closing for now.

PR #2644, opened 2026-05-01
cleanup CAPNP_DEBUG_TYPES ifdef

mikea:
I don't think we care about gcc/clang abi compatibility anymore?


kentonv (approved):
I think this was fixed in GCC 15 in any case (at least judging from the linked bugs).

mikea:
sorry, clicked wrong button, reverting

mikea:
> sorry, clicked wrong button, reverting

no, all is well. I overreacted - the build failure was fixed by https://github.com/capnproto/capnproto/pull/2643

PR #2647, opened 2026-05-06
use kj::ArrayPtr<const char> for statusText

mikea:
Rust's strings are not 0-allocated so passing statusText requires extra allocation.

@kentonv lmk if you think this is ok and I'll create downstream PRs.


kentonv:
I think we need to think about this more holistically and decide what happens to the headers.

In C++ the status text string is often owned by the `HttpHeaders` since it comes out of the same buffer. Same for the URL in the request.

Though I guess here you are specifically modifying the server-side `HttpService::Response` callback API, where maybe that's less true. Here, the string is often just a static string from the app. Far enough.

Still, we talked before about changing the whole model here to make these things be pass-by-move instead of reference. What happened to that plan?

PR #2648, opened 2026-05-07
Add OneOf::clone()

harrishancock:
Mirrors the existing `kj::Array<T>::clone()` (and `kj::Maybe<T>::clone()`) APIs: returns a deep-cloned OneOf with each variant cloned (via clone()) or copied (via copy ctor), depending on which trait the variant satisfies.

Per-variant requires `Cloneable<T> || Copyable<T>`. Unlike `Maybe<T>::clone()` which accepts only `Cloneable<T>` — because a `Maybe<T>` for copyable T is itself copyable, making clone() redundant — OneOf can be non-copyable (when at least one variant is non-copyable) even when other variants are copyable. Supporting `Copyable` per variant lets the mixed-trait case (e.g. `OneOf<int, kj::String>::clone()`) work: int copied, String cloned.

Both `&` and `const&` overloads are provided so types whose clone() is non-const (e.g. `kj::Rc<T>::clone()` mutates a refcount) can be cloned in non-const contexts.

Result type is `OneOf<decltype(_::copyOrClone(v))...>`, so type-changing clones work: `OneOf<ArrayPtr<T>>::clone()` returns `OneOf<Array<T>>`.

### DIFF
--- a/c++/src/kj/one-of-test.c++
+++ b/c++/src/kj/one-of-test.c++
@@ -20,6 +20,7 @@
 // THE SOFTWARE.
 
 #include "one-of.h"
+#include "refcount.h"
 #include "string.h"
 #include <kj/compat/gtest.h>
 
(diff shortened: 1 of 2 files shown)


harrishancock:
Looks like the Linux-clang (18, bazel-dbg) build flaked:
```
//src/capnp:compiler/evolution-test                                     TIMEOUT in 30.0s
```

kentonv (approved):

harrishancock:
Force-push was just to add a couple test cases: clone of default-initialized and moved-from values.

PR #2649, opened 2026-05-11
Support Bazel 9 in V1

spinningthelock:
This PR adds Bazel 9 support to V1 branch
* Removes WORKSPACE file as its now unused
* Updates all the third party packages to latest version
* Replaced boringssl with openssl to match the CMake file


kentonv:
@fhanau can you look at this?

fhanau:
I can confirm that the build works with these changes for me locally (on macOS). Let's keep using boringssl though, this matches what we do on the v2 branch.
One other concern I have is that this will break the Bazel Windows build – maintaining that required several changes when upgrading to new Bazel versions as can be seen on the v2 branch. That being said I'm not sure if that configuration is being actively used (CI on the master branch doesn't test it, and the cmake MSVC Windows build should still work) so maybe that's fine?

spinningthelock:
@fhanau I updated PR to use boringssl. Updated it to latest version in BCR.

fhanau:
> @fhanau I updated PR to use boringssl. Updated it to latest version in BCR.

Looks like CI is failing with updated boringssl since this branch still uses C++14. To preserve compatibility, can you keep the old boringssl version or update to whatever the last version supporting C++14 is?

spinningthelock:
yeah. I will try to find a one that supports C++14 and Bazel 9. I think I ran into the same issue when I was creating the initial PR. Then decided to using openssl since thats what cmake version of code does. That was faster than finding one that is C++14 and Bazel 9.

spinningthelock:
@fhanau solution was to not update boringssl. Builds and tests on my local linux machine

spinningthelock:
For my erudition, why is there a deviation between the cmake and bazel deps?

fhanau (approved):

PR #2650, opened 2026-05-12
Auto-roll capnp-cpp MODULE.bazel version in release.sh

pgerell:
Bump capnp-cpp module version in c++/MODULE.bazel from 1.1.0 to 1.5-dev to match the in-progress mainline version.
Updated release.sh to automate this.
#2649 did remove the version number but AFAICT the Bazel Central Registry would need a patch that adds it if not already set.
The version number could be omitted if `publish-to-bcr` is added to the release flow.
https://github.com/bazel-contrib/publish-to-bcr

### DIFF
--- a/c++/MODULE.bazel
+++ b/c++/MODULE.bazel
@@ -1,6 +1,6 @@
 module(
     name = "capnp-cpp",
-    version = "1.1.0",
+    version = "1.5.0",
 )
 
 bazel_dep(name = "platforms", version = "0.0.11")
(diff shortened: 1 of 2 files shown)


kentonv:
Oh I see, this is a version number that needs to be updated with every release.

Perhaps, then, release.sh should be updated to automatically roll this number when appropriate, like it does all the other places where versions show up?

pgerell:
> Oh I see, this is a version number that needs to be updated with every release.
> 
> Perhaps, then, release.sh should be updated to automatically roll this number when appropriate, like it does all the other places where versions show up?

Good call. I've added `c++/MODULE.bazel` to `update_version()` in `release.sh` so it rolls automatically alongside the other files. Set the version to 1.5-dev to be in sync with the rest.

pgerell:
@kentonv Is this what you were asking for?

PR #2654, opened 2026-05-15
RcStringPtr and RcArrayPtr

mikea:


c++/src/kj/array.h
```diff
@@ -22,13 +22,69 @@
...
+
+  inline ArrayPtr<T> asPtr() { return ptr; }
+  inline ArrayPtr<const T> asPtr() const { return ptr; }
+
+  inline RcArrayPtr clone() { return *this; }
+  inline RcArrayPtr clone() const requires (isConst<T>()) { return *this; }
+
+private:
+  ArrayPtr<T> ptr;
+  mutable Rc<RefcountedWrapper<Array<T>>> storage;
```
kentonv, line 85: This shouldn't be mutable and `clone()` shouldn't support `const` since it isn't thread-safe.

c++/src/kj/array.h
```diff
@@ -22,13 +22,69 @@
...
 
 KJ_BEGIN_HEADER
 
 namespace kj {
 
+// =======================================================================================
+// RcArrayPtr
+
+template <typename T>
+class RcArrayPtr: public DisallowConstCopyIfNotConst<T> {
```
kentonv, line 37: Bikeshed: I'd probably call it `RcArray` or `ArrayRc` but not `RcArrayPtr` -- in KJ the `Ptr` suffix means "unowned" but that's not the case here.

c++/src/kj/string.h
```diff
@@ -325,6 +325,53 @@ class String {
...
+    return *this;
+  }
+
+  inline StringPtr asPtr() const { return ptr; }
+
+  inline RcStringPtr clone() const { return *this; }
+
+private:
+  StringPtr ptr;
+  mutable Rc<RefcountedWrapper<String>> storage;
```
kentonv, line 367: Might as well use ConstString instead of String since we're not allowing the content to be modified.

kentonv:
Oh I guess #2655 is expected to supersede this huh?

PR #2655, opened 2026-05-15
Rc<T> without inheritance

mikea:
Wraps into rc counted object and provides same interface without extra allocations.

The only thing that is missing when T does not extend Refcounted is `addRefToThis` which most Ts do not nead.
Makes possible to write `kj::Rc<String>` directly.

### DIFF
--- a/c++/src/kj/memory.h
+++ b/c++/src/kj/memory.h
@@ -85,6 +85,8 @@ class AtomicRefcounted;
 class Refcounted;
 
 namespace _ {  // private
+template <typename T>
+concept IsPolymorphic = _kj_internal_isPolymorphic((T*)nullptr);
 
 template <typename T, typename U>
 concept DerivedFrom = requires(const T* t) { static_cast<const U*>(t); };
--- a/c++/src/kj/memory.h
+++ b/c++/src/kj/memory.h
@@ -85,8 +85,6 @@ class AtomicRefcounted;
 class Refcounted;
 
 namespace _ {  // private
-template <typename T>
-concept IsPolymorphic = _kj_internal_isPolymorphic((T*)nullptr);
 
 template <typename T, typename U>
 concept DerivedFrom = requires(const T* t) { static_cast<const U*>(t); };
--- a/c++/src/kj/refcount.h
+++ b/c++/src/kj/refcount.h
@@ -37,6 +37,16 @@ namespace kj {
 template<typename T>
 class Rc;
 
+template <typename T, typename... Params>
+Rc<T> rc(Params&&... params);
+
+namespace _ {  // private
+
+template <typename T> class RcWrapper;
+template <typename T> class RcOwnWrapper;
+
+}  // namespace _ (private)
+
 class Refcounted: private Disposer {
   // Subclass this to create a class that contains a reference count. Then, use
   // `kj::refcounted<T>()` to allocate a new refcounted pointer.
@@ -102,6 +112,9 @@ private:
 
   template <typename T>
   friend class Rc;
+
+  template <typename T> friend class _::RcWrapper;
+  template <typename T> friend class _::RcOwnWrapper;
 };
 
 template <typename T, typename... Params>
@@ -110,14 +123,6 @@ inline Own<T> refcounted(Params&&... params) {
   // initial reference to the object.  More references can be created with `kj::addRef()`.
 
   return Refcounted::addRefInternal(new T(kj::fwd<Params>(params)...));
-}
-
-template <typename T, typename... Params>
-inline Rc<T> rc(Params&&... params) {
-  // Allocate a new refcounted instance of T, passing `params` to its constructor.
-  // Returns smart pointer that can be used to manage references.
-
-  return Refcounted::addRcRefInternal(new T(kj::fwd<Params>(params)...));
 }
 
 template <typename T>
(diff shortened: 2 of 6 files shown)


c++/src/kj/refcount.h
```diff
@@ -37,6 +37,22 @@ namespace kj {
...
+namespace _ {  // private
+
+template <typename T>
+class RcWrapper;
+
+template <typename T>
+concept Complete = requires { sizeof(T); };
+
+template <typename T>
+concept RcStoresDirect = (!Complete<T>) || DerivedFrom<T, Refcounted>;
```
kentonv: Seems worrying that if T is incomplete, this concept could match it incorrectly. This could lead to some weird subtle bugs?

c++/src/kj/refcount.h
```diff
@@ -226,47 +257,107 @@ class Rc {
...
 };
 
-template <typename T>
-class RefcountedWrapper: public Refcounted {
-  // Adds refcounting as a wrapper around an existing type, allowing you to construct references
-  // with type Own<T> that appears to point directly to the underlying object.
+namespace _ {  // private
 
+template <typename T>
+class RcWrapper: public Refcounted {
```
kentonv: This seems basically the same as `RefcountedWrapper`. Why not reuse it?

c++/src/kj/refcount.h
mikea: Done. There is KJ_DECLARE_DERIVED_FROM_REFCOUNTED now.

c++/src/kj/refcount.h
mikea: It became slightly different thing and I made it independent. The goal rather would be to remove RefcountedWrapper in favor of direct Rc<T> usage.

mikea:
PTAL, I did a major overhaul and added support for polymorphic T. kj::Rc<kj::AsyncInputStream> is now possible.

c++/src/kj/refcount.h
```diff
@@ -226,6 +291,202 @@ class Rc {
...
+protected:
+  T wrapped;
+  // If you experience "field has incomplete type" error, this means you're trying to use
+  // Rc<T> with forward-declared T. 
+  // Add KJ_DECLARE_DERIVED_FROM_REFCOUNTED/KJ_DECLARE_NOT_DERIVED_FROM_REFCOUNTED
+  // after forward declaration.
+};
+
+template <typename T>
+requires IsPolymorphic<T>
```
kentonv: It seems like you're using `IsPolymorphic` to decide whether to wrap the type in `Own<T>` but I don't see why those things should be related. The `requires (!IsPolymorphic<T>)` specialization above seems like it'll work just fine with polymorphic types (and would be more efficient).

The only case where you need this specialization is if you are trying to wrap a type starting from an `Own<T>`.

c++/src/kj/refcount.h
```diff
@@ -226,6 +291,202 @@ class Rc {
...
+  template <typename U, typename... Params>
+  requires (!_::DerivedFromRefcounted<U>)
+  friend Rc<U> rc(Params&&... params);
+
+  template <typename>
+  friend class Rc;
+};
+
+template<typename T>
+class Rc<Own<T>> {
```
kentonv: Shouldn't `Rc<T>` support initializing from `Own<T>`?

`Rc<Own<T>>` seems to leak details of how the `T` was initially constructed into the type. You end up having to pass `Rc<Own<T>>` all over which feels weird?

You could probably avoid this by storing the `T*` separately from `Own<RcWapper>`, in which case the latter can just be `Own<Refcounted>` instead. But that's an extra pointer per `Rc`. You could go further and not wrap `Own` at all, and just have:

```
T* ptr;
Refcounted* refcounted;
```

And then make sure to properly handle disposal in the `Rc` class, though that's obviously more work...

c++/src/kj/refcount.h
mikea: now that I have implemented constructor from kj::Own, `kj::Rc<kj::Own<T>>` is never needed.

c++/src/kj/refcount.h
mikea: well, I have managed to implement everything in 3 pointers and keep only one allocation for non-poly . `kj::rc<T>`.

mikea: scratch that, only 2 pointers are needed.

c++/src/kj/refcount.h
```diff
@@ -142,48 +147,116 @@ Rc<T> Refcounted::addRcRefInternal(T* object) {
...
+    // Concepts require T to be fully defined, but Rc<T> is often used with forward-declared T.
+    // This function is declared as template to help msvc in polymorphic base class case.
+    static_assert(!kj::canConvert<T*, Refcounted*>());
+    auto wrapper = new _::RcWrapper<U>(kj::mv(t));
+    refcounted = wrapper;
+    ptr = wrapper->getWrappedPtr();
+  }
+
+  inline Rc(Own<T> t) noexcept {
+    static_assert(!kj::canConvert<T*, Refcounted*>());
```
ObsidianMinor: Super nit-picky but maybe we should elide the kj namespace more consistently?

ObsidianMinor (approved):
I like it! It's what I always hoped kj::Rc would be

c++/src/kj/refcount.h
mikea: done.

c++/src/kj/refcount.h
```diff
@@ -142,92 +147,195 @@ Rc<T> Refcounted::addRcRefInternal(T* object) {
...
 class Rc {
-  // Smart pointer for reference counted objects.
+  // Rc<T> is a smart pointer providing reference counting capabilities for all kinds of Ts.
+  //
+  // The primary way to obtain new `Rc<T>` instance is ot use `kj::rc<T>(...)` function that will
+  // allocates new T on the heap. Depending on T nature:
+  // - T extends Refcounted. In such case T's `refcount` field is used for counting. 1 allocation.
+  // - T does not extend Refcounted and is not polymorphic. `kj::rc` will over-allocate 
+  //   `RcWrapper<T>` to provide a `refcount`. 1 allocation.
+  // - T extends Refcounted and is polymorphic. `RcOwnWrapper<T>` is used. 2 allocations.
```
kentonv: I think you meant "does not extend Refcounted" here? If it extends Refcounted then it's always polymorphic but that was covered in the first point.

But I don't see why it matters if T is polymorphic. It seems to me you only need `RcOwnWrapper` in the case that we're initializing from `Own<T>`. `RcWrapper<T>` should work fine for both polymorphic and non-polymorphic types, shouldn't it?

c++/src/kj/refcount.h
```diff
@@ -142,92 +147,195 @@ Rc<T> Refcounted::addRcRefInternal(T* object) {
...
   //
-  // There are only three ways to obtain new Rc instances:
-  // - use kj::rc<T>(...) function to create new T.
-  // - use kj::Rc::addRef() and the existing Rc instance.
+  // Rc<T> can also be constructed from:
+  // - kj::Own<T> for all types of T. Always allocates.
```
kentonv: I think if `T` extends `Refcounted` then you don't need to allocate for this case? You can take over the existing `Refcounted` base class.

c++/src/kj/refcount.h
```diff
@@ -142,92 +147,195 @@ Rc<T> Refcounted::addRcRefInternal(T* object) {
...
+  // - T for non-polymorphic non-`Refcounted` Ts with move constructor. Always allocates.
```
kentonv: Again I don't think that polymorphism matters here, except that of course many polymorphic types do not support move-construction.

c++/src/kj/refcount.h
```diff
@@ -142,92 +147,195 @@ Rc<T> Refcounted::addRcRefInternal(T* object) {
...
-  Rc(T* t) : own(t, *t) { }
-  Rc(Own<T>&& t) : own(kj::mv(t)) { }
+  Rc(Refcounted *wrapper, T *ptr) : refcounted(wrapper), ptr(ptr) {}
+  void dispose() {
+    if (ptr == nullptr) return;
+    auto ptrCopy = ptr;
+    auto wrapperCopy = refcounted;
+    refcounted = nullptr;
+    ptr = nullptr;
+    wrapperCopy->dispose(const_cast<RemoveConst<T>*>(ptrCopy));
```
kentonv: This technically passes the wrong pointer in the case that `T` is a multiply-inherited base class of the actual type. That's why `Own::dispose()` needs to do a `dynamic_cast<void*>`.

However, we know that `Refcounted`'s disposer doesn't care what pointer it is passed, so you could simply pass null here.

mikea: done

c++/src/kj/refcount.h
mikea: done as well

c++/src/kj/refcount.h
mikea: the `kj::Own` might have things attached to it, so it is a bit tricky. I think usually you don't have `kj::Own<T extends Refcounted>` - my long term plan is to fully convert to `kj::Rc`, so it should be fine?

c++/src/kj/refcount.h
mikea: I guess I got separate polymorphic case early on this PR and was carrying it over rather mechanically. You're right we don't need anymore.

c++/src/kj/refcount.h
kentonv: Fair point. You could assert that the pointer and disposer match to protect against that case (or even fall back to allocating at runtime if the pointers don't match).

c++/src/kj/memory.h
```diff
@@ -85,6 +85,8 @@ class AtomicRefcounted;
 class Refcounted;
 
 namespace _ {  // private
+template <typename T>
+concept IsPolymorphic = _kj_internal_isPolymorphic((T*)nullptr);
```
kentonv: No longer used?

kentonv (approved):

c++/src/kj/refcount.h
mikea: I wrote the code with fallback first but:

a) it requires declaring too many friends
b) as I said before it optimizes a code path that I want to get rid of. The only use-case for taking over Own<T> I have is for `Own<Async*Stream>`, so voted for a simpler code. Will be happy to revisit this later if this indeed becomes used for Refcounted.

PR #2656, opened 2026-05-20
NFC: remove duplicate getShortDisplayName

RReverser:
Just something I noticed as part of another WIP but was out of scope of that change.

`getShortDisplayName` and `getUnqualifiedName` helpers duplicated each other. Keeping the `getUnqualifiedName` as it has clearer semantics and reusing it in couple more places.

### DIFF
--- a/c++/src/capnp/compiler/capnp.c++
+++ b/c++/src/capnp/compiler/capnp.c++
@@ -1741,7 +1741,7 @@ public:
 
       case schema::Node::STRUCT:
         if (pos == nameParts.end()) {
-          return kj::str("'", schema.getShortDisplayName(), "' cannot be evaluated.");
+          return kj::str("'", schema.getUnqualifiedName(), "' cannot be evaluated.");
         }
 
         // Use the struct's default value.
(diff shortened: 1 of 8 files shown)


kentonv:
Hmmm, in theory "display name" is for display only and so could change or include decorative text, while "unqualified name" has a specific meaning. If we're going to remove one in favor of the other I'd remove `getShortDisplayName()` since it is less-well-defined.

RReverser:
> Hmmm, in theory "display name" is for display only and so could change or include decorative text, while "unqualified name" has a specific meaning.

I thought about that but it seems moot at this point - if they diverge, _then_ it's worth adding another method.

> If we're going to remove one in favor of the other I'd remove `getShortDisplayName()` since it is less-well-defined.

Yeah fair, I'd rather do that too but wasn't sure it would be accepted given that `getShortDisplayName` is the older / more established one.

kentonv:
You'll need to update workerd, there are some call sites.

Does this actually matter for something? Maybe it's not really worth it to bother?

RReverser:
Mostly just a minor annoyance/cleanup, functionally it doesn't matter as per title.

PR #2657, opened 2026-05-20
operator* for kj::Rc and kj::Arc

mikea:
I still have to reach out for it when converting the code.


kentonv (approved):
Yep I've found myself doing `*someRc.get()` a lot, nice to be able to simplify that.

PR #2658, opened 2026-05-21
clang tidy

mikea:
It is driven by compile_flags.txt, so we will keep it up-to-date as a side-benefit.

Motivation is to eventually start using https://clang.llvm.org/extra/clang-tidy/QueryBasedCustomChecks.html (clang-22 I believe)

### DIFF
--- a/.github/workflows/quick-test.yml
+++ b/.github/workflows/quick-test.yml
@@ -18,7 +18,7 @@ env:
   INSTALL_DEPS: |
     install_deps() {
       local clang=$1
-      export DEBIAN_FRONTEND=noninteractive && sudo apt-get update && sudo apt-get install --no-upgrade --no-install-recommends -y clang-$clang libc++-$clang-dev libc++abi-$clang-dev libclang-rt-$clang-dev lld-$clang llvm-$clang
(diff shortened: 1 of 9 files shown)


fhanau (approved):
I assume that the bazel build step is being run as part of the build so that dependencies are fetched and clang-tidy can find the headers it needs – we should add a comment explicitly mentioning this before someone tries to remove the bazel build step, which could easily be misinterpreted as being redundant.

Non-blocking: clang-tidy contains no checks by default so the -*, statement should not be needed

Non-blocking: The `bazel-bin/external/brotli+/_virtual_includes/brotli_inc/` include path does not exist for me whereas bazel-out/aarch64-dbg/bin/external/brotli+/_virtual_includes/brotli_inc/ does, check which path is available for you locally

mikea:
> Non-blocking: clang-tidy contains no checks by default so the -*, statement should not be needed

This wasn't my experience. I had to disable it because lots of checks were running.

> Non-blocking: The `bazel-bin/external/brotli+/_virtual_includes/brotli_inc/` include path does not exist for me whereas bazel-out/aarch64-dbg/bin/external/brotli+/_virtual_includes/brotli_inc/ does, check which path is available for you locally

The one that I have put exists both for me & for CI. Feel free to add extra paths for your architecture, afaik extra ones do not hurt.

PR #2659, opened 2026-05-21
[release-1.3.0] backport HTTP body-size overflow fixes (CVE-2026-32239 + CVE-2026-32240)

dkgkdfg65:
backports `2744b3c012b4` onto release-1.3.0 — the http body-size overflow hardening isn't on this branch yet. one commit covers both CVE-ids:

- CVE-2026-32239: the content-length parser in `HttpInputStreamImpl::getEntityBody` runs the header through strtoull without a leading-digit check, so `Content-Length: -1` silently becomes 18446744073709551615 — a smuggling primitive through a buggy proxy.
- CVE-2026-32240: the chunked path truncates a chunk size >= 2^64 to a 64-bit value instead of rejecting it.

i pulled both parsers out and ran them either side of the commit: pre-fix `Content-Length: -1` is accepted as 2^64-1 and a 17-digit `f` chunk size truncates to 2^64-1; post-fix both are rejected. clean cherry-pick.

upstream: https://github.com/capnproto/capnproto/commit/2744b3c012b4
CVE: https://nvd.nist.gov/vuln/detail/CVE-2026-32239 / https://nvd.nist.gov/vuln/detail/CVE-2026-32240

didn't run the full test suite — change is the parser guards only. if you'd rather pull it in yourselves that's fine, just flagging the gap.

### DIFF
--- a/c++/src/kj/compat/http-test.c++
+++ b/c++/src/kj/compat/http-test.c++
@@ -4038,6 +4038,70 @@ KJ_TEST("HttpServer invalid method") {
   KJ_EXPECT(expectedResponse == response, expectedResponse, response);
 }
 
+KJ_TEST("HttpServer rejects negative Content-Length") {
+  KJ_HTTP_TEST_SETUP_IO;
+  kj::TimerImpl timer(kj::origin<kj::TimePoint>());
+  auto pipe = KJ_HTTP_TEST_CREATE_2PIPE;
+
+  HttpHeaderTable table;
+  BrokenHttpService service;
+  HttpServer server(timer, table, service, {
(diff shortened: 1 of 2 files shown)


kentonv:
Why do this? Just use 1.4.0. The diff between 1.3.0 and 1.4.0 is trivial -- there is no new feature development happening on the 1.x branch.

dkgkdfg65:
fair point — i'd assumed 1.x was a stable branch with periodic patch releases, but if 1.x is essentially 'use the latest tag' with no per-minor backporting then a 1.3-only PR doesn't really fit the workflow. anyone on 1.3.0 should just bump to 1.4.0 and pick up the same fix.

closing this. thanks for the quick reply.

PR #2661, opened 2026-05-21
clang 22, clang-tidy 22

mikea:


fhanau (approved):
Ubuntu 26.04 has been released recently, supports LLVM 22 and should be available in GitHub Actions within days, maybe that's the cleaner approach in the long run.
I stamped this anyway in case you have a short-term use for it

PR #2662, opened 2026-05-22
Clamp rpc Exception type enum.

dom96:
It's possible to construct an exception type which doesn't match the values allowed by the `rpc::Exception::Type` enum, then send it over the wire. This can lead to a out of bounds array access when stringifying the exception type.

This PR fixes this and includes tests which reproduce it.

### Test Plan

Verified that this passes:

```
bazel test --config=asan //src/kj:exception-test //src/capnp:rpc-test
```

Verified that we get an ASAN trigger without the fix in our test:

```
[ TEST ] exception-test.c++:647: KJ_STRINGIFY(Exception::Type) handles out-of-range values
=================================================================
==15625==ERROR: AddressSanitizer: global-buffer-overflow on address 0x0001023c9ff8 at pc 0x0001023467bc bp 0x00016db08210 sp 0x00016db08208
    #0 0x0001023467b8 in kj::operator*(kj::_::Stringifier, kj::Exception::Type)+0x54 (exception-test:arm64+0x1000527b8)
    #1 0x00010232af74 in kj::_::(anonymous namespace)::TestCase647::run()+0xa70 (exception-test:arm64+0x100036f74)
...
```

### DIFF
--- a/c++/src/capnp/rpc.c++
+++ b/c++/src/capnp/rpc.c++
@@ -123,8 +123,18 @@ kj::Exception toException(const rpc::Exception::Reader& exception) {
     }
   }();
 
-  kj::Exception result(static_cast<kj::Exception::Type>(exception.getType()),
-      "(remote)", 0, kj::mv(reason));
+  // Cap'n Proto enums are encoded as raw UInt16 on the wire, so a peer can send any value 0-65535.
+  // We must clamp to valid kj::Exception::Type values to avoid undefined behavior (e.g. out-of-
+  // bounds array access when stringifying the exception type).
+  kj::Exception::Type type = kj::Exception::Type::FAILED;
+  switch (exception.getType()) {
+    case rpc::Exception::Type::FAILED:        type = kj::Exception::Type::FAILED; break;
+    case rpc::Exception::Type::OVERLOADED:    type = kj::Exception::Type::OVERLOADED; break;
+    case rpc::Exception::Type::DISCONNECTED:  type = kj::Exception::Type::DISCONNECTED; break;
+    case rpc::Exception::Type::UNIMPLEMENTED: type = kj::Exception::Type::UNIMPLEMENTED; break;
+  }
+
+  kj::Exception result(type, "(remote)", 0, kj::mv(reason));
   if (exception.hasTrace()) {
     result.setRemoteTrace(kj::str(exception.getTrace()));
   }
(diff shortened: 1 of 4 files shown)


c++/src/capnp/rpc.c++
```diff
@@ -123,8 +123,19 @@ kj::Exception toException(const rpc::Exception::Reader& exception) {
...
+  // Cap'n Proto enums are encoded as raw UInt16 on the wire, so a peer can send any value 0-65535.
+  // We must clamp to valid kj::Exception::Type values to avoid undefined behavior (e.g. out-of-
+  // bounds array access when stringifying the exception type).
+  kj::Exception::Type type;
+  switch (exception.getType()) {
+    case rpc::Exception::Type::FAILED:        type = kj::Exception::Type::FAILED; break;
+    case rpc::Exception::Type::OVERLOADED:    type = kj::Exception::Type::OVERLOADED; break;
+    case rpc::Exception::Type::DISCONNECTED:  type = kj::Exception::Type::DISCONNECTED; break;
+    case rpc::Exception::Type::UNIMPLEMENTED: type = kj::Exception::Type::UNIMPLEMENTED; break;
+    default:                                  type = kj::Exception::Type::FAILED; break;
```
harrishancock:
```suggestion
```

in conjunction with initializing `type` to `FAILED` above.

c++/src/capnp/rpc.c++
```diff
@@ -123,8 +123,19 @@ kj::Exception toException(const rpc::Exception::Reader& exception) {
     }
   }();
 
-  kj::Exception result(static_cast<kj::Exception::Type>(exception.getType()),
-      "(remote)", 0, kj::mv(reason));
+  // Cap'n Proto enums are encoded as raw UInt16 on the wire, so a peer can send any value 0-65535.
+  // We must clamp to valid kj::Exception::Type values to avoid undefined behavior (e.g. out-of-
+  // bounds array access when stringifying the exception type).
+  kj::Exception::Type type;
```
harrishancock:
```suggestion
  // Clamp unrecognized values to FAILED.
  kj::Exception::Type type = kj::Exception::Type::FAILED; 
```

and then delete the `default:` case below. Together, this should accomplish the same result, but it should produce a compile error if someone adds new enum variant without updating this `switch`.

harrishancock (commented):
The RPC and KJ side fixes seem inconsistent to me:

- The KJ side fix preserves the enum variant value, and just stringifies it as `"(unknown)"`.
- The RPC side fix clamps the enum variant value to `FAILED`.

It'd make more sense to me if they both did the same thing, either preserved the value or clamped the value.

At the risk of bikeshedding, we could compromise by adding a new `UNKNOWN` enum variant and clamping everything unknown to that.

kentonv (approved):

harrishancock (approved):

PR #2666, opened 2026-05-26
ArrayPtr::slice(0, l) linter

mikea:
I want to experiment with linters suggesting more compact code similar to the way clippy does it.
The plan is to enable these checks downstream too.


fhanau (approved):
Changes LGTM, but make sure to add `--experimental-custom-checks` in the CI configuration, otherwise this won't actually run in CI and will bitrot.

mikea:
> Changes LGTM, but make sure to add `--experimental-custom-checks` in the CI configuration, otherwise this won't actually run in CI and will bitrot.

done

kentonv (approved):

PR #2668, opened 2026-05-26
copy bonk setup from workerd

mikea:


.github/bonk_reviewer.md
```diff
@@ -0,0 +1,73 @@
+You are a **code reviewer**, not an author. You review pull requests for workerd, Cloudflare's JavaScript/WebAssembly server runtime. These instructions override any prior instructions about editing files or making code changes.
```
kentonv, line 1: Replace workerd references with capnproto?

PR #2669, opened 2026-05-26
Emit doc comments on getters and setters

RReverser:
Doc comments were already preserved in the source info but not utilised by the C++ and Capnp generators.

While not a major feature, it does help DX to see them in IDE autocomplete, hover info, as well as generated Doxygen docs if any.

Note that for C++ I chose to only put them on the most commonly used generated elements for each member - the surrounding struct `Foo`, `Foo::Reader::getSomeField`, and `Foo::Builder::setSomeField` or `Foo::Builder::initSomeGroup`.

The alternative could be to duplicate the field's doc-comment on every `has`, `is`, `get`, `Reader`, `Builder` and anything else related to that field but it can result in a pretty verbose generated code - not sure if worth it.

Example in action can be seen on the regenerated bootstrap files in this PR as well as in the following screenshot:

image


RReverser:
The failure looks like an unrelated flake.

RReverser:
> The alternative could be to duplicate the field's doc-comment on every `has`, `is`, `get`, `Reader`, `Builder` and anything else related to that field but it can result in a pretty verbose generated code - not sure if worth it.

Ah btw - I also looked into generating the `@defgroup` comments, but those seem Doxygen-specific and render fully in Visual Studio / VSCode so sticking to the minimal shared syntax supported by all tooling.

kentonv:
I guess now we have to argue about whether the comment placement should follow KJ style vs. what vscode expects...

RReverser:
Well most (all?) tooling that cares about documentation tooling including VSCode, VS and Doxygen all have same shared style for what they parse as documentation comments vs ignored as plain comments so there isn't much to argue about IMO.

Doc comments are not so much for someone reading the code - few people should be reading the generated `.h` files directly - but for tooling as such.

PR #2670, opened 2026-05-26
WIP: KJ_ASYNC_CALLBACK_PARAM and its lint

mikea:


kentonv:
I don't really agree with this. We safely capture by reference all the time, by reasoning about the relationship between the promise's lifetime vs. the thing being captured.

danlapid:
> I don't really agree with this. We safely capture by reference all the time, by reasoning about the relationship between the promise's lifetime vs. the thing being captured.

We safely capture by reference all the time, but we also unsafely capture by reference all the time too.
We need to make our codebases more resilient to human errors, even at the cost of limiting the kind of code we can write.
It's just reality that we haven't been able to be good enough for C++'s "everything is allowed and you just have to not write bugs".

kentonv:
> We safely capture by reference all the time, but we also unsafely capture by reference all the time too.
> We need to make our codebases more resilient to human errors, even at the cost of limiting the kind of code we can write.
> It's just reality that we haven't been able to be good enough for C++'s "everything is allowed and you just have to not write bugs".

Sure, I agree with all that, but simply banning unowned references doesn't work. Even Rust doesn't do that -- it has lifetimes. We need to reason about lifetimes in C++ too.

PR #2671, opened 2026-05-26
http: add support for informational (1xx) responses

isker:
Ref: https://github.com/cloudflare/workerd/discussions/761

In HTTP, informational (1xx) responses are intermediate responses that typically do not terminate an HTTP response stream. That is to say, an HTTP response may present zero or more informational responses before presenting a terminal response with status code >= 200. (I say "typically" in that a 101 Switching Protocols response, as in WebSocket handshakes, does terminate the HTTP aspect of the response, though the stream does live on.)

So, informational responses change the "arity" of HTTP responses, from 1 to n, requiring API support from HTTP implementations. Because informational responses are niche, many implementations get away without exposing them to callers, either on the sending side by just not allowing them to be sent, or even on the receiving side by silently consuming them and continuing onward. Until now, KJ has been in this state.

Unfortunately, the standardization of 103 Early Hints as a mechanism to optimize asset loading in dynamic web sites means that informational responses are not as niche as they used to be. Browser support for consuming preload Link headers in early hints responses is widespread, so web servers implemented in KJ can benefit from being able to send early hints responses.

So:
- Add a new `HttpService::Response::sendInformational`, which allows sending any informational response without concluding the response.
- This method has a default no-op implementation, both because to do otherwise would be a breaking API change, and because I spotted many implementations in tests or in workerd that ought not have to care.
- Add support for "reading through" 1xx responses to HttpClient (except 101 Switching Protocols - see commentary). I believe that until now, any response streams containing informational responses were not correctly handled by it, in that HttpClient would give callers the first informational response, with no further access to subsequent responses.
- I have not added support for exposing those "read-through" informational responses to the caller: I could not find an API that preserved compatibility while also not feeling mediocre and bolted on. I am also not sure that support is necessary: in a library like KJ, sending informational responses is a plausible application-level concern, but I'm not sure receiving them is.  Given KJ's current scope, informational responses are useless trivia that applications can't do anything with. And if they do want that, they can use the lower-level HttpInputStream.
- On that note, set body length to 0 when reading an informational response entity, like in HttpInputStream.

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -1635,6 +1635,29 @@ public:
     KJ_UNREACHABLE;
   }
 
+  kj::Promise<HttpHeaders::ResponseOrProtocolError> readResponseHeadersSkippingInformational() {
+    // Like readResponseHeaders(), but automatically skips any informational
+    // responses. These are 1xx responses except for 101 Switching Protocols,
+    // which has the unique behavior of terminating HTTP/1 on the connection.
+    // That response needs to be surfaced to callers in order for them to
+    // properly react to this change in connection state.
+    while (true) {
+      auto headersOrError = co_await readResponseHeaders();
+      KJ_SWITCH_ONEOF(headersOrError) {
+        KJ_CASE_ONEOF(response, HttpHeaders::Response) {
+          if (response.statusCode >= 100 && response.statusCode < 200 && response.statusCode != 101) {
+            finishRead();
+            continue;
+          }
+          co_return kj::mv(headersOrError);
+        }
+        KJ_CASE_ONEOF(protocolError, HttpHeaders::ProtocolError) {
+          co_return kj::mv(headersOrError);
+        }
+      }
+    }
+  }
+
   inline const HttpHeaders& getHeaders() const { return headers; }
 
   Promise<size_t> tryRead(void* buffer, size_t minBytes, size_t maxBytes) {
@@ -2268,7 +2291,7 @@ kj::Own<kj::AsyncInputStream> HttpInputStreamImpl::getEntityBody(
       return kj::heap<HttpNullEntityReader>(*this, length);
     } else if (isConnectRequest && statusCode >= 200 && statusCode < 300) {
       KJ_FAIL_ASSERT("a CONNECT response with a 2xx status does not have an entity body to get");
-    } else if (statusCode == 204 || statusCode == 304) {
+    } else if ((statusCode >= 100 && statusCode < 200) || statusCode == 204 || statusCode == 304) {
       // No body.
       return kj::heap<HttpNullEntityReader>(*this, uint64_t(0));
     }
@@ -5592,7 +5615,7 @@ public:
 
     auto id = ++counter;
 
-    auto responsePromise = httpInput.readResponseHeaders().then(
+    auto responsePromise = httpInput.readResponseHeadersSkippingInformational().then(
         [this,method,id](HttpHeaders::ResponseOrProtocolError&& responseOrProtocolError)
             -> HttpClient::Response {
       KJ_SWITCH_ONEOF(responseOrProtocolError) {
@@ -7581,6 +7604,11 @@ kj::Own<HttpService> newHttpService(HttpClient& client) {
 
 // =======================================================================================
 
+kj::Promise<void> HttpService::Response::sendInformational(
+    uint statusCode, kj::StringPtr statusText, const HttpHeaders& headers) {
+  return kj::READY_NOW;
+}
+
 kj::Promise<void> HttpService::Response::sendError(
     uint statusCode, kj::StringPtr statusText, const HttpHeaders& headers) {
   auto stream = send(statusCode, statusText, headers, statusText.size());
@@ -8081,6 +8109,15 @@ private:
         co_return BREAK_LOOP_CONN_ERR;
       }
     }
+  }
+
+  kj::Promise<void> sendInformational(
+      uint statusCode, kj::StringPtr statusText, const HttpHeaders& headers) override {
+    KJ_REQUIRE(statusCode >= 100 && statusCode < 200, "can't call sendInformational() with a non-1xx statusCode");
+    KJ_REQUIRE(currentMethod != kj::none, "can't call sendInformational() after calling send()");
+    httpOutput.writeHeaders(headers.serializeResponse(statusCode, statusText));
+    httpOutput.finishBody();
+    return httpOutput.flush();
   }
 
   kj::Own<kj::AsyncOutputStream> send(
--- a/c++/src/kj/compat/http-test.c++
+++ b/c++/src/kj/compat/http-test.c++
@@ -1703,6 +1703,40 @@ KJ_TEST("HttpInputStream responses") {
   KJ_EXPECT(!input->awaitNextMessage().wait(waitScope));
 }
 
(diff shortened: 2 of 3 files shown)


c++/src/kj/compat/http.c++
```diff
@@ -7581,6 +7601,11 @@ kj::Own<HttpService> newHttpService(HttpClient& client) {
 
 // =======================================================================================
 
+kj::Promise<void> HttpService::Response::sendInformational(
+    uint statusCode, kj::StringPtr statusText, const HttpHeaders& headers) {
+  return kj::READY_NOW;
+}
```
isker, line 7610: As per the PR description:
> This method has a default no-op implementation, both because to do otherwise would be a breaking API change, and because I spotted many implementations in tests or in workerd that ought not have to care.

I am not sure what the appetite or tolerance is for breaking changes in the KJ APIs. If they are acceptable, maybe the rest of what I said is wrong, and it'd be best for every implementation to have to make a choice about this.

isker (commented):
AI disclosure: I prompted my way to an initial solution with Claude, then iterated pretty hard. I am absolutely not fluent in C++, but I do believe I have a working understanding of the different client codepaths involved in the HTTP implementation now. For example, I think it's right that only one of them (`HttpClient::request`) "reads through" informational responses; that's flat out wrong for WebSockets and I think it's also wrong for the low-level HttpInputStream.

The test coverage is very prompted. I did verify all the added tests fail without the rest of the changes.

isker:
I don't think the test failure is related.

c++/src/kj/compat/http.c++
isker: I think another alternative might be to not add a new method at all, instead adding support for informational responses in `send`, and changing the semantics of calling `send` multiple times under these circumstances. The return value in such a case would always be a HttpDiscardingEntityWriter, which is a bit strange, and the overall complexity of `send` would get materially higher, but maybe it's a better set of tradeoffs.

c++/src/kj/compat/http.c++
```diff
@@ -1635,6 +1635,26 @@ public:
...
   }
 
+  kj::Promise<HttpHeaders::ResponseOrProtocolError> readResponseHeadersSkipping1xx() {
+    // Like readResponseHeaders(), but automatically skips any 1xx informational responses.
+    while (true) {
+      auto headersOrError = co_await readResponseHeaders();
+      KJ_SWITCH_ONEOF(headersOrError) {
+        KJ_CASE_ONEOF(response, HttpHeaders::Response) {
+          if (response.statusCode < 200) {
+            // 1xx informational response: no body, release the gate and try again.
```
kentonv: If this was a 101 Switching Protocols response, doesn't ignoring it lead to HTTP desync, request smuggling, etc.?

isker: Hmm. This made me think about security issues that could exist here in general.

First, an answer to a question you did not explicitly ask. If a malicious server sends us an unprompted 101 response (compliant servers ["MUST NOT" send one without an Upgrade header in the request](https://datatracker.ietf.org/doc/html/rfc9110#field.upgrade)), I can see how this "skipping 1xx" behavior leads to a kind of denial of service, in the same way that a server just never responding with any final response is a denial of service. And in the non-malicious case, it's true that this API (`HttpClient::request`) is not suitable for making upgrade requests, but I think that's true of many HTTP clients. So I don't think this is a problem. 

To your actual question, I'm trying to understand what the desync/~~request~~ response smuggling scenario would look like. The malicious server responds with a 101 that says this is a WebSocket connection now, a proxy in front of the server believes it, and the client (us) ignores the 101 and continues to wait for an HTTP response. Then the server sends a normal HTTP/1 response through the connection, prompting... what divergent behavior between the client and the proxy?

My understanding of smuggling/desync attacks is that they involve attacking resources that inherently lie in a request-oriented direction, in attempts to convince a proxy to make supplemental unauthorized HTTP request(s) pointed somewhere else other than the initial intended server resource (basic smuggling) or to cue up a partial request on the server that poisons the state of the connection for subsequent requests (desync). These things are possible because of some combination of these factors:
- Every hop in the request-oriented direction is expecting to receive bytes from the TCP socket that indicate the start of an HTTP request, and they just disagree about how to delimit the received bytes into one or more requests.
- HTTP requests carry information in the request line or headers that address a resource on the server.
- Requests from multiple clients eventually end up traveling over a single TCP connection (most simply from the final proxy to the server), enabling desync from one client to attack another client.

But here there's only one client, only one thing that's going to accept an HTTP response. There's nothing else for the server to make the proxy attack.

To reframe this a bit, take the proxies and the smuggling/desync vulnerabilities away and think about what those vulnerabilities are supposed to enable: unintended or unauthorized access by a client to a real resource on a server, using otherwise conformant HTTP. For this 101 response handling to produce a vulnerability in the same category, you'd have to be able to open a TCP connection to some thing (a proxy? the client?) on the network path and get unauthorized access to some resource on it by sending it an HTTP response, i.e. nonconformant HTTP. If that's possible, it's a serious parsing vulnerability that exists independently of any smuggling/desync vulnerability.

Let me know if you think I'm missing something. I am certainly not an expert in HTTP security.

kentonv: Imagine an HTTP proxy built on KJ HTTP.

An attacker sends a request through the proxy with `Upgrade: websocket`. The request is proxied through unmodified, forwarded on to a server. The server response with 101 switching protocols. The proxy -- running your code -- ignores this line, and then expects the actual HTTP response to come next. But what actually comes next is the WebSocket body. Perhaps the attacker has some control over its content, e.g. maybe the backend server is some multi-tenant app hosting platform and the attacker has arranged for the WebSocket to connect to an app that the attacker controls.

This is where the problem really starts. The proxy now thinks the connection to the server is done, and it can now start a new HTTP request -- perhaps on behalf of an entirely different client -- to the same server, reusing the connection. But the server thinks it is now speaking WebSocket protocol, and perhaps the attacker has some control over the content that is exchanged. Now the attacker gets to receive requests that came from some other user, perhaps containing credentials, etc., and gets to return responses to those which may attack the victim user.

isker: I see. I was only thinking of this KJ HTTP client as the ultimate client, not as a part of a proxy. Indeed, here's what Envoy (a proxy we use that I am fairly familiar with) has to say on this topic:
https://github.com/envoyproxy/envoy/blob/4b8be8a46833c21b2a8e3d03e6994ac65f33004f/source/common/http/http1/codec_impl.cc#L1575-L1584

So 101 should just be an exception to this behavior.

isker: The latest push should fix this.

kentonv: FWIW we estimate KJ HTTP is used to proxy ~5% of all HTTP requests on the public internet. :) (By virtue of being used in Cloudflare Workers.)

This feature will require some careful review and thinking, not just about security but whether this is the API we want to go with. I could imagine instead that we might want to add explicit support for particular 1xx codes rather than leaving it up to the application to interpret them. E.g. "100 continue" feels like a protocol-level feature that needs some explicit API structure to support.

Unfortunately I'm a bit overloaded for at least the next month, and then on vacation for a while. I do think we should support 1xx codes but I'm not sure when I'll be able to look at this, sorry.

isker: Sounds good. I'll ping later this summer.

isker: Here I am again, but I see @jasnell opened #2749.

PR #2672, opened 2026-05-27
Add setBeforeFireHook

erikcorry:
The user can register a function that is called
on every round of the event loop.


kentonv:
For the record: We talked about this and came up with some better ideas for the problem this was meant to solve.

If we do want to go down this route I think a more complete async context mechanism is what we want, possibly including enter/exit hooks. But we've tried to build that before and found it surprisingly expensive, so it might take some careful design.

Issue #907, opened 2019-11-07
Inline structs/strong type alias


vlovich:
Tracking discussion in https://groups.google.com/forum/#!topic/capnproto/KFPRyvmc33c, would be nice if it were possible to define inlineable structs that behaved identically to a group but provided a strong name that could be reused multiple times.

Considerations:
* Do we want to support only "sealed" types or should we support a general mechanism to provide named groups?
* What is the complexity this adds to cap'n'proto?

vlovich:
For the first consideration the question is does an inlineable struct have to have all fields as a single contiguous range or is it possible to add new fields in the discontiguous manner that's appropriate at the call site where it's used. So for example if we start with:

```
struct MyVectorf inlineable {
   x :Float32;
   y :Float32;
}

struct SomeStruct {
  vec @[0..1] :MyVectorf;
  someValue @2 :UInt64;
}
```

Are we then able to add another member in a backwards compatible way for the inlineable struct?

```
struct MyVectorf inlineable {
   x :Float32;
   y :Float32;
   z :Float32;
}

struct SomeStruct {
  vec @[0..1, 3] :MyVectorf;
  someValue @2 :UInt64;
  someOtherValue @4 :UInt64;
}
```

And finally:
```
struct MyVectorf inlineable {
   x :Float32;
   y :Float32;
   z :Float32;
   w: Float32;
   timestamp: UInt64;
   metadata: Text;
}

struct SomeStruct {
  vec @[0..1, 3, 5..7] :MyVectorf;
  someValue @2 :UInt64;
  someOtherValue @4 :UInt64;
}
```

Obviously this example is an abuse of this feature for this as you'd probably be better off having Vector2f/Vector3f/Quaternionf data types but it illustrates a question about whether you can specify a mix of ranges/explicit field numbers for the inlined struct that's specific to the instantiation.

Semi-related is whether inlineable structs can reference any other type (including inlineable structs). At first glance that seems reasonable but then raises questions of how you assign indices. 

Additionally this brings up a question about notation. Do we want field numbers on the inlineable struct that indicates how they map to the inlined range? Do we do it implicitly based upon the field order? Do we want to have it be explicit based on the field name? e.g. `myVec @[x@0, y@1]`.

It might make a lot of sense to cut off complex use-cases in favor of keeping the design simple & providing a path to solve the most useful applications even if more complex ones (e.g. multiple levels of nested inlined structs) would be impossible to solve.

kentonv:
[...]

PR #2675, opened 2026-05-28
Emit using alias information

RReverser:
In our codegen we want to preserve usings for improved DX through specific alias names in the generated documentation and IDE hover info.

In order to achieve that, this PR preserves usings in the CodeGeneratorRequest in a backward-compatible manner.

First, it preserves using declarations themselves as Nodes - which should be backward-compatible in capnp terms, as codegens should already be ready to handle (eg ignore) unknown node types. If you think this can still break real-world codegens, I've considered adding a flag like `--emit-usings` but I feel that would be an overkill for what is officially a supported schema upgrade.

Second, it adds a `usingId` to the `Type` structs - this way, any codegens that don't recognise it or don't support aliases in the target language can safely ignore it and to them the types will look just like before, while any codegen that does support aliases, can look this declaration up and emit a name at the usage site.

Also included changes to C++ and Capnp codegens to make use of this feature.

Example before:

image

And after:

image


kentonv:
This seems cool but it's going to take me a bit to load all my old thoughts on this from cold storage.

One problem I recall: It's very common to use `using` statements for imports, like:

```
using PipelineDef = import "/edgeworker/scheduling/pipeline.capnp".PipelineDef;
```

This alias is clearly not intended to be seen outside of the file. In fact, if the importing file uses the same C++ namespace as the imported file, then the alias will conflict with the original declaration, and also with other aliases defined in other files.

In fact, most `using` declarations are like this: they are really meant only to be used as shortcuts within the file and not meant to be seen beyond it.

So I think I had always intended to create a separate notion of a "public using" or maybe an "alias" which specifically means "I intend this alias to be used outside this module", and only in that case would it be included in generated code...

RReverser:
> In fact, most `using` declarations are like this: they are really meant only to be used as shortcuts within the file and not meant to be seen beyond it.

I think that's still the case - C++ usings as well as, in our case, Rust ones are still visible only in their own namespace / module. I intentionally tried to keep usings separate from the typesystem itself, as we want them to be visible in the generated docs but not change any type semantics or behaviour.

kentonv:
To be clear: I'm saying that many `using` declarations in capnp are explicitly not intended to be visible at all in the generated code. In C++, there is no module scope. Many modules may declare things in the same namespace. These aliases may pollute that namespace or in many cases lead to compiler errors due to duplicate definitions.

I agree that there are some `using` declarations where outputting an alias in the generated code is highly desirable, but overall its a small minority. We would need to create syntax to explicitly opt into this.

RReverser:
> Many modules may declare things in the same namespace.

Ughhhhh. I hoped that just because it's possible doesn't mean people actually do that...

> but overall its a small minority

But why? Pretty much the only reason to declare a using is to have a named shorthand, aside from those conflicts, what would be the reason where it's "bad" to output the same aliases in the generated output? Opt-in is definitely not what we'd like here - the change becomes kinda pointless if it's not compatible with the vast majority of existing schemas.

If breaking C++ output is the primary concern, I agree those edge cases sound bad enough and I'm happy to revert it and keep the C++ codegen as it was before. I spent some time on it mostly for a completeness of the PR and to make these documentation/IDE benefits accessible to the C++ users as well, but in the end C++ library is not my concern - as long as we can get this information into the `CodeGeneratorRequest` for other codegens, we are good.

kentonv:
Again, it's very common to write things like:

```
using SomeType = import "/some/other/module.capnp".SomeType;
```

This does NOT mean that I want to re-export SomeType from my own module. This declaration is only meant to be references within the capnp file itself.

> Opt-in is definitely not what we'd like here - the change becomes kinda pointless if it's not compatible with the vast majority of existing schemas.

Well... the vast majority of existing schemas don't actually intend to re-export their aliases, so arguably the change is breaking them.

If we make it opt-in we can easily update the few that do actually intend their aliases to be exported.

RReverser:
>Well... the vast majority of existing schemas don't actually intend to re-export their aliases, so arguably the change is breaking them.

Yeah, as I said, I'm happy to revert the C++ codegen as it's a C++ specific issue that generating an alias results in its reexport. I agree with you that reexporting is not the intended effect for most schemas.

This way we can put off an opt-in or another targeted fix for C++ for the later future - at least the capnpc-capnp will still work as intended, and other codegens without similar issue can leverage the information from CodeGeneratorRequest as they see fit.

The only other alternative we discussed is parsing `.capnp` manually in our own codegens and I'd _really_ rather not do that - providing the metadata in the CodeGeneratorRequest, similarly to how it includes doc comments, seems like the right approach.

kentonv:
> it's a C++ specific issue that generating an alias results in its reexport

Wait, you're saying it is not your intent to export the aliases in other languages? What is the point of the feature if the aliases are not exported?

RReverser:
> Wait, you're saying it is not your intent to export the aliases in other languages?

Exactly, as I said in the description: 

> for improved DX through specific alias names in the generated documentation and IDE hover info.

It's purely development experience feature for tools reading the generated code and intentionally not meant to be exposed to the actual typesystem - the aliases remain transparent and not nameable by the code.

Exporting them to the generated code would be a whole separate feature and not the intent of this PR.

puremourning:
Just to link together the threads, as Kenton mentioned this one to me.  

I made a similar, but perhaps more general proposal [here](https://github.com/capnproto/capnproto/issues/907#issuecomment-4883725625) which introduces new syntax (`type NewType = <some type>`  for "aliases which survive to the codegen phase" (newtypes), which also can be pasted into structs "inlining them". I say it's more general because it really defines new types in the schema, and the codegen,  and allows for _annotations_ to be applied to the new types (such as $Json.distriminant, etc.). It's opt-in while being fully wire and existing code generator compatible.

PR #2677, opened 2026-05-28
ArrayPtr::write

mikea:
Writing to array ptr and advancing head is extremely common operation. This change enables using arrayptr as buffer directly and is a great shorthand for `ptr.first(size).copyFrom(data)`.

The other design option would be to make write() immutable and return tail, but I am not sure if that is better. Writing is a mutable operation afterall.

### DIFF
--- a/c++/src/kj/common-test.c++
+++ b/c++/src/kj/common-test.c++
@@ -397,6 +397,72 @@ TEST(Common, ArrayAsBytes) {
   }
 }
 
+KJ_TEST("ArrayPtr write") {
+  int raw[] = {0, 0, 0, 0, 0};
+  ArrayPtr<int> head = raw;
+
+  int first[] = {1, 2};
+  head.write(arrayPtr(first));
+
+  int expectedAfterFirst[] = {1, 2, 0, 0, 0};
+  KJ_EXPECT(arrayPtr(raw) == arrayPtr(expectedAfterFirst));
+  KJ_EXPECT(head.begin() == raw + 2);
+  KJ_EXPECT(head.size() == 3);
+
+  int second[] = {3, 4, 5};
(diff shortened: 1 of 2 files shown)


c++/src/kj/common-test.c++
```diff
@@ -397,6 +397,56 @@ TEST(Common, ArrayAsBytes) {
...
+  KJ_EXPECT(head.size() == 3);
+}
+
+KJ_TEST("ArrayPtr write pieces") {
+  int raw[] = {0, 0, 0, 0, 0, 0};
+  ArrayPtr<int> head = raw;
+  int first[] = {1, 2};
+  int second[] = {3};
+  int third[] = {4, 5};
+  ArrayPtr<const int> pieces[] = {arrayPtr(first), arrayPtr(second), arrayPtr(third)};
```
jasnell, line 440: A test variation that verifies the bounds check would be good.

c++/src/kj/common.h
```diff
@@ -2725,6 +2725,17 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
...
   }
 
+  inline void write(kj::ArrayPtr<const T> other) {
+    // Copy data to the head of this pointer, then advance past the copied data.
+    first(other.size()).copyFrom(other); // first will do a bounds check
+    ptr += other.size();
+    size_ -= other.size();
+  }
+
+  inline void write(kj::ArrayPtr<const kj::ArrayPtr<const T>> pieces) {
```
jasnell, line 2736: Documentation comment explaining that the write is not atomic would be helpful. e.g. given two pieces, if the second fails due to bounds check the first is still committed, etc.

jasnell (approved):
couple of nits, otherwise LGTM

PR #2678, opened 2026-05-28
Convert simple memcpy uses to copyFrom

mikea:


fhanau (approved):
Changes LGTM.
Non-blocking: We can expand UBSan coverage internally slightly if you also clean up memcpys in src/kj/async-io.c++ and src/capnp/compiler/lexer.c++ which are currently exempted from the nonnull-attribute check.

PR #2679, opened 2026-05-28
trivial memset fixes

mikea:


fhanau (approved):
Changes LGTM. Style could be more consistent in terms of using `foo = {}` vs. `kj::arrayPtr(foo).fill()`, but this is non-blocking.

mikea:
> Changes LGTM. Style could be more consistent in terms of using `foo = {}` vs. `kj::arrayPtr(foo).fill()`, but this is non-blocking.

I was already trying to use `foo = {}` everywhere. I looked one more time and identified only one case in http.c++ that I have missed. LMK if you notice more.

PR #2682, opened 2026-06-01
kj::WeakRc<T>

mikea:
on top of https://github.com/capnproto/capnproto/pull/2655

### DIFF
--- a/c++/src/kj/refcount.h
+++ b/c++/src/kj/refcount.h
@@ -37,13 +37,36 @@ namespace kj {
 template<typename T>
 class Rc;
 
+template<typename T>
+class WeakRc;
+
+class Refcounted;
+
 template <typename T, typename... Params>
 Rc<T> rc(Params&&... params);
 
 namespace _ {  // private
 
+struct RcWeakControl;
 template <typename T> class RcWrapper;
 template <typename T> class RcOwnWrapper;
+
+}  // namespace _ (private)
+
+namespace _ {  // private
+
+struct RcWeakControl {
+  RcWeakControl(Refcounted* refcounted, void* ptr): refcounted(refcounted), ptr(ptr) {}
+
+  void addRef() { ++refcount; }
+  void dispose() {
+    if (--refcount == 0) delete this;
+  }
+
+  uint refcount = 1;
+  Refcounted* refcounted;
+  void* ptr;
+};
 
 }  // namespace _ (private)
 
@@ -91,7 +114,17 @@ private:
   mutable uint refcount = 0;
   // "mutable" because disposeImpl() is const.  Bleh.
 
+  mutable _::RcWeakControl* weakControl = nullptr;
+
   void disposeImpl(void* pointer) const override;
+
+  _::RcWeakControl* addWeakRefInternal(void* ptr) const {
+    if (weakControl == nullptr) {
+      weakControl = new _::RcWeakControl(const_cast<Refcounted*>(this), ptr);
+    }
+    weakControl->addRef();
+    return weakControl;
+  }
 
   template <typename T>
   static Own<T> addRefInternal(T* object);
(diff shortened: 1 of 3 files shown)


c++/src/kj/refcount.h
```diff
@@ -319,19 +355,126 @@ class Rc {
...
+  }
+
+  kj::Maybe<kj::Rc<T>> tryUpgrade() {
+    if (weakControl == nullptr || weakControl->refcounted == nullptr) return kj::none;
+    auto refcounted = weakControl->refcounted;
+    ++refcounted->refcount;
+    return Rc<T>(refcounted, ptr);
+  }
+
+  kj::Maybe<T&> tryGet() {
```
jasnell, line 419: Nit: I've been finding myself wishing we could use `kj::Ptr<T>` for stuff like this. We need to actively discourage holding raw references.

c++/src/kj/refcount.h
```diff
@@ -319,19 +355,126 @@ class Rc {
...
+  kj::WeakRc<T> addRef() {
+    if (weakControl != nullptr) {
+      weakControl->addRef();
+      return WeakRc(weakControl, ptr);
+    } else {
+      return kj::WeakRc<T>();
+    }
+  }
+
+  kj::WeakRc<T> clone() {
```
jasnell, line 408: Why not just `addRef()`

c++/src/kj/refcount.h
```diff
@@ -319,19 +355,126 @@ class Rc {
...
+
+  template <typename>
+  friend class WeakRc;
 };
 
+template<typename T>
+class WeakRc {
+  // Weak reference to an Rc<T>-managed object. A WeakRc<T> does not keep T alive, but can be
+  // upgraded back to Rc<T> while some strong Rc/Own reference still exists.
+public:
```
jasnell, line 364: If I'm understanding correctly, the `T` itself needs to be `kj::Refcounted`. That's a bit unfortunate. The usage pattern we have in workerd with `workerd::WeakRef<T>` is that `T` does not need to be `kj::Refcounted` itself.

What I'd hope for is something like:

```c++
// Uniquely owned
kj::Own<T> ownT = kj::heap<T>();
kj::Weak<T> weakT = ownT.addWeak();

// Refcounted
k::Rc<U> rcU = kj::rc<U>();
kj::Weak<U> weakU = rcU.addWeak();

kj::Weak<T> otherT = weakT.addRef();
kj::Weak<U> otherU = weakU.addRef();

// When ownT is dropped, the kj::Weak<T> instances are invalidated
// when rcU refcount reaches zero, the kj::Weak<U> instances are invalidated
```

Keep in mind also that we do have `AtomicWeakRef` that is used in at least one critical place in workerd.

mikea: > If I'm understanding correctly, the T itself needs to be kj::Refcounted

not really. It builds on top of kj::Rc<T> for any T and supports any kind of T. The proper wrapper will be allocated accordingly.

jasnell: Ah, does this build then on your concept of `Rc<T>` without `T` having to subclass `kj::Refcounted` then?

c++/src/kj/refcount.h
mikea: there are both. clone() is useful since Maybe and other containers know how to invoke it.

c++/src/kj/refcount.h
mikea: > Ah, does this build then on your concept of Rc<T> without T having to subclass kj::Refcounted then?

yes. as of couple hours ago kj::Rc<T> works for all Ts and this change works for all Ts as well.

c++/src/kj/refcount.h
mikea: ack. added it to my todoooo list.

mikea:
somehow I've got a duplcate PR going: https://github.com/capnproto/capnproto/pull/2713
Closing this one in favor of the newer.

PR #2684, opened 2026-06-02
fix payload length shift in websocket frame header

sahvx655-wq:
The 8-byte extended length in Header::compose uses `payloadLen >> 42` for bytes[5], which drops bits 32-39 and leaks bits 42-49; for frames over 4 GiB the peer decodes a wrong length and desyncs. Should be `>> 32` to match the 56/48/40/32/24/16/8/0 sequence.


kentonv (approved):

kentonv:
omg

kentonv:
thanks!

PR #2685, opened 2026-06-02
Scope testing to project

denialhaag:
We were trying to fetch Cap'n Proto via `FetchContent` in a C++ project and ran into the issue that we couldn't disable the tests due to unconditional calls to `include(CTest)`. The proposed changes should fix this.

It would also be possible to define a, say, `CAPNPROTO_BUILD_TESTS` option that defaults to `PROJECT_IS_TOP_LEVEL`. This would make the configuration more configurable.


c++/CMakeLists.txt
```diff
@@ -6,7 +6,10 @@ set(VERSION 2.0-dev)
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 
-include(CTest)
+if (PROJECT_IS_TOP_LEVEL)
```
harrishancock: We require CMake 3.16+, but PROJECT_IS_TOP_LEVEL was added in CMake 3.21.

Since `if(PROJECT_IS_TOP_LEVEL)` will evaluate to false if the variable is undefined, this will break testing on earlier CMake versions.

We could bump our required version, but Debian Bullseye (current LTS release) still ships with 3.18, and I think we generally try to stay compatible with popular distros' LTS releases.

```suggestion
# PROJECT_IS_TOP_LEVEL was added in CMake 3.21
if (NOT DEFINED PROJECT_IS_TOP_LEVEL OR PROJECT_IS_TOP_LEVEL)
```

CMakeLists.txt
```diff
@@ -1,5 +1,8 @@
 cmake_minimum_required(VERSION 3.16)
 project("Cap'n Proto Root" CXX)
-include(CTest)
+
+if (PROJECT_IS_TOP_LEVEL)
```
harrishancock:
```suggestion
# PROJECT_IS_TOP_LEVEL was added in CMake 3.21
if (NOT DEFINED PROJECT_IS_TOP_LEVEL OR PROJECT_IS_TOP_LEVEL)
```

harrishancock (commented):
Would you mind backporting this to the master branch, as well?

denialhaag:
> Would you mind backporting this to the master branch, as well?

Sure! I've just created #2693. 🙂 

If I'm not mistaken, the backport does not need any additional changes. 🤔

c++/CMakeLists.txt
denialhaag: Ahh, that makes sense! Thanks for the context! 🙂

harrishancock (approved):

PR #2688, opened 2026-06-03
read http byte-range positions as 64-bit

sahvx655-wq:
Reading through tryParseHttpRangeHeader I noticed the range positions come from consumeNumber, which accumulates into a 32-bit uint and silently wraps, and firstPos/lastPos/suffixLength are uint too. But HttpByteRange and the contentLength argument are uint64_t, so any offset at or above 2^32 gets truncated. On an 8 GB resource bytes=5000000000-5000000001 comes back as {705032704,705032705}, and bytes=4294967298- against an 8-byte resource wraps the start down to 2, turning an unsatisfiable range into a satisfiable one that serves the wrong bytes.

This adds consumeNumber64 which accumulates into uint64_t and rejects values that don't fit instead of wrapping, and widens the range locals to match. Added a few range-parsing cases past 2^32.

### DIFF
--- a/c++/src/kj/compat/http.c++
+++ b/c++/src/kj/compat/http.c++
@@ -852,6 +852,30 @@ static kj::Maybe<uint> consumeNumber(char*& ptr) {
   return result;
 }
 
+static kj::Maybe<uint64_t> consumeNumber64(const char*& ptr) {
+  // Like consumeNumber(), but accumulates into a 64-bit value and rejects (rather than silently
+  // wrapping) numbers that don't fit. Used for byte-range positions, which are compared against a
+  // 64-bit content length.
+  const char* start = skipSpace(ptr);
+  const char* p = start;
+
+  uint64_t result = 0;
+
+  for (;;) {
+    const char c = *p;
+    if ('0' <= c && c <= '9') {
+      uint digit = c - '0';
+      if (result > (uint64_t(kj::maxValue) - digit) / 10) return kj::none;
+      result = result * 10 + digit;
+      ++p;
+    } else {
+      if (p == start) return kj::none;
+      ptr = p;
+      return result;
+    }
+  }
+}
+
 static kj::StringPtr consumeLine(char*& ptr) {
   char* start = skipSpace(ptr);
   char* p = start;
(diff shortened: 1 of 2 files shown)


c++/src/kj/compat/http.c++
```diff
@@ -852,6 +852,30 @@ static kj::Maybe<uint> consumeNumber(char*& ptr) {
...
+  const char* start = skipSpace(ptr);
+  const char* p = start;
+
+  uint64_t result = 0;
+
+  for (;;) {
+    const char c = *p;
+    if ('0' <= c && c <= '9') {
+      uint digit = c - '0';
+      if (result > (uint64_t(kj::maxValue) - digit) / 10) return kj::none;
```
kentonv: Does the integer division on every loop make this sort of slow? Is there a better way? I guess it could be like:

```
if ((digit <= 5 && result > 1844674407370955161ull) ||
    (digit > 5 && result > 1844674407370955160ull))
```

Or is the compiler smart enough to figure that out itself?

Probably good to include tests that the top 10 (or more) possible uint64 values are parsed correctly.

sahvx655-wq: Good point. The divisor was a constant 10, so the compiler was already lowering it to a multiply+shift rather than a real divide, but the `- digit` inside the expression meant it recomputed the threshold every iteration. I've reworked it so the bound is `MAX / 10` with a `digit > MAX % 10` tie-break, both of which fold to compile-time constants, so the hot path is now just a couple of comparisons against constants with no division at all. I went with named constants rather than the literal 1844674407370955161 form mainly so it's self-documenting.

For the tests I added the top ten uint64 values (2^64-10 through 2^64-1) plus a few past 2^64. I leant on the suffix path against a 4-byte resource: a suffix length >= content resolves to the whole resource, so a value that parses cleanly comes back as everything while one that's rejected for overflow falls through to unsatisfiable, which lets the two cases actually be told apart. The boundary lands where you'd expect, 2^64-1 parses and 2^64 is rejected. kj-heavy-tests builds with my change (the unrelated async-test failure on this toolchain aside) and the Range header parsing case passes.

sahvx655-wq:
gentle ping

kentonv (approved):
Thanks, this looks great. And sorry again for the delay -- as mentioned on other issues I was unfortunately unable to keep up with email in June.

I think this issue is slightly less serious than the others but I still plan to roll it into the same release.

kentonv:

PR #2690, opened 2026-06-04
Convert simple memcpy uses to copyFrom

mikea:
Round 2 of memcpy conversion. I think there will be 3 & 4.

### DIFF
--- a/c++/src/kj/string-test.c++
+++ b/c++/src/kj/string-test.c++
@@ -55,7 +55,7 @@ TEST(String, Str) {
   Vector<char> v;
   v.addAll(a);
   FixedArray<char, 3> f;
-  memcpy(f.begin(), buf, 3);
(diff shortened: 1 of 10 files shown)


c++/src/kj/string-test.c++
```diff
@@ -55,7 +55,7 @@ TEST(String, Str) {
   Vector<char> v;
   v.addAll(a);
   FixedArray<char, 3> f;
-  memcpy(f.begin(), buf, 3);
+  kj::arrayPtr(f.begin(), 3).copyFrom(kj::arrayPtr(buf, 3));
```
fhanau, line 58: nit: this can be simply `f.asPtr().copyFrom(kj::arrayPtr(buf, 3))`

fhanau (approved):

PR #2693, opened 2026-06-08
Scope testing to project (backport #2685)

denialhaag:
This PR is a backport of #2685.

> We were trying to fetch Cap'n Proto via `FetchContent` in a C++ project and ran into the issue that we couldn't disable the tests due to unconditional calls to `include(CTest)`. The proposed changes should fix this.
> 
> It would also be possible to define a, say, `CAPNPROTO_BUILD_TESTS` option that defaults to `PROJECT_IS_TOP_LEVEL`. This would make the configuration more configurable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,9 @@
 cmake_minimum_required(VERSION 3.10)
 project("Cap'n Proto Root" CXX)
-include(CTest)
+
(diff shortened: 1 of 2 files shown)


harrishancock (approved):

harrishancock (approved):

harrishancock:
FreeBSD failure is a known, unrelated issue.

PR #2694, opened 2026-06-08
fix offset overflow in schema loader struct validation

sahvx655-wq:
SchemaLoader::Validator::validate() checks each slot field's position with `fieldBits * (slot.getOffset() + 1) <= dataSizeInBits`, and the union discriminant with `(getDiscriminantOffset() + 1) * 16 <= dataSizeInBits`. Those offsets are UInt32 values copied straight from the schema node, but the `+ 1` and the multiply run in 32-bit arithmetic, so an offset near 2^32 (0xffffffff being the easy one) wraps the product back down to a small number and the out-of-range field passes validation.

This is the layer that has to get it right: StructBuilder::setDataField()/getDataField() index the data section straight from the offset with no bounds check of their own, on the assumption that the loader already rejected anything that does not fit, so a node accepted here becomes an out-of-bounds read or write the moment a message is built or read against it. I widened the offset arithmetic to 64 bit at the three affected sites so the comparison can no longer wrap; legitimate schemas are unaffected because their offsets are tiny. The added test loads TestAllTypes after setting a UInt64 field offset to 0xffffffff and confirms the loader now rejects it, where previously the malformed node loaded cleanly.

### DIFF
--- a/c++/src/capnp/schema-loader-test.c++
+++ b/c++/src/capnp/schema-loader-test.c++
@@ -234,6 +234,31 @@ TEST(SchemaLoader, Incompatible) {
       loadUnderAlternateTypeId<test::TestAllTypes>(loader, typeId<test::TestListDefaults>()));
 }
 
+TEST(SchemaLoader, OutOfBoundsFieldOffset) {
+  // A field's offset is a UInt32 taken straight from the schema node. The validator must reject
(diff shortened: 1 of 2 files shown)


sahvx655-wq:
gentle ping

kentonv (approved):

Thanks!

kentonv:

PR #2696, opened 2026-06-09
fix segment size overflow in message readers

sahvx655-wq:
1. the message segment table carries an untrusted uint32 word count per segment; InputStreamMessageReader and AsyncMessageReader sum those into a `size_t totalWords`, which wraps on a 32-bit build and slips past the `totalWords <= traversalLimitInWords` check meant to bound the allocation.
2. FlatArrayMessageReader::init vets each segment with `array.size() >= offset + segmentSize`, where the addition wraps the same way.
3. ArrayPtr::slice only bounds-checks under KJ_IREQUIRE, which is compiled out of release builds, so once a check is bypassed the reader hands back segment pointers past the buffer and they get read during traversal.

Sum the words in uint64_t and write the bound as `array.size() - offset >= segmentSize` so neither can overflow. 64-bit builds are unaffected; this removes an out-of-bounds read on 32-bit deployments fed a hostile message.

### DIFF
--- a/c++/src/capnp/serialize-async.c++
+++ b/c++/src/capnp/serialize-async.c++
@@ -139,7 +139,7 @@ kj::Promise<void> AsyncMessageReader::readAfterFirstWord(kj::AsyncInputStream& i
 
 kj::Promise<void> AsyncMessageReader::readSegments(kj::AsyncInputStream& inputStream,
                                                    kj::ArrayPtr<word> scratchSpace) {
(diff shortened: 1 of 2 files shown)


sahvx655-wq:
any update?

kentonv (commented):

Thanks!

kentonv (approved):

kentonv:

PR #2697, opened 2026-06-12
kj::Pin/Ptr quality of life improvements

mikea:


c++/src/kj/memory.h
```diff
@@ -815,6 +822,19 @@ class Ptr {
...
+// This enables niche optimization: Maybe<Ptr<T>> uses ptr == nullptr as "none".
+template <typename T>
+struct MaybeTraits<Ptr<T>> {
+  static void initNone(Ptr<T>* ptr) noexcept { new (ptr, _::PlacementNew()) Ptr<T>(nullptr); }
+  static bool isNone(const Ptr<T>& p) noexcept { return p.ptr == nullptr; }
+
+  // Ptr's move ctor just copies ptr/counter and sets source.ptr to nullptr. Moving a null Ptr is
+  // safe when the null state is constructed via initNone().
+  static constexpr bool noneIsMoveSafe = true;
 };
```
harrishancock, line 841: `kj::Ptr<T>` can be constructed from `kj::Ptr<U>`, so this might also be useful:

```suggestion

  // Allow `Maybe<Ptr<T>>` to be constructed from types convertible to `Ptr<T>`, like `Ptr<U>`.
  static constexpr bool convertingConstructor = true;
};
```

harrishancock (approved):

PR #2698, opened 2026-06-12
Weak<T>

mikea:
Weak<T> works with Pin<T> and does not prevent Pin<T> from being destroyed.

Current implementation is simple, but allocates. I imagine non-allocating version with double-linked list, but not sure if complexity is warranted yet. My plan is to use it in high-level workerd code, where sporadic allocations should be fine.

```
  // Weak<T> is a smart alternative to T& with expiration detection.
  //
  // Weak<T> is obtained from Pin<T>::addWeak(). It does not keep the Pin alive and does not prevent
  // the Pin from moving.
  // Common usage:
  // - KJ_IF_SOME on Weak<T> upgrades to Ptr<T>
  // - assertLive() obtains T& and throws on expired Weak<T>
  // - tryGet() obtains Maybe<T&> directly.
  // - upgrade() method upgrades to Maybe<Ptr<T>>
```

### DIFF
--- a/c++/src/kj/memory.h
+++ b/c++/src/kj/memory.h
@@ -185,44 +185,107 @@ public:
 
 
 // =======================================================================================
-// Ptr Counters
+// Pointer tracking
 
 namespace _ {
-#if KJ_ASSERT_PTR_COUNTERS
+
+class PtrControl;
+
+class WeakCell {
+  // Shared validity cell for kj::Weak<T>. Pin owns one reference while the referent is alive;
+  // Weak owns one reference per weak pointer. When Pin is destroyed or moved, ptr is nulled before
+  // Pin releases its reference, allowing outstanding Weak pointers to observe expiration safely.
+
+public:
+  explicit WeakCell(void* ptr, PtrControl* control): ptr(ptr), control(control) {}
+
+  inline void addRef() { ++refcount; }
+  inline void decRef() { if (--refcount == 0) { delete this; } }
+
+  void* ptr;
+  PtrControl* control;
+
+private:
+  size_t refcount = 1;
+};
 
 void atomicPtrCounterAssertionFailed(const char* const);
 
-class AtomicPtrCounter {
-  // AtomicPtrCounter uses atomic operations to keep track of active pointers.
-  // Since no other memory location is observed, memory_order_relaxed is used.
+class PtrControl {
+#if KJ_ASSERT_PTR_COUNTERS
+  class AtomicPtrCounter {
+    // AtomicPtrCounter uses atomic operations to keep track of active pointers.
+    // Since no other memory location is observed, memory_order_relaxed is used.
+
+  public:
+    inline void dec() {
+      size_t prevCount = count.fetch_sub(1, std::memory_order_relaxed);
+      if (KJ_UNLIKELY(prevCount == 0)) {
+        atomicPtrCounterAssertionFailed("unbalanced inc/dec");
+      }
+    }
+
+    inline void inc() {
+      count.fetch_add(1, std::memory_order_relaxed);
+    }
+
+    inline void assertEmpty() {
+      size_t c = count.load(std::memory_order_relaxed);
+      if (KJ_UNLIKELY(c != 0)) {
+        atomicPtrCounterAssertionFailed("active pointers exist");
+      }
+    }
+
+  private:
+    std::atomic<size_t> count = 0;
+  };
+#endif // KJ_ASSERT_PTR_COUNTERS
 
 public:
-  inline void dec() {
-    size_t prevCount = count.fetch_sub(1, std::memory_order_relaxed);
-    if (KJ_UNLIKELY(prevCount == 0)) {
-      atomicPtrCounterAssertionFailed("unbalanced inc/dec");
+  PtrControl() = default;
+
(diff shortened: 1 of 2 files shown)


c++/src/kj/memory.h
```diff
@@ -815,8 +914,173 @@ class Ptr {
...
+  inline const T& operator*() const { return get(); }
+
+  inline bool operator==(const Pin<T>& other) const { return tryGetPtr() == other.get(); }
+  inline bool operator==(const Weak<T>& other) const { return tryGetPtr() == other.tryGetPtr(); }
+  inline bool operator==(const T* const other) const { return tryGetPtr() == other; }
+
+  inline T& get() {
+    // Obtain a `T&` reference, checking that the referent is still alive.
+    T* ptr = tryGetPtr();
+    KJ_IREQUIRE(ptr != nullptr, "null Weak<> dereference");
```
mikea, line 986: I am debating if we want to have null-checking methods on Weak. I usually err on the side of stricter API and this causes incovenience. Curious what you think?

c++/src/kj/memory.h
```diff
@@ -185,44 +185,123 @@ class NullDisposer: public Disposer {
...
   }
 
-private:
-  std::atomic<size_t> count = 0;
-};
+  inline void releaseWeakCell() {
+    if (weakCell != nullptr) {
+      weakCell->ptr = nullptr;
+      weakCell->control = nullptr;
+      weakCell->release();
```
harrishancock: Nit: I know this is only called from a destructor, but it'd be nice to set `weakCell = nullptr` after releasing it.

c++/src/kj/memory.h
```diff
@@ -643,37 +722,39 @@ class SpaceFor {
...
+    control.moveWeakCellFrom(other.control, &this->t);
   }
 
   inline ~Pin() {
     // Destroy a Pin with underlying object.
     // Undefined behavior when live pointers exist, asserted when KJ_ASSERT_PTR_COUNTERS is defined.
-#if KJ_ASSERT_PTR_COUNTERS
-    ptrCounter.assertEmpty();
-#endif
+    control.assertEmpty();
```
harrishancock: If this throws, I think we leak the WeakCell. Could we `releaseWeakCell()` first?

c++/src/kj/memory.h
harrishancock: I suspect we'll end up wanting null-checking at some point, but I don't feel strongly. I'm happy shipping something stricter until then, though.

One philosophy could be to make it consistent with the Maybe API:

```cpp
// it's like a Maybe
KJ_IF_SOME(p, weak) {  ... }
weak = kj::none;
bool alive = weak == kj::none;

// OR

// it's not a Maybe
KJ_IF_SOME(p, weak.upgrade()) { ... }
weak = nullptr;
bool alive = weak == nullptr;
```

I think supporting `KJ_IF_SOME` for reading, but `nullptr` for nullifying/null-checking would end up looking weird to me.

c++/src/kj/memory.h
```diff
@@ -185,44 +185,123 @@ class NullDisposer: public Disposer {
...
+    return ptr == this->ptr;
   }
 
-  inline void assertEmpty() {
-    size_t c = count.load(std::memory_order_relaxed);
-    if (KJ_UNLIKELY(c != 0)) {
-      atomicPtrCounterAssertionFailed("active pointers exist");
+  inline void moveWeakCellFrom(PtrControl& other, void* newPtr) {
+    ptr = newPtr;
+    weakCell = other.weakCell;
```
harrishancock: Nit: asserting `weakCell == nullptr` would be a defensive improvement, in case we add an assignment operator to `Pin`.

c++/src/kj/memory.h
```diff
@@ -815,8 +914,164 @@ class Ptr {
...
+class Weak {
+  // Weak<T> is a smart alternative to T& with expiration detection.
+  //
+  // Weak<T> is obtained from Pin<T>::addWeak(). It does not keep the Pin alive and does not prevent
+  // the Pin from moving (but does follow it).
+  // Common usage:
+  // - KJ_IF_SOME on Weak<T> upgrades to Ptr<T>
+  // - assertLive() obtains T& and throws on expired Weak<T>
+  // - tryGet() obtains Maybe<T&> directly.
+  // - upgrade() method upgrades to Maybe<Ptr<T>>
```
harrishancock, line 929: Just a thought: I wonder if we should support implicit conversion to `Maybe<Ptr<T>>`. We might get `KJ_IF_SOME` support for free without having to implement `_::readMaybe()` here.

I'm not sure if it would work, or if it's actually a good idea, but it would justify "supports KJ_IF_SOME but not `kj::none` comparison" better to me.

mikea: Upgrading pointer is not free, strictly speaking (requires incrementing a counter), so I am not sure conversion to `Maybe<Ptr<T>>` should be implicit.

Convertion to `Maybe<T&>` _is_ free, but I specifically want KJ_IF_SOME to upgrade, rather give a raw reference.

mikea:
thanks for review, all comments addressed ptal.

harrishancock (approved):

PR #2700, opened 2026-06-15
opt Ptr<T>

mikea:
Since crashing in opt mode is a no-go and throwing at ~T point cannot protect Ptr<T>, this PR keeps existing crashing behavior in debug mode for eaiser contract debugging, but replaces Ptr<T> with checked Weak<T> in opt , thus making it safe to replace T& with Ptr<T> and always get additional security guarantee.

I've split Weak<T> textual move to a separate commit for easier review. There's no new logic, opt Ptr<T> simply wraps Weak<Y>.


harrishancock (approved):

harrishancock:
> but replaces Ptr with checked Weak in debug mode

s/debug/opt/, but it's just the PR description

PR #2702, opened 2026-06-18
use atomic counter for WeakCell

mikea:


kentonv:
Ugh, this looks pretty expensive -- there's a reason we generally avoid atomic refcounting as much as we can.

Shouldn't we be introducing some sort of separate "atomic weak" variant? And it should be an error to hold weak references cross-thread?

harrishancock (approved):

mikea:
> Ugh, this looks pretty expensive -- there's a reason we generally avoid atomic refcounting as much as we can.
> 
> Shouldn't we be introducing some sort of separate "atomic weak" variant? And it should be an error to hold weak references cross-thread?

I'll just revert my `opt Ptr<T>` change while I figure this out. https://github.com/capnproto/capnproto/pull/2704

PR #2712, opened 2026-06-26
extract base PtrTarget class

mikea:
PtrTarget can be used to obtain Ptr/Weak references to this similar to Refcounted. This is a common pattern for listeners in our code.

### DIFF
--- a/c++/src/kj/memory.h
+++ b/c++/src/kj/memory.h
@@ -191,23 +191,33 @@ public:
 // =======================================================================================
 // Pointer tracking
 
+template <typename T>
+class Ptr;
+
+template <typename T>
+class Weak;
+
+template <typename T>
+class Pin;
+
+class PtrTarget;
+
 namespace _ {
 
-class PtrControl;
-
 class WeakCell {
-  // Shared validity cell for kj::Weak<T>. Pin owns one reference while the referent is alive;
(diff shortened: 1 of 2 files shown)


c++/src/kj/memory.h
```diff
@@ -890,15 +940,20 @@ class Ptr {
 
   inline Ptr(T* ptr, _::WeakCell* cell) : ptr(ptr), control(cell->control) { control->inc(); }
 
+  inline Ptr(T* ptr, PtrTarget* control) : ptr(ptr), control(control) { control->inc(); }
+  // Construct a Ptr that refers directly to a PtrTarget-derived object. Used by
+  // PtrTarget::addPtrToThis().
+
   T *ptr;
-  _::PtrControl* control;
+  PtrTarget* control;
```
harrishancock: Nit
```suggestion
  PtrTarget* target;
```

(and elsewhere, of course)

c++/src/kj/memory.h
```diff
@@ -787,7 +837,7 @@ class Pin {
   }
 
   T t;
-  _::PtrControl control;
+  PtrTarget control;
```
harrishancock: Ditto
```suggestion
  PtrTarget target;
```

harrishancock (approved):

PR #2713, opened 2026-06-26
WeakRc<T>

mikea:
Follows Weak<T> design with a separate on-demand weak cell.

### DIFF
--- a/c++/src/kj/refcount-test.c++
+++ b/c++/src/kj/refcount-test.c++
@@ -223,6 +223,23 @@ KJ_TEST("Rc clone") {
   EXPECT_TRUE(b);
 }
 
+KJ_TEST("Rc self-assignment") {
+  bool b = false;
+
+  Rc<SetTrueInDestructor> ref = kj::rc<SetTrueInDestructor>(&b);
+  auto ptr = ref.get();
+  auto refPtr = &ref;
+
+  ref = kj::mv(*refPtr);
+
+  EXPECT_TRUE(ref != nullptr);
+  EXPECT_TRUE(ref.get() == ptr);
+  EXPECT_FALSE(b);
+
+  ref = nullptr;
+  EXPECT_TRUE(b);
+}
+
 KJ_TEST("Rc container clone") {
   bool b = false;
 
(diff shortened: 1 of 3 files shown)


c++/src/kj/refcount-test.c++
```diff
@@ -521,6 +521,250 @@ KJ_TEST("Rc inheritance") {
...
+    KJ_FAIL_EXPECT("expected WeakRc to upgrade while referent is alive");
+  }
+  EXPECT_FALSE(ref->isShared());
+  EXPECT_FALSE(b);
+
+  ref = nullptr;
+  EXPECT_TRUE(b);
+
+  // The WeakRc has now expired.
+  EXPECT_TRUE(weak == nullptr);
```
harrishancock, line 580: Does `weak == ref` in this state?

I originally expected it shouldn't. But I guess they'd both be equal to `nullptr`, so I guess it should.

c++/src/kj/refcount.h
```diff
@@ -262,6 +306,14 @@ class Rc {
     return addRef();
   }
 
+  WeakRc<T> downgrade();
+  // Create a weak reference to the referent. The weak reference does not keep the object alive;
+  // it expires once the last strong Rc<T> is dropped, but can be upgraded back to an Rc<T> while
+  // the object is still alive. See kj::WeakRc<T>.
+
+  WeakRc<T> asWeakRc() { return downgrade(); }
+  // Synonym for downgrade().
```
harrishancock, line 315: Non-blocking nit: I'd omit the synonyms (`asWeakRc()`, `asRc()`), so that people won't wonder which one they're supposed to use. I don't feel strongly, though.

c++/src/kj/refcount-test.c++
```diff
@@ -521,6 +521,250 @@ KJ_TEST("Rc inheritance") {
...
+
+  WeakRc<SetTrueInDestructor> weak4 = nullptr;
+  weak4 = weak3.clone();  // explicit copy + move assign
+  EXPECT_TRUE(weak4 == ref);
+
+  WeakRc<SetTrueInDestructor> weak5 = nullptr;
+  weak5 = kj::mv(weak4);  // move assign
+  EXPECT_TRUE(weak4 == nullptr);
+  EXPECT_TRUE(weak5 == ref);
+
```
harrishancock, line 682: Maybe test self-assignment somewhere, too? I didn't see if you had it.

harrishancock (approved):

c++/src/kj/refcount.h
mikea: I decided to keep them but with a different names: `addWeakRef` and `addStrongRef` for consistency. I'm actually not very sure that upgrade/downgrade are really discoverable.

c++/src/kj/refcount-test.c++
mikea: > Does weak == ref in this state?

yes. They are both null as you corretly noticed. Added expect

c++/src/kj/refcount.h
harrishancock: I like those names. I agree upgrade/downgrade probably aren't as discoverable, though I also like them.

PR #2729, opened 2026-07-17
change Refcounted to start count from 1

mikea:
This has 2 positive effects:

- makes building and accesing weak pointers and constructor easy and straightforward (main reason why I am doing this)
- allows us to detect on-the-stack allocation.

### DIFF
--- a/c++/src/kj/refcount.c++
+++ b/c++/src/kj/refcount.c++
@@ -35,7 +35,21 @@ namespace kj {
 // Non-atomic (thread-unsafe) refcounting
 
 Refcounted::~Refcounted() noexcept(false) {
-  KJ_ASSERT(refcount == 0, "Refcounted object deleted with non-zero refcount.");
+  // A Refcounted object is born with a refcount of 1, so if a subclass constructor throws, the
+  // object is destroyed while refcount is still non-zero. Any weak references created and
+  // published by the constructor must observe that the referent has expired, just as they do when
+  // the last strong reference is dropped normally.
+  if (refcount != 0 && weakCell != nullptr) {
+    weakCell->refcounted = nullptr;
+    weakCell->decRef();
+    weakCell = nullptr;
+  }
+
+  // Suppress the assertion when we're unwinding due to a constructor exception. This is a
+  // debug-only assertion so that release builds do not do expensive unwinding checks.
+  KJ_DASSERT(refcount == 0 || UnwindDetector::uncaughtExceptionCount() > 0,
+      "Refcounted object deleted with non-zero refcount; it appears to have "
+      "been allocated without using kj::rc.");
 }
 
 void Refcounted::disposeImpl(void* pointer) const {
--- a/c++/src/kj/refcount.h
+++ b/c++/src/kj/refcount.h
@@ -120,7 +120,11 @@ protected:
   }
 
 private:
-  mutable uint refcount = 0;
+  mutable uint refcount = 1;
+  // A Refcounted object is born with a reference count of 1: it always comes into existence owned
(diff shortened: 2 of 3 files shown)


c++/src/kj/refcount.c++
```diff
@@ -35,7 +35,8 @@ namespace kj {
 // Non-atomic (thread-unsafe) refcounting
 
 Refcounted::~Refcounted() noexcept(false) {
-  KJ_ASSERT(refcount == 0, "Refcounted object deleted with non-zero refcount.");
+  KJ_ASSERT(refcount == 0, "Refcounted object deleted with non-zero refcount; it appears to have "
+      "been allocated on the stack or inline.");
```
harrishancock: If `T` inherits from `Refcounted` and throws in its constructor, I think this assertion will now spuriously fire.

I suppose we could suppress the assertion if we're already unwinding due to an exception?

kentonv: We should definitely suppress any destructor assertion in cases where we're already unwinding.

The error message suggests that this can only happen through stack or inline allocation, but it can also happen through `new T`, can't it? Or via `kj::heap`, unless you added some compile-time detection to prevent that.

c++/src/kj/refcount.h
```diff
@@ -284,6 +288,10 @@ class Rc {
   }
 
   inline Rc(Own<T> t) noexcept {
+    // Taking over an Own<T> is only supported for non-refcounted types. A type that is already
```
kentonv: Hmm, not sure I see why this needs to be disallowed.

mikea: Current implementation is not really sound for this case. It does not restore the original pointer, but always wraps. Hence re-owning this Rc and creating a weak pointer for it wouldn't do anything good.

This can be fixed by peeking into Own and checking the disposer, but what if someone calls .attach somewhere?

In reality we _never_ reconvert kj::Own<T> back to Rc<T>. This code path is only used for things like `kj::Own<AsyncInputStream> => kj::Rc<AsyncInputStream>` conversion where it is sound.

c++/src/kj/refcount.h
```diff
@@ -120,7 +120,11 @@ class Refcounted: private Disposer {
   }
 
 private:
-  mutable uint refcount = 0;
+  mutable uint refcount = 1;
```
vaci, line 123: Doesn't this mean that refcounted classes can no longer be cheaply memset-zero initialized by the compiler?

mikea: Proper way to create refcounted objects is through `kj::rc` which currently allocates an object on the heap and immediately increments ref count after init. Thus I don't think initializing counter to 1 would create any extra cycles.

vaci: You could just leave it at zero to mean "one reference exists", but I guess you'd lose the ability to detect if `kj::rc` _wasn't_ used...

If most refcounted objects only ever have one reference, that might be a tiny optimization as you'd never actually have to increment or decrement the counter.

c++/src/kj/refcount.c++
mikea: Updated to check for unwinding. I decided to keep this check debug-only since unwinding check is a bit expensive.

> The error message suggests that this can only happen through stack or inline allocation, but it can also happen through new T, can't it? Or via kj::heap, unless you added some compile-time detection to prevent that.

Updated error message. As for kj::heap there's this: https://github.com/capnproto/capnproto/pull/2728 (if you could take a quick look at it?)

c++/src/kj/refcount.h
mikea: I have extracted this to a separate change for easier review: https://github.com/capnproto/capnproto/pull/2732

harrishancock (approved):

PR #2731, opened 2026-07-20
Guard TCP_NODELAY setsockopt with address family check

ViniciusCestarii:
Check if socket family is AF_INET or AF_INET6 before setsockopt(TCP_NODELAY) to prevent relying on platform-specific errno codes to detect when it's unsupported.

I left the errno checks as is for a safer PR and ensure no behavior change.

This also gives a small performance win (one less unnecessary syscall for when the socket is not in INET family).

Not sure if this was already considered or discussed elsewhere or if there's a reason the errno-based approach was preferred. I'd be happy to hear feedback if this is too naive or missing something.

### DIFF
--- a/c++/src/kj/async-io-unix.c++
+++ b/c++/src/kj/async-io-unix.c++
@@ -1354,24 +1354,26 @@ public:
         // Ignore disallowed address.
         return acceptImpl(authenticated);
       } else {
-        // TODO(perf):  As a hack for the 0.4 release we are always setting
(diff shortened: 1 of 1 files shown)


kentonv (approved):
Yeah... looking at it... I don't know why it isn't like that already. Thanks!

(Will merge when CI passes, ping again if I forget.)

ViniciusCestarii:
A CI test timed out. I believe it is unrelated to the changes here.

kentonv:
Yep, that timeout happens from time to time and is unrelated.

PR #2738, opened 2026-07-23
Report WebSocket overloads with close 1013

apeacock1991:
Report overloaded failures on WebSockets as a 1013 close frame rather than a raw 1006 disconnect


kentonv (approved):
Doesn't quite feel to me like this belongs in the http-over-rpc layer, feels more like KJ exceptions should be converted to HTTP/WebSocket error codes within the HTTP/WebSocket wire protocol layer. However, as discussed, this seems harder due to code structure reasons and it doesn't seem worth spending a lot of time on, so this is fine.

PR #2740, opened 2026-07-24
Initial impl of tryReadSync/tryWriteSync

jasnell:
Adds the implementation of tryReadSync and tryWriteSync 

Some benchmarks in workerd, focusing solely on the kj path (no js involvement)

```
------------------------------------------------------------------------------------------
Benchmark                                Time             CPU   Iterations UserCounters...
------------------------------------------------------------------------------------------
PipeRendezvousWrite_64B_Async          302 us          302 us         2225 bytes_per_second=827.941Mi/s
PipeRendezvousWrite_64B_TrySync        215 us          215 us         3225 bytes_per_second=1.13805Gi/s
PipeRendezvousWrite_4KB_Async         22.2 us         22.2 us        30479 bytes_per_second=44.0779Gi/s
PipeRendezvousWrite_4KB_TrySync       17.5 us         17.5 us        40466 bytes_per_second=55.8927Gi/s
PipeRendezvousRead_64B_Async           316 us          316 us         2373 bytes_per_second=792.345Mi/s
PipeRendezvousRead_64B_TrySync         185 us          185 us         3749 bytes_per_second=1.32261Gi/s
PipeRendezvousRead_4KB_Async          22.2 us         22.2 us        31865 bytes_per_second=43.9467Gi/s
PipeRendezvousRead_4KB_TrySync        15.7 us         15.7 us        44479 bytes_per_second=62.319Gi/s
Pump_1MB_FullAsync                    30.2 us         30.1 us        23217 bytes_per_second=32.3936Gi/s
Pump_1MB_SyncRead                     19.3 us         19.3 us        36582 bytes_per_second=50.6375Gi/s
Pump_1MB_FullSync                     9.61 us         9.60 us        76620 bytes_per_second=101.676Gi/s
GzipDecode_1MB_Async                   200 us          200 us         3485 asyncReads=65 bytes_per_second=4.89249Gi/s syncReads=0
GzipDecode_1MB_TrySync                 204 us          204 us         3554 asyncReads=2 bytes_per_second=4.79358Gi/s syncReads=63
```

So we end up saving roughly 21-80ns per chunk of data on stream workloads that are dominated by the i/o. The Gzip bench highlights that there really is no difference when the stream overhead is mostly the processing. These results are exactly what should be expected.

### DIFF
--- a/c++/src/kj/async-io.c++
+++ b/c++/src/kj/async-io.c++
@@ -592,12 +592,13 @@ private:
       // advance that enough data is buffered to satisfy `minBytes`. (Note that any FDs or
       // streams attached to the blocked write are dropped by a plain byte read, matching
       // tryRead()'s behavior.)
-      size_t totalAvailable = writeBuffer.size();
+      // We only need to confirm that at least minBytes are available, not count the full total.
+      size_t availableChecked = writeBuffer.size();
       for (auto& piece: morePieces) {
-        if (totalAvailable >= minBytes) break;
-        totalAvailable += piece.size();
+        if (availableChecked >= minBytes) break;
+        availableChecked += piece.size();
       }
-      if (totalAvailable < minBytes) {
+      if (availableChecked < minBytes) {
         return kj::none;
       }
 
@@ -1488,6 +1489,7 @@ private:
         return false;
       }
 
+      KJ_ASSERT(pumpedSoFar <= amount);
       if (buffer.size() > amount - pumpedSoFar) {
         // Part of this write would fall past the end of the pump and would have to be delivered
         // to the pipe asynchronously, but tryWriteSync() is all-or-nothing, so we must not start.
@@ -1514,6 +1516,7 @@ private:
         return false;
       }
 
+      KJ_ASSERT(pumpedSoFar <= amount);
       size_t total = 0;
       for (auto& piece: pieces) {
         total += piece.size();
--- a/c++/src/kj/async-io.c++
+++ b/c++/src/kj/async-io.c++
@@ -53,16 +53,29 @@
 namespace kj {
 
 Promise<size_t> AsyncInputStream::read(ArrayPtr<byte> buffer, size_t minBytes) {
-  return tryRead(buffer.begin(), minBytes, buffer.size()).then([=](size_t result) mutable {
-    if (result >= minBytes) {
-      return result;
-    } else {
-      kj::throwRecoverableException(KJ_EXCEPTION(DISCONNECTED, "stream disconnected prematurely"));
-      // Pretend we read zeros from the input.
-      buffer.first(minBytes).slice(result).fill(0);
-      return minBytes;
-    }
-  });
+  size_t result;
+
+  // Fast path: complete the read synchronously if the data is already available, avoiding
+  // promise overhead.
+  KJ_IF_SOME(syncResult, tryReadSync(buffer, minBytes)) {
+    result = syncResult;
+  } else {
+    result = co_await tryRead(buffer.begin(), minBytes, buffer.size());
+  }
+
+  if (result >= minBytes) {
+    co_return result;
+  }
+
+  // Short read indicates EOF.
+  kj::throwRecoverableException(KJ_EXCEPTION(DISCONNECTED, "stream disconnected prematurely"));
+  // Pretend we read zeros from the input.
+  buffer.first(minBytes).slice(result).fill(0);
+  co_return minBytes;
+}
+
+Maybe<size_t> AsyncInputStream::tryReadSync(ArrayPtr<byte> buffer, size_t minBytes) {
+  return kj::none;
 }
 
 Maybe<uint64_t> AsyncInputStream::tryGetLength() { return kj::none; }
@@ -79,6 +92,9 @@ Maybe<Own<AsyncInputStream>> AsyncInputStream::tryTee(uint64_t) {
 kj::Promise<size_t> NullStream::tryRead(void* buffer, size_t minBytes, size_t maxBytes) {
   return kj::constPromise<size_t, 0>();
 }
+kj::Maybe<size_t> NullStream::tryReadSync(ArrayPtr<byte> buffer, size_t minBytes) {
+  return size_t(0);
+}
 kj::Maybe<uint64_t> NullStream::tryGetLength() {
   return uint64_t(0);
 }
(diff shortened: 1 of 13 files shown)


mikea:
I think this is promising, but my worry is that current approach requires rewriting a lot of code to reap benefits (and some of that code will be complicated). I wonder if we can speed things up the "easier" way.

Is the main benefit from this change reading from buffers? Do we already return immediately resolved promise there?

jasnell:
> ... can speed things up the "easier" way.

Which way?

This yields significant improvements for both read and write paths. With the correlating changes in workerd, I'm currently able to realize a 6x improvement in benchmarks and there's still plenty of optimization left on the table still.

jasnell:
A couple additional points here:

If we went with an alternative approach of trying to conditionally extract the result of a settled promise, it becomes too easy to shoot ourselves in the foot and accidentally break the optimization.

Let's imagine the following:

```cpp
kj::Promise<size_t> tryRead(void* buffer, size_t min, size_t max) {
  auto size = co_await inner->tryRead(buffer, min, max);
  co_return size;
}
```

We could, theoretically, make the coroutine capable of collapsing the suspension so that the outer promise resolves synchronously/immediately if the inner promise does; at least with some amount of effort. But someone, seeing that this is a coroutine, doesn't realize that that optimization is in place comes along and adds an additional co_await, then suddenly the optimization is broken. Worse, it's broken silently without the ability to easily detect it.

By going with a `tryReadSync` approach, we are specifically requiring the developer to take either the sync or the async paths explicitly. This becomes much more reliable for us to determine if the sync optimization is actually happening.

The second point is that this optimization is specifically geared towards allowing us to perform synchronous reads/writes in workerd streams *while* we are still holding the isolate lock. The current async path means we have to drop then reacquire the lock on every read/write. With `tryReadSync` and `tryWriteSync`, we can attempt the operation while we're holding the lock, then only fallback to the async path dropping the lock when the sync path is not possible. Best case, we can perform the entire pump while we're holding the lock. Worst case, we fallback to the current behavior.

So while this is a larger change, I do think it's well worth it.

jasnell:
@harrishancock @mikea @danlapid .. still waiting on review for this one.

harrishancock:
Ack, gimme a day.

c++/src/kj/async-io.h
```diff
@@ -64,6 +64,36 @@ class AsyncInputStream: private AsyncObject {
...
+  // completed synchronously, in which case the caller should fall back to tryRead().
+  //
+  // Semantics match tryRead(): at least `minBytes` are read, `buffer.size()` acts as the maximum,
+  // and a return value less than `minBytes` indicates EOF. `minBytes` must not exceed
+  // `buffer.size()`.
+  //
+  // tryReadSync() returns kj::none -- and MUST NOT throw -- whenever a synchronous read is not
+  // possible, for example:
+  // - Answering the read would require waiting for asynchronous I/O.
+  // - An asynchronous operation (tryRead() or pumpTo()) is already in progress on this stream.
```
harrishancock: I think this should be considered a programming error, so I'd expect throwing would be a reasonable thing to do.

c++/src/kj/async-io.h
```diff
@@ -122,6 +152,41 @@ class AsyncOutputStream: private AsyncObject {
...
+  // Returns false if the write cannot be completed synchronously, in which case the caller should
+  // fall back to write().
+  //
+  // This is all-or-nothing: either the entire buffer is written synchronously (returns true), or
+  // nothing is written at all (returns false). There is no partial-write case.
+  //
+  // tryWriteSync() returns false -- and MUST NOT throw -- whenever a synchronous write is not
+  // possible, for example:
+  // - The underlying transport would need to perform asynchronous I/O.
+  // - An asynchronous operation (write() or tryPumpFrom()) is already in progress on this stream.
```
harrishancock: Ditto

c++/src/kj/async-io.h
```diff
@@ -122,6 +152,41 @@ class AsyncOutputStream: private AsyncObject {
...
+  // - The stream is in an errored or shut-down state. (The caller's subsequent write() will
```
harrishancock: Why this behavior? I'd think throwing would be reasonable, if the error state is already known.

c++/src/kj/async-io.c++
```diff
@@ -109,6 +122,35 @@ public:
...
+      if (n == 0) return doneSoFar;
+
+      KJ_IF_SOME(amount, input.tryReadSync(arrayPtr(buffer).first(n), 1)) {
+        if (amount == 0) return doneSoFar;  // EOF
+        doneSoFar += amount;
+        if (!output.tryWriteSync(arrayPtr(buffer).first(amount))) {
+          // The output needs to write asynchronously.
+          return output.write(arrayPtr(buffer).first(amount)).then([this]() {
+            return pump();
+          });
```
harrishancock: Non-blocking nit: slight preference for this to be a coroutine. I guess that would mean two nested for loops, though, and possibly inlining `pumpAsync()`, so maybe it's not worth it. Up to you.

c++/src/kj/async-io.c++
```diff
@@ -136,7 +182,9 @@ Promise<uint64_t> unoptimizedPumpTo(
     AsyncInputStream& input, AsyncOutputStream& output, uint64_t amount,
     uint64_t completedSoFar) {
   auto pump = heap<AsyncPump>(input, output, amount, completedSoFar);
-  auto promise = pump->pump();
+  // Use evalNow() to capture any exception thrown by a synchronous read or write into the
+  // returned promise: historically this function never threw synchronously.
```
harrishancock: If it were a coroutine, you'd get this behavior for free.

c++/src/kj/async-io.c++
```diff
@@ -500,6 +595,38 @@ private:
...
+        return kj::none;
+      }
+
+      // tryReadImpl() cannot stop halfway: if it consumes the entire blocked write without
+      // reaching `minBytes`, it completes the write and expects the caller to continue reading
+      // from the pipe asynchronously. A synchronous read cannot do that, so we must verify in
+      // advance that enough data is buffered to satisfy `minBytes`. (Note that any FDs or
+      // streams attached to the blocked write are dropped by a plain byte read, matching
+      // tryRead()'s behavior.)
+      size_t totalAvailable = writeBuffer.size();
```
harrishancock: Non-blocking nit: this early-bailout on line 612 means `totalAvailable` isn't an accurate name. Worth a comment, and/or perhaps wrapping the block in `{ ... }` so the variable goes out of scope.

c++/src/kj/async-io.c++
```diff
@@ -1276,6 +1497,62 @@ private:
       }, teeExceptionPromise<void>(fulfiller, canceler)));
     }
 
+    bool tryWriteSync(ArrayPtr<const byte> buffer) override {
+      if (!canceler.isEmpty()) {
+        // A write to the pump's output is already in progress.
+        return false;
+      }
+
+      if (buffer.size() > amount - pumpedSoFar) {
```
harrishancock, line 1493: Perhaps this assert should go above this line?
```
        KJ_ASSERT(pumpedSoFar <= amount);
```

c++/src/kj/async-io.c++
```diff
@@ -1276,6 +1497,62 @@ private:
...
+      if (!canceler.isEmpty()) {
+        // A write to the pump's output is already in progress.
+        return false;
+      }
+
+      size_t total = 0;
+      for (auto& piece: pieces) {
+        total += piece.size();
+      }
+      if (total > amount - pumpedSoFar) {
```
harrishancock, line 1524: Same comment re: assertion

c++/src/kj/async-io.c++
```diff
@@ -1689,6 +2011,20 @@ public:
     });
   }
 
+  Maybe<size_t> tryReadSync(ArrayPtr<byte> buffer, size_t minBytes) override {
+    if (limit == 0) return size_t(0);
+    KJ_IF_SOME(actual, inner->tryReadSync(
+        buffer.first(kj::min(buffer.size(), limit)), kj::min(minBytes, limit))) {
```
harrishancock: Nit: this is kind of hard to read. Maybe:

```cpp
// Clamp arguments to the limit.
buffer = buffer.first(kj::min(buffer.size(), limit));
minBytes = kj::min(minBytes, limit);
KJ_IF_SOME(actual, inner->tryReadSync, buffer, minBytes) {
  ...
}
```

c++/src/kj/compat/http.c++
```diff
@@ -2512,13 +2524,28 @@ public:
...
     // want to block further writes.
     writeInProgress = false;
   }
 
+  bool tryWriteBodyDataSync(ArrayPtr<const byte> buffer) {
+    // Synchronous counterpart to writeBodyData(): attempt to write the data via the inner
+    // stream's tryWriteSync(). Returns false if a write is already in progress, if queued
+    // writes (e.g. headers or chunk boundaries) have not been observed to drain yet, or if the
+    // inner stream cannot accept the data synchronously.
+    if (writeInProgress || !inBody || !writeQueueDrained) return false;
```
harrishancock: Nit: maybe better?

```cpp
    if (!canWriteBodyData() || !writeQueueDrained) return false;
```

c++/src/kj/compat/http.c++
```diff
@@ -2512,13 +2524,28 @@ public:
...
+    return inner.tryWriteSync(buffer);
+  }
+
+  bool tryWriteBodyDataSync(ArrayPtr<const ArrayPtr<const byte>> pieces) {
+    if (writeInProgress || !inBody || !writeQueueDrained) return false;
```
harrishancock: ditto

c++/src/kj/compat/http.c++
```diff
@@ -2587,6 +2617,12 @@ private:
   // a write throws an exception or is canceled, this remains true forever. In these cases, the
   // underlying stream is in an inconsistent state and cannot be reused.
 
+  bool writeQueueDrained = true;
+  // True if `writeQueue` is known to have fully drained (i.e. it is safe to write body data
+  // directly to `inner` synchronously). We cannot inspect a promise's readiness synchronously,
+  // so this is tracked explicitly: queueWrite() clears it, and the body-write methods set it
+  // after their `co_await` of the queue completes.
```
harrishancock, line 2628: Non-blocking nit: I wish this class had a proper state machine. I realize it's probably not realistic to demand that that happen in this PR, though. :/

c++/src/kj/compat/brotli.c++
```diff
@@ -263,7 +263,69 @@ BrotliAsyncInputStream::~BrotliAsyncInputStream() noexcept(false) {
 Promise<size_t> BrotliAsyncInputStream::tryRead(void* out, size_t minBytes, size_t maxBytes) {
   if (maxBytes == 0) return constPromise<size_t, 0>();
 
-  return readImpl(reinterpret_cast<byte*>(out), minBytes, maxBytes, 0);
+  readInProgress = true;
+  KJ_ON_SCOPE_FAILURE(readInProgress = false);
```
harrishancock: In other streams implementations, if a `tryRead()` fails, we consider the stream broken, and that it is an error to try reading again. This seems to allow that -- though I suppose the `tryReadSync()` caller would just try the async version in that case, and then see the broken state?

Separately, this looks like it could be simplified as a coroutine.

```cpp
if (maxBytes == 0) co_return 0;
readInProgress = true;
KJ_DEFER(readInProgress = false);
co_return co_await readImpl(...);
```

c++/src/kj/compat/gzip.c++
```diff
@@ -191,7 +191,63 @@ GzipAsyncInputStream::~GzipAsyncInputStream() noexcept(false) {
 Promise<size_t> GzipAsyncInputStream::tryRead(void* out, size_t minBytes, size_t maxBytes) {
   if (maxBytes == 0) return constPromise<size_t, 0>();
 
-  return readImpl(reinterpret_cast<byte*>(out), minBytes, maxBytes, 0);
+  readInProgress = true;
+  KJ_ON_SCOPE_FAILURE(readInProgress = false);
```
harrishancock: same comment as in Brotli

c++/src/kj/async-io.c++
```diff
@@ -62,7 +62,20 @@ Promise<size_t> AsyncInputStream::read(ArrayPtr<byte> buffer, size_t minBytes) {
...
+  };
+
+  // Fast path: complete the read synchronously if the data is already available, avoiding
+  // promise overhead.
+  KJ_IF_SOME(result, tryReadSync(buffer, minBytes)) {
+    if (result >= minBytes) {
+      return result;
+    }
+    // Short read indicates EOF. Run the same handling as the async path, using evalNow() so that
+    // the thrown exception becomes a rejected promise, just as it would have been before.
```
harrishancock: This might be simplifiable as a coroutine.

harrishancock (approved):

jasnell:
this ended up needing a significant rebase along with workerd and the internal repo so I ended up squashing. Otherwise things would have gotten much more difficult to handle.

harrishancock:
I took another look, and it looks like some of my comments were thumbed-up but not addressed. Not all of them need to be, but you might want to do a second pass and make sure any that were useful to you are accounted for. Otherwise LGTM.

jasnell:
Ah crap, I forgot to fill in the responses on all of those and now I can't remember the context 😆 ... let me go back and check them

c++/src/kj/compat/http.c++
jasnell: Took a different approach on this one with KJ_REQUIRE that I think is better

c++/src/kj/compat/http.c++
jasnell: Took a different approach on this one with KJ_REQUIRE that I think is better

c++/src/kj/compat/http.c++
jasnell: Agreed... something we can revisit

c++/src/kj/compat/brotli.c++
jasnell: On the second part, yes, the tryReadSync caller would just try the async version and see the broken state. Intentional.

jasnell:
@harrishancock ... applied one of the fixes I missed and added comments to a couple others

jasnell:
Ok, because the API is introduced with `tryWriteSync` and `tryReadSync` as virtuals with a default no-op, we can safely land this now without lock step PRs in workerd and the production repo. I'm going to go ahead and do so, then will separately bump the versions in the other two repos, then will do a separate lockstep PR pair between those two to get things going

jasnell:
Ugh... my appologies for forgetting about the fixup commit that was still in here... that needed to be squashed :-/

PR #2743, opened 2026-07-24
Add MKCALENDAR to recognized HTTP methods

cnharrison:
`MKCALENDAR` is CalDAV’s standard calendar-creation method (RFC 4791 §5.3.1).

Add it to KJ’s recognized HTTP-method set alongside existing WebDAV methods such as `MKCOL`, `PROPFIND`, and `REPORT`.

This unblocks cloudflare/workerd#6877, where workerd currently rejects `MKCALENDAR` before Worker code runs.

### DIFF
--- a/c++/src/kj/compat/http.h
+++ b/c++/src/kj/compat/http.h
@@ -85,7 +85,8 @@ namespace kj {
   MACRO(QUERY) \
   /* https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-05.html */ \
   MACRO(BAN) \
-  /* Non-standard method name requested by a Cloudflare customer. */
+  /* Non-standard method name requested by a Cloudflare customer. */ \
+  MACRO(MKCALENDAR)
 
 enum class HttpMethod {
   // Enum of known HTTP methods.
--- a/c++/src/capnp/compat/http-over-capnp.capnp
+++ b/c++/src/capnp/compat/http-over-capnp.capnp
@@ -187,6 +187,7 @@ enum HttpMethod {
 
   query @26;
   ban @27;
+  mkcalendar @28;
 }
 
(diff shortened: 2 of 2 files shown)


c++/src/kj/compat/http.h
```diff
@@ -85,7 +85,8 @@ namespace kj {
   MACRO(QUERY) \
   /* https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-05.html */ \
   MACRO(BAN) \
```
harrishancock, line 87: Pre-existing nit
```suggestion
  \
  MACRO(BAN) \
```

c++/src/kj/compat/http.h
```diff
@@ -85,7 +85,8 @@ namespace kj {
...
-  /* Non-standard method name requested by a Cloudflare customer. */
+  /* Non-standard method name requested by a Cloudflare customer. */ \
+  MACRO(MKCALENDAR)
```
harrishancock, line 90:
```suggestion
  \
  MACRO(MKCALENDAR) \
  /* https://www.rfc-editor.org/rfc/rfc4791#section-5.3.1 (CalDAV) */
```

harrishancock (commented):
Check out https://github.com/capnproto/capnproto/pull/2379/changes to see the other places that need modification:

- `validateMethod()` in http-over-capnp.c++
- `consumeHttpMethod()` in http.c++

For the latter, `case 'C': EXPECT_REST(MKC,OL)` will probably need to become:

```cpp
            case 'C':
              switch (*p++) {
                case 'A': EXPECT_REST(MKCA,LENDAR)
                case 'O': EXPECT_REST(MKCO,L)
                default: return kj::none;
              }
```

Once that's in, there are a couple places in workerd that'll need updating, too, but that can come after.

PR #2746, opened 2026-07-25
make AsyncTee::Buffer::size() O(1)

danlapid:
pullLoop() checks each branch's buffered size against bufferSizeLimit every iteration, so summing the deque in size() made teeing N blocks O(N^2) and could stall a thread for long stretches. Track the total in produce()/consume()/ asArray() instead. Resolves the existing TODO(perf) at the call site.

### DIFF
--- a/c++/src/kj/async-io.c++
+++ b/c++/src/kj/async-io.c++
@@ -1771,11 +1771,7 @@ class AsyncTee final: public Refcounted {
     uint64_t size() const;
 
     Buffer clone() const {
-      size_t size = 0;
-      for (const auto& buf: bufferList) {
-        size += buf.size();
-      }
-      auto builder = heapArrayBuilder<byte>(size);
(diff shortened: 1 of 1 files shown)


c++/src/kj/async-io.c++
```diff
@@ -2296,14 +2299,18 @@ uint64_t AsyncTee::Buffer::consume(ArrayPtr<byte>& readBuffer, size_t& minBytes)
       bufferList.pop_front();
     } else {
       bytes = heapArray(bytes.slice(amount, bytes.size()));
-      return totalAmount;
+      break;
     }
   }
 
+  KJ_DASSERT(totalSize >= totalAmount);
```
mikea: Why dassert?

mikea (approved):
Lg. Only asserts can be kj_assert

danlapid:
> Lg. Only asserts can be kj_assert

To not waste an opcode on the check but I can make assert

PR #2747, opened 2026-07-27
Maybe::assertSome,assertNone utility function

mikea:
We tend to use KJ_ASSERT_NONNULL(x), but arguably x.assertSome() reads better.

I am not suggestion any wide replacements, just a way to write code in a more fluent way.

### DIFF
--- a/c++/src/kj/maybe-test.c++
+++ b/c++/src/kj/maybe-test.c++
@@ -854,12 +854,32 @@ KJ_TEST("Maybe assertSome") {
   const Maybe<int&> constMaybeReference = value;
   KJ_EXPECT(constMaybeReference.assertSome() == value);
 
-#if defined(KJ_DEBUG) || (defined(KJ_ENABLE_IREQUIRE) && KJ_ENABLE_IREQUIRE)
-  Maybe<int> emptyValue = kj::none;
-  KJ_EXPECT_THROW_MESSAGE("null Maybe<> dereference", (void)emptyValue.assertSome());
+  const Maybe<int> constMaybeValue = value;
+  static_assert(isSameType<decltype(kj::mv(maybeValue).assertSome()), int>());
+  static_assert(isSameType<decltype(kj::mv(constMaybeValue).assertSome()), const int&&>());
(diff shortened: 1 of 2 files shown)


harrishancock (approved):
A possible bikeshed color: `.unwrap()`. But `assertSome()` might feel more KJ-y.

kentonv:
FWIW, I intentionally chose not to do it this way because I wanted asserts to look different from regular code.

But idk, maybe it's fine.

kentonv:
You probably need `&&` variants though.

mikea:
> You probably need `&&` variants though.

done. Also added `assertNone` for completeness.

c++/src/kj/maybe-test.c++
```diff
@@ -837,6 +837,44 @@ KJ_TEST("Maybe") {
...
+
+  Maybe<int> emptyValue = kj::none;
+  emptyValue.assertNone();
+  Maybe<int&> emptyReference = kj::none;
+  emptyReference.assertNone();
+
+#if defined(KJ_DEBUG) || (defined(KJ_ENABLE_IREQUIRE) && KJ_ENABLE_IREQUIRE)
+  KJ_EXPECT_THROW_MESSAGE("null Maybe<> dereference", (void)emptyValue.assertSome());
+  KJ_EXPECT_THROW_MESSAGE("null Maybe<> dereference", (void)emptyReference.assertSome());
+  KJ_EXPECT_THROW_MESSAGE("expected Maybe<> to be none", maybeValue.assertNone());
```
harrishancock, line 881: This looks like a behavior change from `KJ_ASSERT_NONNULL()`. Try the following:

```cpp
Maybe<int> m1 = 123;
KJ_EXPECT(KJ_ASSERT_NONNULL(kj::mv(m1)) == 123);
KJ_EXPECT(m1.assertNone());  // I expect OK

Maybe<int> m2 = 456;
KJ_EXPECT(kj::mv(m2).assertSome() == 456);
KJ_EXPECT(m2.assertNone());  // I expect ERROR
```

mikea: great catch! done.

harrishancock (approved):

PR #2749, opened 2026-07-28
Implement interim (1xx) response support

jasnell:
Adds the ability to send and receive informational (1xx) http responses.

Receiving:

```c++
  HttpClient::RequestOptions options;
  // The handler is invoked for every information response
  // received (up to a strict limit)
  options.informationResponseHandler =
      [&](HttpClient::InformationalResponse&& response) {
    switch (response.statusCode) {
      case kj::HttpInformationalStatus::CONTINUE: { /* ... */ }
      case kj::HttpInformationalStatus::PROCESSING: { /* ... */ }
      case kj::HttpInformationalStatus::EARLY_HINTS: { /* ... */ }
      default: { /* ... */ }
    }
    values.add(kj::str(KJ_ASSERT_NONNULL(response.headers.get(interimHeader))));
    if (retainedHeaders == kj::none) retainedHeaders = response.headers.clone();
  };

  auto req = client->request(
      HttpMethod::GET, "/", HttpHeaders(*table), kj::mv(options));
  // ...
```

Sending:

```c++
// In the HttpService (server side)
// ...
kj::Promise<void> request(
    HttpMethod method, kj::StringPtr url, const HttpHeaders& headers,
    kj::AsyncInputStream& requestBody, Response& response) override {
  HttpHeaders informational(table);
  informational.setPtr(infoHeader, "value");
  response.sendInformational(kj::HttpInformationalStatus::EARLY_HINTS,
      "Early Hints", informational);
  // Can call sendInformational multiple times...
  response.send(204, "No Content", HttpHeaders(table), uint64_t(0));
  return kj::READY_NOW;
}
```

http-over-capnp support is included.

The existing handling of 101 responses for protocol upgrade is preserved.

Optional automatic server-side handling of 100-continue responses is included, but off by default.

### DIFF
--- a/c++/src/kj/compat/http.h
+++ b/c++/src/kj/compat/http.h
@@ -98,6 +98,17 @@ KJ_HTTP_FOR_EACH_METHOD(DECLARE_METHOD)
 #undef DECLARE_METHOD
 };
 
+enum class HttpInformationalStatus {
+  // We intentionally only support a subset of the 1xx informational status code.
+  // Because we don't know what future semantics will be added to future 1xx
+  // codes, we don't support them. They will be ignored when received and we
+  // won't allow sending them. The ones we do support here are known to be
+  // safe to handle at either the kj or application level.
+  CONTINUE = 100,
+  PROCESSING = 102,
+  EARLY_HINTS = 103,
+};
+
 struct HttpConnectMethod {};
 // CONNECT is handled specially and separately from the other HttpMethods.
 
@@ -829,7 +840,7 @@ class HttpClient {
 
 public:
   struct InformationalResponse {
-    uint statusCode;
+    HttpInformationalStatus statusCode;
     kj::StringPtr statusText;
     const HttpHeaders& headers;
   };
@@ -960,7 +971,7 @@ public:
     //
     // `send()` may only be called a single time. Calling it a second time will cause an exception
     // to be thrown.
-    // `statusCode` must be 200 or greater; use sendInformational() for non-101 1xx responses.
+    // `statusCode` must be 200 or greater; use sendInformational() for informational responses.
 
     virtual kj::Own<WebSocket> acceptWebSocket(const HttpHeaders& headers) = 0;
     // If headers.isWebSocket() is true then you can call acceptWebSocket() instead of send().
@@ -976,8 +987,8 @@ public:
     // exception to be thrown.
 
     virtual void sendInformational(
-        uint statusCode, kj::StringPtr statusText, const HttpHeaders& headers);
-    // Sends a non-101 informational response before send() or acceptWebSocket().
+        HttpInformationalStatus statusCode, kj::StringPtr statusText, const HttpHeaders& headers);
+    // Sends a supported informational response before send() or acceptWebSocket().
 
     kj::Promise<void> sendError(uint statusCode, kj::StringPtr statusText,
                                 const HttpHeaders& headers);
--- a/c++/src/kj/compat/http.h
+++ b/c++/src/kj/compat/http.h
@@ -167,7 +167,8 @@ public:
   MACRO(LOCATION, "Location") \
   MACRO(CONTENT_TYPE, "Content-Type") \
   MACRO(RANGE, "Range") \
-  MACRO(CONTENT_RANGE, "Content-Range")
+  MACRO(CONTENT_RANGE, "Content-Range") \
+  MACRO(EXPECT, "Expect")
   // For convenience, these headers are valid for all HttpHeaderTables. You can refer to them like:
   //
   //     HttpHeaderId::HOST
@@ -386,9 +387,11 @@ public:
   struct Request {
     HttpMethod method;
     kj::StringPtr url;
+    bool isHttp10 = false;
   };
   struct ConnectRequest {
     kj::StringPtr authority;
+    bool isHttp10 = false;
   };
   struct Response {
     uint statusCode;
@@ -825,6 +828,17 @@ class HttpClient {
   //   The `url` specified in a request is a full URL including protocol and hostname.
 
 public:
+  struct InformationalResponse {
+    uint statusCode;
+    kj::StringPtr statusText;
+    const HttpHeaders& headers;
+  };
+
+  struct RequestOptions {
+    kj::Maybe<uint64_t> expectedBodySize = kj::none;
+    kj::Maybe<kj::Function<void(InformationalResponse&&)>> informationResponseHandler = kj::none;
+  };
+
   struct Response {
     uint statusCode;
     kj::StringPtr statusText;
@@ -857,6 +871,11 @@ public:
   // `expectedBodySize`, if provided, must be exactly the number of bytes that will be written to
   // the body. This will trigger use of the `Content-Length` connection header. Otherwise,
   // `Transfer-Encoding: chunked` will be used.
+
+  virtual Request request(HttpMethod method, kj::StringPtr url, const HttpHeaders& headers,
+                          RequestOptions options);
+  // Performs a request and synchronously reports informational responses before the final response.
+  // The handler must not retain `statusText` or `headers` without copying them.
 
   struct WebSocketResponse {
     uint statusCode;
@@ -941,6 +960,7 @@ public:
     //
     // `send()` may only be called a single time. Calling it a second time will cause an exception
     // to be thrown.
+    // `statusCode` must be 200 or greater; use sendInformational() for non-101 1xx responses.
 
     virtual kj::Own<WebSocket> acceptWebSocket(const HttpHeaders& headers) = 0;
     // If headers.isWebSocket() is true then you can call acceptWebSocket() instead of send().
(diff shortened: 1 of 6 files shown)


kentonv:
I'm not comfortable adding a general 1xx handler. Each 1xx code seems to have distinct semantics that affect protocol state in a way that I feel cannot be safely handled at the application layer. I think for each code, we either need to handle it explicitly in KJ HTTP itself or we need to reject it outright (break the connection).

FWIW #2671 was another recent attempt at this.

jasnell:
> ... Each 1xx code seems to have distinct semantics that affect protocol state ...

There are only four defined standard 1xx codes.

101 is really the only one that carries any sort of connection-altering semantics and that's handled specifically (by being excluded) here.

100-continue can be handled automatically or by application easily and safely.

102-processing and 103-early-hints both carry no connection-altering semantics. There are no other standardized 1xx codes in common use.

Given that, perhaps instead of allowing the full range of 1xx, would you be more comfortable strictly limiting this to 100, 102, and 103? These are known to be safely handle-able at the application layer, with optional automatic handling of 100.

jasnell:
I've updated the implementation to support only 100, 102, and 103 responses using an enum approach.

danlapid:
Seems like we agree it is a generally good idea, since Kenton's concerns were taken into account in the PR, maybe @mikea + @harrishancock can help with the review?

kentonv:
Hmm, I feel like `100 Continue` is significantly different from 102 or 103 and might need special handling. I'm not 100% sure about this. But `100 Continue` is part of a two-way handshake (the client requests it, then the server returns it, as an instruction to the client to proceed with sending the request body)... seems more complicated, especially the implications for proxies.

Perhaps we could limit to 102 and 103?

c++/src/kj/compat/http.h
```diff
@@ -98,6 +98,17 @@ KJ_HTTP_FOR_EACH_METHOD(DECLARE_METHOD)
 #undef DECLARE_METHOD
 };
 
+enum class HttpInformationalStatus {
+  // We intentionally only support a subset of the 1xx informational status code.
+  // Because we don't know what future semantics will be added to future 1xx
+  // codes, we don't support them. They will be ignored when received and we
```
kentonv, line 104: Unknown codes should be an error, not ignored.

jasnell: That is not valid per the spec. Specifically, it directly contradicts two MUST-level requirements in RFC 9110:

1. Section 15.2 — 1xx-specific rule

> A client MUST be able to parse one or more 1xx responses received prior to a final response, even if the client does not expect one. A user agent MAY ignore unexpected 1xx responses.

The rules:

* `MUST` parse them → erroring/breaking the connection violates this.
* `MAY` ignore them → the spec explicitly grants permission to discard unrecognized 1xx responses.

This PR's "they will be ignored" comment is the spec-compliant behavior.

2. Section 15 — General unrecognized status code rule
 
HTTP status codes are extensible. A client is not required to understand the meaning of all registered status codes.

However, a client MUST understand the class of any status code, as indicated by the first digit, and treat an unrecognized status code as being equivalent to the x00 status code of that class.

For an unrecognized 1xx, the correct fallback is to treat it like a 100, which per Section 15.2.1 means the client can simply discard this interim response. So even the general "fallback to x00" rule leads to "discard it," not "error."

The enum here prevents us from *send* any code other than 100, 102, and 103. *Receiving* any other code, per the spec should be ignored, not be an error.

jasnell:
> ... I feel like 100 Continue is significantly different from 102 or 103 and might need special handling ...

What additional special handling would 100 need that this PR does not already provide?

In this PR, 100 codes have `autoHandle100Continue`, `OnFirstReadInputStream`, `auto100ContinuePending`, `is100Continue()` parsing, `HTTP/1.0` suppression, etc. The adapters in this PR also handle the proxy case correctly also. The other codes do not use these things.

It's true that 100 needs special behavior but that's already provided by this PR. 

While the APIs and some of the expression is different, the model implemented here is not disimmilar to the model Node.js has used and proven to work for nearly a decade now. 

I do *not* want to drop 100 handling from this PR. Adding 100 support is actually a key motivator as the lack of 100 handling in kj and workerd is leading to severe performance penalties when some HTTP clients (like curl) are used to upload data to production workers.

c++/src/kj/compat/http.h
jasnell: Note that there's also the requirement that a Proxy MUST forward unrecognized 1xx codes. Throwing an error when we encounter one violates the spec. The behavior implemented by this PR is actually stricter than what the spec requires. In a proxy case, our proxy application code is only going to receive the known 1xx codes and can only forward those on. So this PR is already locked down more than the spec and *actually* violates a the Proxy MUST forward requirement.

jasnell:
To break this down further... here's an example of the server side with automatic 100 handling in this PR:

```cpp
class UploadService final: public kj::HttpService {
public:
  UploadService(kj::HttpHeaderTable& table): table(table) {}

  kj::Promise<void> request(
      kj::HttpMethod method, kj::StringPtr url, const kj::HttpHeaders& headers,
      kj::AsyncInputStream& requestBody, Response& response) override {
    // Just read the body. The framework will send "100 Continue" the
    // moment we call readAllBytes(), before any data arrives.
    // This is correct as we don't want the client to send data
    // until we're ready to handle it, which is the point of 100
    // to begin with.
    auto body = co_await requestBody.readAllBytes();

    // Process body...
    response.send(204, "No Content", kj::HttpHeaders(table), uint64_t(0));
  }

private:
  kj::HttpHeaderTable& table;
};

// Wire it up with autoHandle100Continue enabled:
kj::HttpServer server(timer, table, service, {
  .autoHandle100Continue = true,
});
```

If someone wanted to manually handle it, they could. Node.js also allows this pattern (as do other runtimes). I suspect tho, that `autoHandle100Continue = true` would be our default.

```cpp
class GatedUploadService final: public kj::HttpService {
public:
  GatedUploadService(kj::HttpHeaderTable& table, kj::HttpHeaderId authHeader)
      : table(table), authHeader(authHeader) {}

  kj::Promise<void> request(
      kj::HttpMethod method, kj::StringPtr url, const kj::HttpHeaders& headers,
      kj::AsyncInputStream& requestBody, Response& response) override {
    // Check auth before telling the client to send the body.
    if (headers.get(authHeader) == kj::none) {
      // Reject — the client never sends the body, saving bandwidth.
      // Equivalent to Node.js sending 401 inside 'checkContinue'.
      response.send(401, "Unauthorized", kj::HttpHeaders(table), uint64_t(0));
      co_return;
    }

    auto contentLength = requestBody.tryGetLength().orDefault(0);
    if (contentLength > MAX_UPLOAD) {
      response.send(413, "Payload Too Large", kj::HttpHeaders(table), uint64_t(0));
      co_return;
    }

    // Headers look good — tell the client to proceed.
    // Equivalent to calling res.writeContinue() in Node.js's 'checkContinue'.
    response.sendInformational(
        kj::HttpInformationalStatus::CONTINUE, "Continue", kj::HttpHeaders(table));

    // Now read the body.
    auto body = co_await requestBody.readAllBytes();
    response.send(204, "No Content", kj::HttpHeaders(table), uint64_t(0));
  }

private:
  kj::HttpHeaderTable& table;
  kj::HttpHeaderId authHeader;
  static constexpr uint64_t MAX_UPLOAD = 10 * 1024 * 1024;
};
```

On the client side:

```cpp
kj::HttpHeaders headers(table);
headers.setPtr(kj::HttpHeaderId::EXPECT, "100-continue");
headers.setPtr(kj::HttpHeaderId::CONTENT_LENGTH, "1048576");

kj::Own<kj::AsyncOutputStream> requestBody;
kj::Maybe<kj::Promise<void>> upload;

kj::HttpClient::RequestOptions options;
options.expectedBodySize = uint64_t(1048576);
options.informationResponseHandler =
    [&](kj::HttpClient::InformationalResponse&& response) {
  switch (response.statusCode) {
    case kj::HttpInformationalStatus::CONTINUE:
      // Server said go ahead — start writing the body.
      // Equivalent to Node.js's req.on('continue', () => { ... })
      upload = requestBody->write(hugeBuffer);
      break;
    default:
      // 102 or 103 — note it, but no protocol action needed.
      break;
  }
};

auto request = client->request(
    kj::HttpMethod::PUT, "/upload", headers, kj::mv(options));
requestBody = kj::mv(request.body);

// Wait for the final response. The informationResponseHandler fires
// synchronously during header parsing, before this promise resolves.
auto response = request.response.wait(waitScope);

if (response.statusCode == 401 || response.statusCode == 413) {
  // Server rejected us before we sent the body — bandwidth saved.
} else {
  // Server accepted. Finish the upload.
  KJ_ASSERT_NONNULL(upload).wait(waitScope);
}
```

This, again, is an established pattern. Nothing all that special is necessary and any special handling that is required is largely internal.

What am I missing?

isker:
A purely self-interested question: is this leading to 103 support in workerd?

jasnell:
> A purely self-interested question: is this leading to 103 support in workerd?

That's part of it.

jasnell:
@kentonv ... still looking to see if we can get this resolved ...

PR #2750, opened 2026-07-29
make WeakRc more thread-safe

mikea:
Move/Clone/Delete are allowed to be performed on different threads. This makes WeakRc much more versatile and safer to use.


harrishancock:
I don't think we should merge this unless we can articulate a real use case.

PR #2752, opened 2026-07-30
Experimental change for kj refcounts to avoid double deletion

erikcorry:
Part of a bug analysis.


erikcorry:
Pretty much the same as https://github.com/capnproto/capnproto/pull/2751

PR #2753, opened 2026-07-30
Rc::downcast should not consume this

mikea:
I think current implementation was a mistake: while it is "eficcient" it is very error prone.


kentonv:
How about having an overload with a `&&` qualifier that still consumes?

PR #2754, opened 2026-07-31
Report the correct fiber stack bounds to ASAN

logan-gatlin:
`FiberStack::Impl` lives at the *top* of the mmap()ed stack area, but that address was
being passed to `__sanitizer_start_switch_fiber()` as if it were the base. ASAN therefore
believed the fiber stack spanned `[Impl, Impl + stackSize - sizeof(Impl))`, a range that is
almost entirely outside the real mapping and ends `stackSize - 2 * sizeof(Impl)` bytes past
the end of it.

This matters because `__asan_handle_no_return()` unpoisons everything from the current stack
pointer up to the recorded stack top. The compiler emits a call to it ahead of every call to a
noreturn function (so, ahead of every `throw`), and ASAN's `longjmp`/`_longjmp` interceptors
call it too (so, on every fiber switch). With the bad bounds, each of those calls cleared the
ASAN poison from up to `stackSize` bytes of whichever unrelated mapping followed the fiber
stack.

When the victim was a V8 cppgc page, the next full GC aborted the process in
`cppgc::internal::CheckMemoryIsInaccessible()`, which requires free-list entries to stay
poisoned between sweeps.

## Changes

- Bounds are now computed in one place, `FiberStack::Impl::asanStackBounds()`, so a caller
  can't pair a correct base with a stale size. All switches onto a fiber go through
  `FiberStack::asanStartSwitchToFiber()`.
- `FiberStack::asanCheckOnFiberStack()` asserts, from the fiber stack itself, that the bounds
  handed to ASAN actually contain it. ASAN accepts bogus bounds silently, so without this
  check nothing catches a layout mistake here. It uses `__builtin_frame_address(0)` rather
  than the address of a local, since with `detect_stack_use_after_return=1` an address-taken
  local may live on ASAN's fake stack.
- Added a regression test that runs fibers at several stack sizes (both endpoints of the
  reported range derive from `stackSize`) and throws on the fiber stack to exercise
  `__asan_handle_no_return()`.
- Both new members are no-ops when not building with ASAN. `StartRoutine::run()` is now
  guarded on `KJ_USE_FIBERS` rather than just the platform, since without fibers `FiberStack`
  has no `impl` member.

Two things are deliberately left inconsistent, and are commented as such:

- On Apple/arm64, `uc_stack.ss_size`/`ss_sp` describe only the last page as a workaround for
  a libc bounds check, but `asanStackBounds()` reports the full mapping. The stack really is
  `stackSize` bytes and the fiber really does grow down through all of it, so ASAN needs to
  know about all of it. Only ucontext is being lied to.
- `switchToMain()` still passes the *original* stack's bounds, not the fiber's, because it is
  switching away from the fiber.

Not verified locally under ASAN; relying on CI.

### DIFF
--- a/c++/src/kj/async-test.c++
+++ b/c++/src/kj/async-test.c++
@@ -1589,6 +1589,47 @@ KJ_TEST("throw from a fiber") {
   KJ_EXPECT_THROW_RECOVERABLE_MESSAGE("test exception", fiber.wait(waitScope));
 }
 
+KJ_TEST("fiber stack bounds reported to ASAN contain the fiber stack") {
+  // Regression test for fiber stacks being described to ASAN starting at the FiberStack::Impl
+  // that lives at the *top* of the stack mapping rather than at the bottom. ASAN silently accepts
+  // whatever bounds it is given, so the only symptom was that __asan_handle_no_return() -- emitted
+  // ahead of every `throw`, and called by ASAN's longjmp interceptor on every fiber switch --
+  // unpoisoned up to a full stackSize worth of whatever mapping happened to follow the stack.
+  //
+  // FiberStack::asanCheckOnFiberStack() does the actual assertion, from the fiber stack, when
+  // built with ASAN; it aborts if the bounds are wrong. All this test has to do is get onto some
+  // fiber stacks and exercise the paths that consume the bounds. Stack sizes other than the 65536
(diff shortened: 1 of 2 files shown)


logan-gatlin:
Fixed by https://github.com/capnproto/capnproto/pull/2803

PR #2755, opened 2026-07-31
keep capabilities alive until their attached FDs have been written

zeebo:
writeDescriptor() reads a capability's descriptor with ClientHook::getFd() and puts the bare number into the kj::Array<int> handed to OutgoingRpcMessage::setFds(). But send() is only required to queue the message: TwoPartyVatNetwork defers the write to kj::evalLast(), and under backpressure the queue can sit for an unbounded time. A descriptor is only guaranteed to stay open as long as the capability that produced it, so a capability dropped in the meantime takes its descriptor with it, and we write a number that has since been closed -- or that the OS has already recycled for an unrelated file, silently handing the peer the wrong one.

This is reachable in practice. A caller that abandons a question before the Return arrives sends a Finish with releaseResultCaps set, which drops the answer's result exports immediately. If such an export held the last reference to an FD-bearing server, that server is destroyed and closes its descriptor while the Return is still sitting in the write queue.

Collect the descriptors alongside a reference to the capability each one came from, and attach those references to the array passed to setFds(), so the capabilities outlive the write.

### DIFF
--- a/c++/src/capnp/capability.h
+++ b/c++/src/capnp/capability.h
@@ -522,6 +522,11 @@ public:
   // If this capability is backed by a file descriptor that is safe to directly expose to clients,
   // returns that FD. When FD passing has been enabled in the RPC layer, this FD may be sent to
   // other processes along with the capability.
+  //
+  // The descriptor must stay open for at least as long as the capability itself. The RPC layer may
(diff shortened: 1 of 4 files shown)


kentonv (approved):

kentonv:
(will merge when CI is done... ping me if I forget)

PR #2756, opened 2026-08-01
Fix `KJ_NO_RTTI` and `KJ_NO_EXCEPTIONS` detection on NetBSD

hebasto:
NetBSD provides a compatibility [stub](https://github.com/NetBSD/src/blob/01bb5fdac48f5ae014ae903602be631c468c5141/sys/sys/cdefs.h#L58-L70) in <sys/cdefs.h> for compilers that lack the `__has_feature` extension:
```c++
#ifndef __has_feature
#define __has_feature(x) 0
#endif
```

Therefore, using `!__has_feature(...)` produces wrong results when using GCC <14.x, where the `__has_feature` extension has not been implemented yet.

This PR fixes the issue.

### DIFF
--- a/c++/src/kj/arena-test.c++
+++ b/c++/src/kj/arena-test.c++
@@ -148,7 +148,7 @@ TEST(Arena, OwnArray) {
   EXPECT_EQ(0, TestObject::count);
 }
 
-#ifndef KJ_NO_EXCEPTIONS
+#if !KJ_NO_EXCEPTIONS
 
 TEST(Arena, ObjectThrow) {
   TestObject::count = 0;
@@ -178,7 +178,7 @@ TEST(Arena, ArrayThrow) {
   EXPECT_EQ(0, TestObject::count);
 }
 
-#endif
+#endif  // !KJ_NO_EXCEPTIONS
 
 TEST(Arena, Alignment) {
   Arena arena;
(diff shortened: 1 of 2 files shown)


kentonv (changes requested):
IMO this is a serious bug in NetBSD. Code is supposed to be able to use `#ifdef __has_feature` to find out if `__has_feature` is supported... NetBSD is incorrectly making it appear to supported when it isn't, and then incorrectly reporting that no actual features are supported even when they are.

This PR seems to work around the problem by no longer trusting `__has_feature(cxx_exceptions)` and `__has_feature(cxx_rtti)` when they return negative values -- only trusting when they return positive. Then we fall back to compiler-specific extensions to detect negative results. But that seems to be a loss on platforms where `__has_feature` is not broken.

Can we instead put in an early check to detect NetBSD specifically, as well as detect if the compiler is a version which does not support __has_feature, and in that case, undef it? That targets the specific platform and compilers where this bug exists and so avoids impact on other platforms.

hebasto:
> Can we instead put in an early check to detect NetBSD specifically, as well as detect if the compiler is a version which does not support __has_feature, and in that case, undef it? That targets the specific platform and compilers where this bug exists and so avoids impact on other platforms.

Thanks! Done.

However, I could imagine a pathological case where a downstream project _always_ expects `__has_feature` to be defined on NetBSD.

kentonv (approved):

PR #2757, opened 2026-08-01
ci: Update actions and images

hebasto:
See individual commits for details.

I will look into FreeBSD tests separately.


hebasto:
> I will look into FreeBSD tests separately.

See https://github.com/capnproto/capnproto/pull/2758.

hebasto:
Rebased on top of the merged https://github.com/capnproto/capnproto/pull/2758.

kentonv:
Wow, looks like this even resolves the FreeBSD 15 kernel bug. Thanks!

PR #2763, opened 2026-08-03
[nfc] Clean up promise continuations

fhanau:
Eliminates a few memory allocations. Based on downstream workerd-promise-ignore-result check.

### DIFF
--- a/c++/src/capnp/dynamic-capability.c++
+++ b/c++/src/capnp/dynamic-capability.c++
@@ -99,6 +99,12 @@ RemotePromise<DynamicStruct> Request<DynamicStruct, DynamicStruct>::send() {
   return RemotePromise<DynamicStruct>(kj::mv(typedPromise), kj::mv(typedPipeline));
 }
 
+kj::Promise<void> Request<DynamicStruct, DynamicStruct>::sendIgnoringResult() {
+  auto typelessPromise = hook->send();
+  hook = nullptr;  // prevent reuse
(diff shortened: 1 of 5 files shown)


c++/src/capnp/dynamic-capability.c++
```diff
@@ -99,6 +99,12 @@ RemotePromise<DynamicStruct> Request<DynamicStruct, DynamicStruct>::send() {
   return RemotePromise<DynamicStruct>(kj::mv(typedPromise), kj::mv(typedPipeline));
 }
 
+kj::Promise<void> Request<DynamicStruct, DynamicStruct>::sendIgnoringResult() {
```
fhanau, line 102: This was added for congruence with and is copied from the non-specialized Request<Params, Results> class, see capability.h. It is being used in json-rpc.c++ now.

c++/src/capnp/dynamic.h
```diff
@@ -560,6 +560,9 @@ class Request<DynamicStruct, DynamicStruct>: public DynamicStruct::Builder {
   RemotePromise<DynamicStruct> send();
   // Send the call and return a promise for the results.
 
+  kj::Promise<void> sendIgnoringResult();
+  // Send the call and return a promise for the results.
```
mikea: Copied comment?

mikea (approved):

c++/src/capnp/dynamic.h
fhanau: good catch!

PR #2218, opened 2025-01-20
ci: Add FreeBSD job to Quick Tests

hebasto:
This PR adds FreeBSD to the CI jobs, inspired by https://github.com/capnproto/capnproto/pull/1796#discussion_r1905782602.

Related comments from https://github.com/capnproto/capnproto/pull/1321:
- https://github.com/capnproto/capnproto/pull/1321#discussion_r698766582:
> Is there really no way to run FreeBSD on GitHub Actions? I was pretty happy that we managed to consolidate our CI a while back...

- https://github.com/capnproto/capnproto/pull/1321#discussion_r698872987:
> but we should have this use `super-test.sh` long-term.

Tested releases:
- Production Release 14.2
- Legacy Release 13.4
- [FreeBSD-CURRENT](https://docs.freebsd.org/en/books/handbook/cutting-edge/#current) snapshot (future 15.0)

Warnings such as the following:
```
clang++: warning: argument unused during compilation: '-pthread' [-Wunused-command-line-argument]
```
have been worked around, but they must be addressed in a follow-up.

Three failing tests has been disabled temporarily.

PR #2765, opened 2026-08-05
[build] Run the tests on FreeBSD

neilpang:
Cirrus CI shut down on 2026-06-01, and .cirrus.yml still named FreeBSD
11.4, 12.2 and 13.0 -- all end of life. This boots a FreeBSD VM under QEMU
on a normal ubuntu-latest runner instead, covering 14.3 and 15.1.

Verified on a fork, on the exact commits proposed here:

https://github.com/neilpang/capnproto/actions/runs/30993106615

```
FreeBSD 14.3   All 3 tests passed   6m17s
FreeBSD 15.1   All 3 tests passed   7m42s
```

### The one FreeBSD-specific fix it needed

`resetCrashHandlers()` ended with `std::set_terminate(nullptr)`. libstdc++
falls back to `abort()` when the handler is null; libcxxrt calls the
pointer. So on FreeBSD a `noexcept` violation jumps to address zero and the
process dies with SIGSEGV instead of SIGABRT:

```
frame #0: 0x0000000000000000
frame #1: libcxxrt.so.1`std::terminate()
frame #2: __clang_call_terminate + 14
frame #3: kj::_::Event::~Event() at async.c++:2191
```

The guard in `~Event()` fires correctly on FreeBSD -- it is only the last
hop that differs, and "Event may not delete itself without permitting
self-destruction" is what notices, because `KJ_EXPECT_SIGNAL(SIGABRT)` then
sees signal 11. Installing a handler that aborts explicitly makes that test
pass on 14.3 and 15.1, and changes nothing on Linux.

I have not checked what the standard says about `set_terminate(nullptr)`, so
I am not claiming either runtime is wrong -- only that they differ.

### Why CXXFLAGS is set explicitly

Worth flagging because it is not FreeBSD-specific. A bare `./configure`
supplies `-O2`, and `common.h:171` turns `KJ_DEBUG` off whenever
`__OPTIMIZE__` is set, which compiles out the `KJ_IREQUIRE` checks that
three tests assert on:

```
common-test.c++:426   KJ_EXPECT_THROW_MESSAGE("Out-of-bounds", head.write(...))
common-test.c++:458   KJ_EXPECT_THROW_MESSAGE("Out-of-bounds", head.write(...))
memory-test.c++:1273  KJ_EXPECT_THROW_MESSAGE("null Weak<> dereference", ...)
```

So `./configure && make check`, which is what the old `.cirrus.yml` did,
fails on Linux too -- I reproduced it there before assuming it was a BSD
problem. `super-test.sh` avoids it by passing `CXXFLAGS="-g"`, and this
workflow does the same.

### Also

Dropping `.cirrus.yml` is a separate hunk; say the word and I will leave it
in place.


kentonv:
Hmm, I don't think the Cirrus CI had actually run on Cap'n Proto in many years. I don't remember the last time I saw it; in fact, I hardly remember it existing at all. I suspect the account set up by the contributor in #1321 expired after a while and we just forgot the config existed.

New FreeBSD CI was set up on the master branch (which is our 1.x release branch) in #2218. This seems to use a similar approach to what you have here.

But for the `v2` branch, I'm not sure we actually want this. The v2 branch is the Cloudflare Workers team's development branch. We intentionally stripped down the CI to stop including a lot of exotic platforms which we didn't need, since before we created the branch, we were getting blocked way too often trying to debug problems on platforms that none of us have easy access to.

If / when we are ready to do a release based on v2, we will definitely bring FreeBSD CI back in (copied over from the master branch). For now, though, the situation is, we aren't able to maintain FreeBSD support on every commit. I will happily accept PRs to fix issues that arise on FreeBSD, but we need to rely on external contributors to submit those asynchronously.

So, sorry, I don't think we can accept this for now.

c++/src/kj/exception.c++
```diff
@@ -925,7 +933,7 @@ void resetCrashHandlers() {
 #endif
 #endif
 
-  std::set_terminate(nullptr);
+  std::set_terminate(&abortHandler);
```
kentonv, line 936: Regarding this fix:

So, apparently, the effects of `set_terminate(nullptr)` are unspecified by the standard.

However, I don't like setting this to a function we define ourselves, either. The default terminate handler does _not_ necessarily simply call `abort()`. Moreover, if the reason we're resetting the handlers is because the KJ library is about to be dynamically unloaded, then setting the handler to another KJ function is clearly wrong.

It seems like the right thing to do would be to record the original terminate handler as returned by the original call to `std::set_terminate()`, then set it back here.

I'd accept a PR implementing that.

PR #2766, opened 2026-08-05
replace non-atomic read in AtomicRefcounted with atomic

mikea:
If delete happens not through destructor, then non-atomic read is not fenced. Probably doesn't matter in practice but tsan complains.

### DIFF
--- a/c++/src/kj/refcount.c++
+++ b/c++/src/kj/refcount.c++
@@ -72,7 +72,13 @@ void Refcounted::disposeImpl(void* pointer) const {
 // Atomic (thread-safe) refcounting
 
 AtomicRefcounted::~AtomicRefcounted() noexcept(false) {
-  KJ_ASSERT(refcount == 0, "Refcounted object deleted with non-zero refcount.");
+#if _MSC_VER && !defined(__clang__)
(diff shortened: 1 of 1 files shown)


harrishancock (approved):

c++/src/kj/refcount.c++
```diff
@@ -72,7 +72,13 @@ void Refcounted::disposeImpl(void* pointer) const {
 // Atomic (thread-safe) refcounting
 
 AtomicRefcounted::~AtomicRefcounted() noexcept(false) {
-  KJ_ASSERT(refcount == 0, "Refcounted object deleted with non-zero refcount.");
+#if _MSC_VER && !defined(__clang__)
+  KJ_ASSERT(KJ_MSVC_INTERLOCKED(Or, acq)(&refcount, 0) == 0,
+      "Refcounted object deleted with non-zero refcount.");
+#else
+  KJ_ASSERT(__atomic_load_n(&refcount, __ATOMIC_ACQUIRE) == 0,
```
kentonv, line 79: I think this could be a relaxed read rather than acquire. Makes no difference on x86 but might make a difference on arm.

mikea: https://github.com/capnproto/capnproto/pull/2774

PR #2773, opened 2026-08-08
Don't fail the listener when a connection is reset before accept() on Windows

petebacondarwin:
A client that completes the TCP handshake and then resets before the server gets around to
accepting leaves a dead connection in the accept queue. `AcceptEx()` reports this as
`ERROR_NETNAME_DELETED` on the overlapped completion, and `FdConnectionReceiver::accept()` failed
the whole listener. The exception propagates out of `accept()`, and callers generally treat an
`accept()` failure as fatal to the accept loop.

Unix does not have this problem because `async-io-unix` has always retried these errors. Linux
`accept()` likewise reports pending network errors from the connection it accepted — the man page
says so explicitly — and `acceptImpl()` treats a guessed-at set of those codes as transient. The
platforms behave the same way here; only the Win32 implementation was missing the logic.

### report in-flight `kj::Exception` from the terminate handler

`std::current_exception()` only reports an exception that is *currently being handled*. When an
exception escapes a `noexcept` function the MSVC ABI reaches `std::terminate()` before it begins
handling it, so the handler saw nothing and printed:

```
*** std::terminate() called with no exception
```

followed by a stack of exception-dispatch frames that says nothing about who threw. Against a
stripped release binary that leaves no way to identify the failure.

KJ tracks its own exceptions independently of the C++ runtime — `getDestructionReason()` already
relies on this — and `InFlightExceptionIterator` is documented safe to use from a signal handler.
Consult it when the runtime reports nothing. A `kj::Exception` names the throw site and carries its
description and trace as data, so the report stays useful with no symbols present.

This recovers `kj::Exception`s thrown through KJ's own throw functions, which is what
`InFlightExceptionIterator` can see; anything thrown by a bare `throw` or by the standard library
still reports nothing, as before.

### retry transient errors from `AcceptEx()`

Classify the statuses that describe a dead queued connection as transient and take the next
connection instead. Which codes those are is a guess, as it is on Unix. Both Win32 and Winsock
spellings appear since an overlapped completion may report either. `ERROR_OPERATION_ABORTED` is
excluded: that means the operation was cancelled, which must propagate.

The reset-before-accept tests added in 7df5bd0 were Unix-only because they build the doomed
connection from raw POSIX socket calls. That setup moves into a helper with a Winsock
implementation so the IPv4 case runs on both platforms. The dual-stack case stays Unix-only; it
exists to exercise a Darwin `accept()` quirk with no Win32 analogue.

### classify Win32 spellings of network errors

`typeOfWin32Error()` recognised only the Winsock spellings, so a failure surfacing as an overlapped
completion status fell through to `Exception::Type::FAILED`. A connection reset reported as
`ERROR_NETNAME_DELETED` therefore did not read as `DISCONNECTED`, and callers which branch on the
exception type — including kj's own reset-before-accept tests — treated it as an unexpected error.

### correct the description of Unix `accept()` behaviour

Addresses review. The comment above the reset-before-accept tests, which predates this branch,
claimed Unix hands over the dead connection and only fails on the first read. It doesn't, and that
claim had propagated into the new `isTransientAcceptError()` comment and into this PR description.
Describe what actually happens, and state plainly that `isTransientAcceptError()` is guessing which
errors belong to the listener and which to the connection, as its Unix counterpart already admits
about itself. Also reports an in-flight `kj::Exception` under the same heading as a
currently-handled one.

## Verification

Red/green on the newly enabled test, run on a fork so as not to put a knowingly-broken build here.

**Without the accept fix**, both Windows jobs fail identically:

```
[ TEST ] async-io-test.c++:3745: accept() with aborted connection - IPv4
kj/async-io-win32.c++:861: failed: AcceptEx(): #64 The specified network name is no longer available.
[ FAIL ] async-io-test.c++:3745: accept() with aborted connection - IPv4 (42.032ms)
```

`#64` is `ERROR_NETNAME_DELETED`, and line 861 is the `KJ_FAIL_WIN32` this change guards. Worth
noting that Windows does *not* drop the reset connection from the backlog before `AcceptEx()` sees
it, so the test is not vacuous.

**With it**, all 16 jobs pass, including `MSVC`, `Windows-bazel-clang-cl`, asan, ubsan and
clang-tidy. The review fixes on top are comments plus one message string, so that evidence stands.

### DIFF
--- a/c++/src/kj/async-io-test.c++
+++ b/c++/src/kj/async-io-test.c++
@@ -3705,8 +3705,43 @@ KJ_TEST("Calling abortRead() after tryRead() raised exception") {
 // On Unix platforms, accept() typically returns the aborted connection, which fails on first
 // read (throws "Connection reset by peer" or returns 0 bytes). The test then calls accept()
 // again to get the valid connection. Some platforms may filter out aborted connections.
+//
+// On Windows the abort surfaces differently: it completes the pending AcceptEx() operation with
+// an error rather than yielding a socket, so the listener discards it and the first accept()
+// returns the valid connection.
 
-#if !_WIN32
+// Open an IPv4 connection to a loopback port and immediately reset it, leaving a dead entry in
+// that listener's accept queue. SO_LINGER with a zero timeout makes close() send RST rather than
+// FIN. This works below kj::Network on purpose: the connection has to die before anything accepts
+// it, which means owning the socket's close behaviour directly.
+void connectAndReset(uint16_t port) {
+#if _WIN32
+  SOCKET s = ::socket(AF_INET, SOCK_STREAM, 0);
+  KJ_ASSERT(s != INVALID_SOCKET);
+  KJ_DEFER(::closesocket(s));
+#else
+  int s = ::socket(AF_INET, SOCK_STREAM, 0);
+  KJ_ASSERT(s >= 0);
+  KJ_DEFER(::close(s));
+#endif
+
+  struct linger lg = {1, 0};
+  sockaddr_in a {};
+  a.sin_family = AF_INET;
+  a.sin_port = htons(port);
+  a.sin_addr.s_addr = htonl(0x7f000001);
+
+#if _WIN32
+  KJ_WINSOCK(::setsockopt(s, SOL_SOCKET, SO_LINGER,
+                          reinterpret_cast<const char*>(&lg), static_cast<int>(sizeof(lg))));
+  KJ_WINSOCK(::connect(s, reinterpret_cast<sockaddr*>(&a), static_cast<int>(sizeof(a))));
+#else
+  KJ_SYSCALL(::setsockopt(s, SOL_SOCKET, SO_LINGER,
+                          reinterpret_cast<const char*>(&lg), sizeof(lg)));
+  KJ_SYSCALL(::connect(s, reinterpret_cast<sockaddr*>(&a), sizeof(a)));
+#endif
+}
+
 KJ_TEST("accept() with aborted connection - IPv4") {
   // -- async-io boilerplate ---------------------------------------------------
   auto io = kj::setupAsyncIo();
@@ -3716,22 +3751,7 @@ KJ_TEST("accept() with aborted connection - IPv4") {
   auto listener = listenerAddr->listen();
   uint16_t port = listener->getPort();
 
-  // Create a connection that will be aborted (sends RST packet)
-  int s = ::socket(AF_INET, SOCK_STREAM, 0);
-  KJ_ASSERT(s >= 0);
-
-  // Configure socket to send RST on close instead of FIN
-  struct linger lg = {1, 0};
-  KJ_SYSCALL(setsockopt(s, SOL_SOCKET, SO_LINGER, &lg, sizeof(lg)));
-
-  sockaddr_in a {};
-  a.sin_family = AF_INET;
-  a.sin_port = htons(port);
-  a.sin_addr.s_addr = htonl(0x7f000001);
-  KJ_SYSCALL(::connect(s, reinterpret_cast<sockaddr*>(&a), sizeof(a)));
-
-  // Close immediately to send RST packet
-  KJ_SYSCALL(::close(s));
+  connectAndReset(port);
 
   // Allow aborted connection to reach accept queue first
   io.provider->getTimer().afterDelay(20 * kj::MILLISECONDS).wait(io.waitScope);
@@ -3785,7 +3805,6 @@ KJ_TEST("accept() with aborted connection - IPv4") {
   KJ_ASSERT(amount == 5);
   KJ_ASSERT(memcmp(buffer, "hello", 5) == 0);
 }
-#endif  // !_WIN32
 
 // ---------------------------------------------------------------------------------------------
 // accept() with aborted connection - dual-stack IPv4/IPv6 listener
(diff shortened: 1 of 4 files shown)


petebacondarwin:
The red and green test runs are at:

[petebacondarwin/capnproto#1](https://github.com/petebacondarwin/capnproto/pull/1)
[petebacondarwin/capnproto#2](https://github.com/petebacondarwin/capnproto/pull/2)

Also note that the issue was debugged on a local Windows 11 machine with help and prompting from Opus.

c++/src/kj/exception.c++
```diff
@@ -634,9 +634,22 @@ namespace {
...
+    // terminate() is reached because an exception escaped a noexcept function, some ABIs -- the
+    // MSVC one in particular -- have not begun handling it yet, so nothing is reported here even
+    // though an exception is very much in flight. KJ tracks its own exceptions independently of
+    // the C++ runtime, so fall back to that rather than reporting nothing at all. Unlike a stack
+    // trace taken here, a kj::Exception names the throw site and carries its own trace, so this
+    // stays useful even when the binary has no symbols.
+    InFlightExceptionIterator iter;
+    KJ_IF_SOME(exception, iter.next()) {
+      message = kj::str("*** std::terminate() called with in-flight kj::Exception: ",
+                        exception, '\n');
```
kentonv: We should use the same error message here as we use for a kj::Exception above.

```suggestion
      message = kj::str("*** Fatal uncaught kj::Exception: ", exception, '\n');
```

c++/src/kj/async-io-win32.c++
```diff
@@ -818,6 +818,39 @@ Promise<Array<SocketAddress>> SocketAddress::lookupHost(
...
+  // AcceptEx() reports the state of the connection it accepted, not the state of the listener. A
+  // client that completes the handshake and then resets before the server gets around to accepting
+  // leaves a dead connection in the queue, and that is what surfaces here. Such an error says
+  // nothing about the listener's health, so accept() must take the next connection rather than
+  // fail: ConnectionReceiver::accept() reports only permanent errors.
+  //
+  // The two platforms surface a reset connection at different points, so this list is not the same
+  // as the one Unix retries on. On Unix accept() succeeds and the reset is only discovered on the
+  // first read, whereas AcceptEx() fails at accept time instead -- so the codes meaning "reset"
+  // have to be treated as transient here, and Unix has no need for them. Both Win32 and Winsock
```
kentonv: These comments about Unix working differently contradict my understanding. On Linux, the man page explicitly states:

> **Error handling**
> Linux accept() (and accept4()) passes already-pending network errors on the new socket as an error code from accept().   This  behavior  differs from  other  BSD socket implementations.  For reliable operation the application should detect the network errors defined for the protocol after accept() and treat them like EAGAIN by retrying.  In the case of TCP/IP, these are ENETDOWN, EPROTO, ENOPROTOOPT,  EHOSTDOWN,  ENONET,  EHOSTUNREACH, EOPNOTSUPP, and ENETUNREACH.

So it seems that Windows and Linux actually work the same. The reason we see this problem only on Windows is because async-io-unix has always been written to handle these errors -- the logic was simply missing from async-io-win32.

I would suggest just deleting the second paragraph of this comment.

c++/src/kj/async-io-test.c++
```diff
@@ -3705,33 +3705,53 @@ KJ_TEST("Calling abortRead() after tryRead() raised exception") {
 // On Unix platforms, accept() typically returns the aborted connection, which fails on first
```
kentonv: Maybe this is why Opus is confused. This pre-existing comment seems to be wrong... might be AI slop itself.

c++/src/kj/async-io-win32.c++
```diff
@@ -818,6 +818,39 @@ Promise<Array<SocketAddress>> SocketAddress::lookupHost(
 
 // =======================================================================================
 
+bool isTransientAcceptError(DWORD error) {
+  // AcceptEx() reports the state of the connection it accepted, not the state of the listener. A
```
kentonv: This part of the comment is a bit misleading too. In fact AcceptEx() may report errors about either the accept socket or the new connection. That's the point of this function -- to decide which kind of error it is.

c++/src/kj/exception.c++
petebacondarwin: Done, applied as suggested.

c++/src/kj/async-io-win32.c++
petebacondarwin: You are right, and the Unix code says as much already — `acceptImpl()` retries on a set of codes with the comment "According to the Linux man page, accept() may report an error if the accepted connection is already broken ... it's hard to say exactly what errors are such network errors and which ones are permanent. We've made a guess here." That is exactly what this function is doing, so the paragraph claiming the platforms differ is deleted.

I have also corrected the same claim in the commit message and the PR description, since it had spread to both.

c++/src/kj/async-io-test.c++
petebacondarwin: That is exactly where I got it from — I took the comment at its word instead of checking, which is on me.

Rewritten to describe what actually varies between platforms (the dead connection may never reach `accept()`, may be reported as an error from it, or may be handed over and fail on first read) and to state what the test actually asserts: that the listener survives and still produces the valid connection either way. The dual-stack test's header repeated the same wrong claim, so that is fixed too, keeping the part about the Darwin `addrlen == 0` quirk which is the reason that test exists.

c++/src/kj/async-io-win32.c++
petebacondarwin: Reworded to lead with that: `AcceptEx()` may fail because the listener is broken or because the connection it was accepting is, and telling the two apart is the job of this function — with the same admission the Unix version makes, that which codes fall on which side is a guess.

kentonv (approved):

petebacondarwin:
Thanks!

petebacondarwin:
Do you want me to do the PR to pull this into workerd?

kentonv:
If you want this to get into workerd promptly then you should do that yeah.

PR #2777, opened 2026-08-11
HashMap callback style api

mikea:
Reference returning api is prone to errors, since it is very easy to invalidate those (e.g. insert into map and trigger rehash).

The only non-trivial part here is returning `Maybe` from functions conditionally calling callbacks. This makes end user code slightly more elegant.

Not to force people `return true` every time, this includes `Maybe<void>` support.

### DIFF
--- a/c++/src/kj/map.h
+++ b/c++/src/kj/map.h
@@ -71,7 +71,14 @@ public:
   Entry& upsert(Key key, Value value);
   // Tries to insert a new entry. However, if a duplicate already exists (according to some index),
   // then update(Value& existingValue, Value&& newValue) is called to modify the existing value.
-  // If no function is provided, the default is to simply replace the value (but not the key).
+  // If no update function is provided, the default is to simply replace the value (but not the
+  // key).
+
+  template <typename UpdateFunc, typename Func>
+  auto upsert(Key key, Value value, UpdateFunc&& update, Func&& callback);
+  // Like upsert(key, value, update), then calls callback(value) on the inserted or updated value.
+  // Returns the callback's decayed result. The callback must not modify this map while it is using
+  // the value.
 
   template <typename KeyLike>
   kj::Maybe<Value&> find(KeyLike&& key);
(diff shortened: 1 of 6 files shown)


c++/src/kj/map.h
```diff
@@ -368,6 +426,16 @@ typename HashMap<Key, Value>::Entry& HashMap<Key, Value>::upsert(
...
   });
 }
+template <typename Key, typename Value>
+template <typename UpdateFunc, typename Func>
+auto HashMap<Key, Value>::upsert(
+    Key key, Value value, UpdateFunc&& update, Func&& callback) {
+  auto& entry = table.upsert(Entry { kj::mv(key), kj::mv(value) },
+      [&](Entry& existingEntry, Entry&& newEntry) {
+    update(existingEntry.value, kj::mv(newEntry.value));
+  });
```
kentonv, line 436: Can this be like:

```suggestion
  auto& entry = this->upsert(
      kj::fwd<Key>(key), kj::fwd<Value>(value), kj::fwd<UpdateFunc>(updateFunc));
```

c++/src/kj/map.h
```diff
@@ -114,6 +151,27 @@ class HashMap {
   // Erase all values for which predicate(key, value) returns true. This scans over the entire map.
 
 private:
+  template <typename Func, typename Arg>
+  static auto callCallback(Func&& callback, Arg&& arg) {
+    using Result = kj::Decay<decltype(callback(kj::fwd<Arg>(arg)))>;
```
kentonv, line 156: Why decay?

kentonv (commented):
Ehhh, I'm skeptical... seems like the idea here is that people aren't supposed to modify the map during the callback. But nothing is stopping them, right? I feel like you'd still get the same bugs: in the callback you call a method which you don't realize may come back and modify the same map, whoops.

Would be more compelling if there was enforcement.

But for that matter you could also extend the existing API to do enforcement: have it return some sort of checked reference type and fail if the reference is accessed after the map has changed, or alternatively fail if the map is changed while a checked reference exists. A lot like Pin/Ptr, etc.

PR #2778, opened 2026-08-12
Refactor kj async build targets

danlapid:
The kj async framework is layered into three targets.

:kj-async-core  Promise machinery: kj::Promise / EventLoop / EventPort, coroutine support,
                and timers. No I/O.
:kj-async-io    Abstract, OS-independent async I/O: the AsyncInputStream / AsyncOutputStream /
                AsyncIoStream interfaces (and their default-method implementations), in-memory
                pipes, the kj::Network / NetworkAddress interfaces, NetworkFilter, and CIDR
                address parsing. No event loop.
:kj-async-os    The concrete OS event loop and I/O implementation: kj::UnixEventPort /
                kj::Win32IocpEventPort and kj::setupAsyncIo().

The layering is strictly one-directional (os -> io -> core). The :kj-async umbrella preserves the historical all-in-one target Existing consumers need no changes.

### DIFF
--- a/c++/src/capnp/BUILD.bazel
+++ b/c++/src/capnp/BUILD.bazel
@@ -8,6 +8,7 @@ capnp_configure()
 
 exports_files([
     "c++.capnp",
+    "rpc.capnp",
 ])
 
 cc_library(
@@ -66,7 +67,8 @@ cc_library(
     include_prefix = "capnp",
     visibility = ["//visibility:public"],
     deps = [
-        "//src/kj:kj-async",
+        "//src/kj:kj-async-core",
+        "//src/kj:kj-async-io",
     ],
 )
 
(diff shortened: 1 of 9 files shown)


c++/src/capnp/compat/BUILD.bazel
```diff
@@ -6,6 +6,8 @@ load("@rules_cc//cc:cc_test.bzl", "cc_test")
 exports_files([
     "json.capnp",
     "std-iterator.h",
+    "byte-stream.capnp",
```
danlapid, line 9: Drive-by but I don't think this is contentious and I do need it.

c++/src/capnp/BUILD.bazel
```diff
@@ -8,6 +8,7 @@ capnp_configure()
 
 exports_files([
     "c++.capnp",
+    "rpc.capnp",
```
danlapid, line 11: Drive-by but I don't think this is contentious and I do need it.

mikea (approved):

PR #2779, opened 2026-08-13
[tsan] XThreadEvent: use atomic load to access state

mikea:
Looks to be the last tsan issue: https://github.com/capnproto/capnproto/actions/runs/31715250495/job/94498232970?pr=2768


kentonv (approved):
I suspect this is theoretically unnecessary because the `DONE` state is set under the same lock that we're locking with all these `.wait()`s but this seems fine to make TSAN happy.

mikea:
> I suspect this is theoretically unnecessary because the `DONE` state is set under the same lock that we're locking with all these `.wait()`s but this seems fine to make TSAN happy.

`while (state != DONE)` was not under lock, and this is what tsan complained about afaiu.

PR #2780, opened 2026-08-13
kj/atomic.h

mikea:
Consolidating all intrinsics usages in a single file. Using std `<atomic>` for windows consistently

### DIFF
--- a/c++/Makefile.am
+++ b/c++/Makefile.am
@@ -144,6 +144,7 @@ noinst_HEADERS = \
 includekj_HEADERS =                                            \
   src/kj/common.h                                              \
   src/kj/units.h                                               \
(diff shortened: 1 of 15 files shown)


harrishancock (approved):
Looks like a nice cleanup. Personally, I'd prefer it if we could just use the std functions across platforms, though perhaps including that header slows down compile times noticeably.

mikea:
> Looks like a nice cleanup. Personally, I'd prefer it if we could just use the std functions across platforms, though perhaps including that header slows down compile times noticeably.

Agree, but that is a bigger discussion and was just tired of inconsistent ifdefs everywhere. Windows implementation choices were particularly dubious, but I guess they predated atomic support in msvc.

PR #2786, opened 2026-08-18
kj/memory.h kj::PtrTarget: allow for descendent classes to invalidate references

jtolio:
Imagine you have

    class MyClass : public kj::PtrTarget {
     public:
      MyClass() : task(run(addWeakToThis())) {}
      ~MyClass() { cleanup(); }

     private:
      static kj::Promise<void> run(kj::Weak<MyClass> weak) {
        KJ_DEFER(weak.assertLive().doSomething());
        co_await kj::Promise<void>(kj::NEVER_DONE);
      }

      void doSomething();
      void cleanup();

      kj::Promise<void> task;
    };

If you have an instance of MyClass and it destructs, it might be surprising to realize that `doSomething` here *runs successfully* and runs *after* `cleanup`.

With this change, now you can add invalidateWeak() to ~MyClass() so that `weak` in `run` won't be able to be upgraded to a live pointer in a partially destructed state.

### DIFF
--- a/c++/src/kj/memory.h
+++ b/c++/src/kj/memory.h
@@ -249,6 +249,18 @@ protected:
     return Weak<Self>(&self, asPtrTarget(self));
   }
 
+  inline void invalidateWeak() {
+    // Permanently invalidate weak pointers to this object. Derived class destructors can call this
+    // before tearing down state that may otherwise be accessed while the rest of the object is
+    // destroyed.
+    if (weakCell == nullptr) {
+      weakCell = new _::WeakCell(nullptr, nullptr);
+    } else {
+      weakCell->ptr = nullptr;
+      weakCell->target = nullptr;
+    }
+  }
+
 private:
   template <typename Self>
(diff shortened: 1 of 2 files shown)


c++/src/kj/memory.h
```diff
@@ -262,6 +262,12 @@ class PtrTarget {
     return Weak<Self>(&self, asPtrTarget(self));
   }
 
+  inline void invalidateWeak() {
+    // Invalidates all existing weak pointers. For optional use in descendent class destructors,
+    // to avoid weak pointers becoming active after only partial destruction.
+    dispose();
```
mikea: dispose has assertEmpty() at the end. I doubt we want it here too.

mikea (changes requested):

c++/src/kj/memory.h
jtolio: oh right, there might be non-weak pointers out too. fixed

mikea (approved):

PR #2789, opened 2026-08-20
kj/async: Define KJ_CO_MAGIC kj::CURRENT_INVOCATION and helpers

jtolio:
This change defines a new operation for KJ_CO_MAGIC that allows you to get a handle to the current coroutine invocation. E.g., you can do

    auto inv = KJ_CO_MAGIC kj::CURRENT_INVOCATION;

and inv will have methods defined on it for determining things about the invocation.

The current three helpers available are isCanceling() (always available), isUnwinding(), and scopeOutcome() (only available for "CoUnwindAware" coroutines). These features allow for coroutine-specific teardown handling that discriminates on *why* the coroutine is tearing down. Was it due to a successful completion? An exception? Or, harder to determine, due to a cancelation?

Why don't existing features already help? 
 * KJ_DEFER runs in all three cases, but cannot easily tell you which case you are in or why. 
 * KJ_ON_SCOPE_SUCCESS fires when a coroutine scope is canceled! If a coroutine scope is canceled and thus destructed without further execution, no exception is thrown, so KJ_ON_SCOPE_FAILURE doesn't fire, and since KJ_ON_SCOPE_SUCCESS is defined with the opposite condition, KJ_ON_SCOPE_SUCCESS does.
 * KJ_ON_SCOPE_FAILURE has edge cases with regard to coroutine resumption on differing stacks and exception handling. It is possible for a coroutine to be resumed with different uncaught exception counters. As a result, UnwindDetector (and thus KJ_ON_SCOPE_FAILURE (and thus KJ_ON_SCOPE_SUCCESS)) have critical corner cases that may misidentify failure in coroutines depending on how they are resumed and reentered.
 
Example use:

```
Promise<void> unwindAwareFoo(CoUnwindAware = {}) {
  auto inv = KJ_CO_MAGIC CURRENT_INVOCATION;
  KJ_DEFER({
    switch (inv.scopeOutcome()) {
    case kj::CoScopeOutcome::SUCCESS:
      // coroutine exited successfully
    case kj::CoScopeOutcome::FAILURE:
      // coroutine exited with exception thrown
    case kj::CoScopeOutcome::CANCELED:
      // coroutine destruction before execution completed.
    }
  });
  co_await stuff();
}
```

### DIFF
--- a/c++/src/kj/async-coroutine-test.c++
+++ b/c++/src/kj/async-coroutine-test.c++
@@ -890,6 +890,144 @@ KJ_TEST("KJ_TRY/KJ_CATCH inside try/catch with promise rejection in coroutines")
   auto result = testCoro().wait(waitScope);
   KJ_EXPECT(result == "kj:caught std:not-caught");
 }
+
+// =======================================================================================
+// KJ_CO_MAGIC CURRENT_INVOCATION
+
+KJ_TEST("KJ_CO_MAGIC CURRENT_INVOCATION returns CoInvocation without suspending") {
+  EventLoop loop;
+  WaitScope waitScope(loop);
+
+  bool reachedEnd = false;
+  auto coro = [&]() -> Promise<void> {
+    auto invocation = KJ_CO_MAGIC CURRENT_INVOCATION;
+    static_assert(isSameType<decltype(invocation), CoInvocation>());
+    reachedEnd = true;
+    co_return;
+  };
+
+  auto promise = coro();
+
+  // Coroutines start eagerly, so had obtaining the reference suspended, the body would not have run
+  // to the end yet.
+  KJ_EXPECT(reachedEnd);
+
+  promise.wait(waitScope);
+}
+
+// =======================================================================================
+// CoInvocation::isCanceling()
+
+template <typename T, typename Coroutine>
+class StateRecorder {
+  // Reports what `query` said about the enclosing coroutine at the moment this object was
+  // destroyed. Meant to live in a coroutine frame, so that its destructor runs on every path out of
+  // the coroutine.
+  //
+  // The coroutine to watch is supplied after construction rather than to the constructor, so that
+  // this can be used as a coroutine parameter: the body cannot obtain the reference until it starts
+  // running, which is after its parameters have been copied into the frame. It also means the
+  // copy left behind at the call site reports nothing when it is destroyed.
+
+public:
+  using Query = T(const Coroutine&);
+
+  StateRecorder(Maybe<T>& result, Query& query)
+      : result(result), query(query) {}
+  ~StateRecorder() {
+    KJ_IF_SOME(c, coroutine) {
+      result = query(c);
+    }
+  }
+
+  void watch(Coroutine c) {
+    coroutine = kj::mv(c);
+  }
+
+private:
+  Maybe<Coroutine> coroutine;
+  Maybe<T>& result;
+  Query& query;
+};
+
+bool canceling(const CoInvocation& invocation) { return invocation.isCanceling(); }
+
+KJ_TEST("isCanceling() is false when a coroutine runs to completion") {
+  EventLoop loop;
+  WaitScope waitScope(loop);
+
+  Maybe<bool> canceled;
+  auto coro = [&]() -> Promise<void> {
+    StateRecorder recorder(canceled, canceling);
+    recorder.watch(KJ_CO_MAGIC CURRENT_INVOCATION);
+    co_await kj::yield();
+  };
+
+  coro().wait(waitScope);
+  KJ_EXPECT(KJ_ASSERT_NONNULL(canceled) == false);
+}
+
+KJ_TEST("isCanceling() is false while an exception propagates out of a coroutine") {
+  EventLoop loop;
+  WaitScope waitScope(loop);
+
+  Maybe<bool> canceled;
+  auto coro = [&]() -> Promise<void> {
+    StateRecorder recorder(canceled, canceling);
+    recorder.watch(KJ_CO_MAGIC CURRENT_INVOCATION);
+    co_await kj::yield();
+    KJ_FAIL_ASSERT("coroutine failed");
+  };
+
+  auto exception = kj::runCatchingExceptions([&]() { coro().wait(waitScope); });
+  KJ_EXPECT(exception != kj::none);
+  KJ_EXPECT(KJ_ASSERT_NONNULL(canceled) == false);
+}
+
+KJ_TEST("isCanceling() is true while destroying a suspended coroutine") {
+  EventLoop loop;
+  WaitScope waitScope(loop);
+
+  Maybe<bool> canceled;
+  auto coro = [&]() -> Promise<void> {
+    StateRecorder recorder(canceled, canceling);
+    recorder.watch(KJ_CO_MAGIC CURRENT_INVOCATION);
+    co_await kj::Promise<void>(kj::NEVER_DONE);
+  };
+
+  {
+    auto promise = coro();
+    KJ_EXPECT(!promise.poll(waitScope));
+    KJ_EXPECT(canceled == kj::none);
+  }
+
+  KJ_EXPECT(KJ_ASSERT_NONNULL(canceled) == true);
+}
+
+Promise<void> watchFromParameter(
+    StateRecorder<bool, CoInvocation> recorder, Promise<void> awaitMe) {
+  recorder.watch(KJ_CO_MAGIC CURRENT_INVOCATION);
+  co_await awaitMe;
+}
+
+KJ_TEST("isCanceling() is false while destroying a completed coroutine") {
+  EventLoop loop;
+  WaitScope waitScope(loop);
+
+  // A coroutine's parameters outlive its body: they are destroyed with the frame, which for a
+  // coroutine that completed happens when its promise is dropped. So this asks isCanceling()
+  // during destruction of a coroutine that was never canceled -- the case that distinguishes being
+  // destroyed from being canceled.
+  Maybe<bool> canceled;
+  {
+    auto promise = watchFromParameter(StateRecorder(canceled, canceling), kj::READY_NOW);
+    KJ_EXPECT(promise.poll(waitScope));
+    KJ_EXPECT(canceled == kj::none);
+  }
+
+  KJ_EXPECT(KJ_ASSERT_NONNULL(canceled) == false);
+}
+
 #endif  // !(__GNUC__ && !__clang__)
 
 }  // namespace
--- a/c++/src/kj/async-coroutine-test.c++
+++ b/c++/src/kj/async-coroutine-test.c++
@@ -44,10 +44,16 @@ static_assert(sizeof(void*) != 8 || alignof(_::CoroutineBase) == 8,
     "CoroutineBase must retain its 64-bit ABI alignment");
 
 template <typename T>
+concept HasScopeOutcome = requires(const T& c) { c.scopeOutcome(); };
+
+template <typename T>
 concept HasIsUnwinding = requires(const T& c) { c.isUnwinding(); };
 
+static_assert(!HasScopeOutcome<CoInvocation>,
+    "scopeOutcome() must be unavailable without kj::CoUnwindAware");
 static_assert(!HasIsUnwinding<CoInvocation>,
     "isUnwinding() must be unavailable without kj::CoUnwindAware");
+static_assert(HasScopeOutcome<CoUnwindAwareInvocation>);
 static_assert(HasIsUnwinding<CoUnwindAwareInvocation>);
 
 template <typename T>
@@ -935,6 +941,7 @@ KJ_TEST("KJ_CO_MAGIC CURRENT_INVOCATION returns CoUnwindAwareInvocation when opt
     static_assert(isSameType<decltype(invocation), CoUnwindAwareInvocation>());
     KJ_EXPECT(!invocation.isCanceling());
     KJ_EXPECT(!invocation.isUnwinding());
+    KJ_EXPECT(invocation.scopeOutcome() == CoScopeOutcome::SUCCESS);
   };
 
   coro().wait(waitScope);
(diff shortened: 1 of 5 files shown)


c++/src/kj/async.c++
```diff
@@ -3058,7 +3058,8 @@ namespace _ {  // (private)
 
 CoroutineBase::CoroutineBase(stdcoro::coroutine_handle<> coroutine, SourceLocation location)
     : Event(location),
-      coroutine(coroutine) {}
+      coroutine(coroutine),
+      uncaughtCountAtEntry(UnwindDetector::uncaughtExceptionCount()) {}
```
jtolio, line 3062: oops. I need to retool this so that we don't pay the uncaught_exceptions penalty for every coroutine creation, and only the ones that opt in using KJ_CO_DEFER and friends.

c++/src/kj/async.c++
```diff
@@ -3118,6 +3119,8 @@ void CoroutineBase::fire() {
   // already know where it's going. But, we don't really know: the `co_await` might be in a
   // try-catch block, so we have no choice but to resume and throw later.
 
+  uncaughtCountAtEntry = UnwindDetector::uncaughtExceptionCount();
```
jtolio, line 3122: and here

c++/src/kj/async.c++
jtolio: fixed

c++/src/kj/async.c++
jtolio: fixed

c++/src/kj/async-inl.h
```diff
@@ -2332,6 +2332,40 @@ class CoroutineBase: public PromiseNode,
...
+  // Coroutines might end in one of three ways - running successfully to the end, throwing
+  // an exception, or getting destructed while parked.
+  // isCanceling can be used when a coroutine is being destructed to determine if the cause
+  // of the destruction is the third case - getting destructed while parked. Note that this method
+  // is only useful during destruction and not after.
+  bool isCanceling() const { return maybeDisposalResults != kj::none && !isDone(); }
+
+  // isUnwinding returns whether this coroutine is currently unwinding based on the current
+  // stack invocation. A problem with using UnwindDetector directly is that a coroutine might be
+  // entered from many different stacks, with different uncaught exception counts at each point.
```
harrishancock: I would expect a coroutine could be entered from only two stacks:

- Its initial invocation
- The event loop calling the coroutine's `Event::fire()`

That doesn't necessarily make the problem you're describing go away, but I'm curious if you're seeing different stacks besides those two?

Note that I'm using "coroutine" to refer to an individual instantiation of a coroutine -- perhaps that's where my different understanding comes from?

c++/src/kj/async-inl.h
```diff
@@ -2370,6 +2404,11 @@ class CoroutineBase: public PromiseNode,
   inline bool isDone() const { return coroutine.done(); }
 #endif
 
+  // uncaughtCountAtEntry is only for the isUnwinding() method. This field is here so we can
+  // smuggle it inside of the existing padding.
+  // Note that when uncaughtCountAtEntry == kj::maxValue, unwinding checking is not enabled.
+  uint uncaughtCountAtEntry = kj::maxValue;
```
harrishancock: I see you've gone to some length to make this performance-neutral.

Still, ideally we want coroutine frame size to go down over time. I worry that adding a new data member is going to make that harder, even if the new member currently fits in padding bytes. Concretely, it forecloses using pointer tagging to optimize out `hasSuspendedAtLeastOnce`, which would benefit all coroutines, whereas it sounds like we don't have many coroutines which would benefit from the new KJ_CO_DEFER feature.

c++/src/kj/async-inl.h
```diff
@@ -2387,6 +2426,23 @@ class CoroutineBase: public PromiseNode,
   // coroutine.destroy() has returned. Our disposer then rethrows as needed.
 };
 
+// CurrentCoroutineConstant is for use in KJ_CO_MAGIC invocations to get a reference to the
+// current CoroutineBase.
+class CurrentCoroutineConstant final {};
+inline constexpr CurrentCoroutineConstant CURRENT_COROUTINE{};
```
harrishancock: Bikeshed: `THIS_COROUTINE`, though it doesn't matter much

c++/src/kj/async-inl.h
```diff
@@ -2387,6 +2426,23 @@ class CoroutineBase: public PromiseNode,
...
+
+// CoroutineAwaiter is for use in KJ_CO_MAGIC invocations to return a reference to the
+// provided CoroutineBase when requested.
+class CoroutineAwaiter {
```
harrishancock: This is a generic enough name that I worry it will be confusing to people who see it. `ThisCoroutineMagicAccessor`?

harrishancock (commented):
It sounds like there are two different problems being solved:

- `KJ_ON_SCOPE_SUCCESS` cannot distinguish between cancellation and success in coroutines.
- `KJ_ON_SCOPE_FAILURE` isn't reliable in coroutines because the uncaught exception count may differ between the initial call and subsequent resumptions, causing confusion between success and failure.

The former can be solved without adding a new data member, but the latter cannot -- at least, not in a performance-neutral way that I can see.

Would it be useful to solve the former without solving the latter? That would remove the main issue I worry about. Just adding a low-level `KJ_CO_MAGIC` accessor for the current coroutine would add very little code to our coroutine implementation.

c++/src/kj/async-inl.h
jtolio: Oh, sorry, what I mean to say is, when Event::fire() runs, the stack depth can vary. The ancestors of the current stack frame are not fixed, and there are more than 2 possible incarnations of the ancestor stack. The coroutine could be being fed from different promises at different points in its execution. I'm probably not being very precise with my wording so let me ground this. Let's say you have:

     Promise<void> myFunction(Promise<void> first, Promise<void> second) {
       printStacks();
       co_await first;
       printStacks();
       co_await second;
       printStacks();
     }

     void main() {
       EventLoop loop;
       WaitScope waitScope(loop);
       auto first = newPromiseAndFulfiller<void>();
       auto second = newPromiseAndFulfiller<void>();

       auto promise = myFunction(kj::mv(first.promise), kj::mv(second.promise));

       first.fulfiller->fulfill();
       loop.run();

       second.fulfiller->fulfill();
       promise.wait(waitScope);
     }
    
When printStacks() runs, there are three separate stacks we see here. The initial invocation, then when the first fulfiller fulfills via the event loop, and then when the second fulfiller fulfills via the waitScope. 

They look like:

    initial coroutine entry:
        printStacks()
        test()

    after first co_await:
        printStacks()
        coroutine_handle::resume()
        CoroutineBase::fire()
        EventLoop::turn()
        EventLoop::run()
        test()

    after second co_await:
        printStacks()
        coroutine_handle::resume()
        CoroutineBase::fire()
        EventLoop::turn()
        waitImpl callback()
        WaitScope::runOnStackPool()
        waitImpl()
        Promise::wait()
        test()

Using std::unhandled_exceptions() assumes that the ancestor stack is the same each time, so you have something to compare to, but in all three cases the stack is different, and might have a differing amount of currently unhandled exceptions. I imagine I could construct more cases. This feels like an extremely nitpicky point, and perhaps I'm wrong to assume this happens much?

c++/src/kj/async-inl.h
jtolio: Yeah, I imagine the majority of Coroutines won't use this. I'm not sure how to fix the other std::unhandled_exceptions count problem though. I was pretty pleased when I had the idea to smuggle it into the padding. I agree I don't want to increase the size.

jtolio:
> \[KJ_ON_SCOPE_SUCCESS cannot distinguish between cancellation and success in coroutines\] can be solved without adding a new data member, but the latter cannot -- at least, not in a performance-neutral way that I can see.

Hmm, you're saying, continuing to use the existing unwind detector logic in KJ_ON_SCOPE_SUCCESS, we could also check if the coroutine is canceling and fix that at least?

> Would it be useful to solve the former without solving the latter?

Right, I am not sure I want that feature if it might mischaracterize exceptions based on the other issue.

c++/src/kj/async-inl.h
jtolio: Oh that's a better name, yes

c++/src/kj/async-inl.h
jtolio: Yeah, that's better, agreed.

c++/src/kj/async-inl.h
jtolio: Though, to be fair, if this unwinding situation is a subtle enough mistake, but we use the unwind detector a lot (or any of its users), perhaps we should be using this in more Coroutines lol. Perhaps we should fix the non-CO KJ_ON_SCOPE_SUCCESS and friends

jtolio: Your suggestion of finding a way to make coroutines opt into this using templates paid off! I pushed a new commit to this stack that eliminates the new member to CoroutineBase.

c++/src/kj/async-inl.h
jtolio: Fixed

c++/src/kj/async-inl.h
jtolio: Fixed

jtolio:
Okay, usage with the new commit (`kj/async: Make coroutine unwind tracking opt-in`) looks like:

    kj::Promise<void> myFunction(kj::CoUnwindAware = {}) {
      KJ_CO_ON_SCOPE_SUCCESS(handleSuccess());
      co_await doStuff();
    }

If you miss adding CoUnwindAware, a new static assert will help remind you.

jtolio:
Callgrind benchmark changes relative to PR base: https://gist.github.com/jtolio/1e89aaaeaa9c0bcc004977701e073a29

c++/src/kj/async-inl.h
```diff
@@ -2332,11 +2337,27 @@ class CoroutineBase: public PromiseNode,
...
+  // of the destruction is the third case - getting destructed while parked. Note that this method
+  // is only useful during destruction and not after.
+  bool isCanceling() const { return maybeDisposalResults != kj::none && !isDone(); }
+
+  template <typename T = void>
+  CoScopeOutcome scopeOutcome() const {
+    static_assert(!isSameType<T, T>(),
+        "KJ_CO_DEFER and KJ_CO_ON_SCOPE_* require a defaulted kj::CoUnwindAware parameter");
+    KJ_UNREACHABLE;
+  }
```
harrishancock:
```suggestion
  CoScopeOutcome scopeOutcome() const KJ_UNAVAILABLE("requires a defaulted kj::CoUnwindAware parameter");
```

Or `= delete("...")` when we upgrade to C++26.

Note that this would also allow us to write a negative compile test, since `KJ_UNAVAILABLE` is SFINAE-visible:

```cpp
static_assert(!requires(const _::CoroutineBase& c) { c.scopeOutcome(); },
    "scopeOutcome() must be unavailable without kj::CoUnwindAware");
static_assert(requires(const _::UnwindAwareCoroutineBase& c) { c.scopeOutcome(); });
```

c++/src/kj/async-coroutine-test.c++
```diff
@@ -25,10 +25,18 @@
...
+#include <kj/vector.h>
 
 namespace kj {
 namespace {
 
+static_assert(sizeof(void*) != 8 || sizeof(_::CoroutineBase) == 104,
+    "CoroutineBase must retain its 64-bit ABI layout");
+static_assert(sizeof(void*) != 8 || alignof(_::CoroutineBase) == 8,
+    "CoroutineBase must retain its 64-bit ABI alignment");
+static_assert(sizeof(_::UnwindAwareCoroutineBase) > sizeof(_::CoroutineBase),
```
harrishancock: I'd suggest pinning this to a number, rather than allowing it to increase. Though, MSVC and gcc may be different. Maybe we should pin to a range.

c++/src/kj/async.h
```diff
@@ -695,6 +695,74 @@ auto coCapture(Functor&& f) {
...
+// An enum of the three different exit scenarios for a coroutine.
+enum class CoScopeOutcome {
+  SUCCESS,
+  // execution ended without an exception.
+
+  FAILURE,
+  // execution ended with an exception.
+
+  CANCELED,
+  // execution didn't end due to coroutine destruction.
```
harrishancock: This reads to me like "it is not the case that execution ended due to destruction".

```suggestion
  // the coroutine was destroyed before execution ended
```

c++/src/kj/async-coroutine-test.c++
```diff
@@ -880,6 +888,606 @@ KJ_TEST("KJ_TRY/KJ_CATCH inside try/catch with promise rejection in coroutines")
...
+    KJ_EXPECT(canceled == kj::none);
+  }
+
+  KJ_EXPECT(KJ_ASSERT_NONNULL(canceled) == false);
+}
+
+// =======================================================================================
+// CoroutineBase::isUnwinding()
+
+bool unwinding(const _::CoroutineBase& c) {
```
harrishancock: In places that use `KJ_CO_MAGIC THIS_COROUTINE` in an unwind-aware coroutine, I think `c` is already `UnwindAwareCoroutineBase&`, right? In which case the `static_cast` is unnecessary. And in regular coroutines, the `static_cast` would be a bug.

```suggestion
bool unwinding(const _::UnwindAwareCoroutineBase& c) {
```

c++/src/kj/async.h
```diff
@@ -695,6 +695,74 @@ auto coCapture(Functor&& f) {
...
+//     {
+//         KJ_CO_ON_SCOPE_CANCELED({
+//           metrics.canceled();
+//         });
+//         co_await otherstuff();
+//     }
+//
+// Naming follows CoScopeOutcome values. See also KJ_CO_DEFER.
+// The enclosing coroutine must have a defaulted kj::CoUnwindAware parameter.
+// Note: these macros will not detect failure correctly if they are run when an exception is
```
harrishancock: Nit: I'm fairly certain that these scenarios are un-constructible. They would require `co_yield` inside a catch handler, right?

c++/src/kj/async-inl.h
```diff
@@ -2387,6 +2407,48 @@ class CoroutineBase: public PromiseNode,
...
+
+protected:
+  void fire() override;
+
+private:
+  uint uncaughtCountAtEntry = UnwindDetector::uncaughtExceptionCount();
+};
+
+template <bool unwindAware>
+using CoroutineBaseFor = std::conditional_t<unwindAware, UnwindAwareCoroutineBase, CoroutineBase>;
```
harrishancock: We should either make sure to include `<type_traits>` or write this in the house style:

```cpp
template <bool unwindAware> struct CoroutineBaseFor_ { using Type = CoroutineBase; };
template <> struct CoroutineBaseFor_<true> { using Type = UnwindAwareCoroutineBase; };

template <bool unwindAware>
using CoroutineBaseFor = typename CoroutineBaseFor_<unwindAware>::Type;
```

harrishancock (dismissed):

harrishancock:
I didn't mean to approve yet, sorry! D: I wanted to discuss adding macros first.

I generally (strongly) prefer trying to make native language facilities work for us before adding more macros.

KJ_DEFER itself is a fairly good example of a macro which I think is worthwhile: the native language ways to say, "run this expression on scope exit" tend to be pretty verbose and syntax-heavy. The macro has its problems (the `-Wdangling-else` bug, see https://github.com/capnproto/capnproto/pull/2550), but that's fixable.

KJ_ON_SCOPE_FAILURE / KJ_ON_SCOPE_SUCCESS, before coroutines, also seem pretty reasonable, though their native language formulation is less bad than KJ_DEFER's, because KJ_DEFER already exists:

```cpp
kj::UnwindDetector detector;
KJ_DEFER({
  if (detector.isUnwinding()) {
    // on scope failure
  }
});
```

Not as succinct perhaps, but the syntax is pretty readable, unlike KJ_DEFER's alternatives.

Adjusting the macros for coroutines, though, starts showing some blemishes. The one you pointed out in chat -- the name KJ_CO_ON_SCOPE_NONSUCCESS -- highlights the problem: what about _NONEXCEPTION (success or canceled), NONCANCELED (success or exception)? Why does one earn its place but not the others? Which ones do we actually have a use for?

So, before we add a proliferation of KJ_CO_ON_SCOPE_ macros, I'd like to see how things look in our code if we treat the new `KJ_CO_MAGIC` atom as the first-class, user-facing interface of the new feature.

Right now, the new `KJ_CO_MAGIC` atom exposes an implementation detail, but it doesn't have to. We could expose a much more tightly scoped interface:

```cpp
// `KJ_CO_MAGIC CURRENT_INVOCATION` in regular coroutines
class CoInvocation {
public:
  bool isCanceling() const { return coro.isCanceling(); };
  bool isUnwinding() const KJ_UNAVAILABLE("requires kj::CoUnwindAware");
  CoScopeOutcome scopeOutcome() const KJ_UNAVAILABLE("requires kj::CoUnwindAware");
private:
  _::CoroutineBase& coro;
};

// `KJ_CO_MAGIC CURRENT_INVOCATION` in unwind-aware coroutines
class CoUnwindAwareInvocation {
public:
  bool isCanceling() const { return coro.isCanceling(); };
  bool isUnwinding() const { return coro.isUnwinding(); };
  CoScopeOutcome scopeOutcome() const { return coro.scopeOutcome(); }
private:
  _::UnwindAwareCoroutineBase& coro;
};
```

Use would look like:

```cpp
Promise<void> foo() {
  auto inv = KJ_CO_MAGIC CURRENT_INVOCATION;
  KJ_DEFER({
    if (inv.isCanceling()) {
      // scope is exiting due to cancellation
    }
  });
}
```

Or in `CoUnwindAware` coroutines:

```cpp
Promise<void> unwindAwareFoo(CoUnwindAware = {}) {
  auto inv = KJ_CO_MAGIC CURRENT_INVOCATION;
  KJ_DEFER({
    if (inv.isUnwinding()) {
      // scope is exiting due to cancellation
    }
  });
}
```

If that really ends up being too verbose, we can add macros in, starting with KJ_CO_DEFER.

c++/src/kj/async-inl.h
jtolio: good call. fixed

c++/src/kj/async-coroutine-test.c++
jtolio: agreed, fixed

c++/src/kj/async.h
jtolio: fixed

c++/src/kj/async-coroutine-test.c++
jtolio: yep! fixed

c++/src/kj/async.h
jtolio: this makes sense now that I think about it, but TIL you can't co_await in a catch!

c++/src/kj/async-inl.h
jtolio: fixed

jtolio:
Great! this is great feedback! I just reracked what turned into 11 commits back into 4, with this feedback incorporated.

c++/src/kj/async.h
```diff
@@ -695,6 +695,39 @@ auto coCapture(Functor&& f) {
...
+  // This interface describes the current invocation, relative to the coroutine's last resumption.
+  // It borrows the coroutine frame and must not outlive the coroutine's returned promise.
+
+public:
+
+  // Coroutines might end in one of three ways - running successfully to the end, throwing
+  // an exception, or getting destructed while parked.
+  // isCanceling can be used when a coroutine is being destructed to determine if the cause
+  // of the destruction is the third case - getting destructed while parked. Note that this method
+  // is only useful during destruction and not after.
```
harrishancock: This is a good point: if someone stores this somewhere that outlives the coroutine, they'll see a UaF.

That risk is almost certainly not worth designing anything to guard against, but could be worth tightening this comment to say it's actually much less than useful after destruction, but actually UB.

Separately, we could mention that this function always returns false before destruction.

c++/src/kj/async.h
```diff
@@ -695,6 +695,97 @@ auto coCapture(Functor&& f) {
...
```
harrishancock: Nit: the capnproto repo still puts all doc comments below what they document

harrishancock (approved):

c++/src/kj/async.h
jtolio: fixed

c++/src/kj/async.h
jtolio: good call, fixed

jtolio:
https://gist.github.com/jtolio/1e89aaaeaa9c0bcc004977701e073a29 is updated

PR #2793, opened 2026-08-24
Experimental assert_arrayptr_counters mode

mikea:
Changes `kj::Array` to keep track of its `ArrayPtrs` when `--//src/kj:assert_arrayptr_counters=True` is set (`KJ_ASSERT_ARRAYPTR_COUNTERS`). Asserts there are no active pointers when array goes away.

Currently this flag is off, but the intention is to work towards enabling it in debug mode.

This PR doesn't contain violation changes, but fixes compilation in this new mode related to the fact that `ArrayPtr` is no longer mem-copiable and its `sizeof` changes.

My experiment show that the scope of changes to clean up everything is reasonable.

### DIFF
--- a/c++/src/capnp/dynamic.h
+++ b/c++/src/capnp/dynamic.h
@@ -717,8 +717,8 @@ private:
     mutable DynamicCapability::Client capabilityValue;
     // Declared mutable because `Client`s normally cannot be const.
 
-    // Warning:  Copy/move constructors assume all these types are trivially copyable except
-    //   Capability.
+    // Copy/move constructors explicitly construct the active member because some reader types
+    // are non-trivial when KJ_ASSERT_ARRAYPTR_COUNTERS is enabled.
   };
 
   template <typename T, Kind kind = kind<T>()> struct AsImpl;
--- a/c++/src/kj/array.h
+++ b/c++/src/kj/array.h
@@ -122,6 +122,9 @@ public:
 // Array
 
 template <typename T>
+void swp(Array<T>& a, Array<T>& b) noexcept;
+
+template <typename T>
 class Array {
   // An owned array which will automatically be disposed of (using an ArrayDisposer) in the
   // destructor.  Can be moved, but not copied.  Much like Own<T>, but for arrays rather than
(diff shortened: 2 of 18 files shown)


c++/src/kj/common.h
```diff
@@ -2529,8 +2538,19 @@ struct Mapper<Maybe<T>> {
 //
 // So common that we put it in common.h rather than array.h.
 
+}  // namespace kj
+
+#if KJ_ASSERT_ARRAYPTR_COUNTERS
+#include "atomic.h"
```
mikea, line 2550: it will take a lot of work to make atomic.h independent from declarations above.

c++/src/kj/array.h
```diff
@@ -131,32 +131,45 @@ class Array {
...
 
   KJ_DISALLOW_COPY(Array);
-  inline ~Array() noexcept(false) { dispose(); }
-
-  inline operator ArrayPtr<T>() KJ_LIFETIMEBOUND {
-    return ArrayPtr<T>(ptr, size_);
+  inline ~Array() noexcept(false) {
+    dispose();
+#if KJ_ASSERT_ARRAYPTR_COUNTERS
+    delete counter;
```
harrishancock: We'll leak if `dispose()` throws.

c++/src/kj/array.h
```diff
@@ -236,6 +254,11 @@ class Array {
     if (disposer == nullptr) return nullptr;
     Array<PropagateConst<T, char>> result(
         reinterpret_cast<PropagateConst<T, char>*>(ptr), size_, *disposer);
+#if KJ_ASSERT_ARRAYPTR_COUNTERS
+    delete result.counter;
```
harrishancock: This seems kinda hacky. Could it be improved with a new (possibly-private) Array constructor which initializes its `counter` like the move constructors do?

c++/src/kj/common.h
```diff
@@ -2617,10 +2683,14 @@ class ArrayPtr: public DisallowConstCopyIfNotConst<T> {
   }
 
   inline operator ArrayPtr<const T>() const {
+#if KJ_ASSERT_ARRAYPTR_COUNTERS
```
harrishancock: I wonder if we could get rid of some of these `#if` conditions if we added `ArrayPtrCounterTracker` as a `[[no_unique_address]]` data member.

c++/src/capnp/dynamic.h
```diff
@@ -717,8 +717,8 @@ class DynamicValue::Reader {
     mutable DynamicCapability::Client capabilityValue;
     // Declared mutable because `Client`s normally cannot be const.
 
-    // Warning:  Copy/move constructors assume all these types are trivially copyable except
-    //   Capability.
+    // Copy/move constructors explicitly construct the active member because some reader types
+    // are non-trivial when KJ_ASSERT_ARRAYPTR_COUNTERS is enabled.
```
harrishancock, line 721: Do you know if the changes in this commit have any impact on performance?

harrishancock (approved):

c++/src/capnp/dynamic.h
mikea: nothing really measurable on http-over-canpn bench: https://gist.github.com/mikea/da83a4565a9811b7e889d7247f1d34c2

c++/src/kj/array.h
mikea: I added `moveCounterFrom` instead that I can use with ifdefs.

PR #2795, opened 2026-08-25
KJ_EXPECT uaf fix

mikea:
The pattern in test.h succesfully crashes under ASAN. We do this quite a lot in our tests (e.g. io-test.c++) so it is worth fixing

uaf: https://gist.github.com/mikea/c2723d17174e9fff62074ce47051878a

pattern example: https://github.com/capnproto/capnproto/blob/v2/c%2B%2B/src/kj/io-test.c%2B%2B#L171

courtesy of experimental array ptr tracking

### DIFF
--- a/c++/src/kj/test.h
+++ b/c++/src/kj/test.h
@@ -80,19 +80,32 @@ private:
   } KJ_UNIQUE_NAME(testCase); \
   void KJ_UNIQUE_NAME(TestCase)::run()
 
+// Check and log the condition inside one function call so that temporaries backing captured
+// assertion operands remain alive until after any failure message has been rendered.
 #if KJ_MSVC_TRADITIONAL_CPP
 #define KJ_INDIRECT_EXPAND(m, vargs) m vargs
 #define KJ_FAIL_EXPECT(...) \
   KJ_INDIRECT_EXPAND(KJ_LOG, (ERROR , __VA_ARGS__));
 #define KJ_EXPECT(cond, ...) \
-  if (auto _kjCondition = ::kj::_::MAGIC_ASSERT << cond); \
-  else KJ_INDIRECT_EXPAND(KJ_FAIL_EXPECT, ("failed: expected " #cond , _kjCondition, __VA_ARGS__))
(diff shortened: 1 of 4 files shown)


c++/src/kj/test.h
```diff
@@ -80,19 +80,32 @@ class TestCase {
...
+// Check and log the condition inside one function call so that temporaries backing captured
+// assertion operands remain alive until after any failure message has been rendered.
 #if KJ_MSVC_TRADITIONAL_CPP
 #define KJ_INDIRECT_EXPAND(m, vargs) m vargs
 #define KJ_FAIL_EXPECT(...) \
   KJ_INDIRECT_EXPAND(KJ_LOG, (ERROR , __VA_ARGS__));
 #define KJ_EXPECT(cond, ...) \
-  if (auto _kjCondition = ::kj::_::MAGIC_ASSERT << cond); \
-  else KJ_INDIRECT_EXPAND(KJ_FAIL_EXPECT, ("failed: expected " #cond , _kjCondition, __VA_ARGS__))
+  do { \
```
harrishancock, line 90: I don't understand the difference in lifetimes between these two formulations. I would expect them to be equivalent.

harrishancock (approved):

c++/src/kj/test.h
mikea: Previously the macro was expanded to something like:

```
 if (auto condition =                                                                                                                                                                   
           MAGIC_ASSERT << heapString("foo").asBytes() == "bar"_kjb);                                                                                                                     
   else                                                                                                                                                                                   
     log(condition); 
```

temporary destruction happenes at the end of expression (if condition)

Now it looks like this:

```
   [&](auto&& condition) {                                                                                                                                                                
     if (!condition) {                                                                                                                                                                    
       log(condition);                                                                                                                                                                    
     }                                                                                                                                                                                    
   }(MAGIC_ASSERT << heapString("foo").asBytes() == "bar"_kjb);                                                                                                                           
```

and end of expression is _after_ full lambda was executed.

PR #2796, opened 2026-08-25
simple array ptr counter tracking fixes

mikea:
None seems to be a real problem, just tightening lifetimes

### DIFF
--- a/c++/src/kj/filesystem.c++
+++ b/c++/src/kj/filesystem.c++
@@ -352,6 +352,9 @@ Path Path::evalImpl(Vector<String>&& parts, StringPtr path) {
 Path Path::evalWin32Impl(Vector<String>&& parts, StringPtr path, bool fromApi) {
   // Convert all forward slashes to backslashes.
   String ownPath;
+  // `path` can be redirected into `ownPath`. Clear it before `ownPath` is destroyed, including
+  // when path validation throws.
+  KJ_DEFER(path = nullptr);
   if (!fromApi && path.findFirst('/') != kj::none) {
     ownPath = heapString(path);
     for (char& c: ownPath) {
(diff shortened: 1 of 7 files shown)


c++/src/kj/filesystem.c++
```diff
@@ -1621,9 +1620,8 @@ private:
 
     Maybe<EntryImpl&> openEntry(String&& name, WriteMode mode) {
       if (has(mode, WriteMode::CREATE)) {
-        EntryImpl entry(kj::mv(name));
```
harrishancock, line 1624: I don't understand how this used to work. I don't see an `EntryImpl` constructor which accepts a `String`.

Meanwhile, I don't understand how it works after this change. Are we returning an `EntryImpl&` with an uninitialized `node` OneOf?

mikea: > I don't understand how this used to work. I don't see an `EntryImpl` constructor which accepts a `String`.

This is old code, there used to be a constructor taking `String&&`.

> Meanwhile, I don't understand how it works after this change. Are we returning an `EntryImpl&` with an uninitialized `node` OneOf?

As far as I can tell nothing has changed - I only dropped String field from EntryImpl. OneOf default constructor initializes tag(0), so it should be a FileNode() here?

harrishancock: > This is old code, there used to be a constructor taking String&&.

Oh, I see now!

> OneOf default constructor initializes tag(0), so it should be a FileNode() here?

`tag` is 1-based, 0 means uninitialized: https://github.com/capnproto/capnproto/blob/v2/c%2B%2B/src/kj/one-of.h#L496

But maybe it's correct as-written? At any rate, this PR doesn't change its behavior.

harrishancock (approved):

PR #2797, opened 2026-08-26
KJ: Clear BlockedPumpTo state after output errors

aaaazqh:
Fixes #2579.

When a downstream write or optimized sub-pump failed, `BlockedPumpTo` released its canceler and rejected the outer pump promise but left `AsyncPipe::state` pointing at the failed pump state. A subsequent write could therefore be forwarded to an output stream that was already unusable.

This change adds `BlockedPumpTo`-specific exception handlers that:

- release the canceler before propagating the error;
- clear the pipe state before invoking promise rejection observers;
- preserve the original downstream exception for both operations;
- cover scalar writes, gather-write partial/exact/full paths, and `tryPumpFrom()`.

The regression tests verify that the failed output receives no subsequent writes, both promise paths retain the original exception, and the pipe remains usable with a new reader.

Validation:

- Debug full CTest: 5/5 passed
- Debug regression tests: 100 consecutive runs passed
- Release full build and regression tests passed
- ASan + UBSan regression tests: 100 consecutive runs passed
- TSan regression tests: 100 consecutive runs passed

### DIFF
--- a/c++/src/kj/async-io-test.c++
+++ b/c++/src/kj/async-io-test.c++
@@ -1602,6 +1602,146 @@ KJ_TEST("Userland pipe pumpTo") {
   KJ_EXPECT(pumpPromise.wait(ws) == 3);
 }
 
+class ControlledErrorOutput final: public AsyncOutputStream {
+public:
+  Promise<void> write(ArrayPtr<const byte>) override {
+    ++writeCount;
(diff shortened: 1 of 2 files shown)


kentonv:
Sorry, but we are not looking for external contributors to point agents at our issues and submit the results. It's much less work for us to have our own agent fix the issue, than to perform the careful code review needed for external contributions.

PR #2798, opened 2026-08-26
KJ_DROP_TEMPS macro

mikea:
When re-assigning arrays, it is often common to compute new arrays based on the views of the old ones. Due to C++ rules these temporaries live to the end of the expression and outlive the original object they referred to.

Rather than introduce temporaries:

    auto newValue = computeNewValue(value);
    value = kj::mv(newValue)

KJ_DROP_TEMPS let's you write it in single line:

   value = KJ_DROP_TEMPS(computeNewValue(value));

This also makes lifetime intention more obvious.

### DIFF
--- a/c++/src/kj/async-io.c++
+++ b/c++/src/kj/async-io.c++
@@ -2617,9 +2617,7 @@ uint64_t AsyncTee::Buffer::consume(ArrayPtr<byte>& readBuffer, size_t& minBytes)
     if (amount == bytes.size()) {
       bufferList.pop_front();
     } else {
-      // Finish reading from `bytes` before replacing its backing array.
-      auto remainder = heapArray(bytes.slice(amount, bytes.size()));
-      bytes = kj::mv(remainder);
+      bytes = KJ_DROP_TEMPS(heapArray(bytes.slice(amount, bytes.size())));
       break;
     }
   }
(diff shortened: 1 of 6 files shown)


harrishancock (approved):
TBH, I'm not sure this earns its keep, but I don't feel too strongly. If this pattern is prevalent across our other codebases, then I could see it being useful.

I'd suggest `KJ_IILE` (immediately-invoked lambda expression), though. More obscure, but it's obscure anyway, and naming it IILE would be a bit more honest about its disadvantages (e.g., cannot `KJ_IILE(co_await x)`).

mikea:
we decided to table this for now and introduce temps if needed.

kentonv:
FWIW I have long intended to introduce `KJ_IIFE(code)` as a more readable alternative to `[&](){code;}()`. The latter is particularly tricky to read when the code is a large block since the `()` at the end is easy to miss and completely changes the meaning.

I've never seen the acronym IILE before -- I think IIFE is more recognizable.

harrishancock:
> I've never seen the acronym IILE before -- I think IIFE is more recognizable.

I've seen it before, but always used IIFE before. Claude insisted it's what cool C++ kids use, and IIFE is strictly for the JS people, so I tried it on for size.

PR #2803, opened 2026-08-27
Fix ASAN fiber stack bounds

erikcorry:
Pass the low address of the fiber stack to ASAN rather than the address of the metadata stored at its high end. The incorrect range could overlap unrelated mappings, causing ASAN to unpoison memory outside the fiber stack during context switches.


mikea (approved):

kentonv:
Does this make #2754 moot?

logan-gatlin:
> Does this make #2754 moot?

Yeah, closing mine

PR #2805, opened 2026-08-28
the last of the assert ptr counters fixes

mikea:
Enables assert ptr counters in dbg mode (no effect on downstream)

### DIFF
--- a/c++/src/capnp/compat/json-rpc.c++
+++ b/c++/src/capnp/compat/json-rpc.c++
@@ -324,10 +324,9 @@ JsonRpc::ContentLengthTransport::ContentLengthTransport(kj::AsyncIoStream& strea
 JsonRpc::ContentLengthTransport::~ContentLengthTransport() noexcept(false) {}
 
 kj::Promise<void> JsonRpc::ContentLengthTransport::send(kj::StringPtr text) {
-  auto headers = kj::str("Content-Length: ", text.size(), "\r\n\r\n");
-  parts[0] = headers.asBytes();
-  parts[1] = text.asBytes();
-  return stream.write(parts).attach(kj::mv(headers));
+  auto header = kj::str("Content-Length: ", text.size(), "\r\n\r\n");
+  co_await stream.write(header.asBytes());
+  co_await stream.write(text.asBytes());
 }
 
 kj::Promise<kj::String> JsonRpc::ContentLengthTransport::receive() {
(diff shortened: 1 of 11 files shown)


c++/src/capnp/compat/json-rpc.c++
```diff
@@ -324,10 +324,9 @@ JsonRpc::ContentLengthTransport::ContentLengthTransport(kj::AsyncIoStream& strea
 JsonRpc::ContentLengthTransport::~ContentLengthTransport() noexcept(false) {}
 
 kj::Promise<void> JsonRpc::ContentLengthTransport::send(kj::StringPtr text) {
-  auto headers = kj::str("Content-Length: ", text.size(), "\r\n\r\n");
-  parts[0] = headers.asBytes();
-  parts[1] = text.asBytes();
-  return stream.write(parts).attach(kj::mv(headers));
+  auto header = kj::str("Content-Length: ", text.size(), "\r\n\r\n");
+  co_await stream.write(header.asBytes());
+  co_await stream.write(text.asBytes());
```
harrishancock: Non-blocking nit: I think this could technically be a behavior change, if the previous `write` happened to be optimized down all the way to an atomic `writev()`.

I'm not sure that makes a difference, but it might be easy enough to avoid with a coroutine frame-allocated temporary ArrayPtr.

c++/src/capnp/compat/json.c++
```diff
@@ -1209,10 +1204,9 @@ private:
     // For `NORMAL` and `FLATTENED`, the index of the field in schema.getFields().
 
     uint prefixLength;
-    kj::String ownName;
   };
 
-  kj::HashMap<kj::StringPtr, FieldNameInfo> fieldsByName;
+  kj::HashMap<kj::String, FieldNameInfo> fieldsByName;
```
harrishancock, line 1209: Non-blocking nit: Are these changes here only needed because HashMap's entries have the key stored before the value? I wonder if flipping their order would fix the issue, if so.

Generally I'm on board with making keys own their own memory, but our JSON parser is already known to be a bit slow.

harrishancock (approved):

c++/src/capnp/compat/json.c++
mikea: no discernible diff: https://gist.github.com/mikea/d310126c213fa38a006a1279c9d1d595

As far as I can tell I simply moved ownership from value to the key in most cases.

c++/src/capnp/compat/json-rpc.c++
mikea: done.

PR #2808, opened 2026-08-31
[build] Start using clang 23

fhanau:
Fix clang-23 warnings
Clang 23 warns about unused templates (including static templates in headers) and variables that are never being read.

[build] Start using clang 23
- Update to clang-23. The minimum tested version moves to 19.
- Test with gcc-16, which is now available for Ubuntu 26.04
- Use __type_pack_element when available – this is supported by clang 20 and GCC 14 and can be used to implement TypeByIndex. Having a non-recursive implementation for this should speed up compiling tuples.
- Update coroutine alloc size to value expected with clang-23.

type_pack_element has been used in V8 for a while, we can benefit from it too.

### DIFF
--- a/c++/src/kj/async-coroutine-alloc-test.c++
+++ b/c++/src/kj/async-coroutine-alloc-test.c++
@@ -177,7 +177,7 @@ KJ_TEST("Coroutine Frame sizes") {
     DebugCoroutineAllocator allocator;
     auto promise = immediateCoroutine(allocator);
     KJ_EXPECT(allocator.totalAllocCount == 1);
-    KJ_EXPECT_CORO_SIZE(allocator.totalAllocSize == 152);
+    KJ_EXPECT_CORO_SIZE(allocator.totalAllocSize == 160);
   }
 
   {
--- a/.github/workflows/quick-test.yml
+++ b/.github/workflows/quick-test.yml
@@ -28,7 +28,7 @@ env:
 
 jobs:
   Linux-musl:
-    runs-on: ubuntu-24.04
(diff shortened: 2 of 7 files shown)


c++/src/kj/async-coroutine-alloc-test.c++
```diff
@@ -177,7 +177,7 @@ KJ_TEST("Coroutine Frame sizes") {
     DebugCoroutineAllocator allocator;
     auto promise = immediateCoroutine(allocator);
     KJ_EXPECT(allocator.totalAllocCount == 1);
-    KJ_EXPECT_CORO_SIZE(allocator.totalAllocSize == 152);
+    KJ_EXPECT_CORO_SIZE(allocator.totalAllocSize == 160);
```
mikea, line 180: Did it increase in new compiler? any idea what's up? there seems to be an extra pointer, I wonder what is it (not that we have much choice).

mikea (approved):

c++/src/kj/async-coroutine-alloc-test.c++
fhanau: Locally, I get 160B with clang-23 under --config=opt (as used in CI), but 152B when turning off thinLTO. clang-22 results in 152B either way. The change appears to have happened between `clang-llvmorg-23-init-484-gf646b915` and `clang-llvmorg-23-init-19482-g53d18800`. thinLTO resulting in a larger size is interesting, but I doubt that tracking down the specific commit here will be worth it unless there's more regressions.

PR #2809, opened 2026-09-03
async-io: fix accidental blocking in SocketAddress::lookupHost()

jclee:
To perform asynchronous address lookup, SocketAddress::lookupHost() runs a blocking getaddrinfo() API call in a separate thread.  However, following a refactoring to use cross-thread fulfillers in a4ae5896, the kj::Thread was no longer moved into the return value, so it was destroyed at the end of the function.  Because kj::Thread's destructor joins the thread, calls to lookupHost() have effectively been synchronous since then.

This commit attaches the kj::Thread to the returned promise, which should allow lookup to proceed without blocking the caller.

### DIFF
--- a/c++/src/kj/CMakeLists.txt
+++ b/c++/src/kj/CMakeLists.txt
@@ -256,6 +256,17 @@ endif()
 
 # Tests ========================================================================
 
+# TODO(someday): maybe also build/run these tests?:
+#   exception-override-symbolizer-test.c++
+#   filesystem-disk-generic-test.c++
+#   filesystem-disk-old-kernel-test.c++
+#   compat/brotli-test.c++
+#   compat/readiness-io-test.c++
+#   compat/http-socketpair-test.c++
+
+# TODO(someday): add a way to detect when a test file is present but missing from
+# cmake build instructions?
+
 if(BUILD_TESTING)
   add_executable(kj-tests
     common-test.c++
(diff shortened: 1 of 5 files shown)


jclee:
The code prior to the refactor also detached the thread on cancellation... not sure if that might be preferable.  Perhaps something like?:

`paf.promise.attach(kj::defer([thread = kj::mv(thread)]() { thread.detach(); }))`

The current code will join the thread on cancellation, but I figured it wasn't a huge concern, considering that we'd been joining the thread before the function even returned.

kentonv (approved):
wow... ugh

kentonv:
Hmm, regarding detaching on cancellation... that sounds like a good idea, but I also wonder if we're doing that, maybe it makes sense to just _always_ detach?

jclee:
Sure, I can update it to always detach.  In that case, if the caller cancels, there's nowhere for an exception to go, if one is thrown, but maybe that's not a big deal, seeing as how it didn't matter in the original code, either.

I guess one advantage of only detaching on cancellation is that if the promise resolves, you know the thread has also ended, whereas if you always detach, you don't know if the thread might still be cleaning up.  Don't know if there are situations where someone might care about that, like if trying to pare down to a single thread before a fork() or setenv() or something.  But I suppose there are other ways to detect outstanding threads, too, and the simplicity of always detaching is appealing.

jclee:
(An automated reviewer pointed out several potential leaks in the win32 threading code.  Added a commit to fix.)

PR #2815, opened 2026-09-08
Keep attached FDs alive through capability wrappers

zeebo:
PR #2755 attempted to keep FDs alive while messages were queued by retaining the `ClientHook` on which `getFd()` was called. This is insufficient for wrappers such as membranes, which can release their inner hook during revocation even while the wrapper remains alive.

`ClientHook::getFd()` now returns an `FdRef` containing the descriptor and its keep-alive reference. Forwarding hooks preserve the underlying `FdRef` rather than substituting a reference to themselves.

For revocable local servers, these references hold FD leases. RPC revocation remains immediate, while the new `revokeAndWait()` and `whenServerDestroyable()` methods let callers wait for outstanding leases to be released before destroying the server.

The two-party transport also explicitly releases FD arrays and their attachments after writes complete, fail, or are canceled, and when queued messages are abandoned. This prevents RPC bookkeeping from retaining FD leases merely because it still holds references to the message objects.


c++/src/capnp/capability.h
```diff
@@ -609,8 +612,17 @@ class RevocableServer {
 
   void revoke();
   void revoke(kj::Exception&& reason);
-  // Revokes the capability immediately, rather than waiting for the destructor. This can also
-  // be used to specify a custom exception to use when revoking.
+  // Revokes the capability immediately. The overload can specify the exception thrown by clients.
+  // This does not wait for file descriptors obtained from the capability to finish sending.
+
+  kj::Promise<void> revokeAndWait();
+  kj::Promise<void> revokeAndWait(kj::Exception&& reason);
```
kentonv, line 619: Hmm, this smells like trouble. RevocableServer is commonly (always?) used in cases where we need to be able to revoke the cap _synchronously_ when it goes out of scope. If failing to wait is a correctness bug, then essentially all uses of RevocableServer are broken.

Two options:
1. Disallow use of getFd() from a capability created by RevocableServer.
2. Bite the bullet and change Capability::Server::getFd() to the new interface, too.

PR #2817, opened 2026-09-14
Moving StringPtr and ArrayPtr should clear the source

mikea:
I was always puzzled by current behavior. Standard behavior makes cleaning up array pointers easier.


harrishancock (approved):
There might be some downstream effects to shake out, but I've wished for this for a long time.

I think the original logic was that these types behave like `T*`, but in practice I think the current behavior is more surprising than not.

mikea:
> There might be some downstream effects to shake out, but I've wished for this for a long time.

As far as I can tell nothing breaks with this change. I'll apply it and will update capnp everywhere.

PR #2818, opened 2026-09-15
kj: implement DatagramPortImpl on Windows

ThomasRubini:
This change implements the DatagramPortImpl UDP listener on Windows.
Tested against workerd tests ([CI](https://github.com/cloudflare/workerd/actions/runs/34991930549/job/104459113039) [test code](https://github.com/cloudflare/workerd/pull/7375/changes#diff-fcd88641e7b6cc50d47f9e968c04efebe15af596aad0a4aa069cf7345c9ff272))


c++/src/kj/async-io-win32.c++
```diff
@@ -1183,6 +1240,181 @@ private:
...
+}
+
+Promise<size_t> DatagramPortImpl::send(
+    Array<WSABUF> buffers, NetworkAddress& destination) {
+  // The operation state owns the send parameters through completion.
+  auto& addr = downcast<NetworkAddressImpl>(destination).chooseOneAddress();
+  auto state = heap<SendState>(kj::mv(buffers), addr);
+  auto op = observer->newOperation(0);
+
+  if (WSASendTo(fd, state->buffers.begin(), state->buffers.size(), NULL, 0,
```
danlapid, line 1269: **SUGGESTION · Code quality — Validate WinSock size conversions**

On 64-bit Windows, `state->buffers.size()` is a 64-bit `size_t`, but the buffer-count parameter of `WSASendTo()` is a 32-bit `DWORD`. The `WSABUF::len` assignments in the two send overloads above likewise narrow `size_t` to 32-bit `ULONG`.

Could we check that the buffer lengths and scatter/gather count fit the destination types before explicitly casting? This would avoid narrowing warnings and prevent oversized inputs from silently wrapping before WinSock can reject them.

Finding from `ai-review`, rerun against this PR’s `v2` base.

danlapid (approved):

PR #2823, opened 2026-09-17
Introduce kj::StickyFlag (newAdaptedPromise approach)- #2830

jtolio:
Note: There are two otherwise identical implementations of this PR. This one (newAdaptedPromise), and the other one https://github.com/capnproto/capnproto/pull/2830 (newPromiseAndFulfiller/ForkedPromise)

This event is a simple interface for one-shot latched signals, or events, which is fairly common in practice.

Instead of having to dance around promises and fulfillers, this type has three easy methods - bool signal(), bool reject(kj::Exception), and kj::Promise&lt;void> whenSignaled(). These behave naturally and without surprise. Further, destruction will reject an otherwise unresolved StickyFlag.

This type won't save a ton of lines of code, but will allow code to be more self descriptive about its goals.


jtolio:
Example PRs that use this change (from before the rename to StickyFlag, will fix once merged): 
 * https://github.com/cloudflare/workerd/pull/7404
 * https://github.com/capnproto/capnproto/pull/2825
 * https://gitlab.cfdata.org/cloudflare/ew/edgeworker/-/merge_requests/14941 (sorry to external reviewers who can't see this)

kentonv:
I don't understand the value here vs. `kj::newPromiseAndFulfiller<void>()`.

jtolio:
> I don't understand the value here vs. kj::newPromiseAndFulfiller&lt;void>().

If there is value to this, it's almost entirely legibility. This gains very little in terms of LoC reduction and so forth. That said, with this as an example commit, I do find it a bit more readable of a pattern: https://github.com/capnproto/capnproto/pull/2825/changes/8a5e9a711d409609e679ccbe0094cf099fdb9389

jtolio:
Another question might be - why isn't this implemented with kj::newPromiseAndFulfiller&lt;void>()? Maybe that's what you were asking. It certainly could be. I had to dance around deciding to lazily or eagerly construct forked promises for later branch additions, etc., and decided that was annoying. The kjtour docs warning warned me towards newAdaptedPromise. So I did that.

jtolio:
Renamed this by request to kj::StickyFlag from kj::AsyncEvent

jtolio:
> Another question might be - why isn't this implemented with kj::newPromiseAndFulfiller&lt;void>()?

I implemented another version this way: https://github.com/capnproto/capnproto/pull/2830

harrishancock:
> If there is value to this, it's almost entirely legibility.

I think it's valuable in terms of writability and teachability, too. I remember teaching a coworker the PaF+ForkedPromise pattern for once-notification. They got it, but I had to re-derive the pattern, convince myself I had it correct, then convince them I had it correct. I would have loved StickyFlag.

I like the much simpler implementation in https://github.com/capnproto/capnproto/pull/2830, approved with a comment.

jtolio:
Closing in favor of https://github.com/capnproto/capnproto/pull/2830

PR #2828, opened 2026-09-17
Rc::project() and Arc::project()

mikea:
project() method can be used to obtain child values that share the same reference count as the root object.

Common example would be to retain field value pointer while claiming the ownership of the original object.


harrishancock (approved):
It looks nice! Astra had a few good suggestions, which I've put on this branch for you to cherry-pick, if you like: https://github.com/capnproto/capnproto/commits/harris/2026-09-18-capnproto-2828-fixes/

- `disown()` should throw if it is a non-identity projected Rc / Arc
- `project()`'s callback should be allowed to safely destroy the source Rc / Arc
- some comment suggestions from me

I also wanted to suggest an rvalue overload, though I don't know if we actually have a use case? Right now, if I have an `Arc<T>` and want to replace it with `Arc<SubT>`, I have no choice but to pay the cost of an `addRef()`. An rvalue overload could avoid that. In fact, supporting _only_ an rvalue overload could make it so you have to explicitly opt into the cost if you want to retain the source handle: `source.clone().project(...)`. But this is all just bikeshedding at this point.

mikea:
@harrishancock thanks. I have made project() method rvalue only. This also addresses the concern with nulling original Rc (it doesn't matter anymore). Added disown check and applied your comments. Landing this to see how I can use it.

c++/src/kj/refcount.h
```diff
@@ -314,11 +321,46 @@ class Rc {
...
+    // must return a reference into an object whose lifetime is covered by the original ownership
+    // claim. Returning the referent itself is supported; the result is then not a projection.
+    // A WeakRc obtained from the result points
+    // to the projected object, but expires based on the lifetime of the original object.
+    using Result = decltype(kj::fwd<Func>(func)(*ptr));
+    static_assert(isLvalueReference<Result>(), "Rc::project() callback must return a reference");
+    using U = typename _::RemoveReference<Result>::Type;
+
+    KJ_IREQUIRE(ptr != nullptr, "null Rc<> projection");
+    auto projected = &kj::fwd<Func>(func)(*ptr);
```
harrishancock: I think you still need to move *this into the stack in order for this to be safe. If the callback has a reference to *this and nullifies it, we have UB.

That's pretty unlikely, but guarding against it should be free.

mikea: fixed

PR #2829, opened 2026-09-18
KJ_IF_SOME support for Rc and Arc

mikea:
Because C++ Rc and Arc can contain null values. I don't fully like it, but we use this fact in couple of places.

While I am contemplating what to do with it, this PR makes the null checking code look idiomatic both for Maybe<T&> and Rc<T>


harrishancock (approved):
I don't like that we do this, either, but I know existing code depends on Rc and Arc being able to be null.

Consider adding documentation to KJ_IF_SOME explaining that you can do this, but new code should prefer to use `Maybe<Rc<T>>`, etc.

mikea:
I decided not to land this one and to convert the code to Maybe instead when I see it.

PR #2830, opened 2026-09-18
Introduce kj::StickyFlag (newPromiseAndFulfiller/ForkedPromise approach)

jtolio:
Note: There are two otherwise identical implementations of this PR. This one (newPromiseAndFulfiller/ForkedPromise), and the other one https://github.com/capnproto/capnproto/pull/2823 (newAdaptedPromise)

This event is a simple interface for one-shot latched signals, or events, which is fairly common in practice.

Instead of having to dance around promises and fulfillers, this type has three easy methods - bool signal(), bool reject(kj::Exception), and kj::Promise&lt;void> whenSignaled(). These behave naturally and without surprise. Further, destruction will reject an otherwise unresolved StickyFlag.

This type won't save a ton of lines of code, but will allow code to be more self descriptive about its goals.


jtolio:
Most discussion is on https://github.com/capnproto/capnproto/pull/2823

c++/src/kj/sticky-flag.c++
```diff
@@ -0,0 +1,89 @@
...
+  Maybe<Exception> exception;
+  Maybe<Own<PromiseFulfiller<void>>> fulfiller;
+  Maybe<ForkedPromise<void>> forked;
+};
+
+StickyFlag::StickyFlag(): state(heap<State>()) {}
+
+StickyFlag::~StickyFlag() noexcept(false) {
+  // Will only reject if not already resolved.
+  reject(KJ_EXCEPTION(FAILED, "StickyFlag destroyed before it was signaled or rejected"));
```
harrishancock: Let's check `ready` first, otherwise this will allocate `kj::Exception`'s implementation unconditionally.

```suggestion
  if (!ready) {
    reject(KJ_EXCEPTION(FAILED, "StickyFlag destroyed before it was signaled or rejected"));
  }
```

harrishancock (approved):

c++/src/kj/sticky-flag.c++
jtolio: fixed

harrishancock (approved):

